PR #165, opened 2022-02-21
RFC: Visual regression testing

sarayourfriend:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/rfcs/20220217-visual_regression_testing.md
+++ b/rfcs/20220217-visual_regression_testing.md
@@ -6,7 +6,6 @@ A proposal to integrate visual regression testing into the Openverse frontend's 
 
 - [x] @obulat
 - [x] @AetherUnbound
-- [ ] Approver
 
 ## Milestone
 
--- a/rfcs/20220217-visual_regression_testing.md
+++ b/rfcs/20220217-visual_regression_testing.md
@@ -5,7 +5,8 @@ A proposal to integrate visual regression testing into the Openverse frontend's 
 ## Reviewers
 
 - [x] @obulat
-- [ ] Approver (TBD)
+- [x] @AetherUnbound
+- [ ] Approver
 
 ## Milestone
 
--- a/rfcs/20220217-visual_regression_testing.md
+++ b/rfcs/20220217-visual_regression_testing.md
@@ -7,7 +7,13 @@ A proposal to integrate visual regression testing into the Openverse frontend's 
 - [ ] Approver (TBD)
 - [ ] Approver (TBD)
 
-**Milestone:** TBD
+## Milestone
+
+TBD
+
+## Exploratory code
+
+https://github.com/WordPress/openverse-frontend/pull/899
 
 ## Background
 
--- a/rfcs/20220217-visual_regression_testing.md
+++ b/rfcs/20220217-visual_regression_testing.md
@@ -164,6 +164,52 @@ Additionally, subsequent pushes to the parent PR that include changes to the sna
 
 Likewise, if we end up needing to run `pixelmatch` directly to generate the diff output, it'd be nice if we could just add those diff outputs as a comment in the parent PR whenever snapshot images are updated.
 
+## Test variants
+
+Most tests should be written against two parameters: language direction and breakpoint.
+
+For language direction, the best way to do this is to just test in the default locale which is LTR and to test in Arabic (`ar`) which is RTL.
+
+For breakpoints, it will be tedious to have to add the iteration for each test case, so I've written a utility that we can use to make this easier. This utility can also be used for our regular end-to-end tests:
+
+```ts
+import { test, PlaywrightTestArgs, TestInfo } from '@playwright/test'
+
+import { Breakpoints, SCREEN_SIZES } from '../../../src/constants/screens'
+
+export const testEachBreakpoint = (title: string, testCb: (breakpoint: Breakpoints, args: PlaywrightTestArgs,testInfo: TestInfo) => Promise<void>) => {
+  SCREEN_SIZES.forEach((screenWidth, breakpoint) => {
+    test.describe(`screen at breakpoint ${breakpoint} with width ${screenWidth}`, () => {
+      test(title, async ({ page, context, request }, testInfo) => {
+        await page.setViewportSize({ width: screenWidth, height: 700 })
+        await testCb(breakpoint, { page, context, request }, testInfo)
+      })
+    })
+  })
+}
+```
+
+It's then used in the test like so:
+
+```ts
+test.describe('homepage snapshots', () => {
+  // Repeat below for `rtl`
+  test.describe('ltr', () => {
+    test.beforeEach(async ({ page }) => {
+      await page.goto('/')
+    })
+
+    testEachBreakpoint('full page', async (breakpoint, { page }) => {
+      await deleteImageCarousel(page)
+
+      expect(await page.screenshot()).toMatchSnapshot({
+        name: `index-ltr-${breakpoint}`,
+      })
+    })
+  })
+})
+```
+
 ## Recommended tooling
 
 * Continue using Playwright to manipulate the browser and use it's `screenshot` function to generate screenshots.
@@ -203,7 +249,7 @@ Once the final version of this RFC is approved, a milestone with issues for each
         - Note: It may be necessary to add additional TypeScript typings in the root level `typings` folder for TypeScript to know that `expect()` includes the `toMatchImageSnapshot` function. Follow the existing examples extending `@types/nuxt` to extend the `@playwright/test` library's `Matchers` interface.
         - Additional note: As mentioned in the section about `jest-image-snapshot` above it may not be possible to get this library to play nicely with Playwright's specific `expect.extend` context. I'd say time box trying to get this to work to an hour or so and then just move on to the next thing. There's an additional task at the end of this list for writing the `pixelmatch` diff generator script in case we need it.
         - **Extra important note:** This library might not work with Playwright, in which case we'll just leave this out; there's a contingency plan for this case below.
-    - Write basic tests for the homepage; a full-page snapshot for `ltr` and `rtl` versions of the page should do just as a start and proof-of-concept.
+    - Write basic tests for the homepage; a full-page snapshot for `ltr` and `rtl` versions of the page using `testEachBreakpoint` should do just as a start and proof-of-concept.
         - Note: It will be necessary to remove the featured images from the page before taking the snapshot or the diff will fail 2/3 of the time due to them being different. The following code can be used to accomplish this:
         ```ts
         const deleteImageCarousel = async (page: Page) => {
--- a/rfcs/20220217-visual_regression_testing.md
+++ b/rfcs/20220217-visual_regression_testing.md
@@ -1,0 +1,220 @@
+# RFC: Visual regression testing
+
+A proposal to integrate visual regression testing into the Openverse frontend's automated testing suite.
+
+## Reviewers
+
+- [ ] Approver (TBD)
+- [ ] Approver (TBD)
+
+**Milestone:** TBD
+
+## Background
+
+### What is "visual regression testing"?
+
+Visual regression testing is a testing strategy that diffs two screenshots reports a failure if there are significant differences between the two.
+
+Playwright is capable of generating and diffing screenshots of the running application  via the [screenshot function](https://playwright.dev/docs/screenshots) available on the `Page` and `Locator` objects. These can be used with `toMatchSnapshot` like so:
+
+```ts
+expect(await page.screenshot()).toMatchSnapshot({
+  name: 'fullpage',
+})
+```
+
+Playwright uses the popular [pixelmatch](https://github.com/mapbox/pixelmatch) library to diff the screenshots. Example diffs [can be found here](https://github.com/mapbox/pixelmatch#example-output). Playwright on its own does not appear capable of outputting the visual-diff image. An alternative to this is covered below.
+
+### Rationale for this style of testing
+
+A lot of the recent regressions that we've experienced on the frontend _on critical paths_ have been visual style regressions. It's difficult, but not impossible, to write "sensible" unit tests against the applied styles. Testing Library offers the [`toHaveStyle`](https://github.com/testing-library/jest-dom#tohavestyle) matcher and we _could_ start utilizing this more, but it would require tedious manual changes to tests that _most of the time_ would just be 1-1 confirmations of the applied Tailwind classes. It also doesn't necessarily test components in composition very well and it would be tedious to write individual specific style tests for each of the scenarios any given widespread component is used in (`VButton`, for example, is used everywhere).
+
+Visual regression testing could be a way of accomplishing more robust tests without the tedious practice of hand-writing tests for specific styles.
+
+### Existing tooling
+
+We already utilize Playwright for our end-to-end tests. Provided we wrote the requisite tests to generate the screenshots, we could stop here and already have visual regression testing.
+
+However, we also need a way of easily reviewing the diffed images. Luckily, [GitHub already has this functionality built into it's PR diff previews](https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files#viewing-differences). This means when an existing screenshot is replaced we'd be able to review and approve this change as a normal part of our PR review process, just like we review code.
+
+### Alternative/additional tooling
+
+Everything in the previous section is basically the "bare minimum" of what we'd need to start utilizing visual regression testing. We could stop there and already be able to take advantage of this technology. However, there are additionally a whole world of visual regression testing tools. In this section I cover one popular external diff reviewing tool (which I ultimately recommend against, for now) and an enhanced image snapshot matcher.
+
+#### Percy
+
+[Percy](https://percy.io/) seems to be far and away the most powerful tool for reviewing image diffs. It is owned by BrowserStack. While BrowserStack has a generous open source sponsorship (that we should definitely eventually utilize to run e2e tests and even potentially generate snapshots on real mobile devices), they do not appear to extend this offering to Percy[^1].
+
+Percy also requires you to specifically upload the files through a GitHub action which would require managing additional secrets. It would also restrict reviewers to those with access to Percy which would unnecessarily exclude casual contributors.
+
+Percy definitely looks very cool and I would recommend creating a BrowserStack account and checking out the Percy demo project. It's fun to poke around in and the UI mostly works very nicely.
+
+However, if I'm being honest, when taking into account the fact that GitHub already has image diffing built into the PR review process, Percy seems like an unnecessary additional complication to the baseline of visual regression testing. Maybe it's something that we would come to realize the value of down the line, but it does not seem to me to be a requirement for the initial phase of visual regression testing.
+
+#### Alternative image snapshot libraries
+
+Playwrights's built in image snapshot-diffing capability works fine for the most part. In fact, it basically does everything you need short of one very useful feature: it throws away the actual generated diff.
+
+[Amex's `jest-image-snapshot`](https://github.com/americanexpress/jest-image-snapshot) fixes this. In addition to failing when snapshots do not match, it will also write the diff image that includes the "hot spots" where the visual differences exist. It also automatically generates the snapshot name, which Playwright's matcher doesn't seem to be capable of (this is usually a standard feature of Jest snapshot matchers otherwise).
+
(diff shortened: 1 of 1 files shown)


AetherUnbound:
I'm gonna un-assign Ronny and I from this PR :sweat_smile: I think some other frontend folks will likely have better context to say than us!

sarayourfriend:
Actually, funnily I would love @AetherUnbound and @rbadillap's input on the docker related stuff in this RFC.

AetherUnbound:
Ha, my bad! I'll take a look at it :slightly_smiling_face:

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,226 @@
...
+```
+3. Add two new package.json scripts:
+```json
+"test:visual-regression": "node ./bin/visual-regression.sh",
+"test:visual-regression:local": "playwright test -c ./test/visual-regression"
+```
+
+All of this together should successfully allow us to run the snapshot tests inside a docker container.
+
+For local debugging purposes, `pnpm test:visual-regression:local --debug` can be used to run the snapshot tests headed in a browser on the host OS. This is useful for debugging the playwright scripts before they actually take a screenshot. Just keep in mind that the final screenshots submitting in a PR need to come from the docker container.
```
obulat, line 149: Can we create the screenshots during the CI action on GitHub, instead of locally in a docker container?
I'm not very comfortable with Docker, and I'm sure many of the community contributors will not be, either. I think it is important not to make Docker installation required for contributing.

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Recommended tooling
+
+* Continue using Playwright to manipulate the browser and use it's `screenshot` function to generate screenshots.
+* Use `jest-image-snapshot` to save and diff screenshots.
+* Run snapshot tests inside a docker container to avoid cross-platform rendering differences.
+* Write a GitHub action to be run [once daily using the `on.schedule` feature](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule) to automatically remove snapshots from the repository history older than a month old.
+* Use GitHub's PR review UI to review visual diffs of snapshots.
+
+If agreeable with the team, I'd also like to take this opportunity to implement TypeScript linting and recommend that we switch to using TypeScript to write Playwright tests. Playwright supports it natively.
```
obulat, line 159: Considering that we've already added typescript parser to ESLint, we can omit this sentence (and replace 'Typescript setup' with 'Conversion of the existing tests to TS'.

obulat (approved):
I'm really excited for this addition.
This implementation only mentions creating a homepage visual regression test as a proof-of-concept. Would writing all search type search pages and single media item detail page be outside of this milestone/plan?

rfcs/20220217-visual_regression_testing.md
sarayourfriend: I guess we could :thinking: That would involve a much bigger process though. I suppose one thing we could do is something like this:

1. You submit a PR
2. GitHub action runs on it to determine if there are snapshot changes
3. If there are snapshot changes, the action creates a new branch based off the PR branch (`<pr-banch>__snapshot-update` or something) that then opens PR targeting your branch with the updated snapshots for review

I'd be happy to explore that :slightly_smiling_face: 

On the other hand, with the `visual-regression.js` script I'm proposing, the interaction with docker should be very minimal, literally just having it installed, you'd have to do nothing else for it to work.

If it really is a hurdle though the supplemental PR with the snapshot updates could do the trick though.

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,226 @@
...
+        const deleteImageCarousel = async (page: Page) => {
+            const element = await page.$('[data-testid="image-carousel"]')
+            await element.evaluate((node) => node.remove())
+            element.dispose()
+        }
+        ```
+    - Create a `README.md` in the `visual-regression` directory with basic instructions for how to run and write visual regression tests. In particular mention our usage of `jest-image-snapshot` instead of the built in snapshot matcher.
+- [ ] Complete https://github.com/WordPress/openverse-frontend/issues/890 if it isn't already done.
+- [ ] Configure CI to run visual regression tests.
+    - Write a new workflow to run the visual regression tests. You should be able to copy the existing e2e workflow and replace the call to `e2e:ci` to `visual-regression:ci`.
```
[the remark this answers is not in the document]
sarayourfriend: Yes! I was just following the pattern that existing but 100%, I'll update this to reflect the new pattern.

sarayourfriend:
> Would writing all search type search pages and single media item detail page be outside of this milestone/plan

I think so. We can write follow up issues to create individual snapshots but it felt outside the scope of this RFC where I wanted to focus on the infrastructure/tooling aspect of it rather than the details of writing the tests (which should mirror whatever best practices we develop for the e2e tests really).

rfcs/20220217-visual_regression_testing.md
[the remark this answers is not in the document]
sarayourfriend: Updated in https://github.com/WordPress/openverse/pull/165/commits/f411fda12a196fb5e18b9559bfd923805bbf4b6e

rfcs/20220217-visual_regression_testing.md
sarayourfriend: Good catch, removed those sections from the RFC in https://github.com/WordPress/openverse/pull/165/commits/f411fda12a196fb5e18b9559bfd923805bbf4b6e

rfcs/20220217-visual_regression_testing.md
obulat: Wow, I didn't realize how big a hurdle removing Docker would be. I guess if nothing works, we could checkout the contributors' PRs ourselves and run the visual regression checks.

On Windows Home edition, which is very widespread, you cannot install Docker. You have to set up WSL2 to be able to run it, which can be intimidating.

sarayourfriend: > On Windows Home edition, which is very widespread, you cannot install Docker

:cold_sweat: What?! That's incredible... I'm surprised they let you install WSL if you can't install Docker :confounded:

The GH Action probably wouldn't be too bad to implement, it'd just take some extra work; in particular we'd need to make sure it keeps the supplemental branch up to date and that the process for how to use those PRs is clear, probably via some automated comments left in the parent PR.

I'll sketch this idea out in the RFC so that it's recorded and try to think through some of the edge cases as well. We can decide whether to spend time implementing it later (I'm inclined to say we should, I didn't consider the fact that some folks might not be able to run Docker at all). In any case, it seems like something that would be nice to have eventually anyway.

This actually makes me think we could add a workflow to generate the diff images and upload them as a comment when the snapshots fail. That would be especially helpful and would prevent folks from having to install the pixelmatch library locally (I don't think it's cross platform to Windows).

sarayourfriend: Added notes about this and the proposed solution using supplemental PRs in the RFC here https://github.com/WordPress/openverse/pull/165/commits/eb2b8b54ed3f4256ea8e173ff76dd16a59a366d8

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,240 @@
...
+
+**Important caveat to this section!**
+
+I ended up trying out `jest-image-snapshot` in https://github.com/WordPress/openverse-frontend/pull/899 but running the matcher fails on some error about `_counters` being undefined. It looks like Playwright, while it uses the jest `expect` library, does not support the same matcher context as Jest proper does. To use `jest-image-snapshot` we might have to switch to `jest-playwright` which would be less than ideal as it's [recommended against by the Playwright project](https://playwright.dev/docs/test-runners#jest--jasmine) (though for what reason exactly I'm really not sure).
+
+We can still _try_ to find a workaround for this, but if it ends up being too much trouble I'd say just leave it out of the final roadmap for this feature and we can use the built-in Playwright tool. Playwright will output the before and after images for failed snapshot comparisons into a `test-results` folder at the top level of the project, so I figure the worst case scenario is that we could write a little script to walk through that directory and run `pixelmatch` on them to get the diff image.
+
+## Repository size considerations
```
AetherUnbound, line 74: We can also use an extension like [git-lfs](https://git-lfs.github.com/) (Git Large File Storage) to keep the images entirely separate from the actual codebase but still part of the same repository. Git LFS can be a pain to work with, but I believe you can work on the codebase without having to interact with it (e.g. download the files) at all.

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,240 @@
...
+  '/src/bin/visual-regression.sh',
+]
+
+if (process.argv.includes('-u')) {
+  args.push('-u')
+}
+
+spawnSync('docker', args, {
+  stdio: 'inherit',
+})
```
AetherUnbound, line 131: I believe this line is missing the 3 trailing backtics and is causing the rendering to get messed up (I'd put up a "suggestion" it but the markdown is not happy about it lol).

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,240 @@
...
+#!/bin/bash
+# This script is meant to run inside the playwright docker container, hence the assumed `/src` directory
+cd /src
+
+# Use npm to avoid having to install pnpm inside the container, it doesn't matter in this case
+npm run test:visual-regression:local -- $1
+```
+3. Add two new package.json scripts:
+```json
+"test:visual-regression": "node ./bin/visual-regression.sh",
```
AetherUnbound, line 143: Should this be a different file?
```suggestion
"test:visual-regression": "node ./bin/visual-regression.js",
```

rfcs/20220217-visual_regression_testing.md
AetherUnbound: > This actually makes me think we could add a workflow to generate the diff images and upload them as a comment when the snapshots fail.

I was thinking we could do something like this! I know the Codecov bot does something similar for code diffs (example here: https://github.com/AllenCellModeling/aicsimageio/pull/377#issuecomment-1046298561). It would be more work though, as you mention.

On the Docker note: while I like the idea of building this so someone doesn't necessarily have to have Docker locally and the checks can still run on GHA, I don't think it's an unreasonable constraint to expect contributors to have Docker installed. Both the catalog and the API stacks function almost exclusively out of containers for local development, so we're already imposing that as a prerequisite in those cases. I don't see a problem extending that to the frontend as well.

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Cross platform rendering
+
+Browsers will sometimes render minute differences between platforms, even with the exact same webpage. This means that contributors running Linux might generate different screenshots from those running macOS and again from those running Windows, even if they're all using the same browser and version.
+
+To avoid this, other visual regression testing libraries like BackstopJS support running tests inside a docker container and using the browser inside the container. We can do the same with Playwright (it's just not as easy).
+
+I was about to figure out "a way" of doing it. Here's what I came up with:
+
+1. Create a `./bin/visual-regression.js` file with the following:
```
AetherUnbound, line 92: I think this approach in general is solid! I've used tactics like it in the past to accomplish similar tasks. My only other thought is that we might get tighter/easier integration if we use a JS Docker API library rather than running a subprocess ourselves. I know very little about the space, but maybe [`node-docker-api`](https://www.npmjs.com/package/node-docker-api) could work? It's definitely not a requirement, but it may help with things like log capture or mounts.

AetherUnbound (approved):
Looks great! I have a few thoughts/potential suggestions, but I would back moving forward with the proposal as-is :smile: :desktop_computer: :eyes: :robot:

rfcs/20220217-visual_regression_testing.md
sarayourfriend: Oh hey that's a nice idea, I didn't even consider that. Given that it's a single `docker run` command though, I'm reluctant to add another dependency for it :thinking:

rfcs/20220217-visual_regression_testing.md
sarayourfriend: You know, that crossed my mind but I don't know why I didn't pursue that idea further. It seems like exactly the right solution for this problem (if indeed it turns out to be a problem cause like I said the snapshot images are pretty tiny)

I have also heard that git lfs can be a pain, but it's gotten better in the past couple years as I understand it.

rfcs/20220217-visual_regression_testing.md
AetherUnbound: Yea, I totally hear that :slightly_smiling_face: I mean, we're doing pretty much [the same thing in the API](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/test/integration_test.py#L299), so it's not a necessary change! I only bring it up because I know these libraries typically tend to handle the text streaming aspects a bit better.

rfcs/20220217-visual_regression_testing.md
sarayourfriend, line 131:
````suggestion
})
```
````

File: rfcs/20220217-visual_regression_testing.md [hunk 8]
```diff
@@ -0,0 +1,241 @@
+# RFC: Visual regression testing
+
+A proposal to integrate visual regression testing into the Openverse frontend's automated testing suite.
+
+## Reviewers
+
+- [x] @obulat
+- [ ] Approver (TBD)
```
sarayourfriend, line 8:
```suggestion
- [x] @AetherUnbound
- [ ] Approver
```

rfcs/20220217-visual_regression_testing.md
sarayourfriend: Fixed! BTW to be able to use backticks inside of `` `inline code text` `` or ```` ```multi-line suggestions``` ```` you have to use an extra back tick on the fence than what you're using inside the block.

For example:

``` `` `inline code text` `` ```

Will render:

`` `inline code text` ``

And for multi-line it's the same:

`````
````
```py
pring("Hello, world!")
```
````
`````

Will render

````
```py
print("Hello, world!")
```
````

Might be easier to edit this comment and inspect the markdown content of it if it is still confusing :slightly_smiling_face:

**Comment by dhruvkb** [comment]
This plan looks really solid to me and I can get behind the value of visual testing for the codebase, considering how we've seen things break down before. I have two main concerns:

- My past experience with visual regression testing (in CC Vocabulary) has been that it's extremely flaky. Depending on things like the operating system and browser versions, tests were inconsistently passing or failing with the diffs being things moved by half a pixel or so. We had integrated it into Vocabulary using Puppeteer, then gradually disabled it and then removed it entirely because it was a pain to maintain and validate these microscopically different diffs. Animations and dynamic content (like search results) can also cause (solvable) issues in testing. 
- 
- We will also need to factor responsiveness into the tests. So for every test case, we should ideally have 5 versions of the snapshot, one for the most common resolution at each breakpoint. While this doesn't guarantee the site holding up well between two breakpoints, it ensures that things aren't broken at the popular resolutions.

**Comment by sarayourfriend** [comment]
> My past experience with visual regression testing (in CC Vocabulary) has been that it's extremely flaky. Depending on things like the operating system and browser versions, tests were inconsistently passing or failing with the diffs being things moved by half a pixel or so. This was even failing inside Docker (because Docker works differently on macOS and Linux). We had integrated it into Vocabulary using Puppeteer, then gradually disabled it and then removed it entirely because it was a pain to maintain and validate these microscopically different diffs. Animations and dynamic content (like search results) can also cause (solvable) issues in testing.

I've never worked directly with visual regression testing before but _everyone_ (and I mean _everyone_) who I know that _has_ relays almost the same experience as you. It's really unfortunate.

It might be good for us to have a contingency plan/tolerance level for flakiness/unpredictability... or something like that. Basically a point at which we'll call it and say "this isn't giving us enough value for the headache that it is to maintain". I do hope that adding the GitHub action to publish the diff output might help this a little bit, even if it's just to be able to say, "yup this failed test is fine (false negative), update the snapshot and call it a day" or something like that.

> We will also need to factor responsiveness into the tests. So for every test case, we should ideally have 5-6 versions of the snapshot, one for the most common resolution at each breakpoint. While this doesn't guarantee the site holding up well between two breakpoints, it ensures that things aren't broken at the popular resolutions.

Yes! I should have mentioned this in the RFC, thank you for bringing it up. We should probably create some kind of utility that iterates over our `screens` constants and then wraps a test in a `describe` block and passes in the breakpoint to apply to the page.

I'll add a suggestion for that in the RFC.

**Comment by dhruvkb** [comment]
Another point was a shoutout to Percy which you mentioned in the RFC. I've used Percy a little bit (that two years ago). It had a great UI for showing the image diffs side by side for every PR that you can approve or flag which directly translates to a PR's mergeability. 

If we can get a FOSS license for Percy that might make the flakiness much less of a hassle.

**Comment by sarayourfriend** [comment]
> If we can get a FOSS license for Percy that might make the flakiness much less of a hassle. The [BrowserStack open-source discount](https://www.browserstack.com/open-source) does not include Percy afaik.

Yeah that would be great but as you said, it doesn't appear to be included in the perks for OSS projects at the moment :disappointed:

**Comment by zackkrida** [comment]
This looks good to me. I'm curious what folks would think about a GitHub action that replies to PRs with a list of before/after screenshots to be diffed manually by reviewers, rather than relying on a known, flaky diffing algorithm. Some benefits I see there:

- Less work to set up
- Avoid flakiness 
- Don't have to store snapshots

Maybe this could instead be a good "Plan B" if we find the final implementation to give us too many false results.

File: rfcs/20220217-visual_regression_testing.md (hunk 1 above)
Comment (krysal) [reply rfcs/20220217-visual_regression_testing.md:151]: Shouldn't snapshots be updated consciously on purpose? If they update automatically then in which case those tests would fail?

**Comment by sarayourfriend** [comment]
> I'm curious what folks would think about a GitHub action that replies to PRs with a list of before/after screenshots to be diffed manually by reviewers, rather than relying on a known, flaky diffing algorithm

That is exactly the proposed workflow... except that the proposal I think is better in that it just relies on GitHub's image comparison diffing _and_ also proposes we upload the pixelmatch diff output to help identify what pixelmatch is seeing is the difference between the two bitmaps.

File: rfcs/20220217-visual_regression_testing.md (hunk 5 above)
Comment (AetherUnbound) [reply rfcs/20220217-visual_regression_testing.md:132]: TIL you can use 4 backtics when you need to insert 3!

Comment (AetherUnbound) [reply rfcs/20220217-visual_regression_testing.md:132]: TIL!! This rocks!

File: rfcs/20220217-visual_regression_testing.md (hunk 1 above)
Comment (sarayourfriend) [reply rfcs/20220217-visual_regression_testing.md:151]: They would only update intentionally and fail the CI on the main PR until they've been updated and committed. It's up to reviewers to ensure that and changes committed to the snapshots are appropriate and match the expected changes from the rest of the PR.

**Comment by sarayourfriend** [comment]
I'm going to merge this even though it's missing the third reviewer I wanted. It's been open for over two weeks now and there are no outstanding objections from the team. Dhruv and Zack also reviewed (but didn't explicitly approve) so I'm assuming there's been enough front-end eyes on this.

File: rfcs/20220217-visual_regression_testing.md [hunk 9]
```diff
@@ -0,0 +1,288 @@
+# RFC: Visual regression testing
+
+A proposal to integrate visual regression testing into the Openverse frontend's automated testing suite.
+
+## Reviewers
+
+- [x] @obulat
+- [x] @AetherUnbound
+- [ ] Approver
```
sarayourfriend, line 9:
```suggestion
```

rfcs/20220217-visual_regression_testing.md
```diff
@@ -0,0 +1,287 @@
...
+A proposal to integrate visual regression testing into the Openverse frontend's automated testing suite.
+
+## Reviewers
+
+- [x] @obulat
+- [x] @AetherUnbound
+
+## Milestone
+
+TBD
```
sarayourfriend, line 12:
```suggestion
https://github.com/WordPress/openverse-frontend/milestone/7
```

sarayourfriend:
I've created the milestone and issues for this RFC https://github.com/WordPress/openverse-frontend/milestone/7

I skipped creating the issue to create the supplemental PR with the updated snapshots. I think Madison is right that it's a reasonable assumption that code-based contributors are able to run docker locally. If this turns out to be a big problem for our contributors then we can revisit adding the supplemental PR.

I also didn't create the issue for the git history cleanup script. It probably won't be an issue and the ticket would just sit open for a long time. If we do find our repo size to be bloated after some time, a solution is already documented in this RFC and we can revisit it then.

PR #171, opened 2022-02-22
RFC: 3D Model Support

AetherUnbound:
Combined from WordPress/openverse-frontend#889 and WordPress/openverse-catalog#370

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/rfcs/20220221-3d_model_support.md
+++ b/rfcs/20220221-3d_model_support.md
@@ -116,6 +116,11 @@ Many of the fields towards the end are taken from examples on SketchFab (e.g. ht
 - 3D models are often heavy files. We should lazy-load and defer loading of these resources as much as possible.
 - We should probably use *thumbnails* rather than actually render 3D models in the search results. We *could* consider something like [SketchFab's search results](https://sketchfab.com/search?q=snacks&type=models), where they render a 3D model on hover of the search results, when on desktop.
 - We'll need to find a standardized 3d model viewer that will work with various formats. We *really, really* don't want to write our own code here. We will also need to decide if we want to use this universally, or use SketchFab's viewer for SketchFab models, which is an excellent viewer.
+- We need to consider low power devices and low bandwidth connections. In both of these instances we'll want to try and significantly limit the number of downloaded assets. Some solutions for this might be:
+  - Write new composables and/or components to detect user bandwidth and power (see https://github.com/GoogleChromeLabs/react-adaptive-hooks as an example in React!)
+  - Only display thumbnails on list views
+  - Have a 'play' button on single result views, that triggers the loading of code-split model viewer code
+  - Simplify mobile models by default, by using flat colors instead of textures, limiting vertice counts, or other approaches. We need to research what is possible here, and what might give a user a false negative impression about a model. We don't want to inadvertently destroy someone's work to the point that users won't want to download it.
 
 ## Prior Art
 
@@ -179,3 +184,4 @@ Design work for the new components must be completed, or at least in a good enou
 - What features do people need in a 3D model search engine? What functionality does our model viewer need to have?
 - Should we use SketchFab's viewer for SketchFab models?
 - Some models are meant to be 3D printed; some are meant to be used as assets in digital art and games; others are more like standalone pieces of artwork. There are probably different user stories for each of these types. How can we optimize our UI for all of these different experiences?
+
--- a/rfcs/20220221-3d_model_support.md
+++ b/rfcs/20220221-3d_model_support.md
@@ -1,6 +1,6 @@
 # RFC: 3D Model Support
 
-- [ ] [Approver TBD]
+- [x] @sarayourfriend
 - [ ] [Approver TBD]
 
 ## Rationale
--- a/rfcs/20220221-3d_model_support.md
+++ b/rfcs/20220221-3d_model_support.md
@@ -34,6 +34,20 @@ We may be able to use previous work regarding the addition of audio to the catal
 
 WordPress/openverse#19
 
+## Design requirements
+
+### New content type
+- [ ] Create a 3D model icon
+- [ ] Create an icon indicating that the image is a 3d model screenshot/preview
+
+### Search results
+- [ ] Add 3d model results in the All content results
+- [ ] Design a 3d model results page
+- [ ] Review and include filter components if necessary
+
+### Single result view
+- [ ] Design the 3d model single view
+
 ## Technical implementation
 
 ### 1. Add 3D models to the catalog
--- a/rfcs/20220221-3d_model_support.md
+++ b/rfcs/20220221-3d_model_support.md
@@ -1,0 +1,181 @@
+# RFC: 3D Model Support
+
+- [ ] [Approver TBD]
+- [ ] [Approver TBD]
+
+## Rationale
+
+The Openverse team and members of the community have identified a number of sources of CC-licensed 3D models. We would like to add these models to Openverse, as they can be useful for game development, art, 3D printing, etc. In order to incorporate them into Openverse, we need to add a new media type and provide infrastructure in the frontend for rendering the models.
+
+## Overview
+
+Openverse will add first-class support for 3D models, alongside audio and images. This means 3D models will have their own table in the Catalog, and their own API endpoints. Adding these to the frontend will require some new components and utilities, but should be more straightforward than the implementation of audio because we are not conducting a complete redesign of the site in parallel.
+
+This outline of work is predicated on the idea that we've identified sources of 3D models and will seek to flesh out the details regarding metadata fields, database migrations, API development, and frontend work.
+
+This work describes 3D models as a specific entity, exclusive of 2D models. 2D models can be covered under our `image` media type and identified via a separate category.
+
+# 3D Models in the Catalog & API
+
+## Some potential risks
+
+* As it stands currently, each new media type means an additional data refresh. Data refreshes cannot be run simultaneously at present, so we may reach a saturation point where all of our data cannot be refreshed weekly.
+* We do not have any standard database migration approach for the catalog, so all changes suggested here will likely be ad hoc unless #235 is addressed prior to this implementation.
+* The API, while it does have a migration tool, does not have a standardized migration application process. We will need to run the migration manually when the time comes.
+
+## Some other considerations
+
+* It would be ideal to have this under a feature flag in the API. We could perform deployments and testing on our staging setup with the flag enabled, but still make other deployments to production with the flag disabled. We should be able to run the API for images & audio successfully with the feature flag disabled AND without having run the migrations.
+* As we work through this on both the catalog and the API, we should begin compiling information for a document describing the process for adding a **new media type**. This will be useful for future media types we are sure to add!
+
+## Prior work
+
+We may be able to use previous work regarding the addition of audio to the catalog/API as a guide. These updates required a fair amount of generalization, which we may not need to do a second time around.
+
+WordPress/openverse#19
+
+## Technical implementation
+
+### 1. Add 3D models to the catalog
+
+- [ ] Define necessary fields & schema for `model_3d` table & associated data
+- [ ] Add DDL to local Postgres image: https://github.com/WordPress/openverse-catalog/tree/main/docker/local_postgres
+- [ ] Create a popularity calculation view for `model_3d`
+- [ ] Add the database columns to `db_columns.py`: https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/storage/db_columns.py
+- [ ] Add the TSV columns to `tsv_columns.py`: https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/storage/tsv_columns.py
+- [ ] Create a `Model3DStore` subclass of [`MediaStore`](https://github.com/WordPress/openverse-catalog/blob/9538f38401e5428b32dc14c5d5ba7ef50af87a96/openverse_catalog/dags/common/storage/media.py#L39): 
+- [ ] Update the [new provider API script template](https://github.com/WordPress/openverse-catalog/tree/main/openverse_catalog/templates) with the `model_3d` provider
+- [ ] Create one/several provider API scripts & associated DAGs
+
+### 2. Add 3D models to the API
+_Note: Borrowed heavily from WordPress/openverse#19_
+- [ ] Django models for `model_3d`
+- [ ] Add DRF serializers for `model_3d`
+- [ ] Add DRF endpoints for `model_3d`
+- [ ] Make the thumbnail server work with model thumbnails
+
+### 3. Add 3D models to the data refresh
+- [ ] Add Elasticsearch `model_3d` models
+- [ ] Ensure the data refresh works for 3D models
+- [ ] Add 3D model sample data into the sample data CSV for testing
+
+
+## Open Questions
+
+* What data is important to store in the catalog? Some of this is encapsulated in the proposed schema, but there may be other important attributes beyond those listed.
+* Will it be important to differentiate the "type" of 3D model that a particular record refers to? (E.g. 3D printer model, game asset, artwork, etc.)
+* Many model sites come with asset groups, should we have a table similar to `AudioSets` for these cases?
+
+## Proposed schema
+
+Below is the proposed schema for the `model_3d` table in the catalog database. It is subject to change based on discussions that happen here and elsewhere.
+
+Many of the fields towards the end are taken from examples on SketchFab (e.g. https://sketchfab.com/3d-models/charmanders-midnight-snack-57866c01da20414ba81885206acbcc85)
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `identifier` | `UUID` | Identifier for the record |
+| `created_on` | `timestamp` | Time the record was created |
+| `updated_on` | `timestamp` | Time the record was updated |
+| `ingestion_type` | `varchar` | What generated the record |
+| `provider` | `varchar` | Provider API that the model comes from |
+| `source` | `varchar` | Source within the provider (e.g. NASA on Flickr), typically the same as `provider` |
+| `foreign_id` | `varchar` | Identifier that the provider uses for this model |
+| `foreign_landing_url` | `varchar` | URL on the provider that displays the "single-result" page for this model |
+| `url` | `varchar` | URL to the actual asset itself (e.g. the image URL) |
+| `thumbnail` | `varchar` | Thumbnail URL |
+| `filesize` | `integer` | Size of the file |
+| `license` | `varchar` | License for the model |
+| `license_version` | `varchar` | Version of the model's license |
+| `creator` | `varchar` | Creator ID of the model from the provider |
+| `creator_url` | `varchar` | URL to the creator's page on the provider's site |
+| `title` | `varchar` | Title of the model |
+| `meta_data` | `jsonb` | Uncategorized metadata for the model |
+| `tags` | `jsonb` | Tags that describe the model, for use in searching |
+| `watermarked` | `bool` | Whether the model is watermarked - is this needed for 3D models? |
+| `last_synced_with_source` | `timestamp` | Time the record was last updated from the provider |
+| `removed_from_source` | `bool` | Whether the model has been removed from the upstream source |
+| `category` | `varchar` | Category for the model (is this where we'll make the distinction between 3D printers, game assets, etc.?) |
+| `filetype` | `varchar` | File type of the primary file for the model |
+| `alt_files` | `jsonb` | Additional file types for the model |
+| `vertices` | `integer` | Number of vertices within the model |
+| `textures` | `integer` | Number of [textures](https://en.wikibooks.org/wiki/Blender_3D:_Noob_to_Pro/Materials_and_Textures) the model uses |
+| `materials` | `integer` | Number of [materials](https://en.wikibooks.org/wiki/Blender_3D:_Noob_to_Pro/Materials_and_Textures) the model uses |
+| `uv_mapping` | `bool` | Whether the model contains [UV Mapping](https://en.wikipedia.org/wiki/UV_mapping) |
+| `vertex_colors` | `bool` | Whether the model contains [vertex colors](https://help.sketchfab.com/hc/en-us/articles/210765623-Vertex-Colors) |
+| `pbr` | `bool` | Whether the model supports [physically based rendering](https://en.wikipedia.org/wiki/Physically_based_rendering) |
+| `rigged_geometries` | `bool` | Whether the model supports [rigging for animation](https://dreamfarmstudios.com/blog/what-is-3d-rigging/) |
+| `animations` | `integer` | Number of animations the model has |
+| `morph_geometries` | `integer` | Number of [morph targets for animation](https://en.wikipedia.org/wiki/Morph_target_animation) |
+
+
+# 3D Models in the frontend
+
+## Some potential risks
+
+- 3D models are often heavy files. We should lazy-load and defer loading of these resources as much as possible.
+- We should probably use *thumbnails* rather than actually render 3D models in the search results. We *could* consider something like [SketchFab's search results](https://sketchfab.com/search?q=snacks&type=models), where they render a 3D model on hover of the search results, when on desktop.
+- We'll need to find a standardized 3d model viewer that will work with various formats. We *really, really* don't want to write our own code here. We will also need to decide if we want to use this universally, or use SketchFab's viewer for SketchFab models, which is an excellent viewer.
+
+## Prior Art
+
+- We already show 3D models as [part of images](https://search.openverse.engineering/search/image?q=dogs&source=sketchfab,thingiverse), which is great! Fewer scripts to write on the catalog side.
+- We also already show Sketchfab's 3D model viewer for their single results: https://search-staging.openverse.engineering/image/00a6a7a4-6605-4cba-bacf-a08b0da546b9
+
+## Some code considerations
+
+- When a unique identifier is necessary, we should use the string constant `const MODEL_3D = '3d model'` to refer to 3D models. Compare this to `AUDIO` for audio files, for example.
+- Similar to the catalog/API, 3D model support should be enabled via a feature flag. This work should be able to be committed directly to `main` without a base feature branch needed. Meta search support for 3D models does not need to be feature flagged.
+- e2e tests should be added for any new routes; and updated for existing routes that are modified. PRs will be blocked in the abcense of these tests.
+- New content types should always launch in `beta` until we reach a certain number of results and/or are confident in our UI components (this metric(s) need to be determined).
+- Many of the todos identified here can be broken down into even smaller pieces in the future, particularly around the user interface.
+
+## Technical implementation
+
+### 1. Add 3D Models to Meta Search
+
+#### Prerequisites
+
+- Need a list of 3D model providers to use in meta search. This could be a mix of sources we intend to use in the Catalog but also some that don't have APIs and will stay in meta search after launching 3D model support.
+
+#### Todos
+
+- [ ] Add 3D models to the content switcher components:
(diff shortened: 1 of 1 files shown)


rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+	- [ ] Render the model info
+	- [ ] Render the attribution information
+	- [ ] Render the related models
+- [ ] Create a new `VThreeDModel.vue` component that supports displaying 3D models of various file formats
+
+## Open Questions
+
+- What model formats do we want/need to support?
+- What features do people need in a 3D model search engine? What functionality does our model viewer need to have?
+- Should we use SketchFab's viewer for SketchFab models?
```
sarayourfriend, line 180: Copying my remarks and @zackkrida's additional background information from the issue:

I would really really like to avoid SketchFab's viewer if possible.

1. It is not free software
2. It has much [looser/worse privacy guarantees](https://help.sketchfab.com/hc/en-us/articles/203059088-Compatibility#troubleshooting-scripts) than we've tried to achieve with Openverse so far

On the other hand, their viewer is extremely convenient and works very well.

In a brief search I found this [js-3d-model-viewer](https://github.com/cgwire/js-3d-model-viewer) library that is MIT licensed. It's not great, but I wonder if we could help improve it and add some of the features we might want (like UI buttons for controlling zoom when a mousewheel isn't available, recentering, snapping to specific axis, etc).

Currently it only support OBJ files though and has poor texture support (the latter of which is something we could probably spend some time contributing against). These might be unacceptable limitations though.

I guess it depends on how much upfront work we're willing to do for this or if it's something we'd rather just get out the door and be able to move on to video.

**From Zack:**

Regarding viewers, I chatted a bit with @antpb today who works on wp core media, but also around open source projects tackling standardizing 3D content performance, and might have some recommendations for us here smile

sarayourfriend (commented):
Exciting!!!

rfcs/20220221-3d_model_support.md
antpb: Hey there! I have been pretty heavy involved in 3D web development over the last several years and have been particularly focused on Three.js and glTF as my go-to stack for rendering 3D in a browser. I have a plugin that was released a couple weeks ago that officially supports three.js, WebXR and glTF files in the WordPress block editor. You can try the plugin here: https://wordpress.org/plugins/three-object-viewer/

I feel pretty strongly that glTF, fbx, Usdz, and obj will be the most asked for supports. React three fiber is a library that supports gltf and fbx out of the box and I have an example of an obj loader relative to three.js that I would be willing to contribute. React Three Fiber is what I'm using for the render and it is very plug and play with Gutenberg, so implementations of 3D seem like they would be more consistent with a library such as this. Hopefully my plugin will help give inspiration around how people want to use 3D files on their sites. I have many more features like animations and interactions in the pipeline!

Separately from the plugin above, I am part of a group working toward standards around 3D content on the web. The group is called Open Metaverse Interoperability Group: https://omigroup.org and it has been very successful in building a community of leaders in the open metaverse pursuit. All code within the group adheres to the w3c requirements and is licensed MIT. The group is looking for wider support of these extensions as time goes on so Openverse seems a really great pairing! 

( More on the w3c side of OMI here: https://www.w3.org/community/metaverse-interop/  )

The problem currently that the group is solving is around a lack of standards with glTF when it comes to more rich interactions (such as audio, interactive elements, etc). Every engine or renderer requires a unique implementation to achieve these features which means the content (3D object) usually only works for one. The group is solving part of this problem through glTF extensions which allow for interoperability across many environments. Currently the team has developed an audio emitter spec that has resulted in a node package that allows for adding a single parser to support both spatial and global audio in a three.js scene. These standards are not specific to three.js which is what makes the support of the gltf-extensions interoperable across many engines. An example of an extension spec is here: https://github.com/omigroup/gltf-extensions

Also here is the plug and play three.js package that enables files with these extensions to work like magic: https://www.npmjs.com/package/three-omi

A live example of this audio emitter can be found here: https://3ov.xyz/omi-audio-demo/

The above package was implemented in my plugin and you can explore the different features at my plugin's website here: https://3ov.xyz/

Hopefully this info helps. I realized I dumped a bunch of links but I'm always happy to talk about this topic as I have given it _a lot_ of thought. :D

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+We may be able to use previous work regarding the addition of audio to the catalog/API as a guide. These updates required a fair amount of generalization, which we may not need to do a second time around.
+
+WordPress/openverse#19
+
+## Technical implementation
+
+### 1. Add 3D models to the catalog
+
+- [ ] Define necessary fields & schema for `model_3d` table & associated data
+- [ ] Add DDL to local Postgres image: https://github.com/WordPress/openverse-catalog/tree/main/docker/local_postgres
```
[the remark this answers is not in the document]
zackkrida: Data Definition Language, so here it means the sql file that creates the 3d models table in local postgres. It's not a true migration in the sense that it's not run by a migration tool in a managed way, it's basically a schema. 

I do think we could clarify this for contributors though!

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Some potential risks
+
+- 3D models are often heavy files. We should lazy-load and defer loading of these resources as much as possible.
+- We should probably use *thumbnails* rather than actually render 3D models in the search results. We *could* consider something like [SketchFab's search results](https://sketchfab.com/search?q=snacks&type=models), where they render a 3D model on hover of the search results, when on desktop.
+- We'll need to find a standardized 3d model viewer that will work with various formats. We *really, really* don't want to write our own code here. We will also need to decide if we want to use this universally, or use SketchFab's viewer for SketchFab models, which is an excellent viewer.
+
+## Prior Art
+
+- We already show 3D models as [part of images](https://search.openverse.engineering/search/image?q=dogs&source=sketchfab,thingiverse), which is great! Fewer scripts to write on the catalog side.
```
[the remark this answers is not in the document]
zackkrida: Or just delete the existing image data for these providers from the catalog, and then run the new provider scripts. I'll add a note about this.

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+
+- [ ] Add 3D models to the All Results view
+  - [ ] Create a `VModelCell.vue` component in the [`VAllResultsGrid` directory](https://github.com/WordPress/openverse-frontend/blob/f5b7518c3942a2ede49d5c7fef2831d30e9867a9/src/components/VAllResultsGrid)
+  - [ ] Add the logic to display models in the [`VAllResultsGrid.vue`](https://github.com/WordPress/openverse-frontend/blob/8214217f84099a83b9767f6a5c73fdbdf278bc95/src/components/VAllResultsGrid/VAllResultsGrid.vue#L1) component.
+- [ ] Create a new single model route `pages/model/_id.vue` for viewing a 3D model in detail.
+	- [ ] Render the model
+	- [ ] Render the model info
+	- [ ] Render the attribution information
+	- [ ] Render the related models
+- [ ] Create a new `VThreeDModel.vue` component that supports displaying 3D models of various file formats
```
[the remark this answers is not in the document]
zackkrida: Yeah, these items are very generic certainties, and we'll definitely need a new, more detailed RFC for implementing the designs. I'll also add a note about that here.

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+
+The Openverse team and members of the community have identified a number of sources of CC-licensed 3D models. We would like to add these models to Openverse, as they can be useful for game development, art, 3D printing, etc. In order to incorporate them into Openverse, we need to add a new media type and provide infrastructure in the frontend for rendering the models.
+
+## Overview
+
+Openverse will add first-class support for 3D models, alongside audio and images. This means 3D models will have their own table in the Catalog, and their own API endpoints. Adding these to the frontend will require some new components and utilities, but should be more straightforward than the implementation of audio because we are not conducting a complete redesign of the site in parallel.
+
+This outline of work is predicated on the idea that we've identified sources of 3D models and will seek to flesh out the details regarding metadata fields, database migrations, API development, and frontend work.
+
+This work describes 3D models as a specific entity, exclusive of 2D models. 2D models can be covered under our `image` media type and identified via a separate category.
```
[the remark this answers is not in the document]
antpb: I think textures are hard to anticipate the UV maps for every object. There are some common objects like maybe humanoid characters that would have very similar textures for things like hair, but overall it's on an object by object basis in my experience. Things like compression also manipulate those UV maps so textures may be inconsistent this way.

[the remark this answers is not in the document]
antpb: I do think textures in general are a category separate from 3D useful for many cases! I use subtle textures all the time in 3D assets but I think they have more utility than just 3D objects

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+| `meta_data` | `jsonb` | Uncategorized metadata for the model |
+| `tags` | `jsonb` | Tags that describe the model, for use in searching |
+| `watermarked` | `bool` | Whether the model is watermarked - is this needed for 3D models? |
+| `last_synced_with_source` | `timestamp` | Time the record was last updated from the provider |
+| `removed_from_source` | `bool` | Whether the model has been removed from the upstream source |
+| `category` | `varchar` | Category for the model (is this where we'll make the distinction between 3D printers, game assets, etc.?) |
+| `filetype` | `varchar` | File type of the primary file for the model |
+| `alt_files` | `jsonb` | Additional file types for the model |
+| `vertices` | `integer` | Number of vertices within the model |
+| `textures` | `integer` | Number of [textures](https://en.wikibooks.org/wiki/Blender_3D:_Noob_to_Pro/Materials_and_Textures) the model uses |
```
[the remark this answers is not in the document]
antpb: Number of textures comes into play when you have multiple materials. Material count has a pretty big impact on performance and often 3D browser experiences recommend around 20 materials max without having performance issues. (Like iOS just crashes sometimes) Worth looking at is the guidelines that Mozilla Hubs recommends for their 3D scenes. They mention how textures relate to performance and why they matter with a mention of sketchfab: https://hubs.mozilla.com/docs/spoke-optimization.html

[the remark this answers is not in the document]
antpb: When you save files in that platform, it gives you a performance readout. These might be good just to note what folks care about when creating content.

Performance Check from Spoke

panchovm:
To @sarayourfriend's point

> attribution recommendations/guidelines might be slightly different for 3D models. In particular I'm thinking about when you use them for a game or an animated film or something, what's the correct way to give attribution?

The attribution box shows the attribution info required instead of guiding how to place it depending on the format. To your point, we might add a CC-recommended documentation link for people using 3D models. Although I have not seen anything similar.

rfcs/20220221-3d_model_support.md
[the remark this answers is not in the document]
AetherUnbound: Oops, oversight on my part! Would adding a reference here to, 1) what a DDL is and 2) a current example, help here?

rfcs/20220221-3d_model_support.md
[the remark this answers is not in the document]
AetherUnbound: Thank you for your input @antpb! I am no 3D-model expert, so any other fields you think might be useful to have here would be welcome :smile:

rfcs/20220221-3d_model_support.md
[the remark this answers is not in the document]
AetherUnbound: Yea great call, it would be easiest to purge them from the catalog's `image` table and run the new provider as Zack describes.

rfcs/20220221-3d_model_support.md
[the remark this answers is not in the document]
sarayourfriend: I feel like it's domain knowledge or something that I should have known :sweat_smile: You do already link to the folder of existing DDLs, I just didn't know that SQL was DDL.

rfcs/20220221-3d_model_support.md
zackkrida: Despite the fact that Openverse's frontend is written in Vue.js, it _could still be_ entirely appropriate to use react-three-fiber regardless. Thanks for your thoughts!

sarayourfriend: I don't think we should load react for this, but it does seem like there are some Vue equivalents: https://github.com/troisjs/trois

It appears to be more or less the same thing as react-three-fiber. It would need some work to get [this issue closed](https://github.com/troisjs/trois/issues/70) before we could use it though.

zackkrida: I wouldn't be surprised if some renderers are even larger than react and react-dom, so while it's unlikely it's not completely off the table for me, but Trois looks amazing! Well maintained and it would be fun to contribute anything we find missing.

sarayourfriend: That's true. I guess either way with this work we'll need to be careful to make sure we're lazy-loading/bundle splitting the rendering libraries and tools anyway; they are indeed quite heavy.

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+- [ ] Add Elasticsearch `model_3d` models
+- [ ] Ensure the data refresh works for 3D models
+- [ ] Add 3D model sample data into the sample data CSV for testing
+
+
+## Open Questions
+
+* What data is important to store in the catalog? Some of this is encapsulated in the proposed schema, but there may be other important attributes beyond those listed.
+* Will it be important to differentiate the "type" of 3D model that a particular record refers to? (E.g. 3D printer model, game asset, artwork, etc.)
+* Many model sites come with asset groups, should we have a table similar to `AudioSets` for these cases?
```
sarayourfriend, line 67: Aside: It appears that AudioSets is actually just a json blob on the Audio row.

Why is that not a foreign key I wonder?

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+
+### 3. Add 3D models to the data refresh
+- [ ] Add Elasticsearch `model_3d` models
+- [ ] Ensure the data refresh works for 3D models
+- [ ] Add 3D model sample data into the sample data CSV for testing
+
+
+## Open Questions
+
+* What data is important to store in the catalog? Some of this is encapsulated in the proposed schema, but there may be other important attributes beyond those listed.
```
sarayourfriend, line 65: This is probably a good question to call out for @antpb to give feedback directly on (or Anthony, if you know any other folks who'd be willing to give input here that'd be :100: as well)

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+* Will it be important to differentiate the "type" of 3D model that a particular record refers to? (E.g. 3D printer model, game asset, artwork, etc.)
```
sarayourfriend, line 66: I think yes, if we can get this metadata it'd be a useful filter :slightly_smiling_face:

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+| `rigged_geometries` | `bool` | Whether the model supports [rigging for animation](https://dreamfarmstudios.com/blog/what-is-3d-rigging/) |
+| `animations` | `integer` | Number of animations the model has |
+| `morph_geometries` | `integer` | Number of [morph targets for animation](https://en.wikipedia.org/wiki/Morph_target_animation) |
+
+
+# 3D Models in the frontend
+
+## Some potential risks
+
+- 3D models are often heavy files. We should lazy-load and defer loading of these resources as much as possible.
```
sarayourfriend, line 116: What do we need to do for this? Does this just mean using intersection observer or something like that to prevent them from loading?

If IntersectionObserver isn't available on a platform, should we fall back to deferring loading some other way?

Do we need to try to eagerly fetch resources under certain circumstances (maybe we start loading the 3d model after hover or focusing on a result for a certain amount of time).

We probably need to invest more time into figuring out how we're going to demonstrate that the resource is still loading. We have [this issue for adding a loading animation for the audio](https://github.com/WordPress/openverse-frontend/issues/1001) (in fact I thought that issue already existed but it must have fell through the cracks when we discussed it). Whatever we do there could inform what we do for 3d models as well.

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+- We should probably use *thumbnails* rather than actually render 3D models in the search results. We *could* consider something like [SketchFab's search results](https://sketchfab.com/search?q=snacks&type=models), where they render a 3D model on hover of the search results, when on desktop.
+- We'll need to find a standardized 3d model viewer that will work with various formats. We *really, really* don't want to write our own code here. We will also need to decide if we want to use this universally, or use SketchFab's viewer for SketchFab models, which is an excellent viewer.
```
sarayourfriend, line 118: If we could flesh out more details about graceful degradation on low power devices that would be great. As mentioned above, 3d models can be heavy assets. We should avoid crashing someones low-power phone by loading too many (or potentially _any_) of them. Unless there are ways to process the assets that could accommodate low memory or small CPU; for example, is it possible to automatically decrease the number of vertices on a model or something like that?

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+- [ ] Add 3D models to the content switcher components:
+  - [ ] https://github.com/WordPress/openverse-frontend/blob/bbccc36ed49186b369cc4a69693fbc6d1411f29a/src/components/VContentSwitcher/VContentSwitcherButton.vue
+  - [ ] https://github.com/WordPress/openverse-frontend/blob/7d74f033ddc4d35889d7c43756c36150beee4192/src/components/VContentSwitcher/VContentTypes.vue
+- [ ] Create a `model/index.vue` endpoint for 3D models in the [pages directory](https://github.com/WordPress/openverse-frontend/blob/037f2efb768fdbdcff4b875e7f7c271ff50df59f/src/pages)
+- [ ] Add 3D model sources to the [legacy sources lookup](https://github.com/WordPress/openverse-frontend/blob/268532a05004b72fb91b77a875f3a3c1d7fb0599/src/utils/get-legacy-source-url.js#L1)
+- [ ] Add 3D models to the [media constants](https://github.com/WordPress/openverse-frontend/blob/4aeb9fd0d9f893f2df1752d6bb1e52351b8c3d35/src/constants/media.js#L13) file, making sure it is set to the `ADDITIONAL` status.
+
+### 2. Setup the data for 3D models
+
+- [ ] Create a feature flag for 3D model support. All 3D model features will be behind this.
```
sarayourfriend, line 152: Can we specify what "Create a feature flag" means exactly _vis_ https://github.com/WordPress/openverse/pull/171/files#r811500414. I feel like regardless of the _how_ behind the feature flag flipping mechanism it should get put inside a composable, right? Probably a `readonly`-after-initialization Pinia store?

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,181 @@
...
+- [ ] Generate mock/sample api response data for 3D models.
+- [ ] Create a new media service for Models in the [data directory](https://github.com/WordPress/openverse-frontend/blob/fe24ec9bf6a4d3ad429901cef53744dccd93ed5d/src/data)
+- [ ] Update the [media constants](https://github.com/WordPress/openverse-frontend/blob/4aeb9fd0d9f893f2df1752d6bb1e52351b8c3d35/src/constants/media.js#L1) file with new constants for models, and under the `BETA` flag.
+- [ ] Update the [media store](https://github.com/WordPress/openverse-frontend/blob/f9f549beace4f76480c8cbf1e91d8940a0aeef2b/src/store/media.js#L347) with 3D model support.
```
sarayourfriend, line 156: This link will probably stale depending on the timing with #174's implementation.

rfcs/20220221-3d_model_support.md
AetherUnbound: Hmm, I don't think it is? It looks like it's a separate table to me: https://github.com/WordPress/openverse-api/blob/c5daff0a090883d2232394872293796734689640/api/catalog/api/models/audio.py#L186

```
openledger> \dt audio*;
+--------+------------------+-------+--------+
| Schema | Name             | Type  | Owner  |
|--------+------------------+-------+--------|
| public | audio            | table | deploy |
| public | audiolist        | table | deploy |
| public | audiolist_audios | table | deploy |
| public | audioset         | table | deploy |
+--------+------------------+-------+--------+
SELECT 4
Time: 0.087s
openledger> \d audioset;
+---------------------+--------------------------+-------------------------------------+
| Column              | Type                     | Modifiers                           |
|---------------------+--------------------------+-------------------------------------|
| id                  | integer                  |  not null                           |
| created_on          | timestamp with time zone |  not null default CURRENT_TIMESTAMP |
| updated_on          | timestamp with time zone |  not null default CURRENT_TIMESTAMP |
| foreign_identifier  | character varying(1000)  |                                     |
| title               | character varying(2000)  |                                     |
| foreign_landing_url | character varying(1000)  |                                     |
| creator             | character varying(2000)  |                                     |
| creator_url         | character varying(2000)  |                                     |
| url                 | character varying(1000)  |                                     |
| filesize            | integer                  |                                     |
| filetype            | character varying(80)    |                                     |
| thumbnail           | character varying(1000)  |                                     |
| provider            | character varying(80)    |                                     |
+---------------------+--------------------------+-------------------------------------+
Indexes:
    "api_audioset_pkey" PRIMARY KEY, btree (id)
    "api_audioset_url_key" UNIQUE, btree (url)
    "unique_foreign_identifier_provider" UNIQUE, btree (foreign_identifier, provider)
    "api_audioset_foreign_identifier_28c1db59" btree (foreign_identifier)
    "api_audioset_foreign_identifier_28c1db59_like" btree (foreign_identifier varchar_pattern_ops)
    "api_audioset_url_200f9fb4_like" btree (url varchar_pattern_ops)
    "audioset_provider_cc061f30" btree (provider)
    "audioset_provider_cc061f30_like" btree (provider varchar_pattern_ops)

Time: 0.551s
```

I could be misinterpreting your question :slightly_smiling_face:

sarayourfriend: I was going off of the DDL in the catalog, but maybe it gets turned into a real table when transferred to the API DB?

https://github.com/WordPress/openverse-catalog/blob/0c5b43fbab8cba491817ebb647b77e17114a7353/docker/local_postgres/0007_openledger_audio_view.sql#L90-L103

https://github.com/WordPress/openverse-catalog/blob/6c172033e42a91bcd8f9bf78fd6b933a70bd88bf/archive/openledger_sql/03_create_audio_schema.sql#L33-L35

AetherUnbound: Ah yea, that's exactly what happens!

https://github.com/WordPress/openverse-api/blob/1d186bcf9d1df29d917bd79847840f82cf20e0ce/ingestion_server/ingestion_server/tasks.py#L115

I can elaborate a bit more on this point to denote the distinction (since it'll likely be a similar setup).

sarayourfriend: Is that more performant for ingestion because it eliminates a FK? It seems counter-intuitive that storing _more_ data overall would be better (especially as jsonb).

AetherUnbound: I'm almost certain that's the reason that's the reason why this is in the primary table as `jsonb` and pulled out into a view, maybe @obulat and @krysal might know more?

zackkrida:
I thought of something I need to add to the RFC:

In the frontend 'All Results' view, should we make the media type _also_ some kind of filter? Since we will have more than 2 media types, will we allow permutations like '3d models and images' or 'images and audio' (basically excluding one of the three types)? Or will we just keep the choice 'all' vs. 'single' content type?

rfcs/20220221-3d_model_support.md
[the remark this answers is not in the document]
zackkrida: Yeah, 'texture' might be a good category to add to our images, once we have a source with those kinds of assets.

rfcs/20220221-3d_model_support.md
zackkrida: Added a note that this depends on the feature flag RFC.

zackkrida:
@WordPress/openverse-developers Can we get more reviews on this? If you've already looked at it in the past, please know I made some requested updates today.

zackkrida:
@panchovm here's some design thoughts I had. I think we should add a design section to the RFC before merging it. What do you think? If you have any recommendations or tasks I didn't think of please let me know. Once you've reviewed I can add these to the RFC.

- [ ] Design/select an icon for the 3D content type
- [ ] Design how 3D models look in the 'all results' view
  - Need a way of distinguishing 3D models from images, since we will only show model thumbnails
- [ ] Design results grid when the chosen content type is '3D models'
- [ ] Design 3D model single result view
  - [ ] How will 3D models look in our viewer? Will they have show their default textures, show a single color, etc.? This isn't an CC-licensed asset, but here's [an example](https://creazilla.com/nodes/3061-lion-3d-model) of viewing the original textures *and* a plain color model.
  - [ ] What will the interaction look like? There should be a 'play' button, i.e, models aren't loaded by default. Some sites, like [thingiverse](https://www.thingiverse.com/thing:5238596), include actual photographs of the models, not just the 3d renders. How/where do we display those? Like album art, with the metadata, or some other way?
  - [ ] Any special metadata UI we need that doesn't already exist?

sarayourfriend:
> How will 3D models look in our viewer? Will they have show their default textures, show a single color, etc.? This isn't an CC-licensed asset, but here's [an example](https://creazilla.com/nodes/3061-lion-3d-model) of viewing the original textures and a plain color model.

It didn't occur to me that for MVP for this we could avoid having to solve the interactive viewer problem altogether and just choose some angled poses and static renders like the link you shared does. It would also let us defer figuring out how to accommodate low power devices.

I took a closer look at Trois a couple weekends ago and it needs a _lot_ of work to make it Vue 2/3 compatible. It makes heavy usage of the Options API in ways that would be difficult and take a long time to help pick apart into composables. We could do it, but it would probably take a lot of someone's time on the team over a long period of time. And it seems like the maintainer is (understandably) more interested in maintaining and enhancing existing features rather than adding Vue 2 support.

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,211 @@
+# RFC: 3D Model Support
+
+- [ ] [Approver TBD]
```
sarayourfriend, line 3: Tentatively approving. The backend plan seems good to me. I think we'll want another RFC with more specific implementation details once we have frontend designs but this has been a really good starting point for having us discuss and think about the various hurdles we'll need to make sure designs are aware of.

```suggestion
- [z] @sarayourfriend
```

sarayourfriend (approved):

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,211 @@
...
+
+* It would be ideal to have this under a feature flag in the API. We could perform deployments and testing on our staging setup with the flag enabled, but still make other deployments to production with the flag disabled. We should be able to run the API for images & audio successfully with the feature flag disabled AND without having run the migrations.
+* As we work through this on both the catalog and the API, we should begin compiling information for a document describing the process for adding a **new media type**. This will be useful for future media types we are sure to add!
+
+## Prior work
+
+We may be able to use previous work regarding the addition of audio to the catalog/API as a guide. These updates required a fair amount of generalization, which we may not need to do a second time around.
+
+WordPress/openverse#19
+
```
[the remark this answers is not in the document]
AetherUnbound: I love it, thanks @panchovm!

panchovm (approved):

rfcs/20220221-3d_model_support.md
```diff
@@ -0,0 +1,225 @@
+# RFC: 3D Model Support
+
+- [x] @sarayourfriend
+- [ ] [Approver TBD]
```
zackkrida, line 4:
```suggestion
- [x] @panchovm 
```

PR #180, opened 2022-03-08
Create a workflow to label a PR based on its linked issues

### LINKED ISSUE #75: [Feature] Automatically label PRs with the aspect and goal labels
## Problem

Automatically set the aspect and goal labels on PRs to be the same as their linked issues.

## Alternatives

We can also have a default "needs labels" label that is applied on PRs by default. That way we won't have to find linked issues and just set this label.

## Additional context

Since GitHub API does not offer a way to quickly find the issue linked to a PR, scraping the issue page is an [inelegant but functional solution](https://stackoverflow.com/a/60780499).

## Implementation

### LINKED ISSUE #175: PRs created with labels via GH CLI fail label check
## Description

It would appear that there are some cases where the label check fails even though a PR is opened with the correct labels. These failures persist across re-runs of the label check GHA and are only resolved by removing the labels and re-adding them in a separate step.

## Reproduction

1. Submit a PR using the GH CLI (`gh pr create`)
2. During the metadata step, add the required labels (e.g. `priority: medium`, `aspect: dx`, `goal: improvement`)
3. Observe that the PR label check fails even though the labels exist on the PR
4. (Optional) Re-run the label check GHA and observe that it fails again

## Expectation

If a PR is submitted with the appropriate labels, the PR label check should pass.

## Additional context

See WordPress/openverse-catalog#372

## Resolution

dhruvkb:
## Fixes

Fixes #175 by @AetherUnbound 
Fixes #75 by @dhruvkb 

## Description

This PR sets up the action to label PRs based on info from their linked issues. It follows this algorithm:

- if the PR has linked issues, iterate over them to find the issue with all 3 categories of labels (`aspect: *`, `goal: *` and `priority: *`)
  - if a fully-labelled linked issue is found, apply the labels from that issue on the PR
- else, apply the `🚦 status: awaiting triage` label.

## Testing Instructions

GitHub actions are notoriously hard to test. But the Python script can be tested locally.

0. Populate the `.env` file with a GH access token in the `ACCESS_TOKEN` variable.

1. Run the script as 
```
pipenv run python label_pr.py --pr-url <any PR url that ACCESS_TOKEN can access>
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/label_pr.yml
+++ b/.github/workflows/label_pr.yml
@@ -1,0 +1,34 @@
+on:
+  repository_dispatch:
+    types:
+      - label_pr
+
+env:
+  ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+  PR_URL: ${{ github.event.client_payload.pr_url }}
+
+jobs:
+  label_pr:
+    name: Label PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.9"
+
+      - name: Install dependencies
+        working-directory: ./python
+        run: |
+          python -m pip install --user --upgrade pip
+          python -m pip install --user pipenv
+          pipenv install --deploy
+
+      - name: Label PR
+        working-directory: ./python
+        run: |
+          pipenv run python label_pr.py \
+            --pr-url $PR_URL
--- a/python/label_pr.py
+++ b/python/label_pr.py
@@ -125,26 +125,22 @@ def main():
     linked_issues = get_linked_issues(pr.html_url)
     log.info(f"Found {len(linked_issues)} linked issues")
 
-    is_labelled = False
-    if linked_issues:
-        for issue in linked_issues:
-            issue = get_issue(gh, issue)
-            labels = issue.labels
-            if all(
-                [
-                    (aspect := get_label_of_cat("aspect", labels)),
-                    (priority := get_label_of_cat("priority", labels)),
-                    (goal := get_label_of_cat("goal", labels)),
-                ]
-            ):
-                log.info(f"Aspect label: {aspect}")
-                log.info(f"Priority label: {priority}")
-                log.info(f"Goal label: {goal}")
-                pr.set_labels(aspect, priority, goal)
-                is_labelled = True
-                break
-
-    if not is_labelled:
+    for issue in linked_issues:
+        issue = get_issue(gh, issue)
+        labels = issue.labels
+        if all(
+            [
+                (aspect := get_label_of_cat("aspect", labels)),
+                (priority := get_label_of_cat("priority", labels)),
+                (goal := get_label_of_cat("goal", labels)),
+            ]
+        ):
+            log.info(f"Aspect label: {aspect}")
+            log.info(f"Priority label: {priority}")
+            log.info(f"Goal label: {goal}")
+            pr.set_labels(aspect, priority, goal)
+            break
+    else:
         log.info("Could not find properly labelled issue")
         pr.set_labels("🚦 status: awaiting triage")
 
--- a/python/label_pr.py
+++ b/python/label_pr.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+import argparse
+import logging
+import re
+from typing import Optional
+
+import requests
+from bs4 import BeautifulSoup
+from github import Github, Issue, Label, PullRequest
+from shared.data import get_data
+from shared.github import get_client
+from shared.log import configure_logger
+
+
+log = logging.getLogger(__name__)
+
+# region argparse
+parser = argparse.ArgumentParser(description="")
+parser.add_argument(
+    "--pr-url",
+    dest="pr_url",
+    metavar="pr-url",
+    required=True,
+    help="the URL for the PR to label and/or check for labels",
+)
+
+
+# endregion
+
+
+def get_owner_repo_num(html_url: str) -> tuple[str, str, int]:
+    """
+    Get the owner and repo name and issue/PR number from the HTML URL.
+
+    :param html_url: the URL from which to get the owner and repo name and issue/PR num
+    :return: the owner and repo name and the issue/PR number
+    :raise: ``ValueError``, if the data URL does not match the pattern
+    """
+
+    pattern = re.compile(
+        r"https://github.com/(?P<owner>.+)/(?P<repo>.+)/(issues|pull)/(?P<num>\d+)"
+    )
+    if (match := pattern.match(html_url)) is not None:
+        return match.group("owner"), match.group("repo"), int(match.group("num"))
+    raise ValueError("Could not identify owner and repo name and issue/PR number")
+
+
+def get_issue(gh: Github, html_url: str) -> Issue:
+    """
+    Get the ``Issue`` instance from a GitHub issue's HTML URL.
+
+    :param gh: the GitHub client
+    :param html_url: the HTML URL of the issue
+    :return: the ``Issue`` instance corresponding to the given URL
+    """
+
+    org, repo, issue_num = get_owner_repo_num(html_url)
+    return gh.get_organization(org).get_repo(repo).get_issue(issue_num)
+
+
+def get_pull_request(gh: Github, html_url: str) -> PullRequest:
+    """
+    Get the ``PullRequest`` instance from a GitHub PR's HTML URL.
+
+    :param gh: the GitHub client
+    :param html_url: the HTML URL of the PR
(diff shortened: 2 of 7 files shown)


python/label_pr.py
```diff
@@ -0,0 +1,153 @@
...
+        for issue in linked_issues:
+            issue = get_issue(gh, issue)
+            labels = issue.labels
+            if all(
+                [
+                    (aspect := get_label_of_cat("aspect", labels)),
+                    (priority := get_label_of_cat("priority", labels)),
+                    (goal := get_label_of_cat("goal", labels)),
+                ]
+            ):
```
AetherUnbound, line 139: Clever!

python/label_pr.py
```diff
@@ -0,0 +1,153 @@
...
+                log.info(f"Aspect label: {aspect}")
+                log.info(f"Priority label: {priority}")
+                log.info(f"Goal label: {goal}")
+                pr.set_labels(aspect, priority, goal)
+                is_labelled = True
+                break
+
+    if not is_labelled:
+        log.info("Could not find properly labelled issue")
+        pr.set_labels("🚦 status: awaiting triage")
```
AetherUnbound, line 149: I _think_ this could be replaced with a [`for...else` loop](https://book.pythontips.com/en/latest/for_-_else.html) (we may need to default `linked_issues` to an empty list)

```suggestion
    for issue in linked_issues:
        issue = get_issue(gh, issue)
        labels = issue.labels
        if all(
            [
                (aspect := get_label_of_cat("aspect", labels)),
                (priority := get_label_of_cat("priority", labels)),
                (goal := get_label_of_cat("goal", labels)),
            ]
        ):
            log.info(f"Aspect label: {aspect}")
            log.info(f"Priority label: {priority}")
            log.info(f"Goal label: {goal}")
            pr.set_labels(aspect, priority, goal)
            break
    else:
        log.info("Could not find properly labelled issue")
        pr.set_labels("🚦 status: awaiting triage")
```

AetherUnbound (commented):
I like it! So with this new flow, I _should not_ add labels to a PR when creating it from the `gh` CLI, is that correct? (since they'll be added automatically).

It also seems like the PR check will pass if we have `awaiting triage` as a label on the PR, is that the outcome we want? I would assume that the label check should fail if that's the case in order to alert the submitter that the labels should be updated (in case a linked issue couldn't be found). 

Frustrating too that the API doesn't have a easy way to get linked issues!

.github/workflows/label_pr.yml
```diff
@@ -0,0 +1,34 @@
+on:
+  repository_dispatch:
+    types:
+      - label_pr
+
+env:
+  ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
sarayourfriend, line 7: Could we use `github.token` for this (or `GITHUB_TOKEN`) instead? Or does that not have enough permissions? I think it should :thinking: Would be nice to get rid of a secret to manage is all.

python/label_pr.py
```diff
@@ -0,0 +1,153 @@
...
+    :param url: the URL to the GitHub UI for a PR
+    :return: the list of HTML URLs for linked issues
+    """
+
+    res = requests.get(url)
+    if res.status_code != 200:
+        return []
+
+    soup = BeautifulSoup(res.text, "html.parser")
+    divs = soup.find_all("div", **{"class": re.compile("css-truncate my-1")})
```
sarayourfriend, line 89: Very obscure selector, this one! I'm surprised it's reliable but it does seem to be based on my own poking around the page. Hopefully they don't change the margins :stuck_out_tongue:

sarayourfriend (approved):
This is great! I love the solution you picked and impressed by the whole repository dispatch thing, I had no idea that was possible. Glad our contributors won't have to worry about this any more.

zackkrida (approved):
This is so awesome! One question: it's not clear to me when the checks will fail now. It seems like in the case where there _isn't_ a linked PR, and the user doesn't add labels manually, we'd want the 'awaiting triage' label added _and_ for the checks to fail, so that a maintainer could update the PR with correct labels.

sarayourfriend:
> and the user doesn't add labels manually, we'd want the 'awaiting triage' label added and for the checks to fail, so that a maintainer could update the PR with correct labels.

I suppose the problem with that is that new contributors who don't have write access to the repo aren't able to take action on the failing checks, which can be frustrating and confusing.

I wonder if we shouldn't disable checks altogether on fork PRs until a maintainer has removed the awaiting triage label and added the correct labels.

zackkrida:
> I suppose the problem with that is that new contributors who don't have write access to the repo aren't able to take action on the failing checks, which can be frustrating and confusing.

True. As long as we don't end up with any unlabeled PRs getting merged, which we can monitor, this should be okay as-is.

> I wonder if we shouldn't disable checks altogether on fork PRs until a maintainer has removed the awaiting triage label and added the correct labels.

That's a really compelling idea! Could be much less stressful for contributors.

.github/workflows/label_pr.yml
dhruvkb: We should make a separate issue to remove the `ACCESS_TOKEN` from all the workflows it's used it and use `GITHUB_TOKEN` in all those places.

python/label_pr.py
dhruvkb: Thanks!

.github/workflows/label_pr.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: Label PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.9"
```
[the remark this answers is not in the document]
dhruvkb: Nothing specific, I've just not tested the code to work on 3.10 as I'm using 3.9 locally.

.github/workflows/label_pr.yml
```diff
@@ -0,0 +1,34 @@
+on:
+  repository_dispatch:
+    types:
+      - label_pr
```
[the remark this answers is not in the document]
dhruvkb: Yeah, I love that this is possible but there's also a lot of things missing, notably the ability to callback at the end of a workflow.

python/label_pr.py
dhruvkb: It is indeed very random, I came across this hack on a forum thread and was very skeptical that a margin selector would be specific enough (but it works perfectly so 🤷 ).

python/label_pr.py
```diff
@@ -0,0 +1,153 @@
...
+
+    org, repo, pr_num = get_owner_repo_num(html_url)
+    return gh.get_organization(org).get_repo(repo).get_pull(pr_num)
+
+
+def get_linked_issues(url: str) -> list[str]:
+    """
+    Get the list of linked issues from the GitHub UI by parsing the HTML. This is a
+    workaround because GitHub API does not provide a reliable way to find the linked
+    issues for a PR.
```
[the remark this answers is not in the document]
AetherUnbound: Ah good point, I've done that a few times myself

.github/workflows/label_pr.yml
dhruvkb: Created #182.

dhruvkb:
So the idea is that the checks should only fail when someone has the ability to fix them.
- contributors cannot label their PRs so those checks failing for them doesn't make sense
- maintainers can alter labels so failing checks for them has value

So PRs will always be labelled automatically on creation (yay!), either correctly (if linked to an issue) or as '🚦 status: awaiting triage'. For a contributor who cannot edit labels, checks will pass in either case.

As a maintainer, you can edit the labels if the default values are not right according to you, by changing the labels, you can end up in an invalid state (incomplete state but not awaiting triage) so the checks will fail for you.

python/label_pr.py
dhruvkb: Great use of `for`-`else`. Thanks!

.github/workflows/label_pr.yml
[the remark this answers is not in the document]
dhruvkb: Created #183 to tackle this.

dhruvkb:
I'm merging this to get the repository dispatch into production, let's not merge the sync PRs right away, as there might be changes needed.

python/label_pr.py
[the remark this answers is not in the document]
sarayourfriend: Ah, did not consider that!

sarayourfriend:
> True. As long as we don't end up with any unlabeled PRs getting merged, which we can monitor, this should be okay as-is.

To be fair @zackkrida none of our CI workflows are required so the situation there has always been precarious and up to the responsibility of the people merging PRs.

PR #202, opened 2022-03-25
Update feature_request.md label template to remove priority and aspect

zackkrida:
Removes priority and aspect labels from the feature request template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,7 +1,7 @@
 ---
 name: Feature request
 about: Suggest a new feature or improvements to an existing one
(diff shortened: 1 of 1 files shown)


sarayourfriend (approved):
LGTM. At this point, since we are removing these, even `improvement` could be dubious considering it could be an "internal" improvement as well :slightly_smiling_face:

AetherUnbound (approved):
Agreed on the improvement note, I think that can go too :slightly_smiling_face:

zackkrida:
I'm fine with removing improvement, just trying to think of how we could reasonably get a list of all of our feature requests without it, if we ever wanted that 🤔

dhruvkb:
I vote for keeping the "goal: improvement" label, hard to assume an outsider would be raising an internal improvement feature request. If a contributor were opening it, they might be able to change the label to whatever they deem appropriate.

PR #251, opened 2022-06-07
Update weekly_report.js to remove old auth token

zackkrida:
This script isn't being used but I think it'd be best to remove this old, deleted key.

### DIFF
--- a/automations/js/src/weekly_report.js
+++ b/automations/js/src/weekly_report.js
@@ -21,7 +21,7 @@ if (!date) {
 }
 
 // get auth key
-const key = process.argv?.[3]
+const ghKey = process.argv?.[3]
 if (!key) {
   throw new Error(
     'Please supply a GitHub access token with public repo permissions'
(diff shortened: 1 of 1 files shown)


sarayourfriend (approved):

automations/js/src/weekly_report.js
```diff
@@ -20,6 +20,14 @@ if (!date) {
   )
 }
 
+// get auth key
+const key = process.argv?.[3]
```
stacimc, line 24: Should this be `ghKey` as used below?

obulat (approved):

automations/js/src/weekly_report.js
zackkrida, line 24:
```suggestion
const ghKey = process.argv?.[3]
```

automations/js/src/weekly_report.js
```diff
@@ -20,6 +20,14 @@ if (!date) {
   )
 }
 
+// get auth key
+const ghKey = process.argv?.[3]
+if (!key) {
```
zackkrida, line 25:
```suggestion
if (!ghKey) {
```

automations/js/src/weekly_report.js
zackkrida: yes, thanks

PR #257, opened 2022-06-09
Add basic API load testing script

sarayourfriend:
## Description

This adds a basic locust load testing script for the API. It's pretty naive but the idea is to just hit the API very hard and force it to fall over. So far I haven't been able to make it actually fall over, maybe the user count needs to go up into the hundreds. I also haven't tested against v2.5.3 which was the problematic version.

## Testing Instructions

Checkout this branch and `cd load_testing`. Run `just install` and then `just api` (make sure you have the API running locally). Confirm via `docker stats` that the API is being hit with a bunch of requests. `just api` won't exit on it's own so use "control + c" to stop it or pass a time argument like `just api -t 10m` (this will run the load test for 10 minutes).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ env/
 .vscode/
 
 grafana.db
+__pycache__
--- a/load_testing/Dockerfile
+++ b/load_testing/Dockerfile
@@ -1,0 +1,18 @@
+FROM debian:11-slim as words
+
+RUN apt-get update && apt-get install -y wamerican
+
+FROM locustio/locust:2.9.0
+
+COPY --from=words /usr/share/dict/american-english /usr/share/dict/american-english
+
+RUN pip install --upgrade pipenv
+
+COPY Pipfile .
+COPY Pipfile.lock .
+
+COPY Pipfile Pipfile.lock ./
+
+RUN pipenv install --system --deploy --dev
+
+ENTRYPOINT ["locust", "--headless"]
--- a/load_testing/Pipfile
+++ b/load_testing/Pipfile
@@ -1,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+locust = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.10"
--- a/load_testing/Pipfile.lock
+++ b/load_testing/Pipfile.lock
@@ -1,0 +1,632 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "a124a17edc02ce2985b18b005ecfb452ee08179c30df252ed60f11d86da314a0"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.10"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "brotli": {
+            "hashes": [
+                "sha256:12effe280b8ebfd389022aa65114e30407540ccb89b177d3fbc9a4f177c4bd5d",
+                "sha256:160c78292e98d21e73a4cc7f76a234390e516afcd982fa17e1422f7c6a9ce9c8",
+                "sha256:16d528a45c2e1909c2798f27f7bf0a3feec1dc9e50948e738b961618e38b6a7b",
+                "sha256:19598ecddd8a212aedb1ffa15763dd52a388518c4550e615aed88dc3753c0f0c",
+                "sha256:1c48472a6ba3b113452355b9af0a60da5c2ae60477f8feda8346f8fd48e3e87c",
+                "sha256:268fe94547ba25b58ebc724680609c8ee3e5a843202e9a381f6f9c5e8bdb5c70",
+                "sha256:269a5743a393c65db46a7bb982644c67ecba4b8d91b392403ad8a861ba6f495f",
+                "sha256:26d168aac4aaec9a4394221240e8a5436b5634adc3cd1cdf637f6645cecbf181",
+                "sha256:29d1d350178e5225397e28ea1b7aca3648fcbab546d20e7475805437bfb0a130",
+                "sha256:2aad0e0baa04517741c9bb5b07586c642302e5fb3e75319cb62087bd0995ab19",
+                "sha256:3496fc835370da351d37cada4cf744039616a6db7d13c430035e901443a34daa",
+                "sha256:35a3edbe18e876e596553c4007a087f8bcfd538f19bc116917b3c7522fca0429",
+                "sha256:3b78a24b5fd13c03ee2b7b86290ed20efdc95da75a3557cc06811764d5ad1126",
+                "sha256:40d15c79f42e0a2c72892bf407979febd9cf91f36f495ffb333d1d04cebb34e4",
+                "sha256:44bb8ff420c1d19d91d79d8c3574b8954288bdff0273bf788954064d260d7ab0",
+                "sha256:4688c1e42968ba52e57d8670ad2306fe92e0169c6f3af0089be75bbac0c64a3b",
+                "sha256:495ba7e49c2db22b046a53b469bbecea802efce200dffb69b93dd47397edc9b6",
+                "sha256:4d1b810aa0ed773f81dceda2cc7b403d01057458730e309856356d4ef4188438",
+                "sha256:503fa6af7da9f4b5780bb7e4cbe0c639b010f12be85d02c99452825dd0feef3f",
+                "sha256:56d027eace784738457437df7331965473f2c0da2c70e1a1f6fdbae5402e0389",
+                "sha256:5913a1177fc36e30fcf6dc868ce23b0453952c78c04c266d3149b3d39e1410d6",
+                "sha256:5b6ef7d9f9c38292df3690fe3e302b5b530999fa90014853dcd0d6902fb59f26",
+                "sha256:5cb1e18167792d7d21e21365d7650b72d5081ed476123ff7b8cac7f45189c0c7",
+                "sha256:61a7ee1f13ab913897dac7da44a73c6d44d48a4adff42a5701e3239791c96e14",
+                "sha256:622a231b08899c864eb87e85f81c75e7b9ce05b001e59bbfbf43d4a71f5f32b2",
+                "sha256:68715970f16b6e92c574c30747c95cf8cf62804569647386ff032195dc89a430",
+                "sha256:6b2ae9f5f67f89aade1fab0f7fd8f2832501311c363a21579d02defa844d9296",
+                "sha256:6c772d6c0a79ac0f414a9f8947cc407e119b8598de7621f39cacadae3cf57d12",
+                "sha256:6d847b14f7ea89f6ad3c9e3901d1bc4835f6b390a9c71df999b0162d9bb1e20f",
+                "sha256:76ffebb907bec09ff511bb3acc077695e2c32bc2142819491579a695f77ffd4d",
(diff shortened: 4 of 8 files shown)


sarayourfriend:
> since those result in direct DB reads instead of only hitting Elasticsearch, and would also be an easy way to exponentially increase the amount of traffic. Maybe for each search the script could then visit each single result from the response?

Very good call! Will add this.

sarayourfriend:
Curiously walking through and requesting all the returned results actually causes _less_ stress to the API and ES but _did_ stress the database! I'm not sure why this is, I guess a search request uses more CPU from the API than a single result (which is just a pass through to Postgres and marshalling the DB response to JSON, at the end of the day). Very interesting! I will keep tinkering.

sarayourfriend:
Okay, I think part of the problem is that I'm writing a single task that does all these things:

1. Query search endpoint for 10 random words
2. Iterate through the results and query each one and it's related and thumbnail

I suspect that doing all of that in a single task is a problem somehow? For some reason I'm only getting requests to the stats endpoints to actually run. It seems like maybe locust isn't distributing the users between the requests or something? I've been scouring the docs to try and find an answer to this but I haven't found anything helpful there on this topic so far :confused: 

I'll come back to this tomorrow :+1:

sarayourfriend:
@AetherUnbound can you confirm whether this is still needed?

AetherUnbound:
I think this is definitely worth continuing! The script I wrote is merely testing API response time, it's not doing accurate or heavy load testing whatsoever (at 1 request/min). So a tool that can throw a lot more traffic at the service would still be useful IMO 😄

sarayourfriend:
Sounds good. It's probably not high priority anymore though right? Because we don't need this to help reproduce the production slow downs?

AetherUnbound:
Yea that's true, it can probably be dropped in priority level 🙂

sarayourfriend:
I still haven't figured out why Locust isn't making some of the requests in the tasks I've created. It's very odd :thinking: I can't find any information to debug with either... Maybe someone else could see if they're seeing the same thing I am, where running `just api` only makes the stats requests and never makes the API query requests?

If I can't figure it out I'll probably try switching to some other load testing tool. Locust has a whole bunch of caveats too as a Python app due to the GIL. Having to spin up workers and such is a big pain. [Gatling](https://gatling.io/docs/gatling/) seems more straightforward despite being a Scala DSL.

sarayourfriend:
I did more research on some tools over the weekend. I think we could get away with using `ab` (https://httpd.apache.org/docs/2.4/programs/ab.html) for this. Alternatively, we could try using Erlang's `Tsung` library, which is configured using XML, but is very fast and has none of the problems of Locust or Artillery or the other solutions written in woefully limited languages (when it comes to this kind of thing).

I'm going to start by trying `ab` out as I think it's likely the simplest solution. We'll still run it in a docker container so we don't have to worry about installing tools like that locally and to make it easier to potentially run this scheduled in CI against the dev API or something like that in the future.

sarayourfriend:
I'm kind of stumped at the moment. The `ab` command running in the `api.sh` script doesn't work. If I exec into the container, however, and run the command manually (retrieve the desired command by echoing the command in `api.sh` instead of running it) it works. I feel like there must be some bash edge case that I'm not aware of here that is causing it not to run the way I expected...

Going to come back to this later today or tomorrow.

sarayourfriend:
I still haven't been able to figure this out. Can someone else on @WordPress/openverse-maintainers take a look and see if they can reproduce the "too many arguments" error I see and maybe spend a few minutes debugging? I feel like there's some obvious bash/docker problem that I'm just not aware of and have had no luck finding in my research on the topic.

sarayourfriend:
You're seeing what I'm seeing. If you echo the `ab` command in the `api.sh` script (you need to have the `env.sh` setup with a key based on the template btw, I'll add this to the PR description) then run it in your local terminal (or exec into the container and run it if you don't have apache-bench installed locally) then it works fine, but running it inside a shell script produces the too many arguments error or it silently errors out and never makes any network requests or anything. This latter thing, which is happening now, is actually still a "too many arguments" error but it's getting swallowed up somehow :thinking: Maybe `set -xe` would help debug this now that I think of it.

sarayourfriend:
I was thinking about this (accidentally) on a recent long flight from Australia to the US and realized that `ab` isn't going to work for our purposes unless we run the load testing script against the ALB DNS name. However, luckily, I think it should work because [we have a wide open CORS policy in the API](https://github.com/WordPress/openverse-api/blob/c2df644e5faec7013412080a0a258117ae4ccaf8/api/catalog/settings.py#L280).

That way we'll bypass the API's caching in Cloudflare and won't need to worry about the query string being the same for all `n` requests that `ab` makes.

That is... if we can figure out how to get `ab` working at all. When I'm back from vacation[^1] I will write up everything I've tried with `ab` and give clearer description for what I've tried. Sorry @stacimc that I didn't do that earlier.

[^1]: I'm writing this now because I am worried I would forget it... after 15 hours on a plane my grasp of reality is merely tenuous and any and all ideas are liable to slip away without a trace.

sarayourfriend:
@stacimc I updated the README with more instructions for the environment setup. Let me know if you feel anything is still missing. I left out the specific steps for updating the Redis set for the live environment because it's all based on implementation details of our particular current deployment and didn't seem appropriate for this document.

stacimc:
Okay, I've spent a bunch of time playing around with this -- a lot of my problems came from having the API checked out at the ingestion server refactor, which apparently caused issues 😓

Long story short I finally got it working for me by changing the `$host` to `host.docker.internal`, and then adding the same to [ALLOWED_HOSTS](https://github.com/WordPress/openverse-api/blob/main/api/catalog/settings.py#L47) in the api. I'm a little confused about whether this deviates from the intention/if I'm doing something wrong here, though?

So, the current state is that `just api` will make 100 requests to `v1/images` with `page_size=500` and a random word, and then exit. This is now working for me too 🎉

load_testing/api.sh
```diff
@@ -0,0 +1,31 @@
+#! /usr/bin/env bash
+
+set -e
+
+source ./.env.sh || (echo 'Please create a `.env.sh` file based on the `.env.sh.template` file.' > /dev/stderr && false)
+
+host=${1:-http://localhost:8080}
```
stacimc, line 7: Why is `http://localhost:8080` used as a default here, but `http://172.17.0.1:8000` in the justfile?

load_testing/.env.sh.template
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env sh
+
+export ACCESS_TOKEN = "<not_set>"
```
zackkrida, line 3:
```suggestion
export ACCESS_TOKEN="<not_set>"
```

If I copy this file and add the key, the spaces result in an `./.env.sh: line 3: export: `=': not a valid identifier` error.

load_testing/api.sh
sarayourfriend: It should probably just require that a host name is passed instead of duplicating the default, right?

To address the issue you had where you have to change the host to `host.docker.internal` anyway, I'm going to add platform detection to the justfile and determine the correct "localhost" default to use from there and remove this one from the script.

sarayourfriend:
> Long story short I finally got it working for me by changing the $host to host.docker.internal, and then adding the same to [ALLOWED_HOSTS](https://github.com/WordPress/openverse-api/blob/main/api/catalog/settings.py#L47) in the api. I'm a little confused about whether this deviates from the intention/if I'm doing something wrong here, though?

That's correct. I'm not sure what to do about the ALLOWED_HOSTS. Why do we restrict them like that in the API anyway, I wonder... or at least not make it an environment variable that is configured in the specific deployment?

zackkrida:
I also got this working with @stacimc's suggestion to switch to and allow `host.docker.internal`. Here's the resulting output.html:

```html
<p>
 This is ApacheBench, Version 2.3 <i>&lt;$Revision: 1879490 $&gt;</i><br>
 Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/<br>
 Licensed to The Apache Software Foundation, http://www.apache.org/<br>
</p>
<p>
INFO: GET header == 
---
GET /v1/images/?q=spoonbill%27s&page_size=500 HTTP/1.0
Authorization: Bearer Xu62j3LuLslGLIwpq4zhbmRDmL2NDg
Host: host.docker.internal:8000
User-Agent: ApacheBench/2.3
Accept: */*


---
..done


<table >
<tr ><th colspan=2 bgcolor=white>Server Software:</th><td colspan=2 bgcolor=white></td></tr>
<tr ><th colspan=2 bgcolor=white>Server Hostname:</th><td colspan=2 bgcolor=white>host.docker.internal</td></tr>
<tr ><th colspan=2 bgcolor=white>Server Port:</th><td colspan=2 bgcolor=white>8000</td></tr>
<tr ><th colspan=2 bgcolor=white>Document Path:</th><td colspan=2 bgcolor=white>/v1/images/?q=spoonbill%27s&page_size=500</td></tr>
<tr ><th colspan=2 bgcolor=white>Document Length:</th><td colspan=2 bgcolor=white>0 bytes</td></tr>
<tr ><th colspan=2 bgcolor=white>Concurrency Level:</th><td colspan=2 bgcolor=white>4</td></tr>
<tr ><th colspan=2 bgcolor=white>Time taken for tests:</th><td colspan=2 bgcolor=white>0.183 seconds</td></tr>
<tr ><th colspan=2 bgcolor=white>Complete requests:</th><td colspan=2 bgcolor=white>100</td></tr>
<tr ><th colspan=2 bgcolor=white>Failed requests:</th><td colspan=2 bgcolor=white>0</td></tr>
<tr ><th colspan=2 bgcolor=white>Total transferred:</th><td colspan=2 bgcolor=white>0 bytes</td></tr>
<tr ><th colspan=2 bgcolor=white>HTML transferred:</th><td colspan=2 bgcolor=white>0 bytes</td></tr>
<tr ><th colspan=2 bgcolor=white>Requests per second:</th><td colspan=2 bgcolor=white>546.75</td></tr>
<tr ><th colspan=2 bgcolor=white>Transfer rate:</th><td colspan=2 bgcolor=white>0.00 kb/s received</td></tr>
<tr ><th bgcolor=white colspan=4>Connection Times (ms)</th></tr>
<tr ><th bgcolor=white>&nbsp;</th> <th bgcolor=white>min</th>   <th bgcolor=white>avg</th>   <th bgcolor=white>max</th></tr>
<tr ><th bgcolor=white>Connect:</th><td bgcolor=white>    0</td><td bgcolor=white>    0</td><td bgcolor=white>    3</td></tr>
<tr ><th bgcolor=white>Processing:</th><td bgcolor=white>    1</td><td bgcolor=white>    1</td><td bgcolor=white>   10</td></tr>
<tr ><th bgcolor=white>Total:</th><td bgcolor=white>    1</td><td bgcolor=white>    1</td><td bgcolor=white>   13</td></tr>
</table>

````

sarayourfriend:
Another tool we could use that looks great and appears to easily support multiple workers without an issue (virtual users in this tool's parlance): https://github.com/grafana/k6

zackkrida (approved):
The cross-platform script worked, lgtm!

stacimc:
The cross-platform fix also works for me :) It might be nice to add a comment about updating `ALLOWED_HOSTS` in the api in the README.

Can you also update the PR description? There's been a lot of fluctuation because of the technical difficulties and I'm not sure what the exact goals of the PR are for now.

stacimc:
No worries, I do that all the time 😄 And big plus one to the approach. LGTM!

stacimc (approved):

PR #279, opened 2022-07-19
Add k6 load testing script

### LINKED ISSUE #278: K6 based API load testing script
## Problem

In https://github.com/WordPress/openverse/pull/257 we introduced an overly-simplistic load testing script based on `ab`. This is insufficient to adequately load the API to simulate actual production traffic.

## Description

To replace this, let's create a set of [k6 VU scripts](https://github.com/grafana/k6) to attack the staging API.

Off the top of my head, we could have VU scripts that do the following:
1. Request a random word at page_size=20 and then request all the pages and individual documents that exist for it including requesting the related results for each. Do this for all media types.
2. Same as the first entry but with page_size=500.
3. Request each individual provider using the source query param. Paginate through the provider completely until the API returns an error about deep pagination. Do this with page_size=20 and for all media types.
4. The same as the third entry but with page_size=500.

## Alternatives

We could use an alternative load testing scripting solution like Tsung or Gatling. Tsung uses XML however and do not allow programmatic execution (scripts must be defined before hand). Gatling uses JVM (Scala, Kotlin, or Java, your pick) and has a nice programmatic API, but will require a heavier docker container set up and is more complicated. It also introduced a new language to our stack that we could avoid with k6.

## Implementation

krysal:
## Fixes

Fixes #278 by @sarayourfriend 

## Description

Creates the scenarios described in the issue.

### Locally running the API

Remember to add `host.docker.internal` to `ALLOWED_HOSTS` in the API.

```sh
just k6 "-e API_TOKEN=<your_token> -e API_URL=localhost:8000"
```

> **Warning**
> The container can not be built on macOS given that  "repository 'https://dl.k6.io/deb stable InRelease' doesn't support architecture 'arm64'".

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
Okay! I've added the other permutations described as part of the scenarios. It really throws a lot at the API 😳 I feel like I'm making things a bit too complex with how I'm setting up the various scenarios 🤔 let me know if there are some obvious ways to make things simpler, I'm sure I'm doing some js-newbie wacky stuff!

I'm also not sure where to go next with this 😅 So I'm open to thoughts on next steps! CC @WordPress/openverse-maintainers

automations/js/package-lock.json
```diff
@@ -1,310 +1,6 @@
...
-        "webidl-conversions": "^3.0.0"
-      }
-    },
-    "node_modules/wrappy": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/wrappy/-/wrappy-1.0.2.tgz",
-      "integrity": "sha1-tSQ9jz7BqjXxNkYFvA0QNuMKtp8="
-    }
-  },
+  "lockfileVersion": 1,
```
sarayourfriend, line 3: It looks like you might be using a different (lower) version of `npm` than was being used before :thinking: 

I wonder if we could just switch to `pnpm` like we use in the frontend repository anyway... `package-lock.json` is a hot mess :disappointed: 

Opened this issue to fix this: https://github.com/WordPress/openverse/issues/281

load_testing/k6/search.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    sleep(SLEEP_DURATION);
+
+    group("Related requests", () => {
+        console.info(
+            `Requesting all ${media_type} related from "${param}" at page ${page}`
+        );
+        const responses = http.batch(getUrlBatch(relatedUrls, "related_url"));
+        extraFailed = responses.map((r) => responseFailed(r, "related")).some(f => f)
```
sarayourfriend, line 42: What is `extraFailed` for? I see it assigned in both groups but I don't see it used anywere unless I am missing it.

sarayourfriend (commented):
Maybe we should move the `package-lock` and `package` json files up to the root of the repository so that `load_testing` also gets to use them? Otherwise the "installation" process is kind of strange... 

We should also update the `pre-commit-config.yaml` to include the new JS files in the prettier hook.

Overall this is very cool though! The code and approach looks good to me.

Should we continue to try to run these inside of a docker container so that we don't have to install k6 locally? It would ease being able to run these on GitHub actions or something as part of future automations (and locally as well, I prefer downloading a docker image to having to follow instructions to install a binary tool, when possible).

obulat:
This is really great! I managed to run this script and bombard my local API with requests. I think I did something wrong when setting up the app's throttled class because all `page_size=500` requests errored out with 401.

It would be nice to improve the documentation and use `just` for running this script in Docker, as @sarayourfriend suggested.

krysal:
Thanks a lot for picking up this while I was AFK @AetherUnbound! 😄 There is a [docker image for k6](https://hub.docker.com/r/grafana/k6/), so I'll work on adding it here.

load_testing/justfile
```diff
@@ -6,7 +6,10 @@ build:
     docker build . -t openverse-load-testing:latest
 
 run script host: build
-    docker run --rm -it -v {{invocation_directory()}}:/app openverse-load-testing:latest ./{{ script }}.sh {{ host }}
+    docker run --rm -it -v {{ invocation_directory() }}:/app openverse-load-testing:latest ./{{ script }}.sh {{ host }}
 
 api host=("http://" + DOCKER_HOST + ":8000"):
     @just run api {{ host }}
+
+k6 env="-e API_URL=https://api-dev.openverse.engineering/v1/": build
```
sarayourfriend, line 14: Given we also have to pass in the `API_TOKEN` (otherwise the page_size=500 requests will fail), it could be nice for environment variables to be explicitly given as inputs rather than having to pass them individually through `-e` docker run syntax.

What do you think? Or maybe we should encourage the use of a .env file for the justfile? Just loads .env's automatically from what I understand as long as we update `set dotenv-load := false` to `set dotenv-load := true`.

krysal: I like the use of a .env file and made those changes, that is what you were thinking with "enforcing" it?

sarayourfriend: Yes! That works great. Would we want to pick up the API_URL from there as well? :thinking:

load_testing/justfile
```diff
@@ -1,12 +1,15 @@
...
 
 run script host: build
-    docker run --rm -it -v {{invocation_directory()}}:/app openverse-load-testing:latest ./{{ script }}.sh {{ host }}
+    docker run --rm -it -v {{ invocation_directory() }}:/app openverse-load-testing:latest ./{{ script }}.sh {{ host }}
 
 api host=("http://" + DOCKER_HOST + ":8000"):
     @just run api {{ host }}
+
+k6 env="-e API_TOKEN=$API_TOKEN": build
+    docker run --rm {{ env }} -v {{ invocation_directory() }}/k6:/app openverse-load-testing:latest k6 run main.js
```
sarayourfriend, line 15: This might be a slightly easier to use (more flexible) version:

```suggestion
k6 url="https://api-dev.openverse.engineering" +extra_args="": build
    docker run --rm -e API_TOKEN=$API_TOKEN -e API_URL={{ url }} -v {{ invocation_directory() }}/k6:/app {{ extra_args }} openverse-load-testing:latest k6 run main.js
```

What do you think?

sarayourfriend (approved):
LGTM! Using this script I was able to get the staging elastic search datanodes to all reach about 95% CPU utilization! Fantastic!

load_testing/justfile
krysal: I guess that the `env` can already serve to pass multiple arguments and not just environment variables 🤔 it just needs to be enclosed in "" like the code example in the first comment, but we could also pass arguments to k6 with:

```sh
k6 docker_args="-e API_TOKEN=$API_TOKEN" +k6_args="": build
    docker run --rm {{ docker_args }} -v {{ invocation_directory() }}/k6:/app openverse-load-testing:latest k6 run main.js {{ k6_args }}
```

Does it look better like this?

sarayourfriend: Well, if you decide to pass any extra args now, you'll have to also pass the API_TOKEN arg. Given that one _always_ needs to be passed, I was thinking we could remove it from the potential to be removed. If it needs to be overriden for some reason, I think `-e` arguments will override previous ones with the same name.

krysal: Ahh, I understand, then your suggestion is fine for me!

sarayourfriend (commented):
Getting the extra k6_args option in would be great too, I didn't have that in my suggestion but your version did.

AetherUnbound (approved):
Feels a little weird to approve since I worked on this but ✅ from me 😄

PR #294, opened 2022-08-18
Stop syncing CODEOWNERS files

zackkrida:
Each repository is going to move to being managed by an individual team, so the value in syncing the CODEOWNERS files, which will be unique per-repo, is significantly reduced.

Alternatively, we could track and sync a unique CODEOWNERS file for each repository here, if folks preferred, so they could be updated all at once in the future instead of requiring a discrete PR be made manually for each repo.

### DIFF
--- a/.github/CODEOWNERS_FRONTEND
+++ b/.github/CODEOWNERS_FRONTEND
@@ -1,2 +1,0 @@
-# All files are owned by the Openverse Frontend Developers team
-* @WordPress/openverse-frontend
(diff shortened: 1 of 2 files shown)


sarayourfriend:
> if folks preferred, so they could be updated all at once in the future instead of requiring a discrete PR be made manually for each repo

Zero preference here. Seems unlikely to change that much after this outside of repo specific settings. Even now it has not changed that much outside of refining this process.

sarayourfriend (approved):

AetherUnbound (approved):
Same here, I'm fine with having them in the respective repos.

PR #301, opened 2022-09-14
Add dead link scripts

zackkrida:
Adds the dead link script by @sarayourfriend to this primary repo.


sarayourfriend (commented):
I think we can delete the public suffix list. Otherwise LGTM (but also I wrote it so of course it does :laughing:).

PR #312, opened 2022-10-06
RFC: Frontend event tracking

zackkrida:
This is an RFC which outlines a solution for collecting analytics events on the frontend. Please see the ['files changed' diff](https://github.com/WordPress/openverse/pull/312/files) for the full text.

## Reviewers

- @krysal 
- @dhruvkb 

## Deadline

2022-1-16

### DIFF
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -63,7 +63,10 @@ Unless all conditions are met the `sendEvent` function will simply no-op.
 
 A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
 
-This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
+This proposal will require the creation of some new user interface elements to toggle analytics tracking:
+
+1. A small popup banner that shows up to users, which informs them that Openverse collects anonymized analytics information to make the service better. There are "dismiss" and "learn more" calls to action. "Learn more" directs users to:
+2. An analytics opt-out checkbox on the `/privacy` page. We also add information describing our analytics collection, and link to the relevant code in our GitHub repository.
 
 ### Events
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -257,14 +257,23 @@ async function sendEvent<T extends AnalyticsEventTypes>(
 ): Promise<any> {
   /**
    * The default values to send with every analytics event.
+   * Any sensitive or personally-identifying information
+   * sent here is *not* stored, rather used to derive
+   * anonymized session, device, and browser statistics.
+   *
+   * The items not stored directly include: width, height,
+   * ipAddress, and userAgent
    */
   const defaults = {
     timestamp: Date.now(),
     path: window.location.pathname,
-    // The dimensions of the user's device
-    width: window.innerWidth,
-    height: window.innerHeight,
     referrer: document.referrer || null,
+    language: i18n.locale, // collected from the global Nuxt i18n object
+    // These fields are not stored directly, rather used to derive anonymized values
+    width: window.innerWidth, // The width of the user's device
+    height: window.innerHeight, // The height of the user's device
+    ipAddress: "", // Should be read from the X-Forwarded-For header server-side, will need to be stored securely during the user's session
+    userAgent: navigator.userAgent,
   };
   /**
    * Actually send the event here, with some additional values.
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -4,8 +4,8 @@ A proposal to emit frontend events based on user interactions and page visits.
 
 ## Reviewers
 
-- [ ] @sarayourfriend
-- [ ] @dhruvkb
+- [ ] @krysal - For API experience
+- [ ] @dhruvkb - For frontend and DX experience
 
 ## Rationale
 
@@ -42,15 +42,15 @@ curl -i -X POST https://plausible.io/api/event \
   --data '{"name":"pageview","url":"http://dummy.site","domain":"dummy.site","screen_width":1666}'
 ```
 
-There are many points in the code execution at which we could transform our custom events into ones supported by other platforms, if we need to. By adopting and implementing our own custom solution we can start collecting data as soon as possible without being blocked by time spent evaluating and implementing other analytics systems.
+There are many points in the code execution at which we could transform our custom events into ones supported by other platforms. By adopting and implementing our own custom solution, we can start collecting data as soon as possible without being blocked on time spent evaluating and implementing a full analytics system.
 
 ## The technical details
 
-We should create a system for collecting events on the frontend. Each event will have a unique name, collect some default data about the enviroment (timestamp, referer, and so on) and can have an optional data payload containing string and numerical values.
+We will create a system for collecting events on the frontend. Each event will have a unique name, collect some default data about the enviroment (timestamp, referer, and so on) and can have an optional data payload containing string and numerical values. I'd like to limit the payload shape to strings and numbers to keep it JSON serializable and simple. Allowing for complex nested objects might prove too complex to query in the future. I think this is a reasonable constraint but am open to alternatives.
 
 ### Sending events
 
-These events will be sent as `POST` requests to a server route on the frontend, which will then route them to our desired analytics service. In the future it may make sense to move this server route to a standalone microservice but for now a [Nuxt server route](https://v3.nuxtjs.org/guide/directory-structure/server/#server-routes) should be sufficient to handle the analytics events. This endpoint is essentially a proxy between the frontend and our 'true' analytics service.
+These events will be sent as `POST` requests to a server route on the frontend, which will then route them to our desired analytics service. In the future, it may make sense to move this server route to a standalone microservice, but for now a [Nuxt server route](https://v3.nuxtjs.org/guide/directory-structure/server/#server-routes) should be sufficient to handle the analytics events. This endpoint is essentially a proxy between the frontend and our 'true' analytics service that makes sure we're always sending events server-side.
 
 #### Privacy
 
@@ -67,7 +67,7 @@ This proposal will require the creation of a user settings page _or_ an analytic
 
 ### Sample implementaton and event types
 
-Below I've created a TypeScript type for all allowed anlytics events, and a simple function for sending events. This is where you can see a list of proposed events and their payloads, along with the default values which will be passed to all events.
+Below I've created a TypeScript type for all the proposed anlytics events, and a simple function for sending events. For each event you'll see the payload alongside the default values which will be passed to _all_ events.
 
 ```ts
 // src/utils/events.ts
@@ -163,25 +163,73 @@ type AnalyticsEvents = {
     query: string;
   };
   /** The user clicks the CTA button to the external source to use the image */
-  GET_MEDIA: {};
+  GET_MEDIA: {
+    /** The unique ID of the media */
+    id: string;
+    /** The slug (not the prettified name) of the provider */
+    provider: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
   /** The user clicks one of the buttons to copy the media attribution */
-  COPY_ATTRIBUTION: {};
+  COPY_ATTRIBUTION: {
+    /** The unique ID of the media */
+    id: string;
+    /** The format of the copied attribution */
+    format: "plain" | "rich" | "html";
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
   /** The user reports a piece of media through our form */
-  REPORT_MEDIA: {};
+  REPORT_MEDIA: {
+    /** The unique ID of the media */
+    id: string;
+    /** The slug (not the prettified name) of the provider */
+    provider: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
   /** The user plays, pauses, or seeks an audio track.
    * @todo: This potentially requires throttling.
    **/
-  AUDIO_INTERACTION: {};
+  AUDIO_INTERACTION: {
+    /** The unique ID of the media */
+    id: string;
+    event: "seek" | "play" | "pause";
+    /** The slug (not the prettified name) of the provider */
+    provider: string;
+  };
   /** The user visits a CC license description page on CC.org  **/
-  VISIT_LICENSE_PAGE: {};
+  VISIT_LICENSE_PAGE: {
+    /** The slug of the license the user clicked on */
+    license: string;
+  };
   /** The user visits a creator's link in the single result UI  **/
-  VISIT_CREATOR_LINK: {};
+  VISIT_CREATOR_LINK: {
+    /** The unique ID of the media */
+    id: string;
+    /** The permalink to the creator's profile */
+    url: string;
+  };
   /** The user right clicks a single image result, most likely to download it. **/
-  RIGHT_CLICK_IMAGE: {};
+  RIGHT_CLICK_IMAGE: {
+    /** The unique ID of the media */
+    id: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
   /** The user uses the 'back to search' link on a single result **/
-  BACK_TO_SEARCH: {};
+  BACK_TO_SEARCH: {
+    /** The unique ID of the media */
+    id: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
   /** The visibility of the filter sidebar on desktop is toggled **/
-  TOGGLE_FILTERS: {};
+  TOGGLE_FILTERS: {
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
 };
 type AnalyticsEventTypes = keyof AnalyticsEvents;
 
@@ -226,14 +274,16 @@ async function sendEvent<T extends AnalyticsEventTypes>(
 This needs more elaboration, as it's a very rough outline. For example, specific events will require more detailed planning.
 
 - [ ] Create "Event Tracking" GitHub milestone in the frontend repository
-- [ ] Create a GitHub issue for each of the following todos
+- [ ] Design: design a privacy and analytics opt-out checkbox on the Privacy page
+- [ ] Create an issue for the `ANALYTICS_ENABLED` feature flag
 - [ ] Create an issue for the creation of the 'sendEvent' function and for the api route to post analytics events to
-- [ ] Create a single issue for each event in the sample implementation
+- [ ] Create a single issue for each event in the sample implementation. Each issue will need its specific implementation (when the event is called, what user DOM interaction triggers it, etc.) approved by frontend developers before it is implemented.
 
 ## Concerns / Pitfalls
 
 - Related to content safety, this is an area where we are surfacing user information in the form of text, or even malicious URLs, to Openverse developers. These searches could include explicit or sensitive material. We're also potentially surfacing explicit image results to anyone working with this data.
-- We're building something here that is commonly drop-in when 3rd party tools are used. Is this a poor use of time?
+- I think it's great to build this decoupled from a specific back-end, but if we never actually implement a backend this would largely be a wasted effort.
+- We're building something here that is commonly drop-in when 3rd party tools are used. At least for page views; custom events still require custom work. Is this a poor use of time?
 
 ### Prior Art
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -268,7 +268,7 @@ type AnalyticsEvents = {
     /** The unique ID of the media */
     id: string;
     /** If the result is a related result, provide the ID of the 'original' result */
-    related: string | null;
+    relatedTo: string | null;
     /** The media type being searched */
     mediaType: SearchType;
     /** The slug (not the prettified name) of the provider */
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -63,7 +63,7 @@ Unless all conditions are met the `sendEvent` function will simply no-op.
 
 A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
 
-This proposal will require the creation of a user settings page _or_ a analytics opt-out checkbox on the soon to be created `/privacy` page.
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the soon to be created `/privacy` page.
 
 ### Sample implementaton and event types
 
@@ -162,11 +162,19 @@ type AnalyticsEvents = {
     /** The search term **/
     query: string;
   };
+  /** The user clicks the CTA button to the external source to use the image */
   GET_MEDIA: {};
+  /** The user clicks one of the buttons to copy the media attribution */
   COPY_ATTRIBUTION: {};
+  /** The user reports a piece of media through our form */
   REPORT_MEDIA: {};
+  /** The user plays, pauses, or seeks an audio track.
+   * @todo: This potentially requires throttling.
+   **/
   AUDIO_INTERACTION: {};
+  /** The user visits a CC license description page on CC.org  **/
   VISIT_LICENSE_PAGE: {};
+  /** The user visits a creator's link in the single result UI  **/
   VISIT_CREATOR_LINK: {};
   /** The user right clicks a single image result, most likely to download it. **/
   RIGHT_CLICK_IMAGE: {};
@@ -186,21 +194,46 @@ async function sendEvent<T extends AnalyticsEventTypes>(
   payload?: AnalyticsEvents[T]
 ): Promise<any> {
   /**
-   * Actually send the event here, with some additional values:
-   *
-   * - Session ID
-   * - Current path
+   * The default values to send with every analytics event.
    */
-  // Default options are marked with *
-  navigator.sendBeacon;
-  const response = await fetch(`/api`, {
-    method: "POST", // *GET, POST, PUT, DELETE, etc.
+  const defaults = {
+    timestamp: Date.now(),
+    path: window.location.pathname,
+    // The dimensions of the user's device
+    width: window.innerWidth,
+    height: window.innerHeight,
+    referrer: document.referrer || null,
+  };
+  /**
+   * Actually send the event here, with some additional values.
+   * I would have prefered to use `navigator.sendBeacon` here
+   * but it is blocked by most ad blockers.
+   *
+   * We might not want to call our api route "analytics" to
+   * further avoid blocking by adblockers, but I'll use that
+   * name for clarity here.
+   */
+  const response = await fetch(`/analytics`, {
+    method: "POST",
     headers: { "Content-Type": "application/json" },
-    body: JSON.stringify(payload),
+    body: JSON.stringify({ ...defaults, ...payload }),
   });
-  return response.json(); // parses JSON response into native JavaScript objects
 }
 ```
+
+### Implementation Plan
+
+This needs more elaboration, as it's a very rough outline. For example, specific events will require more detailed planning.
+
+- [ ] Create "Event Tracking" GitHub milestone in the frontend repository
+- [ ] Create a GitHub issue for each of the following todos
+- [ ] Create an issue for the creation of the 'sendEvent' function and for the api route to post analytics events to
+- [ ] Create a single issue for each event in the sample implementation
+
+## Concerns / Pitfalls
+
+- Related to content safety, this is an area where we are surfacing user information in the form of text, or even malicious URLs, to Openverse developers. These searches could include explicit or sensitive material. We're also potentially surfacing explicit image results to anyone working with this data.
+- We're building something here that is commonly drop-in when 3rd party tools are used. Is this a poor use of time?
 
 ### Prior Art
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -63,7 +63,7 @@ Unless all conditions are met the `sendEvent` function will simply no-op.
 
 A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
 
-This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the soon to be created `/privacy` page.
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
 
 ### Sample implementaton and event types
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -9,7 +9,7 @@ A proposal to emit frontend events based on user interactions and page visits.
 
 ## Rationale
 
-Openverse developers have very little insight into the behaviors and interests of our users. We do not have any means of observing if they actually use the features we build in the intended manner, or at all.
+Openverse contributors have very little insight into the behaviors and interests of our users. We do not have any means of observing if they actually use the features we build in the intended manner, or at all.
 
 There are multiple ways to address this, like conducting user surveys or test sessions. One other way we would like to gather this information is through analytics, specifically logging events performed by users. I'll share a full list of the events later in this proposal, but here are some examples:
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -255,6 +255,8 @@ type AnalyticsEvents = {
   TOGGLE_FILTERS: {
     /** The media type being searched */
     mediaType: MediaType;
+    /** The state of the filter sidebar after the user interaction. */
+    state: "open" | "closed";
   };
 };
 type AnalyticsEventTypes = keyof AnalyticsEvents;
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -435,13 +435,12 @@ async function sendEvent<T extends AnalyticsEventTypes>(
 
 ### Implementation Plan
 
-This needs more elaboration, as it's a very rough outline. For example, specific events will require more detailed planning.
-
 - [ ] Create "Event Tracking" GitHub milestone in the frontend repository
-- [ ] Design: design a privacy and analytics opt-out checkbox on the Privacy page
-- [ ] Create an issue for the `ANALYTICS_ENABLED` feature flag
-- [ ] Create an issue for the creation of the 'sendEvent' function and for the api route to post analytics events to
-- [ ] Create a single issue for each event in the sample implementation. Each issue will need its specific implementation (when the event is called, what user DOM interaction triggers it, etc.) approved by frontend developers before it is implemented.
+- [ ] Design: design a privacy and analytics opt-out checkbox on the Privacy page. Also a pop-up banner to be shown to users letting them know Openverse uses analytics (can use the existing audio notice for this)
+- [ ] Create the `ANALYTICS_ENABLED` feature flag which defaults to `off`/`false`
+- [ ] Create an issue for the creation of the `sendEvent` function and the api route to post analytics events to. The api route will post events to an `ANALYTICS_ENDPOINT` environment variable. If that variable is unset and/or the `ANALYTICS_ENABLED` flag is off, the analytics events will not send.
+- [ ] Create each event in the sample implementation. Each event will need a detailed description of the payload and trigger (the actual user interaction that fires off the event). Events will need to account for accessibility concerns. Will keyboard users' actions trigger the events the same way as mouse users?
+- [ ] Future: Create a PR to turn `ANALYTICS_ENABLED` to `on`/`true`, set the `ANALYTICS_ENDPOINT` environment variable, and do any munging of data necessary for the decided on analytics service.
 
 ## Concerns / Pitfalls
 
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -76,7 +76,7 @@ Below I've created a TypeScript type for all the proposed anlytics events, and a
 // src/utils/events.ts
 
 // This type is only for illustrative purposes and would be imported from elsewhere.
-type MediaType = "image" | "audio" | "all";
+type SearchType = "image" | "audio" | "all";
 
 /**
  * A list of all events allowed by our analytics server. For each event
@@ -94,7 +94,7 @@ type AnalyticsEvents = {
    */
   SUBMIT_SEARCH: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
   };
@@ -127,7 +127,7 @@ type AnalyticsEvents = {
     /** The value of the filter  */
     value: string | boolean;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
   };
@@ -140,9 +140,9 @@ type AnalyticsEvents = {
    */
   CHANGE_CONTENT_TYPE: {
     /** The previously-set media type */
-    previous: MediaType;
+    previous: SearchType;
     /** The new media type */
-    next: MediaType;
+    next: SearchType;
     /** The name of the Vue component used to switch content types. */
     component: string;
   };
@@ -160,7 +160,7 @@ type AnalyticsEvents = {
    */
   REACH_RESULT_END: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
     /** The current page of results the user is on. */
@@ -177,7 +177,7 @@ type AnalyticsEvents = {
    */
   INSUFFICIENT_RESULTS: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
     /** The number of results returned */
@@ -194,7 +194,7 @@ type AnalyticsEvents = {
    */
   LOAD_MORE_RESULTS: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
     /** The current page of results the user is on. */
@@ -209,7 +209,7 @@ type AnalyticsEvents = {
    */
   VIEW_EXTERNAL_SOURCES: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
     /** The current page of results the user is on. */
@@ -227,7 +227,7 @@ type AnalyticsEvents = {
     /** The full URL of the source */
     url: string;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The search term */
     query: string;
   };
@@ -268,7 +268,7 @@ type AnalyticsEvents = {
     /** If the result is a related result, provide the ID of the 'original' result */
     related: string | null;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The slug (not the prettified name) of the provider */
     provider: string;
     /** The search term */
@@ -285,7 +285,7 @@ type AnalyticsEvents = {
     /** The slug (not the prettified name) of the provider */
     provider: string;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
   };
   /**
    * Description: The user clicks one of the buttons to copy the media attribution
@@ -299,7 +299,7 @@ type AnalyticsEvents = {
     /** The format of the copied attribution */
     format: "plain" | "rich" | "html";
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
   };
   /**
    * Description: The user reports a piece of media through our form
@@ -315,7 +315,7 @@ type AnalyticsEvents = {
     /** The slug (not the prettified name) of the provider */
     provider: string;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
   };
   /** Description: The user plays, pauses, or seeks an audio track.
    * @todo: This potentially requires throttling.
@@ -364,7 +364,7 @@ type AnalyticsEvents = {
     /** The unique ID of the media */
     id: string;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
   };
   /**
    * Description: The user uses the 'back to search' link on a single result
@@ -375,7 +375,7 @@ type AnalyticsEvents = {
     /** The unique ID of the media */
     id: string;
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
   };
   /**
    * Description: The visibility of the filter sidebar on desktop is toggled
@@ -384,7 +384,7 @@ type AnalyticsEvents = {
    */
   TOGGLE_FILTERS: {
     /** The media type being searched */
-    mediaType: MediaType;
+    mediaType: SearchType;
     /** The state of the filter sidebar after the user interaction. */
     state: "open" | "closed";
   };
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -265,6 +265,8 @@ type AnalyticsEvents = {
    *   - Are there popular searches that do not result in result selection?
    */
   SELECT_SEARCH_RESULT: {
+    /** The unique ID of the media */
+    id: string;
     /** If the result is a related result, provide the ID of the 'original' result */
     related: string | null;
     /** The media type being searched */
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -4,7 +4,7 @@ A proposal to emit frontend events based on user interactions and page visits.
 
 ## Reviewers
 
-- [ ] @krysal - For API experience
+- [x] @krysal - For API experience
 - [x] @dhruvkb - For frontend and DX experience
 
 ## Rationale
--- a/rfcs/20221006-frontend_event_tracking.md
+++ b/rfcs/20221006-frontend_event_tracking.md
@@ -5,7 +5,7 @@ A proposal to emit frontend events based on user interactions and page visits.
 ## Reviewers
 
 - [ ] @krysal - For API experience
-- [ ] @dhruvkb - For frontend and DX experience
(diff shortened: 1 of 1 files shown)


rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,294 @@
...
+- [ ] Design: design a privacy and analytics opt-out checkbox on the Privacy page
+- [ ] Create an issue for the `ANALYTICS_ENABLED` feature flag
+- [ ] Create an issue for the creation of the 'sendEvent' function and for the api route to post analytics events to
+- [ ] Create a single issue for each event in the sample implementation. Each issue will need its specific implementation (when the event is called, what user DOM interaction triggers it, etc.) approved by frontend developers before it is implemented.
+
+## Concerns / Pitfalls
+
+- Related to content safety, this is an area where we are surfacing user information in the form of text, or even malicious URLs, to Openverse developers. These searches could include explicit or sensitive material. We're also potentially surfacing explicit image results to anyone working with this data.
+- I think it's great to build this decoupled from a specific back-end, but if we never actually implement a backend this would largely be a wasted effort.
+- We're building something here that is commonly drop-in when 3rd party tools are used. At least for page views; custom events still require custom work. Is this a poor use of time?
```
sarayourfriend, line 286: I think my main concern with this approach is that we would not be able to easily track a series of events for a particular user session, unless Plausible or the other backend somehow does this for us. Though if we're proxying the events through the Nuxt server, then cookie based approaches for that on the analytics tool end of things probably will not work.

This would be useful for example if we wanted to analyse what behaviours people follow when they fall off the funnel and measure A -> B -> C type funnels. For example, under the current proposal, would we be able to know if someone started from the homepage, entered a search, refined the search, and then landed on a particular result, then clicked to visit the upstream provider? To do that we'd need to be able to tie events to each other through some kind of signifier. The signifier would need to be "sticky" so that if someone opened a new tab on the search results page (opening a single result in the new tab) then we'd still be able to tell if that search session led to a click through to the provider or attribution copy event, for example (and crucially distinguish it from someone landing directly on a single result page from an unrelated link).

My understanding is that when you use 3rd party scripts directly, these sorts of things are automatically tracked together for you.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Prior Art
```
sarayourfriend, line 288: Did you take a look at GoatCounter, by any chance? It has a Django middleware plugin that makes it pretty easy to get analytics from our API usage as well, not just from the frontend: https://www.goatcounter.com/

Implementing it would be pretty trivial as well (in my experience) and it supports data attribute based event tracking. It allows you to easily configure click events by assigning attributes to the elements that might be clicked on. It also supports an event function API similar to the one you described that we could use and you can send events directly to the API endpoint too so we could proxy if it we do indeed want to do that.

I've used it for a personal project and the dashboard is also quite nice to use.

One question I have for any of these providers though is whether we're able to easily get the data out so that we can do analysis on it ourselves in the future if we outgrow the capabilities of the tool itself. In particular if we started deriving scores for results based on how likely they are to be clicked on when they appear in a search, we would need to be able to ETL the data out of the analytics provider and into our own data warehouse. If the tool does _not_ support that then IMO it is a non-starter.

sarayourfriend (commented):
One infrastructure question I have: do you have any inclination over whether we would self-host any of the providers you mentioned? Do you have opinions about whether that is worth doing? Does it make extracting analytics data easier if we did (or potentially possible if there's a solution we really like but doesn't support continuous data export)?

rfcs/20221006-frontend_event_tracking.md
zackkrida: Oh, that's actually an accidental omission on my part and i will amend the RFC. All of the drop-in sessions do a bit of work to generate a session id, typically cookie-based. We could do that as well and pass in the session id as a default with every event.

zackkrida: Oh and in the case of Plausible, they actually take the user's IP and UA string, hash it with a rotating salt, to detect users and sessions. Details here:

https://plausible.io/data-policy#how-we-count-unique-users-without-cookies

So in that case we would just send requests to the self-hosted version of Plausible the correct headers and it should work automatically. We'd post our events to this endpoint: https://plausible.io/docs/events-api

zackkrida: GoatCounter's design doc for session ids is really great too: 
https://github.com/arp242/goatcounter/blob/master/docs/sessions.markdown

rfcs/20221006-frontend_event_tracking.md
sarayourfriend: I may be misunderstanding what you mean by custom events, but I believe it does support them, at least in my understanding of them, I use them in my personal project: https://www.goatcounter.com/help/events#sending-events-from-javascript-44

It handles page views automatically but click events can be done either via JS (linked directly above) or via the data attributes I mentioned.

sarayourfriend: Self hosting GoatCounter is pretty easy too, if we want to go that route. If it's something we'd want to do, it'd be nice to evaluate the different providers to see what the effort should be. I'd be surprised if any are much more than a single docker container rigged up to a database though.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,294 @@
...
+Events will _only_ be collected and sent under the following conditions:
+
+- An `ANALYTICS_ENABLED` feature flag is set to `true`.
+- The user has not opted out of analytics by setting the `allowAnalytics` value in our UI cookie store to `false`.
+
+Unless all conditions are met the `sendEvent` function will simply no-op.
+
+A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
+
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the soon to be created `/privacy` page.
```
obulat, line 66: The page is already merged :)

rfcs/20221006-frontend_event_tracking.md
zackkrida: I can clarify what I mean about the events. My language was definitely imprecise so apologies.

As fas as I could tell GoatCounter's "javascript events" are still _meant to_ be page view events. Based on the docs they have a strict payload signature that only allows for an ID, title, and referrer for each event. I think we would benefit from arbitrary payloads (enforced via types for each kind of event) to allow us to ask some pretty specific questions of our data. GoatCounter's events can also only be triggered by click, which while is a good way of detecting user interactions is also a bit too limiting, IMO. For example, say we wanted to keep track of how long users spend on the page, we'd want our event to be triggered by the `beforeunload` event, or if we wanted to trigger an event based on right clicks (as one event in my proposal does), we wouldn't be able to do it with GoatCounter.

rfcs/20221006-frontend_event_tracking.md
zackkrida, line 66:
```suggestion
This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
```

zackkrida: Revised, thanks!

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,294 @@
...
+  };
+  /**  The user opens the menu which lists pages. */
+  OPEN_PAGES_MENU: {};
+  /** Whenever the user selects a result from the search results page. */
+  SELECT_SEARCH_RESULT: {
+    /** If the result is a related result, provide the ID of the 'original' result */
+    related: string | null;
+    /** The media type being searched */
+    mediaType: MediaType;
+    /** The search term **/
```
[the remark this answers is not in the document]
zackkrida: That's a great idea; I'll add it now. OF course, that would be a fantastic way to see which provider's results are most popular!

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,294 @@
...
+    query: string;
+  };
+  /**  The user opens the menu which lists pages. */
+  OPEN_PAGES_MENU: {};
+  /** Whenever the user selects a result from the search results page. */
+  SELECT_SEARCH_RESULT: {
+    /** If the result is a related result, provide the ID of the 'original' result */
+    related: string | null;
+    /** The media type being searched */
+    mediaType: MediaType;
```
zackkrida, line 161:
```suggestion
    mediaType: MediaType;
    /** The slug (not the prettified name) of the provider */
    provider: string;
```

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+Events will _only_ be collected and sent under the following conditions:
+
+- An `ANALYTICS_ENABLED` feature flag is set to `true`.
+- The user has not opted out of analytics by setting the `allowAnalytics` value in our UI cookie store to `false`.
+
+Unless all conditions are met the `sendEvent` function will simply no-op.
+
+A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
+
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
```
dhruvkb, line 66: I couldn't find it but I was certain we had a discussion regarding the design of a public preferences page where users could toggle flags for beta features and enable/disable functionality like tracking and mature content filters.

The current `/preferences` page is not designed for general usage.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+   * Whenever the user scrolls to the end of the results page.
+   * Useful to evaluate how often users load more results or click
+   * on the external sources dropdown.
+   * */
+  REACH_RESULT_END: {
+    /** The media type being searched */
+    mediaType: MediaType;
+    /** The search term **/
+    query: string;
+  };
```
dhruvkb, line 128: Similar to `REACH_RESULT_END`, we should record search queries that generate a very small number (`INSUFFICIENT_RESULTS`) or no results at all (`ZERO_RESULTS`).

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   * The default values to send with every analytics event.
+   */
+  const defaults = {
+    timestamp: Date.now(),
+    path: window.location.pathname,
+    // The dimensions of the user's device
+    width: window.innerWidth,
+    height: window.innerHeight,
+    referrer: document.referrer || null,
```
dhruvkb, line 255: I am not sure if collecting the following information is invasive but it would be beneficial to have information about the browser (name and version), the operating system (name and version) and the device. We typically ask for this information in our bug reports.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+  };
+  /**
+   * Actually send the event here, with some additional values.
+   * I would have prefered to use `navigator.sendBeacon` here
```
dhruvkb, line 259: I couldn't find any documentation about ad-blockers blocking the `navigator.sendBeacon` API. Rather what ad-blockers seem to be doing is blocking requests to analytics and tracking URLs passed to `sendBeacon`.

Here is empirical evidence with sample size 1.
Screenshot 2022-12-03 at 3 25 26 PM

Screenshot 2022-12-03 at 3 25 38 PM

Most ad-blockers will use [EasyList](https://easylist.to/easylist/easylist.txt) so if we avoid endpoints that match the list, we should be able to use `sendBeacon`.

dhruvkb (approved):
From a preliminary read, I like the RFC. The idea for avoiding vendor lock-in is nice. The collection of events are very comprehensive. The TypeScript types will be very helpful during implementation.

I will read it a few more times but this looks good to me!

rfcs/20221006-frontend_event_tracking.md
sarayourfriend: > GoatCounter's events can also only be triggered by click

This isn't true, you can send events from JavaScript on arbitrary actions (re-linking the doc I shared above): https://www.goatcounter.com/help/events#sending-events-from-javascript-44

> Based on the docs they have a strict payload signature that only allows for an ID, title, and referrer for each event

Gotcha! That does seem to be a disqualifying limitation, good call!

rfcs/20221006-frontend_event_tracking.md
sarayourfriend: I'm not even sure that specific width and height are necessary for the questions we currently have and would like us to consider collecting _as little_ information as possible. The information that might be useful for us is the same kind of stuff in the UI state cookie, general breakpoint (maybe?, I'm sceptical...) and whether it's a mobile UA (definitely useful considering the significant design differences). For bugs, we'd ideally want to rely on Sentry for that I would think?

dhruvkb: I didn't mean to suggest using analytics as bug reports, the reference to the issue template was just to say that there is precedent for asking for this info. While on one hand, I can get behind collecting as coarse information as possible, but on the other, it might help to how certain interactions are affected by the devices in use.

sarayourfriend: That's true, device capability (low-spec devices for example) _could_ be useful, I didn't consider that. Some of this information will not be easily derived from flattened UA string we have now though and some (like RAM and CPU capacity) will require permission prompts when UA Client Hints are rolled out: https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API

We can reliably get browser and device type (I think) without permission prompts.

rfcs/20221006-frontend_event_tracking.md
zackkrida: This is a good idea. I wonder if it should be the same event `INSUFFICIENT_RESULTS` but include the result count in the payload. This could be really helpful for making determinations about what types of content users _want_ that we don't currently offer.

I think loading results _less than_ one full page would be a good trigger to the event.

rfcs/20221006-frontend_event_tracking.md
zackkrida: I think initially we could just add a checkbox to the `/privacy` page, as suggested here, but I would definitely like to expand to a preferences page in the future, where we could include this setting as well.

rfcs/20221006-frontend_event_tracking.md
zackkrida: uBlock origin blocks it by default, under the default enabled setting "Disable hyperlink auditing":
 
CleanShot 2022-12-06 at 13 21 42

Details here:

https://github.com/gorhill/uBlock/wiki/Dashboard:-Settings#disable-hyperlink-auditing

rfcs/20221006-frontend_event_tracking.md
dhruvkb: Going into bikesheddding territory here, but the privacy page, to me at least, is somewhere I would go if I wanted to read a bunch of legalese about how my data is collected and handled and not where I'd expect to find options to opt-out of data collection. 

But in the short-term, and in the absence of other preference toggles, a page for a single checkbox would be weird.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+
+## The technical details
+
+We will create a system for collecting events on the frontend. Each event will have a unique name, collect some default data about the enviroment (timestamp, referer, and so on) and can have an optional data payload containing string and numerical values. I'd like to limit the payload shape to strings and numbers to keep it JSON serializable and simple. Allowing for complex nested objects might prove too complex to query in the future. I think this is a reasonable constraint but am open to alternatives.
+
+### Sending events
+
+These events will be sent as `POST` requests to a server route on the frontend, which will then route them to our desired analytics service. In the future, it may make sense to move this server route to a standalone microservice, but for now a [Nuxt server route](https://v3.nuxtjs.org/guide/directory-structure/server/#server-routes) should be sufficient to handle the analytics events. This endpoint is essentially a proxy between the frontend and our 'true' analytics service that makes sure we're always sending events server-side.
```
obulat, line 53: Since we are still on Nuxt 2 😢, we should probably also include [Nuxt serverMiddleware](https://nuxtjs.org/docs/configuration-glossary/configuration-servermiddleware/) as a Nuxt 2 alternative to Nuxt server routes.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Privacy
+
+Events will _only_ be collected and sent under the following conditions:
+
+- An `ANALYTICS_ENABLED` feature flag is set to `true`.
+- The user has not opted out of analytics by setting the `allowAnalytics` value in our UI cookie store to `false`.
```
obulat, line 60: How do we plan to allow users to opt out of analytics? Do we need design for a banner/page/something else?

rfcs/20221006-frontend_event_tracking.md
obulat: Agree with the bikeshedding comment, my version of this sentence would be
"...but the privacy page, to me at least, is somewhere I would **_NOT_** go ~if I wanted to~ so **I don't have to** read  legalese"

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+```ts
+// src/utils/events.ts
+
+// This type is only for illustrative purposes and would be imported from elsewhere.
+type MediaType = "image" | "audio" | "all";
+
+/**
+ * A list of all events allowed by our analytics server.
+ */
+type AnalyticsEvents = {
```
obulat, line 81: Can we also add the user language to the events?

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+  BACK_TO_SEARCH: {
+    /** The unique ID of the media */
+    id: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
+  /** The visibility of the filter sidebar on desktop is toggled **/
+  TOGGLE_FILTERS: {
+    /** The media type being searched */
+    mediaType: MediaType;
```
obulat, line 233: It might be useful to also add the state of the filter after toggling.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+    name: string;
+    /** The full URL of the source */
+    link: string;
+    /** The media type being searched */
+    mediaType: MediaType;
+    /** The search term **/
+    query: string;
+  };
+  /**  The user opens the menu which lists pages. */
+  OPEN_PAGES_MENU: {};
```
obulat, line 155: How about also adding events for clicking on external links in the navigation or content pages? Get involved, API, WordPress.org links and the links to the providers on the About page.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+ * user has opted-out of analytics.
+ */
+async function sendEvent<T extends AnalyticsEventTypes>(
+  key: T,
+  payload?: AnalyticsEvents[T]
+): Promise<any> {
+  /**
+   * The default values to send with every analytics event.
+   */
+  const defaults = {
```
dhruvkb, line 249: I just realized, we should also be recording the language/locale in use. Also, if we retrieve some broad information about the IP such as country, but not actually store the IP, would it be considered invasive?

For reference, GoatCounter records both country and language.

rfcs/20221006-frontend_event_tracking.md
dhruvkb: @zackkrida would recording the user's IP be considered invasive, even if we  rotate it daily?

sarayourfriend: > We could also look at a custom event for a given user and all previous VIEW_PAGE events, something like:

What does this mean? In case you meant that we would change the event name for each user (like by prefixing some user hash to the event name or something along those lines), I don't think that's possible or reliable, at least. From what I understand a lot of these tools are not optimised for the event name to vary that often (per user) and they should be stable. Sometimes you even have to configure the event and its payload specifically in the tool first before they even get ingested (GTM is like this, no experience with Plausible to say either way).

rfcs/20221006-frontend_event_tracking.md
sarayourfriend: That sounds good. The derived data will probably be a _lot_ easier (and faster) to query as well.

rfcs/20221006-frontend_event_tracking.md
panchovm: I am thinking of two complementary solutions.

For the short term, showing a banner similar to most cookie consent messages aligned at the bottom. And for the long term, a toggle element in a profile settings menu to enable/disable; in line with the `Show mature content` toggle we talked about in the prioritization meeting.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+### Implementation Plan
+
+This needs more elaboration, as it's a very rough outline. For example, specific events will require more detailed planning.
+
+- [ ] Create "Event Tracking" GitHub milestone in the frontend repository
+- [ ] Design: design a privacy and analytics opt-out checkbox on the Privacy page
+- [ ] Create an issue for the `ANALYTICS_ENABLED` feature flag
+- [ ] Create an issue for the creation of the 'sendEvent' function and for the api route to post analytics events to
```
krysal, line 281: Perhaps will be better to rephrase these in terms of development rather than project management?
> - [ ] Create the `ANALYTICS_ENABLED` feature flag
> - [ ] Create the 'sendEvent' function and the api route to post analytics events to
...

krysal (commented):
This a great document, it's exciting to start work that enables knowing more about the Openverse users and how we could improve the experience. The rationale to focus on the frontend side and avoid vendor lock-in sounds good to me. From your explanation, it seems certain that we will be able to achieve this regardless of the service we choose to store the statistics.

I'd just like to see a list of questions about what kind of behaviors we want to learn more about, and how to measure it. That will guide the type of events we need to create and have some discussions of each one earlier. We don't need a ton of events to begin with, just a handful of well-defined events and stats that will guide future decisions.

Some questions I had in mind are:
- What are the most/less used filters?
- How often is the attribution of the media used (copied)?
- What are the not yet supported sources more looked for? (the links we suggest when there are no results for a search term)
- etc

It can be explained the other way around, what will event X be used for? This would be really helpful, and I believe we can all contribute to this area as well, @WordPress/openverse-frontend.

rfcs/20221006-frontend_event_tracking.md
dhruvkb: Just saw @obulat's comment https://github.com/WordPress/openverse/pull/312#discussion_r1041931113 above in a re-read.

sarayourfriend:
Thanks for confirming.

rfcs/20221006-frontend_event_tracking.md
zackkrida: I've added the fields necessary for session and funnel tracking, along with details about security, in 34fea2a57093c38c25ae218a8fc4ac197d3bea6d. As-written we're collecting enough information to be compatible with many 3rd party services. Following the scope of this PR I'll leave the specific implementation of session id generation and the storing of event payloads to the backend service.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Fields and documentation added in 34fea2a57093c38c25ae218a8fc4ac197d3bea6d.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Added in 34fea2a57093c38c25ae218a8fc4ac197d3bea6d. Thanks!

rfcs/20221006-frontend_event_tracking.md
zackkrida: Great idea, added in 853028f.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Added language tracking in 34fea2a57093c38c25ae218a8fc4ac197d3bea6d.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Maybe, I do think the issue creation is very important as each event needs to be able to be implemented by someone reading the issue exclusively, who may not have read the RFC. The importance of the issues themselves makes me want to keep track of their creation.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Mentioned in 7265661.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Added in 38d1916.

zackkrida:
@krysal I've added questions each event aims to answer in the comments of each event. I've also made many updates based on user feedback; please see the commit history for detail on each of these additions. This is ready for a final re-review and anyone can provide additional feedback. Thank you!

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,296 @@
...
+    /** The unique ID of the media */
+    id: string;
+    /** The format of the copied attribution */
+    format: "plain" | "rich" | "html";
+    /** The media type being searched */
+    mediaType: MediaType;
+  };
+  /** The user reports a piece of media through our form */
+  REPORT_MEDIA: {
+    /** The unique ID of the media */
```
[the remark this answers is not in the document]
sarayourfriend: > This actually gives me an idea, we may want to add the version number of Openverse to the events as another default field.

Doesn't sound like a bad idea generally.

> Is this event necessary since we are already receiving the report?

@krysal is on to something here that we should not quickly dismissed. I hadn't considered it earlier, but we should consider that long-term analysis of our usage would probably logically include our own data in addition to the events themselves.

However, the reports don't record who did it or if they're done in the same session, so this kind of thing might be useful for identifying if there are specific "power users" or "power sessions" where people go and report lots of things all at once. Not sure what we would do with that information though :slightly_smiling_face: 

If the question is "how many do we receive on average" etc, then we can set up a Grafana dashboard eventually to pull that from the database itself, rather than denormalising the data into an event stream separate from the actual data itself.

This is more of a general principle/consideration though. I don't know how/if it applies specifically to the events here. Something to consider for these and future events though. We want to have a wholistic approach to our analytics that incorporates all the data streams we have, not just user event data.

sarayourfriend:
@zackkrida Something that just occurred to me, do you think we could implement RUM performance data into the events here or should we use a separate tool for it? RUM performance data can be super useful for identify performance issues that only affect a small subsection of clients. It could also supplement things like Google Lighthouse scores and trying to debug issues with specific parts of the lighthouse score.

zackkrida:
@sarayourfriend I think that's a great idea. It honestly didn't occur to me either. We could definitely do this by creating _new_ events which track [Core Web Vitals](https://web.dev/vitals/)-esque events, and maybe by adding performance timing information to appropriate events already defined here. 

My only initial concern is that those events might add considerable time and testing to implement, and I am undecided if they should be in the first 'mvp' of this event tracking or if they would be better-suited to a standalone project. One consideration pushing me towards waiting is that Cloudflare already has some tracking of RUM metrics. They provide the 50th, 75th, 90th, 99th percentile performance for a few metrics:

- Cumulative Layout Shift (CLS)
- First Input Delay (FID)
- Largest Contentful Paint (LCP)

[Details here](https://developers.cloudflare.com/analytics/web-analytics/understanding-web-analytics/core-web-vitals). We can evaluate that for Openverse and see if it's sufficient prior to doing our own implementation. A good test would actually be to use that CloudFlare information to see if @obulat's cookie PR shows a reduction in layout shift!

Do you think that's reasonable? Is it preferable to prioritize RUM in this RFC?

sarayourfriend:
> One consideration pushing me towards waiting is that Cloudflare already has some tracking of RUM metrics.

> Do you think that's reasonable? Is it preferable to prioritize RUM in this RFC?

I think it is reasonable to intentionally exclude it, but that we should prioritise writing a follow up RFC that includes a wholistic approach to analysis that would include data aside from multiple sources (database, Cloudflare, custom events). It seems to me this would fit quite naturally in the scope of the RFC you mentioned @dhruvkb is interested in writing regarding the usage of the data (including the backend provider for event ingestion).

panchovm:
I read the questions we would like to get answers to, and I think we are covering most use cases.

@krysal pointed out the amount of copying attribution information, and I would link it with clicking on `Use this [content type]` button to see the relationship. This flow is crucial to see whether using a piece of content is concatenated to copying the attribution information.

Here are other questions I am curious to know per feature/section:
1. Use of filters: How often is the filter opened/closed during a search? There is a question about visibility preference, but I am curious if users shift between showing and hiding the sidebar during one search.
2. Browsing on All content results: When opening content details from All content results, do users keep browsing on the same content type or go back to All content results? The point behind this is understanding whether All content results works as showcasing content type and Openverse potential, or as a result page to keep browsing.
3. Using meta search: After how many loading results do users use the meta search feature? To see users' tolerance for getting no "relevant results", or results not matching what they expect to see.

Great document. It describes clearly what we aim for.

krysal (commented):
Thanks for adding the questions. I have yet to review each one of the **21 types of events** we aim to collect but I have a question that goes in line with what Sara asked. You said that after the approval of this RFC, "we implement this with a no-op backend", does this means we will build the functions to collect the data but yet not store it anywhere?

To me it seems this requires a log of the events tracked _as a minimun_, that could go in the same instance(s) of Nuxt, it doesn't even need to reach the API. Although the log file would surely need rotational maintenance due to all the per-session events that would be handled, it would grow incredibly fast. This is the most agnostic storage backend I can think of. Otherwise, if we don't save this anywhere _this would largely be a wasted effort_, as you mentioned in the second bullet point of Concerns / Pitfalls.

A more advanced solution to analyze the data would be great but I can see that as part of another RFC for sure.

zackkrida:
@krysal I'll address your other comments too, but regarding the implementation of this and the concern that "this would largely be a wasted effort":

Even if we knew what backend we were using _today_, we still have to implement the frontend event tracking. Which will take time; likely weeks. I am not saying we can't or shouldn't work on the backend, only that this frontend work shouldn't be _coupled to_ that backend work. I don't think having all the event tracking code written, but disabled behind a flag to not actually post anywhere is a waste of time. 

Also, I am recommending here that each event be written as a separate small issue. To me, it makes sense to write and add these events individually while the feature is disabled. Instead of enabling events one by one into a live analytics system. This way we can turn on analytics tracking at one time and all of the initial events outlined here will have the same start date.

I _would_ definitely recommend we come up with a backend as soon as we reasonably can, something like you suggested, or a single postgres table somewhere. Anywhere we can store the actual events and query them later, import them into a UI, etc. The only point I'm suggesting is that the frontend work, which needs to happen regardless, isn't blocked by the backend.


Regarding the number of events, is this section:

> each one of the **21 types of events**

meant to suggest you think there are too many events here for the initial implementation? If so, that sounds like important information to share. I would definitely be open to narrowing the list to a smaller subset of events for the initial implementation if anyone felt that was a better path forward and explained why 😄.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,458 @@
...
+- An `ANALYTICS_ENABLED` feature flag is set to `true`.
+- The user has not opted out of analytics by setting the `allowAnalytics` value in our UI cookie store to `false`.
+
+Unless all conditions are met the `sendEvent` function will simply no-op.
+
+A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
+
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
+
+### Sample implementaton and event types
```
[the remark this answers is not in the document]
zackkrida: I could change to "Events and an example implementation" if you like. The events are definitely meant to be the final list but the `sendEvent` function is probably 80-90% of the actual code, with some subtleties not addressed here (like reading some default values from Nuxt context, for example).

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,458 @@
...
+   *   - How many searches do we serve per day/month/year?
+   *   - What are the most popular searches in Openverse?
+   *   - Which media types are the most popular?
+   *   - Do most users search from the homepage, or internal searchbar?
+   */
+  SUBMIT_SEARCH: {
+    /** The media type being searched */
+    mediaType: MediaType;
+    /** The search term */
+    query: string;
```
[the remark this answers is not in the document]
zackkrida: Definitely. The current page (`{path: window.location.pathname}`) is one of the default values collected with every event.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,458 @@
...
+Events will _only_ be collected and sent under the following conditions:
+
+- An `ANALYTICS_ENABLED` feature flag is set to `true`.
+- The user has not opted out of analytics by setting the `allowAnalytics` value in our UI cookie store to `false`.
+
+Unless all conditions are met the `sendEvent` function will simply no-op.
+
+A previous version of this proposal suggested honoring the `navigator.doNotTrack` browser method, but upon inspection this was deprecated in 2018 and is no longer advised.
+
+This proposal will require the creation of a user settings page _or_ an analytics opt-out checkbox on the `/privacy` page.
```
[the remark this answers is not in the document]
zackkrida: We do not need to show a cookie banner. We want to give the user the ability to opt-out of analytics, so a place with a checkbox to disable analytics is the only requirement.

@panchovm has suggested a banner as one way of showing that, which makes sense. How will users know they _can_ disable analytics without us telling them we are using analytics?

In that case, we would probably also want an event so we can track how many users disable analytics 🤔 Not sure if we can even do that though, it's an odd thing. We want to capture that they are expressing their desire to not have their behavior captured ♾️

rfcs/20221006-frontend_event_tracking.md
zackkrida: I think this is where we need @panchovm's suggestion for a banner announcing analytics to first-time users. Otherwise, who would _know_ to check the privacy page for analytics information? I'll resolve this and add some mention of the banner.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Refined the proposal here in 0816064.

rfcs/20221006-frontend_event_tracking.md
zackkrida: I've updated the implementation plan. As you suggested I switched to remove the 'create issue' text, I think you're correct there. More detail has been added too.

krysal:
@zackkrida I see your point on the backend planning not blocking the frontend work, and agree. Maybe I just have said that as a whole project, this RFC alone feels sort of _incomplete_, but if a subsequent one will come after this, then it makes sense to start this implementation in parallel to the planning of the backend 💯

> Regarding the number of events, is this section:
>  >  each one of the 21 types of events
>
> meant to suggest you think there are too many events here for the initial implementation?

No, I think is a good number! 😄 I just would need to give it a look at another time, so I'll come back to this next week. Thanks for addressing some of the concerns so quickly!

krysal:
Also, if @obulat approves before I get to it, that can count as my approval well 😄 Given @obulat is one of the main contributors to the frontend her opinion is extremely relevant here.

AetherUnbound (commented):
This is excellent! I appreciate and understand the scope of this RFC and why it excludes any infrastructure effort (for now). The events seem carefully considered and would serve to answer most questions I could think to ask 🙂

sarayourfriend:
> Although the log file would surely need rotational maintenance due to all the per-session events that would be handled, it would grow incredibly fast

We could use logs as a temporary backend but:

1. It is relatively (though not prohibitively) expensive to store, as Krystle mentioned
2. It is expensive to query, both monetarily and in time
3. It offers us relatively zero easily-usable immediate feedback for any of this data and would require additional time that I assume most dedicated analytics providers automatically do (event frequency, session browsing, etc)
4. Because it is technically unstructured we'd have to parse logs ourselves to get them into an eventual data lake for further processing (for example, document relevance derivation) or extract it from the logstash's index (probably not possible with CloudWatch and difficult otherwise)

On the other hand, it is extremely portable given we could switch our log transport to use Loki instead of CloudWatch via Grafana Cloud. Loki's approach is more appropriate for that kind of thing than CloudWatch's... but again, we lose the significant benefits that an actual analytics provider would give us.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+   *     or search pages?
+   *   - How many audio plays do we get?
+   */
+  AUDIO_INTERACTION: {
+    /** The unique ID of the media */
+    id: string;
+    event: "seek" | "play" | "pause";
+    /** The slug (not the prettified name) of the provider */
+    provider: string;
+  };
```
krysal, line 335: I think it would be valuable to include the audio component in question.

```suggestion
  /**
   * Description: The user plays, pauses, or seeks an audio track.
   * @todo: This potentially requires throttling.
   *
   * Questions:
   *   - Do users interact with media frequently?
   *   - Is it more common to playback audio on single results
   *     or search pages?
   *   - How many audio plays do we get?
   */
  AUDIO_INTERACTION: {
    /** The unique ID of the media */
    id: string;
    event: "seek" | "play" | "pause";
    /** The slug (not the prettified name) of the provider */
    provider: string;
    /** The name of the Vue component used on the interaction, e.g. the global or main player. */
     component: string;
  };
```

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+   */
+  OPEN_PAGES_MENU: {};
+  /**
+   * Description: The user clicks to a link outside of Openverse.
+   * Questions:
+   *   - What types of external content do users seek?
+   *   - Are there external resources we should make more visible?
+   *   - Is there content we might want to add to Openverse directly?
+   */
+  EXTERNAL_LINK_CLICK: {
```
krysal, line 248: Is this just for external links in the pages menu or is so general as it sounds? Because if it's for all the external links, then it will overlap with the `EXTERNAL_LINK_CLICK` and `VISIT_LICENSE_PAGE` events, for example. I personally find the specific ones more valuable. I'm not sure when this event will be triggered.

krysal (commented):
I read all the events and left some questions to clarify some points. Approving after that should be quick!

dhruvkb:
I came across some documentation about the Reporting API [[MDN]](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API) and [[Chrome]](https://developer.chrome.com/articles/reporting-api/). It's an additional piece that can be factored in as an extra data source on the frontend to capture some dev-specific metrics like crashes and deprecations.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+
+```shell
+curl -i -X POST https://plausible.io/api/event \
+  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.121 Safari/537.36 OPR/71.0.3770.284' \
+  -H 'X-Forwarded-For: 127.0.0.1' \
+  -H 'Content-Type: application/json' \
+  --data '{"name":"pageview","url":"http://dummy.site","domain":"dummy.site","screen_width":1666}'
+```
+
+There are many points in the code execution at which we could transform our custom events into ones supported by other platforms. By adopting and implementing our own custom solution, we can start collecting data as soon as possible without being blocked on time spent evaluating and implementing a full analytics system.
```
rwidom, line 45: Is there a way to surface the implicit infrastructure decisions here? For example, where will the event data be stored and what limitations might that place on future plans?

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+
+## The technical details
+
+We will create a system for collecting events on the frontend. Each event will have a unique name, collect some default data about the enviroment (timestamp, referer, and so on) and can have an optional data payload containing string and numerical values. I'd like to limit the payload shape to strings and numbers to keep it JSON serializable and simple. Allowing for complex nested objects might prove too complex to query in the future. I think this is a reasonable constraint but am open to alternatives.
+
```
rwidom, line 50: I just started scrolling through, and maybe this question is totally obvious already, but are we tracking the search terms/string that people enter anywhere? Seems like it would be important context for everything else below, right?

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+- [ ] Create each event in the sample implementation. Each event will need a detailed description of the payload and trigger (the actual user interaction that fires off the event). Events will need to account for accessibility concerns. Will keyboard users' actions trigger the events the same way as mouse users?
+- [ ] Future: Create a PR to turn `ANALYTICS_ENABLED` to `on`/`true`, set the `ANALYTICS_ENDPOINT` environment variable, and do any munging of data necessary for the decided on analytics service.
+
+## Concerns / Pitfalls
+
+- Related to content safety, this is an area where we are surfacing user information in the form of text, or even malicious URLs, to Openverse developers. These searches could include explicit or sensitive material. We're also potentially surfacing explicit image results to anyone working with this data.
+- We're building something here that is commonly drop-in when 3rd party tools are used. At least for page views; custom events still require custom work. Is this a poor use of time?
+- I think it's great to build this decoupled from a specific back-end, but if we never actually implement a backend this would largely be a wasted effort.
```
rwidom, line 461: Oh, wow. Yeah, in general, I think it's very difficult to evaluate the success of a data collection effort without looking at the data itself. Am I right in understanding that this is a data collection design that is separate from actually storing any data?

rfcs/20221006-frontend_event_tracking.md
zackkrida: Yes! The search query of the user is captured in a majority of the events outlined in the proposal. It's not captured with _every_ event as there are many to which it isn't relevant.

rfcs/20221006-frontend_event_tracking.md
zackkrida: I've commented [here](https://github.com/WordPress/openverse/pull/312#discussion_r1062735808) with some context about this question, thanks for asking!

rfcs/20221006-frontend_event_tracking.md
zackkrida: I've commented [here](https://github.com/WordPress/openverse/pull/312#discussion_r1062735808) with some context about this question, thanks for asking!

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+Openverse contributors have very little insight into the behaviors and interests of our users. We do not have any means of observing if they actually use the features we build in the intended manner, or at all.
+
+There are multiple ways to address this, like conducting user surveys or test sessions. One other way we would like to gather this information is through analytics, specifically logging events performed by users. I'll share a full list of the events later in this proposal, but here are some examples:
+
+- When a user clicks on a search result from the search results page
+- When a user copies the attribution text
+
+By collecting this data, anonymized and with user consent (through opt-out), we can make better decisions about what to build and increase the overall usefulness of Openverse.
+
+### Why is this distinct from a general-purpose, cross-repo Analytics RFC?
```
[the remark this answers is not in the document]
rwidom: Thanks. I did read that, but I guess I'm just so data-biased that it didn't really sync in. :)

sarayourfriend:
@dhruvkb Sentry already collects that data. Do you think it's necessary to duplicate it in our own event stream?

dhruvkb:
I read some post about the ReportingObserver API and it seemed like a good integration so I shared those links. I didn't know Sentry already collected that which is my bad, I didn't do my due diligence there. 

Based on a cursory look at their docs, it seems like it is an [optional integration that could be enabled](https://docs.sentry.io/platforms/javascript/configuration/integrations/plugin/#reportingobserver). Nonetheless, if that info is (or can be) collected in Sentry, that's a better, more logical place to have that info.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Events
+
+Below I've created a TypeScript type for all the proposed anlytics events, and a simple function for sending events. For each event you'll see the payload alongside the default values which will be passed to _all_ events.
+
+```ts
+// src/utils/events.ts
+
+// This type is only for illustrative purposes and would be imported from elsewhere.
+type MediaType = "image" | "audio" | "all";
```
obulat, line 79: Since it include "all, it should be `SearchType` (`MediaType` is for the actual types of media, not the aggregate "All content")

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+    query: string;
+  };
+  /**
+   * Description: Whenever a user changes the content type
+   * Questions:
+   *   - Which content types are most popular?
+   *   - Is there interest in the non-supported content types?
+   *   - Do users switch content types? Where in their journeys?
+   */
+  CHANGE_CONTENT_TYPE: {
```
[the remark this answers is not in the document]
obulat: I think `content` type is more appropriate here because `all` is also included. (`mediaType` in the frontend repository refers specifically to the kinds of media like images or audio, and `searchType` can include the combined view of `All media`)

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+  /**
+   * Whenever a search results in less than a full page
+   * of results or no results.
+   *
+   * Questions:
+   *   - How often do searches provide limited or no results?
+   *   - Do users visit external sources when results are insufficient?
+   *   - What do users do after a search has no results?
+   */
+  INSUFFICIENT_RESULTS: {
```
obulat, line 178: All results view is an edge case here: do we consider results 'insufficient' if only one of the media types has less than a full page of results? Maybe we can specifically add which results are insufficient in this case?

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+    /** The search term */
+    query: string;
+  };
+  /**
+   * Description: The user opens the menu which lists pages.
+   * Questions:
+   *   - How often is this menu used?
+   *   - Is this menu visible enough?
+   */
+  OPEN_PAGES_MENU: {};
```
obulat, line 240: Noting here that this will mostly track mobile users because on desktop the pages menu is always open.

sarayourfriend:
> Based on a cursory look at their docs, it seems like it is an [optional integration that could be enabled](https://docs.sentry.io/platforms/javascript/configuration/integrations/plugin/#reportingobserver). Nonetheless, if that info is (or can be) collected in Sentry, that's a better, more logical place to have that info.

@dhruvkb (and for anyone else unaware) It is enabled by default via the NuxtJS plugin that we use to configure Sentry: https://sentry.nuxtjs.org/sentry/options#clientintegrations

This issue in the frontend repository includes links to some issues that were surfaced from the ReportingObserverAPI in Sentry: https://github.com/WordPress/openverse-frontend/issues/1610

I only know about this because I had to triage those issues when they first appeared :sweat_smile: Unless the integration is broken we should still be getting ReportingObserverAPI issues.

rfcs/20221006-frontend_event_tracking.md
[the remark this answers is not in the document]
zackkrida: I think I prefer to keep @obulat's suggestion, as it matches the frontend code better. Does that sound okay @panchovm and @krysal?

rfcs/20221006-frontend_event_tracking.md
zackkrida: In this case, I'd only feel it necessary to report if the _aggregated set_ of results was insufficient.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+  LOAD_MORE_RESULTS: {
+    /** The media type being searched */
+    mediaType: MediaType;
+    /** The search term */
+    query: string;
+    /** The current page of results the user is on. */
+    resultPage: number;
+  };
+  /**
+   * Description: When a user opens the external sources popover.
```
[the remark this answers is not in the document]
zackkrida: I think "External Sources" is a well-defined concept in the frontend (See this page: https://production.openverse.org/about for details) but I do follow the confusion you're describing. I could use a suggestion for how to document that better.

rfcs/20221006-frontend_event_tracking.md
zackkrida: Good point! I conceived of this for the old header design.

rfcs/20221006-frontend_event_tracking.md
```diff
@@ -0,0 +1,469 @@
...
+  /**
+   * Description: Whenever the user selects a result from the search results page.
+   * Questions:
+   *   - Which results are most popular for given searches?
+   *   - How often do searches lead to clicking a result?
+   *   - Are there popular searches that do not result in result selection?
+   */
+  SELECT_SEARCH_RESULT: {
+    /** If the result is a related result, provide the ID of the 'original' result */
+    related: string | null;
```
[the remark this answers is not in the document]
zackkrida: It does! I've fixed it to include that. `related` will refer to when a piece of media is selected from a single result  view:

CleanShot 2023-01-12 at 14 55 50

rfcs/20221006-frontend_event_tracking.md
zackkrida: Fantastic! Added in 5ea8215.

zackkrida:
@krysal thank you so much! You identified some critical missing data in a few of these events. This is ready for re-review when you are able.

rfcs/20221006-frontend_event_tracking.md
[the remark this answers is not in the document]
sarayourfriend: A potentially clearer name: `relatedTo` rather than `related` which does sound to me more like it'd be whatever the related image was, not the original.

rfcs/20221006-frontend_event_tracking.md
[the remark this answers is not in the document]
krysal: If that is the convention on the FE that is perfect then! Thanks for explaining it.

rfcs/20221006-frontend_event_tracking.md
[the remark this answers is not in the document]
krysal: ^1000 times what Sara said, please.

krysal (approved):
LGTM! Thanks for going through all the comments and clarifications. This is a well-grounded start for the frontend analytics projects.

zackkrida:
@WordPress/openverse I'm not able to figure out this linting issue. The precommit and linting works locally without any changes. Any ideas what's going on here?

sarayourfriend:
@zackkrida I had the same inability to reproduce the linting error locally. It turned out it needed a rebase. I rebased and pushed the linting fix.

PR #336, opened 2022-11-24
RFC: Monorepo

### LINKED ISSUE #205: Write an RFC for monorepo support
After the discussion in #192, it seems like we're ready for an RFC for monorepo adoption. This will focus on the _specific implementation_, how to best transition, and what specific configuration and automation changes we will need to make.

dhruvkb:
## Fixes

Fixes #205 by @zackkrida 

## Description

This PR writes the monorepo RFC.

Proof of Concept can be seen at <s>https://github.com/dhruvkb/monopenverse</s> https://github.com/dhruvkb/monoverse.

## Testing Instructions

0. Read the RFC.
1. Give your Cs.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -15,19 +15,49 @@ For a comprehensive discussion about the pros, the cons and the counterpoints to
 
 This RFC summarily lists the benefits and then, with the twin assumptions of a monorepo being ultimately beneficial and the decision to migrate being finalised in the above discussion, proceeds to go into the implementation details.
 
-### Exclusive benefits of monorepo
+### Benefits of monorepo
 
-This only includes things that cannot be accomplished without the use of a monorepo.
+1. Single place for everything:
 
-1. Single place to go for issues, PRs and all activity. Currently tickets are scattered across several repos, and any tickets that could benefit more than a single layer must be opened in each of the different repos.
+   **Current criticism:** We currently have many repos, and issues and PRs spanning all of them. While this makes it easier for us as maintainers to focus our efforts, it's not easy for contributors. Let's say you were a new contributor looking for good-first Python issues. We shouldn't expect them to search in 3 repos `openverse`, `openverse-api` and `openverse-catalog`. We address this by making tools like Overvue or using search terms like this:
 
-1. Singular copy (different from synced independent copies) of scaffolding code such as Git hooks, lint rules and common workflows. This is distinctly better than elaborate sync workflows.
+   ```
+   is:open is:issue repo:WordPress/openverse-catalog repo:WordPress/openverse repo:WordPress/openverse-api repo:WordPress/openverse-frontend repo:WordPress/openverse-infrastructure sort:updated-desc
+   ```
 
-1. Central place for all technical documentation, enabling documentation for different parts of the stack to cross-reference other pieces and stay current with changes in other places.
+   The search term above perfectly illustrates the problem: we forgot about the extension. It's unwieldy and hard to quickly reach and share.
 
-1. Enables the infra to deploy the code to coexist with the code itself. Apart from the private secrets that will still need to be encrypted, the IaC files could be organised identical to the code.
+   **Monorepo solution:** We could use GitHub's own filters to narrow down what we're looking for.
 
-1. Milestones that can span across multiple layers of the stack are only possible in GitHub. This is a limitation imposed by GitHub and there is no workaround for this.
+1. Meta-issues:
+
+   **Current criticism:** If an issue spans more than a single layer of the stack, we need to open a meta issue in `WordPress/openverse`, open sub-issues in each of the different repos, then manually close meta issues after the sub-issues are closed. Same goes for PRs. We make individual PRs for every layer and then have to cross-reference them so that reviewers can see the full picture. Meta issues are good when a work needs to split into subtasks, but they are not good for cross-repo work splitting, especially when the work happens completely outside the knowledge of the meta-issue.
+
+   **Monorepo solution:** A monorepo allows our cross-layer PRs to be viewed more holistically and be reviewed as a complete change.
+
+1. No more sync:
+
+   Current criticism: We use complex sync workflows to keep files in sync. Some workflows need to by synced to some repos only. Some workflows shouldn't even be in the repo they're synced from. Some files need subtle differences so we compile Jinja templates for them. We also sync GitHub labels and branch management rules. It's a mess (I would know!).
+
+   **Monorepo solution:** A monorepo would just eliminates all of this and saves the time and effort that goes into maintaining these systems.
+
+1. Unified documentation:
+
+   **Current criticism:** Having many repos, each with its own doc site means two things. Common docs such as contribution process needs to be repeated several times and repo-specific docs get siloed and can only reference each other with external links. Also changing docs in one repo will break any links pointing to it.
+
+   **Monorepo solution:** A better system would be one cohesive doc site, for which the API already has a framework that other repos can just use.
+
+1. Infra included:
+
+   **Current criticism:** Our deployment workflows have code duplication. Secrets are stored in lots of repos, we keep secrets synced using Terraform. Containers are used in the infra repo but published in their individual repos.
+
+   **Monorepo solution:** Monorepo enables the infra to coexist with the code (albeit in a separate module). Apart from the (encrypted) private secrets, the IaC could be open-sourced similar to the rest of the codebase. Our deployment workflows can share code and deployment secrets.
+
+1. GitHub Milestones:
+
+   Milestones are confined by repository boundaries. To have milestones that cover issues in different layers of our stack, the only way is for them to be in a monorepo. This is a limitation imposed by GitHub and there is no workaround for this.
+
+The overarching theme is that there are workarounds for everything. We have been working with split repos quite productively for over a year. My proposition is the the monorepo solutions are better than workarounds.
 
 The [integration](#step-4-integration) section in the latter part of the document describes more interesting outcomes made possible by the monorepo. They may or may not be exclusive to monorepos but they're surely made easier by it.
 
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -59,7 +59,7 @@ improvement to convince everyone to migrate.
    compile Jinja templates for them. We also sync GitHub labels and branch
    management rules. It's a mess (I would know!).
 
-   **Monorepo solution:** A monorepo would just eliminates all of this and saves
+   **Monorepo solution:** A monorepo eliminates all of this and saves
    the time and effort that goes into maintaining these systems.
 
 1. Unified documentation:
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -79,9 +79,10 @@ This is a quick process.
    - `.editorconfig`
    - `justfile`
    - `.pre-commit-config.yaml`
-   - `.prettierignore` (symlink into the `frontend/` directory)
-   - `.eslintrc.js` (symlink into the `frontend/` directory)
-   - `.eslintignore` (symlink into the `frontend/` directory)
+   - `prettier.config.js`
+   - `.prettierignore`
+   - `.eslintrc.js` (need to update references in nested `eslintrc.js` files)
+   - `.eslintignore`
    - <s>`.gitignore`</s> (better to move it into the `frontend/` directory and update some absolute paths)
 
 1. Create the final commit on the `WordPress/openverse-frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -9,113 +9,194 @@
 
 ## Rationale
 
-For a comprehensive discussion about the pros, the cons and the counterpoints to each see [discussion](https://github.com/WordPress/openverse/issues/192). Some of the more nuanced points are listed below, biased towards the overall benefits of a monorepo to justify the RFC. This RFC also proceeds to go into the implementation details hoping that the benefits are cumulatively enough of an improvement to convince everyone to migrate.
+For a comprehensive discussion about the pros, the cons and the counterpoints to
+each see [discussion](https://github.com/WordPress/openverse/issues/192). Some
+of the more nuanced points are listed below, biased towards the overall benefits
+of a monorepo to justify the RFC. This RFC also proceeds to go into the
+implementation details hoping that the benefits are cumulatively enough of an
+improvement to convince everyone to migrate.
 
 ### Benefits of monorepo
 
 1. Single place for everything:
 
-   **Current criticism:** We currently have many repos, and issues and PRs spanning all of them. While this makes it easier for us as maintainers to focus our efforts, it's not easy for contributors. Let's say you were a new contributor looking for good-first Python issues. We shouldn't expect them to search in 3 repos `openverse`, `openverse-api` and `openverse-catalog`. We address this by making tools like Overvue or using search terms like this:
+   **Current criticism:** We currently have many repos, and issues and PRs
+   spanning all of them. While this makes it easier for us as maintainers to
+   focus our efforts, it's not easy for contributors. Let's say you were a new
+   contributor looking for good-first Python issues. We shouldn't expect them to
+   search in 3 repos `openverse`, `openverse-api` and `openverse-catalog`. We
+   address this by making tools like Overvue or using search terms like this:
 
    ```
    is:open is:issue repo:WordPress/openverse-catalog repo:WordPress/openverse repo:WordPress/openverse-api repo:WordPress/openverse-frontend repo:WordPress/openverse-infrastructure sort:updated-desc
    ```
 
-   The search term above perfectly illustrates the problem: we forgot about the extension. It's unwieldy and hard to quickly reach and share.
+   The search term above perfectly illustrates the problem: we forgot about the
+   extension. It's unwieldy and hard to quickly reach and share.
 
-   **Monorepo solution:** We could use GitHub's own filters to narrow down what we're looking for.
+   **Monorepo solution:** We could use GitHub's own filters to narrow down what
+   we're looking for.
 
 1. Meta-issues:
 
-   **Current criticism:** If an issue spans more than a single layer of the stack, we need to open a meta issue in `WordPress/openverse`, open sub-issues in each of the different repos, then manually close meta issues after the sub-issues are closed. Same goes for PRs. We make individual PRs for every layer and then have to cross-reference them so that reviewers can see the full picture. Meta issues are good when a work needs to split into subtasks, but they are not good for cross-repo work splitting, especially when the work happens completely outside the knowledge of the meta-issue.
+   **Current criticism:** If an issue spans more than a single layer of the
+   stack, we need to open a meta issue in `WordPress/openverse`, open sub-issues
+   in each of the different repos, then manually close meta issues after the
+   sub-issues are closed. Same goes for PRs. We make individual PRs for every
+   layer and then have to cross-reference them so that reviewers can see the
+   full picture. Meta issues are good when a work needs to split into subtasks,
+   but they are not good for cross-repo work splitting, especially when the work
+   happens completely outside the knowledge of the meta-issue.
 
-   **Monorepo solution:** A monorepo allows our cross-layer PRs to be viewed more holistically and be reviewed as a complete change.
+   **Monorepo solution:** A monorepo allows our cross-layer PRs to be viewed
+   more holistically and be reviewed as a complete change.
 
 1. No more sync:
 
-   Current criticism: We use complex sync workflows to keep files in sync. Some workflows need to by synced to some repos only. Some workflows shouldn't even be in the repo they're synced from. Some files need subtle differences so we compile Jinja templates for them. We also sync GitHub labels and branch management rules. It's a mess (I would know!).
+   Current criticism: We use complex sync workflows to keep files in sync. Some
+   workflows need to by synced to some repos only. Some workflows shouldn't even
+   be in the repo they're synced from. Some files need subtle differences so we
+   compile Jinja templates for them. We also sync GitHub labels and branch
+   management rules. It's a mess (I would know!).
 
-   **Monorepo solution:** A monorepo would just eliminates all of this and saves the time and effort that goes into maintaining these systems.
+   **Monorepo solution:** A monorepo would just eliminates all of this and saves
+   the time and effort that goes into maintaining these systems.
 
 1. Unified documentation:
 
-   **Current criticism:** Having many repos, each with its own doc site means two things. Common docs such as contribution process needs to be repeated several times and repo-specific docs get siloed and can only reference each other with external links. Also changing docs in one repo will break any links pointing to it.
+   **Current criticism:** Having many repos, each with its own doc site means
+   two things. Common docs such as contribution process needs to be repeated
+   several times and repo-specific docs get siloed and can only reference each
+   other with external links. Also changing docs in one repo will break any
+   links pointing to it.
 
-   **Monorepo solution:** A better system would be one cohesive doc site, for which the API already has a framework that other repos can just use.
+   **Monorepo solution:** A better system would be one cohesive doc site, for
+   which the API already has a framework that other repos can just use.
 
 1. Infra included:
 
-   **Current criticism:** Our deployment workflows have code duplication. Secrets are stored in lots of repos, we keep secrets synced using Terraform. Containers are used in the infra repo but published in their individual repos.
+   **Current criticism:** Our deployment workflows have code duplication.
+   Secrets are stored in lots of repos, we keep secrets synced using Terraform.
+   Containers are used in the infra repo but published in their individual
+   repos.
 
-   **Monorepo solution:** Monorepo enables the infra to coexist with the code (albeit in a separate module). Apart from the (encrypted) private secrets, the IaC could be open-sourced similar to the rest of the codebase. Our deployment workflows can share code and deployment secrets.
+   **Monorepo solution:** Monorepo enables the infra to coexist with the code
+   (albeit in a separate module). Apart from the (encrypted) private secrets,
+   the IaC could be open-sourced similar to the rest of the codebase. Our
+   deployment workflows can share code and deployment secrets.
 
 1. GitHub Milestones:
 
-   Milestones are confined by repository boundaries. To have milestones that cover issues in different layers of our stack, the only way is for them to be in a monorepo. This is a limitation imposed by GitHub and there is no workaround for this.
+   Milestones are confined by repository boundaries. To have milestones that
+   cover issues in different layers of our stack, the only way is for them to be
+   in a monorepo. This is a limitation imposed by GitHub and there is no
+   workaround for this.
 
-The overarching theme is that there are workarounds for everything. We have been working with split repos quite productively for over a year. My proposition is the the monorepo solutions are better than workarounds.
+The overarching theme is that there are workarounds for everything. We have been
+working with split repos quite productively for over a year. My proposition is
+the the monorepo solutions are better than workarounds.
 
-The [integration](#step-4-integration) section in the latter part of the document describes more interesting outcomes made possible by the monorepo. They may or may not be exclusive to monorepos but they're surely made easier by it.
+The [integration](#step-4-integration) section in the latter part of the
+document describes more interesting outcomes made possible by the monorepo. They
+may or may not be exclusive to monorepos but they're surely made easier by it.
 
 ## Migration path
 
-First we will merge the API and the frontend repos into `WordPress/openverse`. This decision was made for the following reasons.
+First we will merge the API and the frontend repos into `WordPress/openverse`.
+This decision was made for the following reasons.
 
-1. API and frontend are tightly linked. The frontend is a direct consumer of what the API produces.
+1. API and frontend are tightly linked. The frontend is a direct consumer of
+   what the API produces.
 
-1. The API and frontend form the "service" side of Openverse that directly faces the users (both API consumers and Search engine users).
+1. The API and frontend form the "service" side of Openverse that directly faces
+   the users (both API consumers and Search engine users).
 
-1. The frontend uses ECS deployments and the API is well on the same track. This makes it possible for them to share some deployment code.
+1. The frontend uses ECS deployments and the API is well on the same track. This
+   makes it possible for them to share some deployment code.
 
-1. I am very familiar with the scripts repo, the API and the frontend so merging them would be easier. Adding a third component would make the task daunting.
+1. I am very familiar with the scripts repo, the API and the frontend so merging
+   them would be easier. Adding a third component would make the task daunting.
 
-1. Merging incurs a productivity hit for the initial transition. So merging everything in one swoop is not ideal. While we merge these three, effort can be diverted to the catalog.
+1. Merging incurs a productivity hit for the initial transition. So merging
+   everything in one swoop is not ideal. While we merge these three, effort can
+   be diverted to the catalog.
 
-1. The API’s comprehensive tooling for developer documentation can benefit frontend devs and create a unified docs site for contributors.
+1. The API’s comprehensive tooling for developer documentation can benefit
+   frontend devs and create a unified docs site for contributors.
 
-1. The merge of two JavaScript codebases provides fertile ground for testing `pnpm` workspaces.
+1. The merge of two JavaScript codebases provides fertile ground for testing
+   `pnpm` workspaces.
 
-   - It also allows us to merge the browser extension later and split the design system/component library stuff into a separate package.
+   - It also allows us to merge the browser extension later and split the design
+     system/component library stuff into a separate package.
 
-1. The API is already organised by stack folders so the `frontend/` directory will fit right in with the others like `api/` and `ingestion_server/`. Similarly the scripts repo is nicely organised in folders, reducing conflicts.
+1. The API is already organised by stack folders so the `frontend/` directory
+   will fit right in with the others like `api/` and `ingestion_server/`.
+   Similarly the scripts repo is nicely organised in folders, reducing
+   conflicts.
 
-1. The API and frontend share identical tooling for Git hooks, linting and formatting. We will fight our tools less and encounter minimal friction.
+1. The API and frontend share identical tooling for Git hooks, linting and
+   formatting. We will fight our tools less and encounter minimal friction.
 
-   - The frontend's approach for `pre-commit` expanded this type of usage to the API as well!
+   - The frontend's approach for `pre-commit` expanded this type of usage to the
+     API as well!
 
-   - We're expanding the use of double-quoted strings to JavaScript to further unify our style guides.
+   - We're expanding the use of double-quoted strings to JavaScript to further
+     unify our style guides.
 
-1. The entire system can be integration tested during releases. The real API, populated with test data, can replace the Talkback server as long as we disable network calls and make output deterministic.
+1. The entire system can be integration tested during releases. The real API,
+   populated with test data, can replace the Talkback server as long as we
+   disable network calls and make output deterministic.
 
-The `WordPress/openverse` repo will absorb the `WordPress/openverse-api` and `WordPress/openverse-frontend` repos. The `WordPress/openverse-catalog` will also be merged, _later_.
+The `WordPress/openverse` repo will absorb the `WordPress/openverse-api` and
+`WordPress/openverse-frontend` repos. The `WordPress/openverse-catalog` will
+also be merged, _later_.
 
 ### Reference
 
-I'm following the steps listed below in a fork at <s>[@dhruvkb/monopenverse](https://github.com/dhruvkb/monopenverse/)</s> [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/). You can refer to the fork, but note that it is a comes from a place of haste and has not been treated with the same level of love and care that the final treatment will receive.
+I'm following the steps listed below in a fork at
+<s>[@dhruvkb/monopenverse](https://github.com/dhruvkb/monopenverse/)</s>
+[@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/). You can refer to
+the fork, but note that it is a comes from a place of haste and has not been
+treated with the same level of love and care that the final treatment will
+receive.
 
 ### Step 0: Prerequisites
 
 #### Get the timing right
 
-The first step will be to release the API and frontend, call a code freeze on both of them and pause work on both. This is to prevent the repos from continuing to drift as we merge a snapshot of them into the `WordPress/openverse` repo.
+The first step will be to release the API and frontend, call a code freeze on
+both of them and pause work on both. This is to prevent the repos from
+continuing to drift as we merge a snapshot of them into the
+`WordPress/openverse` repo.
 
-This can prove difficult given how productive our team is, so we will need to channel this productivity towards the catalog in the meantime. I can foresee the end-to-end migration taking one week (ideal scenario) to becoming workable again, and another week (for us to iron out any gaps in the docs and references).
+This can prove difficult given how productive our team is, so we will need to
+channel this productivity towards the catalog in the meantime. I can foresee the
+end-to-end migration taking one week (ideal scenario) to becoming workable
+again, and another week (for us to iron out any gaps in the docs and
+references).
 
 ##### Timeline breakdown
 
-- Day 1: Merging the repos and resolving conflicts, restoring broken workflows except deploys
+- Day 1: Merging the repos and resolving conflicts, restoring broken workflows
+  except deploys
 - Day 2: Restoring deployment workflows including staging auto-deploy
 - Day 3: Transfer of issues from individual repos to monorepo
 - Day 4: Documentation fixes
 - Day 5: Housekeeping
 
-The second week is planned as a buffer in case any of these tasks ends up taking more time than a day, if something breaks or if someone falls ill etc. The ideal scenario is that we're completely back next week, the worst one takes two.
+The second week is planned as a buffer in case any of these tasks ends up taking
+more time than a day, if something breaks or if someone falls ill etc. The ideal
+scenario is that we're completely back next week, the worst one takes two.
 
-Note that in the transition period nothing will break. The old repos will continue to exists as they are, till we ensure everything works and then we archive the current split repos.
+Note that in the transition period nothing will break. The old repos will
+continue to exists as they are, till we ensure everything works and then we
+archive the current split repos.
 
 ### Step 1: Merge with histories
 
-This should be quick save for a few merge conflicts. In case of conflict copy the code from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
+This should be quick save for a few merge conflicts. In case of conflict copy
+the code from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
 
 Add remotes.
 
@@ -158,8 +239,10 @@ $ git commit -m "Merge branch 'main' of openverse-api"
 - https://github.com/dhruvkb/monoverse/blob/main/.gitignore
 - https://github.com/dhruvkb/monoverse/blob/main/api/.gitignore
 - https://github.com/dhruvkb/monoverse/blob/main/ingestion_server/.gitignore
-- https://github.com/dhruvkb/monoverse/blob/main/ingestion_server/test/.gitignore (unchanged)
-- https://github.com/dhruvkb/monoverse/blob/main/load_testing/.gitignore (unchanged)
+- https://github.com/dhruvkb/monoverse/blob/main/ingestion_server/test/.gitignore
+  (unchanged)
+- https://github.com/dhruvkb/monoverse/blob/main/load_testing/.gitignore
+  (unchanged)
 - https://github.com/dhruvkb/monoverse/blob/main/nginx/.gitignore (unchanged)
 
 #### Merging frontend
@@ -200,7 +283,8 @@ Switch to the monorepo.
 $ git pull frontend main --allow-unrelated-histories
 ```
 
-Since we fixed these files when merging the API, almost all these conflicts are redundant.
+Since we fixed these files when merging the API, almost all these conflicts are
+redundant.
 
 | Expected conflicts      | Resolution                      |
 | ----------------------- | ------------------------------- |
@@ -214,9 +298,13 @@ Since we fixed these files when merging the API, almost all these conflicts are 
 
 #### Housekeeping
 
-1. Create "stack: \*" labels to help with issue and PR management. Spoiler/foreshadowing: these labels will be used for more things later.
+1. Create "stack: \*" labels to help with issue and PR management.
+   Spoiler/foreshadowing: these labels will be used for more things later.
 
-1. Migrate issues from `WordPress/openverse-frontend` and `WordPress/openverse-api`. We can directly transfer the issues, retaining all their comments. Apply the "stack: frontend" / "stack: backend" label to them after moving.
+1. Migrate issues from `WordPress/openverse-frontend` and
+   `WordPress/openverse-api`. We can directly transfer the issues, retaining all
+   their comments. Apply the "stack: frontend" / "stack: backend" label to them
+   after moving.
 
 ```bash
 # Substitute repo with WordPress/openverse-frontend and WordPress/openverse-api
@@ -236,25 +324,31 @@ $ gh api \
     -R WordPress/<repo>
 ```
 
-With this done, we can archive the API and frontend repo. An optional notice may be added to the `README.md` files for clarification before archiving.
+With this done, we can archive the API and frontend repo. An optional notice may
+be added to the `README.md` files for clarification before archiving.
 
 ### Step 1. Restore functionality
 
 #### Combine linting
 
-All lint steps can be combined in `.pre-commit-config.yaml`. This also implies the CI jobs can now be merged.
+All lint steps can be combined in `.pre-commit-config.yaml`. This also implies
+the CI jobs can now be merged.
 
-1. Remove pre-commit scripts from `frontend/package.json` and the `install-pre-commit.sh` script.
+1. Remove pre-commit scripts from `frontend/package.json` and the
+   `install-pre-commit.sh` script.
 
 1. Remove `lint` job from CI, there are plenty of those.
 
 #### `pnpm` workspace
 
-1. Move `frontend/.pnpmfile.cjs` outside to the root directory, update the reference to `frontend/package.json`.
+1. Move `frontend/.pnpmfile.cjs` outside to the root directory, update the
+   reference to `frontend/package.json`.
 
-1. Remove `frontend/.npmrc` created earlier because `pnpm` will automatically use the one in the root of the workspace.
+1. Remove `frontend/.npmrc` created earlier because `pnpm` will automatically
+   use the one in the root of the workspace.
 
-1. Create `pnpm-workspaces.yaml` file, see https://github.com/dhruvkb/monoverse/blob/main/.pre-commit-config.yaml.
+1. Create `pnpm-workspaces.yaml` file, see
+   https://github.com/dhruvkb/monoverse/blob/main/.pre-commit-config.yaml.
 
 1. Update the `package.json` files, see the following:
 
@@ -264,7 +358,8 @@ All lint steps can be combined in `.pre-commit-config.yaml`. This also implies t
 
 1. `pnpm i` in the monorepo root.
 
-1. Update the recipe `pnpm` in `frontend/justfile` to include `--filter "openverse-frontend"`.
+1. Update the recipe `pnpm` in `frontend/justfile` to include
+   `--filter "openverse-frontend"`.
 
 1. `git commit -m "Setup workspaces"`
 
@@ -272,7 +367,9 @@ All lint steps can be combined in `.pre-commit-config.yaml`. This also implies t
 
 #### New actions
 
-To clean up the workflows we will define three new actions. The code for all three is available at [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
+To clean up the workflows we will define three new actions. The code for all
+three is available at
+[@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
 
 1. `setup-env` to setup Just, Node.js and Python or a subset of these.
 1. `load-img` to download Docker images into `/tmp` and load them in Docker.
@@ -290,26 +387,38 @@ Updates:
 - `migration_safety_warning.yml`
 - `generate_pot.yml`
 
-With this done, the development on the API and frontend can continue inside their subdirectories. The development of both parts will be independent, at least until we reach [long-term consolidation](#step-5-long-term-consolidation).
+With this done, the development on the API and frontend can continue inside
+their subdirectories. The development of both parts will be independent, at
+least until we reach [long-term consolidation](#step-5-long-term-consolidation).
 
 #### Deployment
 
 ##### Staging
 
-The soon-to-be-ECS-based API and ECS-based frontend will continue to deploy to staging via the CI + CD pipeline, with deployment starting as soon as all CI checks have passed. They will use similar code as the frontend auto-deploy for staging used currently.
+The soon-to-be-ECS-based API and ECS-based frontend will continue to deploy to
+staging via the CI + CD pipeline, with deployment starting as soon as all CI
+checks have passed. They will use similar code as the frontend auto-deploy for
+staging used currently.
 
-These will be separate jobs with specific [path-based filters](https://github.com/dorny/paths-filter).
+These will be separate jobs with specific
+[path-based filters](https://github.com/dorny/paths-filter).
 
 ##### Production
 
-For production, we will not be able to use GitHub Releases and will have to use a manually-triggered workflow to build the assets and tag them. The tag can be set via a workflow input (simple) or can be calculated based on the update scope of major, minor or patch (not as simple).
+For production, we will not be able to use GitHub Releases and will have to use
+a manually-triggered workflow to build the assets and tag them. The tag can be
+set via a workflow input (simple) or can be calculated based on the update scope
+of major, minor or patch (not as simple).
 
 ### Step 3. Housekeeping and DX cleanup
 
-There will be a few rough edges that I cannot foresee and we can continuously fix those as we spot them. But up to this point we should be in a position where
-we can continue to build the API and the frontend independently but from one repo.
+There will be a few rough edges that I cannot foresee and we can continuously
+fix those as we spot them. But up to this point we should be in a position where
+we can continue to build the API and the frontend independently but from one
+repo.
 
-1. The action `banyan/auto-label` will need to be configured (`auto-label.json`) to add the "stack: \*" labels based on the modified directory.
+1. The action `banyan/auto-label` will need to be configured (`auto-label.json`)
+   to add the "stack: \*" labels based on the modified directory.
 
 ### Step 4. Documentation merge
 
@@ -329,6 +438,9 @@ I will need more information about this because IANAL.
 
 ### Step 5. Long-term consolidation
 
-This is the long term combination of code for the frontend and the API. Ideas like end-to-end testing go here. This is beyond my imagination at the moment, and more importantly, beyond the scope of this RFC. It will surely be covered in future RFCs.
+This is the long term combination of code for the frontend and the API. Ideas
+like end-to-end testing go here. This is beyond my imagination at the moment,
+and more importantly, beyond the scope of this RFC. It will surely be covered in
+future RFCs.
 
 Thanks for reading and providing feedback.
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -84,7 +84,7 @@ This is a quick process.
    - `.eslintignore` (symlink into the `frontend/` directory)
    - <s>`.gitignore`</s> (better to move it into the `frontend/` directory and update some absolute paths)
 
-1. Create the final commit on the `WordPress/frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
+1. Create the final commit on the `WordPress/openverse-frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
 
 1. Merge this repo's `main` branch into the `WordPress/openverse-api` repo's `main` branch (see Git docs for `--allow-unrelated-histories`). There will be some conflicts but they will be small and infrequent. [[implementation details](#conflict-resolution)]
 
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -6,8 +6,8 @@
 
 ## Reviewers
 
-- [ ] &lt;your name here&gt;
-- [ ] &lt;your name here&gt;
+- [ ] @zackkrida
+- [ ] @sarayourfriend
 
 ## Rationale
 
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -328,8 +328,14 @@ $ gh api \
     -R WordPress/<repo>
 ```
 
-With this done, we can archive the API and frontend repo. An optional notice may
-be added to the `README.md` files for clarification before archiving.
+With this done, we can archive the API and frontend repo. The following notice
+will be added to the `README.md` files for clarification before archiving.
+
+> **Note**
+>
+> This repository has moved to
+> [WordPress/openverse](https://github.com/wordpress/openverse) as part of a
+> monorepo.
 
 ### Step 1. Restore functionality
 
--- a/rfcs/20221124-monorepo.md
+++ b/rfcs/20221124-monorepo.md
@@ -195,8 +195,12 @@ archive the current split repos.
 
 ### Step 1: Merge with histories
 
-This should be quick save for a few merge conflicts. In case of conflict copy
-the code from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
+This should be quick save for a few merge conflicts. In case of conflict refer
+to the resolution adopted by
+[@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/), and apply the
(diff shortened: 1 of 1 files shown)


rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,122 @@
...
+
+For a comprehensive discussion about the pros, the cons and the counterpoints to each see [discussion](https://github.com/WordPress/openverse/issues/192). This is not the purpose of this RFC.
+
+This RFC summarily lists the benefits and then, with the twin assumptions of a monorepo being ultimately beneficial and the decision to migrate being finalised in the above discussion, proceeds to go into the implementation details.
+
+### Exclusive benefits of monorepo
+
+This only includes things that cannot be accomplished without the use of a monorepo.
+
+1. Single place to go for issues, PRs and all activity. Currently tickets are scattered across several repos, and any tickets that could benefit more than a single layer must in opened in each of the different repos.
```
zackkrida, line 22:
```suggestion
1. Single place to go for issues, PRs and all activity. Currently tickets are scattered across several repos, and any tickets that could benefit more than a single layer must be opened in each of the different repos.
```

small typo

dhruvkb:
After extensive testing in the PoC, everything is working. In the PoC, no existing functionality is broken and a lot of things are simplified, merged or otherwise improved.

In my testing, the monorepo migration was a net positive so I suggest we move forward with it.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+This only includes things that cannot be accomplished without the use of a monorepo.
+
+1. Single place to go for issues, PRs and all activity. Currently tickets are scattered across several repos, and any tickets that could benefit more than a single layer must be opened in each of the different repos.
+
+1. Singular copy (different from synced independent copies) of scaffolding code such as Git hooks, lint rules and common workflows. This is distinctly better than elaborate sync workflows.
+
+1. Central place for all technical documentation, enabling documentation for different parts of the stack to cross-reference other pieces and stay current with changes in other places.
+
+1. Enables the infra to deploy the code to coexist with the code itself. Apart from the private secrets that will still need to be encrypted, the IaC files could be organised identical to the code.
```
sarayourfriend, line 28: > the IaC files could be organised identical to the code.

Just a nit but this isn't entirely true. We could _maybe_ spread _some_ Terraform modules out this way but it would be a pain to have so many places to search for modules. If the Terraform code will be merged into the monorepo as well then it'd be better off assuming that it will live, in totality, in a separate directory.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+1. The API’s comprehensive tooling for developer documentation can benefit frontend devs and create a unified docs site for contributors.
+
+1. The API is already organised by stack folders so the `frontend/` directory will fit right in with the others like `api/` and `ingestion_server/`.
+
+1. The API and frontend share identical tooling for Git hooks, linting and formatting. We will fight our tools less and encounter minimal friction.
+
+   - In fact, we employ a number of hacks to install and configure pre-commit for the frontend. Merging it with the API eliminates the need for such hacks.
+
+1. The entire system can be integration tested during releases. The real API, populated with test data, can even replace the Talkback server.
```
sarayourfriend, line 56: Interesting idea! For this to be feasible, we would need to come up with a definitive solution to preventing the API from making network requests when it executes a search. Right now Playwright tests do not depend _at all_ on any external network and that needs to remain the case.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+The `WordPress/openverse-api` repo will absorb the `WordPress/openverse-frontend` repo. The `WordPress/openverse-catalog` will also be merged, _later_.
```
sarayourfriend, line 58: Will there be a period of time during which frontend work is happening only in `openverse-api`? Is the long-term plan to merge everything into `openverse-api` and then eventually rename that to just `openverse`?

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Reference
+
+I'm following the steps listed below in a fork at [@dhruvkb/monopenverse](https://github.com/dhruvkb/monopenverse/). You can refer to the fork, but note that it is a comes from a place of haste and has not been treated with the same level of love and care that the final treatment will receive.
+
+### Step 0: Prerequisites
+
+#### Get the timing right
+
+The first step will be to release the frontend, call a code freeze and pause work on it. This is to prevent the frontend repo from continuing to drift as we merge a snapshot of it with the API.
```
sarayourfriend, line 68: Should we schedule a bug bash before this so that we can identify and resolve any outstanding issues before we enter the period of time where we won't be able to without great disruption?

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+This can prove difficult given how productive our team is, so we will need to channel this productivity towards the catalog in the meantime. I can foresee the end-to-end migration taking one week (ideal scenario) to one fortnight (worst case scenario).
+
+### Step 1: Merge with histories
+
+This is a quick process.
+
+1. Move the entire content of frontend inside a `frontend/` directory, except the following top-level files and folders. Please comment if you can add to this list.
```
sarayourfriend, line 76: What is the plan for making `pnpm` work in the monorepo? Do we want to go ahead and making `pnpm` live at the root of the repository and treat the frontend as a `pnpm` monorepo package? https://pnpm.io/workspaces

I think this would be necessary for us to be able to (reasonably) merge the browser extension and the JS that is currently in WordPress/openverse into the monorepo... otherwise we'll be fighting weird fights with pnpm in different directories :thinking:

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+
+   - `.github/`
+   - `.editorconfig`
+   - `justfile`
+   - `.pre-commit-config.yaml`
+   - `.prettierignore` (symlink into the `frontend/` directory)
+   - `.eslintrc.js` (symlink into the `frontend/` directory)
+   - `.eslintignore` (symlink into the `frontend/` directory)
```
sarayourfriend, line 84: Why do these need to be symlinked?

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+   - <s>`.gitignore`</s> (better to move it into the `frontend/` directory and update some absolute paths)
+
+1. Create the final commit on the `WordPress/frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
```
sarayourfriend, line 87:
```suggestion
1. Create the final commit on the `WordPress/openverse-frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
```

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+- CODE_OF_CONDUCT.md (both repos)
+- CONTRIBUTING.md (both repos)
+- CONTRIBUTORS.md (API only; also why?)
+- DOCUMENTATION_GUIDELINES.md (API only)
+- TESTING_GUIDELINES.md (frontend only)
+- DEPLOYMENT.md (frontend only)
+
+I will need more information about this because IANAL.
+
+- LICENSE (both repos)
```
sarayourfriend, line 178: Given Openverse still uses MIT we have probably very little to worry about here, FWIW.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,178 @@
...
+### Step 5. Documentation merge
+
+The following documentation files will need reorganisation or merge.
+
+- README.md (both repos)
+- CODE_OF_CONDUCT.md (both repos)
+- CONTRIBUTING.md (both repos)
+- CONTRIBUTORS.md (API only; also why?)
+- DOCUMENTATION_GUIDELINES.md (API only)
+- TESTING_GUIDELINES.md (frontend only)
```
sarayourfriend, line 173: We would probably want to move this into the frontend directory, I would think. It is extremely frontend specific. Or would it go into the Sphinx docs?

dhruvkb: My long-term goal is for Sphinx docs to be the single source of reference for all developers, not just API devs.

Some Markdown files that people generally look for in the repo root could be left there but ideally, even they would just point to the relevant Sphinx pages for the real stuff.

rfcs/20221124-monorepo.md
dhruvkb: Nice, but I had a vague feeling that merging repositories could have legal implications for those that use them somehow.

rfcs/20221124-monorepo.md
dhruvkb: Yes. The timeline will be 
1. pause work on the frontend
2. merge `WordPress/openverse-frontend` into `WordPress/openverse-api`
3. transfer issues and other relevant GitHub models 
4. get fronted functional
5. resume frontend work in `WordPress/opnenverse-api`
6. archive `WordPress/opnenverse-frontend`

The one week to one fortnight period mentioned in the RFC is the time between step 1 and step 5.

rfcs/20221124-monorepo.md
dhruvkb: I think it can be done (maybe something like an environment variable flag to not make requests) and I can envision it being glorious when it works.

rfcs/20221124-monorepo.md
dhruvkb: That was an issue. Currently I had everything working with `pnpm` with `package.json` moved into the `frontend/` directly (no package manager at the root), but I definitely should look into the `workspaces` feature.

rfcs/20221124-monorepo.md
dhruvkb: That's an error on my part. I was performing lots of experiments in https://github.com/dhruvkb/monopenverse and with linting to work so I tried almost everything. When I got it to work finally, I forgot to update the RFC with the right approach.

dhruvkb: Fixed in f2122d8.

rfcs/20221124-monorepo.md
dhruvkb: Ah, my lack of knowledge in the IaC space is really showing here. I would appreciate if you could add any infra issues (especially any dealbreakers) and any infra benefits that you see popping up due to this merge. Thanks!

rfcs/20221124-monorepo.md
dhruvkb: Yes, that would be ideal (maybe even necessary). Before embarking on this potential two-week freeze, it would be best to have the frontend be in a state that could serve two weeks without maintenance.

If we have no high or critical priority tickets, that would indicate optimal time to work on this move.

rfcs/20221124-monorepo.md
sarayourfriend: From an infra perspective I don't think there are any benefits really. The only thing I could see is if we did this issue: https://github.com/WordPress/openverse-infrastructure/issues/220. The benefit of that issue could be a more easily reusable workflow with better variables in the deployment workflows we have now. Basically everything that is used to define the deployment workflows are variables that we could theoretically move into GitHub, thought it's just more manual management of those details. We could try to sync them other ways like with a GitHub secret of a JSON blob that gets passed through so the workflow call is something like:

```yaml
- uses: ./actions/deploy.yml
  with:
    config: ${{ secrets.NUXT_DEPLOYMENT_CONFIG }}
```

And then the workflow parses the JSON to get the configuration.

Note: For folks without access to the infrastructure repository, the issue above describes switching to a callable workflow instead of a composite action. I've found composite actions a little more tedious so I avoided trying to make the current deployment workflows super modular. I think a callable workflow would be easier to work with for a _lot_ of reasons and could open up the possibility of making a more module/generic single-deployment workflow file a little more realistic.

We could do all of this in our current setup, but a monorepo would end up with less places to sync the secrets to and less places to put the deployment action.

Actually, that reminds me, if the infrastructure is in the same repository, the deployment workflow outputs could be automatically merged whenever changes are made to them. Right now we have to manually sync them from the infrastructure repository to wherever they're needed. So that would be a nice simplification of that process as well (one less thing to remember to do!) :slightly_smiling_face:

rfcs/20221124-monorepo.md
sarayourfriend: If we are planning to do a bug discovery session during the Istanbul meetup in January, maybe targeting this for late February (or later/earlier depending on what bugs we find) could be good, provided everyone is on board. It'd be nice to know that ahead of time as well so that we could block off the time and expect frontend feature work to undergo a stoppage at a specific period of time rather than just "sometime in the future". Did you have any particular alternative scheduling ideas in mind?

rfcs/20221124-monorepo.md
sarayourfriend: I love that plan!

rfcs/20221124-monorepo.md
dhruvkb: The bug bash we had in Venice was very productive so repeating it seems logical. But there are two potential issues with that:
- We only identified and filed issues in the last meetup, fixing them afterwards.
- I can't speak for everyone but my productivity immediately after the meetup was very low.

IMO the best time for this migration would be when all the frontend devs' productivity has gone down (less activity = easier migration) so if we could iron out the bugs **before** the meetup, the meetup would be the best time to migrate this because we'd all be together to collaborate in the migration process.

We do need to set down a date for this but we should only do this after the iframe removal project is deployed, and has been stable for a week or two. Other than that, we can make any time frame work.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Enables the infra to deploy the code to coexist with the code itself. Apart from the private secrets that will still need to be encrypted, the IaC files could be organised identical to the code.
+
+1. Milestones that can span across multiple layers of the stack are only possible in GitHub. This is a limitation imposed by GitHub and there is no workaround for this.
+
+The [integration](#step-4-integration) section in the latter part of the document describes more interesting outcomes made possible by the monorepo. They may or may not be exclusive to monorepos but they're surely made easier by it.
+
+## Migration path
+
+First we will merge the API and the frontend. This decision was made for the following reasons.
```
stacimc, line 36: This reasoning makes a lot of sense to me, and I'm sold on merging API/frontend first and separately from the Catalog. How much more difficult do you anticipate eventually merging the Catalog to be? Do you think there's value in planning that work first, or can that be tabled until after the API/frontend are merged?

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
+↖️ Table of Contents
+
+# RFC: Monorepo
+
+**Status:** 🚧 WIP, comments are welcome nonetheless
+
+## Reviewers
+
+- [ ] &lt;your name here&gt;
+- [ ] &lt;your name here&gt;
```
zackkrida, line 10:
```suggestion
- [ ] @zackkrida 
- [ ] @sarayourfriend 
```

Does this reviewer pairing seem okay? My rationale is that @sarayourfriend perhaps works across the most repositories the most often, and also authored some of the more critical actions related to releasing and deploying. And myself because for now I've been the most involved in the planning of projects.

rfcs/20221124-monorepo.md
dhruvkb: I am fairly confident that merging the catalog would not cause major pain later on, mostly because it's another Python project like the API and ingestion server.

We could definitely plan for the eventual catalog-API merge, but I abstained from that in the RFC because
- I am not the most experienced with the catalog so I felt a bit underqualified to write about that in the RFC
- I didn't want to delay the RFC till I could understand the nuances of the catalog enough to include the portion about merging it

rfcs/20221124-monorepo.md
dhruvkb: I like this. But I do want to stress that this affects everyone so I'd appreciate if everyone took at least one look (no comments necessary).

zackkrida: Definitely, I think it's very reasonable to expect everyone to read and consider the proposal.

zackkrida:
One note regarding translations I'd like to make sure is mentioned in this RFC:

We will have to contact a member of the WP Meta team (who works on wp.org) to change the path at which GlotPress reads our translation files, to both update to the new repository and filepath. I believe the current path used is `https://github.com/WordPress/openverse-frontend/blob/translations/openverse.pot`

rfcs/20221124-monorepo.md
sarayourfriend: Maybe we should try to do a bug bash before the meetup then. We could even start the monorepo migration _during_ the meetup given the core contributors will not likely be spending much time making direct frontend contributions. That way we could easily pair on any unexpected issues that crop up.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+The following conflicts may occur during merge.
+
+- `.prettierignore`: concatenate
+- `.pre-commit-config.yaml`: use from [dhruvkb/monopenverse](https://github.com/dhruvkb/monopenverse)
+- Workflows can conflict but they can be renamed and kept alongside each other, _for now_.
+
+#### Issue transfer
+
+As far as I can tell, issue transfer can only be performed via the GitHub GraphQL API ([docs](https://docs.github.com/en/graphql/reference/mutations#transferissue)) and not via the REST API. From my limited testing, transferred issues seem to retain labels (provided they exist in the target repo).
```
zackkrida, line 108: I wrote a script in my previous monorepo exploration to transfer issues from one repo (`openverse/service` is the sample name I use, `WordPress/openverse-frontend` would be a concrete example, into another repo:

`gh api -X GET repos/openverse/service/issues -f pulls=false -f state=all --jq '.[].number' --paginate | xargs -n 1 -I num gh issue transfer num openverse/monorepo -R openverse/service`

In testing it worked quite well, and quickly.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+   - `.editorconfig`
+   - `justfile`
+   - `.pre-commit-config.yaml`
+   - `prettier.config.js`
+   - `.prettierignore`
+   - `.eslintrc.js` (need to update references in nested `eslintrc.js` files)
+   - `.eslintignore`
+   - <s>`.gitignore`</s> (better to move it into the `frontend/` directory and update some absolute paths)
+
+1. Create the final commit on the `WordPress/openverse-frontend` repo. After the merge we might want to add a notice about the migration to the `README.md` file but GitHub's built-in archival process could suffice here.
```
zackkrida, line 88: I think we _definitely_ want to add a README notice and use the GitHub archiving functionality together. The README information will crucially link to the new repository. We should also update the repo description with the new url as well, here:

CleanShot 2022-12-02 at 13 20 13

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+I'm following the steps listed below in a fork at [@dhruvkb/monopenverse](https://github.com/dhruvkb/monopenverse/). You can refer to the fork, but note that it is a comes from a place of haste and has not been treated with the same level of love and care that the final treatment will receive.
+
+### Step 0: Prerequisites
+
+#### Get the timing right
+
+The first step will be to release the frontend, call a code freeze and pause work on it. This is to prevent the frontend repo from continuing to drift as we merge a snapshot of it with the API.
+
+This can prove difficult given how productive our team is, so we will need to channel this productivity towards the catalog in the meantime. I can foresee the end-to-end migration taking one week (ideal scenario) to one fortnight (worst case scenario).
```
zackkrida, line 70: > I can foresee the end-to-end migration taking one week (ideal scenario) to one fortnight (worst case scenario).

Could you unpack this time estimate a bit more for me? I am curious because I could see these tasks taking an hour or so and being treated like a deploy, where a few folks work on it together synchronously. I also think that would have benefit of minimizing the disruption to contributors. Are there long-running tasks I'm not considering?

zackkrida:
This looks great, @dhruvkb. I'm really excited to move forward with this and hope it's what the team wants. Some questions:

1. This RFC only covers the frontend + api combination. Would there be a follow-up RFC for merging the Catalog and plain `WordPress/openverse` repositories later on? 
2. Do we need to write a follow up implementation plan? I see a lot of implementation details here which are great, but I also see some details you may not have considered, or perhaps considered out of scope for this RFC if the intention is to write a separate implementation plan:
 - The root `README.md`. Will we write a new one? We should draft that earlier so we have it prepared, and perhaps move the current API README into the `api/` directory. For example, we'll need to explain the new `just install` will install frontend and API dependencies, things like that.
 - What will the unified repository be called? `openverse-api` would no longer make sense. 
 - We will need to revise any and all existing documentation that references the frontend and api repositories and update them with reference to the new, combined repo. 
 - Just reiterating here that we'll need to update the path at which GlotPress reads our translation files, to both update to the new repository and filepath. I believe the current path used is https://github.com/WordPress/openverse-frontend/blob/translations/openverse.pot
 - We will need to write migration instructions for contributors, who will have local frontend repositories that now point to an archived repo, or api contributors who will now have significant file conflicts with the api repo.
3. Have you observed the duration of lint/required status actions to know if it will be a hindrance to development? For example, say we're attempting to push a small frontend hotfix, will `actionlint` now take dramatically longer? 

When I clone `monopenverse` it's still quite performant and responsive in VS Code, which is great. I was concerned about the size being too memory-intensive to search and things, but it's working fine. 

I observed one problem with prettier in VS Code, the root-level `prettier.config.js` depends on the `prettier-plugin-tailwindcss` which is in the `node_modules` directory of the frontend. VS Code doesn't seem to like this and looks for prettier in the repo root. To prevent future issues like this I do think it would be worth trying @sarayourfriend's workspaces approach to lift up node dependencies into the root of the repo, which will also somplify things when we combine the scripts (WordPress/openverse) and browser extension (should we renew our efforts to support it) repos into this one.

zackkrida:
Also, 👍 the combined codeowners file is excellent!

zackkrida:
Here are the `scc` stats on the combined `monopenverse` repo, just for anyone curious:

```shell
monopenverse on  main +21/-21 [!] via  v16.14.2
❯ scc
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
Python                     205     17329     2039      1391    13899        430
TypeScript                 174     12013     1410      1817     8786       1211
Vue                        142     15803     1460      1537    12806        339
JavaScript                  93      8145      899       758     6488        330
SVG                         53       295        0         0      295          0
Markdown                    44      1963      621         0     1342          0
YAML                        33     20749     2454       123    18172          0
JSON                        23      6839        7         0     6832          0
SQL                         14      1837      554       577      706          0
ReStructuredText            11       526      130         0      396          0
Shell                        7       268       39        31      198         18
TypeScript Typings           6       107       10        12       85          2
HTML                         5       213       19         5      189          0
gitignore                    5       194       35        39      120          0
CSS                          4       210       27        28      155          0
CSV                          4     20004        0         0    20004          0
Docker ignore                3        52        3         6       43          0
Dockerfile                   3       222       64        72       86         15
License                      2        42        8         0       34          0
Plain Text                   2      3022        3         0     3019          0
XML                          2       126        0         0      126          0
───────────────────────────────────────────────────────────────────────────────
Total                      835    109959     9782      6396    93781       2345
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop (organic) $3,179,185
Estimated Schedule Effort (organic) 21.34 months
Estimated People Required (organic) 13.23
───────────────────────────────────────────────────────────────────────────────
Processed 31224426 bytes, 31.224 megabytes (SI)
───────────────────────────────────────────────────────────────────────────────
```

dhruvkb:
^ Sorry, accidental click!

Reply to @zackkrida's comment https://github.com/WordPress/openverse/pull/336#issuecomment-1335670139:

1. Yes, we could write a second RFC for merging the catalog with the API+frontend repo although I imagine it would be a lot simpler than merging with the frontend due to the catalog being Python.

2. I wanted to include the implementation steps in the RFC but there might be scope to be more thorough.  I'll update the RFC with more implementation details. A separate RFC would not be ideal because this one already has parts of implementation.
 
    1. I focused mostly on the technical implementation compared to text so that we can be sure that there won't be any unexpected issues when merging. There might be many documentation pieces that will need to be updated after the merge.
    2. I have no ideas for the name. Tbh, `openverse` would've been ideal but that's already in use by the scripts repo. We could consider merging both the API and the frontend into `WordPress/openverse` but that'll need more experimentation and massive updates to the RFC.
    3. Yes.
    4. This is not a breaking change. The existing GlotPress links will be valid and the switch will be seamless. The new repo will also serve the same `translations` branch (action verified working successfully)
    5. Any users of the API will not face conflicts updating there code. The final merged API structure will be free of merge conflicts and will have a linear history so `git pull` would suffice. As for the frontend repo, people should just delete their local folders (provided there isn't work they want to save).

3. The actions are very fast. Most new steps added in CI run in parallel. The lint action does have some additional steps but they do not slow things down. In my testing, nothing took exceptionally more time that it does now.

4. The root level `prettier.config.js` does depend on `plugin-tailwindcss` but it does not depend on `node_modules` of the frontend. Prettier's execution is managed by `pre-commit` which populates the node_modules for it as described in `.pre-commit-config.yaml`. VS Code might complain because it does not know about this but in practicality nothing breaks.

5. There are still valid reasons to try pnpm workspaces and I have started exploration into that. Will update the RFC with my findings.

Action items
- explore `pnpm` workspaces [pt 5]
- see if `WordPress/openverse` might be a candidate for the merge making the `openverse` name available for the repo [pt 2.ii]

rfcs/20221124-monorepo.md
dhruvkb: The missing implementation details (such as writing new documentation, updating references) are what will take the maximum time in the 1-2 week time range I proposed for this plan.

The technical merge is a and getting the functionality working is a day or two. The remaining 3 days are for identifying these issues and fixing them. I kept 5 more days as a backup in case something unforeseen comes up but given we do sufficient planning beforehand, we will not need the second week (underpromise, overdeliver 😉).

It might be hard to make a draft PR with the changes (because of the merge conflicts) but one thing we could do is start a new branch, do the merge on it and eventually replace `main` with it. That seems like a great way to see everything in action but a lot of (branch-name dependent actions will be affected when we promote the other branch).

In any case, the frontend cannot be merged via the GitHub UI, it will have be CLI. In all my usage of GitHub I don't think it's possible to make a PR from any repo to another. The GitHub UI only provides merge functionality from a fork to the upstream which is not this case.

AetherUnbound:
This RFC looks great! I'm not a required reviewer but I think the plan outlined here is solid 😄 Thanks for all your experimentation and work to set this up @dhruvkb. Would we want to make issues once this is approved on the API repo to track the merging of all of the documents/workflows you outline here?

dhruvkb:
Yes, @AetherUnbound. I will make a milestone and several smaller issues once the team (or atleast a majority of the team, with no one veto-ing) agrees that a monorepo would be beneficial.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+For a comprehensive discussion about the pros, the cons and the counterpoints to each see [discussion](https://github.com/WordPress/openverse/issues/192). This is not the purpose of this RFC.
+
+This RFC summarily lists the benefits and then, with the twin assumptions of a monorepo being ultimately beneficial and the decision to migrate being finalised in the above discussion, proceeds to go into the implementation details.
+
+### Exclusive benefits of monorepo
+
+This only includes things that cannot be accomplished without the use of a monorepo.
+
+1. Single place to go for issues, PRs and all activity. Currently tickets are scattered across several repos, and any tickets that could benefit more than a single layer must be opened in each of the different repos.
```
krysal, line 22: Not something exclusive of a monorepo, you can already [view all the issues of Openverse repositories aggregated in a single page](https://github.com/issues?q=is%3Aopen+is%3Aissue+repo%3AWordPress%2Fopenverse-catalog+repo%3AWordPress%2Fopenverse+repo%3AWordPress%2Fopenverse-api+repo%3AWordPress%2Fopenverse-frontend+repo%3AWordPress%2Fopenverse-infrastructure), including the private infrastructure repo.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Singular copy (different from synced independent copies) of scaffolding code such as Git hooks, lint rules and common workflows. This is distinctly better than elaborate sync workflows.
+
+1. Central place for all technical documentation, enabling documentation for different parts of the stack to cross-reference other pieces and stay current with changes in other places.
```
krysal, line 26: What is wrong with old normal links? We can already cross-reference info if it's necessary.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Enables the infra to deploy the code to coexist with the code itself. Apart from the private secrets that will still need to be encrypted, the IaC files could be organised identical to the code.
+
+1. Milestones that can span across multiple layers of the stack are only possible in GitHub. This is a limitation imposed by GitHub and there is no workaround for this.
```
krysal, line 30: I see manual meta issues as good enough workaround.

krysal (commented):
Leaving short comments as you already said, this is a significant change and affects everyone's development and workflow.

I was a bit interested in reading a document describing the many benefits that a monorepo could bring, given the excitement of the rest of the team. Unfortunately, this has not changed my mind. The most interesting parts seem to be still in an hypothetical field, and given the level of disruption of this movement, I'd prefer to remain with the same split repositories.

Ultimately, it seems a matter of preference. The previous discussion stated several cons and "faux pros" ignored here and there. I don't see the point of commenting on the rest of the document if the team would like to follow this path no matter what and thinks it would be beneficial. I hope it really improves the workflow for everyone else (I'm not so sure it would be my case).

sarayourfriend:
> Ultimately, it seems a matter of preference. The previous discussion stated several cons and "faux pros" ignored here and there. I don't see the point of commenting on the rest of the document if the team would like to follow this path no matter what and thinks it would be beneficial. I hope it really improves the workflow for everyone else (I'm not so sure it would be my case).

FWIW, @krysal I am also pretty sceptical that this would dramatically change or fix any of the problems I personally see as the biggest hurdles the project currently faces in processes and development. On the other hand, it would unequivocally simplify certain DevEx issues like the various synced workflows and linting configurations. Whether that is actually valuable in the face of the trade-offs is a discussion that I feel hasn't been given its due course. I would encourage other folks to share their objections to moving forward with a monorepo, agnostic of the implementation details, just as you have.

For my part I was hoping that the RFC would have a point by point summary of the pros and cons of a monorepo alongside potential responses for each (especially the cons) and explicit alternative solutions to the solutions that are currently presented as being benefits of the monorepo. But I also understand why Dhruv did not take that on. The discussion has happened several times in different places, both in GitHub, in private discussions, and in Make Slack. Because we have not historically been great at coming to actionable conclusions in our discussions, some of those questions continue to linger.

Additionally, it feels that those conversations never really led to a "vote" or consensus, at least not from my perspective. We don't even really have a super clear way of doing that at the moment. I wish we did so that we could make sure folks are on board with large systemic changes like the ones implied by a monorepo change without accidentally leaving any objections unheard and unaddressed.

I appreciate your individual comments on the benefits. I had a lot of the same remarks but kept them to myself because I assumed everyone was on board and I didn't want to re-hash anything that I thought had already been discussed elsewhere (not in specific places, this was just a bad assumption on my part). But if we want to have a comprehensive pro/cons to each monorepo and not-monorepo, I would be fully on board with that and happy to contribute in the research as well as the discussion. The points you've mentioned in response to the benefits listed in the RFC seem like a sensible starting point for that discussion.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
+↖️ Table of Contents
+
+# RFC: Monorepo
+
+**Status:** 🚧 WIP, comments are welcome nonetheless
```
zackkrida, line 5: Is this still accurate? Or can we consider this complete? I would suggest if it's ready we add a completion date that review should be done by here and in the PR description. Moving forward we'll use a two-week deadline. This has already been open for 12-ish days, as I recall, but some folks are starting to share concerns, so I would suggest December 14th as a due date allowing for sufficient time for additional feedback.

rfcs/20221124-monorepo.md
zackkrida: > I see manual meta issues as good enough workaround.

They definitely might be. We would need to establish some conventions and patterns for using them more effectively, but GitHub _did_ recently fix some bugs with showing the titles of issues from other repos. Here's an example of a cross-repo meta issue:

https://github.com/WordPress/openverse/issues/341

It's really nice! Meta issues are also much more flexible than milestones, as milestones do not allow for comments or discussion. Regardless of switching to a monorepo or not we might want to use meta issues more _anyway_.

One _minor_ benefit to the monorepo is that you get autocompletion of issues in the current repo, but not for cross-repo issues: 

CleanShot 2022-12-06 at 12 12 32

zackkrida:
I can't find the comment but I'm still weary about one local development concern, that my VS Code prettier, which I typically enable to run on file save, doesn't work, with the following error:

```shell
["ERROR" - 12:25:44 PM] Error resolving prettier configuration for /Code/monopenverse/frontend/src/components/VAllResultsGrid/VAllResultsGrid.vue
["ERROR" - 12:25:44 PM] Cannot find module 'prettier-plugin-tailwindcss'
```

Are the experiments with pnpm workspaces you mentioned available anywhere, which might fix this issue?

dhruvkb:
Please see https://github.com/dhruvkb/monoverse for a version that includes `WordPress/openverse`. That is an improved version where the benefits of code de-duplication (especially in GH workflows) are more evident.

rfcs/20221124-monorepo.md
krysal: - What kind of bulk actions do you tend to do or would like to be able to do? I'm trying to imagine the use for mixed issues/PRs.
- I think the issue with the `gh` CLI can probably be solved with `just` commands on the `openverse` repo, which would require some configurations of course, but do you need to see often _all_ the open issues? For development one only typically need to focus on a well-defined subset.
- It's part of the GitHub UI in the sense that the link is always on the GH header, you only need to add the filters for the repos. I don't think new contributors need to look at all the issues to start either, approaching one individual level of the stack is more manageable.

zackkrida: @krysal I don't disagree with your comments. There are potential solutions and workarounds for many of the things that a monorepo offers. I don't think the benefits of a monorepo need to be things that _only a monorepo can do_, but things that a _monorepo might do a bit better_ also need to be considered. 

Regarding contributors looking at lists of issues, I don't think they need to see _all_ issues, either. But consider the ways users might look at the issues page. Speaking for myself, when I hear about a new project I'd like to contribute to I tend to:

- Click on the 'issues' tab in the GitHub UI to look at the top 5-10 issues and see what's actively being discussed, worked on, and engaged with in a project
- Sort by "most commented" to see what the biggest or most popular discussions are

I think it's fundamentally simpler for a contributor who wants to look at Openverse issues to go to one repo and click on the "issues" tab in the UI, rather than type filters or click on a link which was found externally.

stacimc:
I am still in favor of this change. I see most of the cons as either implementation issues, or concerns about performance/discoverability for folks focused on only one part of the stack. As long as it's still possible to easily spin up only the portions of the stack you want, I wouldn't consider this a con. The organizational structure described should still make it very easy to focus on the parts of the codebase you care about, and the codebase itself is not prohibitively large.

In particular, I am pretty excited about making it easier to run the entire stack locally and end-to-end. I've encountered multiple scenarios where I needed to test the flow of data through the catalog, ingestion server, and API DB. This required some tinkering, and the next time I needed to test the flow I found that something had changed in the configuration and my old notes no longer applied. While it's of course possible to get this all working without a monorepo, I'm concerned it would be more brittle and difficult to document, and I imagine particularly painful for community contributors/newcomers to the codebase.

I would also like to see documentation grouped in one place. I think this would be particularly beneficial for the catalog/ingestion server/API.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,179 @@
...
+## Reviewers
+
+- [ ] @zackkrida
+- [ ] @sarayourfriend
+
+## Rationale
+
+For a comprehensive discussion about the pros, the cons and the counterpoints to each see [discussion](https://github.com/WordPress/openverse/issues/192). This is not the purpose of this RFC.
+
+This RFC summarily lists the benefits and then, with the twin assumptions of a monorepo being ultimately beneficial and the decision to migrate being finalised in the above discussion, proceeds to go into the implementation details.
```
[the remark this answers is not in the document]
dhruvkb: Right, I am sorry for assuming that we had a unanimous agreement. I have expanded the benefits section in 6ef0168 to present a stronger argument.

rfcs/20221124-monorepo.md
dhruvkb: Hyperlinks can break when documentation is reorganised or reworded. Having references to the changed docs being in the same repo makes it easy for the PR author to update it as a part of the same PR and prevent the links from rotting or breaking. 

[Sphinx explains it better than I can](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#cross-referencing-with-sphinx).

dhruvkb:
Needless to say, as the author of the RFC and as a long-time contributor to our automations, I favour merging our repos into a monorepo. But the reasons I support it aren't based on long-term hypothetical ideas and massive improvements to our processes. Those future possibilities (like catalog to frontend E2E tests), although enabled by a monorepo, are just that: future possibilities. Exciting as they are, to make a decision solely based on them would not be wise.

But the real-world present-day impact will be in the lots and lots of small things that monorepos eliminate our workarounds for. The net positive magnitude of the monorepo is the elimination of these workarounds and the time and effect we spend maintaining, upgrading and syncing them.

We have been very productive with split repos, so not using a monorepo has not prevented us from doing anything. It won't magically unlock something new. But seeing that it can make our work more streamlined and also open us up to future possibilities, I see no reason to not give it a shot.

rfcs/20221124-monorepo.md
zackkrida: I'd still really like to see a breakdown of the tasks involved in the merge, their execution order, and a time estimate for each step. We could do that when we write an implementation plan for this RFC. Two days maximum for a code freeze isn't too disruptive but definitely a concern in the case of high priority or critical bugs we might identify, even with a bug bash and small code freeze beforehand.

rfcs/20221124-monorepo.md
stacimc: This has been a pain point for documenting our data models and ingestion/data refresh processes. IMO that's one concrete place where documentation (and the process of generating/updating that documentation) could improve in a monorepo.

stacimc:
> I said in the previous conversation that merging the API and Catalog repositories seemed like a natural starting point and probably will be actually required at some point anyway, given the plans to turn the ingestion process into an Airflow DAG, and the need to test the data flow, as you mentioned.

@krysal Totally agreed! The Catalog/API merge is definitely what I'm most excited about, although in my case I may be biased 😅 I am not personally exposed to much of the syncing/maintaining work that @dhruvkb references around the API and frontend and my own workflows likely won't change as a result of the work described here so far.

The case for merging frontend/API first as a quick win on the path to a monorepo felt [compelling enough to me](https://github.com/WordPress/openverse/pull/336#discussion_r1037461011), but I'm willing to concede that I may be evaluating the value of this work based on perceived future benefits of the _entire project_ being a monorepo.

For those who've expressed concern, I'm curious if you would feel differently if the catalog inclusion were also scoped/planned? Said another way: would this work still be worth doing on its own, without the catalog inclusion? I read @dhruvkb's comment here as _yes_:

> the reasons I support it aren't based on long-term hypothetical ideas and massive improvements to our processes [...] But the real-world present-day impact will be in the lots and lots of small things that monorepos eliminate our workarounds for

But that's something I feel less able to speak to, as I'm unlikely to be affected.

rfcs/20221124-monorepo.md
dhruvkb: Thanks! I've included the command in the RFC itself in f2a9d81.

rfcs/20221124-monorepo.md
sarayourfriend: > There are potential solutions and workarounds for many of the things that a monorepo offers

For me it isn't just this, but we should also be clear that there are trade-offs to using a monorepo, both long term and temporary. To me it's important that we agree that any benefits we think we'll get from it are worth the trade offs. That's why the conversation cannot just be "what are the benefits", it must be "what are the benefits and why are they worth making the change rather than working with what we have now". Monorepos can make a lot of meta infrastructural things more complex. The ci_cd workflow in the API repository is already complicated. A lot of actions will now run that don't need to, like Playwright for changes to the API. Playwright takes a long time to run. Is that a trade off we want? If we don't, we'll need to make a files filter, but that can only be applied to an entire workflow, so now we have to have separate workflows for different aspects of the projects, which is something we intentionally worked away from at one point. Is that something we are willing to accept?

A change like this is not just pure, uncomplicated benefits. It's not that we can do everything we're doing now but easier. There will be trade offs. Even just a bigger repository to check out can be an issue, more history to dig through, more complicated releases, etc.

sarayourfriend:
@dhruvkb How do we plan to manage releases automatically triggering deployments? Do you envision that we would have a new separate workflow that would trigger new releases of the application (including creating tags for specific applications etc)? How would the tags relate to only a specific app? Do we need GitHub tags?

Deployment isn't covered in this at all and is left as an open question but it just occurred to me that it's a significant complication and would require infrastructure work that _someone_ has to do before a monorepo is workable, including changes to our automated deployment process that would need to be at least as well documented as the current process.

dhruvkb (commented):
@sarayourfriend for the deployments I need your input, as I have already demonstrated my lack of knowledge in that area. I have the following thoughts, which I hope to refine with your comments, and then I will add to the RFC.

Our repos have different release cadences, and some repos (like `WordPress/openverse`) are doing no releases at all. So we cannot go with the traditional approach of cutting releases for all versions simultaneously. That's not an option at all for us.

**Staging:** Staging releases are deployed whenever `main` changes. We can continue to do that with the added scope of filtering by directory to only deploy staging changes for the API and the frontend when files concerning those two have been affected. Most of the current workflows for the frontend will be usable as it and we can reuse parts of them for the corresponding workflows for the API.

**Prod:** Prod is different because that currently requires us to create releases on GitHub. In a discussion you mentioned frontend releases being "significantly less dramatic" and that workflow triggers could replace releases for deployments and I think that's reasonable. The only loss I see is that GitHub releases are an automatic changelog which we will lose and possibly have to [maintain manually](https://keepachangelog.com/en/1.0.0/).

In that sense, monorepo releases, until we have a uniform and regular cadence are a downgrade from the split repos. Personally I am willing to accept that downgrade for the upsides of release code sharing between all ECS-deployed components (and other benefits) but I can totally how others might not consider the tradeoff as positive.

sarayourfriend:
Your suggestion makes sense to me, Dhruv. The only detail I'll add is that the staging deployment workflows will need to be separate workflow files so that they can take advantage of the files filter, unless there was another approach you think could be taken as easily. Separating them would make sense to me though, because it would also have the effect of creating separate and distinct deployment logs per-application and per-environment (if we had a `nuxt_staging.yml` `nuxt_production.yml`, `django_staging.yml`, `django_production.yml`, etc.). The non-ECS-ified applications would in the meantime still need a workflow to trigger docker image tagging with a new version. From where we would derive that version, I'm not sure. It's nice that the release workflow currently increments it automatically for us. Should we enter it manually, or could we add inputs in the workflow to indicate how to bump the version (whether patch, minor or major)? More on semver below though.

As for the separate changelogs, there are tools that do them automatically that I've seen other projects use. I wonder if they're able to be scoped by directory and such. Emotion, for example, uses [changesets](https://github.com/changesets/changesets) which even proclaims to have a focus on monorepos. I haven't looked at all to see if it would work cleanly with a mixed-language monorepo like ours will be or just with a JS monorepo. We might be able to "hack" it to make it work in any case though, if there aren't other tools available like it that would work better for us.

It might also tie us closer to semver. I don't know what the general opinion on the team is of whether we want to stick to it or not long term, so whether that is a pro or a con I'm not able to say.

In any case, I think your plan is perfect with these additional details:

1. Separate workflows per-app/per-environment so that we have deployment logs separated along the same lines with added bonus we can use the built-in files filter.
2. Non-ECS apps still need to have the ability to tag and publish docker images on command when we're ready to deploy them, to replace the GitHub release approach we take right now.

rfcs/20221124-monorepo.md
obulat: The period of the code freeze is a big concern for me. 

Having a bug bash before it is a really great idea. It's really difficult to plan it because we probably need a sync meeting to identify bugs, [very rough estimates] a day or 2 to document all of them, and then 1-3 weeks to fix everything, especially considering the slow pace of the PR review process. We would probably only want to fix the `high` and `critical` priority bugs.

I wonder how this process be affected by the features that we plan to work on now: navigation menu improvements and the homepage. Also, if we plan to work on the addition of analytics next, would that need to be finished before the monorepo move? Or fully postponed till after the move?

Starting the monorepo migration during meetup is great for the reasons Sara mentions. However, I'm not sure that the process itself is a great use of our time at the meetup. Someone will have to be working on the migration steps during the meetup and in the week after that.

dhruvkb:
Yes @sarayourfriend we can definitely use separate workflows; that's much clearer and easier to understand too. Any common actions like setup can be extracted to actions and re-used using inputs.

obulat (commented):
Thank you for providing a more detailed rationale, @dhruvkb.

I was first thinking of the monorepo as a `lateral` change, using Zack's description. However, after reading through the rationale again, I am excited about several points.

- The easier way of running the whole stack locally, and keep all the local settings for it up-to-date, that Staci mentions.
- No more sync. It's nice to get rid of the sync PRs, but more importantly, we will be able to test the changes in all of the code right away instead of waiting for the sync PRs' CI results (like when the json changes in openverse caused the failure of frontend CI due to different prettier settings).
- Unified issues and PRs, and milestones. 

My biggest concern is the time of the code freeze. I would like to see a really detailed plan of what needs to be done during this transition time. This should also allow us to do as much work as possible before the code freeze. Maybe some things can be done under a feature flag or some similar mechanism?

I also don't quite understand @sarayourfriend 's comment on CI. Would the CI workflows be combined? I.e., would a change that only touches the frontend also run the CI workflows for other parts of the monorepo? The frontend CI runs are already pretty long (15-20 minutes). If they become longer, I would probably be against the move to a monorepo.

sarayourfriend:
> I also don't quite understand @sarayourfriend 's comment on CI. Would the CI workflows be combined? I.e., would a change that only touches the frontend also run the CI workflows for other parts of the monorepo? The frontend CI runs are already pretty long (15-20 minutes). If they become longer, I would probably be against the move to a monorepo.

My suggestion was to split them for the reason you stated, that the frontend ones can be quite long. We want to be able to easily skip the actions that are not relevant for the changes, for example, no Playwright runs when the API changes. That is of course unless we make the change Dhruv suggested and have Playwright tests use the actual local API, in which case we'd want to run the Playwright tests for the API.

One thing, though, is that I don't think we need to worry about them getting _longer_ than the currently longest workflow, unless we do something that makes them run in series. The only thing I would be able to anticipate is that runs for the API might be longer if we make the Playwright change... I don't think that would happen for a while though. Presumably the Airflow CI wouldn't ever get longer.

If things go way too long to run for every PR (some projects have this) we could have multiple test suites. I think Firefox does this where they have tests that run for each change but a separate, much larger and more comprehensive suite of tests that run on a once-a-day basis that can take hours to complete. If we wanted to follow this the hardest part would be triaging each test and deciding whether it's worth waiting for them for each change set. I suspect something like the full-end-to-end ingestion->refresh->frontend test kind of thing would be one of these "nightly" test cases.

All of that to say: I don't think the overall workflow run would get longer and if they did, there are strategies we can take to mitigate their growth. The benefit to splitting the CI workflows per app is just that we can more easily have them run only when they absolutely should using the files filter (which is only available on a workflow level, as far as I know, not easily on a job or step level).

sarayourfriend:
@dhruvkb Can you add the infrastructure considerations to the document with explicit steps like the other aspects? This would help to make sure we didn't miss and big things in that regard.

AetherUnbound:
I think I share the same biases as @stacimc but I'm also in favor of this move for the same (long-term) reasons 🙂

zackkrida:
@dhruvkb I'd like to give this a final review as soon as possible. Could you share what fina edits, if any, you intend to make? I just want to make sure I don't do a comprehensive review of what's here prior to a large changeset.

rfcs/20221124-monorepo.md
```diff
@@ -79,13 +75,17 @@ First we will merge the API and the frontend repos into `WordPress/openverse`. T
...
 
 1. The API and frontend share identical tooling for Git hooks, linting and formatting. We will fight our tools less and encounter minimal friction.
 
-   - The frontend's approach for `pre-commit` inspired the RFC for expaning this type of usage to the API as well!
+   - The frontend's approach for `pre-commit` expanded this type of usage to the API as well!
+
+   - We're expanding the use of double-quoted strings to JavaScript to further unify our style guides.
 
-1. The entire system can be integration tested during releases. The real API, populated with test data, can even replace the Talkback server as long as we can turn off all network calls and enable 100% reliable output.
+1. The entire system can be integration tested during releases. The real API, populated with test data, can replace the Talkback server as long as we disable network calls and make output deterministic.
```
obulat, line 88: I love this! 🚀

obulat: We would have to make sure that the API response types _never_ differ from the sample data, though. For example, if we say that the license URL is always present, and it's present in the test data, but absent in the real API responses, that would break the app even though the tests pass.

dhruvkb: Next step would be integrating the catalog so the supplier of the data would also be a part of the tests 👍.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+
+1. No more sync:
+
+   Current criticism: We use complex sync workflows to keep files in sync. Some
+   workflows need to by synced to some repos only. Some workflows shouldn't even
+   be in the repo they're synced from. Some files need subtle differences so we
+   compile Jinja templates for them. We also sync GitHub labels and branch
+   management rules. It's a mess (I would know!).
+
+   **Monorepo solution:** A monorepo would just eliminates all of this and saves
```
zackkrida, line 62:
```suggestion
   **Monorepo solution:** A monorepo eliminates all of this and saves
```

small grammar nit

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+   in a monorepo. This is a limitation imposed by GitHub and there is no
+   workaround for this.
+
+The overarching theme is that there are workarounds for everything. We have been
+working with split repos quite productively for over a year. My proposition is
+the the monorepo solutions are better than workarounds.
+
+The [integration](#step-4-integration) section in the latter part of the
+document describes more interesting outcomes made possible by the monorepo. They
+may or may not be exclusive to monorepos but they're surely made easier by it.
```
zackkrida, line 101: 👍

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+     API as well!
+
+   - We're expanding the use of double-quoted strings to JavaScript to further
+     unify our style guides.
+
+1. The entire system can be integration tested during releases. The real API,
+   populated with test data, can replace the Talkback server as long as we
+   disable network calls and make output deterministic.
+
+The `WordPress/openverse` repo will absorb the `WordPress/openverse-api` and
```
zackkrida, line 151: Another great benefit to this that we've mentioned previously is that `WordPress/openverse` is a great, simple name and logical entrypoint to the code for our users.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+| .pre-commit-config.yaml | Uncomment the commented regions |
+| .gitignore              | Use existing                    |
+| prettier.config.js      | Use existing                    |
+| README.md               | Use existing                    |
+| CONTRIBUTING.md         | Use existing                    |
+
+#### Housekeeping
+
+1. Create "stack: \*" labels to help with issue and PR management.
+   Spoiler/foreshadowing: these labels will be used for more things later.
```
zackkrida, line 302: Perhaps this should happen _before_ the migration? If we created the labels in advance, we could then apply each label to _all issues_ in the respective repositories using the cli (I think this would be a very simple command to write), _before_ transferring. The labels would be preserved during the transfer.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+    -n 1 \
+    -I num \
+  gh issue transfer \
+    num \
+    WordPress/openverse \
+    -R WordPress/<repo>
+```
+
+With this done, we can archive the API and frontend repo. An optional notice may
+be added to the `README.md` files for clarification before archiving.
```
zackkrida, line 328:
```suggestion
With this done, we can archive the API and frontend repo. The following notice will
be added to the `README.md` files for clarification before archiving.
```

```md
> **Note**
>
> # This repository has moved to [WordPress/openverse](https://github.com/wordpress/openverse) as part of a monorepo.
```

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+   Milestones are confined by repository boundaries. To have milestones that
+   cover issues in different layers of our stack, the only way is for them to be
+   in a monorepo. This is a limitation imposed by GitHub and there is no
+   workaround for this.
+
+The overarching theme is that there are workarounds for everything. We have been
+working with split repos quite productively for over a year. My proposition is
+the the monorepo solutions are better than workarounds.
+
+The [integration](#step-4-integration) section in the latter part of the
```
zackkrida, line 99:
```suggestion
The [integration](#step-4-integration) section in the latter part of the
```

This appears out of date, as Step 4 now refers to a documentation merge.

rfcs/20221124-monorepo.md
```diff
@@ -0,0 +1,446 @@
...
+scenario is that we're completely back next week, the worst one takes two.
+
+Note that in the transition period nothing will break. The old repos will
+continue to exists as they are, till we ensure everything works and then we
+archive the current split repos.
+
+### Step 1: Merge with histories
+
+This should be quick save for a few merge conflicts. In case of conflict copy
+the code from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
```
zackkrida, line 199:
```suggestion
This should be quick save for a few merge conflicts. In case of conflict copy
the code from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/).
```

This instruction feels a little clear to the point of being potentially destructive. For example, if some code that was merge conflicted is waaay out of date in `monoverse`. Should this be edited to specify something like 'configuration files' instead of any code? Even that might be concerning, for example if we've updated an action version in the real repository that is behind in monoverse.

Or is monoverse meant to be kept up to date with the real repositories?

zackkrida (changes requested):
The rationale and migration path look great to me. I am really sold on the idea of a monorepo and excited for the benefits I forsee.

I still think, however, that our implementation plan needs more refinement. What is here is a great start but there is a lot of specific cleanup and steps I would like to see documented. 

For this reason, I would approve this PR as-is (with implementation details removed) if we create a separate implementation plan. Otherwise there are things I would like to see expanded on:

## 1. Infra repo considerations

In the infra repository, we have code which syncs many things across the repositories:

- The required status checks and branch protections
- The collaborators and their levels of access
- Mapping secrets across multiple repositories (which you do mention)

What should the timing of changes be there? My thought would be that we should create a PR in advance which corrects the collaborators, repos, and secrets mapping by removing the to-be-archived repos. Then we would apply the terraform changes and merge the PR right after the actual repos are migrated.

## 2. Small details

Some small thoughts that are important but I'm not sure where to put them:

- What preparation of the frontend and API repos do we need to do before the merge? No open PRs? For how long?
- We should make the PR minimum review count for each repo > 2 on the day of the monorepo merge, to block any accidental PR merges.

## 3. The timing and implementation order

I suspect we could do much of the implementation work in advance, to dramatically simplify the day-of changes necessary to implement the monorepo. The idea would be do as much work non-destructively and in advance as possible, so that our code freeze can be as short as possible. The general idea would be:

1. Create all stack labels far in advance, and sync them to all repositories. On the day of launch use a cli command to bulk apply `frontend` and `api` stack labels to _all open issues_ in each respective repository. When issues are transferred later, the labels will be preserved. We can also make the 'this repo is archived' notice PRs in advance of the actual monorepo merge too.

2. In `WordPress/openverse`, create a `monorepo shared files` PR containing all of the merged/unified files from `monopenverse` that you listed:

| Expected conflicts      | Resolution                                                                                            |
| ----------------------- | ----------------------------------------------------------------------------------------------------- |
| .github/CODEOWNERS      | Folderwise separation of owners, use from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/) |
| .prettierignore         | Merge entries from all repos, use from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/)    |
| .pre-commit-config.yaml | Python + Node.js linting, use from [@dhruvkb/monoverse](https://github.com/dhruvkb/monoverse/)        |
| justfile                | Split into many smaller `justfile`s and one root<sup>a</sup>                                          |
| .gitignore              | Split into many smaller `.gitignore`s and one root<sup>b</sup>                                        |
| CONTRIBUTING.md         | Write new, referencing doc site                                                                       |
| README.md               | Write new, referencing doc site                                                                       |

This PR would be merged on the day of the monorepo migration as a first step. Crucially here, we address and fix all 'Step 4. Documentation merge' issues prior to the monorepo merge, rather than after it has happened. 

3. In `frontend`, create a branch which moves all code into the `frontend/` directory and removes the root-level docs which would conflict with the monorepo.

On the day of the merge we make sure this branch is up to date with `main` changes and ready to go. When we combine it with the `openverse` repo we use the command you mentioned (`git pull frontend main --allow-unrelated-histories`) but with the name of this created branch, _instead of_ main.

4. The same approach is followed for the API, creating a branch with  any conflicted files removed.
5. After the root level doc and action PR is merged both histories are immediately merged in and the issue transfer commands are run. Then we apply the terraform changes from the infra PR, to the new list of combined required status checks, secrets to be synced, and so on. Finally, we archive the repos.

This general approach allows us to have all of the combined root-level files prepared in advance in one PR, and to make the repo-specific file changes made in advance on a branch before the actual monorepo merge. I believe it would reduce our "down time" to a few hours on one day. 

I would also suggest we have 1-3 people working on it synchronously like a launch to iron out issues on the day. 

Does that make sense? Is there something major I'm overlooking? I really appreciate your acknowledgement that there will definitely be things we miss, and things that might break after combining the repos. 

---

In general, I want to be really really careful and detailed about the actual implementation plan before we execute this. the best approach might be to keep this open indefinitely, after holiday afk, for further implementation discussion, rather than making two separate document. Alternatively, like I said I would support approving this PR if the plan was to make a separate implementation plan. 

I'll also be awaiting my co-reviewer, @sarayourfriend's thoughts which I will take into consideration too.

rfcs/20221124-monorepo.md
```diff
@@ -248,7 +280,29 @@ To clean up the workflows we will define three new actions. The code for all thr
 
 #### Update workflows
 
-With this done, the development on the API and frontend can continue inside their subdirectories. The development of both parts will be independent. At least until we reach [long-term consolidation](#step-5-long-term-consolidation).
+Workflows with major changes:
+
+- `ci_cd.yml` from the API will absorb `ci.yml` from the frontend
```
sarayourfriend, line 382: What approach will we take to avoid running API tests when only frontend code is changed and vice-versa? Combining the CI workflows will preclude us from using the workflow `files` filter. Did you have a different solution in mind?

rfcs/20221124-monorepo.md
```diff
@@ -248,7 +280,29 @@ To clean up the workflows we will define three new actions. The code for all thr
...
+
+With this done, the development on the API and frontend can continue inside their subdirectories. The development of both parts will be independent, at least until we reach [long-term consolidation](#step-5-long-term-consolidation).
+
+#### Deployment
+
+##### Staging
+
+The soon-to-be-ECS-based API and ECS-based frontend will continue to deploy to staging via the CI + CD pipeline, with deployment starting as soon as all CI checks have passed. They will use similar code as the frontend auto-deploy for staging used currently.
+
+These will be separate jobs with specific [path-based filters](https://github.com/dorny/paths-filter).
```
sarayourfriend: What is the benefit of using this `paths-filter` action rather than the built-in GitHub workflow `paths` settings for workflow triggers? Similar to my other question: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events

rfcs/20221124-monorepo.md
```diff
@@ -248,7 +280,29 @@ To clean up the workflows we will define three new actions. The code for all thr
...
+
+##### Production
+
+For production, we will not be able to use GitHub Releases and will have to use a manually-triggered workflow to build the assets and tag them. The tag can be set via a workflow input (simple) or can be calculated based on the update scope of major, minor or patch (not as simple).
```
sarayourfriend: Another way to simplify this is for us to switch to SHA based tags for release deployment. Rather than creating a new build, just use the build tagged with the `HEAD` SHA.

rfcs/20221124-monorepo.md
dhruvkb: Yes, that makes more sense than transferring the issues, keeping a list of them and then applying labels.

rfcs/20221124-monorepo.md
dhruvkb: This actions allows us to filter jobs instead of applying the filter to the entire workflow. This allows us to have a complete workflow where the job to deploy images to GHCR can run if all the CI test jobs have passed.

If the deploy is in a separate workflow, I don't think it's possible to run a workflow if a different workflow has passed.

rfcs/20221124-monorepo.md
dhruvkb: We can use the [`path-filter` action](https://github.com/dorny/paths-filter) to do that.

rfcs/20221124-monorepo.md
dhruvkb: The slight issue with SHA tags would be that it's harder to identify versioning with them. Given two hashes, it's not possible to say which is the newer version.

rfcs/20221124-monorepo.md
dhruvkb: Keeping `monoverse` (which was a testing ground and experimentation repo) updated witn the three repos being merged will be a huge investment of time and effort with no significant return. I've updated the instructions in 5763d16 to **not copy** the code but rather the resolution logic from the commits.

rfcs/20221124-monorepo.md
dhruvkb: I updated your comment to ensure that merging it wouldn't break linting rules.

dhruvkb:
I would prefer that the implementation plan be here and thoroughly discussed in one place rather than a partially complete RFC that's approved and merged without a concrete implementation plan. Keeping this PR open seems like a fair tradeoff for centralising all conversation about the entire undertaking.

rfcs/20221124-monorepo.md
sarayourfriend: It is possible to have cross-workflow dependencies, actually: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

I only have two concerns with using the actions:

1. Extremely repetitive and annoying code to maintain in the workflows (need to guard each frontend/api job with the same repeated step at the top of the job). Note that we could use YAML anchors to fix this _slightly_ (still more repetition here compared to `paths`) but [`actionlint` does not currently support them](https://github.com/rhysd/actionlint/issues/133). We'd either need to contribute an upstream PR or stop using `actionlint`, the latter of which I think is unwise given how helpful it is at catching syntax errors locally.
2. It still actually uses a runner and worker time to complete whereas needs bypasses it entirely, therefore not eating up any of our quota. Considering the sheer number of jobs we're going to end up having in a single repository by the end of this migration, cutting down on the number of jobs that are run is prudent in this respect. This is one reason why we added all the concurrency configuration to our workflows, for example.

If you think either of these are not a concern, then that is fine. The first should be documented as an expectation and accepted trade off. I think the second would bite us later on, but I can accept that I may be being _too_ cautious.

One more thing that crossed my mind as I was writing this is whether our concurrency settings will need to change for jobs. My hunch is _no_, but it's worth looking into to confirm as they can be quite nasty to think through.

rfcs/20221124-monorepo.md
sarayourfriend: Ah, good point!

rfcs/20221124-monorepo.md
dhruvkb: I agree that `actionlint` has a lot of value so losing it would not be advisable. Let me update some of the code in `monoverse` to use `workflow_run`. Hopefully that'll allow us to keep the jobs distributed.

dhruvkb: `workflow_run` workflows depend on the ["OR" combination](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) of multiple workflows. This is different from `needs` which depends the ["AND" combination](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) of multiple jobs.

The annoying to maintain code might be true, but it can be overcome by abstracting the `paths-filter` job into a separate action with some `inputs` (similar to the `setup-env` action), thus limiting the annoyance to a single file.

Point 2 could be an issue as it will spin up a job and cancel it if the criterion fails, an alternative to that could be to have a job that only matches the condition of the paths. Let's say jobs called "is_api", "is_frontend" etc. The other jobs can `needs` on them so that they're skipped if their code paths are untouched. I think it will be a bit of a hassle to manage the pass/fail/skip cascades though so it's something that needs a bit of a dry-run first.

zackkrida (approved):
Originally it was very important for me to have an implementation plan outlined in this PR, but as I understand @dhruvkb has completed multiple dry-runs of this migration in testing environments and is confident in the procedure to run it. Since this is also a one-off migration which will be executed by @dhruvkb and the timing of it occuring after our openverse.org launch tomorrow is ideal, I am comfortable approving this PR as-is. 

Once our [project process](https://github.com/WordPress/openverse/pull/366) is more refined, I'll feel stricter about adhering to it.

AetherUnbound (approved):
Looking forward to it!

PR #359, opened 2023-01-05
Run label check on every commit and label change

AetherUnbound:
## Description

Follow up to https://github.com/WordPress/openverse-infrastructure/pull/326 which required label checks for all merges into main, this was causing issues where a maintainer had to add & remove labels to re-trigger the check so a PR could be merged.

This action was previously set up to only run when the labels on a PR change. Since we've made this a required check, it makes sense to have this run unconditionally on every PR commit so that the check can pass (or fail) appropriately prior to a merge.

## Testing Instructions

We might be able to see on this PR!

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/pr_label_check.yml
+++ b/.github/workflows/pr_label_check.yml
@@ -4,6 +4,12 @@ name: PR label check
 on:
   pull_request:
     types:
+      # Default types
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
+      - opened
+      - reopened
+      - synchronize
+      # Label is changed
       - labeled
       - unlabeled
 


AetherUnbound:
Shoot, now I'm worried that it'll interfere with the "Label new PR" action since those run quick and it might fail if the PR is submitted without labels but gets them from that action 😭 Let me see if we can actually expand the set of events rather than revert to the default set.

stacimc (approved):
I think this makes sense! LGTM, interesting problem. Thanks for jumping on it so fast

AetherUnbound:
Okay, I think I've got a solution! This should now cover all the default PR actions (commits, opening, reopening, etc) _as well as_ when labels are updated. I'm going to let it run after this push, then try to change the labels to make sure the actions get triggered again.

AetherUnbound:
Worked like a charm! 🥳

zackkrida (approved):
Ah, very nice, thanks for the work here!

stacimc (approved):
Re-approving, good catch :)

krysal (approved):
I was about to say it doesn't make much sense to me to run this action on every commit but it seems necessary, as this is the default config the [GH action project shows](https://github.com/agilepathway/label-checker#using-the-label-checker-action). Thanks for debugging this and finding the solution!

PR #362, opened 2023-01-08
Content Moderation/Trust and Safety: Initial user stories

sarayourfriend:
## Description

This document is still in its preliminary stages and is meant as a place to collaborate on the list of user stories and derived technical and process requirements that we will need to embark on for trust and safety. Identifying the _minimal_ list of requirements for content moderation and trust and safety can come after we've gotten a general view of the full scope of the initial set of needs.

Many Openverse sponsored contributors may be able to discuss these in person next week and collaborate on expanding this document. To participate otherwise, please leave any suggested user stories or technical and process requirements or assumptions that are missing for the list of user stories.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/rfcs/trust-and-safety/20230109-trust_and_safety_overview.md
+++ b/rfcs/trust-and-safety/20230109-trust_and_safety_overview.md
@@ -1,0 +1,204 @@
+# Trust and Safety in the Openverse: preliminary overview and exploration
+
+This document seeks to explore the facets of Trust and Safety that Openverse
+needs to consider as we undertake the important work of building systems and
+tools for content moderation. The bulk of this practice is a list of user
+stories and accompanying assumptions and technical and process requirements
+needed to meet the assumptions.
+
+Note that a distinct feature of Openverse _as it exists today, in early 2023_ is
+that there is no method for anyone to upload content directly to us. Every
+single provider is manually added to the catalogue. We currently rely heavily on
+the fact that our content providers have their own content moderation policies
+and procedures. The majority of our content providers (to the best of our
+knowledge) that have user generated content automatically scan that content for
+illegal materials. The other providers are either unlikely to ever include
+illegal materials (GLAM institutions) or manually review each submission by hand
+(WordPress Photo Directory). This may not always be the case. One of the future
+goals for Openverse is for websites (and WordPress sites, in particular) to be
+able to automatically add themselves as providers of openly licensed content.
+Once that is possible, then it will become even more imperative that Openverse
+scans materials for illegal materials. Even before then, however, I think it is
+prudent for us to consider doing it earlier because running the risk of hosting
+and distributing illegal content (which is also universally heinous content) is
+not something Openverse ever wants to be involved in, even if it is only because
+one of our upstream content providers made a mistake.
+
+## Process user stories
+
+1. As an Openverse user, I expect results that are removed from upstream
+   providers due to containing illegal materials will be automatically removed
+   from Openverse so that Openverse does not accidentally continue to host
+   materials that have been removed from the provider
+1. As an Openverse content moderator, I want to be able to queue the removal of
+   a result from the index without a prior report existing so that I can skip
+   the report generation step
+1. As an Openverse content moderator, I want to be able to prevent results from
+   showing for a search so that searches do not include results that may not
+   follow Openverse's content policies
+1. As an Openverse content moderator, I can review pending media reports and
+   make documented decisions about the validity and needed action for a report
+   so that manual or automated systems know what to do with a particular
+   reported result
+1. As an Openverse content moderator or user, I expect that a result that has
+   been removed from Openverse will not re-appear even if uploaded by a
+   different user so that duplicate uploads will not cause already removed
+   results from re-appearing
+1. As an Openverse content moderator, I expect that a result marked for removal
+   will also remove any other results that may be duplicates of the removed
+   result so that I do not have to manually remove results that have similar
+   perceptual hashes
+1. As an Openverse content moderator, I want a clear definition of what is and
+   is not allowed to exist within Openverse so that I can take decisive action
+   on content reports
+1. As an Openverse developer, I want a clear definition fo what is and is not
+   allowed to exist within Openverse so that I can build tools that identify
+   results that should not be accessible in Openverse without them needing to be
+   discovered by a content moderator or user
+1. As an Openverse user, I want results with sensitive content to require an
+   explicit opt-in to view so that I am not exposed to sensitive content without
+   my consent
+1. As an Openverse user, I want to know why a result is marked as "sensitive"
+   without needing to view the result itself so that I can make an informed
+   decision about whether a sensitive (but not policy-violating) result is
+   relevant to what I am searching for
+1. As an Openverse content moderator, I want to be able to bulk remove results
+   from a given search so that a search that has mixed material (results that
+   follow and results that do not follow Openverse's content policies) can be
+   easily dealt with
+1. As an Openverse content moderator, I want to be able to remove all results
+   from a particular creator so that if a problematic creator is discovered and
+   all (or practically all, or mostly all) their content does not follow
+   Openverse's content policies, they can be easily removed
+1. As an Openverse content moderator, I want to be able to mark all results from
+   a creator as needing review so that if a problematic creator is discovered we
+   don't need to manually mark each result for review (or removal, depending on
+   the severity of the creator's issues)
+1. As an Openverse content moderator, I want to be able to upstream moderation
+   decisions to providers when appropriate so that Openverse can be a good
+   steward of the commons and also help improve the quality of providers'
+   content
+1. As an Openverse user, I expect that Openverse will never include material
+   already known to be illegal because Openverse is not a place for distribution
+   of illegal materials
+1. As an Openverse content moderator, I expect that results with sensitive
+   textual material will be automatically raised for review so that I can
+   quickly identify results that may need moderation action without a user
+   needing to file a report or a content moderator searching them out
+1. As an Openverse user, I expect that sensitive textual material has a useful
+   content warning so that I can make an informed decision about whether I want
+   to read a piece of text
+1. As an Openverse user, I expect to be able to clearly understand what content
+   on the Openverse website is from Openverse and what content is from the
+   providers so that I understand whether I need to make upstream reports of
+   creators or specific content to providers or to Openverse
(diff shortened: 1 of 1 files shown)


rfcs/trust-and-safety/20230109-trust_and_safety_overview.md
```diff
@@ -0,0 +1,266 @@
...
+and procedures. The majority of our content providers (to the best of our
+knowledge) that have user generated content automatically scan that content for
+illegal materials. The other providers are either unlikely to ever include
+illegal materials (GLAM institutions) or manually review each submission by hand
+(WordPress Photo Directory). This may not always be the case. One of the future
+goals for Openverse is for websites (and WordPress sites, in particular) to be
+able to automatically add themselves as providers of openly licensed content.
+Once that is possible, then it will become even more imperative that Openverse
+scans materials for illegal materials. Even before then, however, I think it is
+prudent for us to consider doing it earlier because running the risk of hosting
```
obulat, line 37: Tiny nit: we are not really "hosting" the media. We host the metadata and cache the thumbnails, but the media hosting itself is still done by the providers.

sarayourfriend: That's a good qualification to make, though I'm not sure how to do it without making the qualification over and over again. Do you have a handy turn of phrase that could be used here to indicate the extent to which we _do_ host some aspects of the content (textual and visual, when it comes to thumbnails)?

dhruvkb: The word "present" seems to be valid in our case. We don't "host" the content but we "surface" and "showcase" the content in our web space.

sarayourfriend: Fair enough. I still think there's an important technical distinction to draw that we have extra obligations to ensure that our independently hosted versions (cached thumbnails) of results are correctly managed, but I'll change the language to be "present" with a separate qualification elsewhere.

sarayourfriend: Changes made in https://github.com/WordPress/openverse/pull/362/commits/d63707395a9007415047bfe4a1a99efab5490ec3

rfcs/trust-and-safety/20230109-trust_and_safety_preliminary_overview_and_exploration.md
```diff
@@ -0,0 +1,285 @@
...
+
+Museum and library collections often include historical material with sensitive
+content. For example, a museum may hold, catalogue, and distribute a photograph
+with its original caption even if the caption describes the subject in an
+offensive or inaccurate way. Especially common examples of this are images of
+racialised people with original captions that use slurs or other derogatory
+language to describe the subject. I (Sara) do not know if this is something that
+currently exists in Openverse, but it is something we could discuss with
+providers if we discovered it to ensure that we are capturing the relevant
+metadata. For example, some providers may include a note clarifying that a
```
obulat, line 71: In general, I agree that we should show the provider's notes.
The implementation details can be complicated by the way the notes are displayed on the providers' sites or in their APIs and the way we ingest that data. I tried searching for some racial slurs on Openverse. One example I found is from Boston Public Library. We have ingested this item from their Flickr stream, which does not have any notices. However, the same item is also hosted on digitalcommonwealth.org, and their it has a notice banner at the top of the page: https://www.digitalcommonwealth.org/search/commonwealth:fq977w05r. I'm not sure if it's returned from the API, though.

rfcs/trust-and-safety/20230109-trust_and_safety_preliminary_overview_and_exploration.md
```diff
@@ -0,0 +1,285 @@
...
+probably already has a position on this kind of thing, and we should lean on
+pre-existing definitions and policies in that regard. For everything else,
+however, we'll need to make our own decisions about how to represent the
+material. One option is to generously blur visual content that may be offensive
+or otherwise sensitive. Additionally, adding content warnings as much as
+possible, both for visual and textual material.
+
+## Process user stories
+
+1. As an Openverse user, I expect results that are removed from upstream
```
obulat, line 110: Any ideas on _how_ we can do that? When we re-ingest the items, we add the new items, update the existing ones, and do not remove the ones that were not present for the upstream provider. I don't think there's a way to know why the item was removed from the provider.
I wonder if adding some sort of scanning process for items that are not present at the provider when re-ingesting would help. We should also consider adding an "antitboosting" parameter (is there a word for the action that is opposite of search rank boosting?) to all of such items, assuming that they were removed from the provider for some reason.

rfcs/trust-and-safety/20230109-trust_and_safety_preliminary_overview_and_exploration.md
```diff
@@ -0,0 +1,285 @@
...
+   my works being removed from Openverse
+1. As a creator, I want to be able to mark my own content with specific content
+   warnings so that I can make my content more accessible
+1. As a user, I want to be able to mark or suggest content warnings for results
+   so that I can help make Openverse catalogued content more accessible
+
+- Assumptions:
+  - Upon reingestion from a provider, the catalogue is able to note when results
+    have been removed from upstream providers and eagerly remove them from the
+    Openverse API without needing a full data refresh to occur
```
obulat, line 186: Does this mean that we want to remove all the items that were not returned from the provider API when reingesting? This might mean that we remove a lot of items if the provider changes the API or if the API errors out for any reason. However, re-checking each individual item that wasn't present during reingestion also requires time and resources.

I wonder if adding some sort of scanning process for items that are not present at the provider when re-ingesting would help. Or we could also consider adding an "antitboosting" parameter (is there a word for the action that is opposite of search rank boosting?) to all of such items, assuming that they were removed from the provider for some reason.

sarayourfriend: Oh you're right. Even from a non-technical perspective, it's a difficult question. Expanding on your list, I can think of a bunch of reasons why a work might not appear in the provider API anymore, and there's almost no way for us to distinguish between them.

1. The creator deletes their account or an individual validly licensed work. In this case, the works are still CC licensed (assuming they were correctly licensed to begin with, i.e., not stolen) and part of the commons and would ideally still be accessible. We've discussed this a couple of times in the past and mused about whether Openverse has a role in "preserving" the commons somehow (maybe by uploading things to Archive.org or something like that).
2. DMCA takedown. We shouldn't distribute these.
3. Sensitive material takedown. We shouldn't distribute these either _probably_ assuming we agree with the provider. For example, the provider could be participating in government censorship that we don't care to be a part of, do we preserve these as in the first case then, assuming they're correctly licensed CC works?
4. Illegal material takedown. We shouldn't distribute these.
5. API errors (as you mentioned).
6. API changes (also as you mentioned).

Each of these are different, and I don't think we could easily tell the difference between any of them in an automated or even manual way without heavy provider involvement.

One thing to note that I forgot about when writing this is that we will eventually stop serving those results in search because the links will be dead. They'll disappear from search after the cached success response expires (30 days from the first appearance in search). The thumbnail will continue to exist though, and I think the single result will as well.

Complicated issue. I don't have any concrete suggestions for this at the moment.

> I wonder if adding some sort of scanning process for items that are not present at the provider when re-ingesting would help. Or we could also consider adding an "antitboosting" parameter (is there a word for the action that is opposite of search rank boosting?) to all of such items, assuming that they were removed from the provider for some reason.

Boosting them downward is an interesting idea. You might be able to apply negative or fractional boosting scores to documents in ES that would cause their scores to plummet. At that point though, I wonder if we could "soft delete" them by setting a flag that would just exclude them from search? Maybe we should just play it safe and exclude them from single results as well for now until we have clearer understandings of what the alternatives and implications of those alternatives would be?

Very tricky issue!

rfcs/trust-and-safety/20230109-trust_and_safety_preliminary_overview_and_exploration.md
sarayourfriend: That's a great example of the kind of thing we should do. For sensitive textual content we could pretty reliably attach a notice of our own for items like that.

That specific one does not look unique to the item though, it seems like a generic one (like we could do just by scanning textual content for specific words). If there are GLAM providers that do have handwritten catalogue notes that are accessible to us in this vein we should include them though.

zackkrida:
Just leaving this here: it's an approach to programmatic cache clearing in CloudFlare:

https://community.cloudflare.com/t/worker-recipe-cache-purge-proxy/29978

sarayourfriend:
@WordPress/openverse What do folks want to do this with this PR? Shall I close it? Shall I merge it? Will things from the offline discussion y'all had be added to this?

AetherUnbound:
I think I'd like to see this merged even if it doesn't yet include feedback from our discussion!

zackkrida:
@sarayourfriend I'd like to undraft and merge these. They're great. We can have someone update them later after publishing the notes from the offline sync session we had on content safety.

rfcs/trust-and-safety/20230109-trust_and_safety_preliminary_overview_and_exploration.md
zackkrida: One discussed approach to this was to create `{mediaType}-removed` tables in the Catalog DB in which we insert any records removed from the main tables. We could match the schema of the main tables but also add `removed_on` and `removed_reason` columns which would indicate the date of removal and the reason for removal. As discussed, we likely can't always find the specific reason why an image was removed/became unavailable from the provider (although perhaps _some_ providers return different error codes in different situations, as an example, and we could leverage that) but this column would also be used for media _we_ remove from Openverse (for various content safety and copyright reasons).

This would allow us to create a DAG or other mechanism in the future to re-crawl the items in the `-removed` tables, either all of them or only ones with specific values in the `removed_reason` columns (for example we could only re-crawl items that 503 errored during the data refresh).

sarayourfriend: Sounds great to me.

AetherUnbound (approved):
This was so thorough and really helpful for our discussion about content safety! User stories are a great way of approaching this kind of scoping and I'll try to use this approach in the future for other projects 😄

PR #446, opened 2023-02-21
Use a hack to compile the frontend container with workspaces

dhruvkb:
## Description

This PR updates the `Dockerfile` and the `build-frontend` job to
- use the frontend locally as a part of the workspace
- build the Docker image as a non-workspace JS package

To do this, the PR employs a hack whereby a monorepo like structure is set up inside the Docker image itself by outer using files from the monorepo like `.npmrc`, `pnpm-lock.yaml` and `.pnpmfile.cjs`. Also a fake `pnpm-workspaces.yaml` file is generated that only contains on sub package, the frontend itself.

I have verified that an image generated in this way is very similar in size to the non-workspaced image. Also the root lockfile has been pinned in a way that very much prefers the package versions specified in the previous frontend-only lockfile (notably Nuxt and Vue packages) so that accidental updates are not introduced.

## Testing Instructions

Download the `frontend.zip` artifact from the CI + CD summary page. Then extract and load it into Docker as follows.

```console
$ unzip frontend.zip
$ docker load --input frontend.tar
$ docker image ls
```

You can start the Nuxt server from the downloaded image. See the frontend running properly at http://localhost:50243.

```console
$ docker run --rm -p 50243:8443 openverse-frontend:latest
```

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "version": "2.2.1",
   "scripts": {
-    "prepare": "pnpm i18n:get-translations --en-only",
+    "prepare": "pnpm i18n:get-translations --en-only && cp -r test/locales/* src/locales",
```
obulat, line 7: Is `prepare` only run for tests? If not, this will cause problems because we are copying the test locales (specifically generated for Playwright tests, Arabic, Russian, Spanish and English only).

templates/.pre-commit-config.yaml.jinja
```diff
@@ -103,8 +103,7 @@ repos:
           - --max-warnings=0
         additional_dependencies:
           {% block eslint_dependencies %}
-          - eslint@8.15.0
-          - eslint-config-prettier@8.5.0
+          - eslint@8.28.0
```
obulat, line 106: Why do we remove `eslint-config-prettier`?

frontend/package.json
dhruvkb: The `prepare` script is run everytime the package is set up (immediatey after `pnpm install`), but then running `pnpm i18n`, which we do before building the Docker images, will overwrite the test locales with the actual locales downloaded from GlotPress.

templates/.pre-commit-config.yaml.jinja
dhruvkb: It's not removed, it was listed twice in the list with different versions so I deduped that. I have updated the template in 3857de0.

.github/workflows/ci_cd.yml
```diff
@@ -832,7 +838,11 @@ jobs:
     name: Publish Docker images
     runs-on: ubuntu-latest
     # prevent running on fork PRs
-    if: (github.event_name == 'push' && github.repository == 'WordPress/openverse') || (github.event_name == 'release' && github.repository == 'WordPress/openverse')
+    if: |
+      always() &&
+      (needs.test-ing.result == 'success' || needs.test-ing.result == 'skipped') &&
```
zackkrida, line 843: Side note: It's hard not to read this as "needs testing"

zackkrida (approved):
@dhruvkb could you just explain the motivation for this? It works great, but I'm curious what problems exist with the way the repo is presently working that this solves.

dhruvkb:
@zackkrida the main motivations for going through this extra effort were

- This layout is more in line with the `pnpm` definition of a monorepo, where a single `pnpm` command from the root of the repo can set up all packages.
- Being a monorepo allows us to break out some parts of the frontend (like the components and design system) into smaller reusable pieces in true JS monorepo fashion.
- This layout effectively integrates the frontend into the package instead of it being siloed away as a disjoint separate project.
- A number of `pnpm` invocations needed the `--ignore-workspace` flag earlier, forgetting which was easy and caused confusing problems. It's not needed anymore.

All in all, this was the intended approach from the start but the issues with `pnpm deploy` not working with shamefully hoisted deps was preventing us from going ahead with it. This hack works around that and allows us to use this better approach.

obulat (approved):
I tested running dev, and then removing `node_modules` and running `just install` and `just frontend/run test`, and it worked well. Great hack!

PR #761, opened 2023-02-22
Wrap README image in an openverse.org link

zackkrida:
Makes this image a bit more useful by leveraging it as a link to Openverse.org.


obulat (approved):

dhruvkb (approved):
Weirdly the image looks bigger in the diff, but I like it!

image

zackkrida:
@dhruvkb I noticed that too, it's just an irregularity of how the diffs render. The "after" view has some margin-left applied and it's behaving oddly with the 100% image.

PR #775, opened 2023-02-25
Set rate-limit headers on each response

### LINKED ISSUE #755: Build UI for API consumers to get their key and check their usage (original #335)
This issue has been migrated from the [CC Search API repository](https://github.com/creativecommons/cccatalog-api/issues/335)

```
Author: kgodey
Date: Fri Jul 19 2019
Labels: 🏷 status: label work required,🙅 status: discontinued
```

--------
### Original Comments: 

*annatuma* commented on Thu Jan 23 2020:

>Tentatively slotted for Q3 in Roadmap. Design mockups from UCB student collaboration are a starting point here, when the time comes.
[source](https://github.com/creativecommons/cccatalog-api/issues/335#issuecomment-577438608)

dhruvkb:
## Fixes

Fixes #755

## Description

This PR
- updates audio views to use the same throttling as corresponding image views
- drops an unused throttle class
- documents the non-recoverable nature of the API keys
- adds `X-RateLimit-` headers to the responses to keep tabs on API usage
- documents the presence of these headers

## Testing Instructions

0. Check out the PR and start the services.
1. Read the documentation at `http://localhost:50280/v1/`.
2. Set `DISABLE_GLOBAL_THROTTLING` env var to `false` in `env.docker`.
3. Make unauthenticated requests and see your `X-RateLimit-Available` value go down to zero.
4. Start getting 429 responses.
5. Make authenticated requests and see your much higher `X-RateLimit-Available` value.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal (dismissed):
This does not fix #755 but I think it's a good idea! Could you add some tests for the new headers?

dhruvkb:
@krysal you're right about it not doing what the issue says but I think this is the more pragmatic, developer-friendly way to provide usage information for an API. This is what ]
GitHub does too, they don't have a web UI, just [HTTP headers](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limit-headers).

I'll add the tests.

dhruvkb:
Turns out that writing tests for this is really hard because in the dev environment where the tests are being run, the throttling behaviour is turned off by default which writes over the DRF throttling settings.

api/catalog/api/views/media_views.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs):
...
 
+    def check_throttles(self, request):
+        """
+        Set the rate-limit headers and copy ``super`` implementation for raising
+        throttling exceptions.
+        """
+
+        throttle_durations = []
+        for index, throttle in enumerate(self.get_throttles()):
+            if not throttle.allow_request(request, self):
```
sarayourfriend, line 57: Rather than copying this method so that we can add the line to modify headers, do you think we could augment `allow_request` on the throttle class to do it? It'll have access to headers via the `self` parameter and could add them before returning the regular result of `allow_request`.

We'd have to change the method for finding the suffix for the headers though, as we wouldn't have the index in there. I thought we could use the `scope` name, but the TenPerDay style throttles don't have a scope. We could either add unique scopes to them or find the next suffix from the existing headers. I think I prefer the scope approach personally as it is human-readable and would probably help debugging a question from users "why am I being throttled" this way, but I don't have a strong preference.

This is also not a blocking request for a change, but I do think it would be nice to not have to keep this method copied and have it potentially drift from the upstream implementation (however unlikely that might be) or update the comment below referencing a DRF issue number to clarify that.

api/test/unit/utils/throttle_test.py
```diff
@@ -124,3 +125,25 @@ def test_abstract_oauth2_id_rate_throttle_does_not_apply_if_token_app_rate_limit
...
+    request = request_factory.get("/")
+
+    # Send three requests. The third one should be throttled.
+    for _ in range(3):
+        response = view(request)
+
+    headers = [h for h in response.headers.items() if "X-RateLimit" in h[0]]
+    assert [
+        ("X-RateLimit-Limit-anon_burst", "2/hour"),
+        ("X-RateLimit-Available-anon_burst", "0"),
```
sarayourfriend, line 148: I think the unit test suggestion I made is slightly incomplete, in that we'd probably want to confirm that the first two requests also have the headers and that especially `Available` is the correct value each time.

sarayourfriend (changes requested):
Excellent! I love the cleaned up version even more. With respect to requiring a scope, we could also fall back to the throttle class name if the scope isn't available, rather than leaving it out. That way at least it would be consistent. That is not a blocking suggestion, however.

I left a comment on the unit test to request slightly better coverage of the available header, otherwise this PR is good-to-go from my perspective.

api/test/unit/utils/throttle_test.py
dhruvkb: I added those tests in 909cc3e.

stacimc (approved):
LGTM, tested great locally. Very clear to follow 💯

api/test/unit/utils/throttle_test.py
```diff
@@ -129,8 +130,11 @@ def test_abstract_oauth2_id_rate_throttle_does_not_apply_if_token_app_rate_limit
 
 @pytest.mark.django_db
 def test_rate_limit_headers(request_factory):
+    cache.clear()  # This is needed between multiple runs on the same computer.
```
sarayourfriend, line 134: This will destroy any local data used in redis cache which could interrupt someone doing other testing if they chose to run the entire test suite. Can we use the mock redis fixture for this instead? That way this test doesn't destroy the local environment's data.

sarayourfriend (approved):
One change request on the tests to make them more contained, but I won't block on it. LGTM otherwise.

api/test/unit/utils/throttle_test.py
dhruvkb: The `redis` fixture at the top of the file has `autouse=True`. I think it's using that fixture by default.

dhruvkb: Changed the line to only remove caches pertaining to throttling in a186341.

PR #773, opened 2023-02-24
Prevent injection in weekly report script

dhruvkb:
## Fixes

Fixes #408 by @dhruvkb

## Description

This PR
- upgrades the `last_week_tonight.js` file to MJS
  - for improvements like top-level `await`
  - to enable updating `node-fetch` to v3 which does not support CJS imports
- deletes the `weekly_report.js` file which does something similar but is unused
- upgrades `node-fetch` to v3 (see point above)
- adds HTML escaping in issue titles to prevent issues like #408 from causing injection

Also I can't bear to see Renovate struggling with #404.

## Testing Instructions

You can invoke the script locally.

```console
$ env ACCESS_TOKEN="<GitHub PAT>" MAKE_USERNAME="<Make username>" MAKE_PASSWORD="<Make app password>" just report
```

If you don't have an app password, read [this post](https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/) by [George Stephanis](https://profiles.wordpress.org/georgestephanis/).

## Checklist

- [x] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

[...]

PR #780, opened 2023-02-26
Use Axios instead of node-fetch

dhruvkb:
## Fixes

Fixes #773 by @dhruvkb 

## Description

This PR
- replaces the use of `node-fetch` in automations with Axios

  Shamefully hoisting conflicting versions of `node-fetch` can break either the frontend or the automations (depending on the hoist order). This will most likely be fixed after Nuxt 3 migration (see nuxt/nuxt#14146) but for now, not using `node-fetch` is stabler.

- updates the scope of files that trigger the frontend test-suite

  The frontend being a workspace package means that changes to the root `package.json` and `pnpm-lock.yaml` will affect it.

- removes duplicated fields from inner `package.json`

  `pnpm` will read these fields from the root `package.json` anyway. The `Dockerfile` for Playwright which was coupled to the `packageManager` field has been decoupled as well.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -6,3 +6,6 @@ frontend:
   - frontend/**
   - package.json
   - pnpm-lock.yaml
+  - pnpm-workspace.yaml
+  - .npmrc
+  - .pnpmfile.cjs
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -4,3 +4,5 @@ ingestion_server:
   - ingestion_server/**
 frontend:
   - frontend/**
+  - package.json
+  - pnpm-lock.yaml
--- a/automations/js/package.json
+++ b/automations/js/package.json
@@ -5,9 +5,9 @@
   "version": "0.0.0",
   "dependencies": {
     "@octokit/rest": "19.0.7",
+    "axios": "^0.27.0",
     "js-yaml": "^4.1.0",
     "k6": "0.0.0",
-    "node-fetch": "^3.3.0",
     "nunjucks": "^3.2.3"
   }
 }
--- a/automations/js/src/last_week_tonight.mjs
+++ b/automations/js/src/last_week_tonight.mjs
@@ -116,7 +116,7 @@ const postActivities = (activities) => {
   const token = Buffer.from(`${username}:${password}`).toString('base64')
 
   return axios.post(
-    `posts`,
+    'posts',
     {
       title: `A week in Openverse: ${startDate} - ${endDate}`,
       slug: `last-week-openverse-${startDate}-${endDate}`,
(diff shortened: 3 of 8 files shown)


frontend/bin/playwright.sh
```diff
@@ -1,10 +1,16 @@
 #! /usr/bin/env sh
 set -e
 
+package_manager() {
```
obulat, line 4: Nice deduplication!

obulat (commented):
@dhruvkb, I've noticed that Nuxt is using https://github.com/unjs/node-fetch-native as a replacement of `node-fetch` that has a backwards (replacing `node-fetch` as is) and forward (works the same as the native fetch will) compatibility. Do you think it might be better to use that instead of axios?

dhruvkb:
That would be fine too I guess. I went with `axios` because 

- using `node-fetch` v3 was breaking Nuxt
- we already use `axios` in the frontend

.github/filters.yml
```diff
@@ -4,3 +4,5 @@ ingestion_server:
   - ingestion_server/**
 frontend:
   - frontend/**
+  - package.json
+  - pnpm-lock.yaml
```
krysal, line 8: Should these files also be included?
```suggestion
frontend:
  - frontend/**
  - package.json
  - pnpm-lock.yaml
  - pnpm-workspace.yaml
  - .npmrc
  - .pnpmfile.cjs
```

automations/js/package.json
```diff
@@ -5,9 +5,9 @@
   "version": "0.0.0",
   "dependencies": {
     "@octokit/rest": "19.0.7",
+    "axios": "^0.27.0",
```
krysal, line 8: Can't automations use the last version of Axios?

dhruvkb: I stuck to the version used in the frontend to prevent conflicts in shamefully hoisted dependencies.

automations/js/src/last_week_tonight.mjs
```diff
@@ -112,27 +112,29 @@ const getRepoHtml = ({ repo, mergedPrs, closedIssues }) => {
...
 
-  return fetch(`${MAKE_SITE_API}/posts`, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-      Authorization: `Basic ${token}`,
-    },
-    body: JSON.stringify({
+  return axios.post(
+    `posts`,
```
krysal, line 119:
```suggestion
    'posts',
```

PR #397, opened 2023-02-13
Provider tally extraction script

AetherUnbound:
## Fixes

Fixes #380 by @AetherUnbound

## Description

This PR adds a provider tally extraction script which can be used to pull out the provider-related counts from Redis.

I also moved the `dead_links` and `load_testing` utilities into a common `utilities` directory for easier grouping within this repo.

I plan to take the CSV generated from production and throw it into Google Drive, I'll share the generated graphs on the issue once the analysis is done!

## Testing Instructions

See the README for instructions on how to run the script.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #786, opened 2023-02-27
Add stack label if available, make get-changes composite action

AetherUnbound:
## Description

This PR adds the "stack" category as an _optional_ label to add for the `label_pr` workflow. It also moves the `get-changes` step into a separate, composite action which can be reused.

The PR adds the auto-stack-adding action to the `ci_cd` workflow; I was initially going to add it to the `label_pr` since I'm concerned about race conditions happening between the `label_pr` action and the stack-labeling step. But it would be a bit more work to convert `PR_URL` into fields that the label action can actually use, so I held off for now. Let me know if y'all think that would be a better approach, I can take the time to set that up.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -37,8 +37,10 @@ jobs:
       ingestion_server: ${{ steps.paths-filter.outputs.ingestion_server }}
       frontend: ${{ steps.paths-filter.outputs.frontend }}
     steps:
-      - uses: actions/checkout@v3
-      - uses: ./.github/actions/get-changes
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Get changes
+        uses: ./.github/actions/get-changes
         id: paths-filter
 
   get-image-tag:
@@ -103,12 +105,20 @@ jobs:
       - get-changes
 
     steps:
-      - uses: actions/checkout@v3
-      - name: Add backend stack label
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Add API stack label
         uses: actions-ecosystem/action-add-labels@v1
-        if: ${{ needs.get-changes.outputs.api == 'true' || needs.get-changes.outputs.ingestion_server == 'true' }}
+        if: ${{ needs.get-changes.outputs.api == 'true' }}
         with:
-          labels: "🧱 stack: backend"
+          labels: "🧱 stack: api"
+
+      - name: Add ingestion server stack label
+        uses: actions-ecosystem/action-add-labels@v1
+        if: ${{ needs.get-changes.outputs.ingestion_server == 'true' }}
+        with:
+          labels: "🧱 stack: ingestion server"
 
       - name: Add frontend stack label
         uses: actions-ecosystem/action-add-labels@v1
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -102,6 +102,26 @@ jobs:
           just precommit
           just lint
 
+  add-stack-label:
+    name: Add stack label
+    runs-on: ubuntu-latest
+    needs:
+      - get-changes
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Add backend stack label
+        uses: actions-ecosystem/action-add-labels@v1
+        if: ${{ needs.get-changes.outputs.api == 'true' || needs.get-changes.outputs.ingestion_server == 'true' }}
+        with:
+          labels: "🧱 stack: backend"
+
+      - name: Add frontend stack label
+        uses: actions-ecosystem/action-add-labels@v1
+        if: ${{ needs.get-changes.outputs.frontend == 'true' }}
+        with:
+          labels: "🧱 stack: frontend"
+
   ############################
   # API and ingestion server #
   ############################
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -36,16 +36,10 @@ jobs:
...
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
-
-      - id: paths-filter
-        name: Check changed paths
-        uses: dorny/paths-filter@v2
-        with:
-          filters: ./.github/filters.yml
+      - uses: actions/checkout@v3
```
dhruvkb, line 40: nit, sorry: Jobs should have names.

```suggestion
      - name: Checkout repository
        uses: actions/checkout@v3

```

automations/python/label_pr.py
```diff
@@ -195,7 +195,11 @@ def main():
             log.info(f"Aspect label: {aspect}")
             log.info(f"Priority label: {priority}")
             log.info(f"Goal label: {goal}")
-            pr.set_labels(aspect, priority, goal)
+            labels_to_add = [aspect, priority, goal]
+            if stack := get_label_of_cat("stack", labels):
+                log.info(f"Stack label: {stack}")
+                labels_to_add.append(stack)
```
dhruvkb, line 201: This can be an issue because it will allow the labeller action to pass (only applying aspect, priority and goal labels) while causing the `check_stack_label` action to fail when the stack label is absent.

If that feels okay to do, I would recommend doing the same for the aspect, priority and goal labels so that the action applies as many labels as it can reliably infer from the issue. Currently it's all or none for these 3 categories.

.github/workflows/ci_cd.yml
```diff
@@ -102,6 +96,26 @@ jobs:
...
+    needs:
+      - get-changes
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Add backend stack label
+        uses: actions-ecosystem/action-add-labels@v1
+        if: ${{ needs.get-changes.outputs.api == 'true' || needs.get-changes.outputs.ingestion_server == 'true' }}
+        with:
+          labels: "🧱 stack: backend"
```
dhruvkb, line 111: The ingestion server and the API have different labels. "🧱 stack: api" and "🧱 stack: ingestion server". To the best of my knowledge, the "🧱 stack: backend" label was a placeholder when importing issues for both projects from the `openverse-api` repo.

.github/workflows/ci_cd.yml
```diff
@@ -102,6 +96,26 @@ jobs:
...
           just lint
 
+  add-stack-label:
+    name: Add stack label
+    runs-on: ubuntu-latest
+    needs:
+      - get-changes
+
+    steps:
+      - uses: actions/checkout@v3
```
dhruvkb, line 106: nit, sorry: Jobs should have names.

```suggestion
      - Checkout repository
        uses: actions/checkout@v3

```

dhruvkb (changes requested):
👍 I love the idea of making `get-changes` a composite action! I have one important suggestion and some nits.

.github/workflows/ci_cd.yml
obulat: Yes, I added the "backend" stack label to make the issue import easier. It would be nice to review all the "backend" issues and apply the correct stack some time in the future.

automations/python/label_pr.py
obulat: +1 to making the stack labels mandatory just like the other three. It would help us filter issues and PRs.

AetherUnbound: I think "all or none" becomes a more difficult approach the more mandatory labels we have, since it adds even more work (which could have already been accomplished in an automated way) by maintainers when something goes wrong. If it's okay with y'all, I'll modify this approach so the PR labeling adds as many labels as it can, then we can rely on the label checks to catch the ones that are missing.

.github/workflows/ci_cd.yml
AetherUnbound: I see, I think when I made this the `backend` label was the only one available 🤔 Perhaps not though, I'll change this so it references the right one!

obulat (approved):
LGTM!

automations/python/label_pr.py
dhruvkb: I agree with setting as many labels as can be inferred (along with the awaiting triage) if not all labels could be inferred.

AetherUnbound: Ahhh, that's a great call! I'll make that change as well.

AetherUnbound: Done in 68f2292ab28217a1b1d3639ce8fe1dfb129897f6

AetherUnbound:
Ah, looks like the add stack label action is failing because it's only expecting 1 stack. Seems like we could have a case where multiple stacks absolutely make sense (part of the point of the monorepo IIRC!) so I'll try to fix that too.

AetherUnbound:
Okay, I think I've got the case where there are multiple stack labels covered, along with all our standard cases. I've run the script against an example PR here and it appropriately added all the stack labels (along with the other required ones): https://github.com/WordPress/openverse/pull/397#event-8662904079

dhruvkb (approved):

PR #804, opened 2023-03-01
RFC + POC: Add Plausible for analytics

dhruvkb:
## Description

This PR sets up a Plausible instance in the Docker Compose orchestra. This will be used to record events emitted by frontend analytics.

This is an RFC advocating to use Plausible as our analytics solution and a proof-of-concept showing how it could be set up in development.

## Testing Instructions

0. Bring up the services.
1. Visit https://localhost:50288.
2. Login with the credentials 'deploy@example.com' / 'deploy'.
3. You should see a site named 'localhost'.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
From an infrastructure perspective, I appreciate the considerations you've made and think a cloud managed offering is great.

Can you spec out the pricing for the cloud offering from Plausible? Is it just a flat monthly fee?

Also, as a funny aside, I tried to visit plausible.io but of course my home router blocks it on a DNS level as an analytics provider. I'll have to make an exception for it if we do indeed end up using it. I suppose that's also a minor trade-off from an Openverse project perspective of using the cloud hosted option rather than self-hosted, uBlock and tools like PiHole (etc) would likely not block our own self-hosted version. I think this may have already been covered elsewhere (in the frontend RFC, maybe, the discussion is too big to look through) and I do not mean it to be read as a recommendation to self-host.

dhruvkb:
> I tried to visit plausible.io but of course my home router blocks it on a DNS level as an analytics provider.

I think our Nuxt SSR server [acting as a proxy](https://github.com/nuxt-community/proxy-module) and forwarding analytics calls it receives to Plausible will cover this case.

rfcs/analytics/20230307-backend_service.md
```diff
@@ -84,3 +84,28 @@ provision all the following services:
 If we move to our own instance sometime in the future, we can
 [export data](https://plausible.io/docs/export-stats) and migrate that to our
 instance.
+
+## Pricing
+
+[Plausible pricing](https://plausible.io/#pricing) is based on number of events
+/ month. Each page view and each custom event recorded counts towards the total
```
zackkrida, line 91:
```suggestion
[Plausible pricing](https://plausible.io/#pricing) is based on number of page views
/ month. Each page view and each custom event recorded counts towards the total
```

rfcs/analytics/20230307-backend_service.md
```diff
@@ -84,3 +84,28 @@ provision all the following services:
...
+quota. The annual plan is 10x the monthly price i.e. 2 months free.
+
+There is no cap on
```
zackkrida, line 94:
```suggestion
There is no cap on:
```

rfcs/analytics/20230307-backend_service.md
zackkrida: Ah, just read your next sentence and didn't get to delete my suggestion it time. Well, yes we'd quite quickly reach this then, especially considering we're already doing around 150k _page views_ a day.

zackkrida: I'll reach out to the plausible folks and see if they have any recommendations. Since we have a single site and are a non-profit there might be some wiggle room to discuss alternative pricing.

There's also the 30 day trial, which we could spin up for development purposes at some point.

dhruvkb: Screenshot 2023-03-08 at 4 48 57 PM

sarayourfriend:
> I think our Nuxt SSR server [acting as a proxy](https://github.com/nuxt-community/proxy-module) and forwarding analytics calls it receives to Plausible will cover this case.

Right, I forgot about this. It looks like we'd need to use [this API](https://plausible.io/docs/events-api) on the server side for the frontend implementation. Should we document there here or update the frontend implementation plan to reflect the decision. Plausible's API, for example, uses a specific `props` field on the request body for the custom event payload, rather than pulling from the root request body as the frontend RFC suggests. I don't know if just using an environment variable is sufficient there as each provider probably has some separate little details like this.

Additionally, does this mean we do not actually need to include Plausible's script into the app then? If so, I'm confused exactly how would page views work? Would we post those page view events ourselves? We probably already need to anyway in order to track client side navigation, right?

I appreciate that this RFC is only to recommend specifically using Plausible, but I think it would be worth us briefly revisiting the implementation details of the implementation plan RFC to cover these edge cases. The frontend RFC is missing the following details, from my perspective:

- Do we need to install the script if we're proxying the events via Nuxt (I suspect not). We probably want this reflected in this RFC as well for consistency. However, all the documentation on Plausible's website about proxying them still including serving the script, just from our own domain instead of from Plausible.
    - It's worth noting that [the example for NextJS](https://plausible.io/docs/proxy/guides/nextjs) (the closest thing to Nuxt they have on their documentation) does not proxy via the Events API, it just serves the script from our own domain. We could also use the [Nginx strategy](https://plausible.io/docs/proxy/guides/nginx).
- We need to identify the specific implementation details of the Nuxt server middleware code and how it integrates with Plausible. In particular, do we need to abstract this for some reason or could we just hard code the calls to Plausible, including the shape of the request body.
- We need to iron out how page views and navigations will be tracked.

If we use the events API directly from our own server, it looks like we don't need the script at all. But then we need additional work to make sure page views and navigations are being tracked in addition to the custom events we've defined.

If we use the script, then we do not need the server middleware, but do need to recommend a specific way of proxying the script (my preference is the Nginx method) and then slightly revise the analytics composable to call the `plausible` function rather than sending the API request.

I understand the intention behind the API endpoint, but we need to the implementation details of each approach, probably in a follow-up RFC. However, for that to happen, it'd be good to verify we do indeed want to use the API endpoint approach rather than proxying the script (which, again, the main pro is that it automatically handles page views for us).

Anyway, none of this is a blocker other than to ask: can we re-affirm whether we'll use the proxied script or the custom API endpoint and then commit to a small follow-up RFC to clarify the implementation details of whichever approach we use?

dhruvkb:
In my local setup, what I am doing is using the Plausible tracker (`vue-plausible`) integration with the API URL set to the same as the dev server's. This means that the JS integration automatically sends `pageview` events to `https://localhost:8443/api/events`.

Using the proxy module, we just reroute these incoming API calls to the actual Plausible URL, which would be `http://localhost:50288/api/events` (locally, and `https://plausible.io/api/events` in production). This gives us the benefit of both worlds:

- the tracker can collect all the information Plausible needs on the client side
- our proxy allows us to bypass restrictions like DNS and adblockers

All that is to say the following two things.

- This PR merely makes a recommendation to use Plausible and sets up a dev environment for development purposes. Holding this up on implementation details of how we would integrate Plausible with Nuxt is not right.
- The Plausible-Nuxt integration is already somewhat developed in #844 (blocked by this PR). I'll update that PR and push my local setup to reflect my text above so your concerns (which are very valid and justified, I might add) could be addressed there. 

Sorry if I come across as aggressive or defensive. I think the only factors that should hold this PR up are reasons that we think Plausible should not be used or if some other service is better suited to our purposes.

sarayourfriend:
> Sorry if I come across as aggressive or defensive. I think the only factors that should hold this PR up are reasons that we think Plausible should not be used or if some other service is better suited to our purposes.

Not at all, I didn't realise a PR was already in the works, it makes sense to just iron these details out there as you suggested. Thanks for referencing it here.

I have some thoughts based on what you've shared, but I'll put them on the PR. Overall, it looks like we're on the right track, my comments are just detail oriented questions.

rfcs/analytics/20230307-backend_service.md
```diff
@@ -0,0 +1,111 @@
...
+
+- Plausible has an NPM package that can be bundled with the app (whereas Umami
+  needs a `<script>` tag).
+- Plausible has a generally available cloud platform (whereas Umami's is in
+  beta).
+- there was some talk of it being used in other WordPress endeavours, so it
+  makes more sense to use it.
+- I've already done some experimental work with it, and it might save us the
+  effort if that existing work were to be used.
+
```
obulat, line 73: Very strong pros for Plausible!

PR #820, opened 2023-03-02
Fix `always()` logic for dependent tasks in CI/CD workflow

AetherUnbound:
## Description

I noticed that we weren't receiving any API deployment messages in Slack, even though we've merged PRs which _should_ have initiated an API deploy. After some investigation, I found that we were requiring changes made to the ingestion server prior to deploying the API:

https://github.com/WordPress/openverse/blob/a2782e8b45300be669c90b4da1abd801caa0263b/.github/workflows/ci_cd.yml#L926

Since these services are distinct, we shouldn't need changes made to the ingestion server in order to deploy the API. This PR removes that requirement.

## Testing Instructions

Merge and hope :slightly_smiling_face:

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -923,7 +923,7 @@ jobs:
   deploy-api:
     name: Deploy staging API
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && (needs.get-changes.outputs.ingestion_server == 'true' || needs.get-changes.outputs.api == 'true')
+    if: github.event_name == 'push' && needs.get-changes.outputs.api == 'true'
     needs:
       - get-image-tag
       - get-changes
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -900,7 +900,10 @@ jobs:
   deploy-frontend:
     name: Deploy staging frontend
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && needs.get-changes.outputs.frontend == 'true'
+    if: |
+      !failure() && !cancelled() &&
+      github.event_name == 'push' && needs.get-changes.outputs.frontend == 'true' &&
+      (needs.nuxt-playwright-e2e.result == 'success' && needs.nuxt-playwright-vr.result == 'success' && needs.storybook-playwright.result == 'success')
     needs:
       - get-image-tag
       - get-changes
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -926,7 +926,10 @@ jobs:
   deploy-api:
     name: Deploy staging API
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && needs.get-changes.outputs.api == 'true'
+    if: |
+      !failure() && !cancelled() &&
+      github.event_name == 'push' && needs.get-changes.outputs.api == 'true' &&
+      (needs.test-ing.result == 'success' && needs.publish-images.result == 'success')
     needs:
       - get-image-tag
       - get-changes
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -186,14 +186,14 @@ jobs:
         run: just ingestion_server/test-local
 
       - name: Upload ingestion test logs
-        if: always()
+        if: success() || failure()
         uses: actions/upload-artifact@v3
         with:
           name: ing_logs
           path: ingestion_server/test/ingestion_logs.txt
 
       - name: Print ingestion test logs
-        if: always()
+        if: success() || failure()
         run: cat ingestion_server/test/ingestion_logs.txt
 
   #######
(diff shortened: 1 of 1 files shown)


obulat (commented):
I don't understand why we should need this change. I read the old code as "If anything in the API **or** the ingestion server changes, deploy". So, it **should** deploy if there are only changes in the API. 
With the changes here, will deploy not happen when the ingestion server is changed? Don't we want that?

obulat:
I compared two runs (after the dependabot PRs were merged). [One](https://github.com/WordPress/openverse/actions/runs/4309734336) had changes to the API, and the [other](https://github.com/WordPress/openverse/actions/runs/4310292072) - to ingestion server.

The ingestion server one did deploy. I think the difference is in the "Publish Docker images": it does not run if the ingestion server tests didn't run.
This `needs` step is not run, and the API is not deployed if changes are in API only:
https://github.com/WordPress/openverse/blob/1621ed0d68daf62420f997c97745f361191a9825/.github/workflows/ci_cd.yml#L931

AetherUnbound:
> Sorry, I was comparing the wrong runs: the "Publish Docker images" only runs on the "push" event, and it does run when there are api changes: https://github.com/WordPress/openverse/actions/runs/4309530451

This run is a good example though - the ingestion server tests didn't run and so a staging deployment _of the API_ wasn't run, even though it should be. At least, that's what I suspect is happening 🙂 GitHub Actions doesn't give you much info beyond "this was skipped" so it's kind of hard to tell haha.

dhruvkb:
@obulat's understanding is correct, the deploy happens if either API or ingestion server has changed.

We will need to use the expanded form of the `if` condition which is seen in some other jobs.

dhruvkb:
Screenshot 2023-03-02 at 10 12 50 AM

What I don't get is why `deploy-api` was skipped even though all  its dependencies have passed.

AetherUnbound:
Ah okay, thank you both. I was misunderstanding the situation and the code I altered. That said, I think the particular line change I made might still apply, since we don't have ingestion server deployments set up automatically (so no need to deploy the API on ingestion server changes). What do y'all think? I'm also confused about that @dhruvkb, I'll try to look into it more today.

AetherUnbound:
Here are three example runs. All 3 affect the API, but the first 2 do not initiate API deployments. The only difference I can see from the 3rd (which _did_ have a successful run) was the fact that the ingestion server tests were not skipped in this case:
- https://github.com/WordPress/openverse/actions/runs/4309189828
- https://github.com/WordPress/openverse/actions/runs/4314531114
- https://github.com/WordPress/openverse/actions/runs/4310292072

AetherUnbound:
@dhruvkb I think this _has_ to be an interaction with the `always() &&` logic and these lines in the `publish-images` task:
https://github.com/WordPress/openverse/blob/1621ed0d68daf62420f997c97745f361191a9825/.github/workflows/ci_cd.yml#L854-L855

I suspect this gets us into some weird behavior where the job itself runs but due to its dependencies being skipped it registers as "skipped" for downstream tasks.

AetherUnbound:
Ha, this behavior is exactly described in the next comment on the original suggestion 😅 🤦🏼‍♀️ https://github.com/actions/runner/issues/491#issuecomment-882886236

AetherUnbound:
OKAY I think I got it! Per some of the discussion in that linked thread, I've changed may of our `always()` calls to something more appropriate, since `always()` will run even if the job is cancelled or fails. I also added that condition to the API deploy step so that it ignores the skipped state of its grandparents and runs when it should. Fingers crossed that this does it! 🤞🏼 🍀

.github/workflows/ci_cd.yml
```diff
@@ -923,7 +926,10 @@ jobs:
   deploy-api:
     name: Deploy staging API
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && (needs.get-changes.outputs.ingestion_server == 'true' || needs.get-changes.outputs.api == 'true')
+    if: |
+      !failure() && !cancelled() &&
+      github.event_name == 'push' && needs.get-changes.outputs.api == 'true' &&
+      (needs.test-ing.result == 'success' && needs.publish-images.result == 'success')
```
dhruvkb, line 932: A similar change could be needed for the `deploy-frontend` action above which also depends on the sometimes-skipped `nuxt-playwright-e2e`, `nuxt-playwright-vr` and `storybook-playwright` jobs.

dhruvkb (approved):
LGTM, except for a request to update the frontend `if` condition as well. I won't block it though.

It's unfortunate that the only way to check if these changes work is push and pray.

.github/workflows/ci_cd.yml
AetherUnbound: Ah, thanks for that callout!

AetherUnbound: I've taken a stab at this here: 8bb3fef7823b7512ce11690b512cdd790f8cae31 - do you think that looks right?

obulat (approved):
It's unfortunate that `always()` acts in such an nonintuitive way in this case. Thank you for sharing the steps of your investigation!
🤞

.github/workflows/ci_cd.yml
dhruvkb: That looks logically right 👍. You could break it into several lines for readability but nonetheless, it should work just fine.

PR #843, opened 2023-03-05
Add preferences for analytics

dhruvkb:
## Fixes

Fixes #821 by @dhruvkb

## Description

This PR
- adds a switch view to the checkbox based on the WordPress design language
- removes dummy example feature flags
- adds feature flag for analytics
- updates the preferences UI to match mockups from #791
- adds documentation for the feature flag system

## Checklist

- [x] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #844, opened 2023-03-06
Implement analytics in Nuxt

### LINKED ISSUE #825: Set up wrangling for events
## Problem

Whether we want to plug directly into a third party (like Plausible) or build our own analytics solution (using open-source DBs and API frameworks), we need some pipeline to wrangle the events from the frontend to a format acceptable to the service.

## Description

In the interest of avoiding vendor lock-in, we can put a thin layer between the frontend and the analytics service, which will receive events from the frontend and will appropriately wrangle and forward them to the right service.

This can be any of the following:
- a separate microservice
- an API endpoint
- a Nuxt server middleware
- a Nuxt server route

## Alternatives

Building a decoupled backend sounds great but in reality, choosing a platform early on makes the most sense as it avoids us having to build our middle layer and we can use native SDKs from the platform to directly ship data to the service.

## Additional context

RFC: https://github.com/WordPress/openverse/blob/3fa9f03937d2af356535f158f7f80b20a6bd79d2/rfcs/analytics/20221006-implementation_plan_frontend.md

dhruvkb:
## Fixes

Fixes #825 by @dhruvkb

## Description

This PR implements a framework in Nuxt for sending analytic events to Plausible. It also adds an event as an example.

## Testing Instructions

0. Check out the PR and run
   - the backend setup, `just up`
   - the frontend dev server `just frontend/run dev`
0. Open Plausible in a new tab at `localhost:50288`.
1. Visit some URLs in the dev server and see stats roll in.
2. [Set up a custom event `CLICK_HOME_GALLERY_IMAGE`](https://plausible.io/docs/custom-event-goals#3-create-a-custom-event-goal-in-your-plausible-account) in the Plausible UI.
3. Click on some homepage gallery images.
4. See custom event goals showing up in the lower part of the UI.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
Blocking this on other PRs #804 and #843.

frontend/src/composables/use-analytics.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Send a custom event to Plausible. Mandatory props are automatically merged
+   * with the event-specific props. This function is meant to be used on the
+   * server side because it uses their API instead of using the integration.
+   *
+   * @param name - the name of the event being recorded
+   * @param payload - the additional information to record about the event
+   */
+  const sendCustomEventApi = async <T extends EventName>(
```
sarayourfriend, line 107: So I could be totally wrong about this, and please let me know if I'm missing something big here. I really think we might be able to significantly simplify this implementation. I think we're doing a lot of duplicated work here by proxying the API call in this particular way. The recommended proxy approach is to have something like Nginx just forward the call from the script to plausible and let the script handle the rest of the parameters. See the example here: https://plausible.io/docs/proxy/guides/nginx

In that case, we would still be using our own domain for the events, but we don't have to handle _anything_ near the level of complexity in this function because we know that Plausible is passing the event through anyway. We would still have the flexibility to configure Nginx locally to send events to the docker-compose plausible instance rather than the production instance.

There is the possibility that some events would be fired on the server-side as well, like the `INSUFFICIENT_RESULTS`, unless we intentionally defer it to happen on the client side. If we want to defer it to happen on the client side, we need a mechanism to do that (a store? I don't know). However, I think this event could be implemented in the Django API instead and we could still track the event and even forward the information Plausible uses to calculate the user identifier from there.

Looking at the list of events from the events RFC, the only events that have this quality are the server rendered event and the insufficient results one. The rest all do come as a result of user interaction with the page and would never happen during the server rendering of a page.

If we can eliminate those events from the frontend code, then we pretty significantly simplify this stuff because `sendCustomEventApi` and `sendCustomEvent` can (a) ignore the difficulties of isomorphism, (b) people writing code that fires events don't need to choose from two different similarly named functions, (c) we eliminate the need to maintain a lot of this code and would no longer need the server middleware to forward events (we'd rely on Nginx to forward them for us instead).

Additionally, I think even the server rendered event might be redundant with the [entry pages](https://plausible.io/docs/metrics-definitions#entry-pages) metric. Any entry page (and their rates) would show is the number of server rendered pages (though requests already do that). Looking at that metrics page, I think we're duplicating a lot of stuff potentially. For example, do we need to gather screen dimensions if [Plausible already tracks it](https://plausible.io/docs/metrics-definitions#screen-size) (though it only tracks width, not height).

Here's basically the pit-fall that comes from not having considered the particular provider ahead of time. Plausible appears to have a great deal of things built into that we'd be making redundant with some of the stuff we're writing into the application here (all of which comes with added complexity). It made sense to assume that none of this stuff would be tracked when we weren't using a specific analytics provider, but now that we know exactly who we're using, we can rely on their default behaviours. I assume that if we were to switch providers it would be a significant enough deal that we'd need to do more work anyway. Trying to hand-spin this stuff on our own to not feel as locked-in might be a red herring.

Anyway, my recommendation here boils down to:

1. Use the regular Nginx approach documented by Plausible to proxy the script and the events API on our own domain.
2. Use the default Plausible window function to send custom events. If we need a wrapper to handle some details automatically then :+1:, but it can call the regular Plausible script function, and we'd avoid additional complexity.
3. Send the "insufficient results" custom event from Django.
4. Confirm that Plausible doesn't already track things for us that we're tracking. Entry page as an alias for "server rendered page" is a good example of this. Screen width as well (do we absolutely _need_ height?).

Overall, I think this would significant reduce the complexity of this work.

sarayourfriend (commented):
Some initial feedback. Forgive me if I am missing something that invalidates my recommendation. This is exciting work and I don't want to delay it. I believe my recommendations (if valid) make the PR get in faster because I think they would massively reduce the amount of code required to implement the baseline for event tracking.

frontend/src/middleware/page-views.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * This middleware sends events to Plausible.
+ *
+ * - For the initial page load, executed on the server, it sends a
+ *   `SERVER_RENDERED` event.
+ * - For subsequent page loads, executed on the client, it sends a
+ *   `VIEW_PAGE` event between every navigation.
+ *
+ * These events are in addition to the `pageview` events automatically sent by
+ * the Plausible integration.
```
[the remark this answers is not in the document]
dhruvkb: - The `SERVER_RENDERED` event was experimental (given the draft nature of this PR) to see how we could potentially send events from the server. This event is actually not needed because SSR or not, the pageview event is sent from the client for each page visited.

- The `VIEW_PAGE` event written in the RFC prior to choosing Plausible an a platform. If we were to go with a different platform (or build one ourselves), we'd have to add this `VIEW_PAGE` event but with Plausible mandating `pageview`, it's not needed.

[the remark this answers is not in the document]
dhruvkb: I intended to remove both of them before undrafting. The point of this PR being a draft was to run as many experiments as possible while other PRs are blocking it.

frontend/src/composables/use-analytics.ts
dhruvkb: This was an experiment, in public view, to see the boundaries of our event sending system, it was not supposed to go into production as-is (hence the draft PR).

In the end, the implementation of analytics is as follows.

- `vue-plausible` bundles the Plausible tracker into the code, removing the need to include via a script. This allows us to not be blocked by mechanisms like adblockers or DNS denylists because the tracker is shipped with the app code itself.

- `@nuxtjs/proxy` module is used to send all Plausible calls to the same domain as the app, so those calls will be made to `https://openverse.org/api/event`. The SSR server (Express afaik) will forward them to `https://plausible.io/api/event` (or wherever we host our Plausible instance), again bypassing adblockers and DNS denylists.

- All events will be sent from the client side, for the most part, so they will have access to `window`. The cookie code has been removed 7e06869. It was a test to see if a server, given enough info in cookies can emulate analytics sent from clients, not production ready and to be dropped before undrafting.

- Plausible tracks a few things that we don't need, like IP addresses which it uses to identify unique users. Considering that, like you said, I have not included IP in our custom mandatory props. Plausible also tracks screen width, like you said, but does not make it available to our custom goals (only using it to determine device afaik) so we must track that (and height for added measure) on our own. Plausible does not charge for data contained per event so we can make it as rich as we like.

- Server side events like `INSUFFICIENT_RESULTS` can be considered in a later phase of this project after this "MVP". We have plenty of learnings to derive from other events before we get to a point where lack of events data will prevent us from drawing conclusions.

dhruvkb:
@sarayourfriend thanks for your review. I really appreciate you taking so much effort in going through this messy PR and providing suggestions. My only concern is that you were a bit too early to review it (given it was a draft), because most of the points you mentioned were already under implementation and you could have saved yourself a lot of trouble had you reviewed it a bit "later".

That said, again thanks for reviewing it. I hope that now that I've cleaned it up, you might find yourself agreeing with a lot of the work done here. Please review it again? Thanks!

sarayourfriend:
> My only concern is that you were a bit too early to review it (given it was a draft), because most of the points you mentioned were already under implementation and you could have saved yourself a lot of trouble had you reviewed it a bit "later". Most issues you raised were points of experimentation that would have never made it to the final PR.

Gotcha, sorry about that. I misunderstood what we discussed in the Plausible RFC issue and thought you'd wanted me to share feedback here, folding in what I'd pointed out in the RFC. Apologies for moving too quickly.

The [summarised changes you shared here](https://github.com/WordPress/openverse/pull/844#discussion_r1131337598) look great, I'll review the code today, but I'm sure it's great.

One potential improvement is that we _can_ use Nginx for the redirect, and it could be better performance wise to not tie up the event loop with those redirects, even if they're fast. Just something to note if we notice any degradation in performance, considering these requests happen for every new client page load.

frontend/nuxt.config.ts
```diff
@@ -328,6 +332,22 @@ const config: NuxtConfig = {
       },
     },
   },
+  proxy: {
+    "/api/event": "http://localhost:50288", // the key is appended as the path
```
sarayourfriend, line 336: Should this proxy configuration be changed based on `isProd` to point to plausible? 

```suggestion
    "/api/event": isProd ? "https://plausible.io" : "http://localhost:50288", // the key is appended as the path
```

sarayourfriend (commented):
Looks great other than the question regarding the proxy configuration :slightly_smiling_face:

frontend/nuxt.config.ts
dhruvkb: The prod link is skipped here because, like you said, it's better to proxy at the NGINX level instead of at the app level. I assume if we added an NGINX level proxy, the `isProd` would be irrelevant?

dhruvkb:
@sarayourfriend I am really thankful and appreciative that you reviewed it as quickly and as thoroughly as you did.

frontend/nuxt.config.ts
sarayourfriend: I'm so sorry about this Dhruv, but I had the frontend infrastructure configuration incorrect in my head :slightly_frowning_face: I had written a comment saying that you were correct, that if we're going to use Nginx to proxy, then we can remove the express proxy handler in production and only use it for localhost. However, after looking at the frontend Dockerfile and the infrastructure configuration, I realised that I had conflated the API and the frontend setup. The frontend does not use a reverse proxy currently, so we'll need to proxy using the express plugin for now indeed.

In that case, we do need the change here to point it to plausible.io in production so that when the script requests `openverse.org/api/events` it gets forwarded to `plausible.io/api/events`.

One outlying question: Are we going to have plausible running in staging? If so then we might need to configure the script domains using an environment variable so that we can include the correct subdomain per deployment stage. If not then this isn't an issue, but we probably need to have another environment variable to disable it in staging.

As for Nginx, I think there is an issue _somewhere_ to add it to the frontend, in particular to free up Nuxt from needing to serve static files. I can't find the issue right now, but I'll keep searching. If I cannot find it, I'll write up the issue and mention that in addition to serving the static files, it should also include the plausible proxy and the removal of the express middleware in production from the Nuxt app.

dhruvkb: Added the option to set the proxy using env var in 70f0dcd.

dhruvkb:
> One outlying question: Are we going to have plausible running in staging? If so then we might need to configure the script domains using an environment variable so that we can include the correct subdomain per deployment stage. If not then this isn't an issue, but we probably need to have another environment variable to disable it in staging.
> - @sarayourfriend 

> i don’t think we want analytics actively running on staging
> - @zackkrida 

Plausible does not prevent us from sending events to it even if the site does not exist. So we could technically let the staging server keep sending info to Plausible that would be discarded by them.

When we do need to track events, we can register the staging site on our Plausible cloud account, do the tracking we need to do, and then delete the site (and associated data) thus stopping tracking again. Also the feature flag for analytics would be switched off in staging during which no events will be sent to Plausible.

```bash
$ curl --location 'https://plausible.io/api/event' \
--header 'Content-Type: application/json' \
--data '{
  "n": "pageview",
  "u": "https://staging.openverse.org/",
  "d": "staging.openverse.org",
  "r": "https://staging.openverse.org/",
  "w": 999,
  "h": 2
}'

ok

$ curl --location 'https://plausible.io/api/event' \
--header 'Content-Type: application/json' \
--data '{
  "n": "CUSTOM_EVENT",
  "u": "https://staging.openverse.org/",
  "d": "staging.openverse.org",
  "r": "https://staging.openverse.org/",
  "p": "{\"key\":\"value\"}",
  "w": 999,
  "h": 2
}'

ok
```

sarayourfriend (approved):
I've tested this locally, and it looks good to me. Everything works, following the testing instructions. Plausible's UI is simple but easy to use.

I noticed ~two~ three things as I was playing around with Plausible and the custom events. None of these are blocking issues and should be addressed in follow-up issues or clarified in documentation whether they're expected behaviour.

1. The "entry" pages appears to be tied to a specific user/session. If a person reloads a page, it doesn't think of it as a new "entry" and so does not fully capture every server rendered page.
2. The width and height props on custom events are going to be interesting to use and I wonder if there are other ways of encoding the data, perhaps as an additional prop (breakpoint, for width, like we do with the UI state cookie?). It currently gives a separate bar graph for every single instance of every single width, so at a glance, at least from the Plausible UI, it's basically useless. We can transform this ourselves once we're analysing the data in our own pipelines, but if we're doing anything "at a glance" with these, we'll want to consider encoding ranges as a separate prop to make it easier for a human to understand.
image
3. To test things on localhost, you _must_ visit with "localhost" as the domain. You cannot do what I tried to do (for far too long before I realised what was happening) and visit the site on another device on your local network at the LAN IP. We'd need to set up some kind of expected hosts file configuration for being able to test on multiple devices like `openverse.local` so that other devices can have the correct referrer on the request to Plausible. This is worth clarifying in the documentation (along with any other limitations we might be aware of in testing) and opening an issue to explore ways of fixing it that won't intrude on the average use case.

Anyway, like I said, none of these are blockers or should be addressed in this PR. This PR is good to go from my perspective and I'm exciting to start getting to use this!

dhruvkb:
Good point, @obulat! I'll add tests in a separate PR, which will focus on the usage of this utility to record all the custom events.

Issue #791, opened 2023-02-28
Safe content browsing flow


panchovm:
## Problem
The current browsing experience hides content marked as mature to allow a safe experience when displaying results. However, and as defined in project #377, we decided to allow visitors to opt-in browsing and seeing results marked as sensitive.

### Issues related
Project #377

- User stories stated as “As an Openverse user”: #362
- Blurring single results: #505
- Sensitive content filter: [#2153](https://github.com/WordPress/openverse-frontend/pull/2153)

### PR
- #824 

## Proposal
Before diving into details, here is a flow for the search results and landing in a single result in both `xl` and `xs` breakpoints.

#### `xl` breakpoint
https://user-images.githubusercontent.com/895819/221883944-a9111e27-d220-40fe-8e35-d4d57b51208f.mp4

#### `xs` breakpoint
https://user-images.githubusercontent.com/895819/221884009-31f84839-9fdb-45aa-8dd4-86db06677db4.mp4

#### `xl` breakpoint
https://user-images.githubusercontent.com/895819/221884291-2690270d-b1d2-46a5-ae22-58976a828188.mp4

#### `xs` breakpoint
https://user-images.githubusercontent.com/895819/221884311-4c28b45b-89dd-4b0c-97e0-d3ef35062913.mp4

### Opt-in settings in footer
Even when the safe browsing settings do narrow the results, placing the toggles in the filter sidebar felt without much context during the prototype tests. Because of that, and since browsing mode sets a site-wise experience, putting the component out of the filter section made more sense.

This design proposes updating the footer again. I acknowledge the efforts of simplifying this component, but this design goes further in simplicity by reducing its height and making the WordPress mention more simple. The idea is open to discussion so please leave your thoughts.

On small devices, the popover behaves as a modal. Same as other modals but revamping it with updated styles.

Having this component in the footer would also allow us to add other opt-in settings like privacy and, in the future, replace the area with a user settings one once we start exploring the collection feature. Note the exploration below.

<img src="https://user-images.githubusercontent.com/895819/221887769-f2e6aa99-4cb5-4007-867f-bdaad51dddcf.png" width="600">

### Blurring page
As noted in the user stories (#362), the single result page blurs all content area with white at 80% plus a noise texture as it keeps the header and footer clean. The CTA area is fixed in the center, and the Go to search button takes you to the homepage. I kept the page height to avoid hiding the footer as some visitors might want to click on any of those links or change the site language.

[...]

PR #862, opened 2023-03-08
Add feature flag for fake marking results as sensitive

### LINKED ISSUE #854: Add non-production feature flag for marking half of results as sensitive
For the sensitive content flow we are designing in #791 we need a way to test sensitive results. For two reasons, I'd like to create a feature flag which marks 50% of results as sensitive:

- Folks testing this functionality should not have to see *actual* sensitive content if they do not want to.
- Because marking results as sensitive is a largely manual process, we do not actually have many results marked as sensitive.

By creating this feature flag we can mark non-sensitive results as sensitive in order to test these features. I propose when this flag is active, we manipulate the API results to set every other result to have a true value in their `mature` field. 

This flag will also need to set `mature=true` to API requests.

This flag should be `disabled` in production, and switchable everywhere else.

dhruvkb:
## Fixes

Fixes #854 by @zackkrida 

## Description

This PR adds a feature flag `fake_sensitive` to mark roughly half the results as mature. It uses a random (but deterministic) approach which allows results to be marked mature consistently and across Pinia stores and browsing sessions.

This PR makes no UI changes. It is limited in scope to just adding the feature flag and setting up its behaviour. UI changes are already necessitated by #791 so this PR does not put in effort that's destined to be repeated.

## Testing Instructions

0. Enable the flag 'fake_sensitive' from `/preferences`.
0. Search for somethine.
   - You will see the console logs printing `Fake mature <uuid4>` for ~50% of results.
   - This list should remain the same no matter how many times you refresh the page.
0. Pick any ID out of this list and see the single result page for that ID. 
   - You should get the `Fake mature <uuid4>` log again because it also works on the single media store.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/types/media.ts
+++ b/frontend/src/types/media.ts
@@ -44,6 +44,8 @@ export interface Media {
 
   tags: Tag[]
   fields_matched?: string[]
+
+  mature: boolean
 }
 
 export interface ImageDetail extends Media {
--- a/frontend/feat/feature-flags.json
+++ b/frontend/feat/feature-flags.json
@@ -1,5 +1,13 @@
 {
   "features": {
+    "fake_sensitive": {
+      "status": {
+        "staging": "switchable",
+        "production": "disabled"
+      },
+      "description": "Mark 50% of results as mature to test content safety.",
+      "defaultState": "off"
+    },
     "external_sources": {
       "status": {
         "staging": "switchable",
--- a/frontend/src/stores/media/index.ts
+++ b/frontend/src/stores/media/index.ts
@@ -2,7 +2,7 @@ import { defineStore } from "pinia"
 
 import axios from "axios"
 
-import { log, warn } from "~/utils/console"
+import { warn } from "~/utils/console"
 import { hash, rand as prng } from "~/utils/prng"
 import prepareSearchQueryParams from "~/utils/prepare-search-query-params"
 import type { DetailFromMediaType, Media } from "~/types/media"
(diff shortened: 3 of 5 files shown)


zackkrida (approved):
This approach works perfectly. Awesome. Do you think it's worth creating a little utility function for the mature detection so the logic doesn't need to be updated in two places? Something like:

```
const fakeMature = (item) => {
  if (prng(hash(item.id))() < 0.5) return
  item.mature = true
  log("Fake mature", item.id)
}
```

I don't think it's necessary but something to consider.

I also tried this with some blurring, which was a fun way to test:

CleanShot 2023-03-08 at 08 31 45

dhruvkb:
Good point, I'll extract it to a separate utility 👍

frontend/src/types/media.ts
```diff
@@ -44,6 +44,8 @@ export interface Media {
 
   tags: Tag[]
   fields_matched?: string[]
+
+  mature: boolean
```
obulat, line 48: Wow, I didn't know we didn't have this in the type! Great catch! Is it always available, or should we mark it as optional?

obulat (approved):
Really nice! I have a question about `mature` type inline, but don't want to block the PR.

frontend/src/types/media.ts
dhruvkb: It's always available.

Issue #377, opened 2023-02-06
Filter and blur sensitive results by term matching


zackkrida:
| Start Date | Project Lead    | Actual Ship Date |
| ---------- | --------------- | ---------------- |
| 2023-02-06 | @sarayourfriend | TBD              |

## Description

This project matches Openverse images against a list of sensitive terms. All sensitive single image results on the frontend are blurred and viewing them is opt-in for all users.

Development work is ongoing and tracked in a milestone: https://github.com/WordPress/openverse/milestone/12

## Documents

- [x] [Trust and Safety in the Openverse: Preliminary User Stories and Exploration](https://docs.openverse.org/projects/proposals/trust_and_safety/20230109-trust_and_safety_preliminary_overview_and_exploration.html) - A general doc relevant to all content safety projects
- [x] [Project Proposal](https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html)
- [Implementation Plans (link to requested plans)](https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#required-implementation-plans)
  - [x] https://github.com/WordPress/openverse/pull/911
    - This wasn't noted as a requested IP in the project proposal as it was already being worked on by Zack, but it is part of this project.
  - [x] https://github.com/WordPress/openverse/issues/937
    - [Milestone here](https://github.com/WordPress/openverse/milestone/5)
  - [x] https://github.com/WordPress/openverse/issues/938
- [x] [Design Proposal](https://github.com/WordPress/openverse/issues/791)

## Issues

- [x] #854 
- [x] #1169 
- [x] API Milestone: https://github.com/WordPress/openverse/milestone/5
- [x] Frontend Milestone: https://github.com/WordPress/openverse/milestone/12

## Prior Art

- https://github.com/WordPress/openverse-api/pull/1108 | A potential full-fledged implementation for the API
- https://github.com/WordPress/openverse/issues/750
- https://github.com/WordPress/openverse/issues/721

zackkrida:
## Update 2023-03-07



Frontend design work is taking place here. This work assumes that the MVP of this functionality will allow the user to show sensitive results, and to elect to blur sensitive results in the search results view (with blurring as the default value). 

[...]

PR #873, opened 2023-03-09
Project Proposal: Detecting, filtering, and blurring results that include sensitive terms

sarayourfriend:
Project proposal for #377 

## Due date

Review for this proposal should aim to be completed by 24 March 2023, roughly two weeks from when this PR was undrafted.

## Notes

I've added a `project-proposal.md` template file based on the approach I took to assist future project proposal writers in getting started. The sections are based on the requirements of projects proposals described in the project process documentation.

Reviewers: Please make note of the two outstanding questions. I've made recommendations for each, but I would like particular consideration from y'all on these two points, whether you agree or disagree with my recommendations.

I pinged @krysal and @AetherUnbound semi-randomly as the reviewers, but if y'all aren't able to review this proposal over the next two weeks, please let me know and I'll find other folks to do so.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
+# Project Proposal: Detecting sensitive textual content
+
+## Reviewers
+
+- [ ] @stacimc or @aetherunbound
+- [ ] @krysal
+
+## Project summary
+
+Detect results with sensitive textual content by checking their textual content
```
obulat, line 10: Would adding the level where detection takes place be useful here? "Detect in the API"?

sarayourfriend: I tried to avoid implying any specific implementation details in this summary and to just cover the broad strokes of what the feature should entail that would otherwise be duplicated in the Requirements section below.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+Right now, the recommendation is to store the "include sensitive results"
+setting in session storage and the "do not blur sensitive results" setting in
+the ephemeral application state. "Include sensitive settings" should not be
+reset by a page reload, but "do not blur sensitive results" should. To
+summarise:
+
+| Setting                       | Default | Persistence location            | Behaviour implications                                                                                                                                              |
+| ----------------------------- | ------- | ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Include sensitive results     | Off     | Session storage                 | Cannot be manipulated through the query params. Does not reset on page reload or in new tabs. Does reset when the browser is restarted or when the session expires. |
```
obulat, line 139: Thank you for explaining the reasons behind not wanting to use the query param, it makes total sense.

By "Session storage", do you mean a session that is stored on the server, or `Window.sessionStorage`?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+   burdensome. The simplicity of the CSS filter makes BlurHash look monstrous,
+   but in the grand scheme of things, it's actually a relatively simple approach
+   to blurring the images.
+2. It avoids performance issues with relying on low-spec clients to apply
+   Gaussian blurs to potentially several images.
+3. BlurHash hashes have significant future value and all potential API level
+   performance issues can be ameliorated through caching hashes with Redis.
+
+The only significant downside that I think exists is the delay between
+unblurring and displayin7.29.1g the full image due to needing to download the
```
obulat, line 293:
```suggestion
unblurring and displaying the full image due to needing to download the
```

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+For reviewers: **What are you feelings on this issue and why?**. One alternative
+to going straight for BlurHash, if we wanted to get blurring out the door
+quickly is to treat BlurHash as a future "nice to have" and use CSS blurring in
+the meantime. This would allow us to get a fully working feature slightly more
+quickly. I'm not totally convinced it'd be _that_ much faster though. Like I
+said above, BlurHash looks complicated, but really only in comparison to how
+simple the CSS approach is. Anything would look complicated when compared to
+just adding a CSS filter style to something.
+
+##### Important detail7.29.1
```
obulat, line 307:
```suggestion
##### Important detail
```

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+unblurred.
+
+For reviewers: **What are you feelings on this issue and why?**. One alternative
+to going straight for BlurHash, if we wanted to get blurring out the door
+quickly is to treat BlurHash as a future "nice to have" and use CSS blurring in
+the meantime. This would allow us to get a fully working feature slightly more
+quickly. I'm not totally convinced it'd be _that_ much faster though. Like I
+said above, BlurHash looks complicated, but really only in comparison to how
+simple the CSS approach is. Anything would look complicated when compared to
+just adding a CSS filter style to something.
```
obulat, line 305: What if we had a single generic blurred image instead of trying to blur the one that needs it? This way, the user would not be able to discern what's behind the blur, but is it really necessary?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+should still work, but it should just be an alias for
+`include_sensitive_results` and should be marked as deprecated in or altogether
+excluded from the API documentation.
+
+##### Implementation details
+
+This will likely be implemented as a secondary index of results that do not
+include the sensitive terms, as explored in
+https://github.com/WordPress/openverse-api/pull/1108.
+
```
obulat, line 51: Would determining the list of "sensitive terms" be a part of this project?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+  therefore "safe").
+
+There may also be other viable ways of performing this determination in
+Elasticsearch, but the detail remains. If the determination is made in Python,
+then it is easy to share the implementation between the search endpoint and the
+single result endpoint (and both can benefit from the Redis caching). If the
+determination is made in Elasticsearch and benefits from Elasticsearch's full
+text search, then we may need to change the single result endpoint to pull the
+result in question from Elasticsearch rather than Postgres, leveraging the
+`_source` property on hits. This could be a big, complex change, though and
```
obulat, line 92: I am not sure we have access to the `_source` property on hits. @dhruvkb, could you confirm?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I think you're asking about the technical implementation of a session-stored variable for this (apologies if I am misunderstanding). My understanding is that it will need to be a cookie for the technical implementation to work as the server will need to know on SSR requests whether to send the `include_sensitive_results` parameter.

Here is the MDN documentation about this particular cookie lifetime setting: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_the_lifetime_of_a_cookie

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: Yes! I should add that as another one of the requested documents aside from the implementation plans. I believe @zackkrida is already sort of working on this: https://github.com/WordPress/openverse/issues/377#issuecomment-1458653924

I suspect that was delayed due to Zack being sick, or I just haven't seen it yet today :slightly_smiling_face:

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: Elasticsearch DSL does indeed include `_source` on the Hit data: https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/response/hit.py#L24-L27

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: > This way, the user would not be able to discern what's behind the blur

I think it's desirable to be able to sort of "guess" at what's behind the blur. For example, if you notice a lot of red, maybe the image includes a large amount of blood or something you'd rather avoid seeing, even though some sensitive results might be of interest for your query.

On the other hand, there are issues with this assumption as well, as in the case of this Mastodon issue: https://github.com/mastodon/mastodon/issues/10729

Our context might be different from Mastodon, though we do include Flickr, which is social media, and may have the same issues the Mastodon users are describing in that issue.

If we used a single generic blurred image, how would we choose it? We decided against using anything associated with Openverse (like the logo) to avoid implicitly associating Openverse with sensitive content. Are there other reasonable options that would potentially mislead people? Maybe a blank box or some other solid colour would be appropriate :thinking:

docs/projects/templates/project-proposal.md
zackkrida, line 1: This template looks great as-is. No changes.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
zackkrida: Yep, I'm working on this today still! Expecting to publish in an hour or two.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+## Reviewers
+
+- [ ] @aetherunbound
+- [ ] @krysal
+
+## Project summary
+
+Detect results with sensitive textual content by checking their textual content
+against a list of sensitive terms. Once results with sensitive terms in their
+textual content are determined, blur those results on the frontend.
```
zackkrida, line 12:
```suggestion
Detect results with sensitive textual content by checking their textual content
against a list of sensitive terms. Once results with sensitive terms in their
textual content are determined, blur those results on the frontend.
```

I think it might also be important to note that we _are_ making an inference about image contents based on the textual content matching.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+The Django API is able to filter out results that include sensitive terms in
+their textual content. It does so without degrading search performance.
+
+Filtering of results with sensitive terms in their textual content occurs when
+`include_sensitive_results=True`. This parameter will supplant the `mature=True`
+as a more comprehensively and descriptively named parameter. `mature=True`
+should still work, but it should just be an alias for
+`include_sensitive_results` and should be marked as deprecated in or altogether
+excluded from the API documentation.
```
zackkrida, line 44: 👍 Great idea.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+the API results in a new result field named `sensitivity`. Sensitivity should be
+an array that should be empty for results where `mature = false` and the textual
+content does not contain any sensitive terms. For results marked as `mature`,
+the array should include a `"mature"` string. For results with sensitive terms
+in their textual content, the array should include a `"sensitive_text"` string.
+At this time we will not be denoting _what_ sensitive terms are present nor will
+we be developing any categories of terms. This approach leaves open the
+potential for that in the future, however, by allowing the array to include
+specific tokens for particular categories of sensitivity, whether derived from a
+categorisation of the sensitive terms list, or from an image classification API.
```
zackkrida, line 64: Does "token" here mean to some reference to a sensitivity category, like the id of a row in a `sensitivity_categories` table, for example? I just want to make sure there isn't anything else happening here I wasn't understanding.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+Note that this designation needs to happen both in the search endpoint and in
+the single result endpoint. Each endpoint may need a different approach to
+achieve this. In particular, the approach taken will depend on whether
+Elasticsearch is used to determine whether a result needs to be designated as
+having sensitive textual content or if this is done in pure Python.
```
zackkrida, line 70:
```suggestion
achieve this. In particular, the approach taken will depend on whether
Elasticsearch or Python is used to mark results as sensitive.
```

Just a writing nit, no comments on the subject matter here.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+##### Implementation details
+
+There are two broad approaches that can be taken for this. I am actively
+consulting with people more familiar with Elasticsearch for the best way to do
+this, but the broad strokes of this are that we will either:
+
+- Loop over results in Python and using a Regex to determine if textual content
```
zackkrida, line 78:
```suggestion
- Loop over results in Python and use a Regex to determine if textual content
```

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+Sensitive results on the search results page are blurred. Users can unblur and
+reblur individual results through a simple and clear interaction on the search
+results page. In addition to being able to opt-in to having sensitive results
+included in their query, users can disable blurring by default. This parameter
+is stored in the application state. If users disable blurring by default, there
+is not an option to "reblur" images.
+
+Results unblurred by the user on the search results page are also unblurred if
+they visit the single result page for that result. Results that have not been
+unblurred remain blurred upon navigating to the single results page.
```
zackkrida, line 121: This specificity is appreciated, very nice.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+If a user lands directly on a single result page for a result with sensitive
+textual content, the result is always be blurred by default. This is true
```
zackkrida, line 124:
```suggestion
textual content, the result is always blurred by default. This is true
```

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+| ----------------------------- | ------- | ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Include sensitive results     | Off     | Session storage                 | Cannot be manipulated through the query params. Does not reset on page reload or in new tabs. Does reset when the browser is restarted or when the session expires. |
+| Do not blur sensitive results | Off     | Application state (Pinia store) | Cannot be manipulated through the query params. Resets on page reloads and in new tabs and is independent of the browser session.                                   |
+
+Compared to Google and DuckDuckGo image searches' behaviour, ours would be more
+or less equivalent (with additional safety from blurring). Neither Google nor
+DDG, as far as I can tell, store the safe search setting in the query
+parameters. Instead, they store it as a session variable. Neither have an option
+to blur sensitive results, so there is no comparison to be made along those
+lines.
```
zackkrida, line 147: I appreciate the comparison to other search engines.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+- Can have a performance impact on the thumbnail endpoint
+- Introduces a delay between when the image is unblurred when it is displayed
+  due to needing to download the actual thumbnail
+- Could have different performance issues if decoding the hash is a heavy
+  operation for low-spec clients
+  - Looking at the code, I highly doubt this because it's a
+    [fast bit of maths](https://github.com/mad-gooze/fast-blurhash/blob/main/index.js)
+    that (due to the nature of BlurHash) ignores almost all the content of the
+    image rather than a Gaussian blur which needs to consider ever aspect of the
+    image in order to apply the blur
```
zackkrida, line 260: I don't have any comparisons, but it does require each image render to a canvas element, and I don't know the performance implications of having 20 canvas nodes and some (performant, small) decoding JS executing vs. many evocations of a css filter.

I did find an alternative solution that generates images but _also_ then expects you to apply a css blur on top of the generated placeholders. I'm not advocating for this solution but I found its README comparison of different approaches an interesting, but limited, read:

https://github.com/transitive-bullshit/lqip-modern

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+3. BlurHash hashes have significant future value and all potential API level
+   performance issues can be ameliorated through caching hashes with Redis.
+
+The only significant downside that I think exists is the delay between
+unblurring and displaying the full image due to needing to download the
+full unblurred image. However, I think this is balanced out by the bytes saved
+over the network in _not_ sending images that might potentially never be
+unblurred.
+
+For reviewers: **What are you feelings on this issue and why?**. One alternative
```
zackkrida, line 298: I 100% agree the future benefits of blurhash and love providing the hashes themselves _as a feature_ to our users. It could be _extremely_ useful for API consumers and also dramatically increases the safety of Openverse _integrations_ using the API. So I am fully convinced blurhash does _more_ to contribute to safety in Openverse as a project beyond the frontend.

That said, the blur CSS is _so trivial_ to implement and allows for rapid, iterative adjustment based on user feedback. If we hear that lots of users think it's _too_ blurry or insufficiently blurry, we could adapt it. Because it's literally a one-line thing to implement, I would advocate for **initializing this project with CSS blurring** and creating a new content safety project to implement blurhash in the future. 

This is just one opinion, and I fully admit that it's a painful one for me to make. but I do also fundamentally think separating these things would allow us to ship meaningful safety improvements faster, without limiting our ability to implement blurhash later or having wasted time on a temporary solution (since the CSS is so trivial).

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+- `SENSITIVE_RESULTS_TOGGLED`
+  - Sent each time the "include sensitive results" setting is toggled. Include a
+    prop noting if the setting was being toggled on or off.
+- `SENSITIVE_RESULT_UNBLURRED`
+  - Sent each time a result is unblurred. This should include a property
+    denoting whether it on the single result or search results page.
+- `SENSITIVE_RESULT_REBLURRED`
+  - Sent each time a result is reblurred. This should include a property
+    denoting whether it on the single result or search results page.
```
zackkrida, line 331: Would we include the media ID with these?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+quickly. I'm not totally convinced it'd be _that_ much faster though. Like I
+said above, BlurHash looks complicated, but really only in comparison to how
+simple the CSS approach is. Anything would look complicated when compared to
+just adding a CSS filter style to something.
+
+##### Important detail
+
+The success of this project rests on the completion of its technical
+implementation. However, we can add the following events to the API and frontend
+to better understand the usage of the features.
```
zackkrida, line 311: Could the motivation / questions these events might answer be included? For example, I am not sure if `SENSITIVE_RESULT_BLURRED`, for example, is intended to evaluate how much the feature is used or to identify specific media items users _really don't want to see_ for, perhaps, removal from Openverse.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Infrastructure
+
+There are no infrastructure changes anticipated for this project.
+
+## Marketing
+
+We can coordinate with marketing to describe the new feature and its motivations
+and celebrate the increased safety for users, especially young people, using
+Openverse.
```
zackkrida, line 353: I would also emphasize the history Openverse has with educators and students specifically. This change would be huge for them and a great way to reach out to and reestablish ties with those communities. @cogdog ([alan levine](https://www.oeglobal.org/about-us/people-of-oeg/staff/alan-levine/)) is a contributor who comes to mind as someone to help with this outreach. He already does lots of advocating for Openverse.

zackkrida (approved):
This is excellent. I left some small suggestions and opinions on the subjects you asked about. I am happy to approve this. Do you think this document needs to make a final decision about blurhash vs. CSS? If so I can rescind my approval and wait for more discussion on that to occur.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
zackkrida: Done, here: https://github.com/WordPress/openverse/pull/911

AetherUnbound:
I am in the process of reviewing this and will finish my initial questions tomorrow morning!

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: Oh, I'll also note that the issue is slightly different from Mastodon in that on Mastodon images are often marked with author-composed "content warnings". We won't have those, at least not for the time being (until we implement some hypothetical categorisation of sensitive terms, for example. In our case, the blurred image would be an even more valuable signal for someone deciding whether a result might be relevant to their query.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: It's intentionally non-specific. It could be a string token literal (`sensitive_terms::violence`) or an obfuscated ID if we felt that was reasonable/necessary. In either case it's a constant that a client (our frontend or others) can use to display additional information about the result or even ignore it completely.

However, those would all be future ideas. My goal here was to avoid recommending us use a new boolean property (or overloading the `mature` property) as that boolean would become quickly obsolete if we expanded the tokenised reasons list and started using it for display information.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I'll add a note to this respect.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: Interesting, I like that approach. The examples are helpful for understanding what the unblurred images end up being. They're almost entirely unintelligible, especially for complex images.

> I don't know the performance implications of having 20 canvas nodes and some (performant, small) decoding JS executing vs. many evocations of a css filter.

I could measure this. My personal phone is on the far lower end of "mid-spec" and my personal laptop runs a single-core ARM process that only _just_ manages to render web video.

Maybe we could add a request for an implementation plan up front that requests:

1. Test each solution's performance and figure out which one actually performs best on low-spec hardware
2. Make a recommendation for which one to follow

The lqip approach doesn't look like it would be more work than BlurHash. If it has a significant performance advantage for low-spec hardware over BlurHash, that would make it a better candidate, even though it doesn't necessarily have the same long-term usability as BlurHash for other parts of Openverse.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I agree with this. I'll change the proposal to make this the recommendation. That allows us to push off deciding between BlurHash or some other solution like LQIP.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: Can do. Coming in the next commit.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I added a comment in the next commit.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I can't think of how we would use it now but considering it is free to include props on custom events, I don't see why not.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
obulat: Thank you for clarification! I was thrown off by the name resemblance to `Window.sessionStorage`, which is reset between tabs and is more ephemeral than `localStorage`. I see that you weren't describing that.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
zackkrida: That makes perfect sense, thanks.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,438 @@
...
+
+This option has the following downsides:
+
+- It is a basic Gaussian blur and may not sufficiently obscure the image if a
+  low number is used or may create an unsettling blob effect if a higher number
+  is used
+- Gaussian blur may not be as aesthetically pleasing as the alternative
+- It requires the image to be downloaded to be blurred by the client
+
+Initial I speculated that client-side Gaussian blurring may have adverse effects
```
zackkrida, line 235:
```suggestion
Initially I speculated that client-side Gaussian blurring may have adverse effects
```

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
cogdog: FWIW Pixabay places a black overlay with text "Adult Content" (more or less like a Mastodon content warning) that toggles off when a user selects option to remove safe search filters. No, it does not provide a hint to the image.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,438 @@
...
+
+- Significantly more complicated than Gaussian blur
+- Requires conditionally calculating the hash on the API side in the thumbnail
+  endpoint
+  - This would be done by using a new query parameter on the thumbnail endpoint
+    to request the blurred hash
+- Can have a performance impact on the thumbnail endpoint
+- Introduces a delay between when the image is unblurred when it is displayed
+  due to needing to download the actual thumbnail
+- Could have different performance issues if decoding the hash is a heavy
```
AetherUnbound, line 269: Much of this section discusses the thumbnail endpoint - do we use thumbnails for the single result page or do we load the whole image? If it's the latter, since we're not serving the images themselves, would we need to switch to using the thumbnail for sensitive results?

AetherUnbound (commented):
Thank you for the proposal and template! I've included a few questions, as well as answers to your outstanding questions from my own perspective 😄

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
AetherUnbound: Per @zackkrida: It appears that we hotlink the upstream image in the single result case. I imagine that would impact how we handle navigating to the single result page for sensitive content and what the blurring process is for that, is that correct?

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,438 @@
...
+https://github.com/WordPress/openverse-api/pull/1108.
+
+#### Designation of results with sensitive terms
+
+Results that include sensitive terms in their textual content are so denoted in
+the API results in a new result field named `sensitivity`. Sensitivity should be
+an array that should be empty for results where `mature = false` and the textual
+content does not contain any sensitive terms. For results marked as `mature`,
+the array should include a `"mature"` string. For results with sensitive terms
+in their textual content, the array should include a `"sensitive_text"` string.
```
krysal, line 67: This is speaking of the Response serializers, which sounds great to me. However, do you know how the manually reported content will fit into this process? For example, will sensitive terms be included as tags? Do we need to add more columns? Will the `api_mature<media>` tables may be deprecated?

krysal (commented):
This is a great proposal, it would be incredible to (re)add this feature to Openverse! I agree with most of what is written. My outstanding questions are related to the API implementation, which I know it's in planning still but I'd like to discuss it at a high level. When I talked to Zack about this initiative, I thought of it more as a work on the Catalog.

For what I read, the classification of the sensible content will happen at query time, either in Python or ES. I can see this as a first step to expedite the detection of the sensible media, given that presumably, it will take us a significant time to classify all our existing content. Now, this makes me wonder:
- will these discoveries be saved to the database? I can see having this info stored useful for analytical purposes
- is it a temporary solution (the API part) until we implement a sort of classifier DAG or add the checks to the MediaStores classes?

I'm intrigued by the ES solution since I saw your proof of concept PR, I hadn't thought of it before. I'm worried it may put too much burden on the cluster but I'm eager to see it working for us.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I should change it to "Session cookie" to avoid the confusion, thank you!

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: That's an interesting alternative, thanks cogdog. We could do something similar with "Sensitive Content" for the text if folks felt it was a better approach.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,379 @@
...
+
+### API
+
+#### Filtering
+
+The Django API is able to filter out results that include sensitive terms in
+their textual content. It does so without degrading search performance.
+
+Filtering of results with sensitive terms in their textual content occurs when
+`include_sensitive_results=True`. This parameter will supplant the `mature=True`
```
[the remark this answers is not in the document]
sarayourfriend: Oops, I goofed on the phrasing of sentence. You're right, it should read the opposite, that the _inclusion_ of results with sensitive terms in their textual content occurs when that param is `True`. Thank you!

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: You're right, there's some nuance that needs to be teased out here.

The frontend does use the thumbnail on the single results page if navigating to the result from the search page. This allows us to display the result (albeit at lower quality) until we have the full image ready to display. It prevents some weird pop-in behaviours from happening. In that case, we're already using the thumbnail and so we could delay downloading the full image until it is unblurred.

In the case where someone lands directly on the single results page (without navigating from the search page), we do not use the thumbnail. We'd need to have the frontend decide to use the frontend instead of the full image initially for sensitive results and then only load the full image once it's been "unblurred".

This is mostly moot with the CSS filter approach though, as in that case we can always just apply the filter client side, regardless of where the image is coming from. We would only need to consider these nuances if we went with an API-side blurring approach, BlurHash, LQIP, etc.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
```diff
@@ -0,0 +1,438 @@
...
+For reviewers: **What are you feelings on this issue and why?**. Do you think it
+is worth investing time up front in BlurHash, LQIP, or another solution; or is
+it fine to push that off to a later date as a potential future improvement to
+the feature?
+
+## Measurements and success criteria
+
+The success of this project rests on the completion of its technical
+implementation. However, we can add the following events to the API and frontend
+to better understand the usage of the features.
```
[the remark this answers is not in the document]
sarayourfriend: Yes, definitely, I will make this change :+1:

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
sarayourfriend: I hadn't considered the mature reports and had to look at the code to learn how they worked again, specifically, how the information gets incorporated into search. Forgive me if you already know this, but it was new to me and to understand my answer this is necessary context, so for my sake and for others who might not be familiar I am including it.

When the mature report is created, the [`.save` method updates the document in ES](https://github.com/wordpress/openverse/blob/7c92ba5bdff84b2771dc25a946bb371600906e0e/api/catalog/api/models/media.py#L362) to set `mature=True` and forces a refresh of the index.

There are a couple details here:

1. This will automatically work with the proposed solution, the reports just get folded into the rest of the documents already labelled as "mature". These documents would have the "mature" string in the sensitivity array.
2. I believe this is slightly broken in that this information is not preserved between data refreshes. As far as I can tell, there is no mechanism to re-apply the setting of `mature=true` for results that have confirmed reports during data refresh. Once a new index is created, the mature setting on those documents will go back to `false`. I tried digging for something that would prevent this, but I can't find it. If you take a look at the code and confirm my suspicion, I'll open an issue for us to investigate how to fix this. I suspect we would need to use a similar backport strategy as the others in discussion to get the confirmed reports from API to catalogue. We could also use Elasticsearch ingestion pipelines to flip the mature flag on documents as they're being ingested based on the mature report tables in the API database. Yet another alternative is to iterate through the mature report table after indexing and set the flag on the documents in a batch update. All are probably complex solutions to implement with their own pitfalls that would need to be considered.

So that being said, if the mature content reporting process didn't have that data refresh issue, I think it would work perfectly without modification with this proposal. If we fix that issue, then it will work with the proposal, but there's nothing this proposal needs to do to handle that bug. It is a separate bug and doesn't impede the sensitive terms detection discussed here. We would keep the mature tables around, provided we don't decide a different way of handling them is necessary in the course of fixing the bug described.

> will sensitive terms be included as tags? Do we need to add more columns?

Do you mean tags on the Elasticsearch documents? I don't believe it's necessary and would make search more complicated as we'd have to somehow prevent those sensitive terms from affecting the overall score of the document. Each tag would be adding an instance of the term on the document's overall textual content.

As for adding new columns, I don't believe it's necessary. Right now we're not classifying the particular terms into categories, so the only token that would exist in the sensitivity array is `sensitive_text`.

I guess it is worth clarifying explicitly that documents labelled "mature" are separate entirely (at least conceptually) from those we find have potentially sensitive textual content.

sarayourfriend:
Thanks for the feedback, Krystle. I think it's fine to discuss the broad strokes of the API implementation to make sure we've got the right idea in mind as far as how everything will wire up, and especially to make sure we're planning to implement things in the right places. As you mentioned, there is an alternative approach that could see this implemented partially at the catalogue level.

> For what I read, the classification of the sensible content will happen at query time, either in Python or ES.

This is mostly true, but I want to clarify that for the initial filtering, it will always happen in ES, as a separate re-indexing step, during data refresh. At least, that's the solution that I found previously that avoids us needing to include the exclusion of sensitive terms into every single query that doesn't want them (the majority of queries).

That being said, there's the secondary issue of how to determine which results in a set that we know might include sensitive results, are the sensitive ones. That's the part that is still up in the air and could happen in either ES or Python. I hope to know more about which approach we should take after getting answers back from folks more expert in Elasticsearch than me. I have one viable solution for doing it in Elasticsearch that I think would work very well and have little to no performance impact for us. The viability of that solution still needs to be verified with folks who know ES better than me.

> it will take us a significant time to classify all our existing content

This is also my assumption. If we had the recrawler working and ready to go, we might be able to implement the classification work there and have it part of the data before it is ingested into Elasticsearch. That would allow us to avoid the re-indexing with every data refresh. However, I don't believe we yet know how long the recrawling process will take. If there's more up to date information about this, then it's worth looking into if the process takes a reasonable amount of time to run.

> will these discoveries be saved to the database? I can see having this info stored useful for analytical purposes

It would be great to have this data saved somewhere. Both for analytical purposes but also to save us needing to do the reindex into the filtered index at every data refresh (if we're able to save it in the catalogue). The query to find results that match the sensitive content would not theoretically be a significant burden on our cluster, but I should test this during a downtime with the production cluster to make sure. If that's the case, then we can at least easily query for all documents that include sensitive content and then do whatever we want with those results, within the limitations of Elasticsearch's pagination.

More comprehensive analysis that wouldn't be limited by ES's pagination would require saving the data elsewhere, hopefully backported to the catalogue or managed by the recrawler.

> is it a temporary solution (the API part) until we implement a sort of classifier DAG or add the checks to the MediaStores classes?

Definitely! If there are any ways for us to get this information before documents are indexed, that is 100% preferable to this solution. As above, my understanding is that this isn't viable at the moment. I'm happy to explicitly label the querying approach a stop-gap until we're able to implement something less ad-hoc in the future.

Thanks again for the review and feedback!

docs/projects/templates/project-proposal.md
```diff
@@ -0,0 +1,40 @@
...
+<!-- How do we measure the success of the project? How do we know our ideas worked? -->
+
+## Participants and stakeholders
+
+<!-- Who is working on the project and who are the external stakeholders, if any? -->
+
+## Infrastructure
+
+<!-- What infrastructural considerations need to be made for this project? If there are none, say so explicitly rather than deleting the section. -->
+
```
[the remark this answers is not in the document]
sarayourfriend: Added! Excellent suggestion.

sarayourfriend:
Thanks @krysal and @AetherUnbound for the initial reviews :pray: It sounds like everyone is on the same page as far as implementing blurring first with the CSS filter.

I wanted to highlight for both of y'all, as reviewers, this discussion that @obulat started: https://github.com/WordPress/openverse/pull/873#discussion_r1136315785

Do either of y'all have concerns over specifically blurring over other obfuscation methods like using a placeholder or hiding the image behind a solid colour with explanatory text? If so, do you feel it's necessary to hash out those particulars in this proposal or is it something you'd be comfortable handling as a separate proposal to be written and approved before the frontend implementation plan is written?

cc @panchovm as I believe the designs would need to keep this in mind as well. Have you got any opinion on whether we need to more carefully consider other obfuscation methods?

panchovm:
The proposal is very clear. Thanks for putting all the info and testing the possible blur implementations.

I agree with how Mastodon [approached the feature](https://github.com/mastodon/mastodon/issues/10729), but for this project, I have no concerns over the solution we are taking. Regarding the blur effect, I am ok with any path we take as all three solutions are in line with the design idea I have discussed.

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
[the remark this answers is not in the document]
AetherUnbound: Phew, thanks for correcting that! I was worried that something was flying very far above my head 😅

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
AetherUnbound: Ah, okay that's a great point! I didn't realize that we could punt on these nuances with the CSS blur approach. That's even further argument for it at this time 🙂

rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md
AetherUnbound: Great call about considering the data refresh pieces of this! I looked at the code, and it appears that when building the index we **do already consider** deleted and mature records:

https://github.com/WordPress/openverse/blob/0f90f84d4bf856297356a918e4cc4fe42d07ffc2/ingestion_server/ingestion_server/queries.py#L28-L38

That function is called in the update & reindex steps, and I believe the `mature` value returned there is applied to the index. So if the record exists in `api_mature{media}`, it will have `mature=True` in the resultant index:

https://github.com/WordPress/openverse/blob/0f90f84d4bf856297356a918e4cc4fe42d07ffc2/ingestion_server/ingestion_server/indexer_worker.py#L58-L60

I was unsure about this myself because my initial assumption mirrored your own Sara, I'm glad that @dhruvkb ensured that we handled this appropriately during the refresh 😄

AetherUnbound (approved):
Similarly I (personally) think that the blurring approach is more appealing than using a placeholder image. As others states it gives variety on our search results page and gives the user _some_ indication of what the content is. The placeholder would not give any context, and until we develop more detailed ways to provide user with context before they decide to view an image marked as sensitive media, this would mean that images anywhere along the gradient of "sensitive" could be behind every placeholder. I'd personally rather have a sense of what the image looks like before viewing it, rather than have no clue whatsoever (especially since we also don't have the user-generated content warnings that Mastodon has).

Thanks for the clarifications on everything Sara, I think this proposal looks great! We can refine the Elasticsearch implementation details based on what we hear back from the Elasticsearch experts.

krysal (approved):
Thanks, Sara! Everything is clear with your thoughtful answers, so I feel confident approving this proposal ✅ 

Edit: About the blurring vs other obfuscation methods, I slightly prefer the blurring for the same reasons that Madison and Zack mention, nothing more to add.

PR #882, opened 2023-03-09
Only run stack label addition step on pull requests

AetherUnbound:
## Description

Small fix/follow up to #786, this prevents the "Add stack label" action from running on anything that isn't a pull request. See a failed run which occurred on the "push" action here: https://github.com/WordPress/openverse/actions/runs/4369574126/jobs/7643514186

## Testing Instructions

Merge :slightly_smiling_face:

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -95,7 +95,7 @@ jobs:
 
   add-stack-label:
     name: Add stack label
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && github.repository == 'WordPress/openverse'
     runs-on: ubuntu-latest
     needs:
       - get-changes
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -95,6 +95,7 @@ jobs:
 
   add-stack-label:
     name: Add stack label
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     needs:
       - get-changes


obulat (approved):
Let's test it 😄

dhruvkb (commented):
Also add the condition to not run it on PRs from forks because it fails.

dhruvkb:
I have to ask why the job is a part of the `ci_cd.yml` workflow? Wouldn't the `label_new_pr.yml` workflow, which calls the `label_pr.yml` workflow, which runs the `label_pr.py` script apply the label to the PRs automatically?

zackkrida:
@dhruvkb I think that question is best answered by @AetherUnbound and we should probably get clarification on that before moving forward here.

AetherUnbound:
I added this note in the description of #786:

> The PR adds the auto-stack-adding action to the `ci_cd` workflow; I was initially going to add it to the `label_pr` since I'm concerned about race conditions happening between the `label_pr` action and the stack-labeling step.

I'm not sure that assumption/issue still holds, and in fact we no longer need to call the `label_new_pr` workflow using a remote dispatch since they coexist in the same repo (@dhruvkb has helpfully made #897 for covering this effort). Let me take another look at this and see if those race conditions are a real concern. If anything we can merge this (plus the suggested fork fix) for now and rework this down the road so at least merges to main no longer fail on this step.

dhruvkb (approved):
Thanks @AetherUnbound for looking into it again. I should've been more careful reviewing #786. I'll approve and unblock this for now because it's a positive change nonetheless.

AetherUnbound:
Okay I had some time to look into this today and I'm not sure how we could change this without making it overly complicated. The "add stack label" step currently in `ci_cd.yml` relies on `get-changes`, which is specific to the monorepo. The `label_pr.yml` workflow which is dispatched from all of our repos is agnostic to which repo it's working on. Thus it would be harder for us to add the stack labeling to the label PR step because `get-changes` wouldn't always be available. We _could_ allow it as a parameter to the script similar to `PR_URL`, but then we'd need to add logic to the script to be able to process that input and add the necessary label. So I think keeping this in `ci_cd.yml` is easiest _for now_, since this is a monorepo-specific action and the `ci_cd.yml` workflow is monorepo-specific.

PR #883, opened 2023-03-09
Deploy to staging automatically on release

AetherUnbound:
## Description

This is a small change to the CI/CD workflow to automatically initiate a deployment to staging when a release is cut.

## Testing Instructions

N/A

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -924,7 +924,8 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       !failure() && !cancelled() &&
-      github.event_name == 'push' && needs.get-changes.outputs.frontend == 'true' &&
+      (github.event_name == 'push' || github.event_name == 'release') &&
+      needs.get-changes.outputs.frontend == 'true' &&
       (needs.nuxt-playwright-e2e.result == 'success' && needs.nuxt-playwright-vr.result == 'success' && needs.storybook-playwright.result == 'success')
     needs:
       - get-image-tag
@@ -953,7 +954,8 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       !failure() && !cancelled() &&
-      github.event_name == 'push' && needs.get-changes.outputs.api == 'true' &&
+      (github.event_name == 'push' || github.event_name == 'release') &&
+      needs.get-changes.outputs.api == 'true' &&
       (needs.test-ing.result == 'success' && needs.publish-images.result == 'success')
     needs:
       - get-image-tag


dhruvkb:
Will we be cutting releases in the future? I'm not opposed to the idea but I thought that was no longer possible due to the out-of-sync release cadence of the projects inside the monorepo.

AetherUnbound:
> Will we be cutting releases in the future? I'm not opposed to the idea but I thought that was no longer possible due to the out-of-sync release cadence of the projects inside the monorepo.

We will! We'll just be able to selectively deploy a release against either the prod frontend or API (or both) at our own discretion 🙂

zackkrida:
I'm not necessarily opposed to this, but I'm not understanding what it is for. No rush but could you explain the motivation somewhere @AetherUnbound?

AetherUnbound:
> I'm not necessarily opposed to this, but I'm not understanding what it is for. No rush but could you explain the motivation somewhere @AetherUnbound?

I believe the previous behavior for the frontend was that, when a release was cut, the frontend would automatically be deployed at that tag. This PR maintains that behavior and also applies it to the API as well. The alternative would be for us to manually deploy a release tag to staging come on which seems unnecessarily manual IMO.

dhruvkb:
@AetherUnbound the main issue with releases is that it will tag the entire codebase at a particular commit (every layer of the stack).

Apologies if I'm misunderstanding, but it seems that with this change, we will be deploying every layer on every release, in sync, with the same version number, even if nothing has changed or something is incomplete?

AetherUnbound:
> Apologies if I'm misunderstanding, but it seems that with this change, we will be deploying every layer on every release, in sync, with the same version number, even if nothing has changed or something is incomplete?

I think a point of clarification would be that we will be deploying every layer _**to staging**_ on every release, in sync, with the same version number, even if nothing has changed. Even if something is incomplete, we already deploy both the frontend and API to staging on every push to `main` (depending on if the service code is affected) so whatever code is out there already would merely be re-deployed. This simply prevents us from having to initiate both the staging and production deployments manually during the release process.

dhruvkb:
Just noting that some of this is ideologically conflicting with #895 which tries to skip as many steps as possible (including image building and publishing) if files haven't changed.

AetherUnbound:
Ahh, that's an interesting point. I think it makes sense to build & push images for all services when a release is made, though now that you mention the job skipping I'm not even sure that'd happen with our workflows the way they're currently set up 😅 I guess it depends on how `get-changes` computes changes for a given release; i.e. does it compute the changes made since the last release? Since this still requires that the images be pushed for a tag in order to actually run the deployment for that tag on a service, I don't think it's necessarily in conflict (although that _would_ mean that deployments would only happen to staging on a release _if there were changes since the last release_, again assuming that's how `get-changes` is computed for releases).

AetherUnbound:
We discussed this publicly here and will be going with an alternative approach: https://wordpress.slack.com/archives/C02012JB00N/p1678808559132249

Issue #413, opened 2023-02-16
Add option to sort search results by created_on


obulat:
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2022-09-07 | 2023-02-15 | @dhruvkb  | 2023-02-15  |

## Description

This project adds the ability to sort media items on the basis of the `created_on` field (which corresponds to the date when the item was first indexed by Openverse). This is an experimental feature, so the query parameters are prefixed with `unstable_`. 

## Documents

- [x] Implementation PR https://github.com/WordPress/openverse-api/pull/916

## Issues

Note: This Project thread issue is added retroactively to record the yearly progress in the Project board

PR #891, opened 2023-03-10
Add ability to boost search results by authority

obulat:
## Description

This PR adds 2 search query parameters to the API to allow us to test boosting the results by their `authority` score. To make sure that no one relies on this test parameter, the query parameter `unstable__authority`.

By default, the boost is the same as the `popularity` boost: 10 000. You can also fine-tune the `authority` boost by adding a float value as `unstable__authority_boost`. This number will be multiplied with the default boost. So, if you want the authority boost to be lower than the `standard_popularity` boost, you can set it to a value lower than 1 (e.g., 0.5)

## Testing Instructions

Run the API and go to [http://localhost:50280/v1/images/?q=cat&unstable__authority=true&unstable__authority_boost=1.0](http://localhost:50280/v1/images/?q=cat&unstable__authority=true&unstable__authority_boost=1.0). Locally, you should see that the first 3 results are from Stocksnap (authority score 85) and not from Flickr (authority score 75)

Go to [http://localhost:50280/v1/images/?q=cat&unstable__authority=true&unstable__authority_boost=0.5](http://localhost:50280/v1/images/?q=cat&unstable__authority=true&unstable__authority_boost=0.5) - and you should see much more mixed results.

If you search for cat with `unstable__authority_boost=0.0`, you would probably only see Flickr results on the first page.
## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -376,10 +376,7 @@ def search(
 
     if settings.USE_RANK_FEATURES:
         feature_boost = {"standardized_popularity": DEFAULT_BOOST}
-        if (
-            "unstable__authority" in search_params.data
-            and search_params.data["unstable__authority"]
-        ):
+        if search_params.data["unstable__authority"]:
             feature_boost["authority_boost"] = (
                 search_params.data["unstable__authority_boost"] * DEFAULT_BOOST
             )
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -33,6 +33,7 @@ URL = "url"
 PROVIDER = "provider"
 DEEP_PAGINATION_ERROR = "Deep pagination is not allowed."
 QUERY_SPECIAL_CHARACTER_ERROR = "Unescaped special characters are not allowed."
+DEFAULT_BOOST = 10000
 
 
 class RankFeature(Query):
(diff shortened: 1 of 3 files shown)


zackkrida (dismissed):
This is so interesting to play with, thanks for doing this 😄 

Like @dhruvkb did for #413 we might want to put this behind authentication when we ship it. 

Great idea to make the boost configurable though; we can check various searches and thresholds to try and get the right value for production dialed in.

stacimc (approved):
This is so cool ✨ I love the option to adjust the boost coefficient as well.

This looks good! I'm seeing Stocksnap over-represented in local search results for all searches even _before_ applying the boost, which makes it hard to see 'dramatic' results when testing, but I definitely confirmed the boost is being applied via logs.

For example, if I search 'food' with no boost applied, the first result is Flickr and the next 19 are Stocksnap. The order does not change at all with a boost of `0.5` or even `1.0`, but if I bump the boost all the way up to the maximum `10.0` the Flickr result gets bumped down to the tenth place in the results. I guess that's just an extremely popular result 😆 I'm excited to play around with this with way more data!

api/catalog/api/controllers/search_controller.py
```diff
@@ -374,7 +375,15 @@ def search(
             s = s.query("simple_query_string", fields=["tags.name"], query=tags)
 
     if settings.USE_RANK_FEATURES:
-        feature_boost = {"standardized_popularity": 10000}
+        feature_boost = {"standardized_popularity": DEFAULT_BOOST}
+        if (
+            "unstable__authority" in search_params.data
+            and search_params.data["unstable__authority"]
+        ):
```
dhruvkb, line 382: `unstable__authority` will always be present in `search_params.data` because it has a default value. Just the second half of the condition should suffice.

```suggestion
        if search_params.data["unstable__authority"]:
```

dhruvkb (approved):
LGTM, except minor suggestion. Approving to unblock.

PR #898, opened 2023-03-11
 Use upstream thumbnail if available

### LINKED ISSUE #675: Use `thumbnail_url` for thumbnail generation when present
## Problem

Some providers (like SMK) may link to quite large images in their `image_url`. This can cause unacceptably slow response times or even timeouts when generating thumbnails via our thumbnail service.

## Description

We already have a `thumbnail_url` available on the Image model which we can use.

When the `thumbnail_url` is available, we should send this to the thumbnail service instead of the `image_url`. This should be a small change [here](https://href.li/?https://github.com/WordPress/openverse-api/blob/f0c44d571249a2631eef5a13ea0d56702210daf3/api/catalog/api/views/image_views.py#L109):

```
image_url = image.thumbnail or image.url
```
_(using the serialized field names)_

## Additional context

https://github.com/WordPress/openverse-catalog/issues/698 tracks updating the SMK provider script in the Catalog to populate `thumbnail_url` with a link to a smaller image size.

## Implementation

krysal:
## Fixes

Fixes #675 by @stacimc

## Description

Sends the URL of the upstream thumbnail to Photon if it's available and it's **not** from Phylopic (since these are not working currently), so we can get the thumbnail faster and prevent timing out when the image is too big. See [this comment](https://github.com/WordPress/openverse/issues/675#issuecomment-1464969665) from the linked issue for more details on the thumbnail investigation.

## Testing Instructions

You can manually set a thumbnail URL from the database for any image and test if it's retrieved from the `/thumb` endpoint.

## Checklist

      errors.

[best_practices]: https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/views/image_views.py
```diff
@@ -107,17 +106,10 @@ def oembed(self, request, *_, **__):
...
-        if not image_url:
-            raise NotFound("Could not find image.", 404)
-
-        # Hotfix to use scaled down version of the image from SMK
-        # TODO Remove when this issue is addressed:
-        # TODO https://github.com/WordPress/openverse-catalog/issues/698
-        if "iip.smk.dk" in image_url:
-            width = settings.THUMBNAIL_WIDTH_PX
-            image_url = re.sub(r"!\d+,", f"!{width},", image_url)
+        # TODO: Remove the Phylopic check once the thumbnails are fixed.
```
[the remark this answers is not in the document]
krysal: Forgot to remove the comment, good catch! My plan is to leave it as is and correct the underlying data (in the catalog DB), so yes, it's the second option you mention.

api/catalog/api/views/image_views.py
```diff
@@ -107,17 +106,7 @@ def oembed(self, request, *_, **__):
     )
     def thumbnail(self, request, *_, **__):
         image = self.get_object()
-
-        image_url = image.url
-        if not image_url:
-            raise NotFound("Could not find image.", 404)
```
sarayourfriend, line 113: Is removing this check intentional/are we just relying on the upstream response from photon to return 404 in this case?

sarayourfriend (approved):
LGTM, just one question but not a blocker as I don't think it'll break anything.

api/justfile
```diff
@@ -83,7 +83,7 @@ collectstatic:
 #########
 
 # Run API tests inside the Docker container
-test *args: _all-up
+test *args:
```
dhruvkb, line 86: I think `test` needs to depend on `wait-up`, hence the merge conflict. I think using out this file as it exists on the `main` branch should be fine.

dhruvkb (approved):
LGTM!

api/justfile
krysal: Thanks Dhruv!

PR #904, opened 2023-03-13
Save cleaned up data during the cleanup step

### LINKED ISSUE #654: Add a list of domains that are known to support TLS to the cleanup step
## Problem

During the cleanup step, if the URL is invalid because it lacks the scheme (`http` or `https`), we add the scheme to the URL. To decide whether to add secure `https` or the insecure `http`, we check the domains for TLS support by pinging the secure version. We also create a cache for these responses. 
The inefficiency is in the way we re-create this cache on every data refresh in every cleanup worker:

https://github.com/WordPress/openverse-api/blob/3204b6e8f911335da2c810fb4c00f27068d21ab2/ingestion_server/ingestion_server/cleanup.py#L194

The ping sometimes fails for unrelated reasons, and we end up deciding that, for example, Flickr does not support TLS:
`2023-02-07 04:32:58,700 INFO cleanup.py:218 - TLS cache: {'www.flickr.com': False}`. 

Looking through the logs, I found other problems:
-  `.musee-mccord.qc.ca` is also incorrectly marked as not supporting TLS. This is caused by the redirect to a new domain `collections.musee-mccord-stewart.ca`
- `.thorvaldsensmuseum.dk` is marked as not supporting TLS because it redirects to `www.` address

## Description

We should create the cache object from a default object with some domains pre-filled. This is an object I ended up with when trying this locally and looking through the logs:
```
tls_cache = {
        "www.flickr.com": True,
        "commons.wikimedia.org": True,
        "https://www.eol.org/": True,
        ".eol.org": True,
        ".digitaltmuseum.org": True,
        ".geograph.org.uk": True,
        "www.geograph.org.uk": True,
        ".animaldiversity.org": True,
        "capl.washjeff.edu": False,
        "fotoarkivet.thorvaldsensmuseum.dk": True, // This endpoint requires API key so it's marked as False
        "www.thorvaldsensmuseum.dk": True,
        ".flora-on.pt": True
    }
```

## Additional context

Adding this cache will make sure that we add secure URLs to most image URLs that support them. This is important when we use the data saved here to update the upstream during data normalization.

### LINKED ISSUE #861: Add additional logging to the cleanup process
## Problem

Currently, the [cleanup](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/cleanup.py) process only records the final number of cleaned rows. We want to have more information on how many rows are affected and what type of cleanup is applied (malformed URL fixed, tag filtering) in order to develop a strategy to apply these changes on the upstream database and remove the steps from the ingestion server.

## Additional context

Previous attempt at https://github.com/WordPress/openverse-api/pull/1126

obulat:
## Fixes

Fixes #861 by @krysal
Fixes #654 by @obulat

## Description

This PR adds more logging for the data refresh, but its main goal is to be a proof-of-concept of saving the data during weekly data refresh as a preparation step for data normalization.

Data refresh image cleanup steps:

- Add http or https protocol to URLs that don't have a scheme in "url", "creator_url", "foreign_landing_url" fields
- Clean up tags:
    - tags that contain anything from the TAG_CONTAINS_DENYLIST set
    - AI-generated tags ("provider": "clarifai") with confidence level below TAG_MIN_CONFIDENCE = 0.90

~~This PR also adds a Wikimedia title cleanup step that removes File: prefix and file extension suffix from the image title. This step was added because in the Openverse Inserter PR it was specifically pointed out that those titles are bad for UX.~~ The Wikimedia title cleanup step can be added after during the second run of this PR in prod.

There is also a step that we need to add to the cleanup process for incorrect utf-8 tags, but I think we should add it in a later refresh (gist with the implementation) so as the cleanup step does not become much longer.

This PR saves one file per cleaned field in a tsv format. The files contain the image identifier and the cleaned data. I don't know where the best place to save them is - all suggestions welcome!

## Testing Instructions

Replace `sample_data/sample_images.csv` with the file in this gist (https://gist.github.com/obulat/b31e43b131352b8f6cd66a2dd87061d8), and run `just recreate` (or `just start` -> `just init`, if you haven't run the API before). You should see the tsv files recreated, logging about the cleaned fields:
```
2023-02-05 10:11:52 2023-02-05 07:11:52,223 INFO cleanup.py:276 - Finished saving cleaned data in 0.059366464614868164
2023-02-05 10:11:52 2023-02-05 07:11:52,223 INFO cleanup.py:353 - Batch finished, records/s: cleanup_rate=229.1038788216563
2023-02-05 10:11:52 2023-02-05 07:11:52,223 INFO cleanup.py:354 - Fetching next batch. Records cleaned so far: 1899, counts: {'tags': 320, 'url': 2, 'creator_url': 404, 'foreign_landing_url': 294, 'title': 224}
2023-02-05 10:11:52 2023-02-05 07:11:52,238 INFO cleanup.py:362 - Cleaned all records in 8.332297325134277 seconds, counts: {'tags': 320, 'url': 2, 'creator_url': 404, 'foreign_landing_url': 294, 'title': 224}
```

Also, check the logs about updated fields/values and TLS_CACHE.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -217,16 +217,15 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
     start_time = time.time()
     cleaned_values = {field: [] for field in all_fields}
     for row in rows:
+        source, _id, identifier = row["source"], row["id"], row["identifier"]
+
         # Map fields that need updating to their cleaning functions
-        source = row["source"]
-        _id = row["id"]
-        identifier = row["identifier"]
+        fields_to_update = {**global_field_to_func}
         if source in sources_config:
             source_field_to_func = sources_config[source]["fields"]
             # Merge source-local and global function field mappings
-            fields_to_update = {**global_field_to_func, **source_field_to_func}
-        else:
-            fields_to_update = global_field_to_func
+            fields_to_update |= {**source_field_to_func}
+
         # Map fields to their cleaned data
         cleaned_data = {}
         for update_field in fields_to_update:
@@ -246,9 +245,9 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
                 )
         # Generate SQL update for all the fields we just cleaned
         update_field_expressions = []
-        for field in cleaned_data:
-            update_field_expressions.append(f"{field} = {cleaned_data[field]}")
-            cleaned_values[field].append((identifier, cleaned_data[field]))
+        for field, clean_value in cleaned_data.items():
+            update_field_expressions.append(f"{field} = {clean_value}")
+            cleaned_values[field].append((identifier, clean_value))
 
         if len(update_field_expressions) > 0:
             update_query = f"""UPDATE {temp_table} SET
@@ -315,8 +314,7 @@ def clean_image_data(table):
     fields_to_clean = set()
     for p in sources:
         _fields = list(table_config["sources"][p]["fields"])
-        for f in _fields:
-            fields_to_clean.add(f)
+        fields_to_clean.update(_fields)
 
     cleanup_selection = (
         f"SELECT id, identifier, source, "
@@ -340,6 +338,7 @@ def clean_image_data(table):
         num_workers = multiprocessing.cpu_count()
         num_cleaned = 0
         cleaned_counts_by_field = {field: 0 for field in fields_to_clean}
+        cleanable_fields_for_table = _get_cleanable_fields("image")
 
         while batch:
             # Divide updates into jobs for parallel execution.
@@ -359,7 +358,7 @@ def clean_image_data(table):
                         batch[start:end],
                         temp_table,
                         source_config,
-                        _get_cleanable_fields("image"),
+                        cleanable_fields_for_table,
                     )
                 )
             pool = multiprocessing.Pool(processes=num_workers)
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -3,7 +3,7 @@ Functions for processing data when it is imported into the catalog.
 
 This includes cleaning up malformed URLs and filtering out undesirable tags.
 """
-
+import csv
 import logging as log
 import multiprocessing
 import time
@@ -49,6 +49,17 @@ TAG_CONTAINS_DENYLIST = {
 # Filter out low-confidence tags, which indicate that the machine-generated tag
 # may be inaccurate.
 TAG_MIN_CONFIDENCE = 0.90
+
+# We know that flickr and wikimedia support TLS, so we can add them here
+TLS_CACHE = {
+    "www.flickr.com": True,
+    "commons.wikimedia.org": True,
+    "https://www.eol.org/": True,
+    ".geograph.org.uk": True,
+    ".eol.org": True,
+    ".digitaltmuseum.org": True,
+    "www.geograph.org.uk": True,
+}
 
 
 def _tag_denylisted(tag):
@@ -154,6 +165,17 @@ _cleanup_config = {
 }
 
 
+def _get_cleanable_fields(table):
+    """
+    Extract global and sources-specific field names from
+    _cleanup_config for specific table.
+    """
+    cleanable_fields = []
+    for source in _cleanup_config["tables"][table]["sources"].values():
+        cleanable_fields += list(source["fields"].keys())
+    return cleanable_fields
+
+
 class TlsTest:
     """
     Test URLs to add the correct protocol when missing and use HTTPS when available.
@@ -184,19 +206,21 @@ class TlsTest:
         return True
 
 
-def _clean_data_worker(rows, temp_table, sources_config):
+def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
     log.info("Starting data cleaning worker")
     global_field_to_func = sources_config["*"]["fields"]
     worker_conn = database_connect()
     log.info("Data cleaning worker connected to database")
     write_cur = worker_conn.cursor(cursor_factory=DictCursor)
     log.info(f"Cleaning {len(rows)} rows")
-    tls_cache = {}
+
     start_time = time.time()
+    cleaned_values = {field: [] for field in all_fields}
     for row in rows:
         # Map fields that need updating to their cleaning functions
         source = row["source"]
         _id = row["id"]
+        identifier = row["identifier"]
         if source in sources_config:
             source_field_to_func = sources_config[source]["fields"]
             # Merge source-local and global function field mappings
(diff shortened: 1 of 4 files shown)


ingestion_server/ingestion_server/cleanup.py
```diff
@@ -50,6 +50,17 @@
...
+# We know that flickr and wikimedia support TLS, so we can add them here
+TLS_CACHE = {
+    "www.flickr.com": True,
+    "commons.wikimedia.org": True,
+    "https://www.eol.org/": True,
+    ".geograph.org.uk": True,
+    ".eol.org": True,
+    ".digitaltmuseum.org": True,
+    "www.geograph.org.uk": True,
+}
```
sarayourfriend, line 62: How did the others get added? Are they similar to Flickr and Wikimedia in that we just know that they do support TLS?

If that's the case, would it be worth manually testing providers for this and adding them to the list (understanding how tedious that is)? Or, is it something we need to monitor/update over time due to the potential for this status to change (I suppose, most likely, that someone starts to support it that previously didn't)?

Would moving this into Redis make sense at all (as an entirely separate issue) or is there a different future change that would persist this TLS support status?

sarayourfriend (approved):
This appears to have worked perfectly for me locally using the sample images file you shared.

Should we switch to using that sample images file permanently to ensure this feature is tested on a regular basis?

Here are the sample output logs I found to show it working:

```
openverse-ingestion_server-1  | 2023-03-13 22:32:11,922 INFO cleanup.py:258 - TLS cache: {'www.flickr.com': True, 'commons.wikimedia.org': True, 'https://www.eol.org/': True, '.geograph.org.uk': True, '.eol.org': True, '.digitaltmuseum.org': True, 'www.geograph.org.uk': True, 'www.eol.org': True}
openverse-ingestion_server-1  | 2023-03-13 22:32:11,922 INFO cleanup.py:259 - Worker committing changes...
openverse-ingestion_server-1  | 2023-03-13 22:32:11,923 INFO cleanup.py:265 - Worker finished batch in 3.2522239685058594
openverse-ingestion_server-1  | 2023-03-13 22:32:14,006 INFO cleanup.py:200 - https://musee-mccord.qc.ca/ObjView/M965.199.10008.jpg:403
openverse-ingestion_server-1  | 2023-03-13 22:32:14,006 INFO cleanup.py:103 - Tested domain .musee-mccord.qc.ca
openverse-ingestion_server-1  | 2023-03-13 22:32:14,006 INFO cleanup.py:243 - Updated url for 74454cfd-489d-4c7a-bdda-d7eef06d6d2b from '{dirty_value}' to '{clean}'
openverse-ingestion_server-1  | 2023-03-13 22:32:14,007 INFO cleanup.py:200 - https://musee-mccord.qc.ca/ObjView/5344.jpg:403
openverse-ingestion_server-1  | 2023-03-13 22:32:14,007 INFO cleanup.py:103 - Tested domain .musee-mccord.qc.ca
```

Are there any useful unit tests for us to add for this change? I'm not requesting changes for it because I'm not certain that testing log output is 100% necessary. However, if we're going to rely on it for analysis or something else like that it might be good to add a unit test to re-enforce the expected format.

sarayourfriend:
Oh, one requested change. I just noticed my git state was messy after reviewing this PR. Can we add the files to gitignore so they don't appear locally as changes?

ingestion_server/ingestion_server/cleanup.py
obulat: I've added these manually, by looking through the logs and adding the ones that were being tested. Your logs suggest that `.musee-mccord.qc.ca` should have also been added :)

>If that's the case, would it be worth manually testing providers for this and adding them to the list (understanding how tedious that is)? Or, is it something we need to monitor/update over time due to the potential for this status to change (I suppose, most likely, that someone starts to support it that previously didn't)?

In short, I think that the TLS support check, the way it's done right now, should go away after we finish the cleanup step (1-2 refreshes to get the updated TSV and 1-2 update DAG runs could be enough).

We do not test the URLs that have insecure `http` scheme for TLS support.  The main reason we _were_ testing for TLS support was to add a best scheme to the URLs that _don't have it_ (to convert urls like `www.flickr.com/image/path` to **https://**`www.flickr.com/image/path`). There are not so many such rows in the database, mainly the ones that were ingested before the `ImageStore` improvements in the catalog, that were also not re-ingested. When we use the TSV from the cleanup run to update the catalog, all of the URLs will have a scheme, whether it is `http` or `https`, so the URL cleanup function will be parsing the URL and not running the TLS support checks because the scheme will not be `""`.

https://github.com/WordPress/openverse/blob/2646c5ead465603b42c70f58a190f7b50861d698/ingestion_server/ingestion_server/cleanup.py#L83-84

Do you think we should monitor for TLS support status? If so, I think this should be a separate issue. We could add a cleanup function to test domains with `http` for TLS support and report all of them, and then test them and update the URLs if the support changes.

obulat:
> Should we switch to using that sample images file permanently to ensure this feature is tested on a regular basis?

I'm not sure. The end goal for these changes is to remove the cleanup step. Or at least to remove the functions that we can remove. So, optimally, we would not have any data in the catalog that does not have a scheme in the URLs, and tags that are denylisted or badly-formed. Then, this sample data would be wrong. Does it make sense to add these rows to sample data until we update the catalog, and remove it after we're done?

krysal (approved):
Perfect! Then, do you have a plan for the produced files here?

ingestion_server/ingestion_server/cleanup.py
sarayourfriend: Thanks for the explanation and motivation for this feature.

> Do you think we should monitor for TLS support status?

If we're hand-maintaining the list then yes, I think we should revisit it periodically or else the cleanup step here will apply the incorrect transformations.

I agree that that is out of scope of this issue, _sort of_, but if we're expanding the list of sites we're automatically applying the transformation to, then it does make the matter slightly more pressing as the area of effect is slightly wider. The added providers are small though, I think, so it's negligible. In any case, I agree it's a separate issue. I wanted to mention it in case it needs to be explicitly documented as such in a GitHub issue.

If it's something that will go away soon though, due to some other mechanism that will render this step unnecessary, then we can ignore it altogether and just document in the code that it's a temporary hold-over.

obulat:
> Perfect! Then, do you have a plan for the produced files here?

No, actually that's what I need help with. What's the best way of getting these TSV files from the Ingestion server to somewhere where the catalog can use them? I assume we should upload them to S3. Would it be more practical to somehow do it manually to avoid managing secrets here? @krysal ? @AetherUnbound ?

obulat:
I am going to merge this PR as is, and we can download the files from the box and upload them to S3 manually.

Hopefully, we will remove this process soon after we clean up the data, so it is an acceptable solution for that.

sarayourfriend:
@obulat Sounds good. The unit tests for ingestion server failed, I've re-run them to see if they pass.

Is there an issue for tracking the following up work?

AetherUnbound:
This can certainly be merged as-is, but I think it would be pretty straightforward to upload the files to S3! As Sara mentions, I don't _think_ it'd require any explicit permissions management, perhaps besides some IAM/role changes. We could even make a new `data-refresh` bucket specifically for this kind of thing.

PR #914, opened 2023-03-14
Use profiles in Docker Compose

### LINKED ISSUE #866: Use profiles in Docker Compose
## Problem

The Docker Compose file starts a lot of services that may not always be necessary.

## Description

Using [Docker Compose profiles](https://docs.docker.com/compose/profiles/), we can enable a subset of services to be spun up, as needed. This becomes even more useful when we add even more services from the catalog such as the 2 Airflow instances.

## Alternatives

Splitting up the Docker Compose file into smaller files, or launching individual services by name are options but using profiles is the most ergonomic option.

dhruvkb:
## Fixes

Fixes #866 by @dhruvkb

## Description

This PR configures Docker Compose to use profiles so that not all services need to be started at once. This allows us to selectively start services pertaining to the `api`, `ingestion_server` and `frontend`.

At the cost of increasing complexity, there is the scope of further cutting down the services needed (for example, the API doesn't always need the ingestion server, indexer worker and NGINX running). This PR does not do that, opting instead for a 1:1 package-profile mapping.

## Testing Instructions

Try the following recipes for all packages like the `api`, the `ingestion_server` and the `frontend`:

- `just <service>/profile-up`
- `just <service>/wait-up`
- `just <service>/init`

Also try running them from the root to make them work for all packages.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
@dhruvkb This PR isn't targeting main. Is that the intention? Just want to clarify before reviewing if there's another PR that also needs to be reviewed in conjunction to or before this one :thinking:

dhruvkb:
@sarayourfriend this was building on top of the Docker reorganisation PR. With that merged, the base automatically changed to `main`. I'll fix the conflicts immediately.

.github/workflows/ci_cd.yml
```diff
@@ -486,8 +486,8 @@ jobs:
 
       - name: collectstatic
         run: |
-          just _all-up
           just api/collectstatic
+        # `api/collectstatic` uses `api/dj` which depends on `api/wait-up`
```
sarayourfriend, line 490: This comment feels like a response to removing the `_all-up` line, I feel like you could just leave it out (and add a comment in the PR to explain the change) :sweat_smile:. The comment is confusing otherwise. Even with the context here I spent a few seconds trying to understand what the comment was documenting.

sarayourfriend (approved):
This is awesome! What a nice feature.

Is `profile-up` a non-negotiable name? Would `up` also be usable or is there a name conflict? `just api/up` is ever so slightly more convenient for typing over and over again :sweat_smile:

Does this change also mean it would now be viable to start running the frontend in the docker-compose stack with everything else instead of needing to run it on the host?

dhruvkb:
@sarayourfriend `up` does seem like a better name, I'll update the PR to reflect that. I named the recipe `profile-up` initially to not conflict with the existing `up` recipe and it just stuck.

I'm not sure what the benefit running the frontend locally inside Docker during development offers. It can be done if wanted/needed, but I'd prefer that to be a separate issue, discussion and PR to get this merged faster.

krysal (commented):
Great, profiles come in very handy to this use case 👍

It could be a good opportunity to fill out the documentation as well.

dhruvkb:
@krysal I would like to address the docs in a separate PR.

stacimc (approved):
Worked great for me! Docs as a follow up is fine with me as long as that's planned. Do we have an issue open for that or are you planning on fast-following?

dhruvkb:
I have already started working on the docs, will have the PR up by tomorrow.

sarayourfriend:
> I'm not sure what the benefit running the frontend locally inside Docker during development offers. It can be done if wanted/needed, but I'd prefer that to be a separate issue, discussion and PR to get this merged faster.

Definitely not for this PR. I think the benefit is making it unified and less reliant on local environment things like volta and pnpm versions and have that configured more universally in a Docker container. I'll cut an issue for it though, definitely not something I was expecting as part of this PR.

PR #932, opened 2023-03-16
use old nginx image name

zackkrida:
DO NOT MERGE. Used as the branch for a workflow, to deploy an old version of the API which needs the old name of the nginx image.

In the future when we rename a package we should push over all the historical versions as well.


sarayourfriend:
@zackkrida Is this PR still needed or can it be closed?

PR #933, opened 2023-03-16
Add application name to DB

zackkrida:
Adds our application name to the database options. This can be useful for identifying the source of queries in PostgreSQL's `pg_stat_activity` view or in query logs.


krysal (approved):

AetherUnbound (approved):
LGTM

api/catalog/settings.py
```diff
@@ -260,6 +260,9 @@ def _make_cache_config(dbnum: int, **overrides) -> dict:
         "USER": config("DJANGO_DATABASE_USER", default="deploy"),
         "PASSWORD": config("DJANGO_DATABASE_PASSWORD", default="deploy"),
         "NAME": config("DJANGO_DATABASE_NAME", default="openledger"),
+        "OPTIONS": {
+            config("DJANGO_DATABASE_APPLICATION_NAME", default="openverse-api"),
+        },
```
sarayourfriend, line 265: Is `OPTIONS` really meant to be a set? Is a key missing?

sarayourfriend (changes requested):

api/catalog/settings.py
AetherUnbound: Oh good catch

zackkrida: GAH, fixed. Thank you.

krysal: It's a [dictionary](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-OPTIONS) but I can't more specific information for Postgres yet.

AetherUnbound (approved):

krysal (approved):

PR #944, opened 2023-03-17
Absorb `build-nginx` job into `build-images` job

### LINKED ISSUE #899: Include `collectstatic` step inside API Dockerfile
## Problem

Currently the Dockerfile for the API (https://github.com/WordPress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/api/Dockerfile) uses the following build stages.

- `builder`
- `nginx`
- `api`

Here the `nginx` build stage requires that `just api/collectstatic` be run beforehand so that it can be copied into the image from the static files path. 

https://github.com/WordPress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/api/Dockerfile#L42

So to build the `api_nginx` image, the API needs to be spun up with the entire services orchestra just to collect static files.

https://github.com/WordPress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/.github/workflows/ci_cd.yml#L408

## Description

This problem can be avoided if staticfiles are collected during the build process itself. And copied into the `nginx` target from the appropriate stage. This will also drop the need for the NGINX build to be a separate job in CI.

https://github.com/WordPress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/.github/workflows/ci_cd.yml#L384

dhruvkb:
## Fixes

Fixes #899 by @dhruvkb

## Description

This PR merges `build-nginx` inside `build-images` thus making it the comprehensive image building job for the entire workflow. It also makes the build process faster by parallelising using matrix and removing the need to spin up the entire Docker Compose system.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
LGTM. Works fine locally. This will cause the published API image to increase by the size of the static directory. The last published image is 453 MB and this PR increases it to 480 MB. Locally the static directory for me is 25.6 MB so the change in size tracks. This will be (useless) duplicated between the API and the nginx image.

This PR does simplify things though. While it'd be nice to remove this duplication, the correct way to do it would be through shared mounts, which we just need to spend time configuring for ECS. It's not a big hurdle to configure this. I've opened an issue in the infrastructure repository to track this work.

dhruvkb:
@sarayourfriend yes, the API image does grow by ~25MB. I felt that the tradeoff was worth it given how much complexity it eliminates by being able to build the NGINX image in the matrix with all the others.

zackkrida (approved):
I support this change, the size increase seems worth the benefits, and it does work locally 👍

krysal:
How were you all able to see the size of the new image? After `recreate` my `web` service is only 67 MB in size and exits immediately after starting.

PR #954, opened 2023-03-20
Make `label` prop required for VIconButton

### LINKED ISSUE #618: Add required `aria-label` prop to VIconButton
## Problem

The IconButton component should always have an aria- label unless it is `aria-hidden`, which it shouldn't probably ever be considering it's an interactive part of the page. Without the label screen readers don't have the necessary information to describe the button to the user.

## Description

We should add a required prop for the `aria-label` attribute to enforce that it is always passed when using this component.

## Implementation

obulat:
## Fixes

Fixes #618 by @sarayourfriend

## Description

This PR adds a required prop named `label` to `VIconButton` to make sure that a button always has an accessible name. It uses a `label` prop instead of `aria-label` attr to make the validation easier, especially for elements that wrap around `VIconButton`.

## Testing Instructions

Try removing the `label` prop for one of the `VIconButton`s. You should see a warning in the console.
The CI failures will be fixed by #1137.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VHeader/VHeaderMobile/VSearchBarButton.vue
+++ b/frontend/src/components/VHeader/VHeaderMobile/VSearchBarButton.vue
@@ -71,7 +71,7 @@ export default defineComponent({
      * The label to use as accessible name for the button (aria-label).
      */
     label: {
-      type: String,
+      type: [String, Object] as PropType<TranslateResult>,
       required: true,
     },
   },
--- a/frontend/src/components/VHeader/VHeaderMobile/VSearchBarButton.vue
+++ b/frontend/src/components/VHeader/VHeaderMobile/VSearchBarButton.vue
@@ -1,6 +1,7 @@
 <template>
   <VIconButton
     class="border-tx bg-tx"
+    :label="label"
     :button-props="{ variant: 'plain--avoid' }"
     v-on="$listeners"
   >
@@ -65,6 +66,13 @@ export default defineComponent({
     rtlFlip: {
       type: Boolean,
       default: false,
+    },
+    /**
+     * The label to use as accessible name for the button (aria-label).
+     */
+    label: {
+      type: String,
+      required: true,
     },
   },
 })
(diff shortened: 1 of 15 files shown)


zackkrida (approved):
This is a good approach to requiring a standard html attribute. LGTM.

frontend/src/components/VIconButton/VIconButton.vue
```diff
@@ -63,6 +64,13 @@ export default defineComponent({
       type: Object as PropType<ButtonProps>,
       default: () => ({ variant: "plain" }),
     },
+    /**
+     * the label for the button; This is used for accessibility purposes.
+     */
+    label: {
+      type: String,
```
dhruvkb, line 71: If we make the `label` prop accept translate results, we can drop the `.toString` from all the different places where it is passed.

```typescript
import { TranslateResult } from 'vue-i18n'
```

```suggestion
      type: [String, Object] as PropType<string | TranslateResult>,
```

frontend/src/components/VHeader/VHeaderMobile/VSearchBarButton.vue
```diff
@@ -66,6 +67,13 @@ export default defineComponent({
       type: Boolean,
       default: false,
     },
+    /**
+     * The label to use as accessible name for the button (aria-label).
+     */
+    label: {
+      type: String,
```
dhruvkb, line 74: If we make the `label` prop accept translate results, we can drop the `.toString` from all the different places where it is passed.

```typescript
import { TranslateResult } from 'vue-i18n'
```

```suggestion
      type: [String, Object] as PropType<string | TranslateResult>,
```

dhruvkb (changes requested):
There is a scope to simplify the changes.

dhruvkb:
@obulat I think you're missing an `import type` in these commits^.

obulat:
> @obulat I think you're missing an `import type` in these commits^.

Updated locally. It's so much cleaner now, thank you for this improvement, @dhruvkb !

dhruvkb (approved):

PR #973, opened 2023-03-20
Use formless default browsable API renderer

### LINKED ISSUE #970: Reporting HTML view `SELECT`s all media records
The reporting views for each media type should only accept `POST` requests. Currently, they allow `GET` requests and Django renders an HTML form for submitting a report. This form contains a dropdown of all identifiers for each media model (images and audio, currently), which is an _extremely_ expensive query that taxes our production database and threatens to take down Openverse.

These views are currently blocked in Cloudflare and set to redirect to the API homepage until we fix the problem. `POST` requests to the endpoints continue to work as expected.

## Proposed solutions

1. Try only allow POST methods in Django by adding `@action(detail=True, methods=["POST"])` above the `def report` view logic. I tried this and the page successfully errors and returns json, but still renders the HTML form!

  ```json
  {
      "detail": "Method \"GET\" not allowed."
  }
  ```
  
2. Find a way to modify and remove the form from the HTML view
3. Something I haven't thought of yet.

sarayourfriend:
## Fixes

Fixes #970 by @zackkrida 

## Description

The view in question in the issue is the browsable API view. You'll get a 405 method not allowed, but the form below is the culprit here.

The new renderer added in this PR removes this form from all the views to prevent this kind of thing happening in any of our browsable views. This has the benefit of still maintaining the browsable API views which are quite useful to all developing the API or using it for their own purposes.

## Testing Instructions

Run the branch (`just up`) and visit the browsable API views. Test something like http://localhost:50280/v1/images/2982c0bf-9c26-4543-a712-4d50a117ded8/report/ specifically. On `main` you'll see a form underneath the JSON blob reporting the method not allowed error. On this branch, you'll see no form.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/settings.py
+++ b/api/catalog/settings.py
@@ -145,7 +145,7 @@ REST_FRAMEWORK = {
     "DEFAULT_VERSIONING_CLASS": "rest_framework.versioning.URLPathVersioning",
     "DEFAULT_RENDERER_CLASSES": (
         "rest_framework.renderers.JSONRenderer",
-        "rest_framework.renderers.BrowsableAPIRenderer",
+        "catalog.api.utils.drf_renderer.BrowsableAPIRendererWithoutForms",
         "rest_framework_xml.renderers.XMLRenderer",
     ),
     "DEFAULT_THROTTLE_CLASSES": (
--- a/api/catalog/api/utils/drf_renderer.py
+++ b/api/catalog/api/utils/drf_renderer.py
@@ -1,0 +1,27 @@
+from rest_framework.renderers import BrowsableAPIRenderer
+
+
+class BrowsableAPIRendererWithoutForms(BrowsableAPIRenderer):
+    """
+    Renders the browsable api, but excludes the forms.
+
+    See https://github.com/WordPress/openverse/issues/970
+
+    CC BY 3.0 Brad Montgomery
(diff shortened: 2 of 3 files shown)


api/catalog/settings.py
```diff
@@ -145,7 +145,7 @@
     "DEFAULT_VERSIONING_CLASS": "rest_framework.versioning.URLPathVersioning",
     "DEFAULT_RENDERER_CLASSES": (
         "rest_framework.renderers.JSONRenderer",
-        "rest_framework.renderers.BrowsableAPIRenderer",
```
dhruvkb, line 148: I'm curious to know why not turn off the browsable API in production.

dhruvkb (approved):
LGTM!

api/catalog/settings.py
zackkrida: We discussed in Slack, but generally we find it quite useful while browsing the API in a browser!

dhruvkb: In production?

zackkrida (approved):
LGTM. Confirmed that this does not execute the `SELECT * from images` query as well.

PR #986, opened 2023-03-22
Remove XML from the API

### LINKED ISSUE #982: Remove XML support from Django API
## Problem

The Django API depends on this package to enable XML support: https://pypi.org/project/djangorestframework-xml/

However, we don't document this XML support anywhere nor document the expected output.

Over the last month we've received 4000 requests, but those are almost entirely all in the last week, during which there was a production incident related to the browsable API pages. I suspect that these XML requests are coming from something crawling the browsable API pages and clicking the links to the different formats you can request in: 
image

## Description

Remove the XML renderer and the package providing that capability. Keeping an unused and undocumented XML renderer is at best a niche feature that no one uses and at worst a security vulnerability (search "XML vulnerabilities" for a taste of what I mean here).

## Alternatives

Leave XML in but document it as an intended feature, including the format of its output.

obulat:
## Fixes

Fixes #982 by @sarayourfriend 

## Description

The PR removes the XML renderer and a test that was using it.

@sarayourfriend, do you think we should keep the XML for `oEmbed` endpoint in light of @dhruvkb's [comment](https://github.com/WordPress/openverse/pull/986#discussion_r1144922318)?
If so, I wonder if it would be possible to remove XML and only have it for a single endpoint?

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/test/image_integration_test.py
```diff
@@ -125,27 +124,6 @@ def test_oembed_endpoint_for_json():
     assert parsed["license_url"] == "https://creativecommons.org/licenses/by/2.0/"
 
 
-def test_oembed_endpoint_for_xml():
```
dhruvkb, line 128: The [oEmbed spec](https://oembed.com/) allows response to be either JSON or XML so while dropping XML is not a spec violation, it will be a breaking change for the API.

obulat: I couldn't get the oEmbed endpoint to work :( Has it been working?
I tried this URL:
https://api.openverse.engineering/v1/images/oembed/?format=json&url=https%3A//www.flickr.com/photos/7788419@N05/15218475961
And the same without the `format=json` param, and with `format=xml` - always Cloudflare gateway error 502. 

Is there a redirect in Cloudflare?

api/Pipfile.lock
```diff
@@ -960,6 +960,56 @@
             "markers": "python_version >= '3.7'",
             "version": "==4.1.0"
         },
+        "orjson": {
```
dhruvkb: @obulat if `orjson` is needed but is not listed as a dependency, it can be wiped out in a future lock or sync. I recommend adding it to the `Pipfile` as a dependency.

api/test/image_integration_test.py
dhruvkb: Yeah, I'm all for dropping these niche endpoint but I also don't want to cause breakages for those who are using them. Data on the usage of this endpoint would be nice.

zackkrida: here's some!

CleanShot 2023-03-22 at 14 46 35@2x
CleanShot 2023-03-22 at 14 46 17@2x

dhruvkb: I think we can delete it if it's only been used 8 times in the last month.

zackkrida: Especially considering at least half of them were probably Olga testing 😁

sarayourfriend: There is context for the removal suggestion in the issue as well. I had the same question as Dhruv and didn't want to remove it if it was being used. We've had several thousand XML requests over the last week, but before that it's basically no one. The usage of the XML is definitely weird for the last week!

sarayourfriend:
>  If so, I wonder if it would be possible to remove XML and only have it for a single endpoint?

This is possible, for sure, we just have to define the renderer classes on the View rather than in the list of default renderer classes. More info here in the DRF documentation about how to do that: https://www.django-rest-framework.org/api-guide/renderers/#setting-the-renderers

My preference is to remove it entirely, even if it makes the oEmbed endpoint technically out of spec. Considering the usage I think this is fair, but I'm of course open to being wrong about that. If other folks feel strongly it should be kept, then scoping it to the one endpoint would be a reasonable compromise rather than removing it entirely.

api/test/image_integration_test.py
zackkrida: @sarayourfriend I definitely think your "an api consumer was crawling the ui and following links" theory makes a lot of sense because it would also potentially explain the GET requests to the reporting endpoint.

sarayourfriend: That's exactly what I was thinking too. We don't include the report URL in the result object, but it is in the "additional actions" list on the detail URL's browsable page, so very clickable for a crawler.

sarayourfriend (approved):
LGTM!

api/catalog/api/serializers/image_serializers.py
```diff
@@ -148,11 +148,12 @@ class OembedSerializer(BaseModelSerializer):
     """
 
     version = serializers.ReadOnlyField(
-        help_text="The oEmbed version number. This must be 1.0.",
+        help_text="The oEmbed version number, always set to 1.0.",
+        required=True,
```
obulat, line 152:
```suggestion
        help_text="The oEmbed version number, always set to 1.0.",
```

obulat:
Sorry, @sarayourfriend, I accidentally clicked on "Request review" next to your name, and there's no undo :)

dhruvkb (approved):
The changes look good to me.

PR #987, opened 2023-03-22
Add tag app release action

### LINKED ISSUE #976: Cutting a release does not successfully run CI/CD workflow
## Description

I cut a release in anticipation of deploying the production API today. My understanding of the production deployment process was:

1. Cut a release
2. Wait for CI/CD to pass and publish the images
3. Run the workflow for the production API deployment with the release name as the image tag

However, step 2 never passes: https://github.com/WordPress/openverse/actions/runs/4475172448

I knew we'd deployed production recently, so I looked at the production API deployment workflow logs and saw that we'd just been deploying production to a SHA.

@zackkrida you're the only one I saw doing this so I'm pinging you as I'm not sure who else to ask: did you know that this process was broken (is this a known issue)? Was it decided that we would deploy production API and Nuxt to SHAs rather than release version tags?

If that is the case... should we do away with the "release" process altogether?

Additionally, I'll need to know if this is the case in order to update #907 with that information.

cc @AetherUnbound and @dhruvkb as y'all have the most time working on the workflow files, maybe y'all will know the details here.

## Additional context

Marking as critical because not having a clearly documented production release process is a recipe for disaster. If someone who didn't know where to look to find out how to do this otherwise needed to deploy production in an emergency, they'd be lost, and quite unjustly so.

sarayourfriend:
## Fixes

Fixes #976 by @sarayourfriend 

## Description

This PR adds a new `Release app` workflow. This workflow requires the choice of an app (api, ingestion server, or frontend) and an existing image tag for that app. It then does the following:

1. Generate a new release tag based on the UTC timestamp when the workflow is dispatched
2. Create a new docker image tag based on the existing image(s) with the timestamp (`rel-{timestamp}`). The docker image tag does not include the app name as it'd be redundant with the rest of the image name
3. Creates and publishes a release with the changes since the previous release. Until each application has it's own release generated by this workflow, they will all the last "General" release as the "previous release". Currently that is v4.0.1. Once each of them have their own releases from this workflow, however, from then on they will derive the changes between the previous release for that app only up until the commit corresponding to the docker image SHA.
   - This step also creates the git tag. We do not need to manually do it ourselves as creating a GitHub release also creates the git tag
4. Pulls the text from the release (the changelog) and generates a Markdown file to add to the documentation site that includes all the changelogs (starting with when we begin using this process).
5. If run for the API or frontend, then it also dispatches the corresponding production deployment workflow

## Testing Instructions

This was tested extensively on this branch. A summary of that can be found here: https://github.com/WordPress/openverse/pull/987#issuecomment-1482270090

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


.github/release_drafter_api.yml
```diff
@@ -0,0 +1,30 @@
...
+    label: "✨ goal: improvement"
+  - title: Internal Improvements
+    labels:
+      - "🤖 aspect: dx"
+      - "🧰 goal: internal improvement"
+  - title: Bug Fixes
+    label: "🛠 goal: fix"
+change-template: "- $TITLE (#$NUMBER) @$AUTHOR"
+include-labels:
+  - "🧱 stack: api"
```
AetherUnbound, line 22: Being able to make this distinction is SO COOL, WOW! 🤯

.github/workflows/tag-app-release.yml
```diff
@@ -0,0 +1,123 @@
+name: Tag app release
+
+on:
+  workflow_dispatch:
+    inputs:
+      app:
+        type: string
+        required: true
+        description: Application to tag for. Either `ingestion_server`, `api`, or `frontend`.
```
AetherUnbound, line 9: Could this potentially be of type `choice`? [Example](https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/). That way we wouldn't have to let this be a free-form field and perform the validation step below explicitly.

documentation/changelogs/api/2023.03.22.04.56.29.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Bug Fixes
+
+- Switch to internal header on single results (#981) @zackkrida
+- Add GH_TOKEN to the gh steps (#984) @obulat
+- Add decoding of the strings that don't have backslashes (#979) @obulat
+- Update to new link for pipenv install instructions (#962) @kk311y
+- Treat any non 200 status as failure for thingiverse (#940) @sarayourfriend
+- Skip build and publish job if nothing to do (#977) @dhruvkb
+- Make `searchTerm` for VAudioTrack and VImageCell optional (#955) @obulat
```
AetherUnbound, line 27: Should this be part of the API changelog? It doesn't appear to edit any API files nor does it have the API stack label 🤔

documentation/changelogs/api/2023.03.22.04.56.29.md
```diff
@@ -0,0 +1,33 @@
+# 2023.03.22.04.56.29
+
+## Improvements
+
+- Absorb `build-nginx` job into `build-images` job (#944) @dhruvkb
```
AetherUnbound, line 5: One downside of having this as a markdown doc and not in an official release is that the reference numbers are no longer auto-linked by GitHub (i.e. `#944` won't go anywhere when this is published to our docs site). Is it possible to update the action to hard-link these?

.github/workflows/tag-app-release.yml
```diff
@@ -0,0 +1,123 @@
...
+        # We don't actually ever use these releases, but because there's no easier
+        # way to generate the release notes for an application than this action,
+        # and because the action doesn't support generating the notes without
+        # adding them to a release (draft or no), we still let the release get created
+        # and then just copy the body from the result of this action.
+        uses: release-drafter/release-drafter@v5
+        id: release-drafter
+        with:
+          config-name: release_drafter_${{ inputs.app }}.yml
+          version: ${{ steps.tag.outputs.tag }}
```
AetherUnbound, line 91: One concern with this - what does this diff _against_ when it's drafting the release? If it is naively only the last release that was made, we either get in a situation where:
1. Every subsequent changelog includes all previous changes since this workflow was introduced _plus_ the new changes
2. Every changelog only includes the changes _made to the `app` in question since whenever this workflow was last run_, regardless of `app` (e.g. if the runs were `api` -> `frontend` -> `api`, the latter `api` changelog would only have changes from `frontend` -> `api`.

AetherUnbound (changes requested):
Generally I love the idea behind this PR, but I have some concerns about the end result of these changes.


Regarding your outstanding questions:
1. I also would prefer to keep git tags. I would prefer to have the tags prepended by the app name rather than appended, that way when viewing the tags in an alphabetically sorted list, it's easy to see when the tags were made for an individual service.
2. I think either having this run as part of the deployment workflows or having this kick off the deployment workflow makes a lot of sense. Would this be called as part of the staging deployment workflow, and then the tag would be used manually for production? That would make the most sense to me, given we usually want to deploy to staging prior to prod.

.github/workflows/tag-app-release.yml
sarayourfriend: Yes, definitely. I totally overlooked that.

documentation/changelogs/api/2023.03.22.04.56.29.md
sarayourfriend: As noted in the commit that adds these files, they're just here for testing: https://github.com/WordPress/openverse/pull/987/commits/1a757d19bee7b4f57068a48e105132affe81355f

I copied the file from the existing draft release, it has nothing to do with the release-drafter workflow configurations introduced in this PR.

documentation/changelogs/api/2023.03.22.04.56.29.md
sarayourfriend: Good point! The easiest way is probably through this sphinx extension: https://github.com/sloria/sphinx-issues

We could also update the change template in release-drafter (https://github.com/release-drafter/release-drafter#change-template-variables) but I think the sphinx plugin would be more broadly useful in the end as it would apply to _all_ docs, not just these ones.

.github/workflows/tag-app-release.yml
sarayourfriend: Oh jeez, that's a glaring issue with using release-drafter in this way. I believe it would just compare to the previous GitHub release. I think specifically we'd end up with the first case, each changelog including all previous changes for the app in subsequent changelogs.

I think the only way to get around this is going to be to generate the change list ourselves or find another action that allows you to define which commit to start from.

We could also fork release-drafter to add that capability to it ourselves (and while we're add it, the ability to stop it from creating the draft release and instead just build the markdown output).

I'll dig around and see if there's another action we can use or if we'll need to take an independent approach here.

sarayourfriend:
> I also would prefer to keep git tags. I would prefer to have the tags prepended by the app name rather than appended, that way when viewing the tags in an alphabetically sorted list, it's easy to see when the tags were made for an individual service.

Agreed, will do this in the next commits.

> I think either having this run as part of the deployment workflows or having this kick off the deployment workflow makes a lot of sense. Would this be called as part of the staging deployment workflow, and then the tag would be used manually for production? That would make the most sense to me, given we usually want to deploy to staging prior to prod.

It hadn't occurred to me to create the tag during staging: but then what's the point of the changelogs? I think this process is only useful if we're going to have any delay between staging and production release, which right now I think we should without doing live integration testing.

The way I envisioned this being used was:

1. Staging deploys like normal, to a SHA
2. Once we're happy with the set of changes for an app in staging, we copy the SHA for the current deployment and then run this workflow

The only modification I could imagine is to make the production deployment workflows call this new one to generate the tag from the supplied SHA and then also immediately deploy it. That way we wouldn't have to copy the SHA and then also copy the resulting tag out from this new workflow: one less change to accidentally copy the wrong thing somehow.

.github/workflows/tag-app-release.yml
sarayourfriend: Digging more into it, other people use release-drafter with monorepos, so we should be able to do so as well: https://github.com/release-drafter/release-drafter/issues/1294

It looks like `tag-prefix` might tell release drafter how to find the "previous tag" to use when comparing. The docs aren't clear enough on it though, so I'll need to dig and find if that's the case.

sarayourfriend: Yup, after looking at the code it does just that: when it searches for the "latestRelease" it filters the releases by `tagPrefix`: https://github.com/release-drafter/release-drafter/blob/master/lib/releases.js#L57-L59

_However_, this requires you to actually be using GitHub releases.

Which means we have a few options:

1. Go our own way and don't use release-drafter or use a fork of it modified to look at tags generally, not just GitHub releases
  - Only necessary if we're against GitHub release entirely
  - Big downside of needing a lot of extra work
2. Use release drafter the way guardrail is doing so and have it draft releases for each of our three applications. Create a new workflow triggered by releases that parses the release's name to get the name of the app and tags the image and then triggers the production deployment only for that app (by dispatching the corresponding production deployment workflow)
  - Benefit here is that it's pretty simple
  - One downside is that it will still be stuck using the meaningless semantic versioning. We could update the version template to just increment the major version number every time and drop the minor and patch number: https://github.com/release-drafter/release-drafter#projects-that-dont-use-semantic-versioning
  - Another downside is that we wouldn't have the changelogs on the documentation site, only in GitHub releases, in one, singular list (not split by app other than the prefix)
3. A hybrid of the two and the current approach: Stop drafting releases on an ongoing basis and instead use this workflow to create the release and immediately publish it. Automatically dispatch the production deployment with the tag generated. We would rename the workflow to just "Release" and keep the inputs as they are now.
  - Benefits:
    - Changelogs get added to the documentation site and can be split by application (making it easier to see the progression and cadence for individual apps in a list)
    - Lets us use CalVer instead of SemVer which is a more meaningful way of marking our releases
  - The only downside is that it's slightly more work than the second option. However, almost all of that work has already been done in this PR, we just have to make a few tweaks and fixes.

My vote is to do the third one and because it's the one that requires the medium amount of work of the three, I'll go ahead and update the branch do reflect it and then we can decide between it and the second option. To me the first option is a no-go, I just don't see why we would pour that much effort into this for so little benefit compared to the other two.

I'll make the following changes:

1. Remove the "draft release" workflow that currently creates the drafted releases on an ongoing basis for each push to main (this needs to happen anyway for this PR)
2. Change the workflow name to "Deploy production"
3. Change the app input to a choice, as you suggested in another comment
4. Update the release-drafter configuration to tell it to immediately publish the release
5. Dispatch the correct deployment workflow if the app was `api` or `frontend`.

documentation/changelogs/api/2023.03.22.04.56.29.md
AetherUnbound: Woah, what a cool extension!

.github/workflows/tag-app-release.yml
AetherUnbound: Thanks for doing all this investigation! I think your plan of action sounds great, and looks like we'd be prepending the app name anyway with this approach 💃🏼

documentation/changelogs/api/2023.03.22.04.56.29.md
sarayourfriend: https://github.com/sloria/sphinx-issues didn't work due to it relying on a custom sphinx role, but I found this SO answer with a handy bit of code that works perfectly for us: https://stackoverflow.com/questions/72424748/myst-parser-auto-linking-linkifying-references-to-bug-tracker-issues

Tested it locally with this PR:

image

The "code" was just `Test #987` :tada: 

This is in the latest commit.

sarayourfriend:
I've implemented the changes from this conversation: https://github.com/WordPress/openverse/pull/987#discussion_r1147015594

Thanks heaps @AetherUnbound for the review. You caught some nasty issues with my first attempt here.

sarayourfriend:
@AetherUnbound This workflow would also make the "bump" script in the infrastructure config no longer work for the ingestion server, at least not with its current implementation. We'll need to add some tag-prefix filtering there as well to search for the latest _ingestion server_ specific version from the repo rather than the way it works now.

sarayourfriend:
I've removed the changelog files I was using for testing. If you want to see how the whole thing will look on the documentation site, checkout https://github.com/WordPress/openverse/pull/987/commits/905bcbf24fb9bce357ec72c205297d8999509a26 and then run the site locally (`just documentation/live`).

sarayourfriend:
Thankfully the testing revealed a number of configuration issues. Everything up until the release-drafter works. However, release-drafter _requires_ the configuration file to live on `main`. As this PR hasn't yet been merged, that's not the case.

Therefore, in the interest of being able to test this, I think we should add the `release-drafter-api.yml` in this PR to the `main` branch in a separate PR. We only need to add one to be able to finish testing.

I'm hoping that the release-drafter part will "just work" but I'm convinced that something about the configuration is ever so slightly off still and will need tweaking.

I'll open the PR to add the api configuration to main for this PR to be able to use it to finish testing.

sarayourfriend:
Further testing is blocked until https://github.com/WordPress/openverse/pull/1011 is merged

sarayourfriend:
Alright! This is fully working now.

Here's a screenshot of the generated release (I deleted the release as it wasn't a genuine one):

image

Here's the closed PR that added the changelogs for that fake release: https://github.com/WordPress/openverse/pull/1016

Notice that the changelog only includes API changes, nothing else :tada: 

You can see the release tags added to the `openverse-api` and `openverse-api_nginx` images here: https://github.com/wordpress/openverse/pkgs/container/openverse-api and https://github.com/wordpress/openverse/pkgs/container/openverse-api_nginx

I've deleted the git tags from the repository, but here's the list before I deleted them:

image

In short: the new action does what it's supposed to do!

1. Create a git tag for the specified SHA (and only if the SHA corresponds to a published docker image)
2. Add the release tag to the docker image
3. Publish a release (this is an ongoing thing we'll need to allow release drafter to pull only the changes since the last release... it relies explicitly on the GitHub releases, not tags!)
4. Open a PR with the changelogs added to the documentation site

After I remove the testing code, it'll also deploy the production API or frontend with the new docker image tag.

documentation/changelogs/api/index.md
```diff
@@ -0,0 +1,10 @@
+# API changelogs
+
+Changelogs for <https://api.openverse.engineering/>.
+
+```{toctree}
+:glob:
+:titlesonly:
+
+*
+```
```
sarayourfriend, line 10: This applies to all the index files in the changelog directories: the reason for this configuration is that it lists only the titles of the changelogs in ascending order. We might want to reverse this to have it list them in descending order so that the most recent changelog is at the top (as with most changelogs).

AetherUnbound: I definitely find descending-order changelogs to be the most helpful, if that's a quick change to make!

AetherUnbound (approved):
This is fantastic, thanks for all the effort you put into getting this to work and making sure it functions as expected! I have a bunch of minor notes, but nothing to block this PR (all could be a quick add or fast-follow for this). I'm really excited to see how this new process works 🚀

documentation/changelogs/api/index.md
```diff
@@ -0,0 +1,10 @@
+# API changelogs
+
+Changelogs for <https://api.openverse.engineering/>.
+
+```{toctree}
+:glob:
+:titlesonly:
+
```
zackkrida, line 8:
```suggestion
:glob:
:reverse:
:titlesonly:


```

I _believe_ this is all we need to do to swap the order!

https://github.com/sphinx-doc/sphinx/pull/2527

documentation/changelogs/frontend/index.md
```diff
@@ -0,0 +1,10 @@
+# Frontend changelogs
+
+Changelogs for <https://openverse.org/>.
+
+```{toctree}
+:glob:
+:titlesonly:
+
```
zackkrida, line 8:
```suggestion
:glob:
:reverse:
:titlesonly:

```

documentation/changelogs/index.md
```diff
@@ -0,0 +1,18 @@
...
+"Release app" workflow.
+
+The lists of changes are derived using the "stack" labels and are organised by
+type of change. The documents are generated using the
+[`release-drafter`](https://github.com/release-drafter/release-drafter)
+workflow.
+
+```{toctree}
+:maxdepth: 1
+:glob:
```
zackkrida, line 13:
```suggestion
:maxdepth: 1
:glob:
:reverse:

```

documentation/changelogs/ingestion_server/index.md
```diff
@@ -0,0 +1,10 @@
+# Ingestion server changelogs
+
+Changelogs for the internal ingestion server service.
+
+```{toctree}
+:glob:
+:titlesonly:
```
zackkrida, line 7:
```suggestion
:glob:
:reverse:
:titlesonly:
```

zackkrida (approved):
This is absolutely amazing. I left a suggestion for reversing the changelogs but I'd love to get this merged and tested on monday.

documentation/index.md
```diff
@@ -4,6 +4,8 @@ This documentation is for developers who are building Openverse. If you want to
 use the Openverse API, you should instead refer to the
 [API consumer documentation](https://api.openverse.engineering/).
 
+Test #987
+
```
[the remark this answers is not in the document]
sarayourfriend: Yes :sweat_smile:

documentation/changelogs/api/index.md
sarayourfriend: Yup! Should be.

.github/workflows/release-app.yml
```diff
@@ -0,0 +1,165 @@
...
+
+          if [ "${{ inputs.app}}" = "api" ]; then
+            docker buildx imagetools create ghcr.io/wordpress/openverse-api_nginx:${{ inputs.image-sha }} --tag ghcr.io/wordpress/openverse-api_nginx:${{ steps.tag.outputs.image-tag }}
+          fi
+
+      - name: Create and publish release
+        uses: release-drafter/release-drafter@v5
+        id: release-drafter
+        with:
+          config-name: release-drafter-${{ inputs.app }}.yml
```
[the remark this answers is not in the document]
sarayourfriend: I don't know if it is possible. The documentation says that it should be in the `.github` folder: https://github.com/release-drafter/release-drafter#action-inputs

I tried to look in the code to see if it would support directories and it's a bit of a rabbit hole that I didn't try to find the end of. If we wanted to keep looking, we'd need to start here and see how this config method resolves the file: https://probot.github.io/api/latest/classes/context.Context.html#config

It also makes it seem like the file should just be in `.github` though.

documentation/_ext/link_issues.py
```diff
@@ -0,0 +1,370 @@
+"""Issue linking w/ plain-text autolinking, e.g. #42
```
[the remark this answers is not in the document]
sarayourfriend: Is there a specific process for doing so aside from including it in the repository?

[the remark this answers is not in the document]
AetherUnbound: It seems like with the `myst-parser`, another extension we added, we were able to include it as a package: https://github.com/WordPress/openverse/blob/185d9e3c2bb35b088a083159cd29466a9127b673/documentation/Pipfile#L12

I assumed we might be able to follow a similar path with this, using their installation instructions: https://github.com/ignatenkobrain/sphinxcontrib-issuetracker#installation

.github/workflows/release-app.yml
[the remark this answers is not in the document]
AetherUnbound: Fair enough 🙂

documentation/_ext/link_issues.py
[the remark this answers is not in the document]
sarayourfriend: Gotcha. No we can't use the official package as it doesn't work with modern versions of Sphinx, hence the linked SO answer.

Issue #200, opened 2022-03-25
Stop binding ports in the reservable range locally


sarayourfriend:
## Problem

We currently bind ports in the reservable range, except for the API proxy in the frontend repository.

This means that our local development services can clash with system services (not to mention default ports of other projects that are also binding in the reserveable range; how many things do you know that run on 3000 or 8080 or 8000?)

For example SyncThing runs on port 8384 by default which is fine because it's a non-ephemeral system service.

Also there's always the possibility that this will happen again: https://developer.apple.com/forums/thread/682332

And there's nothing wrong with an OEM adding a system service in that range, the range is specifically reserved for such things. We're just lucky that nothing has conflicted with our choices (or the default values) thus far.

## Description

There is actually a range of ports available for dynamic and ephemeral services: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports

Let's switch our services to use those so that we avoid conflicting with other projects or, indeed, other running applications that could be critical to a developers flow or conflict with a new system service from an OEM or from the dev themselves.

It'd be nice if we just used a solid block of ports so that it's easier to remember where each service is running locally. For example we could do the following, going lowest in the stack to highest, with the main services grouped together and then auxiliary services (Databases, Redis, etc) grouped after those.

* Airflow -> 65000
* API -> 65001
* Nuxt -> 65002
* Everything else (multiple PG instances, redis, etc) -> 6500{3..}

Alternatively, the other direction as Madison suggested:

* Nuxt -> 65000
* API -> 65001
* Airflow -> 65002
* Everything else (multiple PG instances, redis, etc) -> 6500{3..}

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

AetherUnbound:
I actually think that the reverse might be best, start with frontend -> API -> Airflow. The higher the port, the "further back in the stack". As you mention, databases and other supported services could just be added on incrementally as needed. I like this idea, because I can never remember which port something is on and using an explicitly _different_ port would allow me to bookmark that on localhost as well :slightly_smiling_face: 

I think that having a monorepo (#192) would make this organization a _lot_ easier, rather than having to cross reference which next port in 3 different repos is available for some new service.

sarayourfriend:
> I actually think that the reverse might be best, start with frontend -> API -> Airflow.

[...]

PR #990, opened 2023-03-22
Retry `up` recipe in case port is occupied

dhruvkb:
## Fixes

Addresses https://github.com/WordPress/openverse/issues/200#issuecomment-1257716418

## Description

Since there are flaky occurrences of ports being in use, this PR makes `just up` run thrice with 5 second pauses so that the port can be grabbed. I think this should work because the issue goes away during the re-run of the failing jobs.

## Testing Instructions

As seen in the screenshot, the first run is the output when a service is bound to port 50254 and thus the `db` service, which needs it cannot run. The second run is the output when that port is available.

Screenshot 2023-03-22 at 9 13 26 PM

If the port is freed up within 15 seconds, the `up` command will succeed and the CI+CD jobs will continue to run.

## Checklist

- [X] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [X] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
@dhruvkb It seems we were occupied by the same concerns recently :) I wrote this issue proposing an alternative solution: https://github.com/WordPress/openverse/issues/1035

This one doesn't add _too_ much complexity, so I don't think the approach in this PR isn't worth doing. But I think the solution I proposed in the linked issue would solve the problem outright with absolutely no chance of flakiness: provided it is actually possible.

dhruvkb:
@sarayourfriend your observation about not needing to bind ports is true for CI. That's seems like a concrete avoidance of the problem compared to bandaging on top of it (that this PR does). It does add a little bit of complexity, however `just dc` takes care of assigning the right Compose config file, so the overriding of multiple Compose files can be handled by `just` itself.

Nonetheless, retrying `up` does not cause any negative effects so I'll merge this PR if it gets the two approvals. If #1035 is resolved first, I'll close this PR as unmerged.

obulat (approved):
LGTM!

stacimc (approved):
Looks good and discussion [here](https://github.com/WordPress/openverse/pull/990#issuecomment-1484518542) makes sense to me 👍

PR #996, opened 2023-03-23
Implementation Plan: Filtering and designating sensitive results in the API

### LINKED ISSUE #937: Implementation Plan: API sensitive terms filtering and designation
## Description

Part of https://github.com/WordPress/openverse/issues/377
Project Proposal: https://github.com/WordPress/openverse/blob/main/rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md#required-implementation-plans

The implementation plan must cover the following key aspects in addition to the baseline expectations:

- Must cover efficiently filtering out results with sensitive textual content in addition to including the sensitivity designation on each result.
- Consider the details outlined in the [filtering](https://github.com/WordPress/openverse/blob/main/rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md#implementation-details) and [designation](https://github.com/WordPress/openverse/blob/main/rfcs/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.md#implementation-details-1) sections in the project proposal.

Refer to https://github.com/WordPress/openverse-api/pull/1108 for reference on using `reindex`.

~**Blocked on consultation with Elasticsearch experts**~

sarayourfriend:
## Fixes

Fixes #937 by @sarayourfriend 

Opened 30 March, **due 13 April**.

## Description

I've requested reviews from Staci and Dhruv as the IP touches Airflow, API, and ingestion server changes. Staci and Dhruv are both knowledgeable in at least one or several of these areas. If either of y'all are unable to review this, please let me know and I'll ask someone else.

## Testing Instructions

Read the IP. I do not have any outstanding questions, but please focus on the plan and make sure that it makes sense and is feasible. If you have suggestions for how to break any steps up into separate PRs, please share those especially.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
+++ b/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
@@ -91,6 +91,23 @@ sensitive result designation than I had originally devised. This solution is
 easily 10x more intelligible and simpler to document and implement than the
 others I had considered.
 
+## Feasibility
+
+Whether the approach in this implementation plan is feasible was an open
+question for some time. After extensive discussions with folks who know
+Elasticsearch well, I decided on this approach and was able to test it in
+staging. The full process, index creation, reindex, refresh, was run on a
+production dataset from before iNaturalist was added. This means the origin
+index was smaller than our current one, but still large: 550 million documents.
+In staging, the whole process took just over six and a half hours. The resulting
+index had 444 million documents, meaning it excludes approximately 20% of the
+origin index.
+
+Therefore, I consider this approach feasible: it is sufficiently quick, it
+produced a dataset of unsurprising size (once quoted terms were used to prevent
+false positives), and is easy to implement with the tools we already have, use,
+and (mostly) understand.
+
 ## Technical description of plan
 
 Each heading refers to one of the steps in the previous section.
--- a/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
+++ b/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
@@ -278,14 +278,10 @@ To enable running filtered index creation independently of the data refresh
 DAGs, we will create a separate DAG that will be triggered by the data refresh
 DAGs using the
 [`TriggeredDagRunOperator`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/trigger_dagrun/index.html).
-The new DAG will be named `<media_type>_filtered_index_creation`[^passive-name].
-It will not run on a schedule (only triggered), it will have a max concurrency
-of one, and it will accept the following configuration variables:
-
-[^passive-name]:
-    The passive voice name of this DAG is not ideal, but
-    `create_filtered_<model_name>_index_dag` reads as a DAG factory function
-    rather than a dag named "create filtered index".
+The new DAG will be named `create_filtered_<media_type>_index`. It will not run
+on a schedule (only triggered), it will have a max concurrency of one, and it
+will accept the following
+[Airflow params](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html):
 
 - `model_name`: The model being filtered. Used as the alias of the origin index
   to use.
@@ -299,25 +295,22 @@ data refresh is currently happening, but you want to create a new filtered
 index. You can theoretically kick off the reindex job from the existing index.
 However, after the data refresh is finished, the previous index will be deleted!
 This will prevent the reindex from completing successfully. Therefore, we cannot
-run filtered index creation if a data refresh is currently underway.
-Additionally, if filtered index creation is underway, then we cannot run a data
-refresh either. Even though the data refresh process takes considerable time
-before it's ready to delete the previous index, we don't want to play with race
-conditions.
+create a filtered index if a data refresh is currently underway. Additionally,
+if filtered index creation is underway, then we cannot run a data refresh
+either. Even though the data refresh process takes considerable time before it's
+ready to delete the previous index, we don't want to play with race conditions.
 
-Rather than trying to use a lock, we will read the DAG status directly from the
-DAG and fail the filtered index creation DAG run if the corresponding data
-refresh DAG is running, e.g.:
+Rather than trying to use a complex lock, we will read the data refresh DAG
+status at the start of the filtered index creation DAG run to check if the
+corresponding data refresh DAG is running, e.g.:
 
 ```py
-from data_refresh.dag_factory import image_data_refresh, audio_data_refresh
+from data_refresh.dag_factory import image_data_refresh
 
-def filtered_index_creation_dag():
-    # ...
-    if model_name == "image":
-        if image_data_refresh.get_active_runs():
-            # fail dag
-    # etc for other media types
+@task
+def prevent_concurrently_running_during_data_refresh():
+    if image_data_refresh.get_active_runs() and not :
+        # fail dag
 ```
 
 In the data refresh, however, rather than failing if the filtered index creation
@@ -342,20 +335,20 @@ making requests to the ingestion server. It must do the following:
    - The
      [data refresh DAG does this here](https://github.com/WordPress/openverse-catalog/blob/75990f356ed6d79601ecf3ecd96e48c3932acb47/openverse_catalog/dags/data_refresh/data_refresh_task_factory.py#L167-L174).
 1. Create a new suffix to use if one is not provided in the variables
-1. Call the `create_and_populate_filtered_index` action in the ingestion server
+1. Call the `CREATE_AND_POPULATE_FILTERED_INDEX` action in the ingestion server
    and wait for completion
    - See existing data refresh DAG strategy for making and waiting for ingestion
      server requests
-1. Call the `point_alias`[^no-promote-action] action in the ingestion server and
+1. Call the `POINT_ALIAS`[^no-promote-action] action in the ingestion server and
    wait for completion.
 1. Delete the previous filtered index whose canonical name we retrieved in the
    first step (`DELETE_INDEX` action)
 
 [^no-promote-action]:
-    We cannot use the `promote` action used by the regular data refresh flow
+    We cannot use the `PROMOT` action used by the regular data refresh flow
     because while that action does point the alias to the newly created index,
     it also promotes the API tables, which is not necessary for filtered index
-    creation. Therefore, we use the `point_alias` action which only points
+    creation. Therefore, we use the `POINT_ALIAS` action which only points
     aliases and nothing else.
 
 Because the image and audio data refreshes run concurrently, it is necessary for
@@ -367,9 +360,9 @@ and `audio_filtered_index_creation` DAGs.
 ### Query the filtered index (overview step 3)
 
 We must add a new boolean query parameter to the search query serialiser,
-`unstable__include_sensitive_results`. This parameter will default the disabled
-state. We will remove the `unstable` designation during the wrap-up of this
-implementation plan. This parameter should also reflect the state of the
+`unstable__include_sensitive_results`. This parameter will default to the
+disabled state. We will remove the `unstable` designation during the wrap-up of
+this implementation plan. This parameter should also reflect the state of the
 `mature` parameter: when `mature` is enabled, the new parameter should also be
 enabled. This prepares us to deprecate the `mature` parameter when we remove the
 `unstable` designation from the new parameter.
--- a/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
+++ b/documentation/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
@@ -1,0 +1,338 @@
+# Implementation Plan: Filtering and designating results with sensitive textual content
+
+## Reviewers
+
+- [ ] TBD
+- [ ] TBD
+
+## Project links
+
+[Project Thread](https://github.com/WordPress/openverse/issues/377)
+[Project Proposal](./20230309-project_proposal_detecting_sensitive_textual_content.md)
+[Milestone (pending)](#)
+[Implementation Plan: Managing A Sensitive Terms List](https://github.com/WordPress/openverse/pull/911)
+
+Please refer to additional prior art linked in the project thread. The most
+significant technical prior art is this exploratory PR for creating a filtered
+index: https://github.com/WordPress/openverse-api/pull/1108
+
+## Expected outcomes
+
+- Results are filtered efficiently based on a list of sensitive terms.
+- API results include a call-out on the result objects designating individual
+  results that have sensitive textual content.
+
+## Tools and dependencies
+
+We will not need any new tools on the API side for this work.
+
+## Feature flags
+
+We should introduce a single "feature flag" on the API as an undocumented and
+unstable query parameter `unstable__include_sensitive_results`.
+
+## API versioning
+
+There are no API versioning concerns. We already filter sensitive content based
+on provider data, we'll just be expanding upon that existing feature by
+incorporating a new signal into the filter. I do not believe this requires a new
+API version prefix or distinctly new parameter.
+
+## Areas touched
+
+Implementing this plan will require making changes in the following areas:
+
+- API
+- Ingestion server
+- Airflow data refresh DAG factory
+
+## Terms
+
+- "Origin index": This is the index we have now that includes all documents.
+- "Filtered index": This is the index we will have after this implementation
+  plan is implemented. Its documents are derived from the origin index but
+  exclude any that have sensitive terms in their textual content.
+
+## Overview of implementation process
+
+In depth descriptions of each of these steps are covered below. Each step will
+be implemented as a single pull request including appropriate tests.
+
+1. Update the ingestion server with a new action to use the reindex API to
+   generate a new index based on the origin index that only includes documents
+   that do not have sensitive terms in the three textual content fields
+   (`title`, `description`, `tags.name`). This is the filtered index.
+   - Requires the following:
+     - SENSITIVE_TERMS injected as new environment variable. See
+       [the related IP](https://github.com/WordPress/openverse/pull/911) for how
+       this will be implemented in Terraform.
+     - Ingestion server action to create the new index using `reindex` and an
+       update the promote action to allow it to promote (point alias) for the
+       filtered index
+     - Updates to the `load_sample_data.sh` script to generate the filtered
+       index
+     - A benign list of terms to use for local testing that matches results in
+       the sample data
+1. Update the data refresh DAG factory code to call the new filtered index
+   actions.
+1. Update the API with the new query parameter and query the filtered index when
+   not including sensitive results.
+1. Update the API to query the filtered index by document `_id` to deduce which
+   documents in a query that includes sensitive results do not have sensitive
+   terms. Use the resulting list of "safe" IDs to inverse the check and mark
+   documents that are sensitive. Simultaneously, update the media serialisers to
+   add the new `sensitivity` field. The serialiser field derives the array
+   entries based on the results of the previous step and the existing `mature`
+   field.
+
+Many thanks to @robfelty for helping me find a much simpler approach to
+sensitive result designation than I had originally devised. This solution is
+easily 10x more intelligible and simpler to document and implement than the
+others I had considered.
+
+## Technical description of plan
+
+Each heading refers to one of the steps in the previous section.
+
+### Ingestion server (overview step 1)
+
+To efficiently filter sensitive content, we will create a secondary "filtered"
+index using Elasticsearch's
+[`reindex` API](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-reindex.html).
+This API allows us to create a new index based on an existing one, while also
+applying a query to the existing index. In effect, we can create a large negated
+`multi_match` filter to exclude all results that contain any of the sensitive
+terms in their textual content. This will happen in the ingestion server as a
+separate step during data refresh.
+
+An example of the query built as a Python object iterating over a list of
+sensitive terms follows:
+
+```py
+query = {
+    "bool": {
+        "must_not": [
+            {
+                "multi_match": {
+                    "query": term,
+                    "fields": ["tags.name", "title", "description"],
+                }
+            }
+            for term in SENSITIVE_TERMS
+        ]
+    }
+}
+```
+
+The name of the filtered index will follow this pattern:
+`{source_index}-filtered`. The filtered index alias will follow the same
+pattern. This allows the API to easily query the filtered index, simply by
+appending `-filtered` to the name of the index that was already being filtered.
+For example, if the API is querying the filtered images index, it takes the
+already designated `image` index name and appends `-filtered`: `image-filtered`.
+
+The creation of the new index follow the same pattern of creating the regular
+index: create the index, wait for index creation to finish, then issue the
+command to point the alias to the new index. The ingestion server action to
+create the filtered index will be called `create_filtered_index`. The ingestion
+server action to promote the newly created filtered index will be called
+`promote_filtered_index`. These actions follow the same API as the existing
+`create` and `promote` actions.
+
+> **Note**
+>
+> There is a notable exception to the similarity of this process with the
+> regular index creation step. Specifically, the filtered index creation does
+> not need to push the documents to Elasticsearch using the indexer workers: ES
+> already has the documents in the "origin" index.
+
+### Airflow data refresh DAG factory (overview step 2)
+
+Adding these new steps to the Airflow data refresh DAG is simple and only
+requires adding the two new actions to the
+[existing `action_data_map`](https://github.com/WordPress/openverse-catalog/blob/4b5811aa240ba3ee2d92adea3a8abf88eb1e8aa4/openverse_catalog/dags/data_refresh/data_refresh_task_factory.py#L182)
+with the target alias also configured on the `promote_filtered_index` action,
+just as it is with the regular `promote` action.
+
+### Query the filtered index (overview step 3)
+
+We must add a new boolean query parameter to the search query serialiser,
+`unstable__include_sensitive_results`. This parameter will default to `False`.
+We will remove the `unstable` designation during the wrap-up of this
+implementation plan. This parameter should also reflect the state of the
+`mature` parameter: when `mature` is `True`, the new parameter should also be
+`True`. This prepares us to deprecate the `mature` parameter when we remove the
+`unstable` designation from the new parameter.
+
+When `unstable__include_sensitive_results` is `False` (default behaviour), query
+the filtered index by appending the `-filtered` suffix to the index name. Update
+the existing `mature` parameter handling to check
+`unstable__include_sensitive_results`. When the new parameter is `False`,
+results with the `mature` set to `true` should also be filtered out.
+
+When `unstable__include_sensitive_results` is `True`, query the origin index
+(existing behaviour).
+
+### Derive the list of sensitive document IDs based on the filtered index (overview step 4)
+
+When `unstable__include_sensitive_results` is `True`, we need to derive which
+documents in the results have sensitive textual content. To do this, we will
+rely on the fact that those documents are _not_ in the filtered index.
+Therefore, we can pull the list of `_id`s for the documents retrieved from the
+origin index and query for their presence in the filtered index. If they are in
+the filtered index, then we know they do not have sensitive textual content.
+Something along the following lines of pseudocode:
+
+```py
+results = query_origin_index()
+result_ids = set({r["_id"] for r in results})
+
+results_in_filtered_index = query_filtered_index(results_ids)
+ids_in_filtered_index = set({r["_id"] for r in results_in_filtered_index})
+
+# Use set arithmatic to derive the list of sensitive documents
+sensitive_text_result_ids = result_ids - ids_in_filtered_index
+```
+
+Add the resulting `sensitive_text_result_ids` set to the result serialiser
+context so that the media serialisers can reference them to derive the sensitive
+field:
+
+```py
+sensitivity = serializers.SerializerMethodField()
+
+def get_sensitivity(self, obj):
+    result = []
+    if obj["mature"]:
+        result.append("provider_supplied_mature")
+
+    if obj["identifier"] in self.context["sensitive_result_ids"]:
+        result.append("sensitive_text")
+
+    return result
+```
+
+#### Mature results
+
+Results with the `mature` field set to `true` will still be included in the
+"filtered" index. While it is easy to exclude these results from the filtered
+index, we cannot do so without making it harder to derive the list of results
+with sensitive textual content based on presence in the filtered index. This is
+because if the filtered index also excludes documents marked mature, then the
+`origin_result_ids - filtered_result_ids` will generate a set exclusive of
+documents marked mature _and_ those with sensitive textual content. Critically
+this means that results in the origin index that are mature but do not have
+sensitive text would get marked as having sensitive text based on their presence
+in the `sensitve_result_ids` set.
+
+The examples below are meant to illustrate this point as it's a bit of a
+slippery concept without concrete examples (at least that's what I found when
+writing this document).
+
+```
+origin_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+  {id: 2, mature: true}, # no sensitive text
+  {id: 3, mature: false}, # has sensitive text
+  {id: 4, mature: true}, # has sensitive text
+]
+
+filtered_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+]
+
+origin_index_result_ids - filtered_index_result_ids
+# => [1, 2, 3, 4] - [1] = [2, 3, 4]
+```
+
+Note how the final resulting list of IDs includes results marked mature, but
+that do not have sensitive textual content (result 2). Unfortunately there is no
+way around this as if we were to exclude the results marked `mature` from the
+origin result IDs in the final set subtraction, we'd end up with the following
+situation:
+
+```
+origin_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+  {id: 2, mature: true}, # no sensitive text
+  {id: 3, mature: false}, # has sensitive text
+  {id: 4, mature: true}, # has sensitive text
+]
+
+filtered_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+]
+
+origin_index_less_mature_result_ids = [r for r in origin_index results if not r["mature"]]
+# => [1, 3]
+
+origin_index_less_mature_result_ids - filtered_index_result_ids
+# => [1, 3] - [1] = [3]
+```
+
+Notice how we still have a problem: we've excluded the result that has sensitive
+text _and_ is marked mature (result 4).
+
+It's tempting to think we could move the mature filtering to _after_ when we've
+derived the list of sensitive result IDs. However, the problem will remain the
+same: we will exclude results that are both marked mature and have sensitive
+textual content:
+
+```
+origin_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+  {id: 2, mature: true}, # no sensitive text
+  {id: 3, mature: false}, # has sensitive text
+  {id: 4, mature: true}, # has sensitive text
+]
+
+filtered_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+]
+
+sensitive_result_ids = origin_index_result_ids - filtered_index_result_ids
+# => [1, 2, 3, 4] - [1] = [2, 3, 4]
+
+sensitive_results = [r for r in origin_index_results if r["_id"] in sensitive_result_ids and not r["mature"]]
+# => [3]
+```
+
+Result 4 is still excluded because we can't tell in the final operation whether
+result 4 was not in the set of IDs from the filtered index _only_ because it is
+marked mature or if it also has sensitive textual content.
+
+The only way I can think of to solve this is to have an index solely of
+documents that _do_ have sensitive textual content and querying it to derive the
+final list of sensitive results without needing to rely on the diversely
+filtered index. However, this adds a big re-indexing burden (yet another
+additional index) that can be bypassed altogether if we just exclude mature
+results from the filtered index to begin with.
+
+## Wrap up
+
+To wrap up the implementation, once the final step has been completed and
+results are filtered or correctly marked as sensitive, we will promote the
+`include_sensitive_results` parameter by removing the `unstable` designation.
+Additionally, we will deprecate the `mature` parameter by removing it from the
+documentation. To maintain API version backwards compatibility, it should
+continue to operate as an undocumented alias for the `include_sensitive_results`
+parameter.
+
+## Feature flags
+
+Because the new functionality will all be either internal (ingestion server and
+catalogue) or behind an unstable query parameter, we do not need any new feature
+flags to manage the rollout of this work.
+
+## Work streams
+
+While the ingestion server and catalogue change could be implemented at roughly
+the same time, it's probably useful to implement the ingestion server change
+first so that the catalogue change can actually be tested in the local
+environment with the new API functions. That means **there is a single work
+stream for this implementation plan**.
+
+The frontend implementation is _not_ entirely blocked by this plan as the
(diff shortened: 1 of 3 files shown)


rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+- [ ] TBD
+- [ ] TBD
+
+## Project links
+
+[Project Thread](https://github.com/WordPress/openverse/issues/377)
+[Project Proposal](./20230309-project_proposal_detecting_sensitive_textual_content.md)
+[Milestone (pending)](#)
+[Implementation Plan: Managing A Sensitive Terms List](https://github.com/WordPress/openverse/pull/911)
```
zackkrida, line 13:
```suggestion
- [Project Thread](https://github.com/WordPress/openverse/issues/377)
- [Project Proposal](./20230309-project_proposal_detecting_sensitive_textual_content.md)
- [Milestone (pending)](#)
- [Implementation Plan: Managing A Sensitive Terms List](https://github.com/WordPress/openverse/pull/911)
```

nit for readibility

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+Results with the `mature` field set to `true` will still be included in the
+"filtered" index. While it is easy to exclude these results from the filtered
+index, we cannot do so without making it harder to derive the list of results
+with sensitive textual content based on presence in the filtered index. This is
+because if the filtered index also excludes documents marked mature, then the
+`origin_result_ids - filtered_result_ids` will generate a set exclusive of
+documents marked mature _and_ those with sensitive textual content. Critically
+this means that results in the origin index that are mature but do not have
+sensitive text would get marked as having sensitive text based on their presence
+in the `sensitve_result_ids` set.
```
stacimc, line 346: Just double checking my understanding:  the goal is to make it possible to determine if an API result was marked sensitive due to `provider_supplied_mature` vs matching `sensitive_text`, but (at least for now) they will be treated exactly the same.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+In depth descriptions of each of these steps are covered below. Each step will
+be implemented as a single pull request including appropriate tests.
+
+1. Update the ingestion server with a new action to use the reindex API to
+   generate a new index based on the origin index that only includes documents
+   that do not have sensitive terms in the three textual content fields
+   (`title`, `description`, `tags.name`). This is the filtered index.
+   - Requires the following:
+     - SENSITIVE_TERMS injected as new environment variable. See
```
AetherUnbound, line 67: Nit:
```suggestion
     - `SENSITIVE_TERMS` injected as new environment variable. See
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+Additionally, we will deprecate the `mature` parameter by removing it from the
+documentation. To maintain API version backwards compatibility, it should
+continue to operate as an undocumented alias for the `include_sensitive_results`
+parameter.
+
+## Feature flags
+
+Because the new functionality will all be either internal (ingestion server and
+catalogue) or behind an unstable query parameter, we do not need any new feature
+flags to manage the rollout of this work.
```
stacimc, line 445: Since `include_sensitive_results` is by default False, even though the actual query parameter is unstable won't this be immediately rolled out in production once the first data refresh including the new index changes completes?

Do we have a planned way to roll back/query from the original index in case of some problem with the filtered index in production? Is that necessary?

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+   terms. Use the resulting list of "safe" IDs to inverse the check and mark
+   documents that are sensitive. Simultaneously, update the media serialisers to
+   add the new `sensitivity` field. The serialiser field derives the array
+   entries based on the results of the previous step and the existing `mature`
+   field.
+
+Many thanks to @robfelty for helping me find a much simpler approach to
+sensitive result designation than I had originally devised. This solution is
+easily 10x more intelligible and simpler to document and implement than the
+others I had considered.
```
AetherUnbound, line 92: Agreed, this makes a lot more sense on initial read to my Elasticsearch-novice experience!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+reindex API. While the reindex API can create new indexes (if the destination
+index specified does not exist), it does not support configuring the index in
+the same request.
+
+There are two approaches to ensuring the correct configuration of the filtered
+index:
+
+1. Create the new filtered index before calling reindex, using the
+   [`index_settings`](https://github.com/wordpress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/ingestion_server/ingestion_server/es_mapping.py#L1)
+   function to configure this index in the same was as the origin index for the
```
AetherUnbound, line 172:
```suggestion
   function to configure this index in the same way as the origin index for the
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+     will apply the template settings automatically.
+
+The first is my recommendation as it matches our current approach and is trivial
+to implement. The code sample at the end of this section demonstrates it in the
+call to `es.indices.create`.
+
+The second option, using index templates, is an interesting alternative that we
+could explore in the future, especially once we remove the ingestion server.
+However, it breaks significantly from our current approach to index
+configuration and I do not think is worth pursuing as part of this project.
```
AetherUnbound, line 186: Agreed!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+            },
+            "dest": {"index": destination_index},
+        },
+        slices="auto",
+        wait_for_completion=True,
+    )
+
+    self.refresh(index_name=destination_index, change_settings=True)
+
+    self.ping_callback()
```
AetherUnbound, line 229: So straightforward!! 😮

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+   optimal number of slices to use to parallelise the reindex.
+1. The source index name is not passed explicitly to the API: instead the method
+   must build it. This means we match the API of other calls (making updates to
+   the data refresh DAG much simpler) and can avoid needing to parse the model
+   name from the source index name for the call to `index_settings`.
+1. We explicitly refresh the index after it is finished reindexing to prime it
+   to be searched. ES only automatically refreshes indexes that are being
+   actively used, and the index should be refreshed before it starts being used.
+
+The creation of the new index follow the same pattern of creating the regular
```
AetherUnbound, line 247:
```suggestion
The creation of the new index follows the same pattern of creating the regular
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+point the alias to the new index. This process is encapsulated in two actions:
+the `create_and_populate_filtered_index` action described above, and the
+existing `point_alias` action.
+
+> **Note**
+>
+> While both processes have a "copy" step, they are notably different. The
+> origin index receives its documents from Postgres via ingestion workers, which
+> push the documents into Elasticsearch. The filtered index receives its
+> documents via the reindex API and does not require the ingestion workers.
```
AetherUnbound, line 258: Thanks for pointing out this distinction 🙂 We should make sure that the code path you describe above doesn't spin up the indexer worker instances since they won't be needed.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+### Derive the list of sensitive document IDs based on the filtered index (overview step 4)
+
+When `unstable__include_sensitive_results` is `True`, we need to derive which
+documents in the results have sensitive textual content. To do this, we will
+rely on the fact that those documents are _not_ in the filtered index.
+Therefore, we can pull the list of `_id`s for the documents retrieved from the
+origin index and query for their presence in the filtered index. If they are in
+the filtered index, then we know they do not have sensitive textual content.
+Something along the following lines of pseudocode:
```
AetherUnbound, line 304: This is very much a nit, but there's some oscillation between `True`/`true` and `False`/`false` here - is there any importance behind that difference? If not, would it make sense to make them consistent across the document?

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+```py
+action_data_map: dict[str, dict] = {
+    "ingest_upstream": {},
+    "promote": {"alias": target_alias},
+    "create_and_populate_filtered_index": {},
+    "point_alias": {"alias": f"{target_alias}-filtered"}
+}
+```
+
```
stacimc, line 276: Checking I understand:  any changes to the `SENSITIVE_TERMS` list (addition or removal of terms) will not be reflected in search results until the next data refresh, right?

I'm hoping the data refresh will become much more stable soon so this won't be a problem🤞, but if we needed to it sounds like we could also manually trigger just the relevant `create_and_populate_filtered_index` and `point_alias` steps on the ingestion server to refresh these results without a full data refresh.

Do you think it would be worth having the steps to create the filtered index in a separate DAG? We could then trigger the DAG from the data refresh using a [TriggerDagRunOperator](https://docs.astronomer.io/learn/cross-dag-dependencies#triggerdagrunoperator) so that it would continue to be an automatic part of standard data refreshes, but we could also run it independently. We _would_ need to make sure the new DAG can't be manually triggered while a data refresh is already running, which would require a bit of tweaking.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+documents marked mature _and_ those with sensitive textual content. Critically
+this means that results in the origin index that are mature but do not have
+sensitive text would get marked as having sensitive text based on their presence
+in the `sensitve_result_ids` set.
+
+The examples below are meant to illustrate this point as it's a bit of a
+slippery concept without concrete examples (at least that's what I found when
+writing this document).
+
+```
```
AetherUnbound, line 352:
```suggestion
```py
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+# => [1, 2, 3, 4] - [1] = [2, 3, 4]
+```
+
+Note how the final resulting list of IDs includes results marked mature, but
+that do not have sensitive textual content (result 2). Unfortunately there is no
+way around this as if we were to exclude the results marked `mature` from the
+origin result IDs in the final set subtraction, we'd end up with the following
+situation:
+
+```
```
AetherUnbound, line 374:
```suggestion
```py
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+Notice how we still have a problem: we've excluded the result that has sensitive
+text _and_ is marked mature (result 4).
+
+It's tempting to think we could move the mature filtering to _after_ when we've
+derived the list of sensitive result IDs. However, the problem will remain the
+same: we will exclude results that are both marked mature and have sensitive
+textual content:
+
+```
```
AetherUnbound, line 401:
```suggestion
```py
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+  {id: 2, mature: true}, # no sensitive text
+  {id: 3, mature: false}, # has sensitive text
+  {id: 4, mature: true}, # has sensitive text
+]
+
+filtered_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+]
+
+origin_index_result_ids - filtered_index_result_ids
```
AetherUnbound, line 364: Suggested change for clarity:
```suggestion
sensitive_result_ids = origin_index_result_ids - filtered_index_result_ids
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+]
+
+filtered_index_results = [
+  {id: 1, mature: false}, # no sensitive text
+]
+
+origin_index_less_mature_result_ids = [r for r in origin_index results if not r["mature"]]
+# => [1, 3]
+
+origin_index_less_mature_result_ids - filtered_index_result_ids
```
AetherUnbound, line 389: Suggested change for clarity:
```suggestion
sensitive_result_ids = origin_index_less_mature_result_ids - filtered_index_result_ids
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+Result 4 is still excluded because we can't tell in the final operation whether
+result 4 was not in the set of IDs from the filtered index _only_ because it is
+marked mature or if it also has sensitive textual content.
+
+The only way I can think of to solve this is to have an index solely of
+documents that _do_ have sensitive textual content and querying it to derive the
+final list of sensitive results without needing to rely on the diversely
+filtered index. However, this adds a big re-indexing burden (yet another
+additional index) that can be bypassed altogether if we just exclude mature
+results from the filtered index to begin with.
```
stacimc, line 429: Thank you for these examples! The problem is quite clear and your reasoning here makes sense to me.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
AetherUnbound, line 445: I might be misunderstanding this, doesn't this contradict the `# Feature Flags` section above?

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
AetherUnbound, line 429: Forgive me if I'm missing a piece here - could we exclude mature results from the filtered index as well when we're creating it? Would there be a need to include mature results that aren't sensitive separately, especially since we're going to be excluding those results anyway if the sensitive/mature results aren't requested to be included in the final results?

AetherUnbound (approved):
This is an excellent IP, thanks for the work you've done in exploring the options here! I know I'm not a reviewer but I was curious about how this would turn out on the Elasticsearch side so I gave it a look. Feel free to use my approval for the document or wait until the requested reviewers have a chance to review.

All of my commentary is either surface/syntax changes or questions, nothing blocking!

stacimc (commented):
This is fantastic work and a very interesting read :) Thanks very much for the code samples and examples, this was quite easy to follow even as a newcomer to Elasticsearch!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: What would be required to ensure that? Assuming we don't call the `schedule_distributed_index` function, it should be fine, right? I don't see anywhere else that would cause the workers to get spun up.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: > Forgive me if I'm missing a piece here - could we exclude mature results from the filtered index as well when we're creating it? Would there be a need to include mature results that aren't sensitive separately, especially since we're going to be excluding those results anyway if the sensitive/mature results aren't requested to be included in the final results?

As Staci said, the section deals entirely with the idea of excluding mature documents from the filtered index and describes why it's an impossible solution. @AetherUnbound, your inclination is precisely why I wrote the section this paragraph is under, because I also considered it a possibility and anticipated that others would have the same question. If the explanation is not clear on why it's not possible for us to exclude the mature results from the filtered index, please let me know so that I can revise the section.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+   - Requires the following:
+     - SENSITIVE_TERMS injected as new environment variable. See
+       [the related IP](https://github.com/WordPress/openverse/pull/911) for how
+       this will be implemented in Terraform.
+     - Ingestion server action to create the new index using `reindex` and an
+       update the promote action to allow it to promote (point alias) for the
+       filtered index
+     - Updates to the `load_sample_data.sh` script to generate the filtered
+       index
+     - A benign list of terms to use for local testing that matches results in
```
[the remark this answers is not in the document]
sarayourfriend: Yes, good point. Do you think we should maintain the testing list external from the application, rather than as the default value when reading the environment variable?

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+[existing `action_data_map`](https://github.com/WordPress/openverse-catalog/blob/4b5811aa240ba3ee2d92adea3a8abf88eb1e8aa4/openverse_catalog/dags/data_refresh/data_refresh_task_factory.py#L182)
+with the target alias configured on the `point_alias` action, just as it is with
+the regular `promote` action.
+
+```py
+action_data_map: dict[str, dict] = {
+    "ingest_upstream": {},
+    "promote": {"alias": target_alias},
+    "create_and_populate_filtered_index": {},
+    "point_alias": {"alias": f"{target_alias}-filtered"}
```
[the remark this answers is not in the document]
sarayourfriend: Yes! We can't use `promote` for that reason exactly, because it also promotes the API table, which isn't necessary for the filtered index.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,458 @@
...
+
+### Query the filtered index (overview step 3)
+
+We must add a new boolean query parameter to the search query serialiser,
+`unstable__include_sensitive_results`. This parameter will default to `False`.
+We will remove the `unstable` designation during the wrap-up of this
+implementation plan. This parameter should also reflect the state of the
+`mature` parameter: when `mature` is `True`, the new parameter should also be
+`True`. This prepares us to deprecate the `mature` parameter when we remove the
+`unstable` designation from the new parameter.
```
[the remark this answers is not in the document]
sarayourfriend: I hadn't considered this. Personally I'd rather 400 requests with conflicting parameters.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: > Do we have a planned way to roll back/query from the original index in case of some problem with the filtered index in production? Is that necessary?

That's a good point, I hadn't considered the mismatch there. Let's add a Django setting for enabling the new query parameter.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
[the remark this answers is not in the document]
stacimc: I think you're right, even better not to assume either way.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
stacimc: Ah, that makes a lot of sense! Thank you for that link.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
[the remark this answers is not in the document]
stacimc: 💯 Perfect! This is great, I love how easily this fits into the existing flow.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
AetherUnbound: I think that's it! Awesome

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
AetherUnbound: Admittedly I read this section several times and still came away a bit confused 😅 I think the premise was getting to me - my thought was that it doesn't matter for the API whether the `mature` field is `True` or whether we filtered it for sensitive text reasons, the results are excluded all the same. So _in my mind_ making the distinction between `mature` and `sensitive` was unnecessary, but it seems that it might be in some cases which would require the explanation you put out Sara. I guess the core of my question was whether that distinction was important, although you've [answered my question in a later comment](https://github.com/WordPress/openverse/pull/996#discussion_r1162130216)!

I perhaps was getting confused by the distinction between the `mature` API parameter (which we are saying is functionally equivalent to `sensitive` after the implementation of this plan) and the `mature` _field_ on the API (which will stick around in perpetuity) 🤯 I believe I'm all cleared up now!

sarayourfriend: > I perhaps was getting confused by the distinction between the mature API parameter (which we are saying is functionally equivalent to sensitive after the implementation of this plan) and the mature field on the API (which will stick around in perpetuity) exploding_head I believe I'm all cleared up now!

This will be confusing for a while, unfortunately :disappointed:. Hopefully decoupling the parameter name from the various things that it includes will help (i.e., using the broader "sensitive" term to capture "mature" results as well as other types of sensitivity, textual, and machine vision derived).

> So in my mind making the distinction between mature and sensitive was unnecessary, but it seems that it might be in some cases which would require the explanation you put out Sara

I can add another note about this in the section referencing the frontend. The existence of the `sensitivity` field as an array that notes the distinction hopefully communicates the necessity of knowing the difference between them. The reason for that need isn't explicit in this document because the project proposal covers the reason we are making this distinction.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
stacimc: Exactly, I remembered talking about a one-week timeframe being acceptable at some point, but the instability of the data refresh might change that 😞 

> To prevent them running at the same time, is it a matter of adding some sort of lock in Airflow variables? I vaguely recall you'd done some work on another DAG to prevent this as well.

Yes, the data refresh DAGs themselves use a `data_refresh` pool with one slot and a custom `SingleRunExternalDAGSensor` to make sure that only one data refresh DAG runs at a time. We could add these new DAGs to the list of "data refresh DAGs", but that wouldn't be quite right by default -- because the 'parent' DAG will still be running when the filtered index creation DAG starts.

I think the `TriggerDagRunOperator` will allow us to [pass in conf options](https://registry.astronomer.io/providers/apache-airflow/versions/latest/modules/triggerdagrunoperator) when we trigger the filtered index creation DAG, so we should be able to pass a "force" parameter like you suggested. Then we would probably want a new, simple sensor that would `wait` as long as the data refresh is running, but also accept the "force" parameter as an override.

There might be a simpler way to do it, which I can look into, but I think that ought to work!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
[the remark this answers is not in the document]
stacimc: +1, I think that would be the safest approach. Great call out!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: > Then we would probably want a new, simple sensor that would wait as long as the data refresh is running

The data refresh takes so long: would it make more sense for the dag to fail if called during the data refresh rather than waiting? If the only time that kind of error can happen is if we manually call it (this dag itself won't be scheduled, right?) then waiting would be more confusing to me rather than an error that says "I can't do this while the data refresh is going".

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: Whoops! I'm going to fix this in light of https://github.com/WordPress/openverse/pull/996#discussion_r1162007337 as well.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: No reason other than missing the shift bar a couple times :slightly_smiling_face: I'll consolidate to using the Python title cased version.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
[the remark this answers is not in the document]
sarayourfriend: To clarify, you think that we should _not_ just use the default value for the environment variable?

If we don't then we'd need to update `env.docker` for the ingestion server and copy those to the ingestion server settings.

If we set the default in code then it is (in my opinion) slightly easier to understand in testing. We can restrict the default to only apply not in production:

```py
SENSITIVE_TERMS = config(
    "SENSITIVE_TERMS",
    cast=lambda x: x.split("\n"),
    default=( # just an example list
        """
        water
        dogs
        cactus
        """
        if ENVIRONMENT != "production" else None
    )
)
```

We can raise an error in `create_and_populate_filtered_index` if `SENSITIVE_TERMS` is None.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
stacimc: For sure, we could definitely fail the DAG instead. Now you mention it, the data refresh will include the filtered index creation step as is... so there shouldn't theoretically be any reason to wait and re-do it after the data refresh finishes, anyway. That sounds great 👍

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
dhruvkb, line 67: This is a very small detail but I'm wondering if a file stored in S3 and it's URL in the environment variable would be a better option that loading the entire list of sensitive terms from an environment variable. It's not an immediate issue but if we include sensitive terms from multiple languages, the list might be several megabytes which might exceed some character and size limits.

dhruvkb (approved):
LGTM! I have one concern about the use of env vars to store the terms list but everything else about the implementation seems well-thought out and nicely planned to me.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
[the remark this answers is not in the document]
stacimc: The addition of the environment check works just fine and I agree, easier to understand. Your example here is totally fine with me.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: I had thought about doing it that way as well, and I don't remember why I landed on using an environment variable. Alternatively, we could even request the list directly from GitHub, as it will be a public repository storing them.

The difficulty this presents is for testing data that doesn't need to be stored in a public repository. We could either have an `inane-terms.txt` in the repository for testing or use `SENSITIVE_TERMS_URL` as the primary and `SENSITIVE_TERMS` as a fallback that would only be used in non-production environments.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: Ah, actually, the difference between `true` and `True` here _was_ intentional! In Elasticsearch the "mature" property is a JSON field, so the value is indeed `true` with a lowercase T. I'm still changing the language to not reference the specific values, but I just wanted to note that there is indeed a small difference here due to the data representation model that Elasticsearch is actually using.

sarayourfriend:
@AetherUnbound Thank you for the editorial suggestions. I've applied them in https://github.com/WordPress/openverse/pull/996/commits/830f9a747b595362880bcdf117ae8965268a991f. Rather than noting that in each suggestion, I've gone ahead and just resolved the conversations, but I wanted to make sure you knew I was grateful for your suggestions and did apply them. I've also added the clarifications we discussed.

@stacimc I've made the changes we talked about to the feature flags, conflicting parameter handling, and the Airflow DAG configurations. @AetherUnbound I would appreciate you careful review of the DAG situation as well :pray: 

@dhruvkb I made the change you suggested to use a URL instead of putting the list in the environment variables. It's a much more flexible solution that just generally makes more sense. I also added instructions for how to handle this in staging and local environments by using a locally served file from Falcon itself. This seemed like the easiest approach without adding too much in the code for handling fallbacks of different types of settings for different environments.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,581 @@
...
+DAGs using the
+[`TriggeredDagRunOperator`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/trigger_dagrun/index.html).
+The new DAG will be named `<media_type>_filtered_index_creation`[^passive-name].
+It will not run on a schedule (only triggered), it will have a max concurrency
+of one, and it will accept the following configuration variables:
+
+[^passive-name]:
+    The passive voice name of this DAG is not ideal, but
+    `create_filtered_<model_name>_index_dag` reads as a DAG factory function
+    rather than a dag named "create filtered index".
```
AetherUnbound, line 288: We don't need `_dag` in the suffix, do you think `create_filtered_<model_name>_index` works? That sounds fine IMO

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,581 @@
...
+the filtered index creation DAG to also be a DAG factory, in the same style as
+the data refresh DAG factory. This will result in a new DAG per media type. With
+our current media types, that means we will add `image_filtered_index_creation`
+and `audio_filtered_index_creation` DAGs.
+
+### Query the filtered index (overview step 3)
+
+We must add a new boolean query parameter to the search query serialiser,
+`unstable__include_sensitive_results`. This parameter will default the disabled
+state. We will remove the `unstable` designation during the wrap-up of this
```
AetherUnbound, line 371:
```suggestion
`unstable__include_sensitive_results`. This parameter will default to the disabled
state. We will remove the `unstable` designation during the wrap-up of this
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,581 @@
...
+However, after the data refresh is finished, the previous index will be deleted!
+This will prevent the reindex from completing successfully. Therefore, we cannot
+run filtered index creation if a data refresh is currently underway.
+Additionally, if filtered index creation is underway, then we cannot run a data
+refresh either. Even though the data refresh process takes considerable time
+before it's ready to delete the previous index, we don't want to play with race
+conditions.
+
+Rather than trying to use a lock, we will read the DAG status directly from the
+DAG and fail the filtered index creation DAG run if the corresponding data
```
AetherUnbound, line 309: Can you clarify which DAG `the DAG` is here? I think it's the data refresh DAG but clarity would be helpful:
```suggestion
Rather than trying to use a lock, we will read the DAG status directly from the
data refresh DAG and fail the filtered index creation DAG run if the corresponding data
```

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,581 @@
...
+   wait for completion.
+1. Delete the previous filtered index whose canonical name we retrieved in the
+   first step (`DELETE_INDEX` action)
+
+[^no-promote-action]:
+    We cannot use the `promote` action used by the regular data refresh flow
+    because while that action does point the alias to the newly created index,
+    it also promotes the API tables, which is not necessary for filtered index
+    creation. Therefore, we use the `point_alias` action which only points
+    aliases and nothing else.
```
AetherUnbound, line 359: Forgive my pedantry here, but should the various actions all be capitalized? The mixed case is throwing me off in thinking that they might be referring to different things:
```suggestion
1. Call the `POINT_ALIAS`[^no-promote-action] action in the ingestion server and
   wait for completion.
1. Delete the previous filtered index whose canonical name we retrieved in the
   first step (`DELETE_INDEX` action)

[^no-promote-action]:
    We cannot use the `PROMOTE` action used by the regular data refresh flow
    because while that action does point the alias to the newly created index,
    it also promotes the API tables, which is not necessary for filtered index
    creation. Therefore, we use the `POINT_ALIAS` action which only points
    aliases and nothing else.
```

AetherUnbound (approved):
Thanks for making the changes suggested 🙂 I have a few more points of clarification specifically around the changes. No blockers, this looks good to me 👍🏼

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: Great! I will remove this then :sweat_smile:

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: I think the wording is confusing. I'll change it to "we will read the data refresh DAG status at the start of the create filtered index DAG" to remove the ambiguity.

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
sarayourfriend: For clarity yes. These actions correspond 1-to-1 with the `TableIndexer` methods, but the API does ask for them in caps.

sarayourfriend:
Thanks @AetherUnbound. Address your comments in https://github.com/WordPress/openverse/pull/996/commits/a1fad8c575d91ba6fc58c7e6608c78804cbcf137

stacimc (approved):
Fantastic! +1 for injecting the `SENSITIVE_TERMS_URL` as an environment variable. The updates around the DAG all look great to me and document the concurrency problems very clearly.

This looks great, @sarayourfriend!

rfcs/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.md
```diff
@@ -0,0 +1,574 @@
+# Implementation Plan: Filtering and designating results with sensitive textual content
+
+## Reviewers
+
+- [ ] TBD
+- [ ] TBD
```
sarayourfriend, line 6:
```suggestion
- [x] @stacimc
- [x] @dhruvkb
```

PR #997, opened 2023-03-23
Safely call create-or-update-comment when dealing with forks

### LINKED ISSUE #853: Cannot comment about Playwright on PRs from forks
## Description

The action that comments about Playwright failures does not work on PRs from forks. This is a [known limitation](https://github.com/peter-evans/create-or-update-comment#action-inputs) and the workaround (using `pull_request_target`) is not acceptable because we want the tests to run in the context of the PR and not `main`.

## Reproduction

See this [failed run](https://github.com/WordPress/openverse/actions/runs/4330608751/jobs/7591849919).

## Solution

It would seem that the only solution is to print the debugging message in the failed Playwright job itself and skip the comment job for PRs from forks.

sarayourfriend:
## Fixes

Fixes #853 by @dhruvkb 

## Description

I'm not sure if my solution is entirely the best one. I like that it avoids us needing to copy the body text between two separate actions in the workflow, but is it more complicated to have an additional layer sitting between the call to the upstream workflow? If it was normal code I'd say not (we wrap method calls to external APIs all the time to handle our own particularities) but with GitHub actions, it's so many more layers of redirection.

What do y'all think? Should I change this to just inline the conditional on the playwright action only and duplicate the body text?

## Testing Instructions

See if everything works on this PR: frontend size check, playwright run check.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal (approved):
Looks good!

.github/actions/create-or-update-comment/action.yml
```diff
@@ -0,0 +1,36 @@
...
+  issue-number:
+    default: null
+
+runs:
+  using: "composite"
+  steps:
+    - name: Call upstream action
+      # Not fork
+      if: github.repository == 'WordPress/openverse'
+      uses: peter-evans/create-or-update-comment@v2
```
AetherUnbound, line 22: Something I'm not sure of - does/would dependabot update versions within composite actions? I'm trying to find an answer now.

AetherUnbound: Update, I found this link: https://stackoverflow.com/a/74997644/3277713

TL;DR is that we have to add explicit dependabot checks for all our composite action directories 😞

sarayourfriend: Good call, easy enough to add.

dhruvkb:
With the changes in `main`, there are only two usages of the `peter-evans/create-or-update-comment` action:

- `playwright-test-failure-comment`
- `emit-docs`

Out of these, the job `emit-docs` does not run on PRs from forks (and from dependabot).

```js
(
  (
    github.event_name == 'push' &&
    github.repository == 'WordPress/openverse'
  ) ||
  (
    github.event_name == 'pull_request' &&
    github.event.pull_request.head.repo.owner.login == 'WordPress' &&
    github.actor != 'dependabot[bot]'
  )
)
```

so the change will be to expand the `if` condition in `playwright-test-failure-comment.steps[2]` where the action is used, greatly simplifying the PR.

sarayourfriend:
Thanks, @dhruv! I will implement that in the next commit.

sarayourfriend:
@dhruvkb I changed the PR as you suggested, but I am wondering if you have any clever alternatives to putting the help body in the separate step? I wanted to avoid duplicating it, but maybe it's better to just copy/paste it...

zackkrida:
@dhruvkb just following up; @sarayourfriend had a question [here](https://github.com/WordPress/openverse/pull/997#issuecomment-1489459887) and I'm not sure if you've seen it.

dhruvkb (changes requested):
I see no harm in having the message logged every time, so we can drop the 'Log help body' step, and just `echo` to `STDOUT` in the 'Build help body' step itself. That can further simplify the code.

dhruvkb (approved):

PR #1028, opened 2023-03-25
Absorb `actionlint` into pre-commit

### LINKED ISSUE #896: Absorb `actionlint` into pre-commit
## Problem

Currently `actionlint` is a separate workflow that runs on `push` (to the `main` branch) and `pull_request`.

https://github.com/WordPress/openverse/blob/38af21c359703d3faee9e160f1ac69372661d6a2/.github/workflows/actionlint.yml

This is a problem because it reports issues after they have been pushed instead of locally before pushing.

## Description

`actionlint` offers [`pre-commit` hooks](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#pre-commit).

This means that we can
- run the checks locally before committing and pushing
- get rid of the separate workflow and run the check as a part of the Lint job inside the CI + CD workflow.

AdarshRawat1:
## Fixes

Fixes #896 by @dhruvkb 
## Description

**The current setup** of running 'actionlint' as a separate workflow on push and pull requests can be problematic as it only reports issues after they have been pushed, which can be time-consuming and potentially cause problems for the development process.

**This PR** integrates `actionlint` into the pre-commit process. 

**Effects -** 

- developers can run checks locally before committing and pushing, allowing them to catch any issues and fix them before they get pushed to the main branch. 
 
-  eliminates the need for a separate workflow and instead incorporates the check as part of the Lint job inside the CI + CD workflow.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


templates/.pre-commit-config.yaml.jinja
```diff
@@ -81,6 +81,11 @@ repos:
           # See the following documentation for what each rule does:
           # https://www.pydocstyle.org/en/6.2.3/error_codes.html#error-codes
           - --add-ignore=D1,D202,D205,D211,D400
+          
+  - repo: https://github.com/rhysd/actionlint
+    rev: main
+    hooks:
+      - id: actionlint
```
dhruvkb, line 88: - You have some trailing whitespace in this line. That is CI to fail.
- Since we cannot rely on `actionlint` executable to be present, the Docker-based approach makes more sense.
- We also want it to only run when the `.github/worfklows` or `.github/actions` folder has changed, so we can set the [`files` field](https://pre-commit.com/#config-files) to limit it to those folders.

```suggestion
      - id: actionlint-docker
        files: ^\.github/(workflows|actions)/.*$
```

dhruvkb (changes requested):
@AdarshRawat1 thanks for this PR, it is in the right direction but I do have a few changes for you.

templates/.pre-commit-config.yaml.jinja
AdarshRawat1: Thank you, @dhruvkb I have committed the suggested changes.

.pre-commit-config.yaml
```diff
@@ -129,6 +129,11 @@ repos:
     hooks:
       - id: shellcheck
 
+  - repo: https://github.com/rhysd/actionlint
+    rev: main
+    hooks:
+      - id: actionlint
```
dhruvkb, line 135: Please update this to match `.pre-commit-config.yaml.jinja`.

dhruvkb (approved):
Awesome, only one more step left to go!

dhruvkb:
Weirdly `actionlint` is reporting our custom actions from `.github/actions` as incorrect. Actions have a different syntax compared to workflows and I expected `actionlint` to be aware of that. Maybe someone from @WordPress/openverse-maintainers might have any idea.

sarayourfriend:
Actionlint can't lint actions, only workflows: https://github.com/rhysd/actionlint/issues/46

I think the issue is the `files` directive. The default files directive for actionlint already scopes it to `.github/workflows`. Adding the actions directory as well includes files that actionlint doesn't know what to do with. I think removing the files directive altogether is fine and just relying on the default: https://github.com/rhysd/actionlint/blob/main/.pre-commit-hooks.yaml#L14

dhruvkb:
> Actionlint can't lint actions, only workflows (TIL): rhysd/actionlint#46

Wow. Pretty misleading name then!

I agree, dropping `.github/actions/` from the linted files should suffice. That would mean deleting the `files` field altogether because we want to use the default value.

https://github.com/rhysd/actionlint/blob/5562294a5537e12c41501f4b0600f9633c2043cd/.pre-commit-hooks.yaml#L14

sarayourfriend (approved):
LGMT. @dhruvkb I'll open a PR to remove the `actionlint` required check from the repo configuration now that it's folded into pre-commit.

dhruvkb (approved):
This is nice.

sarayourfriend:
Rebasing to get the checks to re-run and update with the modified required checks list which now excludes the no-longer existent `actionlint` workflow.

PR #1036, opened 2023-03-27
Do not bind ports with docker-compose in CI

### LINKED ISSUE #1035: Consider disabling exposed ports in CI to avoid port conflict flakiness
## Problem

CI jobs sometimes fail due to port conflicts.

cf #990 and #200

## Description

One way to avoid this is to not bind ports at all in CI. At a glance, I don't think we need to bind ports _at all_ in CI as we _never_ make requests to running containers in CI: everything happens _inside_ the containers and we just check the output.

To implement this we'd need to remove all port declarations from the base `docker-compose.yml` file and add them in a `docker-compose.development.yml` as overrides instead. See this comment: https://github.com/docker/compose/issues/3729#issuecomment-232473164

This is _not_ feasible to do in a single PR and will require multiple discrete PRs of smaller, reviewable chunks of code to accomplish. I tried this in #1036 and learned the hard way that this is actually a very big change. Here is my rough outline of how I would approach this in discrete PRs:

1. Refactor `load_sample_data.sh` to run inside the docker-compose networking context.
    - Has benefits aside from this project, for example, new contributors do not have to debug environment issues with the script
    - Rather than refactoring the script directly, we could also change it to be a management command. This would make certain things a lot easier, like tracking ingestion server task status. This is a bigger task to review though.
    - Do not yet remove the just recipes used for waiting for ES, etc, as the ingestion server tests also need them. We can go ahead and remove the just recipes used for making ingestion server requests, provided no one uses them outside the load sample data script.
2. Update the API to use docker networking rather than host networking
    - Mostly involves updating references to `localhost` to reference the docker networking name for the service. e.g., `localhost:<es-port>` becomes `es:9200` for the Elasticsearch connection. Etc for Postgres, Redis, and so on.
3. Update the ingestion server to run inside the docker networking context. Do not modify the tests yet though.
4. Get ingestion server tests running without host-bound ports. There are two ways to do this. An "easy" way and a very hard way.
    - Easy way: Run the tests inside a [`docker/compose`](https://hub.docker.com/r/docker/compose/) container. This would spin up the ingestion server tests inside the docker-compose container and not require any host bound ports. The downside here is that we introduce yet another docker image to download to run tests. It may be easier to copy the compose binary into one of the Python image we already use via the [`docker/compose-bin`](https://hub.docker.com/r/docker/compose-bin/) image. This at least prevents us from needing to download the entire OS for the compose image and just reuse one we already have for existing services.
    - Hard way: Change the way the tests are run to not rely on host-bound networking. The reason this is hard is that the tests currently rely on host bound networking for making necessary schema and data changes to the support databases. These can be done through `dc exec` or `dc run` even, but it's a whole heap of work to change this stuff.

There are probably other things that need to be taken into consideration as well, like how the ingestion server gets deployed.

## Alternatives

Do the retry approach that @dhruvkb implemented in #990. This could work but is potentially still flaky and adds complexity to the CI.

sarayourfriend:
## Fixes

Fixes #1035 by @sarayourfriend 

Follow-up to #990 

This is not a viable replacement for #990 as it requires further work to make `load_sample_data.sh` work in CI without bound ports.

cc @dhruvkb

## Description

As described in the issue, the idea here is that we shouldn't need to bind any ports in CI as no part of our CI environment makes requests from the host into the running containers. 

Unfortunately this PR ended up being a bit bigger than I anticipated, primarily due to the fact that I hadn't considered how much we actually rely on host-bound networking. In particular, the `load_sample_data.sh` script needed to change to be able to run inside the context of the `web` container. It needs to run there so that it actually has access to the docker network. Once inside, it no longer has docker-compose, so everything relying on docker-compose needed to be changed. Additionally, because `just` is not available in the web container (and I couldn't justify installing it), everything that previously called out to a just script needed to be inlined into the script. I put these in bash functions when it made sense to do so. I also made changes to the way we waited for task completion by leveraging the ingestion server's task status endpoint rather than trying to infer the state based on the ES index status (which I found was flaky). The `_cat` endpoints are useful, but they're explicitly not for machine reading, according to the ES documentation, so I also removed the usages of those endpoints as much as possible.

It ended up being more changes than I wanted, and I could definitely pair down a number of the changes by moving things like the `wait-for-index` just script directly into the load_sample_data.sh. If folks feel that would be easier to review and would increase confidence in the changes, just let me know, I'm more than happy to do it.

## Testing Instructions

`just down -v` (to clean your local environment) then `just up` and `just init` and make queries to local DB to ensure it all works. This will confirm that the local approach (with host-bound networking) still works. `just recreate` should also still work.

Next, to try the CI version. In addition to it passing in the CI on this PR, you should also be able to do the following locally (after `just down -v`):

```
$ CI=1 just up
$ CI=1 just api/init
$ CI=1 just curl '-H "Accept: application/json" http://web:8000/v1/images/'
```

The final command should yield a JSON blob with the same results that would have shown at localhost in the first pass.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/settings.py
+++ b/api/catalog/settings.py
@@ -62,6 +62,7 @@ if DEBUG:
         "localhost",
         "127.0.0.1",
         "0.0.0.0",
+        "web",  # used when requesting from within the docker network context
     ]
 
 USE_S3 = config("USE_S3", default=False, cast=bool)
--- a/api/justfile
+++ b/api/justfile
@@ -23,12 +23,12 @@ up *flags:
     env COMPOSE_PROFILES="api" just ../up {{ flags }}
 
 # Wait for all profile services to be up
-wait-up: up
+@wait-up: up
     just ../ingestion_server/wait # API profile includes ingestion server
     just wait # API waits for ES in entrypoint
 
 # Load sample data into API via ingestion server
-init: wait-up
+@init: wait-up
     cd .. && just exec "-T web /bin/bash < load_sample_data.sh"
 
 recreate:
@@ -41,27 +41,23 @@ recreate:
 ##########
 
 # Check the health of the service
-@health host="localhost:50280":
-    -curl -s -o /dev/null -w '%{http_code}' 'http://{{ host }}/healthcheck/'
+@health host="web:8000":
+    just ../curl "-s -o /dev/null -w %{http_code} 'http://{{ host }}/healthcheck/"
 
 # Wait for the service to be healthy
-@wait host="localhost:50280":
-    # The just command on the second line is executed in the context of the
-    # parent directory and so must be prefixed with `api/`.
-    just ../_loop \
-    '"$(just api/health {{ host }})" != "200"' \
-    "Waiting for the API to be healthy..."
+@wait host="web:8000":
+    just ../curl "--retry-connrefused --retry 5 --retry-delay 5 http://{{ host }}/healthcheck/"
 
 ########
 # cURL #
 ########
 
 # Make a cURL GET request to service at the given path
-_curl-get path host="localhost:50280":
-    curl "http://{{ host }}/v1/{{ path }}"
+_curl-get path host="web:8000":
+    just ../curl "http://{{ host }}/v1/{{ path }}"
 
 # Make a test cURL GET request to the API
-stats media="images" host="localhost:50280":
+stats media="images" host="web:8000":
     just _curl-get "{{ media }}/stats/" {{ host }}
 
 # Launch a `pgcli` shell in the web container (requires typing credentials)
--- a/api/justfile
+++ b/api/justfile
@@ -29,7 +29,7 @@ wait-up: up
 
 # Load sample data into API via ingestion server
 init: wait-up
-    cd .. && ./load_sample_data.sh
+    cd .. && just exec web load_sample_data.sh
 
 recreate:
     just ../down -v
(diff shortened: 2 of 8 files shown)


sarayourfriend:
I'm not sure what is going on with the API tests job's init failure. When I run `CI=1 just api/test` locally (after `api/init`, also with CI=1) it works fine. I'll keep digging tomorrow if no one else comes up with a solution in the meantime.

sarayourfriend:
Okay, it turns out that the ingestion server tests are a huuuge rabbit hole for this. Even if I am able to figure out what is going on with the API tests (which I haven't looked at today at all), the ingestion server tests will need a lot of changes as they spin up their own docker-compose stack and rely heavily on host-bound networking. It's definitely possible to do, but not in a single PR.

I'm going to set this PR back to draft and decide how best to split it up and then we can decide whether it's worth tackling or if we're okay with just retrying things long-term as in #990. I still think it's worth doing this _at some point_, but having now spent two and a half days working on it, I don't think it is the correct use of our time _right now_. I still want to write up the rough outline of how to tackle this in the future as I've discovered a lot of hidden-ish details that cause hiccups here and there with the process.

PR #1056, opened 2023-03-28
Remove `ARG API_URL` instruction from frontend/Dockerfile

krysal:
## Fixes

Related to WordPress/openverse-infrastructure#431

## Description

This argument may be in the way of us setting the API URL to use at running time.

## Testing Instructions

The image build is not working for me locally but the CI/CD workflows seem to indicate that it works. You could try building the image, or let's merge this and redeploy staging FE.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -39,7 +39,6 @@ RUN pnpm install --offline
 ENV NUXT_TELEMETRY_DISABLED=1
 ENV NODE_ENV=production
 
-ARG API_URL
 ARG SEMANTIC_VERSION
 
 RUN echo "{\"release\":\"${SEMANTIC_VERSION}\"}" > /home/node/frontend/src/static/version.json


sarayourfriend:
@krysal Is the thought process here that the ARG is overwriting what we pass at runtime? According to https://vsupalov.com/docker-arg-env-variable-guide/#arg-and-env-availability, `ARG`s are only available during the docker build and won't be around when the entrypoint or command run in the container.

> [`ARG`s] are only available from the moment they are ‘announced’ in the Dockerfile with an ARG instruction up to the moment when the image is built. Running containers can’t access values of ARG variables. This also applies to CMD and ENTRYPOINT instructions which just tell what the container should run by default.

Removing it won't fix the actual issue, which is that the value for `API_URL` gets inlined during build, because it is part of `env` config: https://nuxtjs.org/docs/directory-structure/nuxt-config#env

You can see this locally if you build the app with a command like this:

```
API_URL=http://example.com pnpm run build:only
```

And then `cd .nuxt` and grep for `example.com`. It'll appear in these files (hashes may differ):

```
utils.js
index.js
dist/server/server.js
dist/client/2672bcb.js
dist/client/8f74f12.modern.js
```

In `utils.js`, for example, you can see it is inlined:

```
➜  .nuxt git:(fix/rm_arg) cat utils.js | grep fake-domain
      env: {"apiUrl":"http://example.com/","filterStorageKey":"openverse-filter-visibility","savedSearchCount":"4","providerUpdateFrequency":"3600000"}
```

I think the correct way to fix this is to make API_URL available at runtime, is to put it in `publicRuntimeConfig`: https://nuxtjs.org/docs/directory-structure/nuxt-config#publicruntimeconfig. I believe either you or Dhruv mentioned that this might be the case. In testing locally I can confirm it almost definitely is the case.

If I grep for `process.env.API_URL` in the built app, I find only one place where it is being read at runtime, in `dist/server/server.js`:

```js
const apiUrl = (_a = process.env.API_URL) !== null && _a !== void 0 ? _a : "https://api.openverse.engineering/";
/**
 * Default environment variables are set on this key. Defaults are fallbacks to existing env vars.
 * All boolean values should be designed to be false by default.
 * providerUpdateFrequency - how often we should re-fetch provider statistics.
 */
const env = {
  apiUrl: apiUrl.endsWith("/") ? apiUrl : `${apiUrl}/`,
  filterStorageKey: "openverse-filter-visibility",
  savedSearchCount: "4",
  providerUpdateFrequency: `${60 * 60 * 1000}` // 1 hour
};
```

That code makes it look like you could overwrite it at runtime, but nothing actually reads that `apiUrl` value. Everything else has it inlined. There are no references to `.apiUrl` or `[("|')apiUrl("|')]` when I grepped the build locally.

krysal:
Thanks a lot @sarayourfriend! Your explanation is quite clarifying of what is going on here. I seems that the `publicRuntimeConfig` suggested by @dhruvkb is indeed the best place to store the `API_URL` and being able to change it at runtime. Looks like it will require some major changes of how that value is used by the different `*-service.ts` files. I didn't have time to look at it but I'll at my come back from AFK if nobody takes it.

I'm converting this to a draft in the meantime. If anyone want to give it a try here or in another PR please go ahead.

PR #1057, opened 2023-03-29
Use context manager for multiprocessing in the ingestion server

obulat:
Follow up on #904

## Description

This PR adds a context manager to the multiprocessing pool in the ingestion server. I originally added it in #904, but without `.join()` it was causing the tests to fail.

I am not sure about the placement of `results` saving: putting it after the call to `join` might mean that there is too much data in memory (all of the data that has been updated from several child processes), but if we handle the results as they come, will there be a racing condition with the files?

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
Based on [this guide](https://superfastpython.com/multiprocessing-pool-context-manager/#Approach_3_Context_Manager) and [the multiprocessing docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool), I think the flow and `close`/`join` calls you have set up now are correct! I would have assumed that using the pool as a context manager would handle everything, but it seems like the `__exit__` method only calls `terminate` (not `close` or `join`). It also seems like `starmap` will block until all tasks are complete. This is in-line with the previous implementation, so if we haven't been running into memory issues already we should be good!

AetherUnbound (approved):
Tested on my machine, works great!

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -351,16 +351,15 @@ def clean_image_data(table):
...
-
-            for result in results:
-                batch_cleaned_counts = save_cleaned_data(result)
-                for field in batch_cleaned_counts:
-                    cleaned_counts_by_field[field] += batch_cleaned_counts[field]
-            pool.close()
+            with multiprocessing.Pool(processes=num_workers) as pool:
+                log.info(f"Starting {len(jobs)} cleaning jobs")
+
+                for result in pool.starmap(_clean_data_worker, jobs):
```
dhruvkb, line 357: Since `pool.starmap` is blocking (like `pool.map`), as per this [StackOverflow answer](https://stackoverflow.com/a/55035444) I don't think the `close()` and `join()` should be needed at the end.

dhruvkb (approved):
Works well for me locally, but I don't understand why 😅.

ingestion_server/ingestion_server/cleanup.py
obulat: Without the calls to `close()` and `join()`, the ingestion test was failing in CI. Probably because the audio ingestion was throwing an error about the index already existing.

PR #1063, opened 2023-03-29
Use label.yml to determine required labels

### LINKED ISSUE #884: Use label.yml to determine required labels
## Description

We currently have a list of the required labels explicitly defined in the `label_pr.py` script:

https://github.com/WordPress/openverse/blob/bc60edc64fe34e59bbda5b880db76f4b1b41b9da/automations/python/label_pr.py#L25

This list is redundant with the `is_required` field in the list of labels defined in `labels.yml`:

https://github.com/WordPress/openverse/blob/89953c6aec63260b9d49b86e0764d667fea4e8a4/automations/data/labels.yml#L17

We could leverage the latter in the script instead to find out which stacks are required. We could also add an `apply_all_available` boolean field to a group to replace the `GET_ALL_CATEGORY_LABELS` value:

https://github.com/WordPress/openverse/blob/bc60edc64fe34e59bbda5b880db76f4b1b41b9da/automations/python/label_pr.py#L27

Tomvth:
## Fixes

Fixes #884  by @AetherUnbound 

## Description

This PR implements both changes mentioned in the issue. 
1) Instead of using a hardcoded list of labels explicitly defined in labels_pr.py, instead now the labels which have set is_required to True are parsed from the labels.yml data file. 
2) The GET_ALL_CATEGORY_LABELS list is replaced and is also dynamically parsed from labels.yml, to do this a new apply_all_available boolean was added for each group. 

Since labels.yml was encoded in utf-8, an optional argument to the load_data function was added to allow specifying the file encoding.

## Testing Instructions

Run the labels_pr.py file, verify that the parsed list equates the original explicitly defined list.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/automations/data/labels.yml
+++ b/automations/data/labels.yml
@@ -15,6 +15,7 @@ colors:
 groups:
   - name: priority
     is_required: true
+    apply_all_available: false
     labels:
       - name: critical
         color: "UNFAVOURABLE"
@@ -35,6 +36,7 @@ groups:
 
   - name: status
     is_required: true
+    apply_all_available: false
     labels:
       - name: ready for work
         color: "LIGHT"
@@ -68,6 +70,7 @@ groups:
   - name: goal
     color: "ffffff"
     is_required: true
+    apply_all_available: false
     labels:
       - name: addition
         description: Addition of new feature
@@ -86,6 +89,7 @@ groups:
     color: "006b75"
     emoji: "🧱"
     is_required: true
+    apply_all_available: true
     labels:
       - name: frontend
         description: Related to the Nuxt frontend
@@ -101,6 +105,7 @@ groups:
         description: Related to the ingestion/data refresh server
 
   - name: tooling
+    apply_all_available: false
     labels:
       - name: sentry
         color: "362d59"
@@ -110,6 +115,7 @@ groups:
   - name: aspect
     color: "04338c"
     is_required: true
+    apply_all_available: false
     labels:
       - name: text
         description: Concerns the textual material in the repository
@@ -132,6 +138,7 @@ groups:
 
   - name: talk
     color: "f9bbe5"
+    apply_all_available: false
     labels:
       - name: question
         description: Can be resolved with an answer
@@ -142,6 +149,7 @@ groups:
 
   - name: tech
     color: "fcdf4a"
+    apply_all_available: false
     labels:
       # Programming languages
       - name: javascript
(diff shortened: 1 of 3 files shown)


automations/data/labels.yml
```diff
@@ -15,6 +15,7 @@ colors:
 groups:
   - name: priority
     is_required: true
+    apply_all_available: false
```
AetherUnbound, line 18: Since `false` is so prevalent here, maybe we should make it a default and allow it to be overwritten only when it's `true`? Similar to `is_required`.

automations/python/shared/data.py
```diff
@@ -12,16 +12,20 @@
 
 
 @cache
-def get_data(file_name: str) -> dict:
+def get_data(file_name: str, encoding: str = None) -> dict:
```
AetherUnbound, line 15: Was this necessary for reading the file? Python _should_ [default to `utf-8` when opening files](https://docs.python.org/3/howto/unicode.html#encodings) if it's not set explicitly. Even if we needed to make this explicit here, we should just default to `utf-8` anyway and avoid the need for the `if encoding` block below.

AetherUnbound (changes requested):
Thank you for the contribution! It looks like the linting checks might be failing, please try running `just lint` and committing the changes. I also have a few other suggestions for how to improve this approach.

automations/python/shared/data.py
Tomvth: It seems that [the encoding used by open() is platform-specific](https://docs.python.org/3/library/functions.html#open), although utf-8 seems to be the most prevelant. On my platform running open() on labels.yml results in `UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 980: character maps to <undefined>`. Specifying "utf-8" as encoding method fixes this. 
I agree on making it the default and I removed the if statement, get_data() now defaults to utf-8.

AetherUnbound (approved):
This is excellent!! Thanks again for contributing 😄

dhruvkb (approved):
I like this very much, thanks @Tomvth!

automations/python/shared/data.py
dhruvkb: What platform are you using @Tomvth? Since these actions will be running inside GitHub Actions, the default encoding will be UTF-8.

PR #1100, opened 2023-03-30
Update other references of media count to 700 million

krysal:
## Description

While looking at the same problem that #1098 resolves I thought of extracting the default meta tags to its own file, so the Nuxt configuration is reduced by a some lines and also creates a constant for the page description :) Not the ideal solution still but the number of items is more centralized.

## Testing Instructions

1. `just frontend/run dev`
2. Visit http://localhost:8443/ and check on the source page the "700 million" description is still showed correctly.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/constants/meta.ts
+++ b/frontend/src/constants/meta.ts
@@ -1,0 +1,46 @@
+/**
+ * The default metadata for the site. Can be extended and/or overwritten per page. And even in components!
+ * See the Nuxt.js docs for more info.
+ * {@link https://nuxtjs.org/guides/features/meta-tags-seo} Nuxt.js Docs
+ */
+
+const NUMBER_OF_RECORDS = "700 million"
+const DESCRIPTION = `Search over ${NUMBER_OF_RECORDS} free and openly licensed images, photos, audio, and other media types for reuse and remixing.`
+
+export const meta = [
+  { charset: "utf-8" },
+  {
+    name: "viewport",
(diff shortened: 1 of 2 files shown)


frontend/src/constants/meta.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * The default metadata for the site. Can be extended and/or overwritten per page. And even in components!
+ * See the Nuxt.js docs for more info.
+ * {@link https://nuxtjs.org/guides/features/meta-tags-seo} Nuxt.js Docs
+ */
```
dhruvkb, line 5: I think this doc comment was associated with the `meta` variable on L10.

dhruvkb (approved):
I like the idea of `NUMBER_OF_RECORDS` being a constant 👍. 

Making it modifiable via an env var would've been nice but it'll break snapshots when changed so I don't mind this solution.

frontend/src/constants/meta.ts
krysal: Right, good catch! Thank you.

zackkrida (approved):
LGTM. I made a follow-up issue to address the fact that these strings aren't localized: #1129

Issue #392, opened 2023-02-10
Search relevancy sandbox


obulat:
| Start Date        | ETA | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2023-04-01 | 2024-04-31 | @AetherUnbound | TBD              |

## Description

Modify the API staging environment to include a proportional subset of media from each provider. Increase the frequency of data refreshes.

This project does not address metrics for measuring and tracking relevancy.

To implement this project, we want to read the production `/stats/` endpoints to get the media totals for each provider, then scale these numbers down to set the ingestion limit per provider in staging.

This project could also include the update of the Elasticsearch cluster (or setting up a new cluster and moving the staging there).

## Documents

[Project homepage](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/index.html)

- [x] https://github.com/WordPress/openverse/pull/1107 - [doc link](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html)
- [x] https://github.com/WordPress/openverse/pull/1154 - [doc link](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.html)
- [x] https://github.com/WordPress/openverse/pull/2133 - [doc link](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.html#outlined-steps)
- [x] https://github.com/WordPress/openverse/pull/2358

## Issues

- [x] https://github.com/WordPress/openverse/issues/1985
- [x] https://github.com/WordPress/openverse/issues/1987
- [x] https://github.com/WordPress/openverse/issues/1989
- [x] https://github.com/WordPress/openverse/issues/1990
- [x] https://github.com/WordPress/openverse/issues/2370
- [x] https://github.com/WordPress/openverse/issues/2371
- [x] #3232
- [x] https://github.com/WordPress/openverse/issues/2372
- [x] #3488 
- [x] #3492 
- [x] #3493

## Milestones

- [ ] Rapid iteration on Elasticsearch indices: https://github.com/WordPress/openverse/milestone/16

## Prior Art

- [Produce test harness for measuring the quality of search results](https://github.com/cc-archive/cccatalog-api/issues/219)
- [Script which added 100 records per provider into the testing ECS database](https://github.com/WordPress/openverse-infrastructure/pull/314)

[...]

PR #1107, opened 2023-03-31
Project Proposal: Search Relevancy Sandbox

AetherUnbound:
## Due date:

2023-04-14

## Assigned reviewers

- [x] @sarayourfriend
- [x] @krysal 
- [x] @zackkrida 

@dhruvkb - I chose you because you've worked on the ingestion server and Elasticsearch in the past and may be familiar (albeit distantly!) with some of these pieces. I know that you're involved in a number of other projects at this moment, so please feel free to opt out if you're currently at capacity.

@sarayourfriend - I chose you because you're familiar with many of the systems at play and have been exploring modifications to Elasticsearch, so your insight into the "user-facing" side of this project would be invaluable.

## Description

This PR is the kick-off for the search relevancy sandbox project: https://github.com/WordPress/openverse/issues/392

### DIFF
--- a/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
+++ b/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
@@ -68,9 +68,9 @@ test API results from all currently available providers.
 
 ### Production data volume
 
-The production-data-volume index will be an index of similar size to production.
-This is how the data refresh is currently set up and does not require any
-additional work to set up.
+The production-data-volume index will be an index of similar size (i.e. on the
+same order of magnitude in document count) to production. This is how the data
+refresh is currently set up and does not require any additional work to set up.
 
 This index will allow us to test how queries and indexing operations perform at
 a data volume consistent with production.
@@ -131,8 +131,8 @@ In the order they should be completed:
    - This plan will describe a DAG which can be triggered to update the staging
      API database.
 2. Rapid iteration on ingestion server index configuration
-   - This plan will describe how rapid iteration of the Elasticsearch index
-     configuration can happen.
+   - This plan will describe how rapid iteration of Elasticsearch index
+     configurations can happen.
    - It should be assessed as part of this implementation plan whether it would
      be easier to convert the ingestion server to ECS or provide a mechanism on
      the existing EC2 infrastructure to update the Elasticsearch index
--- a/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
+++ b/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
@@ -46,11 +46,16 @@ Many of these mechanisms will likely be built in Airflow.
 The proportional-by-provider index will be a subset of the production data,
 constructed so that the number of records by provider is roughly proportional to
 the percentages per provider that exist in production. This index will allow us
-to create tests down the line which might assess result relevancy in an
-automated way, i.e. based on how many results from a particular provider appear
-in results (see the
-[provider tally script](https://github.com/WordPress/openverse/pull/397) for
-examples of these results). This could be done by:
+to iterate rapidly on the following:
+
+- Changes affecting Elasticsearch index configuration (e.g. shard sizes,
+  slicing, etc.)
+- Changes to the ingestion server's configuration which require running a data
+  refresh end-to-end
+- Measuring index performance
+- Integration testing dead link & thumbnail checks across all providers
+
+This could be done by:
 
 1. Querying the `/stats` endpoint of the production API (e.g.
    https://api.openverse.engineering/v1/images/stats/).
@@ -132,7 +137,12 @@ In the order they should be completed:
      API database.
 2. Rapid iteration on ingestion server index configuration
    - This plan will describe how rapid iteration of Elasticsearch index
-     configurations can happen.
+     configurations can happen. This can be done via Airflow interacting
+     directly with Elasticsearch rather than having to process requests through
+     the ingestion server. Airflow would only need to be aware of the
+     [existing index settings](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
+     and how to augment/adjust them for the new index.
+     ([See this comment for further inspiration](https://github.com/WordPress/openverse/pull/1107#discussion_r1155399508))
    - It should be assessed as part of this implementation plan whether it would
      be easier to convert the ingestion server to ECS or provide a mechanism on
      the existing EC2 infrastructure to update the Elasticsearch index
--- a/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
+++ b/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
@@ -1,0 +1,145 @@
+# Project Proposal: Search Relevancy Sandbox
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] dhruvkb
+- [ ] sarayourfriend
+
+## Project summary
+
+<!-- A brief one or two sentence summary of the project's features -->
+
+Develop mechanisms and workflows for sustainably creating & updating our staging
+API database & Elasticsearch cluster, in order to enable testing changes to
+Elasticsearch indices. The staging setup will reflect production in two ways:
+proportionally by provider and with production-level volumes.
+
+## Goals
+
+<!-- Which yearly goal does this project advance? -->
+
+Yearly goal: **Result Relevancy**
+
+## Requirements
+
+<!-- Detailed descriptions of the features required for the project. Include user stories if you feel they'd be helpful, but focus on describing a specification for how the feature would work with an eye towards edge cases. -->
+
+The following should be available once this project is complete:
+
+1. A mechanism by which maintainers can easily update the staging API database
+   with recent data.
+2. A mechanism by which maintainers can create/update a small but
+   proportional-to-production-by-provider index.
+3. A mechanism by which maintainers can create/update a production-data-volume
+   sized index.
+4. A mechanism by which maintainers can easily deploy & iterate on new
+   Elasticsearch indexing configurations
+5. A mechanism by which maintainers can easily point the staging Elasticsearch
+   index aliases (currently `image` and `audio`) to one of the above.
+
+Many of these mechanisms will likely be built in Airflow.
+
+### Proportional by provider
+
+The proportional-by-provider index will be a subset of the production data,
+constructed so that the number of records by provider is roughly proportional to
+the percentages per provider that exist in production. This index will allow us
+to create tests down the line which might assess result relevancy in an
+automated way, i.e. based on how many results from a particular provider appear
+in results (see the
+[provider tally script](https://github.com/WordPress/openverse/pull/397) for
+examples of these results). This could be done by:
+
+1. Querying the `/stats` endpoint of the production API (e.g.
+   https://api.openverse.engineering/v1/images/stats/).
+2. Calculating percentage-per-provider.
+3. Computing the number of results per provider based on these percentages based
+   on the condition that the smallest provider has a minimum `N` results (e.g. a
+   minimum of 10 results).
+4. Pulling a number of results based on the previous step per-provider either
+   into a Postgres table or an Elasticsearch index (the specifics of this step
+   will be left for the implementation plan).
+
+This index will allow us to assess how changes to Elasticsearch index or the API
+will affect results using a smaller total result count. It will also allow us to
+test API results from all currently available providers.
+
+### Production data volume
+
+The production-data-volume index will be an index of similar size to production.
+This is how the data refresh is currently set up and does not require any
+additional work to set up.
+
+This index will allow us to test how queries and indexing operations perform at
+a data volume consistent with production.
+
+## Success
+
+<!-- How do we measure the success of the project? How do we know our ideas worked? -->
+
+This project can be considered a success once any maintainer can make
+adjustments to the staging API database & Elasticsearch index based on the
+requirements described above.
+
+## Participants and stakeholders
+
+<!-- Who is working on the project and who are the external stakeholders, if any? Consider the lead, implementers, designers, and other stakeholders who have a say in how the project goes. -->
+
+- Lead: @AetherUnbound
+- Implementation:
+  - @AetherUnbound
+  - TBD
+- Stakeholders:
+  - Openverse Team
+
+## Infrastructure
+
+<!-- What infrastructural considerations need to be made for this project? If there are none, say so explicitly rather than deleting the section. -->
+
+This project may require infrastructure work necessary to assist with the rapid
+iteration of the data refresh server. Further details will be determined in the
+implementation plan.
+
+The project will also likely involve the creation or modification of Airflow
+DAGs related to the data refresh and infrastructure modification. This will
+require interfacing Airflow with the staging data refresh server, API database,
+and Elasticsearch cluster.
+
+## Accessibility
+
+<!-- Are there specific accessibility concerns relevant to this project? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+
+Not every member of the maintainer team is intimately familiar with Airflow - we
+will need to provide clear instructions for maintainers on how to use Airflow to
+run any of the above mechanisms.
+
+## Marketing
+
+<!-- Are there potential marketing opportunities that we'd need to coordinate with the community to accomplish? If there are none, say so explicitly rather than deleting the section. -->
+
+No marketing is required as this is an improvement internal to the team.
+
+## Required implementation plans
+
+<!-- What are the required implementation plans? Consider if they should be split per level of the stack or per feature. -->
+
+In the order they should be completed:
+
+1. Staging API update procedure
+   - This plan will describe a DAG which can be triggered to update the staging
+     API database.
+2. Rapid iteration on data refresh server index configuration
+   - This plan will describe how rapid iteration of the Elasticsearch index
+     configuration can happen.
+   - It should be assessed as part of this implementation plan whether it would
+     be easier to convert the data refresh server to ECS or provide a mechanism
+     on the existing EC2 infrastructure to update the Elasticsearch index
+     configuration without issuing a deployment (similar to the
+     [dag-sync script for the catalog](https://github.com/WordPress/openverse-catalog/blob/10857e3ee94ae686853984c54d504b152082d4c2/dag-sync.sh)).
+3. Staging data refresh DAGs for both potential index types
+   - This plan will describe the DAG or DAGs which will be used to create/update
+     both the proportional-by-provider and production-data-volume indices.
+   - It will also describe the mechanism by which maintainers can rapidly switch
+     index aliases.
--- a/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
+++ b/rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
@@ -99,13 +99,13 @@ requirements described above.
 <!-- What infrastructural considerations need to be made for this project? If there are none, say so explicitly rather than deleting the section. -->
 
 This project may require infrastructure work necessary to assist with the rapid
-iteration of the data refresh server. Further details will be determined in the
+iteration of the ingestion server. Further details will be determined in the
 implementation plan.
 
 The project will also likely involve the creation or modification of Airflow
 DAGs related to the data refresh and infrastructure modification. This will
-require interfacing Airflow with the staging data refresh server, API database,
-and Elasticsearch cluster.
+require interfacing Airflow with the staging ingestion server, API database, and
+Elasticsearch cluster.
 
 ## Accessibility
 
@@ -130,12 +130,12 @@ In the order they should be completed:
 1. Staging API update procedure
    - This plan will describe a DAG which can be triggered to update the staging
      API database.
-2. Rapid iteration on data refresh server index configuration
+2. Rapid iteration on ingestion server index configuration
    - This plan will describe how rapid iteration of the Elasticsearch index
      configuration can happen.
    - It should be assessed as part of this implementation plan whether it would
-     be easier to convert the data refresh server to ECS or provide a mechanism
-     on the existing EC2 infrastructure to update the Elasticsearch index
+     be easier to convert the ingestion server to ECS or provide a mechanism on
+     the existing EC2 infrastructure to update the Elasticsearch index
      configuration without issuing a deployment (similar to the
      [dag-sync script for the catalog](https://github.com/WordPress/openverse-catalog/blob/10857e3ee94ae686853984c54d504b152082d4c2/dag-sync.sh)).
 3. Staging data refresh DAGs for both potential index types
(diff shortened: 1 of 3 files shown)


rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,145 @@
...
+### Proportional by provider
+
+The proportional-by-provider index will be a subset of the production data,
+constructed so that the number of records by provider is roughly proportional to
+the percentages per provider that exist in production. This index will allow us
+to create tests down the line which might assess result relevancy in an
+automated way, i.e. based on how many results from a particular provider appear
+in results (see the
+[provider tally script](https://github.com/WordPress/openverse/pull/397) for
+examples of these results). This could be done by:
```
sarayourfriend, line 53: I'm slightly sceptical that this is a reasonable goal/down-the-line use case. For it to work we'd not only need proportional-by-provider, we'd also need to scope it by specific queries... at which point we're optimising one or a handful of specific query terms. We don't have a uniform distribution, for example, of "dog" images, or "vase", or "flower", etc. I'm sure Flickr is dominant in all of them, but I'd bet few museums have many results for dog but many do for vases and iNaturalist for flowers, etc.

Overall, I don't think we can actually test relevancy in this way. It _needs_ to be based on user data and a/b style testing to see whether click-through increases. Otherwise, we'll be falling back to what I earlier called the "we know it when we see it" approach, which I have low to zero confidence in and is not a quantitative approach viable in the long-term.

It is worth mentioning this to avoid us couching work in the possibility of being able to "manually test relevancy" in staging, either with actual production data or proportionally-similar-to-production data. There are use-cases for manually testing index changes in staging, but I think primarily for testing things like text analysis changes to address specific problems rather than broad relevancy. Those, in particular, would need to have _specific_ data sets, not just proportional-by-provider data sets.

If we wanted to cover changes to the stemmer for the following, how large a percentage of the production data would we need to be able to actually reliably see results relevant to these changes?

- "universe"
- "animal"
- "news"

It's not unreasonable to think that even 50% of production data would exclude one or the other of these, and these are real examples of stemmer issues we have encountered in production. There are likely heaps of others that will be even more diverse.

The bottom line here is that: proportional-by-provider _sounds_ right, it does sound like a good idea. However, I just don't know what it's actually useful for if it's a naive proportion and not directed for specific query terms. At which point, it's only useful for specific problems, and we need to incorporate the assumption that specific query terms are needed to generate these proportional-by-provider indexes.

It might also be useful for something like an integration test that runs through a representative and random sample of data for each provider and tests thumbnails and other edge cases that we've found vary by provider at times. But then we need to decide what "random" means and how we'd implement it in the context of ES and Postgres, I'm sure each have their own unique and variously relevant approaches to that problem.

sarayourfriend: Additional comment though to say that aside from the motivation, the broad approach described in this section makes sense to me based on the assumptions it was written under. I just want to clarify that I am questioning the assumptions for this section, not the implementation overview in it, which for the assumptions given, make sense to me.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,145 @@
...
+   into a Postgres table or an Elasticsearch index (the specifics of this step
+   will be left for the implementation plan).
+
+This index will allow us to assess how changes to Elasticsearch index or the API
+will affect results using a smaller total result count. It will also allow us to
+test API results from all currently available providers.
+
+### Production data volume
+
+The production-data-volume index will be an index of similar size to production.
```
sarayourfriend, line 71: What does "similar size" mean, in this context? Is the variation down to the fact that any data refresh started _after_ another will have slightly more data due to ongoing ingestion into the catalogue?

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,145 @@
...
+  - TBD
+- Stakeholders:
+  - Openverse Team
+
+## Infrastructure
+
+<!-- What infrastructural considerations need to be made for this project? If there are none, say so explicitly rather than deleting the section. -->
+
+This project may require infrastructure work necessary to assist with the rapid
+iteration of the data refresh server. Further details will be determined in the
```
sarayourfriend, line 102: Does "data refresh server" mean the ingestion server project?

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,145 @@
...
+<!-- What are the required implementation plans? Consider if they should be split per level of the stack or per feature. -->
+
+In the order they should be completed:
+
+1. Staging API update procedure
+   - This plan will describe a DAG which can be triggered to update the staging
+     API database.
+2. Rapid iteration on data refresh server index configuration
+   - This plan will describe how rapid iteration of the Elasticsearch index
+     configuration can happen.
```
sarayourfriend, line 135: What does "the Elasticsearch index" mean? Specifically the "the" isn't clear to me :thinking: ES can have any number of indexes, each with their own independent configuration. Rapidly iterating on index configuration could be as easy as changing the request body to the create index endpoint, which someone could either issue directly in Elasticvue (or similar) or give as input to an Airflow DAG, for example.

I also think we should leverage the re-index API for this. I could see an Airflow DAG meant for iterating on index configuration that works something like this:

1. Trigger the DAG with a new index configuration body.
    - The existing configuration can be generated using the `index_settings` function from the ingestion server project. This function runs fine in a Python shell with zero setup inside the ingestion_server project.
1. The DAG creates the index via the create index endpoint, either using a name provided when the DAG was triggered or by generating a UUID or date-based suffix. This should happen by hitting ES directly, not through the ingestion server.
1. The DAG then calls the reindex API to copy the data from the existing (periodically refreshed?) "production data" index into the new index. If we want to incorporate "proportional data" or have a subset of data, we could also have the DAG take in a query. If we just need to test a stemmer change, for example, we can query for documents that include the terms in question. For broader text analysis changes we'd want to use the full production data (see my comment above). If we do use the proportional-by-provider data, we can allow changing the origin index between the full production data and the proportional-by-provider index.
1. After reindex is done, refresh the new index.

Done! We've just completed one index configuration iteration.

This bypasses the complexity of the ingestion server and wouldn't require any changes to the ingestion server _at all_.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
sarayourfriend: One more thing: I could see proportional-by-provider (but still statistically relevant sample size!) being useful for iterating on changes to how indexing happens: shard sizes, slicing, etc. In that case, the actual data itself doesn't matter, just that we have a statistically significant sample size that can be used to understand how long, on average, documents take to index, etc. If we want to iterate on the ingestion server configuration as well, we can use this to have shorter but still relevant enough runs.

Forgive my repeated clarifications. I just want to make sure that my line of questioning isn't misunderstood to mean that proportional data is outright not useful. I'm specifically sceptical about its usefulness for relevancy testing which I think, by necessity, needs to be done in production, with real user behaviour.

sarayourfriend (commented):
Requesting clarification for a few things.

Overall I think this proposal works, but I think we should mention some specific approaches that we could take that would be way easier than trying to run top-to-bottom data refreshes in staging. Easier both in reduced complexity and in time. We can entirely avoid pushing data from Postgres into Elasticsearch if all we're testing are changes to index configuration: that is worth noting specifically for the implementation plan, especially as it allows us to cut out moving the ingestion server to ECS, a big win in terms of the scale of the downstream work, I would think.

I also wonder if we can bypass data refresh for the proportional-by-provider stuff as well. We could avoid making any changes to the data refresh process or ingestion server code altogether to support the proportional index. For example, a DAG that used [`random_score`](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-function-score-query.html#function-random) to pull random documents for a specific provider and paginated using [`search_after`](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/paginate-search-results.html#search-after) until we have the right number for that provider, would accomplish the process without needing to copy data from Postgres for it or rely on the ingestion server. Especially if we want to eventually move the data refresh process out of the ingestion server project and into Airflow, this might be a good opportunity to start wiring up the ES connection direct from Airflow.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: These are great questions and points! I appreciate you drawing question to the assumed need for this type of index at all. My original motivation behind this was actually the [data transfer script](https://github.com/WordPress/openverse-infrastructure/blob/5b3e2c625d6119d4c8a0ceefbfc3bac4e58b82a5/next/modules/concrete/api/scripts/data-transfer.sh) we used when creating a small sample dataset for the API ECS migration. In that case, we selected 100 rows (not randomized) from each provider for exactly the reasons you describe - to check that there were no issues serving them via the API and that the thumbnails worked. 

My _primary_ motivation here is that I want to have an index that is both **small** and **has data from each provider we serve in production**. The former is a requirement in my mind because I don't want to make any dev wait around for one (or several) days while reindexing happens so they can check that something for a provider works in staging. The reasoning for the latter is to ensure that we have many possible edge cases available (though having the entire dataset indexed also nets us this).

I'm not tied to the proportionality piece of this, my intent was to _maybe_ catch two birds with one stone by having this both be small _and_ proportional, but your reasoning for why it may not be useful makes sense.

Do you think there would still be a benefit to having, like we did in API ECS testing, an index that had 100 records per provider? That could replace this index requirement and we could remove the proportionality aspect described here.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: Yes, that's exactly what I mean but I can be more specific. While I'm glad that the 500M document index worked well for your testing, I think it should be possible for us to match production close enough that staging trails it by a week (and at most a month's) worth of data. I will codify my intentions here a bit better.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: Yes it does, I was torn trying to determine which to put down since the _process_ was renamed to "data refresh", but all our code and server references are to "ingestion server" so I'll change it back.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: I love the idea of leveraging Airflow the way you're describing!

By "the" index I mean - the primary index that will eventually serve as the index behind the `image|audio` alias. In this case/example, that's not what would be iterated on though, I'll try to reword this to be more generic as it relates to indices. The specifics you describe will go into the eventual implementation plan for this step, so I won't include them here.

AetherUnbound (commented):
Thanks for your feedback @sarayourfriend! I'll address a few of the questions you bring up, but I'll wait for your (and others') thoughts on the proportionality before I change that aspect of the proposal.

sarayourfriend:
Just wanted to confirm that I'm AFK until Tuesday, but I will respond here ASAP on my Tuesday morning/your Monday afternoon @AetherUnbound.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
sarayourfriend: As I said in the last comment, my main concern is in establishing an expectation that these data sets enable artificial relevancy tests. Relevancy needs to be tested against real user data. My issue is not with the concept of a smaller data set generally. It is undoubtedly useful for a myriad of use cases other than relevancy: measuring indexing performance, integration testing dead link and thumbnail checks with all providers, etc. Proportionality and randomised selection is especially useful for something like index performance as you need a representative sample. Integration tests theoretically require far fewer results but also benefit from sample diversity so that we cast as wide a net as possible, even in a smaller data set.

I fully agree we should have this ability, I just want us to be clear on the motivations and use cases it is relevant to.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
sarayourfriend: Makes sense, I don't think it is a huge distinction, I just wanted to make sure I was understanding what the possible discrepancy would be. Using production snapshots as described in the first implementation plan makes a lot of sense and clarifies the data lag that would appear when a data refresh runs.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
sarayourfriend: Yeah, unfortunately with already unclear names I think it's worth sticking with the specific one until they're actually changed in code, otherwise it's quite confusing :disappointed:. It's worth noting that "ingestion server" _does_ make sense from an API perspective: the service ingests catalogue data into ES and the API database. Maybe "API ingestion server" would be a good interim name?

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
sarayourfriend: Sounds good. I think it is worth recording the general approach in this document, otherwise the person doing the implementation plan may not know about this being a viable approach and go down a different rabbit hole. A small note in the implementation plan section that we can use the reindex API directly from Airflow to ES rather than passing through the ingestion server would be sufficient. Right now the suggested approach feels heavily skewed towards needing to iterate on the ingestion server code itself and maybe even moving it to ECS, which is a _lot_ more work that we could avoid. I think it would at least triple the scope of the implementation plan if it included moving the ingestion server to ECS vs just hitting ES directly from Airflow and using the reindex API.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: Ah, I see! Thanks for clarifying your intentions - that makes perfect sense. I'll remove the language around using this to assess result relevancy and replace it with some of the motivations you mention.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: I think "ingestion server" is fine to remain as-is here. The catalog _does_ ingestion but I wouldn't consider it an "ingestion server", and we refer to the "data refresh" process separately anyway.

AetherUnbound:
@sarayourfriend I've some clarity via https://github.com/WordPress/openverse/pull/1107/commits/d044dfd4c7f4d98e257abe67014bee19733bd1e3 around the motivation behind the proportionality index & suggestion for implementation plans. I think that covers all of your points of clarification but please let me know if I've missed anything or if the new revisions are unclear.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,155 @@
+# Project Proposal: Search Relevancy Sandbox
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] dhruvkb
+- [ ] sarayourfriend
```
sarayourfriend, line 8:
```suggestion
- [x] sarayourfriend
```

sarayourfriend (approved):
LGTM! Excited for this :rocket:

zackkrida:
@AetherUnbound I'll review this now, in the meantime I just wanted to point out the structure will need to be updated to meet the changes in #1162.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Project summary
+
+<!-- A brief one or two sentence summary of the project's features -->
+
+Develop mechanisms and workflows for sustainably creating & updating our staging
+API database & Elasticsearch cluster, in order to enable testing changes to
+Elasticsearch indices. The staging setup will reflect production in two ways:
+proportionally by provider and with production-level volumes.
```
zackkrida, line 17: > The staging setup will reflect production in two ways:
proportionally by provider and with production-level volumes.

This might be a nit but this sentence reads as confusing to me. If staging is provider and volume-proportionate to production, it would be identical to production! Maybe some rewording to suggest that it can be in _either_ of these states for testing purposes would be useful? In general, the concept that the staging ES cluster can be adapted to have different results at different times is probably worth highlighting in this section.

zackkrida (approved):
This looks great, and I had some thoughts that I hope do not derail this work but that I am curious to hear your thoughts on! The last bullet of the last implementation plan:

> - It will also describe the mechanism by which maintainers can rapidly switch index aliases.

Feels quite important to me and like it may be a bit under-explained by this proposal. As written, the proposal seems to suggest that the DAGs would be used to allow for creating a new staging index that replaces the default, "live" index the staging API is pointed at. Is that correct?

Alternatively, I can see a lot of value in having DAGs that allow for creating and updating staging indices with different qualities, but having a mechanism that allows developers to switch between them much more rapidly, perhaps even a query param-based solution (which would allow, for example, to switch indices via a checkbox on the staging.openverse.org/preferences page). 

To put this all another way, I'm wondering if requirement #5 should be a bit more specific in terms of what "easily" means.

One more thought: It occurs to me that the staging API database is _always_ just a subset of the production database, with the exception of situations where we're developing column changes to the media tables. In that sense, it seems wasteful and costly to even _have_ `image` and `audio` mat views in staging. Instead, I wonder if we could figure something out with foreign data wrappers in Postgres to create the staging matviews from the production database. It's pretty straightforward to create a read-only user for the prod db that we would connect the staging db with, and then read the media from production. I imagine that would also save considerable time in the data refresh process as well for staging when doing a production-data-volume sized index. It would also allow us to save considerable cost and resources. Production CPU usage pretty reliably hovers around 4% so I don't think the extra reads from staging would incur much of a cost at all.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,155 @@
+# Project Proposal: Search Relevancy Sandbox
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] dhruvkb
```
zackkrida, line 7:
```suggestion
- [x] @zackkrida
```

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
```diff
@@ -0,0 +1,155 @@
...
+     directly with Elasticsearch rather than having to process requests through
+     the ingestion server. Airflow would only need to be aware of the
+     [existing index settings](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
+     and how to augment/adjust them for the new index.
+     ([See this comment for further inspiration](https://github.com/WordPress/openverse/pull/1107#discussion_r1155399508))
+   - It should be assessed as part of this implementation plan whether it would
+     be easier to convert the ingestion server to ECS or provide a mechanism on
+     the existing EC2 infrastructure to update the Elasticsearch index
+     configuration without issuing a deployment (similar to the
+     [dag-sync script for the catalog](https://github.com/WordPress/openverse-catalog/blob/10857e3ee94ae686853984c54d504b152082d4c2/dag-sync.sh)).
```
krysal, line 150: I'm eager to get the results of this assessment! If it's worth it, I'd be interested in helping with the migration to ECS.

krysal (approved):
Thanks for writing this proposal. It clarified many of my questions for this project.

rfcs/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md
AetherUnbound: Okay! I'll try to reword this a bit 🙂

AetherUnbound:
> This looks great, and I had some thoughts that I hope do not derail this work but that I am curious to hear your thoughts on! The last bullet of the last implementation plan:

> > - It will also describe the mechanism by which maintainers can rapidly switch index aliases.

> Feels quite important to me and like it may be a bit under-explained by this proposal. As written, the proposal seems to suggest that the DAGs would be used to allow for creating a new staging index that replaces the default, "live" index the staging API is pointed at. Is that correct?

> Alternatively, I can see a lot of value in having DAGs that allow for creating and updating staging indices with different qualities, but having a mechanism that allows developers to switch between them much more rapidly, perhaps even a query param-based solution (which would allow, for example, to switch indices via a checkbox on the staging.openverse.org/preferences page). 

> To put this all another way, I'm wondering if requirement #5 should be a bit more specific in terms of what "easily" means.

That's a great point, but I actually think this would be a good thing to determine in the implementation plan! It was my intention to have a single alias (like production, at least at the current moment) and allow maintainers to swap between these aliases. However, I do quite like the idea of an API setting which might allow dynamically changing which index is used at query time! That would bring increased complexity to the plan and the project that I wasn't initially considering, but if we scope it out as part of the plan, potentially starting with the rapid swapping but laying out a plan for what changing the index at query time might look like, we could even add that capability down the line outside of the bounds of this specific effort!

> One more thought: It occurs to me that the staging API database is _always_ just a subset of the production database, with the exception of situations where we're developing column changes to the media tables. In that sense, it seems wasteful and costly to even _have_ `image` and `audio` mat views in staging. Instead, I wonder if we could figure something out with foreign data wrappers in Postgres to create the staging matviews from the production database. It's pretty straightforward to create a read-only user for the prod db that we would connect the staging db with, and then read the media from production. I imagine that would also save considerable time in the data refresh process as well for staging when doing a production-data-volume sized index. It would also allow us to save considerable cost and resources. Production CPU usage pretty reliably hovers around 4% so I don't think the extra reads from staging would incur much of a cost at all. 

I think a few things might be conflated here - the API databases don't have their own matviews, only the catalog has a matview. The implementation plan in #1154 describes how we could replace the staging API database wholesale without having to worry about the matview refresh aspect, since that will have already been completed in order to create the prod database (on which the snapshot is based). The data refresh steps described in this document are only those which involve indexing data from the database into Elasticsearch - I'll try to make that clearer in the document!

zackkrida:
Oops, accidentally referred to the API DB media tables as matviews. I'll check out #1154! My concern was always writing and storing a subset of all Openverse media to the staging api db when changing the indices and it sounds like that won't be a concern.

sarayourfriend:
> laying out a plan for what changing the index at query time might look like

FWIW, if anyone is looking at this later during implementation planning, this probably doesn't need to be more complicated than a query parameter that determines the index, but we'd need to be careful to juggle things in light of https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html

PR #1118, opened 2023-04-03
Analytics event: BACK_TO_SEARCH

### LINKED ISSUE #1087: Analytics event: `BACK_TO_SEARCH`
## Description

The goal is to 

- add the type for event `BACK_TO_SEARCH` to the `Events` type.

## Reference

https://github.com/WordPress/openverse/blob/839418fabb0972ec3c3ccf48c72fd64cbba03cb9/rfcs/analytics/20221006-implementation_plan_frontend.md?plain=1#L424

## Additional information

- [RFC](https://github.com/WordPress/openverse/blob/main/rfcs/analytics/20221006-implementation_plan_frontend.md#events)

masif2002:
## Fixes

Fixes #1087 by @dhruvkb 

## Description

* Adds `BACK_TO_SEARCH` event to the Events type.
* Fires the event from `VBackToSearchResultsLink.vue` using the `useAnalytics` composable.

### DIFF
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
@@ -3,6 +3,7 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 pe-3 text-xs font-semibold text-dark-charcoal-70 hover:text-dark-charcoal"
     v-bind="$attrs"
+    @click="handleClick()"
   >
     <VIcon name="chevron-left" :rtl-flip="true" />
     {{ $t("single-result.back") }}
@@ -10,7 +11,10 @@
 </template>
 
 <script lang="ts">
-import { defineComponent } from "vue"
+import { defineComponent, PropType } from "vue"
+
+import { useAnalytics } from "~/composables/use-analytics"
+import type { SupportedMediaType } from "~/constants/media"
 
 import VIcon from "~/components/VIcon/VIcon.vue"
 import VLink from "~/components/VLink.vue"
@@ -24,6 +28,35 @@ export default defineComponent({
     VIcon,
     VLink,
   },
+  props: {
+    /**
+     * The unique ID of the media
+    */
+    id: {
+      type: String,
+      required: true,
+    },
+    /**
+     * The media type being searched
+     */
+     mediaType: {
+      type: String as PropType<SupportedMediaType>,
+      required: true,
+    },
+  },
   inheritAttrs: false,
+  setup(props) {
+    const { sendCustomEvent } = useAnalytics()
+    const handleClick = () => {
+      sendCustomEvent("BACK_TO_SEARCH", {
+        id: props.id,
+        mediaType: props.mediaType,
+      })
+    }
+
+    return {
+      handleClick
+    }
+  },
 })
 </script>
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
@@ -3,7 +3,7 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 pe-3 text-xs font-semibold text-dark-charcoal-70 hover:text-dark-charcoal"
     v-bind="$attrs"
-    @click="handleClick"
+    @mousedown="handleClick"
   >
     <VIcon name="chevron-left" :rtl-flip="true" />
     {{ $t("single-result.back") }}
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
@@ -3,7 +3,7 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 pe-3 text-xs font-semibold text-dark-charcoal-70 hover:text-dark-charcoal"
     v-bind="$attrs"
-    @click="handleClick()"
+    @click="handleClick"
   >
     <VIcon name="chevron-left" :rtl-flip="true" />
     {{ $t("single-result.back") }}
@@ -28,6 +28,7 @@ export default defineComponent({
     VIcon,
     VLink,
   },
+  inheritAttrs: false,
   props: {
     /**
      * The unique ID of the media
@@ -39,12 +40,11 @@ export default defineComponent({
     /**
      * The media type being searched
      */
-     mediaType: {
+    mediaType: {
       type: String as PropType<SupportedMediaType>,
       required: true,
     },
   },
-  inheritAttrs: false,
   setup(props) {
     const { sendCustomEvent } = useAnalytics()
     const handleClick = () => {
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
@@ -11,10 +11,10 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType } from "vue"
+import { defineComponent } from "vue"
 
 import { useAnalytics } from "~/composables/use-analytics"
-import type { SupportedMediaType } from "~/constants/media"
+import { useSearchStore } from "~/stores/search"
 
 import VIcon from "~/components/VIcon/VIcon.vue"
 import VLink from "~/components/VLink.vue"
@@ -37,20 +37,15 @@ export default defineComponent({
       type: String,
       required: true,
     },
-    /**
-     * The media type being searched
-     */
-    mediaType: {
-      type: String as PropType<SupportedMediaType>,
-      required: true,
-    },
   },
   setup(props) {
     const { sendCustomEvent } = useAnalytics()
+    const searchStore = useSearchStore()
+
     const handleClick = () => {
       sendCustomEvent("BACK_TO_SEARCH", {
         id: props.id,
-        mediaType: props.mediaType,
+        searchType: searchStore.searchType,
       })
     }
 
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
@@ -32,7 +32,7 @@ export default defineComponent({
   props: {
     /**
      * The unique ID of the media
-    */
+     */
     id: {
       type: String,
       required: true,
@@ -55,7 +55,7 @@ export default defineComponent({
     }
 
     return {
-      handleClick
+      handleClick,
     }
   },
 })
--- a/frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx
+++ b/frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx
@@ -8,7 +8,7 @@ import VBackToSearchResultsLink from "~/components/VBackToSearchResultsLink.vue"
 />
 
 export const Template = (args) => ({
-  template: `<VBackToSearchResultsLink v-bind="args"/>`,
+  template: `<VBackToSearchResultsLink v-bind="args" :id="id" :media-type="mediaType" />`,
   components: { VBackToSearchResultsLink },
   setup() {
     return { args }
--- a/frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx
+++ b/frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx
@@ -8,7 +8,7 @@ import VBackToSearchResultsLink from "~/components/VBackToSearchResultsLink.vue"
 />
 
 export const Template = (args) => ({
-  template: `<VBackToSearchResultsLink v-bind="args" :id="id" :media-type="mediaType" />`,
+  template: `<VBackToSearchResultsLink v-bind="args" :id="id" />`,
   components: { VBackToSearchResultsLink },
   setup() {
     return { args }
--- a/frontend/src/types/analytics.ts
+++ b/frontend/src/types/analytics.ts
@@ -1,4 +1,4 @@
-import type { MediaType } from "~/constants/media"
+import type { MediaType, SearchType } from "~/constants/media"
 import type { ReportReason } from "~/constants/content-report"
 
 /**
@@ -34,8 +34,8 @@ export type Events = {
   BACK_TO_SEARCH: {
     /** The unique ID of the media */
     id: string
-    /** The media type being searched */
-    mediaType: MediaType
+    /** The content type being searched (can include All content) */
+    searchType: SearchType
   }
   /**
    * Description: The user clicks the CTA button to the external source to use the image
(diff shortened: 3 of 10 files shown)


sarayourfriend (changes requested):
This is on the right path, but needs at least one additional change to make it work. Please pass the new id and media type props to the component in the `image/_id/index.vue` and `audio/_id/index.vue` page components. Please also add a unit test to confirm the analytics event is called when the component is clicked.

If possible, please also test the change locally to verify that it is working:

1. Run `just frontend/up && `just frontend/init`
2. Run the app locally: `just frontend/run dev`
3. Log into plausible by visiting http://0.0.0.0:50288 and using the default username and password `deploy@example.com` and `deploy`
4. Add the custom event as a goal in http://0.0.0.0:50288/localhost/settings/goals
5. Verify that clicking the back to search link populates the goal on the main Plausible page for the localhost site (http://0.0.0.0:50288/localhost) at the bottom under the Goal Conversions section

@dhruvkb I'm writing a page for the docs site with this basic setup information so we can link it on these issues for adding new events. Otherwise, anyone who didn't review the PRs adding Plausible would be a little lost on how to start testing it locally.

It would probably also be worthwhile to have a quick PR to set up testing utilities like an easy mock to use to test the function is called as expected, to make it easier for new contributors, as the issues are marked as "good first issue".

masif2002:
Will make the changes and test it locally @sarayourfriend 

I have never written unit tests before. Is there something that can get me started on this?

sarayourfriend:
Sure. Openverse uses `jest`, a popular JavaScript testing library: https://jestjs.io/

Their documentation is fairly comprehensive, and we have lots of examples in our repository for other tests. A good basic example for Openverse specifically is the test suite for `VLicense`: https://github.com/wordpress/openverse/blob/7c92ba5bdff84b2771dc25a946bb371600906e0e/frontend/test/unit/specs/components/v-license.spec.js#L1

We also use [`testing-library`](https://testing-library.com/docs/vue-testing-library/intro/) which you would need to use in the case of the tests for this change by causing a click event on the rendered component. The [basic example for Vue testing](https://testing-library.com/docs/vue-testing-library/examples#basic-example) includes an example for how to fire click events on a rendered component.

You'll need to "spy" on the `sendCustomEvent` function that is returned by `useAnalytics`. To do that, it will be necessary to mock `useAnalytics`, similar to this `useI18n` mock from our own test suite: https://github.com/wordpress/openverse/blob/7c92ba5bdff84b2771dc25a946bb371600906e0e/frontend/test/unit/specs/components/AudioTrack/v-audio-track.spec.js#L38-L43

Something like:

```js
jest.mock("~/composables/use-analytics", () => ({
  useAnalytics: jest.fn(() => ({
    sendCustomEvent: jest.fn()
  })
})
```

Then you can import the mocked `useAnalytics` function to retrieve the mocked `sendCustomEvent` and follow jest's documentation for asserting the calls work as expected: https://jestjs.io/docs/mock-functions

The mock for `useI18n` is an instructive example.

frontend/src/components/VBackToSearchResultsLink.vue
```diff
@@ -3,14 +3,18 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 text-xs font-semibold text-dark-charcoal-70 pe-3 hover:text-dark-charcoal"
     v-bind="$attrs"
+    @click="handleClick()"
```
dhruvkb, line 6: When referring to an event handler function, the parenthesis are not used (because that would invoke the function).

```suggestion
    @click="handleClick"
```

frontend/src/components/VBackToSearchResultsLink.vue
```diff
@@ -26,9 +30,37 @@ export default defineComponent({
     VIcon,
     VLink,
   },
+  props: {
```
dhruvkb, line 33: These must be added below the `inheritAttrs` field to match the [recommended option order](https://eslint.vuejs.org/rules/order-in-components.html).

Also, if you add new required props to a component, you will need to update the usage of that component. This particular component `VBackToSearchResultsLink.vue` is used in the following locations, which need to be updated:

- `frontend/src/pages/audio/_id/index.vue`
- `frontend/src/pages/image/_id/index.vue`
- `frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx`

frontend/src/types/analytics.ts
```diff
@@ -23,6 +24,17 @@ export type Events = {
     /** the identifier of the image */
     identifier: string
   }
+  /**
+   * Click on the 'back to search' link on a single result
+   * 
+   * - Are these links used much? Are they necessary?
+   */
+  BACK_TO_SEARCH: {
+    /** The unique ID of the media */
```
dhruvkb, line 33: Small grammar nit.

```suggestion
    /** the unique ID of the media */
```

frontend/src/types/analytics.ts
```diff
@@ -23,6 +24,17 @@ export type Events = {
...
+    id: string
+    /** The media type being searched */
```
dhruvkb, line 35: Small grammar nit.

```suggestion
    /** the media type being searched */
```

dhruvkb (commented):
@masif2002 thanks for the PR, it's on the right track but needs a little more work before it can be merged.

masif2002:
Hey @sarayourfriend .. I tried testing the changes locally. I set the `BACK_TO_SEARCH` event as a goal, but it does not show up in the Goal Conversions section :/

sarayourfriend:
>  I tried testing the changes locally. I set the BACK_TO_SEARCH event as a goal, but it does not show up in the Goal Conversions section :/

Have you confirmed that the `sendCustomEvent` function is being called as expected? I won't be able to take a closer look at this PR until next Tuesday, but maybe @dhruvkb can help debug if there's a deeper issue.

masif2002:
> Have you confirmed that the `sendCustomEvent` function is being called as expected? 
Nope, the function is not invoked on click.. I'm trying to figure it out

frontend/src/components/VBackToSearchResultsLink.vue
```diff
@@ -3,14 +3,18 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 text-xs font-semibold text-dark-charcoal-70 pe-3 hover:text-dark-charcoal"
     v-bind="$attrs"
+    @click="handleClick"
```
dhruvkb, line 6: I checked out the PR locally and it seems like this handler is not being called, despite being perfectly ok from a code standpoint. In fact, the identical code from a different component `VHomeGallery.vue` works just as expected.

https://github.com/WordPress/openverse/blob/5106dc3e73d0b55f8eb0bfb8375c103c93402c1a/frontend/src/components/VHomeGallery/VHomeGallery.vue#L32

dhruvkb (commented):
The code looks fine but for some reason, attaching the `@click` handler to the `VLink` does not work here. I'll tag @obulat for help in debugging this issue.

frontend/src/components/VBackToSearchResultsLink.vue
masif2002: Yes Dhruv, the `CLICK_HOME_GALLERY_IMAGE` gets fired as expected, but `BACK_TO_SEARCH` event doesn't

dhruvkb: We'll try to investigate, but in the meantime please feel free to work on another issue if you'd like.

masif2002: > We'll try to investigate, but in the meantime please feel free to work on another issue if you'd like.

Sure, thanks!

obulat (commented):
Thank you for your contribution, @masif2002 !

I've tried debugging the issue locally. The link's click event is fired after the page transition is complete and the button is already unmounted. That's why the analytics event is not fired.

I found a (somewhat hacky) solution: we can add a `mousedown.native` event to the `VLink` component, and use that for firing the analytics event.

Here are the changes that worked for me locally:
```
--- a/frontend/src/components/VBackToSearchResultsLink.vue
+++ b/frontend/src/components/VBackToSearchResultsLink.vue
-    @click="handleClick"
+    @mousedown="handleClick"
   >
     <VIcon :icon-path="chevronIcon" :rtl-flip="true" />
     {{ $t("single-result.back") }}
diff --git a/frontend/src/components/VLink.vue b/frontend/src/components/VLink.vue
--- a/frontend/src/components/VLink.vue
+++ b/frontend/src/components/VLink.vue
@@ -5,6 +5,7 @@
     :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
     :to="linkTo"
     v-on="$listeners"
+    @mousedown.native="$emit('mousedown', $event)"
     @click.native="$emit('click', $event)"
   >
```

masif2002:
> I've tried debugging the issue locally. The link's click event is fired after the page transition is complete and the button is already unmounted. That's why the analytics event is not fired.

Awesome, that's good to know. How did you find out that the event is fired only after the page transition?

masif2002:
> I found a (somewhat hacky) solution: we can add a mousedown.native event to the VLink component, and use that for firing the analytics event.

Yea this one works, thank you!
@dhruvkb Should i commit the changes?

dhruvkb:
@obulat how does this work in `VHomeGallery.vue`?

obulat:
> > I've tried debugging the issue locally. The link's click event is fired after the page transition is complete and the button is already unmounted. That's why the analytics event is not fired.
> 
> Awesome, that's good to know. How did you find out that the event is fired only after the page transition?

In the `VLink` component, I added a route watcher that logged when the route changed, and an `onUnmounted` that logged when the link was unmounted, with its href. Not the most optimal way of debugging, but it worked :)

masif2002:
> In the `VLink` component, I added a route watcher that logged when the route changed, and an `onUnmounted` that logged when the link was unmounted, with its href. Not the most optimal way of debugging, but it worked :)

Wow, that's great! But as @dhruvkb said, how does the analytics event work as is in `VHomeGallery.vue` because there is also a page transition happening on click

obulat:
> @obulat how does this work in `VHomeGallery.vue`?

I think I found the answer to this. When you click on a link on the home page, you navigate to a different page that has a different **layout**. The same is true when you click on a page link from a search page. However, here, we stay in the **same layout**. This is the only difference between the events.

sarayourfriend:
@obulat Does that mean that when navigating between base layouts, Vue is, for some reason, keeping the previous layout mounted behind the scenes until it finishes triggering the event? It's interesting that that is the only difference, but I'm super curious _why_ having different layouts between the navigated pages causes the difference in behaviour... :thinking:

obulat:
> @obulat Does that mean that when navigating between base layouts, Vue is, for some reason, keeping the previous layout mounted behind the scenes until it finishes triggering the event? It's interesting that that is the only difference, but I'm super curious _why_ having different layouts between the navigated pages causes the difference in behaviour... 🤔

Apparently, this is a regression that happened in Vue 2.7.x: https://github.com/nuxt/nuxt/issues/10593 and https://github.com/vuejs/vue/issues/12781. I think with a `mousedown` we have a good enough workaround and don't need to investigate further.

frontend/src/components/VBackToSearchResultsLink.vue
```diff
@@ -3,14 +3,18 @@
   <VLink
     class="time inline-flex flex-row items-center gap-2 rounded-sm p-2 pe-3 text-xs font-semibold text-dark-charcoal-70 hover:text-dark-charcoal"
     v-bind="$attrs"
+    @click="handleClick"
```
obulat, line 6:
```suggestion
    @mousedown="handleClick"
```

obulat (commented):
@masif2002, I rebased your PR onto main.
Because the CI did not start, I replaced the `@click` with `@mousedown` to trigger it. We might need to change it after testing.

This PR needs unit tests. Are you interested in writing them? You can find examples in https://github.com/WordPress/openverse/pull/1180. If not, please let us know and we can help finish your work in this PR.

masif2002:
Hey @obulat ! I'm swamped right now and won't be able to write the unit tests for this PR. Thanks for offering to help out, though - please go ahead and take care of it.

obulat:
> Hey @obulat ! I'm swamped right now and won't be able to write the unit tests for this PR. Thanks for offering to help out, though - please go ahead and take care of it.

Thank you for quick response, @masif2002! I'll add tests, then. You are always welcome to contribute when you have more time 😃

frontend/src/components/meta/VBackToSearchResultsLink.stories.mdx
```diff
@@ -8,7 +8,7 @@ import VBackToSearchResultsLink from "~/components/VBackToSearchResultsLink.vue"
 />
 
 export const Template = (args) => ({
-  template: `<VBackToSearchResultsLink v-bind="args"/>`,
+  template: `<VBackToSearchResultsLink v-bind="args" :id="id" :media-type="mediaType" />`,
```
obulat, line 11:
```suggestion
  template: `<VBackToSearchResultsLink v-bind="args" :id="id" />`,
```

obulat (approved):

frontend/test/unit/setup.js
```diff
@@ -19,7 +19,13 @@ Vue.use(VueI18n)
...
 config.stubs["nuxt-link"] = Vue.component("NuxtLink", {
   props: ["to"],
-  template: '<a :href="to" v-on="$listeners"><slot /></a>',
+  methods: {
+    handleClick() {
+      this.$emit("mousedown")
+      this.$emit("click", new MouseEvent("click"))
+    },
+  },
+  template: '<a :href="to" v-on="$listeners" @click="handleClick"><slot /></a>',
```
sarayourfriend, line 28: Can a note be added that this adaptation is necessary to mimic NuxtLink's behaviour (and fix a bug)? If we can remove this after the Nuxt 3 upgrade, noting that as well would be nice.

frontend/test/unit/specs/components/v-back-to-search-results-link.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+const localVue = createLocalVue()
+localVue.use(PiniaVuePlugin)
+describe("VBackToSearchResultsLink", () => {
+  it("should send analytics event when clicked", async () => {
+    const sendCustomEventMock = jest.fn()
+
+    useAnalytics.mockImplementation(() => ({
+      sendCustomEvent: sendCustomEventMock,
+    }))
```
sarayourfriend, line 24: Doesn't this duplicate the mock in `jest.mock` call? It seems like one could be removed in favour of the other.

sarayourfriend (approved):
LGTM. Just small questions but no blockers. Questions are for @obulat, to be clear.

frontend/test/unit/specs/components/v-back-to-search-results-link.spec.js
obulat: Fixed :)

PR #1125, opened 2023-04-03
Prevent non-uniform event names or overwriting default props

sarayourfriend:
## Fixes

There is no issue for this. I don't think it is absolutely necessary and was mostly motivated by wondering if this was even possible to do. If @dhruvkb or @obulat do not feel this is worthwhile to add or maintain, then we can close the PR, no problem.

## Description

Validates the `Events` type by doing some TypeScript magic to validate the following:

- That all event names are in uppercase. I started adding validation to check that `-` and space were not used as separators, but that felt unnecessary and made things even more complicated.
- That payload types do not include prop names that overlap with the default/reserved prop names.

When either of these things is the case, the `EventsOrErrors` type will be a string that has descriptive content for help identifying the problem. Examples:

Error message indicating usage of reserved payload prop names

Error message indicating both usage of reserved payload prop names and incorrect event names

Error message indicating incorrect event names

These error strings won't show as TS will only be complaining that the string type is not assignable to `BaseEventsType`. The error messages are only visible when inspecting the types in an editor with type help like VSCode. I don't know if WebStorm behaves similarly and would reveal the underlying string.

I've also gone ahead and made it possible to have custom events with zero payload which will be necessary for the `OPEN_PAGES_MENU` event. **This part of the PR I believe we should keep regardless as it is fundamentally necessary for implementing the full list of events.**

## Testing Instructions

Checkout the branch and make changes to the Events configuration object that would cause an invalid configuration.

Try out the changes to the `sendCustomEvent` function after adding some new events, including ones that have `never` payload.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/types/validate-analytics.ts
+++ b/frontend/src/types/validate-analytics.ts
@@ -1,0 +1,70 @@
+import type { Events } from "~/types/analytics"
+
+import type { UnionToIntersection } from "type-fest"
+
+type ReservedPropNames =
+  | "timestamp"
+  | "language"
+  | "breakpoint"
+  | "ua"
+  | "os"
+  | "platform"
+  | "browser"
+  | "version"
+  | "origin"
+  | "pathname"
+  | "referrer"
+  | "width"
+  | "height"
+
+type BaseEventsType = Record<string, Record<string, string | number | boolean>>
+
+type MergePayloadTypes<E extends BaseEventsType> = UnionToIntersection<
+  E[keyof E]
+>
+type ValidatePayloads<
+  E extends BaseEventsType,
+  Payloads = MergePayloadTypes<E>
+> = Omit<Payloads, ReservedPropNames> extends Payloads
+  ? E
+  : "Payloads include reserved prop names."
+
+type ExtractEventNames<E> = E extends Record<infer K, unknown>
+  ? K extends string
+    ? K
+    : never
+  : never
+
+type ValidateEventNames<E> = ExtractEventNames<E> extends Uppercase<
+  ExtractEventNames<E>
+>
+  ? E
+  : "Event names must be in SCREAMING_SNAKE_CASE."
+
+type OnlyErrors<A, B> = A extends string
+  ? B extends string
+    ? A | B
+    : A
+  : B extends string
+  ? B
+  : boolean
+
+type ValidateEvents<
+  E extends BaseEventsType,
+  ValidatedEventNames = ValidateEventNames<E>,
+  ValidatedPayloads = ValidatePayloads<E>,
+  Errors = OnlyErrors<ValidatedEventNames, ValidatedPayloads>
+> = Errors extends string ? Errors : E
+
+type IsRecord<E extends BaseEventsType> = E
+
+type EventsOrErrors = ValidateEvents<Events>
+
+/**
+ * If `Events` is not a valid type parameter to `IsRecord` then
+ * the configuration is invalid. `Events` will be set to string
+ * types represnting the specific error case(s) present.
+ *
+ * Exported to avoid needing to ignore the unused variable error.
+ */
+export type IsValid = IsRecord<EventsOrErrors>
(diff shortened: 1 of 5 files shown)


frontend/src/types/validate-analytics.ts
```diff
@@ -0,0 +1,70 @@
+import type { Events } from "~/types/analytics"
+
+import type { UnionToIntersection } from "type-fest"
+
+type ReservedPropNames =
```
dhruvkb, line 5: If we break these into WindowPropNames and IsomorphicPropNames, they can be used as type hints for [`windowProps`](https://github.com/WordPress/openverse/pull/1125/files#diff-d6fe83def5944105827719ecdad03ddd3b8b989855454f17c241c571a0c1f4c1L38) and [`isomorphicProps`]((https://github.com/WordPress/openverse/pull/1125/files#diff-d6fe83def5944105827719ecdad03ddd3b8b989855454f17c241c571a0c1f4c1L19), making it easier to keep this list in sync with the actual props.

frontend/src/types/validate-analytics.ts
```diff
@@ -0,0 +1,70 @@
...
+
+type MergePayloadTypes<E extends BaseEventsType> = UnionToIntersection<
+  E[keyof E]
+>
+type ValidatePayloads<
+  E extends BaseEventsType,
+  Payloads = MergePayloadTypes<E>
+> = Omit<Payloads, ReservedPropNames> extends Payloads
+  ? E
+  : "Payloads include reserved prop names."
```
dhruvkb, line 30: When I was writing the functions, the ability to override the default props was considered a feature. One such use was when I was sending a page navigation event from the router middleware where we want `pathname` to be the new page being navigated to, but `window.location.pathname` gives the current page being navigated from. So we wanted to be able to supply that and override the default.

Considering it's not used anymore we can add this restriction, I'm just adding some historical context.

frontend/src/types/validate-analytics.ts
```diff
@@ -0,0 +1,70 @@
...
+> = Errors extends string ? Errors : E
+
+type IsRecord<E extends BaseEventsType> = E
+
+type EventsOrErrors = ValidateEvents<Events>
+
+/**
+ * If `Events` is not a valid type parameter to `IsRecord` then
+ * the configuration is invalid. `Events` will be set to string
+ * types represnting the specific error case(s) present.
```
dhruvkb, line 66:
```suggestion
 * types representing the specific error case(s) present.
```

dhruvkb (approved):
There is some TS-fu in `validate-analytics.ts` that I don't completely understand but I can attest that the additional checks are helpful and that the changes are a net improvement 👍.

I wonder if ESLint rules would have been more appropriate for these instead of TS checks, but it's all part of the same CI 'Lint' job anyway.

frontend/src/types/validate-analytics.ts
sarayourfriend: I see :thinking: From my perspective I would hope that any of the "default", "always there" props always have the same meaning for every event, rather than having to remember that their meaning can vary. If we run into absolutely needing to override the meaning of a default prop, perhaps we can deal with it then, so that we understand the use-case?

sarayourfriend:
> I wonder if ESLint rules would have been more appropriate for these instead of TS checks, but it's all part of the same CI 'Lint' job anyway.

Oh dear, no you're right. ESLint _is_ the correct answer for this :facepalm: Not just in the fact that the code will be easier for someone who hasn't spent Too Much Time thinking about TypeScript (me) but also because we can have _actually descriptive error messages_ when things aren't configured correctly :facepalm: I'll close this PR for now. TS is an _interesting_ way to approach this problem, but it's not the _right_ way, at least not considering how much clearer and easy to understand an ESLint message would be.

PR #1151, opened 2023-04-07
WIP: Content switcher analytics event

### LINKED ISSUE #1071: Analytics event: `CHANGE_CONTENT_TYPE`
## Description

The goal is to 

- add the type for event `CHANGE_CONTENT_TYPE` to the `Events` type.

## Reference

https://github.com/WordPress/openverse/blob/839418fabb0972ec3c3ccf48c72fd64cbba03cb9/rfcs/analytics/20221006-implementation_plan_frontend.md?plain=1#L187

## Additional information

- [RFC](https://github.com/WordPress/openverse/blob/main/rfcs/analytics/20221006-implementation_plan_frontend.md#events)

zackkrida:
## Fixes

Fixes #1071 by @dhruvkb 

## Description

- Adds a new `useComponentName` composable
- Adds the `CHANGE_CONTENT_TYPE` event and fires it on use of the `VSearchType` and `VContentLink` components
- Adds dev logging to the `sendCustomEvent` method

### TODO

- [ ] Add tests
- [ ] Figure out the terrible `VLink` click event situation

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/composables/use-component-name.ts
+++ b/frontend/src/composables/use-component-name.ts
@@ -1,0 +1,10 @@
+import { getCurrentInstance } from "vue"
+
+/**
+ * Retrieve the name of the component in which
+ * the composable is being used.
+ */
+export const useComponentName = () => {
+  const vm = getCurrentInstance()
+  return vm?.proxy.$options.name
+}
--- a/frontend/src/composables/use-search-type.ts
+++ b/frontend/src/composables/use-search-type.ts
@@ -16,6 +16,10 @@ import { useFeatureFlagStore } from "~/stores/feature-flag"
 
 import { useI18n } from "~/composables/use-i18n"
 
+import { useAnalytics } from "~/composables/use-analytics"
+
+import { useComponentName } from "./use-component-name"
+
 import allIcon from "~/assets/icons/all-content.svg"
 import audioIcon from "~/assets/icons/audio-content.svg"
 import imageIcon from "~/assets/icons/image-content.svg"
@@ -40,6 +44,9 @@ const labels = {
 
 export default function useSearchType() {
   const i18n = useI18n()
+  const componentName = useComponentName()
+  const analytics = useAnalytics()
+
   const activeType = computed(() => useSearchStore().searchType)
 
   const previousSearchType = ref(activeType.value)
@@ -53,6 +60,12 @@ export default function useSearchType() {
 
   const setActiveType = (searchType: SearchType) => {
     if (previousSearchType.value === searchType) return
+
+    analytics.sendCustomEvent("CHANGE_CONTENT_TYPE", {
+      previous: previousSearchType.value,
+      next: searchType,
+      component: componentName ?? "Unknown",
+    })
     useSearchStore().setSearchType(searchType)
     previousSearchType.value = searchType
   }
--- a/frontend/src/types/analytics.ts
+++ b/frontend/src/types/analytics.ts
@@ -1,3 +1,4 @@
+import type { SearchType } from "~/constants/media"
 /**
  * compound type of all custom events sent from the site; Index with `EventName`
  * to get the type of the payload for a specific event.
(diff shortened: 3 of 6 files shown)


sarayourfriend:
@zackkrida Is this PR blocked?

zackkrida:
@WordPress/openverse-frontend apologies but I am going to need to hand this off to someone else. We can assign in the team meeting tomorrow 👍

zackkrida:
Thanks for working on this, @obulat 😄

frontend/src/composables/use-component-name.ts
```diff
@@ -0,0 +1,10 @@
+import { getCurrentInstance } from "vue"
+
+/**
+ * Retrieve the name of the component in which
+ * the composable is being used.
+ */
+export const useComponentName = () => {
+  const vm = getCurrentInstance()
+  return vm?.proxy.$options.name
```
dhruvkb, line 9: I don't think `name` is a required property so having a fallback in case it's `undefined` might be a good idea. A lint check to ensure all our components have names would be a better idea but that might be more work.

frontend/src/composables/use-search-type.ts
```diff
@@ -47,6 +54,12 @@ export default function useSearchType() {
 
   const setActiveType = (searchType: SearchType) => {
     if (previousSearchType.value === searchType) return
+
+    analytics.sendCustomEvent("CHANGE_CONTENT_TYPE", {
+      previous: previousSearchType.value,
+      next: searchType,
+      component: componentName ?? "Unknown",
```
dhruvkb, line 61: `?? "Unknown"` should be moved to the composable itself.

dhruvkb (approved):
LGTM! I have one non-blocking suggestion.

frontend/src/types/analytics.ts
```diff
@@ -101,6 +101,21 @@ export type Events = {
...
+   *   - Is there interest in the non-supported content types?
+   *   - Do users switch content types? Where in their journeys?
+   */
+  CHANGE_CONTENT_TYPE: {
+    /** The previously-set media type */
+    previous: SearchType
+    /** The new media type */
+    next: SearchType
+    /** The name of the Vue component used to switch content types. */
+    component: string
```
sarayourfriend, line 117: I kind of wonder about using the Vue component name here rather than more stable names. We don't change component names often but the header dropdown I think has had its components changed during the redesign. In cases like that, the name of the component might change, even though the interaction area stayed the same :thinking: 

This also makes me wonder whether we should send the feature flag state in the payload for all events? If we have a feature flag that toggles a particular bit of UI or a feature on or off and it suddenly changes how a custom event trends, it would be nice to see if the feature flag preferences were changed for those events.

It isn't an issue now because I don't think we have any feature flags that would qualify for that, but something to consider for the future, especially if we use things like component names which are not necessarily stable, even if the UI area they're inside of remains stable.

On the other hand, using the component name _might_ be a good way to tell the same cases as those feature flags, but probably contingent on the implementation.

No real suggestion for a change here, just some musings about how to find stable identifiers and thing carefully about exactly what each payload item is meant to help us do. If we ask the question, why do we want to know the component name for this event, does the answer lead us to think that a different approach to this payload field would be appropriate?

frontend/src/composables/use-component-name.ts
sarayourfriend: Another thought: should `useComponentName` raise an error if it's used in a context where no component name exists? That would keep the return value stable `string` not `undefined | string`, but would be much stricter than this implementation.

Falling back to "Unknown" without it being intentional, especially if this is used to disambiguate events, doesn't seem like a good long-term practice.

sarayourfriend (changes requested):
This works fine locally, but as with the other event PRs, I have some questions about the payload data, what we're trying to identify with them, and whether we need to tighten the requirements for the `useComponentName` composable or choose a different approach altogether for disambiguating this particular event.

AetherUnbound (approved):
I was able to test both search type changes locally and see that they appeared in the logs, great idea adding that!

frontend/src/composables/use-component-name.ts
obulat: I would like to have a stricter implementation, but when we test the `use-search-type` in unit tests, there is no component, so this will throw errors. Maybe we could create a new issue to see if we have many "Unknown" components in analytics reports, and if we do - fix it?

zackkrida:
@obulat this has two approving reviews despite some dangling feedback from Sara. Could you review and merge, making any necessary follow-up issues?

PR #1154, opened 2023-04-07
Implementation Plan: Update staging database

AetherUnbound:
## Due date:

2023-05-03

## Assigned reviewers

- [x] @krysal
- [x] @stacimc

## Description

This is an implementation plan for the first portion of #1107.

## Current round

This discussion is following the Openverse decision-making process. Information about this process can be found [on the Openverse documentation site](https://docs.openverse.org/reference/decision_making/index.html). Requested reviewers or participants will be following this process. If you are being asked to give input on a specific detail, you do not need to familiarise yourself with the process and follow it.

The deadline for review for this round is **May 3rd**

### DIFF
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -105,7 +105,48 @@ This will be accomplished by:
       until the database removal is complete.
     - `SkipFinalSnapshot` should be set to `True` to avoid creating a final
       snapshot of the database before deletion.
-12. Report the success or failure of the DAG run to Slack.
+12. Once the database rename is complete, the following actions will also need
+    to occur (they can occur simultaneously in the DAG):
+    1. The staging API must be deployed to apply any necessary migrations to the
+       new copy the database. In order to do this we must augment the existing
+       [GitHub API class](https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/common/github.py#L4)
+       with additional functions for **getting the latest GHCR image tag** and
+       **triggering the deployment workflow**. Once those functions exist, the
+       steps for this piece would be:
+       1. Get the latest image tag for the
+          [`openverse-api` package](https://github.com/WordPress/openverse/pkgs/container/openverse-api).
+       2. Trigger the deployment workflow
+          [in a similar manner to the existing CI/CD workflow](https://github.com/WordPress/openverse/blob/1f7b8e670a0f7812494375570d076a7c33142062/.github/workflows/ci_cd.yml#L833-L846)
+          using the tag from the previous step and the `openverse-bot` actor.
+    2. Update the Elasticsearch indices corresponding to each media type. This
+       can be done by using the
+       [`UPDATE_INDEX` action](https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L314)
+       on the ingestion server. The steps for this piece would be:
+       1. Get the date for the last successful run of the update DAG (see the
+          [`_month_check` function](https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/data_refresh/dag_factory.py#L85)
+          from the data refresh DAG for a similar example).
+       2. For each media type:
+          1. Get the current index (see the
+             [`get_current_index` task](https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/data_refresh/data_refresh_task_factory.py#L167-L175)
+             on the data refresh DAG)
+          2. Initiate the
+             [`UPDATE_INDEX` action](https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/api.py#L107-L113)
+             using the date and index suffix retrieved above
+          3. Wait for the index update to complete
+    3. Truncate the OAuth tables to prevent production API applications from
+       working in staging. Each of these truncate operations can be run
+       simultaneously using
+       `TRUNCATE TABLE <table_name> RESTART IDENTITY CASCADE;` in a
+       [`PGExecuteQueryOperator`](https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/common/sql.py#L137).
+       The tables that need to be truncated are:
+       - `api_throttledapplication`
+       - `api_oauth2registration`
+       - `api_oauth2verification`
+       - `oauth2_provider_accesstoken`
+       - `oauth2_provider_grant`
+       - `oauth2_provider_idtoken`
+       - `oauth2_provider_refreshtoken`
+13. Report the success or failure of the DAG run to Slack.
 
 ### Django Admin UI changes
 
@@ -218,8 +259,15 @@ unfamiliar with Airflow.
 
 <!-- What, if any, work within this plan can be parallelized? -->
 
-All of this work will be done as part of a single PR, so it is not
-parallelizable.
+Based on the above steps, the following work can be parallelized:
+
+1. Alterations to the
+   [GitHub API class](https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/common/github.py#L4)
+   to include new methods for pulling the latest GHCR image tag and for
+   triggering a workflow.
+2. The addition of the notification banner to the Django Admin UI.
+
+After the above work is complete, the primary DAG can be written and tested.
 
 ## Rollback
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -1,0 +1,85 @@
+# Implementation Plan: Update Staging Database
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @stacimc
+- [ ] @krysal
+
+## Project links
+
+<!-- Enumerate any references to other documents/pages, including milestones and other plans -->
+
+- [Project Thread](https://github.com/WordPress/openverse/issues/392)
+- [Project Proposal]() _TBD_
+
+## Overview
+
+<!-- A brief one or two sentence overview of the implementation being described. -->
+
+This document describes how we will implement a mechanism for updating the
+staging database with the latest data from the production database.
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+The final product of this plan will be a DAG (scheduled for `@monthly`) which
+will recreate the staging database from the most recent snapshot of the
+production API database. This will be accomplished by:
+
+1.
+
+## Dependencies
+
+### Tools & packages
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+### Other projects or work
+
+<!-- Note any projects this plan is dependent on. -->
+
+## Design
+
+<!-- Note any design requirements for this plan. -->
+
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+## Blockers
+
+<!-- What hard blockers exist which might prevent further work on this project? -->
+
+## API version changes
+
+<!-- Explore or mention any changes to the API versioning scheme. -->
+
+## Accessibility
+
+<!-- Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+
+## Rollback
+
+<!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
+
+## Privacy
+
+<!-- How does this approach protect users' privacy? -->
+
+## Localization
+
+<!-- Any translation or regional requirements? Any differing legal requirements based on user location? -->
+
+## Risks
+
+<!-- What risks are we taking with this solution? Are there risks that once taken can’t be undone?-->
+
+## Prior art
+
+<!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -44,8 +44,9 @@ production API database. This will be accomplished by:
    [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html)
    function (an Airflow operator does not exist for this operation). Namely, the
    following attributes should be gathered and set to ensure they match:
-   - Availability should be a single DB instance (not multi-AZ)
-   - It should be attached to our default VPC
+   - Availability should be a single DB instance (not
+     [multi-AZ](https://docs.aws.amazon.com/fsx/latest/WindowsGuide/high-availability-multiAZ.html))
+   - It should be attached to our default Virtual Private Cloud (VPC)
    - It should not be publicly accessible
    - It will need access to the following VPC security groups: `default`,
      `openverse-vpc-db-access`, `staging-dev-rds-sg`
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -15,7 +15,7 @@
 <!-- Enumerate any references to other documents/pages, including milestones and other plans -->
 
 - [Project Thread](https://github.com/WordPress/openverse/issues/392)
-- [Project Proposal]() _TBD_
+- [Project Proposal](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html)
 
 ## Overview
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -40,48 +40,58 @@ production API database. This will be accomplished by:
 2. In the case where the snapshot is not yet available, wait for the status of
    the most recent snapshot using the
    [`RdsSnapshotExistenceSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsSnapshotExistenceSensor).
-3. Create a new database from this snapshot using boto3's
-   [`restore_db_instance_from_db_snapshot`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/restore_db_instance_from_db_snapshot.html)
-   function (an operator does not exist for this operation). This database will
-   be named in a way that does not conflict with the existing dev database name,
-   e.g. `dev-next-openverse-db`. Careful consideration will need to be made when
-   drafting this step to ensure that the generated database matches the settings
-   for the existing database exactly. Specifically, the following should be
-   matched:
+3. In parallel, gather the attributes of the **staging** database using boto3's
+   [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html)
+   function (an Airflow operator does not exist for this operation). Namely, the
+   following attributes should be gathered and set to ensure they match:
    - Availability should be a single DB instance (not multi-AZ)
    - It should be attached to our default VPC
    - It should not be publicly accessible
    - It will need access to the following VPC security groups: `default`,
      `openverse-vpc-db-access`, `staging-dev-rds-sg`
-   - It should match the dev database's instance size (currently `m5.xlarge`)
+   - It should match the staging database's instance size (currently
+     `m5.xlarge`)
    - It should have 3000 GB of allocated gp2 storage
    - It should use password authentication (this may not need to be changed from
      the default, as the snapshot should contain the same connection
      information)
-4. Wait for the new database to be ready using the
+4. Create a new database from this snapshot using boto3's
+   [`restore_db_instance_from_db_snapshot`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/restore_db_instance_from_db_snapshot.html)
+   function (an operator does not exist for this operation). This database will
+   be named in a way that does not conflict with the existing staging database
+   name, e.g. `dev-next-openverse-db`. The database configuration information
+   from the previous step will be used to ensure the new database matches the
+   old database's configuration exactly.
+5. Wait for the new database to be ready using the
    [`RdsDbSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsDbSensor)
    ([example](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/rds.html#howto-sensor-rdsdbsensor)).
-5. Rename the old staging database to `dev-old-openverse-db` using boto3's
+6. Rename the old staging database to `dev-old-openverse-db` using boto3's
    [`modify_db_instance`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/modify_db_instance.html)
    function (an operator does not exist for this operation). We must set the
    `ApplyImmediately` option here to `True` to ensure this operation happens
    immediately rather than waiting for the next maintenance window.
-6. Wait for the old database rename to be complete with the `RdsDbSensor` (we
+7. Wait for the old database rename to be complete with the `RdsDbSensor` (we
    may need retries on this step, since the database may not be initially
    available/named when the sensor first starts). _**Note**: this will cause a
    temporary outage of the staging API, see
-   [the alternatives section](#alternatives) for why this is the case._
-7. Rename the new database to `dev-openverse-db` using `modify_db_instance`.
-8. Wait for the new database rename to be complete with the `RdsDbSensor` (we
+   [the alternatives section](#alternatives) for why this is the case._ A Slack
+   notification should be sent out at the start of this step to alert the team
+   of the outage, and again once the outage is resolved.
+8. Rename the new database to `dev-openverse-db` using `modify_db_instance`.
+   (Noting that `ApplyImmediately` should be set to `True` here as well.)
+9. Wait for the new database rename to be complete with the `RdsDbSensor` (we
    may need retries on this step, since the database may not be initially
    available/named when the sensor first starts).
-9. If the previous steps fail, rename `dev-old-openverse-db` back to
-   `dev-openverse-db`. Otherwise, `dev-old-openverse-db` can be deleted using
-   the
-   [`RdsDeleteDbInstanceOperator`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsDeleteDbInstanceOperator)
-   (the `wait_for_completion` should be left as `True` here so we don't need a
-   follow-up sensor).
-10. Report the success or failure of the DAG run to Slack.
+10. If the previous steps fail, rename `dev-old-openverse-db` back to
+    `dev-openverse-db`. Otherwise, `dev-old-openverse-db` can be deleted using
+    the
+    [`RdsDeleteDbInstanceOperator`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsDeleteDbInstanceOperator).
+    We should use the following configuration options:
+    - `wait_for_completion` should be left as `True` here so this stop will hang
+      until the database removal is complete.
+    - `SkipFinalSnapshot` should be set to `True` to avoid creating a final
+      snapshot of the database before deletion.
+11. Report the success or failure of the DAG run to Slack.
 
 ## Dependencies
 
@@ -131,6 +141,13 @@ than try and avoid it.
 
 This DAG will need to have clear instructions for how to run it for those
 unfamiliar with Airflow.
+
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+All of this work will be done as part of a single PR, so it is not
+parallelizable.
 
 ## Rollback
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -63,7 +63,9 @@ production API database. This will be accomplished by:
    ([example](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/rds.html#howto-sensor-rdsdbsensor)).
 5. Rename the old staging database to `dev-old-openverse-db` using boto3's
    [`modify_db_instance`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/modify_db_instance.html)
-   function (an operator does not exist for this operation).
+   function (an operator does not exist for this operation). We must set the
+   `ApplyImmediately` option here to `True` to ensure this operation happens
+   immediately rather than waiting for the next maintenance window.
 6. Wait for the old database rename to be complete with the `RdsDbSensor` (we
    may need retries on this step, since the database may not be initially
    available/named when the sensor first starts). _**Note**: this will cause a
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
@@ -32,15 +32,22 @@ The final product of this plan will be a DAG (scheduled for `@monthly`) which
 will recreate the staging database from the most recent snapshot of the
 production API database. This will be accomplished by:
 
-1. Determine the most recent automated **production** snapshot using boto3's
+1. Use a new `SKIP_STAGING_DATABASE_RESTORE`
+   [Airflow Variable](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html)
+   to control whether the DAG should run or not. This will allow us to skip the
+   DAG in the case where we have made diverging chances to the staging database
+   for testing a new feature that we do not want to be overwritten. If this
+   Variable is set to `True`, the DAG should issue a Slack message notifying the
+   maintainers of the skipped run and raise an `AirflowSkipException`.
+2. Determine the most recent automated **production** snapshot using boto3's
    [`describe_db_snapshots`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_snapshots.html)
    function (an Airflow operator does not exist for this operation). This step
    should also check the `Status` value of the response for the most recent
    snapshot to ensure that it is `available`.
-2. In the case where the snapshot is not yet available, wait for the status of
+3. In the case where the snapshot is not yet available, wait for the status of
    the most recent snapshot using the
    [`RdsSnapshotExistenceSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsSnapshotExistenceSensor).
-3. In parallel, gather the attributes of the **staging** database using boto3's
+4. In parallel, gather the attributes of the **staging** database using boto3's
    [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html)
    function (an Airflow operator does not exist for this operation). Namely, the
    following attributes should be gathered and set to ensure they match:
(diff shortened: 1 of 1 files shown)


rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+     `openverse-vpc-db-access`, `staging-dev-rds-sg`
+   - It should match the dev database's instance size (currently `m5.xlarge`)
+   - It should have 3000 GB of allocated gp2 storage
+   - It should use password authentication (this may not need to be changed from
+     the default, as the snapshot should contain the same connection
+     information)
+4. Wait for the new database to be ready using the
+   [`RdsDbSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsDbSensor)
+   ([example](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/rds.html#howto-sensor-rdsdbsensor)).
+5. Rename the old staging database to `dev-old-openverse-db` using boto3's
```
sarayourfriend, line 64: Probably worth noting that we need to pass `ApplyImmediately` to the `modify_db_instance` calls, otherwise RDS will wait until the next maintenance window to make the name change.

It occurred to me that maybe we can just use the maintenance window to avoid some downtime. If we queue the rename of the old database, then queue the rename of the new one, RDS would handle the swap over for us and probably do so faster. However, deleting a db instance always happens immediately, so we can't queue it for the maintenance window _and_ even if we could, doing this would incur the cost of two RDS db instances until the maintenance window rather than just for the brief period of overlap time in this plan.

Anyway, just wanted to note the need for ApplyImmediately as it's not necessarily obvious reading the documentation and would be easy to miss.

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+   the most recent snapshot using the
+   [`RdsSnapshotExistenceSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsSnapshotExistenceSensor).
+3. Create a new database from this snapshot using boto3's
+   [`restore_db_instance_from_db_snapshot`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/restore_db_instance_from_db_snapshot.html)
+   function (an operator does not exist for this operation). This database will
+   be named in a way that does not conflict with the existing dev database name,
+   e.g. `dev-next-openverse-db`. Careful consideration will need to be made when
+   drafting this step to ensure that the generated database matches the settings
+   for the existing database exactly. Specifically, the following should be
+   matched:
```
sarayourfriend, line 50: These could all be copied from a call to [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html) for the existing staging RDS instance, that way we don't have to maintain these settings between the infrastructure repo _and_ this new DAG.

sarayourfriend (commented):
I'm not an assigned reviewer but had some small notes for things that could be clarified to avoid easy mistakes during implementation.

Also: the implementation plan guide [requests identifying the atomic blocks of work that can be split into individual PRs](https://github.com/WordPress/openverse/tree/main/docs/projects#implementation-plans-status-in-rfc). This plan looks like it would probably go into a single PR, despite its complexity. Does that sound right? If so, can it be noted explicitly?

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+This document describes how we will implement a mechanism for updating the
+staging database with the latest data from the production database.
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+The final product of this plan will be a DAG (scheduled for `@monthly`) which
+will recreate the staging database from the most recent snapshot of the
+production API database. This will be accomplished by:
```
stacimc, line 33: I was under the impression that the staging DB was meant to be a subset of the production data, although using the full snapshot would of course be great. I think I may have confused the api DB with the _elasticsearch index_, which _will_ be a subset. Is that right?

stacimc (commented):
Looks great, very clear! And nice to see how many existing operators we'll be able to take advantage of.

AetherUnbound:
Thank you both for your feedback & questions! I will be answering the questions and revising the document over the next few days.

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
AetherUnbound: Oh excellent call, I'll be sure to add this. That's good to be aware of, but yes in this case I'd rather just incur the downtime.

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+   temporary outage of the staging API, see
+   [the alternatives section](#alternatives) for why this is the case._
+7. Rename the new database to `dev-openverse-db` using `modify_db_instance`.
+8. Wait for the new database rename to be complete with the `RdsDbSensor` (we
+   may need retries on this step, since the database may not be initially
+   available/named when the sensor first starts).
+9. If the previous steps fail, rename `dev-old-openverse-db` back to
+   `dev-openverse-db`. Otherwise, `dev-old-openverse-db` can be deleted using
+   the
+   [`RdsDeleteDbInstanceOperator`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsDeleteDbInstanceOperator)
```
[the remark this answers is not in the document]
AetherUnbound: Oh yes we should! I'll make note of that here.

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
AetherUnbound: Excellent call!

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+   [`RdsDbSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsDbSensor)
+   ([example](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/rds.html#howto-sensor-rdsdbsensor)).
+5. Rename the old staging database to `dev-old-openverse-db` using boto3's
+   [`modify_db_instance`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/modify_db_instance.html)
+   function (an operator does not exist for this operation).
+6. Wait for the old database rename to be complete with the `RdsDbSensor` (we
+   may need retries on this step, since the database may not be initially
+   available/named when the sensor first starts). _**Note**: this will cause a
+   temporary outage of the staging API, see
+   [the alternatives section](#alternatives) for why this is the case._
```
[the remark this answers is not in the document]
AetherUnbound: That's a great idea!

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,166 @@
...
+7. Rename the new database to `dev-openverse-db` using `modify_db_instance`.
+8. Wait for the new database rename to be complete with the `RdsDbSensor` (we
+   may need retries on this step, since the database may not be initially
+   available/named when the sensor first starts).
+9. If the previous steps fail, rename `dev-old-openverse-db` back to
+   `dev-openverse-db`. Otherwise, `dev-old-openverse-db` can be deleted using
+   the
+   [`RdsDeleteDbInstanceOperator`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsDeleteDbInstanceOperator)
+   (the `wait_for_completion` should be left as `True` here so we don't need a
+   follow-up sensor).
```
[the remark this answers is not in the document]
AetherUnbound: I think my wording may be ambiguous here - the `wait_for_completion` step means that we avoid the need for a sensor _on the delete operation itself_. The `wait_for_completion` parameter is a feature on the Airflow operators themselves, unfortunately anything we do with the RDS client directly will require a follow-up sensor since I don't _think_ there's a similar mechanism available with the `boto3` API alone. I'll reword this!

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
AetherUnbound: Yes, that's correct! The database itself will have all data, but some of the indices we use may have a subset of that data (e.g. the proportional-by-provider index described in #1107). We have also made the staging database a subset of the production database in the past, but I don't think that's necessary to do going forward.

AetherUnbound:
> Also: the implementation plan guide [requests identifying the atomic blocks of work that can be split into individual PRs](https://github.com/WordPress/openverse/tree/main/docs/projects#implementation-plans-status-in-rfc). This plan looks like it would probably go into a single PR, despite its complexity. Does that sound right? If so, can it be noted explicitly?

Yes, that's correct - my thought was that this would be a single DAG. I'll make this explicit.

AetherUnbound:
@sarayourfriend  and @stacimc - I believe I've applied all of the feedback & clarification from your notes, please take another look when you have a moment!

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,185 @@
...
+
+- [ ] @stacimc
+- [ ] @krysal
+
+## Project links
+
+<!-- Enumerate any references to other documents/pages, including milestones and other plans -->
+
+- [Project Thread](https://github.com/WordPress/openverse/issues/392)
+- [Project Proposal]() _TBD_
```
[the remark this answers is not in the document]
zackkrida: Or perhaps https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html instead

zackkrida:
Hey, just jumping in here to share (and elaborate on) an alternative approach I mentioned in the [project proposal comments](https://github.com/WordPress/openverse/pull/1107#pullrequestreview-1385957729). This isn't meant to discredit the work here or suggest that it isn't a suitable solution. In considering this alternative solution though, I feel that it avoids some of the pitfalls of the proposed solution, providing that I understand the motivation of this project correctly.

If the staging database is just a copy of the production database, why do we even _need_ a staging database? I came up with the following reasons, trying to ignore anything that wasn't DB specific:

1. To test Django admin views and operations which impact Django DB tables
2. To test new tables for new media types

These to me suggest that we'd want the Django-specific tables to be persistent, and not wiped out weekly by restoring from production backup. What we really want is to only sync the media tables. Additionally, I believe access to these media tables only needs to be read-only, as API consumers and the Django admin dashboard do not actually make changes to the media tables.

I can think of three alternative approaches to reading the media from the production database in the staging database. These approaches have some benefits to the proposal here, along with their own tradeoffs:


| Name | Pros | Cons |
| --- | --- | --- |
| 1. Foreign Data Wrapper to point staging media tables to production | - Access to latest data in real-time<br>- Minimal code required<br>- Can downsize staging DB<br>- Simple to secure read-only connection | - Increase in reads to production DB |
| 2. Postgres' "logical replication" feature to keep production and staging media tables in sync | - Near-realtime updates automatically<br>- Staging DB has its own media records which can be deleted and modified | - Limited experience with feature<br> - Increase in load on production DB |
| 3. Foreign Data Wrapper to insert all records from production media tables into staging | - Staging DB has its own media records which can be deleted and modified | - Slower process<br>- Scheduled updates<br>- May duplicate data unnecessarily |

1. Use a foreign data wrapper to point staging media tables to their production counterparts. This means we read all media data directly from the production db without storing it in the staging db. It also means that staging always has access to the latest data in realtime. This also doesn't require us to write much code, at all. From a security perspective it's also quite simple to make sure this is a read-only connection that can't modify the production DB. The only thing to consider here is that we'd be increasing reads to the production DB but I think that's a minor concern. We can also dramatically downsize the staging DB with this approach.
2. Use Postgres' "logical replication" feature keep the production and staging media tables in sync. I don't have experience with this feature, but it allows for Postgres to automatically snyc changes from a table in one DB to another using a pub/sub model. Basically, in the production DB we `CREATE PUBLICATION pub_name FOR TABLE images;` and are able to subscribe to this in the staging DB. The only benefit here is that we get near-realtime updates automatically and the staging DB has its own media records which can be deleted and modified.
3. We could use a foreign data wrapper to insert all records from the production media tables into the staging media tables. We could do this incrementally, based on the created_at date of the newest records in the staging media tables, or we could do something more like the data refresh where we copy over all records. This would be slower, and scheduled, but the staging DB has its own media records which can be deleted and modified.

I'm curious what you think about these ideas! I only fell into this rabbit hole while reviewing the project proposal but I'm excited to look at this problem space in a way I hadn't before.

krysal (commented):
I'm getting to this now. I'll share some comments hoping not to disturb too much (sorry I didn't make it to the first round!). The current round shown is "Revision" but it looks like you already made all the requested changes, so should it be in the "Decision" round? If so then I'd probably approve on the next comment. The implementation plan looks quite detailed and straightforward to me, agree with all on it. Thanks for writing it!

---

Regarding @zackkrida suggestions. Those sound good for only the subset of goals that pertains to the project but keep in mind that staging is still for a general purpose. We need a separate copy of production to test things like deployments and migrations (the ones affecting the main tables are uncommon but still is much necessary to test them before getting to production), so I don't think we can/should go any of those approaches here or in the midterm.

zackkrida:
@krysal my suggestions should make the use cases you identified _easier_, because they don't require that the staging database be rewritten completely to keep the media in sync. As written, the plan here would overwrite any migrated tables weekly, wouldn't it?

krysal:
@zackkrida Reading again, I stuck with the idea of the first approach but the other two seem to be more viable. The [logical replication](https://www.postgresql.org/docs/current/logical-replication.html) especially looks precisely like what we need here:

> - Sending incremental changes in a single database or a subset of a database to subscribers as they occur.
...
> - Giving access to replicated data to different groups of users.
> - Sharing a subset of the database between multiple databases.

It's quite an interesting feature! Though the project doesn't need "near-realtime updates" but rather a data size of near production level. It'd be interesting to know more about the pros & cons of this approach. It sounds like it could be easier to implement (just some SQL commands? needs anything else?) but how much load increase will put on prod DB? It is acceptable? It is compatible with our migrations process? I'll be in favor of whatever is easier to implement and maintain with our current practices.

rfcs/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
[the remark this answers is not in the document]
stacimc: OH got it, that did totally confuse me -- I was thinking of `wait_for_completion` as the task id for a sensor. Got it!

AetherUnbound:
Thanks for your investigation and suggestions @zackkrida! I appreciate your effort to explore a few other options. I do think that recreating copies of production is still the best option for us, for a couple of reasons.

> If the staging database is just a copy of the production database, why do we even _need_ a staging database? I came up with the following reasons, trying to ignore anything that wasn't DB specific:
> 
> 1. To test Django admin views and operations which impact Django DB tables
> 2. To test new tables for new media types
> 
> These to me suggest that we'd want the Django-specific tables to be persistent, and not wiped out weekly by restoring from production backup. What we really want is to only sync the media tables. Additionally, I believe access to these media tables only needs to be read-only, as API consumers and the Django admin dashboard do not actually make changes to the media tables.

I would actually disagree with this last point - we want to sync _all_ the tables, because if we do not we're effectively maintaining two separate, distinct copies of the database. Here are a few scenarios where updating only the base media tables (e.g. `image` or `audio`) would cause more work on our part:

- A new provider is added, so we add a new entry into the "Content Providers" table in production. Since this table (`content_provider`) is not synced, we'd also need to add an entry in staging.
- A media item is marked as mature in production, and addressed in the Admin UI to hide it from searches. Since this table (`api_mature<media>`) is not synced, the mature/reported results would still show up on searches in staging.

Additionally, the sync approach might impede our ability to test migrations against the main media tables in staging. For instance, the recent [removal of `tags_list`](https://github.com/WordPress/openverse/pull/956) could fail to deploy in staging because migrations to the primary image table could be affected/prevented by the syncing mechanism described. Granted, if we are copying the staging database from production weekly then it's possible some migrations would effectively be undone when the update occurs. However, our [zero-downtime deployment policy](https://docs.openverse.org/guides/zero-downtime-database-management.html) paired with the fact that Django will re-run any necessary migrations on the next staging deployment should effectively mitigate any potential issues there.

All that to say, unless we want to duplicate maintenance efforts and potentially impede our ability to run migrations, I think having a full copy of production restored each week would be more desirable than syncing the media tables. I may be misinterpreting what you're saying, so let me know if I missed a point in your explanation that covers the above cases!

@krysal thanks also for taking a look, I'll apply your suggestions now :slightly_smiling_face: Once I've applied your feedback, I'll move this into the "Decision" round.

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,185 @@
...
+   should also check the `Status` value of the response for the most recent
+   snapshot to ensure that it is `available`.
+2. In the case where the snapshot is not yet available, wait for the status of
+   the most recent snapshot using the
+   [`RdsSnapshotExistenceSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsSnapshotExistenceSensor).
+3. In parallel, gather the attributes of the **staging** database using boto3's
+   [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html)
+   function (an Airflow operator does not exist for this operation). Namely, the
+   following attributes should be gathered and set to ensure they match:
+   - Availability should be a single DB instance (not multi-AZ)
```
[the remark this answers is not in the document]
AetherUnbound: I'll add a reference here!

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,185 @@
...
+   function (an Airflow operator does not exist for this operation). This step
+   should also check the `Status` value of the response for the most recent
+   snapshot to ensure that it is `available`.
+2. In the case where the snapshot is not yet available, wait for the status of
+   the most recent snapshot using the
+   [`RdsSnapshotExistenceSensor`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/sensors/rds/index.html#airflow.providers.amazon.aws.sensors.rds.RdsSnapshotExistenceSensor).
+3. In parallel, gather the attributes of the **staging** database using boto3's
+   [`describe_db_instances`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/describe_db_instances.html)
+   function (an Airflow operator does not exist for this operation). Namely, the
+   following attributes should be gathered and set to ensure they match:
```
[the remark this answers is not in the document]
AetherUnbound: The care I mention here _should_ only be necessary when setting up the DAG, especially if we're going to use the `describe_db_instances` call Sara described above to capture the existing settings. So I don't think we'll need any runtime checks here 🙂

AetherUnbound:

The deadline for review for this round is **April 20**

sarayourfriend (changes requested):
Jumping in here during the decision round to raise a blocker. Below I describe the blocking issue I see with it, in light of the discussion Zack and Krystle have been engaging in. In general, I suspect we are overcomplicating this issue, though I may be forgetting a critical detail.

> we want to sync all the tables, because if we do not we're effectively maintaining two separate, distinct copies of the database

If we sync the entire database, we're knowingly deleting reports and any other new data created in staging on an ongoing basis.

Additionally, to Krystle's point, staging is general purpose. If we restore the staging database from a production snapshot, that will also undo any migrations that have been applied to staging but not production. Those migrations would re-run the next time staging is deployed, but that's not guaranteed to happen at any specific time after the production data is imported. If the Django app expects a table to exist (or not exist), but production doesn't have that yet, then staging Django will crash until it is deployed and can re-apply the migrations that aren't yet on production.

To fix that, **this plan needs, at minimum, an additional step to automatically redeploy staging.** To avoid staging repeatedly crashing, we should just stop the staging service until the database is ready to have staging deployed against it again (and re-apply migrations).

On top of that, we need to also acknowledge that this will wipe out the following that are not shared between staging and production:

1. Any Django users added (critical for testing the development of #1176)
2. Reports and deleted/mature images added or removed (also critical for testing the above but relevant today as well)

We will also need to include any future tables in that list, whatever those might be. Data replication would solve that, but it presents the migration issue again. If we're okay with the data wipe outs listed above on a regular basis, then we still need to solve the issue of migrating the staging DB after restoration. I'm not sure that I am okay with monthly wipe outs of that data in staging. It would cause delays in critical feature development that we've agreed to this year, and ones that would be easy to forget are relevant.

My suggested and preferred alternative is to commit to running a data refresh on the same schedule as production. We already have a staging ingestion server anyway. We should use it on a regular basis. This would reduce the scope of this implementation plan to enabling the data refresh DAGs for staging. It prevents the migration issue, prevents data wipe out, and actually forces us to exercise ingestion on staging with staging changes on a regular basis. It is also significantly simpler, requires no new DAGs, and no new downtime in staging.

AetherUnbound:
> I'd like to request one small DevEx improvement: if we know the schedule that the database restoration would happen at, can we add a notice to the top of Django admin with the date of the next database reset in staging? That way folks would have somewhere easy to reference and likely right in front of their eyes just when it'll be most important to them.

This is a great idea! I can add that too, it should be fairly easy to implement since the schedule will just be `@monthly`.

> Can you explain more about this comment?
> 
> > I want to try and avoid a situation where we're actively maintaining two copies of the database
> 
> We _are_ maintaining two databases, even under this plan. Staging will still have migrations run against it that production won't have until later. I think I'm just missing what the significance is of the distinction of "maintaining two separate databases" or what the implications of doing or not doing so are.

Yes, what I mean here is that I don't want us to have to, indefinitely, carry all changes that we make to production _back_ to staging by hand. My examples [above](https://github.com/WordPress/openverse/pull/1154#issuecomment-1512262136) specifically around new providers and media reports fit into this consideration; I want to make sure we aren't having to add every new provider to both staging and production if we want them present in both, or to have to backport all of the mature images from production to staging. Even our recent example of [deleting thumbnails for certain providers](https://wordpress.slack.com/archives/C02012JB00N/p1681830916291229) would be a change that, bereft of any way to pull/update from production, would require us to also make the same change in staging. Having a way to regenerate staging from a copy of production regularly would prevent those cases in the long term, while still allowing us to make divergent changes to staging in the short term for things like feature testing.

sarayourfriend:
One more thought occurred to me, so sorry for repeated "ideas", but I want to make sure the bases are covered.

Based on Zack's suggestion to use the replicated tables, could we copy the data from replicated tables into tables that then get "promoted" to the main table, similar to how the data refresh works? My understanding is that the lengthy part of pushing data into the API is Postgres indexing it and the popularity calculation stuff. Would that get around that issue for staging?

zackkrida:
@AetherUnbound it occurs to me that I was maybe too specific in the example implementations I provided when attempting to share their usefulness. I maintain that they handle some of the use cases you are addressing as good as or preferably to the backup-restoration approach.

What I'm suggesting basically allows us to take different approaches per-table to syncing or pulling information from the production database. There are certainly performance considerations to factor in but as far as I can tell it would allow for exactly the same approach as the backup approach, if we wanted, but also allows for much more

I want to take some time to walk through your examples:

> A new provider is added, so we add a new entry into the "Content Providers" table in production. Since this table (content_provider) is not synced, we'd also need to add an entry in staging.

In your proposal, this would be synced to staging anywhere from 1-31 days after being added to production, depending on which day of the month it is added, correct? So hypothetically, if we enabled Phylopic on April 2nd it would be live in the staging database four weeks later on May 1st? Assuming a monthly sync as described in the proposal.

Using one of the approaches where the staging db is subscribed to, or has read access to, the production database, we could handle this in a number of alternative ways:

1. The staging `content_provider` table is a 1:1 live synced replica of the production table
2. Subscribe the staging `content_provider` table to the prod `content_provider` table, so it gets live updates but can have its own contents (need to deal with conflicts here)
3. Write a DAG or other scheduled task at our preferred interval (daily, hourly, minute) which copies the production `content_provider` rows to staging. Something like `INSERT INTO content_provider SELECT * FROM content_provider_remote ON CONFLICT (id) DO UPDATE SET`.

> A media item is marked as mature in production, and addressed in the Admin UI to hide it from searches. Since this table (api_mature<media>) is not synced, the mature/reported results would still show up on searches in staging.

Again, this sensitive media would be synced to staging anywhere from 1-31 days after being added to production, depending on which day of the month it is added and could be addressed in the same way as the first example.

To be clear, there might be many, many reasons this is a bad idea, but I don't think "we're effectively maintaining two separate, distinct copies of the database" is an accurate one.

zackkrida:
@sarayourfriend just seeing your message here:

> Based on Zack's suggestion to use the replicated tables, could we copy the data from replicated tables into tables that then get "promoted" to the main table, similar to how the data refresh works? My understanding is that the lengthy part of pushing data into the API is Postgres indexing it and the popularity calculation stuff. Would that get around that issue for staging?

There are probably a few ways we could do this. One would be to have different DB schemas with their own `image` tables and use [schema search paths](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH) to seamlessly switch to different `image` tables. With schema search paths you can essentially choose which table to query without modifying the query itself. We could have a schema, for example, where the `images` table is a foreign table which allows us to query the remote images table directly from the production database, without copying any data. Another schema could include a copied subset of the images, stored in staging, which could have migrations performed on it, records modified, whatever.

sarayourfriend:
And that only works if we follow zero-downtime deployments closely, otherwise the schema differences when staging has migrations applied to it would cause staging to incorrectly query the production database.

AetherUnbound:
Thanks everyone for your input! I need to take some time to look over this feedback and decide what to do next with the plan. I'll leave this in the Revision round for the time being.

AetherUnbound:
I want to say thanks again to everyone who's taken time to respond to this implementation plan. This discussion has challenged some of the assumptions I had when drafting the proposal and have brought clarity to both the proposal as it stands and potential alternatives.

For context, staging has been updated historically in one of two ways:
1. A staging data refresh
2. A manual recreation of the database via snapshot in a mechanism which mirrors that which has been codified in this proposal

The former approach is currently inviable due to issues we're having with the data refresh in production. This implementation plan describes the steps necessary to convert the manual process used for recreating the database from a snapshot.

I appreciate the alternate ideas that have been shared with how to approach regular updates to staging. I think in the long run, they are more thorough and would provide us more flexibility. However, in contrast to the plan as written, they _are not_ processes we've performed manually in the past and would require a degree of trailblazing in order to accomplish (specifically regarding the FDW/replication pieces). The current implementation plan represents a codification of a set of steps we know to work and which we have confidence in implementing, albeit with additional checks and caveats that the discussion has surfaced.

I have a few concerns in drafting a new implementation plan with the alternate approach:

- The FDW/replication approach(es) described may have numerous pitfalls, unknowns, or drawbacks which the maintainers are not privy to due to unfamiliarity with the functionality.
- The current plan describes a DAG which would likely take only a few days to implement as it stands.
- Tabling this plan and drafting a new implementation plan would restart this process, requiring an additional two weeks for discussion (in addition to implementation time thereafter).
- If I'm understanding the proposal correctly, what is being proposed as an alternative solution is essentially an ETL framework for production -> staging replication. While such a system could be significantly more flexible, it would by necessity also be more complex. This could likely be its own project entirely.
- This is the first of [three implementation plans this project requires](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html#required-implementation-plans), the other two being "Rapid iteration on ingestion server index configuration" and "Staging Elasticsearch reindex DAGs for both potential index types". Any elongations to the project timeline at this step will also affect the timeline for drafting and implementing those plans as well.
- The DAG described here does not present any **permanent, lasting, and irreversible** changes; it can be disabled at any time in favor of a more thorough update approach down the line.

With all this in mind, I'd like to take the following course of action:

- Create a new project to put on the backlog for fleshing out the production -> staging replication mechanism which @zackkrida has described.
- Add a "Discussed alternate approaches" section to the implementation plan which describes the proposed alternative approach as a guide for how this could be done differently in the future, and the reasons why the current approach was taken.
- Addressing the feedback provided, namely:
    - Ensuring the following steps occur at the end of the database update process:
        - The staging API is deployed
        - A reindex of Elasticsearch is initiated
        - The Oauth tables are truncated
    - Adding the option to skip this DAG by way of Airflow Variable

This will allow us to continue to advance the rest of the project while setting up plans for improvement of the workflow in the future.

If anyone has any **blocking objections** with the above proposed steps, please express them!

sarayourfriend:
> The DAG described here does not present any permanent, lasting, and irreversible changes; it can be disabled at any time in favor of a more thorough update approach down the line.

I agree with this, but I would also like folks who intend to give feedback to explicitly say whether they're okay with data created in staging being destroyed on a regular basis, even with the option to prevent it at any time being available.

> This is the first of [three implementation plans this project requires](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html#required-implementation-plans), the other two being "Rapid iteration on ingestion server index configuration" and "Staging Elasticsearch reindex DAGs for both potential index types". Any elongations to the project timeline at this step will also affect the timeline for drafting and implementing those plans as well.

This is a bit of an aside, but I don't see why this should be the case. The other two implementation plans do not at all appear to depend on this one. Am I missing a connection here?

> A reindex of Elasticsearch is initiated

We hadn't discussed this yet (or else I missed it), but it is an interesting point. Won't this require work to extract the relevant parts of the data refresh? On the surface this seems like something that would have a bit of complexity to it because of how singular our approach to indexing data is right now. Maybe we can use the `UPDATE_INDEX` action for this? From what I understand there's some uncertainty about that action, but it would be cool if we got to start. Mostly just want to point out that this seems like a complicated issue and don't want that glossed over. It certainly appears to challenge the idea that this IP would be a single PR, unless the PR in question was pretty significant.

The list of changes is also missing the callout in Django admin for the next staging data reset, just want to make sure that isn't missed :slightly_smiling_face: 

In any case, I'm aligned with your opinion that the approach as described with the modifications discussed is viable, if not in the long-term, at least in the short term. However, to reiterate, I am especially curious to hear Krystle and Zack's opinion about data generated in staging being wiped out on a regular ongoing basis. It's a non-trivial implication of this plan that is worth being explicit about, including calling it out as an inevitable result in the implementation plan.

stacimc:
> With all this in mind, I'd like to take the following course of action:

+1 @AetherUnbound for the path forward proposed in [this comment](https://github.com/WordPress/openverse/pull/1154#issuecomment-1515652916)! Catching up on the discussion, my main takeaways are that the approach currently described in the document (with a few revisions needed, as you've listed) would be:
  * fast/easy to implement
  * relatively risk free
  * presents no 'downgrade' or regressions in what we _currently_ have for staging DB updates
  * can be easily undone & swapped out for a more robust approach in the future

I'll also add that the alternative idea is _extremely_ cool and definitely seems like a worthwhile investment to me, as well as an idea that I'd hate to lose track of. Creating a new project for it sounds great to me!

AetherUnbound:
> The list of changes is also missing the callout in Django admin for the next staging data reset, just want to make sure that isn't missed 🙂 

Oh thanks for that callout - I've edited it into the list that I've posted so folks can see it when they come online! I'll respond to the other comments tomorrow.

AetherUnbound:
> > This is the first of [three implementation plans this project requires](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html#required-implementation-plans), the other two being "Rapid iteration on ingestion server index configuration" and "Staging Elasticsearch reindex DAGs for both potential index types". Any elongations to the project timeline at this step will also affect the timeline for drafting and implementing those plans as well.
> 
> This is a bit of an aside, but I don't see why this should be the case. The other two implementation plans do not at all appear to depend on this one. Am I missing a connection here?

This was implicit to that comment but I'll make it explicit here: I do not know that anyone else has the intention of writing these implementation plans, so my assumption was that _I_ would be writing these plans. Any time that I take on writing a new implementation plan for this step necessarily would take away from the time available for me two research & write those plan. In short: the shared connection is me, personally, and my availability :slightly_smiling_face:

> > A reindex of Elasticsearch is initiated
> 
> We hadn't discussed this yet (or else I missed it), but it is an interesting point. Won't this require work to extract the relevant parts of the data refresh? On the surface this seems like something that would have a bit of complexity to it because of how singular our approach to indexing data is right now. Maybe we can use the `UPDATE_INDEX` action for this? From what I understand there's some uncertainty about that action, but it would be cool if we got to start. Mostly just want to point out that this seems like a complicated issue and don't want that glossed over. It certainly appears to challenge the idea that this IP would be a single PR, unless the PR in question was pretty significant.

I initially brought this up [in this comment](https://github.com/WordPress/openverse/pull/1154#issuecomment-1513867338) with my note:

> This _does_ bring up a good point though that, regardless of our approach, we'll **need to reindex Elasticsearch with the new data from the staging database**.

Yes, I believe we should be able to leverage `UPDATE_INDEX` at this step and it would be an interesting and useful test to see if our assumptions about that action are correct! In the case that that action would require significant work to make it functional, we could do a full reindex. We will need to leverage the existing data refresh DAG code for this, but it may already be set up in such a way that we only need to invoke task factories. I'll be looking into this as I add the steps here to the existing plan. 

> The list of changes is also missing the callout in Django admin for the next staging data reset

Also making a note here that this is parallelizable work and I'll note as much in the plan revisions.

zackkrida:
I'm okay with the staging data being wiped on a monthly cadence, and reassured by the fact that we'll trigger an ES reindex and wipe the oauth tables after restoring the backup. 

More generally, and to reiterate for others some ideas I've shared with @AetherUnbound:

I'm happy to stick to the original implementation plan as written, not on the basis of which idea is "better" but on the following pragmatic points Madison raised:

> The FDW/replication approach(es) described may have numerous pitfalls, unknowns, or drawbacks which the maintainers are not privy to due to unfamiliarity with the functionality.

> Tabling this plan and drafting a new implementation plan would restart this process, requiring an additional two weeks for discussion (in addition to implementation time thereafter).

> The DAG described here does not present any permanent, lasting, and irreversible changes; it can be disabled at any time in favor of a more thorough update approach down the line.

I also shared the idea that alternative ideas _can_ be useful to validate our existing assumptions and reassure that our path forward is the right one. Taking more time to discuss alternative ideas is rarely time wasted, IMO. 

That said, I _also_ think it's okay in the future to say "that sounds cool, or even better, but for these practical reasons the original plan sufficiently meets our use case..." and save time on engaging with the alternative approach. 

I call this out specifically because I do think using the backup approach may well be _more appropriate_ in this situation without being the "better" solution objectively.

The very very last thing I want to share is how the prod -> staging foreign data wrapper could replace the backup restoration step in this plan _in the future_, potentially more quickly and simply, just so it isn't lost:

- Read the remote production tables from staging
- create corresponding temp tables in the staging DB
- copy all prod data into the temp tables
- swap the staging tables with the temp tables.

That could all be encapsulated in a single SQL function in the staging DB evoked by a DAG. It avoids the downtime and work to provision and wait for databases and snapshots.

AetherUnbound:
I believe I have addressed all the feedback from folks! I'm going to be moving this back into the **Decision Round**, but since I'm going to be gone most of this week and we have a few folks AFK, we'll extend the timeline for response - please respond to this by **May 3rd**.

stacimc (approved):
Looks good! The updates look to capture the feedback well, and thanks for creating #1874. Thanks for the very interesting discussion on this one.

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,305 @@
...
+    [`RdsDeleteDbInstanceOperator`](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/operators/rds/index.html#airflow.providers.amazon.aws.operators.rds.RdsDeleteDbInstanceOperator).
+    We should use the following configuration options:
+    - `wait_for_completion` should be left as `True` here so this stop will hang
+      until the database removal is complete.
+    - `SkipFinalSnapshot` should be set to `True` to avoid creating a final
+      snapshot of the database before deletion.
+12. Once the database rename is complete, the following actions will also need
+    to occur (they can occur simultaneously in the DAG):
+    1. The staging API must be deployed to apply any necessary migrations to the
+       new copy the database. In order to do this we must augment the existing
```
krysal, line 111: If the requirement is to only run migrations it would be easier and faster just to run the django command `python manage.py migrate` in sstaging instance. Are there additional reasons that warrant a deployment?

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
```diff
@@ -0,0 +1,305 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @stacimc
+- [ ] @krysal
```
krysal, line 11:
```suggestion
- [x] @krysal
```

krysal (approved):
The plan is perfectly reasonable and appropiate in my view! Thanks for writing it and add the suggestions @AetherUnbound. 

> I would also like folks who intend to give feedback to explicitly say whether they're okay with data created in staging being destroyed on a regular basis, even with the option to prevent it at any time being available.

To me staging was always a laboratory to try changes and never saw it as a place where data persisted, so I'm okay with it being wiped out on the workflow described here. Having data persisted on stagin would be a new requirement but I still don't see it necessary.

documentation/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md
sarayourfriend: To run a command in staging we'd need to enable [AWS SSM](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html) on the AWS client installed in Airflow. ECS Exec only works if SSM is enabled on the client. I don't know how much work it would be to set that up for the catalog. Staging API deployments take  <=3 minutes. My 2 cents is that it probably isn't worth the hassle for this particular use-case.

AetherUnbound: Yea agreed - I think a deployment would be the easiest thing to kick off given what we have.

AetherUnbound:
Coming back to this, I actually don't think that performing an `UPDATE_INDEX` on the staging primary index is the appropriate move at this time. This is mainly because that action assumes that all data in the database should be in the index, which is not how we're planning on setting staging up at the moment (see #1987):

https://github.com/WordPress/openverse/blob/ee9014441c44f32cf16673661122a3a4d15233d6/ingestion_server/ingestion_server/indexer.py#L325-L338

After #1987 is implemented we could add steps to initiate a full refresh of the indices that IP describes once the restore is complete. If we wanted to use `UPDATE_INDEX` here, we'd need to add some logic to that operation to run the same selections as are planned for #1987 and then compare the dates after that.

Issue #383, opened 2023-02-09
Sensitive content report moderation workflow


obulat:
| Start Date | Project Lead    | Actual Ship Date |
| ---------- | ----------      | ---------------  | 
| 2023-02-13 | @sarayourfriend | TBD              |

## Description

This project creates the sensitive content reporting flow guidelines.

## Documents

- [x] #1176
- [x] #1965
- [x] #1966
- [x] #1967
- [x] ~#1968~
- [x] #1969
- [x] #1970

## Milestones

- [x] Copy changes: https://github.com/WordPress/openverse/milestone/18
- [x] Implementation plan drafting: https://github.com/WordPress/openverse/milestone/8
- [x] Django admin access control and basic functionality changes: https://github.com/WordPress/openverse/milestone/22
- [x] Bulk moderation actions: https://github.com/WordPress/openverse/milestone/24
- [x] ~Metrification: https://github.com/WordPress/openverse/milestone/27~

krysal:
Required work to enable the DAG for reported media pending review created at WordPress/openverse-catalog#513. I noticed it hadn't run yet.

sarayourfriend:
@AetherUnbound has started on the first implementation plan for this project. We probably need to get the next implementation plan also started. The project proposal shows that the API cache management and the initial Django access control implementation plans can be started at any time. If any @WordPress/openverse-maintainers do not currently have an implementation plan they are working on or can take on a new one, please volunteer for one of the ones for this project.

I will start working on https://github.com/WordPress/openverse/issues/1969 next week.

sarayourfriend:
#2126 is pending review from the requested reviewers, @dhruvkb and @krysal. @zackkrida given the delay, do you think it's appropriate for my review of the implementation plan to serve as one of the two approvals for go ahead? @dhruvkb and @krysal, can y'all share an ETA of when you'll be able to review the implementation plan? It's been in the decision round much longer than the expected time span.

I was going to start working on the #1969 IP, but that hasn't happened due to recent incidents and high priority retrospective items taking up lots of time followed by MSR. I hope to be able to start that implementation plan next week.

The recent Redis memory consumption issue has brought to attention that the fine-grained cache IP will need to consider Redis memory usage.

sarayourfriend:
After starting to work on the fine-grained cache implementation plan, I opened a discussion about our approach to caching generally: https://github.com/WordPress/openverse/discussions/2453

[...]

PR #1176, opened 2023-04-12
Project Proposal: Sensitive content report moderation

sarayourfriend:
## Description

Part of https://github.com/WordPress/openverse/issues/383

Tagging @zackkrida and @obulat for review. I don't have strong reasons for choosing y'all but with the potential cross-team dependencies I felt like Zack was a good option. Everyone else is already involved in multiple other discussions, so I selected you Olga, because I hadn't pinged you for a discussion yet and the two you're already reviewing look like they'll be moving into decision soon. We won't start discussing this document until you have availability though, of course!.

[on the Openverse documentation site](https://docs.openverse.org/reference/decision_making/index.html).

## Current round

Blockers should be shared by 2023-05-01.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+
+- [ ] @zackkrida
+- [ ] @olgabulat
+
+## Project summary
+
+<!-- A brief one or two sentence summary of the project's features -->
+
+While Openverse supports reporting results as mature, we do not have a process
+of handling these reports. Likewise, there are significant gaps in our
```
zackkrida, line 15:
```suggestion
for handling these reports. Likewise, there are significant gaps in our
```

micro-nit

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+now and defer that as a potential future improvement. We will also update the
+"reason" for reports from `mature` to `sensitive_content`.
+
+### Volunteer moderator Django admin tools and access control (requirement 2)
+
+Django admin already includes granular access control tools. We'll need to scope
+out the creation of a new Django user group with appropriate permissions, the
+management of the users in that group, and how to get volunteer moderators
+through Cloudflare Access. One suggested implementation is to add a new
+`WordPress/openverse-content-moderators` GitHub group able access Django admin.
```
zackkrida, line 84:
```suggestion
`WordPress/openverse-content-moderators` GitHub group able to access Django admin.
```

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+A new GitHub group would have the additional benefit of allowing discussions
+about iterations to the moderation processes and tools to include everyone
+actually doing the moderation.
```
zackkrida, line 87: I think the GitHub group is an excellent approach 👍

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+    catalog ingestion to evaluate the "usefulness" of a description and develop
+    a score to help increase or reduce descriptions that may not be as useful.
+    This could be expanded to use bigger NLP APIs or text comparison/analysis
+    methods to find other qualities that could factor into the "usefulness" of a
+    description.
+    [I opened this discussion to further explore these ideas](https://github.com/WordPress/openverse/discussions/1175).
+
+Computer vision based enhancements to Django admin are covered in a subsequent
+section as it is sufficiently complicated to warrant its own set of
+considerations.
```
zackkrida, line 118:
```suggestion
Computer vision based enhancements to Django admin are covered in a [subsequent
section](#computer-vision-based-moderator-safety-tools-requirement-4) as it is sufficiently complicated to warrant its own set of
considerations.
```

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+user who performed the action, the date of the action, the type of the action
+(described below), and required notes describing why the action was taken. The
+possible actions are:
+
+- Deindex
+- Mark sensitive
+- Reindex
+- Undo mark sensitive
+
+The last two actions are necessary for "undoing" the first two.
```
zackkrida, line 151: 👍

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+caching middlewares to attach the cache key on the request context to make it
+accessible to downstream middleware that manages the reverse index.
+
+### Moderation response times (requirement 6, optional)
+
+If we are able to identify volunteers to moderate Openverse's content report
+queue, we can work with them to establish expectations about response times. To
+have insight into the process, we'll need a reporting mechanism that tracks the
+number of resolved reports, average response time, and number of open reports on
+an ongoing basis.
```
zackkrida, line 233: > we'll need a reporting mechanism that tracks the
number of resolved reports, average response time, and number of open reports on
an ongoing basis.

It isn't clear to me if this is optional or not, and if it is contingent on identifying volunteers or not? Would we want this even if Openverse maintainers were doing the moderation?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+    CV["Computer vision API integration for\nimage content safety metadata\n(Requirement 4)"]
+    Bulk["Bulk moderation actions\n(Requirement 3)"]
+    Ins["Moderation queue progress insights\n(Requirement 6)"]
+
+    AC --> VU --> CV
+    AC --> Bulk
+    CV --> Ins
+    Bulk --> Ins
+    end
+```
```
zackkrida, line 338: Excellent!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+tracking all the works affected by the action.
+
+For the bulk actions list view, the subclass should have a related join table
+tying the bulk action instance identifier to the result identifiers. For the
+creator actions, the subclass only needs to indicate the identifier for the
+creator. Note that disambiguated creator identification must include the
+provider as the same username across different providers may be used by
+different distinct creators.
+
+### Computer vision based moderator safety tools (requirement 4)
```
zackkrida, line 167: > ### Computer vision based moderator safety tools (requirement 4)

Could this section or the previous about amendments to Django be more explicit that we intend to render the images in the Django UI, and provide text hints? It would probably make sense to decouple the blurring from the computer vision section, actually, because we'd want to blur all reported images by default. We'll need to factor that blurring into the "Content report admin view upgrades" task.

This would help clarify the document, but I think it would also help me as a reviewer to make a determination about whether computer vision is necessary for this project or not. The hints provided by the computer vision API would be supplemental to the mandatory description provider by user reports, but it's already likely that moderators will need to view every reported image. I understand the hints could still improve safety and prepare moderators for what they may see.

Perhaps there would be ways to increase the usefulness of user-reports? For example, beyond a mere text description we could require them to choose content labels in the UI, for things like "violence", "sexual acts", and so on. 

I also wonder about it on the basis of storing the rekognition labels in a way they can be later applied to records in the catalog. That sounds like a potentially complicated and time-consuming dependency to add to this project.

zackkrida (changes requested):
This looks excellent, I've requested changes specifically concerning incomplete information on blurring and displaying images in the Django UI. All other feedback is non-blocking.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+<!-- A brief one or two sentence summary of the project's features -->
+
+While Openverse supports reporting results as mature, we do not have a process
+of handling these reports. Likewise, there are significant gaps in our
+infrastructure that make it difficult to respond to these reports or to lean on
+contributors outside the core maintainers for help processing them. This project
+will address these gaps, make it possible to easily and safely respond to
+content reports, and develop a baseline process for moderating content based on
+reports. It will also rename the "mature content" reports to "sensitive content"
+reports. This project is not the end-all, be-all for content report moderation
```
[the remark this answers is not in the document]
sarayourfriend: That's a great idea. I'd actually like to create a separate document for that, as it'd undoubtedly be referenced by other projects in the future as well. Maybe a "Terms" page at the root of the Trust & Safety directory next to the initial overview & user stories would be a good place to hold ideas like that. Or maybe more generally "guiding principles"? There are some aspects of the sensitive terms IP that we could add there too, like the "this will never be perfect" bit you wrote, which I think is helpful to keep in mind as we work through these things.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+might make sense in spite of those reasons listed above. However, Rekognition
+should be considered the "default" option unless another one provides a
+substantial benefit that invalidates those reasons.
+
+### Cache management (requirement 5)
+
+Our current caching strategy is to cache responses at the Cloudflare level for 1
+month based on query params. While this roughly works for our existing needs, it
+will not be workable for content moderation due to the inability to granularly
+invalidate cached pages via query params. Any Cloudflare cache features that
```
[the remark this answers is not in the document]
sarayourfriend: Correct. The only ways to get around that is through cache-tags which is an enterprise feature. I can add some clarification for why Cloudflare's existing caching doesn't work.

I should also clarify that we could keep Cloudflare caching on but reduce the time to <24 hours or something. That way we still get _some_ benefit without causing a massive delay in cache cleanup when we remove results. We should probably also have a way of purging the Cloudflare cache, in that case though, for when we find particularly problematic results that need to be hidden ASAP, though my grasp on when we would and wouldn't use that is tenuous.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
sarayourfriend: Yes, definitely worth clarifying. That was context I had in my head that I forgot to put down in the proposal, thanks for pointing it out.

> mandatory description provider by user reports

User descriptions on sensitive reports are optional, FWIW.

We could also treat the computer vision hints feature as non-blocking for starting to use the process, i.e., as the first improvement to the process that we'll roll out after the "MVP" is finished.

> I also wonder about it on the basis of storing the rekognition labels in a way they can be later applied to records in the catalog. That sounds like a potentially complicated and time-consuming dependency to add to this project.

My idea for this was to throw them in Redis and defer normalisation until the subsequent project, but I should be clear about that. We need to cache them _somehow_ to avoid making duplicate requests (and paying for them) and so long as the key in Redis is normalised the data would still be accessible for the downstream project.

sarayourfriend: > Perhaps there would be ways to increase the usefulness of user-reports? For example, beyond a mere text description we could require them to choose content labels in the UI, for things like "violence", "sexual acts", and so on.

This is a great idea, but would require more design and frontend development than I think was originally assumed for this project, so I'd want to clear it with @panchovm especially to make sure he has availability to help with something like that at this stage.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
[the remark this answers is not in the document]
zackkrida: A "guiding principles" page sounds like a great idea. I'll resolve this concern.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
zackkrida: > User descriptions on sensitive reports are optional, FWIW.

You're right, thanks for pointing this out. I was thinking of the "Other" reports mandating a 20-min-character description.

sarayourfriend:
Thanks for the early review, Zack. @obulat one of the other discussions you were involved in is closed now, so I'm going to move this into the clarification round.

Based on Zack's review, these are the revisions I will make thus far:

1. Clarify that images related to the content report will be blurred in Django admin and that the image classification metadata is used to give a hint to the moderator before they unblur.
2. Clarify that "storing computer vision API responses" really only amounts to ensuring that whatever method we use to cache responses has stable keys to be referenced in the future by the project to incorporate Rekognition data. The goal is not to decide the final format or storage method for that data, just to keep it in a way that a future project can use it without needing to duplicate the requests.
3. Clarify that the entire last requirement is optional, though ongoing insight into how often reports are resolved and the time it takes to resolve them would be helpful, it isn't a hard requirement of this project. Metrification of moderation activity can happen after the fact without hurting the goals of this project. Ideally we'd leverage something that allows for easy database visualisation like Grafana or [Metabase](https://www.metabase.com/).
4. Add more details to describe why Cloudflare caching cannot be managed programmatically for our use case and use softer language to describe _not_ using Cloudflare. The proposal should leave open the possibility that we would continue to use Cloudflare to cache requests for something like less than 24 or 12 hours (whatever makes sense).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+Requirements are outlined here, with further details below. Each of these
+roughly correspond to a single implementation plan. Requested plans are
+[covered in detail below](#required-implementation-plans).
+
+1. Language is updated to use "sensitive" or "sensitive content" rather than
+   "mature" where possible.
+1. Volunteers outside the core maintainers can access content reports in Django
+   admin and work through a queue of pending reports.
+1. Bulk actions can be taken to deindex or mark sensitive many results.
+1. Use a computer vision API to analyse images on a report to give hints to
```
kaueburiti, line 58: I would suggest to allow the user to control if they want to see the images blurred or not.
Making all images blurred by default and without the option to turn this off can make the moderation work slower.

sarayourfriend:
Thanks for those suggestions @kaueburiti, they're very helpful insights into what moderators expect. Being able to turn blurring off sounds like a great idea and easy to implement without adding much technical scope. I will modify the project proposal to specify that requirement. Would you consider the other two things (grid view and custom keyboard interactions) "must haves" for moderators to be able to do their work, keeping in mind that Openverse's report volume is currently relatively low (less than 2 per day)? Just trying to understand how critical those two tools are for the job so I can know whether to add them as requirements to this project or defer them to a fast-follow project.

kaueburiti:
> Would you consider the other two things (grid view and custom keyboard interactions) "must haves" for moderators to be able to do their work, keeping in mind that Openverse's report volume is currently relatively low (less than 2 per day)?

hmm, I didn't know that the volume was so slow hehe

Well, so I believe that those are not must have at all, but when the volume increases those are very important features in my experience 😅

sarayourfriend:
Okay, thank you for that clarification. Our team doesn't have a good way to demarcate things that are "fast follows" vs what must be completed for a given project to be considered "finished" and I do want to be defensive of scope on this project because it could essentially go on forever. Project there specifically being the Openverse "project" i.e., a unit of work, not the ongoing project of enabling and improving moderation, which as I said in the proposal, is a forever thing for us.

I'll ask the team what the best way to record that is without expanding the scope of this baseline project. Just want to clarify though that it'll get put down as a future requirement :slightly_smiling_face:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
```diff
@@ -0,0 +1,338 @@
...
+actually doing the moderation.
+
+The content report admin views already exist, but need some upgrades to improve
+moderator quality of life:
+
+- Moderator notes, for moderators to supply an explanatory note for their
+  decision. This should be a separate column to the content report description.
+- Improved table organisation by displaying the report ID and making that the
+  clickable column rather than the status (which is confusing).
+- Create a "decision" page that displays all pending and historical reports for
```
obulat, line 96: Would we want this info to be shareable with the content creators?  Do you think we would ever get requests from the authors about why their items were marked as sensitive?

obulat (approved):
The proposal is very thorough and clearly shows how much work needs to be done for this part of the Content safety goal.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
sarayourfriend: I think that falls under the disputes issue that I called out as something we should handle in a separate project. But yes, I think that would be valuable long term, just not part of this baseline project.

sarayourfriend:
I'm moving this proposal into revision. I plan to have the revised document ready by Wednesday, 2023-04-26.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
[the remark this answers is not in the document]
sarayourfriend: > Additionally, and more importantly, we'd need to bust the cache for any cached API response which returns the sensitive result, and we don't have a way of knowing that, right?

If we had custom cache keys, we could still do this. It isn't technically an issue with caching based on query params. We could maintain the same reverse index of result identifiers to cache keys/tags and invalidate those tags in Cloudflare. But that still relies on the Enterprise cache features.

In other words: if we were an Enterprise Cloudflare customer we would still be able to use Cloudflare's caching, everything we need to implement is possible. However, all the features we need are inaccessible to us, at least for the query param based views (search).

sarayourfriend:
I opened a new project thread for us to make further improvements to the moderator tools in the future: https://github.com/WordPress/openverse/issues/1885

sarayourfriend:
This proposal is now in the decision round. Please share any blockers by 2023-05-01 @zackkrida and @obulat.

sarayourfriend:
@zackkrida is AFK until Monday, so it's safe to assume that we might not have a decision on this until at least the day after the current deadline. That's fine, but if you could share your thoughts after the revisions @obulat, I would be grateful to have a head start on planning any additional revisions needed, if any exist.

obulat:
> @zackkrida is AFK until Monday, so it's safe to assume that we might not have a decision on this until at least the day after the current deadline. That's fine, but if you could share your thoughts after the revisions @obulat, I would be grateful to have a head start on planning any additional revisions needed, if any exist.

All the revisions look good, @sarayourfriend. Having input from someone more familiar with the content moderation process at scale also gives confidence. Thank you for your input, @kaueburiti ! 
I have no objections.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
zackkrida: @sarayourfriend just pinging you on this comment for visibility.

zackkrida (approved):
No blockers here. I have some open suggestions for typos and formatting, and one open question about the contents of the doc. LGTM!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.md
sarayourfriend: I added further clarification that this section is contingent on finding the group of moderators.

> Would we want this even if Openverse maintainers were doing the moderation?

Right now I am working off the strong assumption that Openverse maintainers will not be the ones doing the moderation. As was shared with us from professional moderation folks, having non-professional moderation is fraught and likely to get us in trouble. My impression was that the people we spoke to were on board with that, but I didn't get an answer when I explicitly asked if they would help, so I'll follow up and ask again.

However, if Openverse maintainers _do_ end up doing the moderation, then sure, we'll want these things. Particularly if our team is being asked to supply OKRs and KPIs, these are easy and obvious ones to share and track.

PR #1193, opened 2023-04-13
Ensure the generated changelog gets linted, fix backticks in log

AetherUnbound:
## Fixes

Fixes #1189 by @obulat, also fixes an issue preventing the linting from happening on changelog PRs like #1186

## Description

This PR modifies the `release-app` workflow in two ways:

1. It stages the generated changelog in git to _specifically ensure_ that `just lint` runs on it. For whatever reason, leaving it as an unstaged file prevents `just lint` from running on it, even though I cannot reproduce that behavior locally (i.e. unstaged & untracked files _are_ linted locally).
2. It changes the way the body of the changelog is inserted into a file so that backticks (`) don't get interpolated by bash. I've confirmed that using single quotes should prevent that locally, but I can't test this fully yet until we have a release that includes backticks in the file.

**Successful release run**: https://github.com/WordPress/openverse/actions/runs/4694770892/jobs/8323238328
**Already-linted changelog PR**: https://github.com/WordPress/openverse/pull/1192

## Testing Instructions

View the above!

## Checklist

- [x] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

[...]

PR #1815, opened 2023-04-18
Try to prevent additional issues with writing the changelogs to file

zackkrida:
I am _not_ a shell script expert but am curious if this might fix issues I'm still seeing with drafting certain release notes. See the [recent run here](https://github.com/WordPress/openverse/actions/runs/4735022950/jobs/8404730500) which errored due to a "syntax error near unexpected token `(".

### DIFF
--- a/automations/python/workflows/write_changelog.py
+++ b/automations/python/workflows/write_changelog.py
@@ -1,9 +1,9 @@
 import os
 
 
-app = os.environ["INPUTS_APP"]
-tag_date = os.environ["STEPS_TAG_OUTPUTS_DATE"]
-release_body = os.environ["STEPS_RELEASE_DRAFTER_OUTPUTS_BODY"]
+app = os.environ["APP"]
+tag_date = os.environ["DATE"]
+release_body = os.environ["RELEASE_BODY"]
 
 file_path = f"documentation/changelogs/{app}/{tag_date}.md"
 
--- a/automations/python/workflows/write_changelog.py
+++ b/automations/python/workflows/write_changelog.py
@@ -1,12 +1,16 @@
 import os
+from pathlib import Path
 
 
 app = os.environ["APP"]
 tag_date = os.environ["DATE"]
 release_body = os.environ["RELEASE_BODY"]
 
-file_path = f"documentation/changelogs/{app}/{tag_date}.md"
+file_path = Path(__file__).parents[3] / f"documentation/changelogs/{app}/{tag_date}.md"
+file_path.write_text(
+    """
+    # {tag_date}
 
-with open(file_path, "w") as f:
-    f.write(f"# {tag_date}\n\n")
-    f.write(f"""{release_body}\n""")
+    {release_body}
+    """
+)
--- a/automations/python/workflows/write_changelog.py
+++ b/automations/python/workflows/write_changelog.py
@@ -1,0 +1,12 @@
+import os
+
+
+app = os.environ["INPUTS_APP"]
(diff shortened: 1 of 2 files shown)


AetherUnbound:
It looks like that run failed because a `'` character within the changelog body closed the changelog string before the final `'` was supposed to. Unfortunately, I don't think this change will address the problem either because the double quotes in `"${{ steps.release-drafter.outputs.body }}"` puts us back in the same position we were in #1193 where backticks were being interpolated. I think we might need to use a github expression to replace or escape the `'` characters in the changelog as part of `${{ steps.release-drafter.outputs.body }}`.

dhruvkb (commented):
This is not directly helpful but I'd recommend using Python for anything more complicated than a simple `echo`. That way the expression can be interpolated into a triple quotes raw string, escaping everything.

```python
value = r"""${{ text }}"""
```

AetherUnbound (approved):
LGTM! 🤞🏼

zackkrida:
I'm going to try a frontend test run with this PR.

zackkrida:
@AetherUnbound @dhruvkb I made a test run of this but I think the path was incorrect:

https://github.com/WordPress/openverse/actions/runs/4736717050/jobs/8408615526

```
  File "/home/runner/work/openverse/openverse/automations/python/workflows/write_changelog.py", line 10, in <module>
    with open(file_path, "w") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'documentation/changelogs/frontend/2023.04.18.20.33.52.md'
Error: Process completed with exit code 1.
```

I think it might relate to the working directory in the workflow configuration being         `working-directory: automations/python/workflows` and the relative path in the script being `file_path = f"documentation/changelogs/{app}/{tag_date}.md"`.

Perhaps I shouldn't set the working directory and instead invoke ` python  automations/python/workflows/write_changelog.py` in the run command 🤔

AetherUnbound:
Oh shoot, that's totally it. Adding a suggestion for the script now regarding this.

automations/python/workflows/write_changelog.py
```diff
@@ -0,0 +1,12 @@
...
+
+app = os.environ["APP"]
+tag_date = os.environ["DATE"]
+release_body = os.environ["RELEASE_BODY"]
+
+file_path = f"documentation/changelogs/{app}/{tag_date}.md"
+
+with open(file_path, "w") as f:
+    f.write(f"# {tag_date}\n\n")
+    f.write(f"""{release_body}\n""")
```
AetherUnbound, line 12: Potential change to make this resilient to working directory changes

```suggestion
import os
from pathlib import Path


app = os.environ["APP"]
tag_date = os.environ["DATE"]
release_body = os.environ["RELEASE_BODY"]

file_path = Path(__file__).parents[3] / f"documentation/changelogs/{app}/{tag_date}.md"

with file_path.open("w") as f:
    f.write(f"# {tag_date}\n\n")
    f.write(f"""{release_body}\n""")
```

sarayourfriend: FWIW with Pathlib you don't even need to use `open`:

```py
file_path.write_text(
    """
    # {tag_date}

    {release_body}
    """
)
```

sarayourfriend (approved):
LGTM! Good call from @dhruvkb to use Python instead.

automations/python/workflows/write_changelog.py
AetherUnbound: Good point 🙂

PR #1833, opened 2023-04-19
Add filtered index creation DAG

### LINKED ISSUE #1197: Create `create_filtered_<media_type>_index` DAG factory
## Problem

We need a new DAG to be able to trigger the creation of the filtered indexes.

To ease local testing, this is blocked by #1196.

## Description

Please see the technical description from the implementation plan here: https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html#airflow-data-refresh-dag-factory-overview-step-2

Steps to complete:

1. Create a new DAG factory function to create a DAG that completes the steps described in the implementation plan section linked above (search for "It must do the following"). Remember to add the anti-race-condition checks by checking that the data refresh DAG for the media type is not currently active. If it is, the DAG should fail.
1. Trigger the new DAG from the data refresh DAG at the end, passing the `force` parameter to skip the anti-race-condition check. Add a new Airflow Variable to act as the feature flag for triggering the filtered index creation DAG only when it is enabled.

sarayourfriend:
## Fixes

Fixes #1197 by @sarayourfriend 

## Description

Adds new filtered index creation DAGs per media type:
- `create_filtered_image_index`
- `create_filtered_audio_index`

Because interaction with the ingestion server is now shared between two different DAGs (data refresh and this new one), I've abstracted those operators out into their own module with some quality of life improvements to make additional ingestion server interactions easier moving forward. I think there are some additional improvements here, specifically in that the create filtered image index DAG and data refresh DAG look _preeeetty_ similar right now. There may be more code sharing that could be done. However, in the interest of making this PR easier to review and in not abstracting too early, I've left it just at extracting the parts that are 1:1 shared between the two without further levels of abstraction required.

The line count looks big, but ~110 of the new lines are purely documentation/lengthy comments and another 126 are roughly copy/pasted abstractions of code that already existed in the data refresh DAG.

## Testing Instructions

This PR adds new DAGs and changes the data refresh DAGs, so we must test both. Both can easily be tested locally now that the docker-compose stack includes the ingestion server and the catalog.

I've had a hard time trying to understand how best to write unit tests for these DAGs. Based on the fact that the data refresh DAG doesn't have comprehensive unit tests, I'm assuming that it's just too difficult to do right now. I opened another issue to explore writing integration tests that would exercise these DAGs in an automated test setting: #1843.

Anyway, for manual testing, to set up your local environment:

1. `just ingestion_server/up && just catalog/up`
  - You could also use `env COMPOSE_PROFILES="ingestion_server,catalog" just up` if you're wanting to be really slick and don't mind the extra typing :stuck_out_tongue_winking_eye: 
2. If you haven't run API init, you should do that as well: `just api/up && just api/init`. This will give you existing indexes to work off of and test the DAGs against.
3. I would recommend having [Elasticvue](https://app.elasticvue.com/) to easily observe changes to your local indexes. Use the Elasticvue browser extension to avoid needing to fiddle with CORS settings in your local cluster.

You can now visit Airflow at `http://localhost:9090` and login with `airflow` for both username and password.

Filter DAGs by the `data_refresh` tag. Before running any DAGs, note the current state of your local Elasticsearch cluster indexes. You should have the following indexes (and potentially others you can ignore):

- `image-init` with `image` alias
- `audio-init` with `audio` alias
- `image-init-filtered` with `image-filtered` alias
- `audio-init-filtered` with `audio-filtered` alias

Now let's test the data refresh DAGs by manually running `image_data_refresh` and `audio_data_refresh`. Before running them, you need to add the `data_refresh` pool with slots set to `1` via the Airflow admin's Admin > Pools view. Now you can manually trigger them. These should pass without issue. In Elasticsearch, you should observe the following:

- `image-init` is gone and a new `image-{uuid}` index exists and has the `image` alias
- `audio-init` is gone and a new `audio-{uuid}` index exists and has the `audio` alias
- `image-init-filtered` is gone and a new `image-{uuid}-filtered` index exists and has the `image-filtered` alias
- `audio-init-filtered` is gone and a new `audio-{uuid}-filtered` index exists and has the `audio-filtered` alias

The `uuid` of the latter two should match the `uuid` of the former. This indicates that data refresh DAG correctly passed the generated UUID to the triggered filtered index creation DAG. You should also now check the `create_filtered_{media_type}_index` DAGs and confirm they have a successful run. Verify they've been correctly passed the DAG parameters by the data refresh DAG by checking the "Run config" in the graph view for the DAG set on the root DAG information (not individual tasks).

Now we can test running the filtered index creation DAGs on their own. You should manually trigger each of the `create_filtered_image_index` and `create_filtered_audio_index` DAGs without any custom config and with custom origin index and destination index suffixes.

Finally, we need to test that the concurrency check in the filtered index creation DAGs correctly fails the DAG when the corresponding data refresh is running. Rather than scrambling around the UI for that, you should use the airflow CLI:

```
$ just catalog/shell
# airflow dags trigger image_data_refresh && airflow dags trigger create_filtered_image_index`
```

You can try that for the audio DAGs as well, for good measure. When you run this, you should see the data refresh DAG run successfully without issue and the filtered index creation DAG should fail with a sensor timeout error.

**Reviewers: Is there a way to make the error message from the concurrency check task more descriptive? Right now it'll be a timeout error which is confusing if you're not aware that that is intentional.**

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/env.template
```diff
@@ -75,7 +75,7 @@ EMR_TEST_CONN_ID=emr_test
 
 # Connection to the Ingestion Server, used for managing data refreshes. Default is used to
 # connect to your locally running ingestion server.
-AIRFLOW_CONN_DATA_REFRESH=http://172.17.0.1:8001
+AIRFLOW_CONN_DATA_REFRESH=http://ingestion_server:8001
```
sarayourfriend, line 78: This change is now possible thanks to the monorepo effort and no longer requires different settings for different OS environments. We can just go straight through the docker network now.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -488,6 +488,21 @@ def create_and_populate_filtered_index(
...
+        filtered index should be based. If not supplied, the filtered index will be
+        based upon the index aliased to ``model_name`` at the time of execution.
+        :param destination_index_suffix: The suffix to use for the destination index.
+        This can be configured independently of the origin index suffix to allow for
+        multiple filtered index creations to occur against the same origin index, i.e.,
+        in the case where sensitive terms change before a data refresh creates a new
+        origin index and we wish to update the filtered index immediately. If not
+        supplied, a UUID based suffix will be generated. This does not affect the
+        final alias used.
+        """
```
sarayourfriend, line 505: I've added this documentation so that the Airflow param docs can link to the code to indicate how they are used upstream rather than needing to repeat that information.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -533,4 +548,6 @@ def create_and_populate_filtered_index(
 
         self.refresh(index_name=destination_index, change_settings=True)
 
+        if self.progress is not None:
+            self.progress.value = 100  # mark job as completed
```
sarayourfriend, line 552: This was missing from the original implementation and is used rather than the `ping_callback` approach by the catalog to check if a task has completed without error.

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -161,17 +111,36 @@ def create_data_refresh_task_group(
...
-            method="GET",
-            response_check=lambda response: response.status_code == 200,
-            response_filter=response_filter_stat,
+        # If filtered index creation was manually triggered before the data refresh
+        # started, we need to wait for it to finish or the data refresh could destroy
+        # the origin index. Realistically the data refresh is too slow to beat the
+        # filtered index creation process, even if it was triggered immediately after
+        # filtered index creation. However, it is safer to avoid the possibility
+        # of the race condition altogether.
+        # ``execution_dat_fn`` is used to find the most recent run becuase
```
sarayourfriend, line 122:
```suggestion
        # ``execution_date_fn`` is used to find the most recent run becuase
```

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -161,17 +111,36 @@ def create_data_refresh_task_group(
...
+            task_id="wait_for_create_and_populate_filtered_index",
+            external_dag_id=create_filtered_index_dag_id,
+            # Wait for the whole DAG, not just a part of it
+            external_task_id=None,
+            check_existence=False,
+            poke_interval=poke_interval,
+            execution_date_fn=lambda _: get_most_recent_dag_run(
+                create_filtered_index_dag_id
+            ),
+            mode="reschedule",
```
sarayourfriend, line 138: Please let me know if any of this needs further clarification.

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -44,23 +44,20 @@
 - [[Feature] Data refresh orchestration DAG](
```
sarayourfriend, line 44: Aside from the addition of the filtered index creation wait task, the changes in this module are almost all related to the abstractions necessary to share ingestion server communication tasks with the new DAGs.

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -0,0 +1,224 @@
+"""
+# Create filtered index DAG factory
+
+This module creates the filtered index creation DAGs for each media type
+using a factory function.
+
+Filtered index creation is handled by the ingestion server. The DAGs generated
+by the ``build_create_filtered_index_dag`` function in this module are
```
AetherUnbound, line 8: Note for this docstring, this gets rendered as markdown so I believe only one set of backticks should be necessary (same for the rest of this section):
```suggestion
by the `build_create_filtered_index_dag` function in this module are
```

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -0,0 +1,224 @@
...
+
+# Note: We can't use the TaskFlow `@dag` DAG factory decorator
+# here because there's no way to set the media type as a static
+# variable in the context of the DAG because all arguments
+# passed to the DAG factory decorated by `@dag` are thrown
+# away: `@dag` decorated functions' arguments are treated
+# merely as DAG param aliases. We could use `@dag` inside
+# this factory function, but it would require a redundant
+# call to the decorated function and doesn't look like it would
+# provide any additional value whatsoever.
```
AetherUnbound, line 75: Thanks for this note!

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -0,0 +1,224 @@
...
+        if origin_index_suffix:
+            create_payload["origin_index_suffix"] = origin_index_suffix
+        if destination_index_suffix:
+            create_payload["destination_index_suffix"] = destination_index_suffix
+
+        return ingestion_server.trigger_and_wait_for_task(
+            action="CREATE_AND_POPULATE_FILTERED_INDEX",
+            model=media_type,
+            data=create_payload or None,
+            timeout=timedelta(days=1),
```
AetherUnbound, line 119: I think this is something we might want to set up as a per-media-type configuration. We have the data refresh types defined elsewhere: 

https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/data_refresh/data_refresh_types.py#L59-L72

I notice that you're actually using those types as part of this DAG generator, perhaps we can add a new timeout config option there and leverage that value?

catalog/dags/data_refresh/dag_factory.py
```diff
@@ -261,11 +263,31 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
...
         )
 
+        index_suffix = XCOM_PULL_TEMPLATE.format(
+            data_refresh_group.get_child_by_label("generate_index_suffix").task_id,
+            "return_value",
+        )
+        trigger_filtered_index_creation = TriggerDagRunOperator(
+            task_id=f"trigger_create_filtered_{data_refresh.media_type}_index",
+            trigger_dag_id=f"create_filtered_{data_refresh.media_type}_index",
+            params={
```
AetherUnbound, line 273: This needs to be [`conf`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/trigger_dagrun/index.html#airflow.operators.trigger_dagrun.TriggerDagRunOperator), otherwise `params` usage here will be interpreted as [task-level params](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html#task-level-params). I tried running the DAGs locally and the triggered create filtered index DAG failed because this did not appropriately pass `force` and the downstream DAG started before the upstream DAG was complete.

```suggestion
            conf={
```

AetherUnbound (changes requested):
This is awesome! I appreciate some of the abstraction work you've done, I think it will be helpful towards #1154 as well. I'm also glad that it's so easy to set up the ingestion server stack along with the catalog now 💃🏼 

I have a few blocking issues, namely:
- We'll want to allow the timeouts to be configured per-media-type
- The DAG trigger needs to use `conf` instead of `poke`

Per this note:

> Reviewers: Is there a way to make the error message from the concurrency check task more descriptive? Right now it'll be a timeout error which is confusing if you're not aware that that is intentional.

Since we're already wrapping the `check_for_running_data_refresh` sensor in a python function, we could add a `try`/`catch` around `wait_for_filtered_index_creation.execute(context)` which would catch the `AirflowSensorTimeout` and raise a more useful exception!

catalog/dags/data_refresh/dag_factory.py
sarayourfriend: I don't understand :thinking: When I try this locally it does work, the data refresh is able to trigger the filtered index creation, i.e., the params are passed to the DAG.

Can you share more specifics about how you were able to reproduce the issue? The documentation you linked was what I referenced to use the params there, so I'm confused what to understand about it that should indicate the usage of conf instead :sweat:

catalog/dags/data_refresh/create_filtered_index_dag.py
sarayourfriend: Gotcha. It's hard to remember what docstrings are regular markdown and which are RST. Two backticks get rendered fine in Markdown (it's part of the CommonMark spec, even), so for consistency with all other Python docstrings I think I prefer to just use two backticks, rather than needing to think about where I do and don't need to use them inside a Python module.

catalog/dags/data_refresh/create_filtered_index_dag.py
sarayourfriend: Done!

sarayourfriend:
Thanks for the review, @AetherUnbound. Changes are up except for the conf/params change as I don't understand the issue yet. I asked for clarification in the comment thread.

catalog/dags/data_refresh/dag_factory.py
AetherUnbound: Just because the DAG itself gets triggered doesn't necessarily mean that the configuration parameters defined here are sent to the DAG. In my case, the DAG was started but did not receive the configuration parameters. I'm also realizing now that the `params=` you define for the DAG is also incorrect, since those are [DAG-level params](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html#dag-level-params).

To rephrase the two different (but confusingly named Airflow concepts here):
- [**Parameters**](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html) are a way to add variables _into the jinja templating that Airflow tasks use to render their inputs_. These are accessed via their names in templates, e.g. a Param named `x` is going to be accessible via `{{ params.x }}`. 
- [**DAG run configuration**](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#passing-parameters-when-triggering-dags) are the inputs that you can provide _to a specific DAG run_. These are accessible via `dag_run.conf`, e.g. `{{ dag_run.conf["x"] }}`. 

The [template reference](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html#variables) is useful here, although it _looks_ like we might also be able to use `params` for DAG run configuration if we set the `core.dag_run_conf_overrides_params` system configuration 😮

We use `dag_run.conf` as part of the ingestion DAGs: https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/dags/providers/factory_utils.py#L32

This does make me wonder, I will need to try to see if enabling that setting makes this behave as you're expecting, the docs make it seem as simple as that: https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#dag-run-conf-overrides-params. If we can get that to work I think I'd prefer it, especially since `Params` have a lot more configuration options!

AetherUnbound: My understanding of the above evolved as I was writing that comment ☝🏼 I need to head out for the day but the TL;DR is that I think everything should be right based on how you've defined it but maybe I have `AIRFLOW__CORE__DAG_RUN_CONF_OVERRIDES_PARAMS` not set correctly locally or something (or we may need to define it explicitly, even though it's supposed to default to `True`? 🤔)

AetherUnbound: Oh one other thing, the index UUIDs matched up when you ran this locally? For me they did not, another indication that the DAG run config/params weren't working:
Screenshot_2023-04-20_16-31-13

catalog/dags/data_refresh/create_filtered_index_dag.py
AetherUnbound: You're right it does! I didn't realize that and I thought it'd cause problems, feel free to disregard (and TIL)

catalog/dags/data_refresh/dag_factory.py
sarayourfriend: I don't think the DAG configuration is incorrect. Your description of the usage of it precisely matches what I intended: they're DAG level configuration variables that are passed by the DAG to the individual tasks. When you trigger the DAG "w/ config" from the UI and pass these variables they get correctly propogated through the tasks. If you pass `destination_index_suffix` as `"some-other-suffix"` it does work, at least when I run it locally.

I think you are correct about them not being passed. They do need to be passed with `conf`, but the explanation you gave didn't make sense to me. The `params` I've passed to the trigger task are passed to the task, `conf` is passed by the trigger task on to the DAG that it triggers. I didn't read the trigger task's documentation carefully. So you're right there.

Also: I'm not sure how but I must have done something terribly wrong when testing the data refresh DAG locally as I swear I was able to get it to work but now it does not, as you described.

I'll update this to use conf, and verify, but I don't believe there are any other changes needed.

sarayourfriend: Also: I realised we probably want a sensor to wait for the filtered index creation to finish, potentially? :thinking: What do you think?

sarayourfriend: Fixed the param passing in https://github.com/WordPress/openverse/pull/1833/commits/541648466e405bde9c563132ecde8bbc80b2c623

I'll update the testing instructions to share how you can prove it worked.

AetherUnbound: Ah, okay I see what you mean. I think the `params` construct has evolved since I used it last, it used to be more limited in scope. Apologies for my misinterpretation there!

As for a sensor, I think it would be necessary if there are any cases where other operations would need to wait for the filtered index creation to be complete. But isn't that what the `trigger_and_wait_for_task` function does?

AetherUnbound (approved):
Using `conf` makes this work perfectly! Fantastic 🙌🏼 

image

I was also able to encounter the new `ValueError` with the provided run command.

catalog/dags/data_refresh/dag_factory.py
sarayourfriend: Sorry, just to clarify, the sensor I was wondering about would be added to the Data Refresh DAG to wait for the filtered index creation DAG to finish. We'd only need it if we wanted the Data Refresh DAG's full timeline to include the length of the filtered index creation DAG that it triggers. Right now it triggers the DAG and then finishes immediately. I don't know what is more useful from an observability perspective, is all.

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -0,0 +1,230 @@
...
+                    "unless you're absolutely sure of what you're "
+                    "doing. This check exists to prevent race "
+                    "conditions and should not be ignored lightly."
+                ),
+            ),
+            "origin_index_suffix": Param(
+                default=None,
+                type=["string", "null"],
+                description=(
+                    "See https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L495-L497. "  # noqa: E501
```
dhruvkb, line 170: Weirdly, referencing a permalink seems odd considering that it might not reflect the present situation of the ingestion server, but at the same time, referencing `main` would be worse as it might start pointing to something else. Is there a way to not reference the ingestion server code and explain in a comment?

dhruvkb (approved):
LGTM, except for a small nit.

catalog/dags/data_refresh/create_filtered_index_dag.py
sarayourfriend: There are two other options, in my opinion:

1. Repeat the documentation here
  - Pros: Does not reference a static part of the ingestion server code
  - Cons: It will get out of date probably almost as soon as it is committed, as these things go
2. Rather than link to the code, just say "see `Indexer::create_and_populate_filtered_index` documentation in ingestion server"
  - Pros: Never goes out of date unless the method name changes (which would force changes in this DAG as well, because the method name is concretely tied to the task name that the DAG must know
  - Cons: Less convenient

I'd rather do the second if these permalinks aren't to satisfaction.

sarayourfriend: Though, I will say, ideally we'd have task documentation either statically written in the ingestion server documentation or pulled from docstrings in the code.

sarayourfriend:
The DAG docs generation script appears to have flakiness due to linting differences. I don't know why this would be the case. It's quite frustrating for it to pass locally but fail in CI. I don't have any environment changes locally and rebuilding the catalog containers is a no-op for me at the moment so it isn't even a difference in the dependencies or something.

I've pushed a change to avoid unnecessary linting in the documentation generation script, so at least it'll run faster and not pin the CPU to 100% when unnecessarily checking types.

catalog/dags/data_refresh/dag_factory.py
AetherUnbound: Ohhh, I see, thanks for clarifying! I think having that final sensor would make sense, since the data refresh is technically "still running" when the filtered index creation is happening, right?

PR #1858, opened 2023-04-20
Add Cloudflare malicious traffic runbook

zackkrida:
Adds a runbook for identifying and triaging malicious traffic to the Openverse frontend and API. I would like to supplement this, in the future, with an additional general guide for working with cloudflare's UI, but I think this guide is a good start that covers the primary and most critical usecase for Openverse maintainers. 

Also deletes an unused file full of functional programming utilities in JS that's been sitting in the repo for some time 😆


documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+
+1. There is a service disruption to the Openverse API or frontend.
+2. There is a marked increase in resource usage on the API or frontend ECS
+   services that continues to trend to unsustainable levels, which _could lead_
+   to a service disruption.
+
+And:
+
+1. There is a marked increase in traffic visible in our AWS dashboards or
+   Cloudflare analytics.
```
sarayourfriend, line 21: Would it be okay to put the necessary condition first? It's confusing to read in this order:

```suggestion
Use this runbook when there is a marked increase in traffic visible in our AWS dashboards or
   Cloudflare analytics and either of these are true:

1. There is a service disruption to the Openverse API or frontend.
2. There is a marked increase in resource usage on the API or frontend ECS
   services that continues to trend to unsustainable levels, which _could lead_
   to a service disruption.
```

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+can block them through the Cloudflare Web Access firewall.
+
+There are three primary identifiers for targeting traffic. These are ordered
+from largest, most persistent, and most impactful, to smallest and most
+ephemeral. When possible, **blocking ASNs** is the single best way to mitigate
+malicious traffic.
+
+#### Traffic Source Identifiers
+
+- Source ASNs (Autonomous System Numbers): Unique identifiers for a group of IP
```
sarayourfriend, line 59: nit: usually it's clearer to write the abbreviation after the expanded version.

```suggestion
- Source "Autonomous System Numbers" (ASNs): Unique identifiers for a group of IP
```

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+
+### 2. Identify targetable sources of the suspicious traffic
+
+Next, we need to associate the traffic with specific sources. Essentially, we're
+trying to find unique identifiers linked to potentially malicious traffic, so we
+can block them through the Cloudflare Web Access firewall.
+
+There are three primary identifiers for targeting traffic. These are ordered
+from largest, most persistent, and most impactful, to smallest and most
+ephemeral. When possible, **blocking ASNs** is the single best way to mitigate
```
sarayourfriend, line 54: This is the first mention of ASN, so I'd suggest expanding it here first:

```suggestion
ephemeral. When possible, **blocking Autonomous System Numbers (ASNs)** is the single best way to mitigate
```

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+#### Traffic Source Identifiers
+
+- Source ASNs (Autonomous System Numbers): Unique identifiers for a group of IP
+  networks and routers under the control of a single organization, typically an
+  internet service provider or hosting company.
+
+  Typically, a single ASN makes up less than 10% of total Openverse requests in
+  a given timeframe. If this percentage is larger, it could suggest that a user
+  on the ASN is scraping Openverse or accessing Openverse programmatically in an
+  aggressive way.
```
sarayourfriend, line 66: Slight rewording for clarity. The re-wording helps clarify for me what the difference between scraping and other programmatic usages are if they're both malicious.

```suggestion
  A single ASN usually makes up less than 10% of total Openverse requests in
  a given timeframe. If the percentage of requests from a single ASN is higher
  than 10%, it may indicate that one or more users on the ASN are scraping
  Openverse or over using/abusing expected programmatic access patterns.
```

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+
+- Source user agents: Unique user-agent strings sent with requests to Openverse.
+  Nearly all UA strings from web browsers begin with `Mozilla`. Outside of the
+  latest stable version of Chrome on Windows 10, which is the most popular
+  Openverse browser (and usually looks something like
+  `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/{webkit-version} (KHTML, like Gecko) Chrome/{real-chrome-version} Safari/{webkit-version}`),
+  no single user agent should take up more than 10% of total traffic. Excluding
```
sarayourfriend, line 73: > no single user agent should take up more than 10% of total traffic

Won't this become less true over time as UA masking becomes more common? Where does this 10% number come from? Same question for the ASN 10% above as well. When would we know to adjust the expectation of "10%" specifically? Is it possible for us to create a dashboard that shows a rolling distribution of these so that we know what to benchmark against at any given moment?

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+  user agents from things like social media platforms and popular http request
+  software libraries, like `Go-http-client/2.0` or `axios/0.21.4`, as examples.
+
+- IP addresses: A unique identifier for a single machine or user.
+
+  Typically, a single IP address makes up less than 1% of total Openverse
+  requests in a given timeframe. These larger IPs typically belong to "good" web
+  crawlers from places like Google, Bing, and Dotbot, as examples. If this
+  percentage is larger, it could suggest that a user on the ASN is scraping
+  Openverse or accessing Openverse programmatically in an aggressive way.
```
sarayourfriend, line 85: Can you clarify whether the "larger IPs" go over 1%? We wouldn't want to unintentionally block search engine crawlers.

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Steps
+
+1.  Using the view from the previous task, scroll down and look for IP
```
sarayourfriend, line 89: Helpful cross-reference. The "previous task" is ambiguous and the section immediately preceding this one is non-trivial. It would be easy to get lost (I did!)

```suggestion
1.  Using [the view from the previous task](#1-identify-the-timeframe-of-the-traffic-anomalies), scroll down and look for IP
```

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,161 @@
...
+    Addresses, Source ASNs, or Source User Agents which exceed the typical usage
+    thresholds defined above. Create a list of these traffic sources.
+
+```{note}
+Cloudflare will only show the 5 largest of each category; greater than 5 disproportionately-large items of any category would be unprecedented and surprising in Openverse's history.
+```
+
+```{warning}
+If you haven't found anything, it's unlikely that this situation is connected to malicious traffic or specific actors. Instead, it looks like a normal increase in traffic to Openverse. At this point, it's more likely that a change in code or infrastructure has caused a performance drop, exposed by the higher traffic, or that the traffic has grown so much that we need to allocate additional server resources.
+```
```
sarayourfriend, line 99:
```suggestion
If you haven't found anything, it's unlikely that this situation is connected to malicious traffic or specific actors. Instead, it is more likely to be a normal increase in traffic to Openverse. At this point, it's more likely that a change in code or infrastructure has caused a performance drop, exposed by the higher traffic, or that the traffic has grown so much that we need to allocate additional server resources. Consider recent Openverse marketing efforts or WordPress events if the traffic looks organic and we need to attribute it to something.
```
```

File: documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md [hunk 9]
```diff
@@ -0,0 +1,161 @@
...
+```{warning}
+If an ASN has a significant amount of human traffic, it can indicate this ASN is an ISP rather than a hosting company or company which offers a Virtual Private Network. Blocking these types of ASNs can disrupt service to human users.
+```
+
+1. Finally, and optionally, make a web search in your preferred browser for each
+   individual traffic source ID. Many ASNs and IP addresses used for scraping
+   will have negative reports, be on spam lists, or be associated with hosted
+   VPN services. This can be helpful in identifying which are truly malicious.
+   However, our first priority is Openverse's availability to users, so this
+   step may only be necessary in lower-pressure scenarios.
```
sarayourfriend, line 119: Should we _always_ do this, but with the caveat that it can be done after the incident is resolved? It may help prevent long-term accidental or unintentional block against someone we do not want to block.

sarayourfriend (approved):
LGTM! I have several comments and nits for clarity, but none are blockers. This is certainly something we can iterate on and solidify processes for over time. Thank you for sharing your description of this process!

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
zackkrida: So much better, thanks!

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
zackkrida: > but with the caveat that it can be done after the incident is resolved? 

Great idea. Yeah, the reason to make it optional was that it can be time consuming, but doing it after is a great workaround. I'll update!

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
zackkrida: Creating a dashboard could be a good idea. For now I've added a note recommending users look at a time slice of traffic _without_ any known malicious traffic or incidents to compare the percentages.

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,163 @@
...
+    Addresses, ASNs, and User Agents below a graph of known bot traffic. It's
+    very likely you'll see items from your lists in these graphs.
+
+3.  Use the "add filter" button and filter by your listed resources
+    individually. Write down which of the identified resources are comprised of
+    the most `Automated` and `Likely Automated` traffic.
+
+```{warning}
+If an ASN has a significant amount of human traffic, it can indicate this ASN is an ISP rather than a hosting company or company which offers a Virtual Private Network. Blocking these types of ASNs can disrupt service to human users.
+````
```
AetherUnbound, line 114: This looks like it's not quite opened/closed properly, which is making the render look odd. Specifically (suggestions are hard WRT escaping the triple backticks) the first warning here is 4 backticks and doesn't appear to be closed at all.
image

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,163 @@
...
+  to the frontend, but you still feel it should be blocked, consider using the
+  "Managed Challenge" action instead of an outright block. This should allow
+  human users but block automated traffic.
+
+### 4. Analyze reputation of identified traffic sources
+
+Finally, make a web search in your preferred browser for each individual traffic
+source identifier. Many ASNs and IP addresses used for scraping will have
+negative reports, be on spam lists, or be associated with hosted VPN services.
+This can be helpful in identifying which are truly malicious.
```
AetherUnbound, line 141: Should this perhaps be step 3 instead? So that it comes before the instructions for blocking a traffic source.

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
```diff
@@ -0,0 +1,163 @@
...
+## Escalation Procedures
+
+1. If the team member is unable to complete the procedure successfully, escalate
+   the issue to the rest of the team via a ping in our team chat. A maintainer
+   should review your run of the book and repeat it if they find a particular
+   type of traffic anomaly wasn't assessed.
+2. If the team member is unable to complete the procedure successfully, escalate
+   the issue to the rest of the team via a ping in our team chat. Here we will
+   explore more drastic measures for mitigating the service disruption, like API
+   throttling or rate limiting adjustments.
```
AetherUnbound, line 163: These seem duplicated and/or irrelevant - do we just need the first point?

AetherUnbound (commented):
Thanks for drafting this! I love that it can just go up on our docs site with everything else 🎉 I have a few comments. I think adding pictures for one or several of the steps could be useful as well, just to give an impression of the UI (although they aren't necessary).

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
zackkrida: It actually was previously, but I moved it because it can be time consuming and I don't want it to block mitigating the issue. It's cool that this stood out to you though; I will add information explaining its positioning!

AetherUnbound: Ah okay, sounds good!

documentation/meta/traffic/runbooks/identifying-and-blocking-traffic-anomalies.md
sarayourfriend: Rewording it might be fine for now but it is still a bandaid until we have ongoing observability of what the expected rate should be. It probably won't be an issue for a little while still, UA masking isn't that aggressive yet, which is why it isn't a blocker to me.

krysal (approved):
Thank you for writing this guide. It clarified for me the process for dealing with possible outages!

I fixed a little bit the formatting of the note blocks because doing it by suggestion in the GitHub interface was a bit of a mess. Also added an image for someone in the future wich might be unfamiliar with the Cloudflare interface. Hope it helps.

Issue #1240, opened 2023-04-17
Merge API documentation into main docs site


sarayourfriend:
## Problem

This is based on an issue that @obulat reported:

> Add Openverse colors/customize the design slightly to make it clear that it's still Openverse.

## Description

The page in question is https://api.openverse.engineering/v1/

Redoc, the library used for the Django documentation, has essentially zero customisation options. Additionally, it uses React, which we could SSR but would require us invest way more time into this.

Rather than having to maintain two separate documentation websites built on different and hardly related technologies, we could use something like https://github.com/sphinx-contrib/openapi to add our OpenAPI spec into the main documentation site. That would solve the issue Olga mentioned and reduce the number of disparate dependencies we rely on for documentation generation.

The `generate_swagger` management command can already be used to create an `openapi.yml` file, as seen in the check to validate generated specs: https://github.com/WordPress/openverse/blob/697f62f01a32cb7fcf2f4a7627650a113cba40da/api/catalog/management/commands/validateopenapischema.py#L37

CI would need to generate that spec and output it into the documentation site before building the final site. We would also need to make it easy to iterate on the documentation locally.

dhruvkb:
I agree with the idea of consolidating as much documentation as possible into one location but for some reasons, I still prefer the API documentation to be separate. 

Fundamentally, I consider our documentation to be of the following types based on who it is targeted towards.

- Developers building Openverse itself: https://docs.openverse.org, the Sphinx site
- Developers using the Openverse API: https://api.openverse.org, the ReDoc site
- Users of the Openverse search engine: https://openverse.org/search-help, the frontend page

While having them all in central location is good, having them where people will look for them during their usage is better.

sarayourfriend:
I can see your point for search users. For people consuming the API directly, I don't see a significant difference in usability/discoverability between hosting the documentation directly on the API or using the docs site, provided we linked to the docs site's API documentation from the API itself (or redirected to it). It does, however, present a maintenance burden. The different technology used to render documentation is redundant, but also prevents us from being able to create a sense of cohesion in our documentation. The API docs and the docs site feel disjointed at the moment due to having similar but ultimately very different presentation styles.

[...]

PR #1876, opened 2023-04-23
Install drf-spectacular

### LINKED ISSUE #1149: Use `drf-spectacular` to generate API docs in OpenAPI 3 format
## Problem

Currently we use `drf-yasg` to create API schema. It only supports OpenAPI v.2, not the current OpenAPI v.3. There are better instruments available to generate Typescript types and client code for the frontend for OpenAPI v.3.

## Description

There is great documentation for the migration from `drf-yasg` on the `drf-spectacular` website: https://drf-spectacular.readthedocs.io/en/latest/drf_yasg.html

## Alternatives

Keep using `drf-yasg`. 

## Additional context

This migration can help us generate better documentation of the media properties.

A link to the special setup that we use to add the code examples in the documentation: https://jangiacomelli.com/redoc-code-examples-in-drf-using-drf-yasg/

### LINKED ISSUE #1875: Schema for audio_related and image_related is incorrect
## Description

The schemas for the [`audio_related`](https://api.openverse.engineering/v1/#operation/audio_related) and [`image_related`](https://api.openverse.engineering/v1/#operation/image_related) endpoints are incorrect because they describe one instance whereas the response from those endpoints is a list of instances.

## Reproduction

1. Visit the linked schemas above.
4. See error.

## Additional context 

Will be fixed alongside #1149.

dhruvkb:
## Fixes

Fixes #1149 by @obulat 
Fixes #1875 by @dhruvkb 
<s>Address #1240 by @sarayourfriend</s>

## Description

This PR
- removes `drf-yasg` and its configuration
- installs and configures `drf-spectacular`
- adds examples for endpoints that lacked them (e.g. audio waveform)
- updates the OpenAPI documentation code to use `drf-spectacular`'s utils
- aligns documentation for authentication endpoints with that for media
- fixes some inaccuracies in the documentation
- seamlessly merges the authentication text with the authentication endpoints

## Testing Instructions

0. Check out the PR and run the services: `just api/up`
1. See the Redoc UI at `http://localhost:50280/v1/`
2. See the Swagger UI at `http://localhost:50280/v1/swagger/`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


.github/actions/build-docs/action.yml
```diff
@@ -44,3 +57,4 @@ runs:
         mv documentation/_build /tmp/docs
         mv frontend/storybook-static /tmp/docs/storybook
         mv frontend/.tcv-export /tmp/docs/tailwind
+        mv api/api_docs /tmp/docs/api_docs
```
sarayourfriend, line 60: This would run every time `main` is pushed, right? That would mean that the API docs would be updated to reflect _staging_ and be slightly ahead of production. Should we have a separate "staging_docs" area for pushes to `main` and then have the production docs built and published during Release app? :thinking: Originally I thought we could just copy the staging docs, but the production docs should reflect the code running in production which means it needs to handle rollbacks as well.

That's a big downside to including the API docs in the documentation site that I didn't consider before. That alone makes me doubt that this was a good suggestion (sorry!!!). Unless it's not too hard to juggle the documentation updates... What would we have to do? Each time the deployment workflow is dispatched, we'd need to pull the API image for the version being deployed, build the docs, then push them? We could put that into its own workflow that the deployment workflows dispatch but I'm starting to get dizzy thinking about that :dizzy_face: Maybe it wouldn't be so bad? What do you think? I don't think I'm making too much of the docs being ahead of production. Rollbacks are exceptional enough that we could ignore that case, at least for now... then all we'd have to do is rebuild the docs during Release app :thinking: hmmm...

sarayourfriend (commented):
Not a real review yet because I caught this just before logging off for the day and was excited. Just took a peek and I want to do a better review when I am back on Wednesday, but some initial thoughts are here. I especially wanted to share the first comment early as it seems significant and I hadn't considered it before.

- See the comment about production docs syncing. Not sure whether that is a blocker.
- We should truncate the waveform sample response. It's something like 3/5ths of the page length at the moment :joy: 
- drf-spectacular seems great, nice find!
-

dhruvkb:
@sarayourfriend your point about the docs being out of sync with the API is a good one. I'll separate out the part of the PR related to deploying the docs and focus only on the `drf-yasg` to `drf-spectacular` migration.

obulat (commented):
This is a really exciting PR, @dhruvkb! I will be adding more review comments, but I'd like your opinion on the best ways to handle some cases that need updates, but might be out of scope of this PR? Such as the "Add deprecation docs about watermark endpoint" or "Update the link to Openverse Syntax guide" (it currently links to a page on creativecommons).

A meta-issue with points to address after this PR is merged might work well.

api/catalog/api/docs/base_docs.py
```diff
@@ -0,0 +1,85 @@
...
+from drf_spectacular.utils import OpenApiExample, extend_schema
+
+
+refer_curl = (
+    "You can refer to the cURL request samples "
+    "for examples on how to consume this endpoint."
+)
+
+
+def fields_to_md(field_names):
```
obulat, line 14: Do you think it brings a lot of value to have a list of field names, if we give all of them below anyways?

api/catalog/api/serializers/media_serializers.py
```diff
@@ -48,10 +59,11 @@ class MediaSearchRequestSerializer(serializers.Serializer):
         "extension",
         "mature",
         "qa",
-        # "unstable__sort_by",  # excluding unstable fields
-        # "unstable__sort_dir",  # excluding unstable fields
-        # "unstable__authority",  # excluding unstable fields
-        # "unstable__authority_boost",  # excluding unstable fields
+        # Excluded unstable fields, also see `exclude_fields` above.
```
obulat, line 62: Do we have to duplicate the list of excluded fields? What would happen if you removed the commented out fields here?

api/catalog/api/serializers/media_serializers.py
```diff
@@ -471,8 +466,9 @@ class MediaSearchRequestSourceSerializer(serializers.Serializer):
         """
 
         _field_attrs = {
-            "help_text": make_comma_separated_help_text(
-                search_controller.get_sources(media_type).keys(), "data sources"
+            "help_text": (
+                "A comma separated list of data sources; for valid sources see "
+                f"https://api.openverse.engineering/v1/{media_path}/stats/."
```
obulat, line 471: Is there a way go get the base URL programmatically here? If we ever change the version or the base URL?

api/catalog/configuration/spectacular.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Openverse is a search engine for openly-licensed media. The Openverse
+        API is a system that allows programmatic access to public domain digital
+        media. It is our ambition to index and catalog billions of
+        openly-licensed works, including articles, songs, videos, photographs,
+        paintings, and more. Using this API, developers will be able to access
+        the digital commons in their own applications.
+        """
+    ),
+    "TOS": "https://docs.openverse.org/terms_of_service.html",
```
obulat, line 18: Can we move the TOS to the API consumer docs somehow? Not necessarily in this PR.

api/catalog/api/docs/base_docs.py
dhruvkb: It's not very useful, I'd be willing to remove it if more people feel like it's not useful.

api/catalog/api/serializers/media_serializers.py
obulat: Would be nice to add something like "please use the source_name as a value"

api/catalog/api/serializers/media_serializers.py
dhruvkb: This list is only used to generate the names of fields in the docs. They are listed here and commented so that they can be un-commented when the fields are finally considered stable.

.github/actions/build-docs/action.yml
dhruvkb: You're right @sarayourfriend. I've reverted the changes related to merging of the Redoc and the docs site, keeping this slightly more focused on the upgrade to `drf-spectacular`.

api/catalog/api/serializers/media_serializers.py
sarayourfriend: Using Django's `resolve` would do it, I believe (and have the correct links per environment, too) but when I last tried to implement that it caused circular imports :cry:

sarayourfriend (approved):
LGTM, a welcome improvement. Kudos for finding the new library.

dhruvkb:
@obulat was the one who discovered it, and named it in the issue.

obulat (approved):
> So I think as long as there is a way to access the API without authentication, this section will have to be null.

I was selfishly hoping to test the `unstable__` search query parameters from the docs :) Really hope we can add this in the future, but it is surely out of scope for this PR.

Thank you for handling this migration so quickly, and enabling us to improve the data flows between the stacks!

obulat:
> @obulat was the one who discovered it, and named it in the issue.

The library is featured very prominently on the `drf-yasg` docs 😄 
 https://drf-yasg.readthedocs.io/en/stable/readme.html#openapi-3-0-note

api/catalog/management/commands/validateopenapischema.py
```diff
@@ -1,42 +0,0 @@
...
-
-        validate_spec(spec_dict, validator=openapi_v2_spec_validator)
-
-        invalid_paths = [p for p in spec_dict["paths"].keys() if not p.endswith("/")]
-
-        assert invalid_paths == [], (
-            "All paths must end with a trailing slash. "
-            f"Found {len(invalid_paths)} missing a trailing "
-            f"slash: {', '.join(invalid_paths)}"
-        )
```
sarayourfriend, line 33: @dhruvkb I've just realised that removing this management command accidentally ended us validating the trailing slashes. I didn't catch this during the PR review :cry: 

I created this issue to follow up on this: https://github.com/WordPress/openverse/issues/2452

PR #1884, opened 2023-04-24
Add checks for required parameters to Provider scripts (K-R)

### LINKED ISSUE #1883: Provider API value checks should filter out all falsy values, not only `None`
## Problem

In Provider API scripts, we often check if a required value exists using `if <value> is None` checks:

https://github.com/WordPress/openverse/blob/f5c6c7ad9a34c25c8a7afafcee7b95b34686e625/catalog/dags/providers/provider_api_scripts/brooklyn_museum.py#L119-L124

However, this lets other falsy values (such as empty strings) to pass the test, which might cause issues like https://github.com/WordPress/openverse-catalog/pull/1102

## Description

Since non-None falsy values (empty strings, empty dictionaries) do not make sense as values for required parameters such as foreign landing url or url, we should replace the `is None` checks with `if value` checks.

## Alternatives

The current solution with checks for `None` works in most cases, and the checks could be done in the `MediaStore` classes. However, it moves the check back in the ingestion process, and makes it longer unnecessarily.

## Additional context

The template for provider scripts should be updated, too.

obulat:
## Fixes

Fixes #1883 by @obulat

## Description

This PR replaces `is None` checks for required values with `if <value>` checks because falsy values such as empty string or empty list should also short-circuit the provider API scripts.

## Testing Instructions

All the CI tests should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/providers/provider_api_scripts/museum_victoria.py
+++ b/catalog/dags/providers/provider_api_scripts/museum_victoria.py
@@ -126,7 +126,7 @@ class VictoriaDataIngester(ProviderDataIngester):
         height, width, filesize = None, None, None
         media_data = {}
         for size in ["large", "medium", "small"]:
-            if size in media and "uri" in media[size]:
+            if isinstance(media.get(size), dict) and media[size].get("uri"):
                 media_data = media[size]
                 break
         if image_url := media_data.get("uri"):
--- a/catalog/dags/providers/provider_api_scripts/museum_victoria.py
+++ b/catalog/dags/providers/provider_api_scripts/museum_victoria.py
@@ -1,6 +1,7 @@
 import logging
 from typing import TypedDict
 
+from common import constants
 from common.licenses import LicenseInfo, get_license_info
 from common.loader import provider_details as prov
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
@@ -64,16 +65,14 @@ class VictoriaDataIngester(ProviderDataIngester):
             }
 
     def get_record_data(self, data: dict):
-        object_id = data.get("id")
-        if object_id in self.RECORDS_IDS:
+        if not (object_id := data.get("id")) or object_id in self.RECORDS_IDS:
             return None
         self.RECORDS_IDS.add(object_id)
         foreign_landing_url = f"{self.LANDING_PAGE}{object_id}"
 
-        if (media_data := data.get("media")) is None:
-            return None
-        images = self._get_images(media_data)
-        if len(images) == 0:
+        if not (media_data := data.get("media")) or not (
+            images := self._get_images(media_data)
+        ):
             return None
         meta_data = self._get_metadata(data)
         title = data.get("displayTitle")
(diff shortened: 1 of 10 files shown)


AetherUnbound (approved):
Nice! I'm surprised there were not more tests which were affected. Do you think it makes sense to go through and add test cases for the providers which check against a few falsy values?

obulat:
@AetherUnbound, I re-requested your review because this PR has changed a lot since your approval.

catalog/tests/dags/providers/provider_api_scripts/test_metropolitan_museum.py
```diff
@@ -82,8 +82,6 @@ def test_get_batch_data(response_json, expected):
             full_expected_data[0].get("meta_data"),
             id="full_object",
         ),
-        pytest.param({}, None, id="empty_dict"),
-        pytest.param(None, None, id="None"),
```
krysal, line 86: Why are these cases removed? I think it would be good to keep them.

krysal (approved):
LGTM. Thank you for caring about splitting the changes for easier reviews and adding type hints ✨

catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -126,11 +126,10 @@ def _get_image_data(
         height, width, filesize = None, None, None
         media_data = {}
         for size in ["large", "medium", "small"]:
-            if size in media:
+            if size in media and "uri" in media[size]:
                 media_data = media[size]
                 break
-        image_url = media_data.get("uri")
-        if image_url is not None:
+        if image_url := media_data.get("uri"):
```
stacimc, line 132: Could `image_url` still be falsy here?

catalog/dags/providers/provider_api_scripts/nypl.py
```diff
@@ -145,6 +144,20 @@ def get_record_data(self, data):
             images.append(image_data)
         return images
 
+    @staticmethod
+    def _get_title(mods: dict) -> str:
+        title_info = mods.get("titleInfo")
+        if isinstance(title_info, list) and title_info:
+            title_info = title_info[0]
+        return "" if title_info is None else title_info.get("title", {}).get("$")
```
stacimc, line 152: Should this `is None` check also be changed to a check for falsy values?

stacimc (approved):
LGTM, no blocking requests. Thanks for all the additional cleanup and type hints added as well :)

catalog/dags/providers/provider_api_scripts/museum_victoria.py
```diff
@@ -126,11 +126,10 @@ def _get_image_data(
         height, width, filesize = None, None, None
         media_data = {}
         for size in ["large", "medium", "small"]:
-            if size in media:
+            if size in media and "uri" in media[size]:
```
obulat, line 129:
```suggestion
            if isinstance(media.get(size), dict) and media[size].get("uri"):
```

catalog/dags/providers/provider_api_scripts/museum_victoria.py
obulat: I think it shouldn't because we check that `media[size].get("uri")` is not falsy above (I updated the check).

obulat: Unless we didn't find any size in `media`, in which case `media_data` is still `{}`. So we still have to have the check.

catalog/dags/providers/provider_api_scripts/nypl.py
obulat, line 152:
```suggestion
        return "" if not title_info else title_info.get("title", {}).get("$")
```

PR #1889, opened 2023-04-24
Handle more than one value for a url query parameter

### LINKED ISSUE #453: `query[queryDataKey].split is not a function` in search query transform
## Sentry link

https://openverse.sentry.io/share/issue/f7bcd3d050574f78b23936e923453e7c/

## Description

```
TypeError: query[queryDataKey].split is not a function
  File "server.js", line 12723, col 52, in <anonymous>
  File "server.js", in Array.forEach
  File "server.js", line 12709, col 15, in queryToFilterData
  File "server.js", line 4073, col 124, in Object.setSearchStateFromUrl
  File "app:///pinia.prod.cjs", line 341, col 30, in Object.<anonymous>
    ret = action.apply(this && this.$id === $id ? this : store, args);
  File "pages/search.js", line 1948, col 17, in asyncData
  File "server.js", in runMicrotasks
  File "node:internal/process/task_queues", line 96, col 5, in processTicksAndRejections
```

This stacktrace was a bit difficult to decode, but I was able to pin it down to this line:

https://github.com/WordPress/openverse-frontend/blob/94fb35469ad28cc3cfd3d1b9ef877b695925a736/src/utils/search-query-transform.ts#L202

It appears that, while `qquery[queryDataKey]` may not be falsy from the check above, it might not be the correct type which is expected of an object that has the `.split` function available.

## Reproduction

I was unable to reproduce this myself.

### LINKED ISSUE #1046: Frontend URL parsing: filterParameter.split is not a function
## Sentry link

https://openverse.sentry.io/share/issue/44140b9540954051b37d00f635382d4c/

## Description

We need to validate that `filterParameter` is a string and not an array of strings before trying to process it. If the parameter is an array, we should probably just use the first one because we join the values with a comma to pass 2 values (so, `license=by,cc0` instead of `license=by&license=cc0`).

## Reproduction

Go to a URL with 2 `sources` params: 
https://openverse.org/search/image?q=cat&source=thingiverse&source=images

obulat:
## Fixes

Fixes #1046 by @obulat 
Fixes #453 by @AetherUnbound 

## Description

Vue router supports having separate values for a single parameter in the URL: `q=cat&q=dog`. This is parsed to `context.query` as `{ q: ["cat", "dog"] }`. Our code, however, expects the values for parameters to be strings, not arrays. This causes a 500 error in Nuxt for such URLs.

This PR pre-processes the `query` received from Nuxt context to leave only the first parameter if an array is given. We can safely discard the other values because normally we do not set such parameters using the UI. If we need to pass several values, we use a comma-separated value: `{ license: "by,sa" }` instead of an array. 

## Testing Instructions

Run the app and go to `localhost:8443/search/?q=cat&q=dog`. You should receive the results for `cat`, and `dog` should be discarded as a search term. If you do the same on main, you would get a Nuxt error screen.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/utils/search-query-transform.ts
+++ b/frontend/src/utils/search-query-transform.ts
@@ -14,6 +14,9 @@ import {
 import { getParameterByName } from "~/utils/url-params"
 import { deepClone } from "~/utils/clone"
 
+import type { Context } from "@nuxt/types"
+import type { Dictionary } from "vue-router/types/router"
+
 export interface ApiQueryParams {
   q?: string
   license?: string
@@ -166,7 +169,7 @@ export const queryToFilterData = ({
   searchType = "image",
   defaultFilters,
 }: {
-  query: Record<string, string>
+  query: Dictionary<string>
   searchType: SupportedSearchType
   defaultFilters: Partial<Filters>
 }) => {
@@ -262,11 +265,23 @@ export const areQueriesEqual = (
 }
 
 /**
- * Converts the API query `q` parameter to a single string that can
- * be used as a search term. Use the first item if `q` is an array.
- * @param q - the URL `q` parameter
+ * The URL query string can contain multiple values for the same parameter.
+ * This function converts the query string to a dictionary where the values
+ * are always strings. If the parameter has multiple values, the first value
+ * is used.
+ * * @param queryDictionary - the query param dictionary provided by Vue router
  */
-export const qToSearchTerm = (q: string | null | (string | null)[]) => {
-  const searchTerm = Array.isArray(q) ? q[0] : q
-  return searchTerm ? searchTerm.trim() : ""
+export const queryDictionaryToQueryParams = (
+  queryDictionary: Context["query"]
+): Dictionary<string> => {
+  const queryParams = {} as Dictionary<string>
+  Object.keys(queryDictionary).forEach((key) => {
+    const value = queryDictionary[key]
+    // If the parameter is an array, use the first value.
+    const parameter = Array.isArray(value) ? value[0] : value
+    if (parameter) {
+      queryParams[key] = parameter
+    }
+  })
+  return queryParams
 }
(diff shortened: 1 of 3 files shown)


frontend/src/utils/search-query-transform.ts
```diff
@@ -14,6 +14,9 @@ import {
 import { getParameterByName } from "~/utils/url-params"
 import { deepClone } from "~/utils/clone"
 
+import type { Context } from "@nuxt/types"
+import type { Dictionary } from "vue-router/types/router"
```
sarayourfriend, line 18: Just a curiosity: what is the benefit of this type vs `Record<string, string>`?

sarayourfriend (commented):
This looks fine to me, but I question whether this is the right approach (discarding all but the first item) for fields other than `q`. I understand why `q` should just be a single one as it's hard to assume what the intention is (should it be `AND` or `OR` or something else entirely), but for the other ones, why wouldn't we just combine them into a comma separated string instead?

I guess part of this would be answered if I knew more about the context of _when_ this happens to real users. What is the situation where this occurs? If we explicitly do not want to support this type of query, then I think the more expected approach would be to return a 400 Bad Request rather than trying to assume what the requester intended.

obulat:
> I guess part of this would be answered if I knew more about the context of _when_ this happens to real users. What is the situation where this occurs? If we explicitly do not want to support this type of query, then I think the more expected approach would be to return a 400 Bad Request rather than trying to assume what the requester intended.

I think this happens when the users are constructing the URL manually. Moreover, this might be an attempt to hack something as the `q` in such requests can be unusual.

sarayourfriend (approved):
LGTM! I still think returning a 400 instead of still trying to process is the best solution for API usage we do not want to support (especially if we think it might be related to abusive usage, programmatic?), but treating the behaviour as "undefined" and doing whatever is easiest is also fine.

Issue #410, opened 2023-02-16
Additional search views for the frontend


obulat:
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2023-04-01 | 2023-05-30 | TBD          | TBD              |

## Description

To improve the search experience on openverse.org, we want to add more ways of displaying search results:
- add a landing page for providers. These pages would only display media from the selected provider and will also serve as a landing page for the provider on Openverse.
- add pages for tags, making them clickable. These pages would probably look similar to provider landing pages and would show only the media with the selected tag.
- creator landing pages would be similar to the previous pages and would show all the media in Openverse by a specific creator.

As a start, we can create and implement a single template for all three types of pages.

## Documents

- [ ] Project Proposal
- [ ] Implementation Plan

zackkrida:
@WordPress/openverse-maintainers, @panchovm has expressed interest in leading this project and is going to start the kickoff post next week. He'll be AFK some days at the start of April which was the original estimated start date for this project.

fcoveram:
Project Prosal was merged (#1890) and project is ready to continue with Implementation Plans

fcoveram:
As noted [here](https://github.com/WordPress/openverse/issues/2113#issuecomment-1622214018), designs were approved due to no discrepant comments. I will prepare the assets for their hands-off and comment again once done.

As discussed during the prioritization meeting, @obulat will lead the implementation and will write the involved plan.

fcoveram:
* Implementation plan is starting soon
* A new use case for content with several tags was shared in #2589

fcoveram:
Implementation plan started a few days ago and has been discussed in #2676

zackkrida:
Implementation plan was approved and issues have been created. @obulat has started work.

fcoveram:
Dev work continues, and a [Frontend milestone](https://github.com/WordPress/openverse/milestone/17) was created.

To the above, I think we need to add the following tickets to the milestone as they address part of the changes
* #2840 
* #2589

What do you think @obulat?

obulat:
I added the issues to the milestone, and updated the comment to rename "Frontend milestone" to "Milestone" because it contains some API-related issues.

obulat:
## Update 2023-09-06

Merged so far: `additional_search_views` feature flag and `VCollectionHeader` component added, `VAudioCollection` component extracted.

[...]

Issue #630, opened 2022-01-24
Replace external artist links with links to the relevant creator search on Openverse


sarayourfriend:
## Problem

Rather than link to the external resource for the creator, which may or may not give the user quick access to the artist/creator's content, let's replace the links and mentions of creator names everywhere with links to the relevant Openverse search.

## Description

According to @zackkrida we can use the following format to build the search link:

```
/search/audio?searchBy=creator&q=<creator name>&source=<creator source>
```

We need to include the source because there could be two separate creators by the same name on different sources and they should be distinguished (we don't want to misattribute).
 
We could also explore making the source clearer on the artist name to help disambiguate multiple sources coming back with the same creator name representing separate people.

We could also continue to include the external link to the creator's page on the source when it exists as a separate link (?).

This probably needs some extra thought from @panchovm on the UX side to make sure we're delivering the correct information and not making things more confusing.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
We could, yup! If you do, please add it to the project thread. We handled an existing issue similarly for the #377 project (specifically https://github.com/WordPress/openverse/issues/750).

obulat:
I added this issue to the Prior Art section in Project thread #410. Closing this as won't do.

PR #1890, opened 2023-04-24
Project Proposal: Additional search views

fcoveram:
## Due date:
2023-05-12

## Assigned reviewers
- [x] @obulat 
- [ ] @krysal 

## Description
This PR is the kick-off of the Additional search views project: #410


documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Participants and stakeholders
+
+- Design: @panchovm
+- Implementation:
+
+## Infrastructure
+
+## Accessibility
+
```
panchovm, line 33: Do you think we need to add details in the Requirements, Infrastructure, Accessibility, and Required implementation plans sections?

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,36 @@
...
+## Marketing
+
```
panchovm, line 35: Since this projects adds views where people can browse content without a search query, there is a comms and marketing opportunity to convey how openverse is moving towards more ways of discovering content. What do you think?

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,36 @@
...
+## Goals
+
+Search Experience
+
+## Requirements
+
+## Success
+
+The project will be considered successful when each tag, provider, and creator
+has a standalone and shareable page
```
obulat, line 23: I think we should add analytics here, too:
Something like "And we have the analytics events set up to measure how users are using them"

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
obulat: Accessibility: we should mention how page structure and headings on the pages affect it.

Implementation plans: we need the design issue (would 1 be enough, or do you plan to have an issue per search view, @panchovm ?) as an implementation plan. I am not sure that we need an IP for the code. Maybe a single issue is enough? I don't think we need any changes in the API, do we?

panchovm: One design issue introducing mockups for each case is suficient. Regarding changes in the API, I would defer to @krysal or ping maintainers folks in case we need more visions.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
panchovm: I'm not sure about this. We don't have a criterion to mark its utility in relation with other search flows or as a standalone browsing experience. We surely need to assess how these pages are used, but as inputs for understanding how people search and browse and to polish the tool rather than labeling this project as success or not.

obulat: By "adding analytics" I meant adding the code that will send an analytics event when the user opens one of the additional search views. So, when a user opens a provider page, we send an event to the Analytics server with the provider name and other relevant details. Without this event, we would not be able to see if the users are actually opening the page or not. We can, of course, add the events after this project is finished. But I think it would be useful to add the analytics events to any important frontend user-facing changes. Just like we add tests for any visual changes.

This **does not** include the actual collection and analysis of this data. I wasn't suggesting that we decide whether this project is successful by the number of users who opened the views.

Does this comment make my point clearer, @panchovm ? Or do you think that we should not add any code that sends analytics events as part of this project?

dhruvkb: Analytics tracks pageviews by default, so at the very least we should be getting information about how many visits these pages are receiving.

obulat: I think it would be useful to have a separate Plausible event, with properties like {"searchView" : <"tag"|"provider"|"creator">, "detail": <"tagName"|"providerName"|"creatorUrl/creatorName/creatorProvider">} to track this.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,36 @@
...
+The Openverse team has decided to enhance the browsing experience by
+incorporating new content views for providers, tags, and creators. This feature
+will allow users to share collections detached of individual searches.
+
+## Goals
+
+Search Experience
+
+## Requirements
+
```
krysal, line 19: I remember we discussed something on our last meetup related to the navigation with tags, like how the terms in tags will apply in the search, as `or` or `and` logical operations? Do you think is worth adding those here as requirements for the project? It would be helpful to know how they would be applied and to limit ourselves to specific functionalities :)

Other points that could be specified here:
- will the tags from the current details page be clickable to navigate to these new views?
- the new views will include the filter sidebar?
- the new views will show the lastet images/audio added to Openverse first?

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
krysal: Yeah, I don't think we need changes on the API side since it already provides ways for searching by providers, tags, and creators. The frontend will only need to assemble the url with required query parameters.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,36 @@
...
+has a standalone and shareable page
+
+## Participants and stakeholders
+
+- Design: @panchovm
+- Implementation:
+
+## Infrastructure
+
+## Accessibility
```
krysal, line 32: A call for the a11y team of the WordPress project for revision of the mockups would be nice here.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
krysal: Agree! It'd be a good oppportunity to promote the project as well.

krysal (commented):
Thanks for starting this proposal! I left a few questions and suggestions. The main section that I see needs more some information is the **Requirements** as this will guide us on what to implement here. I remember you show us some mockups time ago and those sparkled several questions, that will be a good place to start I think.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
sarayourfriend: FWIW custom events cost money (each one counts towards our "pageviews" for the billable period), so if we can derive the information from an existing one, it's worth doing so. If we truly need another event, of course that is fine, but we should keep in mind the cost and avoid duplicating or denormalising event data.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
sarayourfriend: > I am not sure that we need an IP for the code. Maybe a single issue is enough? I don't think we need any changes in the API, do we?

I think we should have implementation plan(s) for the frontend changes. If the changes are simple then the plan will be small and easy to review and won't add more than a week and a half to the project (maybe less if the process goes faster for a very small IP, as is allowed). We should not skip implementation planning for new views or wholly new designs.

I would suggest two requested issues:

1. The design issue
2. One implementation plan at minimum, to be split depending on the requirements surfaced during design review

> Yeah, I don't think we need changes on the API side since it already provides ways for searching by providers, tags, and creators. The frontend will only need to assemble the url with required query parameters

This is sort of true, but the method for querying is different. When you add individual field queries to the search request, the search term is ignored ([described in detail in our documentation](https://docs.openverse.org/api/reference/search_algorithm.html#individual-field-querying)).

If the design changes allow for doing regular searches against the subset of data, then we would need API changes to accommodate that for sure and it should have its own implementation plan (a third, counting the two I mentioned before).

Anyway, all of that to say, I would make the final list of requested implementation plans dependent on completed designs unless this project proposal is significantly expanded to describe the specific requirements and expected behaviour of the new views. Right now the described requirements are so broad that it's impossible to know what the outcome of the designs would include. This is dangerous because we do not have any guiding light to understand how to cut scope.

sarayourfriend (commented):
I would prefer to see an expansion of the requirements section with detailed expected outcomes: specifically, how the views would work on a general level. This is important for guiding the design discussion, understanding the scope of the project, and even guessing at how many implementation plans would be needed. The existing requirements description is so broad that the number of required implementation plans could be anywhere from as little as a single, small implementation plan to several (up to 3 or 4 I can easily imagine) significant plans, including one for API changes.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
obulat: > If the design changes allow for doing regular searches against the subset of data, then we would need API changes to accommodate that for sure and it should have its own implementation plan (a third, counting the two I mentioned before).

This is a great point to clarify in the proposal! My assumption (based on the what we talked about during the 2023 planning) was that the first iteration that we talk about in **this** project would only include a page with the items that are from the chosen provider, creator or have the selected tag. Roughly, the page would have a large title (tag/provider/creator) and a grid of results for the relevant tag/provider/creator.

I was assuming that it is not possible to do a search within this page. So, you only list all the items with the selected tag, and not search for "parrot" within the images tagged as "parrot".

But I realized after reading the comments that this proposal does not contain information that would support or disprove my assumptions. So, similar to other reviewers, my request is to clarify the scope: is searching within the selected tag/provider/creator supported or not? How are the items sorted (as they come from Elasticsearch, assuming by their popularity score, or by the date when the items were indexed by Openverse)?

obulat: > This is sort of true, but the method for querying is different. When you add individual field queries to the search request, the search term is ignored ([described in detail in our documentation](https://docs.openverse.org/api/reference/search_algorithm.html#individual-field-querying)).

I just realized that my assumption about the way these views would work was wrong.

I was assuming that it would be a view that does _filtering_ rather than searching. So, if you click on a "cat" tag, you would get all the images with the exact "cat" tag. However, the `tags` parameter in the API *searches* through tags, so for `tags=bot` you could get the results for tags like `robot` or even `botulism`: https://api.openverse.engineering/v1/images/?tags=bot. This is definitely *not* what I would expect when I click on a tag called "bot"!

The same is true for creator (it would return any results for creator whose name _contains_ the search term, not the exact creator), which, again, is not what I would expect as a user.

If everyone agrees with me on the idea that we should return _exact_ matches for tags and creator, I think we *would* need API changes. We would need to add a `filter` similar to `source` or `licenses` that would return exact matches and not search within the field text.

Even the `creator` search suggested in https://github.com/WordPress/openverse/issues/630 would search for text within the creator's name and not return the items from the selected creator. Here's a sample of what you see if you search for creator "K": https://api.openverse.engineering/v1/images/?source=flickr&creator=k

panchovm: Thank you all for your thoughts.

It seems that the designs are mandatory for continuing with the Project Proposal and the Implementation Plan. I am fine with holding this document until having a level of approval for the visual changes. In that vein, the scope of this project is framed within two main user stories:

1. As an openverse user, I want to see all results associated with a single tag.
2. As an openverse user, I want to see all results coming from a single provider.
3. As an openverse user, I want to see all results created by a single author.

That being said, and recalling a preliminary approach we had when suggesting this project, this search view does not allow any search. It _might_ allow sorting by different criteria and display content per content type or in the "all content" grid, but that is a discussion we should have in the design ticket.

At the same time, I agree with @obulat that user expectations would mean exact matches for both tags and authors.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
panchovm: Awesome. I will state in the document: Track the uses of these new pages by derivating data from existing events.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
panchovm: Good call. Based on how the discussion unfolded, I will ping a11y folks in the design ticket.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
obulat: If we want to implement exact matches, we would also need an implementation plan for the API changes. Or a section in the Implementation Plan that deals with the API changes in addition to the frontend changes. We would need to add `filter`s for tag and creator. Do you agree, @sarayourfriend ?

As for the provider, we might be able to use the `source` filter. We would need to confirm the if the distinction between `source` and `provider` complicates the things too much. For example, if you decide to click on a provider of a Flickr image from NASA, would you be taken to a Flickr page or a NASA page? Both can be retrieved from the API using `source=flickr` or `source=NASA`.

sarayourfriend: Can you clarify what you mean by "add filters for tag and creator"? Are you talking about the ES boolean query "filter" feature or something else?

Exact match queries should be possible using quotations because we escape quotes, but there's virtually no guarantee that that works as expected: https://github.com/WordPress/openverse/blob/697f62f01a32cb7fcf2f4a7627650a113cba40da/api/catalog/api/controllers/search_controller.py#L373

Indeed, I get the same set of results (at least at a glance) when I query tags for `"bot"` and `bot` (with and without quotes).

> It seems that the designs are mandatory for continuing with the Project Proposal and the Implementation Plan

I think we should define the scope of expected behaviour before design work starts in earnest, mostly to give you, Francisco, and anyone reviewing the designs guidance for preventing scope creep while ensuring they are feature complete.

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
sarayourfriend: Just to give info for the current implementation, right now field specific queries (tags, creator, title) use ["simple query string"](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/query-dsl-simple-query-string-query.html) style query which defaults to "OR".

obulat: I suggest to add the following requirements, based on my understanding of what this project is.

The project creates 3 additional views for the Nuxt frontend.
1. View by tags
The page shows the name of the tag, and all the items of the selected media type (image/audio) that have this exact tag applied (so, for `cat` tag view, only images that have a tag `cat`, not `cats` will be shown). The items are sorted by `ingested_on` field in descending order. The user can go to this view directly, by opening www.openverse.org/tags/image/cat or by clicking on one of the tags on the single result page.

2. View by provider
The page shows the name of the provider, [with their logo and description? Or would that be out of scope for this project?], and all the items from that provider sorted by `ingested_on` field in descending order. 
The user can go to this view directly, by opening www.openverse.org/providers/wikimedia/image/ or by clicking on the provider's name on the single result page.
Note: for providers, I'm not sure what we want to do, so here are some questions: 
- Do we want to use `providers` or  `sources`? The difference is that provider is the *site* where we got the image from, and the source _can_ be the same as the provider, but can also be the collection on that provider. For example, NASA publishes images on Flickr. For these images, `source` is NASA, and the provider is Flickr. Naturally, there are more sources than providers. Do we want to create views for all sources, or only for providers?
- For providers with multiple media types, such as Wikimedia or Europeana, do we want to have an "All content" veiw, or only the separate image/audio views?

3. View by creator
The page shows the name of the creator (together with the provider? otherwise the creators will clash because different creators can use the same name on different providers), and all the items from that creator sorted by `ingested_on` field in descending order. The user can go on this view directly, by opening www.openverse.org/creators/cat_flickr (for a creator named "cat" on Flickr), or by clicking on the creator's name on the single result page.

These views will require changes to:
- Nuxt frontend: adding pages for the specific views. They can reuse the grids and result boxes/squares from the search views.  The pages will only show the results, without the ability to search within them. Because of that, the views will use the "content" header (without the search bar, search type selection or filters button) and will not have a filter sidebar.

- API: adding `filter` parameters that would enable retrieving all items of a certain media type that have the specific property (tag/provider/creator and source pair). These will work similarly to the way filter parameters such as `license` work now, returning only the items with the specific filter. Searching within the filtered items is not in scope of this project.

obulat (commented):
@fcoveram, after thinking a lot about the comments in this project, I realized that we haven't fully shared our assumptions here. Because we have different assumptions, we are talking about different things. For example, some comments talk about searching within these views, whereas I was assuming that these views (in the scope of this project, not in the future) will only be the items that are filtered by the specific tag/provider/creator. I do remember you showing us some experimental designs for improvements in searching by tag/provider/creator within the search bar, but I assumed that those are _not_ in scope for this project.

I've written out my assumptions of what this project is as a comment on the project requirements. Very curious if you agree with what I wrote.
I think we should first finalize the requirements and then do the designs, because otherwise you might spend too much time on work that might turn out to be not what we want to do.

fcoveram:
Thank you all for your thoughts.

I added @obulat’s suggestion with minor copy tweaks to make it more clear for non-dev readers. I framed the requirements keeping out the feature of searching within the result pages. I can envision the complexities and uses cases crushing the experience, but I think we can address this challenge in a future release.

### Providers or Sources 

For the provider or source note, I’m drawn to show sources across the whole site. From a user perspective, it's clearer to show which organization provides the content you want to use, no matter what site that content lives on. This also coexists with the filters and how openverse lists sources.

Following Olga's example, if you click on NASA source, landing on the main Flickr site feels odd. I would expect to see a site only displaying NASA content.

---

Since we are tiering this discussion before any design ticket, I will remove the pause tag to continue polishing the document if needed. What do you think of the changes? Do we need something else to continue with the Implementation Plan(s) and the designs?

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,89 @@
...
+## Requirements
+
+The project adds three additional views for the Nuxt frontend where users
+interact in the following ways.
+
+### Results by tag
+
+Users land on the tag results page by clicking on the tag in the single result
+page or by opening a tag results link (for instance,
+[www.openverse.org/tags/image/cat](http://www.openverse.org/tags/image/cat)).
```
obulat, line 27:
```suggestion
[www.openverse.org/tags/image/cat](https://www.openverse.org/tags/image/cat)).
```

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,89 @@
...
+`ingested_on` field in descending order.
+
+For instance, the `cat` tag page will show only images that have the `cat` tag,
+not items tagged with `cats` .
+
+### Results by source
+
+Users land on the source results page by clicking on the source name in the
+single result page or by opening a source results link (for instance,
+[http://www.openverse.org/providers/wikimedia/image/](http://www.openverse.org/providers/wikimedia/image/))
```
obulat, line 40:
```suggestion
[http://www.openverse.org/providers/wikimedia/image/](https://www.openverse.org/providers/wikimedia/image/))
```

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,89 @@
...
+
+The results page shows the name of the source and all the items from this source
+in the selected media type. The items are sorted by `ingested_on` field in
+descending order.
+
+### Results by creator
+
+Users land on the creator results page by clicking on the creator’s name in the
+single result page or by opening a creator results link (for instance,
+[http://www.openverse.org/creators/cat_flickr](http://www.openverse.org/creators/cat_flickr))
```
obulat, line 50:
```suggestion
[http://www.openverse.org/creators/cat_flickr](https://www.openverse.org/creators/cat_flickr))
```

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,89 @@
+# 2023-04-24 Project Proposal
+
+## Reviewers
+
+- [ ] @obulat
```
obulat, line 5:
```suggestion
- [x] @obulat
```

obulat (approved):
I'm really excited for this project! Thank you for putting it all together, @fcoveram!

The only thing I would like to add here is "Requested Plans":
- API 
- Frontend
- Design issue (if that can be called "implementation plan")

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,95 @@
+# 2023-04-24 Project Proposal
```
AetherUnbound, line 1: Small suggestion for uniformity with the rest of our project proposals
```suggestion
# 2023-04-24 Project Proposal: Additional search views
```

AetherUnbound (approved):
I'm not a requested reviewer, but I think this proposal looks great! I have two suggestions:

1. Updating the title of the proposal to be in line with our other proposals (suggestion made on that line)
2. This doesn't show up in the documentation preview because we don't have an `index.md` file for the additional search views. I'm happy to add that if you'd like @fcoveram, the contents of `additional_search_views/index.md` would be ([example](https://github.com/WordPress/openverse/blob/2e82df0105a94f7b85c48321fc3e05c23edc49bd/documentation/projects/proposals/popularity_optimizations/index.md?plain=1)):

```markdown
# Additional Search Views

```{toctree}
:titlesonly:
:glob:

*
```

AetherUnbound:
Awesome, I've merged `main` into this branch, added the `index.md` file, and added the author field in line with #2066 - you should only need to `git pull` now before you make any other changes!

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Results by source
+
+Users land on the source results page by clicking on the source name in the
+single result page or by opening a source results link (for instance,
+[http://www.openverse.org/providers/wikimedia/image/](https://www.openverse.org/providers/wikimedia/image/))
+
+The results page shows the name of the source and all the items from this source
+in the selected media type. The items are sorted by `ingested_on` field in
+descending order.
```
krysal, line 46: Fair point, this query will be for an author on a single provider since we don't have a way to cross the works of an author through different providers/sources at the moment 👍

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,93 @@
...
+Users land on the tag results page by clicking on the tag in the single result
+page or by opening a tag results link (for instance,
+[www.openverse.org/tags/image/cat](https://www.openverse.org/tags/image/cat)).
+
+The results page shows the name of the tag and all the items of the selected
+media type that have this exact tag applied. The items are sorted by
+`ingested_on` field in descending order.
+
+For instance, the `cat` tag page will show only images that have the `cat` tag,
+not items tagged with `cats` .
```
krysal, line 36:
```suggestion
For instance, the `cat` tag page will show only images that have the `cat` tag,
not items tagged with `cats`.
```
Wouldn't `cats` be also relevant for the `cat` page and vice-versa? It would be a bit of a pain to have to add both tags. CC @obulat @AetherUnbound

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
```diff
@@ -0,0 +1,93 @@
+# 2023-04-24 Project Proposal: Additional search views
+
+**Author**: @fcoveram
+
+## Reviewers
+
+- [x] @obulat
+- [ ] @krysal
```
krysal, line 8:
```suggestion
- [x] @krysal
```
🙌

krysal (approved):
It's looking good! Thanks for detailing on the requirements, the scope is clearer now. Can't wait to see your excellent designs for the new pages, Francisco!

documentation/projects/proposals/additional_search_views/20230424-project_proposal_additional_search_views.md
AetherUnbound: I think these would be created dynamically based on the tags which exist for a result, we wouldn't be manually adding any of the tags to existing media. This would make the tags serve as a strict match on other tags, if I'm understanding it correctly.

fcoveram: I will resolve this conversation to continue with the Implementation Plan(s), and in case we include `cats` within the results, I'll go back to this document to update it.

PR #1892, opened 2023-04-24
Raise `UpstreamThumbnailException` for unsuccessful requests

### LINKED ISSUE #676: UpstreamThumbnailException only raised when there are "ambiguous" problems and not for things outside the 2xx range
## Description

Requests does not raise an exception if a link 404s, you have to check the status code or call `raise_for_status` on the response.

## Reproduction

1.  See the response from `https://api-dev.openverse.engineering/v1/images/aaabb3e4-cbd9-4a55-b16f-634fd9d91483/thumb/` which should be 424ing is instead 404ing (check the actual status code of the response, not the response contents which currently relays the upstream issue).

## Additional context
Noticed while testing the [v2.5.11 release](https://github.com/WordPress/openverse-api/releases/tag/v2.5.11) but did not stop the release because we're still no longer 500'ing.

## Resolution

krysal:
## Fixes

Fixes #676 by @sarayourfriend

## Description

This PR extracts the params building from the `photon.get` method to reduce it a bit, and also calls `raise_for_status` from the response, as suggested in the issue, to make sure we raise the correct custom exception from unsuccessful HTTP response codes.

The `requests.HTTPError` class inherits from `requests.RequestException`[^1] so there is no need to add it explicitly to the `except` block.

[^1]: https://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions

## Testing Instructions

1. To test this manually, try replacing one DB entry's URL with the sample image URL provided in the issue:
```
https://upload.wikimedia.org/wikipedia/commons/8/89/Flag_of_the_Turkic_Council.svg
```
2. Make sure that image doesn't have a thumbnail URL; in that case, make it null.
3. Go to the image's thumbnail URL (`http://localhost:50280/v1/images/<uuid>/thumb/`) and make sure you see a response like the following:
```
HTTP 424 Failed Dependency
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "Failed to render thumbnail. 404 Client Error: File Not Found for url: https://i0.wp.com/upload.wikimedia.org/wikipedia/commons/8/89/Flag_of_the_Turkic_Council.svg?w=600&quality=80&ssl=true"
}
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/photon.py
+++ b/api/catalog/api/utils/photon.py
@@ -27,15 +27,7 @@ HEADERS = {
 }
 
 
-def get(
-    image_url: str,
-    accept_header: str = "image/*",
-    is_full_size: bool = False,
-    is_compressed: bool = True,
-) -> HttpResponse:
-    logger = parent_logger.getChild("get")
-    # Photon options documented here:
-    # https://developer.wordpress.com/docs/photon/api/
+def _get_photon_params(image_url, is_full_size, is_compressed):
     params = {}
 
     if not is_full_size:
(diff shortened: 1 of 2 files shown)


api/catalog/api/utils/photon.py
```diff
@@ -58,35 +50,40 @@ def get(
...
+
+def get(
+    image_url: str,
+    accept_header: str = "image/*",
+    is_full_size: bool = False,
+    is_compressed: bool = True,
+) -> HttpResponse:
+    logger = parent_logger.getChild("get")
+    # Photon options documented here:
+    # https://developer.wordpress.com/docs/photon/api/
```
sarayourfriend, line 64: nit: this comment should probably be in the `_get_photon_params` function instead.

api/test/unit/utils/photon_test.py
```diff
@@ -287,3 +284,15 @@ def test_get_successful_https_image_url_sends_ssl_parameter(mock_image_data):
...
+@pook.on
+def test_get_unsuccessful_request_raises_exception():
+    mock_get: pook.Mock = pook.get(PHOTON_URL_FOR_TEST_IMAGE).reply(404).mock
+
+    with pytest.raises(
+        UpstreamThumbnailException, match=r"Failed to render thumbnail."
+    ):
+        photon_get(TEST_IMAGE_URL)
+
+    assert mock_get.matched
```
sarayourfriend, line 298: Isn't this test semi-redundant with the one that uses mocking to force requests to throw an error? I suppose it's more realistic to do it this way (this feels less implementation-detail oriented) but consolidating the two would be helpful for keeping the test purposes clear.

sarayourfriend (approved):
LGTM. Small non-blocking request for test consolidation but otherwise the changes look like a great improvement.

obulat (approved):
LGTM!

Issue #433, opened 2023-02-18
Popularity calculation optimizations (Matview refresh)


obulat:
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2023-03-13 |            | @stacimc     | TBD              |

## Description

Currently, re-calculating popularity values during data refresh takes a long time. We need to optimize the calculations to make data refresh shorter. ~This is currently scheduled for December, but if some data refresh threshold is crossed, we might have to re-prioritize this.~

## Documents

- [x] [Project Proposal](https://github.com/WordPress/openverse/pull/1153)
- [x] [Implementation Plan](https://github.com/WordPress/openverse/issues/1873)

## Issues

- Milestone: https://github.com/WordPress/openverse/milestone/11

AetherUnbound:
One thought I had while thinking about this for the [iNaturalist data refresh complications](https://make.wordpress.org/openverse/2023/02/21/post-inaturalist-data-refresh-status/) is that we could potentially split this up from a single matview into a separate table that gets joined on in a _non materialized_ view which is used during the data refresh. Right now, the popularity matview is the one that the data refresh uses to copy data from the catalog to the API. This means that if the matview is not refreshed, data that exists in the underlying table _will not make it to the API_. This makes the popularity calculation a prerequisite to any new data being added to the API, which seems like it might not be necessary.

Instead, we could have the `image_view` be a standard view which is a left join between `image` and `image_popularity`. The latter _would_ be a materialized view which would contain popularity values, but the difference here is that it could be refreshed _independently of the data refresh itself_! This would free us up from depending on the popularity calculation for bringing new data into the API.

rwidom:
I thought that part of why we don't, for example, have a separate tags table, is because we don't want to join `image` to anything. Is the idea that this would be faster because `image_popularity` is much smaller?

AetherUnbound:
> I thought that part of why we don't, for example, have a separate tags table, is because we don't want to join `image` to anything. Is the idea that this would be faster because `image_popularity` is much smaller?

That's a really good point I had overlooked, I ran a query and found that the standardized popularity field is way more populated than I had initially expected...Perhaps a join like that will have a higher impact than is desirable 😕 

```
deploy@localhost:openledger> select count(*) from image_view where standardized_popularity is not NULL;
+-----------+
| count     |
|-----------|
| 523503895 |
+-----------+
SELECT 1
[...]

PR #1868, opened 2023-04-21
Add `quicktoctree` directive for easier prioritised globbed ToC reuse

sarayourfriend:
## Fixes

Fixes #1861 by @sarayourfriend

## Description

I'm not sure if this is the best solution. However, it's an interesting one, at least. The alternative is really to copy and paste the same `toctree` options in all the places that use `quicktoctree` now.

This fixes the ordering problem raised in the issue without reintroducing the need to duplicate the `toctree` settings everywhere. However, it does so by introducing a custom directive that just extends the base `TocTree` directive. Maybe that's _too_ clever and we really should just copy/paste those `toctree` settings everywhere.

## Testing Instructions

Try out the new `quicktoctree` directive and confirm it works identically to `toctree` except for the new `type` option which adds the new behaviours.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [N/A] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
[...]

Issue #412, opened 2023-02-16
Clearly document all media properties


obulat:
| Start Date | Project Lead | Actual Ship Date |
| ---------- | ------------ | ---------------- |
| 2023-03-01 | @dhruvkb         | TBD              |

## Description

All the details we collect and store for each media type should be clearly documented, and should be consistent across the stack (Catalog, API and Frontend).

Due to historic reasons during the development of the project, not all properties that are deemed required are available in the data. As a pre-requisite to making the data more consistent, we need to document all the media properties and whether they are available (should be available in the future) in the database or are added by the API/ingestion server/elasticsearch/frontend. 

Also, we need to document whether the properties are nullable, and what their constraints are. Although we have the database constraints, some of them are not reliable.

## Documents

- [x] #855 

## Implementation Plans

- [x] #1995
- [x] #1996

## Prior Art

- Excellent table for `image` fields by @rwidom : https://docs.google.com/spreadsheets/d/1gaVsvFnsYby2iwzRm0Ta9IPhvNixCpQdHByGLo1r_pg/edit?usp=sharing ("image table ddl" tab)
- Related analysis was attempted in the Data normalization issue https://github.com/WordPress/openverse/issues/244

obulat:
## Update 2023-04-26



The [project proposal](https://github.com/WordPress/openverse/pull/1153) has been finalized, and the PR with the Catalog implementation plan (#1904) opened.
The migration to `drf-spectacular` that creates OpenAPI 3 spec in the API is also the foundation work for the API and frontend parts of this project.


### Blockers



None.

dhruvkb:
This project has been shipped, efforts are underway to describe the properties manually across catalog, API and frontend.

zackkrida:
I've begun work on https://github.com/WordPress/openverse/issues/4025 and should have a PR out in the next 1-2 days.

dhruvkb:
#4366 is the final PR which will close #2187 and also this project.

PR #1894, opened 2023-04-24
Implementation Plan: Decoupling Popularity Calculations from Data Refresh

### LINKED ISSUE #1873: Implementation Plan: Popularity calculation optimizations
## Problem

Write an implementation plan for the Popularity Optimizations project, with the goal of re-enabling the data refresh.

Project thread: #433
Project proposal: #1153

## Description

I'll be proposing a plan for decoupling popularity calculations from the data refresh by calculating standardized popularity at time of ingestion.

stacimc:
## Due date:

2023-05-08

## Assigned reviewers

- [x] @AetherUnbound 
- [x] @obulat

Feedback welcome from all; @AetherUnbound and @obulat chosen as reviewers for expertise in the Catalog. Also calling out in advance: because of  folks' availability this week I anticipate that the clarification round may need to be extended, and am happy to do so!

Also pinging @rwidom, whose insight here would be very valuable.

## Description

Fixes #1873 by @stacimc 
Implementation plan for project #433.

## Current round
This discussion is following the Openverse decision-making process. Information about this process can be found [on the Openverse documentation site](https://docs.openverse.org/reference/decision_making/index.html). Requested reviewers or participants will be following this process. If you are being asked to give input on a specific detail, you do not need to familiarise yourself with the process and follow it.

The deadline for review for this round is **May 10, 2023**.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
+++ b/documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
@@ -1,0 +1,486 @@
+# Implementation Plan: Decoupling Popularity Calculations from the Data Refresh
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @AetherUnbound
+- [ ] @obulat
+
+## Project links
+
+<!-- Enumerate any references to other documents/pages, including milestones and other plans -->
+
+- [Project Thread](https://github.com/WordPress/openverse/issues/433)
+- [Project Proposal](https://docs.openverse.org/projects/proposals/popularity_optimizations/20230406-project_proposal_popularity_optimizations.md)
+
+## Overview
+
+<!-- A brief one or two sentence overview of the implementation being described. -->
+
+This document describes a method for separating the popularity calculation steps
+from the data refresh process, allowing the data refresh to be re-enabled
+independently of popularity calculations.
+
+## Background
+
+To improve search relevancy, we boost records in our search results relative to
+a normalized popularity score. Because the process of calculating these scores
+and pushing them into the API DB (referred to as "data refresh") is complex, it
+may be beneficial to briefly outline the tables and views currently used.
+
+For simplicity, I will use `image` as an example, but note that equivalent
+structures exist for `audio` and will exist for any media types added in the
+future. Descriptions are not exhaustive; some details that are not relevant have
+been omitted for brevity.
+
+#### `image`
+
+The 'main' image table. Rows in this table may contain **raw** popularity data
+in the `meta_data` column, such as the number of times this image has been
+viewed or downloaded.
+
+This table is updated by **ingestion**, which may:
+
+- insert newly ingested images
+- update previously ingested images (for example, by updating raw popularity
+  data)
+
+#### `image_popularity_metrics`
+
+A table which contains known 'metrics': that is, the name of the meta*data field
+which contains raw popularity data for each provider. \_Not all providers have a
+configured popularity metric.*
+
+This table only needs to be updated when a metric is added or modified.
+
+#### `image_popularity_constants`
+
+A materialized view which calculates, for each provider that supports popularity
+data, a popularity constant that can be used to normalize raw popularity scores.
+This constant is calculated using the 85th percentile of the given metric (total
+views, etc). Because of this, the constant needs to be periodically updated for
+accuracy as more data is consumed.
+
+Refreshing this view results in the constants being recalculated. We currently
+do this once a month. It must also be done whenever a new metric is added.
+
+#### `image_view`
+
+A materialized view sourced from the `image` table, which adds the calculated
+`standardized_popularity` column. This normalized popularity score is calculated
+by a simple operation using the raw score and the popularity constant.
+
+The effect of refreshing the view is to:
+
+- Add all images which were ingested since the last time the view was refreshed,
+  and calculate their popularity
+- Update existing records which may have been modified on the `image` table
+  since the last refresh
+- Update the `standardized_popularity` for all records to use _the current
+  popularity constant_
+
+In a data refresh, the data from this view is copied into a new table in the API
+DB, which is used to build new Elasticsearch indices and eventually swapped with
+the live tables. As a result, the view **must** be refreshed before the data
+refresh can run, or newly ingested data would not make it to the API.
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+The basic strategy proposed here for decoupling popularity calculations from the
+data refresh is:
+
+1. Add `standardized_popularity` as a `nullable` column on the `image` and
+   `audio` tables. ([link](#add-standardized-popularity-to-the-media-tables))
+1. Update the `provider_dag_factory` to calculate `standardized_popularity` **at
+   ingestion**.
+   ([link](#Update-provider-DAGs-to-calculate-standardized-popularity))
+1. Create a new `popularity_refresh` DAG
+   ([link](#create-a-popularity_refresh-dag)) which:
+   1. Recalculates the popularity constants
+   1. Updates the `standardized_popularity` score for existing records, using
+      the new constants.
+1. Remove the popularity calculation steps from the data refresh DAGs.
+   ([link](#remove-popularity-steps-from-the-data_refresh-dags))
+1. Begin running the data refresh directly from the media tables, and drop the
+   materialized views. ([link](#run-the-data-refresh-from-the-media-tables))
+
+More details about each of these steps, and the order in which they must be
+implemented, is included in the following sections.
+
+### Add standardized popularity to the media tables
+
+This step is simple:
+
+```pgsql
+ALTER TABLE image ADD COLUMN standardized_popularity DOUBLE PRECISION;
+ALTER TABLE audio ADD COLUMN standardized_popularity DOUBLE PRECISION;
+```
+
+The column can (and indeed _must_) be nullable, as not all providers support
+popularity data. Because this configuration exactly matches the schema of the
+current `image_view` and `audio_view` matviews, we will be able to begin running
+data refreshes directly off these tables with very few changes in the ingestion
+server.
+
+It should be noted that conventionally, storing calculated fields would be
+considered a violation of database normalization principles. However, doing so
+in this case will give us much better performance of the data refresh and also
+save a significant amount of space, as it will enable us to drop the
+materialized views.
+
+### Update provider DAGs to calculate standardized popularity
+
+Using the `ProviderDataIngester`, we can easily add standardized popularity
+calculation to all of our providers at once [^1].
+
+1. In `provider_dag_factory.py`, add a new
+   `get_popularity_constants_and_metrics` task which runs prior to `pull_data`.
+   This task will fetch the `metric` and `constant` for this provider from the
+   `<media>_popularity_metrics` table.
+   1. We will need to fetch metrics and constants for _each media type_
+      supported by the provider, and return it as a dictionary. This information
+      will be posted to XComs.
+1. In `provider_dag_factory.py`, read the popularity data from XComs and pass it
+   to the `pull_data` task and into the `init` for the `ProviderDataIngester`.
+1. In `ProviderDataIngester`, create a `get_popularity_enriched_record_data`
+   method as a wrapper for the abstract `get_record_data`. It should:
+   1. Call `get_record_data`
+   1. Calculate standardized popularity using the popularity constant and metric
+      passed in from XComs
+   1. Return the results of `get_record_data`, with `standardized_popularity`
+      appended.
+1. Update `ProviderDataIngester#process_batch` to call
+   `get_popularity_enriched_record_data` instead of `get_record_data` _if
+   popularity constants exist for this provider_. The logic can be skipped for
+   providers that do not support popularity metrics.
+1. Update the Media stores to accept `standardized_popularity`, defaulted to
+   `None`, and write to tsv.
+1. Update the `columns` used in the loader sql to write this column to the
+   database.
+
+When this work is completed:
+
+- Standardized popularity scores will be calculated at ingestion for all new
+  records
+- For non-dated DAGs, popularity scores will immediately be backfilled for
+  historical records as soon as the DAG has a successful run.
+- For dated DAGs, the existing reingestion DAGs will begin slowly backfilling
+  standardized popularity scores for historical records.
+
+[^1]:
+    There is one exception: iNaturalist does not use `get_record_data`, instead
+    doing its data processing in postgres. We do not need to do anything to
+    address this now because we do not track popularity data for iNaturalist;
+    _however_, if we want to add popularity data for it in the future we will
+    need to handle it separately.
+
+### Create a `popularity_refresh` DAG
+
+Once popularity calculation happens at ingestion (and assuming the media tables
+are backfilled -- more on that later), we need only handle the following cases:
+
+- A new popularity metric is added (for a new or existing provider)
+- The popularity constants are recalculated
+
+In both cases, we need to refresh the `<media>_popularity_constants` view (in
+order to recalculate constants). Note that once the constants have been updated,
+_the provider DAGs will immediately pick up the new constant and begin using it
+at ingestion_. All that remains is to update the scores for historical data.
+
+**Note**: It is true that our reingestion DAGs will also immediately begin
+updating the standardized scores of historical data using the new constant.
+However, this process is slow and some reingestion DAGs are currently running on
+sparse schedules. I think it is still worth it to explicitly updates the scores
+in this DAG, especially because that will allow us to use the DAG to do the
+initial backfill.
+
+As with our data refresh DAGs, we will use a factory to generate a
+`popularity_refresh` DAG for each media type. The DAG will do the following:
+
+1. Update the `<media>_popularity_metrics` table to include any newly added
+   metrics.
+1. Refresh the `<media>_popularity_constants` view to recalculate the popularity
+   constants.
+   1. This is done `CONCURRENTLY` so that provider DAGs can continue reading
+      from the view while it updates.
+1. For each unique `provider` in the `<media>_popularity_constants` view,
+   generate a `refresh_<provider>_scores` task. The task will run an `UPDATE` of
+   the `standardized_popularity` on all records matching that provider which
+   were last updated before the task began.
+   1. We may consider running the `refresh_<provider>_scores` tasks in parallel
+      to speed up the update.
+   1. Optionally, we can hard code a `SKIPLIST` of providers that are present in
+      the `<media>_popularity_constants` view, but for which we do not want to
+      create a refresh task. We currently have some providers (Nappy, Rawpixel,
+      Stocksnap) that support popularity data but are not dated, meaning scores
+      for **all** of their records will be updated the next time the DAG runs.
+      _Note that some of these DAGs are on a `@monthly` schedule however, which
+      means skipping them in this DAG could result in delayed recalculation
+      time._
+1. Report to Slack when the scores have finished updating.
+
+Some notes:
+
+- The DAG will initially run on a `None` schedule, while we get a sense of how
+  long it will take to complete. Thereafter we will enable it on a `@monthly`
+  schedule if possible, or quarterly if it proves too long-running.
+- This DAG can run completely independently of the data refresh. The data
+  refresh will always have access to the most recently ingested records, with
+  standardized popularity. They will at worst be calculated using slightly
+  outdated constants, which is the case today in production.
+
+#### Special Considerations: Avoiding Deadlocks and Timeouts
+
+The popularity refresh tasks must be written carefully to avoid deadlocks. The
+key issues to understand are:
+
+- Provider DAGs may be running and upserting historical data in the media tables
+  at the same time that popularity refresh tasks are updating _the same data_.
+- The `upsert_data` step of a provider DAG upserts all records in a single
+  transaction, meaning that the changes are not committed until all rows have
+  been upserted.
+
+If the popularity refresh task _also_ updates all records for a provider in a
+single transaction, a deadlock scenario could happen. Imagine that the records
+'A' and 'B' already exist on the `image` table for Flickr, and the Flickr
+reingestion DAG is reingesting them at the same time the `refresh_flickr_scores`
+task is refreshing their popularity scores. Concurrent updates might happen in
+this order:
+
+| Flickr Reingestion DAG | `refresh_flickr_scores` | Comment                                                                                                                                                               |
+| ---------------------- | ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Upsert record A        | Update record B         |                                                                                                                                                                       |
+| Upsert record B        |                         | `refresh_flickr_scores` has a lock on the row containing record B, so the Flickr Reingestion DAG must wait.                                                           |
+|                        | Update record A         | The Flickr Reingestion DAG has a lock on the row containing record A, so `refresh_flickr_scores` must wait. The transactions are now waiting on each other. Deadlock! |
+
+We can avoid deadlock by noting that writes from a provider DAG will necessarily
+_always have more up-to-date data_: so when a popularity refresh task encounters
+a locked row, we can tell it to simply skip that record using `SKIP LOCKED`
+(example code below).
+
+We have now avoided deadlock but we still have a problem with our example: the
+Flickr reingestion DAG must still wait for `refresh_flickr_scores` to release
+the lock on record B. Our upsert task will hang and likely time out, as the
+popularity refresh task will take a considerably long time to complete.
+
+To avoid this, we must break up the popularity refresh into smaller
+transactions, committing after every 10,000 rows are updated (this number can be
+tweaked as necessary). This ensures that a concurrently running provider DAG
+will only ever have to wait for a second or two before locked rows are released.
+
+The approach is easiest explained with some sample code:
+
+```pgsql
+-- The final product will be generalized by provider and media type.
+-- In a local test, this successfully updated ~1 million Flickr
+-- records in ~6 seconds, while the reingestion DAG was running.
+DO $$
+DECLARE
+  rows_per_loop int := 10000;
+  min_row int := 1;
+  max_row int;
+  max_timestamp timestamp;
+BEGIN
+  -- We will use this to select rows that have not been recently updated.
+  SELECT NOW() INTO max_timestamp;
+
+  -- A temporary table storing the identifiers of all records that need
+  -- to be updated, plus their row number. The row number will be used to
+  -- paginate the results more efficiently than using OFFSET.
+  CREATE TEMP TABLE rows_to_update AS
+    SELECT ROW_NUMBER() over() row_id, identifier
+    FROM image
+    WHERE provider='flickr' AND updated_on < max_timestamp;
+  CREATE INDEX ON rows_to_update(row_id);
+
+  SELECT COUNT(*) INTO max_row FROM rows_to_update;
+
+  -- Loop over `rows_to_update` 10k rows at a time, updating scores.
+  -- Avoid deadlock using SKIP LOCKED
+  FOR i IN min_row..max_row BY rows_per_loop LOOP
+   UPDATE image
+   SET standardized_popularity = standardized_image_popularity(
+       image.provider,
+       image.meta_data
+   )
+   WHERE identifier in (
+       SELECT identifier FROM rows_to_update
+       WHERE row_id >= i AND row_id < i+rows_per_loop
+       FOR UPDATE SKIP LOCKED
+   );
+   -- COMMIT this batch
+   COMMIT;
+END LOOP;
+END;
+$$;
+```
+
+This is much less efficient than doing a single `UPDATE`, but has clear
+advantages:
+
+- It allows provider DAGs to continue running unimpeded.
+- The data refresh can run at any time, even while the refresh tasks are
+  running.
+- Because updates are committed incrementally, each batch of updated scores will
+  be 'live' on the image table as soon as they are committed. Rather than
+  waiting for all rows to be calculated before making the update, as much
+  up-to-date data as possible will be present each time the data refresh runs.
+
+### Remove popularity steps from the data_refresh DAGs
+
+We can simply remove the popularity tasks from the relevant data refresh DAG
+factories. See the `Parallelizable Streams` section for more information on the
+timing of this change.
+
+### Run the data refresh from the media tables
+
+Once the `image` and `audio` tables contain up-to-date standardized popularity
+data, we can update the data refresh to read from these tables rather than their
+respective materialized views. Because the schema of the media tables now
+matches what the ingestion server already expects, we can simply change the
+table name each place it is referenced:
+
+- `_get_shared_cols`
+- `refresh_api_table`
+- `get_copy_data_query`
+
+## Dependencies
+
+### Tools & packages
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+We should not need any additional dependencies.
+
+### Other projects or work
+
+<!-- Note any projects this plan is dependent on. -->
+
+This does not depend on any existing projects.
+
+### Infrastructure
+
+<!-- Note any infrastructure this plan is dependent on. -->
+
+This project directly affects our Catalog database schema, Airflow DAGs, and
+ingestion server. All necessary changes are described in this document.
+
+Notably, because this plan would allow us to drop the materialized `image_view`
+and `audio_view`, we should net some DB storage savings in the long term.
+
+## Alternatives
+
+### Duplicate table
+
+Another approach was considered which, rather than updating records directly on
+the `image` table, would have instead created a duplicate table with updated
+scores and ultimately swapped tables. This has the benefit of avoiding
+concurrent writes to the `image` table, but the following disadvantages:
+
+- While the duplicate table is being built, writes to the `image` table from
+  provider DAGs would need to be propagated to it via a `Trigger`. Thus the same
+  deadlocking precautions are necessary, and the writes must still be batched,
+  offering no improved performance.
+- This approach requires much more space for the duplicate table (although
+  roughly equivalent to the current materialized view approach).
+- This approach is much less efficient because it unnecessarily re-inserts
+  **all** rows, including those that do not support popularity data at all.
+
+### Separate `image_popularity` view
+
+We considered keeping the `image` table as is, and an `image_popularity`
+materialized view equivalent to the current `image_view`. A new,
+non-materialized `image_view` would join `image` and `image_popularity` and be
+used for the data refresh. The advantage is that the `image_popularity` view
+could be refreshed independently of the data refresh; this approach would also
+be very quick and easy to implement. The disadvantages:
+
+- Would require an expensive `join` at the time of data refresh
+- The `image_popularity` view still needs to be refreshed, and will still take
+  an untenably long time.
+- When the data refresh runs, records that have not yet been refreshed in the
+  `image_popularity` view will make it to the API, but without any standardized
+  popularity score.
+
+## Accessibility
+
+<!-- Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+
+As part of this project, we should add documentation for the new popularity
+refresh DAGs and the calculations themselves.
+
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+The order of operations for this project is very important.
+
+1. Add `standardized_popularity` column to media tables.
+2. Update the Media stores to accept `standardized_popularity`, defaulted to
+   None, and write this column to tsv.
+3. Update the provider DAGs to calculate `standardized_popularity` at ingestion,
+   and write to the database.
+4. Create the `popularity_refresh` DAG factory.
+5. Enable the `popularity_refresh` DAGs in production, to backfill scores on the
+   media tables.
+6. Once the `popularity_refresh` DAGs have completed their initial run, update
+   the ingestion server to begin copying data from the media tables rather than
+   the matviews.
+7. Remove the popularity steps from the data refresh DAGs.
+8. Drop the materialized views.
+
+Steps 3 and 4 may be done in parallel.
+
+## Rollback
+
+<!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
+
+These changes will only go into effect when we update the ingestion server to
+begin reading from the media tables. We can rollback by reverting just this last
+change. Consequently we should wait to drop the materialized views until we are
+confident that the process is working.
+
+Because we have a far smaller number of `audio` records, we should test the
+process with `audio` first. We should also test the data refresh in staging.
+
+## Risks
+
+<!-- What risks are we taking with this solution? Are there risks that once taken can’t be undone?-->
+
+This approach would add an additional column to our media tables; it also
+requires a temp table used for storing the ids of rows that need to be updated.
+There will be a net decrease in storage needed when the materialized views are
+finally dropped, but in the interim we will use extra space.
+
+We do not have a clear estimate for how long the popularity refresh DAG will
+take, and a complete run is necessary before we can begin running the data
+refresh. If the popularity refresh takes an extremely long time, this will delay
+our ability to restart the image data refresh. _After_ the initial run, we will
+be able to data refresh without waiting for subsequent popularity refreshes --
+but an extremely long popularity refresh time is still undesirable because it
+could cause our popularity scores to become relatively outdated.
+
+It's also possible that we may see some increase in ingestion time for providers
+with popularity data, due to:
+
+- the addition of the calculations themselves. This should be insignificant as
+  the calculation runs in constant time, and the popularity constant/metric need
+  only be fetched once.
+- concurrent writes from the popularity refresh tasks. The batched approach will
+  prevent provider DAGs from hanging indefinitely while waiting on the
+  popularity refresh, but there may still be some competition for locks which
+  could add several seconds of waiting time. This is also unlikely to be
+  significant, especially as the `upsert_data` steps are not currently a
+  bottleneck in our provider DAGs.
+
+## Prior art
+
+<!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
+
+- [Investigation into data refresh failures due to popularity calculation timeouts post-iNaturalist](https://make.wordpress.org/openverse/2023/02/21/post-inaturalist-data-refresh-status/)
--- a/documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
+++ b/documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
@@ -27,11 +27,12 @@ independently of popularity calculations.
 
 ## Background
 
-To improve search relevancy, we boost records in our search results relative to
-a normalized popularity score. Because the process of calculating these scores
-and pushing them into the API DB (referred to as "data refresh") is complex, it
-may be beneficial to briefly outline the tables and views currently used in the
-catalog database.
+To improve search relevancy, we use a
+[rank feature query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-rank-feature-query.html)
+to calculate relevance scores based (in part) on a `standardized_popularity`
+score. Because the process of calculating these scores and pushing them into the
+API DB (referred to as "data refresh") is complex, it may be beneficial to
+briefly outline the tables and views currently used in the catalog database.
 
 For simplicity, I will use `image` as an example, but note that equivalent
 structures exist for `audio` and will exist for any media types added in the
@@ -98,8 +99,8 @@ data refresh is:
 
 1. Add `standardized_popularity` as a `nullable` column on the `image` and
    `audio` tables. ([link](#add-standardized-popularity-to-the-media-tables))
-1. Update the `provider_dag_factory` to calculate `standardized_popularity` **at
-   ingestion**.
+1. Update the `provider_dag_factory` and `ProviderDataIngester` to calculate
+   `standardized_popularity` **at ingestion**.
    ([link](#Update-provider-DAGs-to-calculate-standardized-popularity))
 1. Create a new `popularity_refresh` DAG
    ([link](#create-a-popularity_refresh-dag)) which:
@@ -293,7 +294,8 @@ BEGIN
 
   -- A temporary table storing the identifiers of all records that need
   -- to be updated, plus their row number. The row number will be used to
-  -- paginate the results more efficiently than using OFFSET.
+  -- paginate the results more efficiently than using OFFSET. The table
+  -- identifier must be unique for each provider to prevent conflict.
   CREATE TEMP TABLE rows_to_update AS
     SELECT ROW_NUMBER() over() row_id, identifier
     FROM image
@@ -363,7 +365,11 @@ We should not need any additional dependencies.
 
 <!-- Note any projects this plan is dependent on. -->
 
-This does not depend on any existing projects.
+This does not _depend_ on any existing projects. However, we should be careful
+to follow along with the
+[Documenting Media Properties](https://github.com/WordPress/openverse/issues/412)
+project which is also in progress, and make sure that schema changes are
+documented in accordance with that work.
 
 ### Infrastructure
 
@@ -403,12 +409,31 @@ used for the data refresh. The advantage is that the `image_popularity` view
 could be refreshed independently of the data refresh; this approach would also
 be very quick and easy to implement. The disadvantages:
 
-- Would require an expensive `join` at the time of data refresh
+- Would require a `join` at the time of data refresh
 - The `image_popularity` view still needs to be refreshed, and will still take
   an untenably long time.
 - When the data refresh runs, records that have not yet been refreshed in the
   `image_popularity` view will make it to the API, but without any standardized
   popularity score.
+
+### Separate `image_popularity` table
+
+Rather than having `popularity_score` as a column on the `image` table, we could
+have a separate `image_popularity` table which contains only the scores, and
+`identifier` for the purpose of joins. Initial popularity would still be
+calculated at ingestion, and written to the new table. The data refresh would be
+run on a (non-materialized) view which joins the two tables. This prevents the
+media tables from becoming larger, and may possibly be faster to update
+popularity. However:
+
+- It is more complex to implement, because updates now have to be made to two
+  tables at ingestion
+- It still requires the same batching logic to avoid deadlocks/timeouts, so it
+  is unlikely to be significantly faster at updating popularity scores
+- It may result in a longer running data refresh because of the join
+- It may result in longer loading times for provider DAGs
+- It will use more space as long as there is a high proportion of records with
+  popularity data, because we need to store two columns for each
 
 ## Accessibility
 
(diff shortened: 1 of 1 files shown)


stacimc:
I've extended the deadline until May 2nd for my own availability, but want to re-emphasize that I'm happy to push it back further for participants! I'll check in closer to the deadline.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+   );
+   -- COMMIT this batch
+   COMMIT;
+END LOOP;
+END;
+$$;
+```
+
+This is much less efficient than doing a single `UPDATE`, but has clear
+advantages:
```
AetherUnbound, line 325: Very thorough thinking here!!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+Once the `image` and `audio` tables contain up-to-date standardized popularity
+data, we can update the data refresh to read from these tables rather than their
+respective materialized views. Because the schema of the media tables now
+matches what the ingestion server already expects, we can simply change the
+table name each place it is referenced:
+
+- `_get_shared_cols`
+- `refresh_api_table`
+- `get_copy_data_query`
```
AetherUnbound, line 351: Do you mind linking static references to these functions here?

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Duplicate table
+
+Another approach was considered which, rather than updating records directly on
+the `image` table, would have instead created a duplicate table with updated
+scores and ultimately swapped tables. This has the benefit of avoiding
+concurrent writes to the `image` table, but the following disadvantages:
+
+- While the duplicate table is being built, writes to the `image` table from
+  provider DAGs would need to be propagated to it via a `Trigger`. Thus the same
```
AetherUnbound, line 387: Is this a Postgres `Trigger` or a `Trigger<x>` Operator in Airflow? (Do you mind linking to the documentation here as well?)

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+used for the data refresh. The advantage is that the `image_popularity` view
+could be refreshed independently of the data refresh; this approach would also
+be very quick and easy to implement. The disadvantages:
+
+- Would require an expensive `join` at the time of data refresh
+- The `image_popularity` view still needs to be refreshed, and will still take
+  an untenably long time.
+- When the data refresh runs, records that have not yet been refreshed in the
+  `image_popularity` view will make it to the API, but without any standardized
+  popularity score.
```
AetherUnbound, line 409: Thank you for including this!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+  max_row int;
+  max_timestamp timestamp;
+BEGIN
+  -- We will use this to select rows that have not been recently updated.
+  SELECT NOW() INTO max_timestamp;
+
+  -- A temporary table storing the identifiers of all records that need
+  -- to be updated, plus their row number. The row number will be used to
+  -- paginate the results more efficiently than using OFFSET.
+  CREATE TEMP TABLE rows_to_update AS
```
AetherUnbound, line 296: Since multiple providers might be updated at once, does this table identifier need to be unique for each provider?

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+This approach would add an additional column to our media tables; it also
+requires a temp table used for storing the ids of rows that need to be updated.
+There will be a net decrease in storage needed when the materialized views are
+finally dropped, but in the interim we will use extra space.
+
+We do not have a clear estimate for how long the popularity refresh DAG will
+take, and a complete run is necessary before we can begin running the data
+refresh. If the popularity refresh takes an extremely long time, this will delay
+our ability to restart the image data refresh. _After_ the initial run, we will
```
AetherUnbound, line 464: I made a note about this above, but if we potentially performed a one-time copy of existing data from the matview into the new `standardized_popularity` columns on the primary tables, we might be able to unblock a few steps without affecting production data!

AetherUnbound (commented):
Sharing my initial thoughts - I think this is an _excellent_ plan. I love the idea of calculating the scores at ingestion time using the most up-to-date constants we have; that strikes a great balance between having data available in production and still having the capacity to update it down the line. I had a few more comments while I was reading the IP but they ended up being answered later in the doc itself 😁

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+This document describes a method for separating the popularity calculation steps
+from the data refresh process, allowing the data refresh to be re-enabled
+independently of popularity calculations.
+
+## Background
+
+To improve search relevancy, we boost records in our search results relative to
+a normalized popularity score. Because the process of calculating these scores
+and pushing them into the API DB (referred to as "data refresh") is complex, it
+may be beneficial to briefly outline the tables and views currently used.
```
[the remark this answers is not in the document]
stacimc: Great point!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+## Overview
+
+<!-- A brief one or two sentence overview of the implementation being described. -->
+
+This document describes a method for separating the popularity calculation steps
+from the data refresh process, allowing the data refresh to be re-enabled
+independently of popularity calculations.
+
+## Background
```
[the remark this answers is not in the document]
stacimc: I agree! I think when we cut tickets we should definitely include a few documentation ones. This will probably need to be cleaned up a bit to include some of the details I left out for the sake of brevity in the IP.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+```pgsql
+ALTER TABLE image ADD COLUMN standardized_popularity DOUBLE PRECISION;
+ALTER TABLE audio ADD COLUMN standardized_popularity DOUBLE PRECISION;
+```
+
+The column can (and indeed _must_) be nullable, as not all providers support
+popularity data. Because this configuration exactly matches the schema of the
+current `image_view` and `audio_view` matviews, we will be able to begin running
+data refreshes directly off these tables with very few changes in the ingestion
+server.
```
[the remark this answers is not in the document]
stacimc: Is it possible to do that without running into the concurrency/deadlocking problems with simultaneous updates coming from the provider DAGs?

I suppose one thing we could do is do the update one provider at a time, and temporarily pause the provider DAG while the update runs (for this initial run).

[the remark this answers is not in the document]
stacimc: We should definitely wait at least until this IP is fully approved before starting something like that, just to make sure we're going with this approach -- but that's a really interesting idea if we can make it work. It would allow us to enable the data refreshes more quickly, _and_ let us do a full production trial of the refresh-from-media-tables before investing the time in the rest of the work 😮

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: I was thinking of a [Postgres Trigger](https://www.postgresql.org/docs/current/plpgsql-trigger.html), but it's a really good point that needs to be clarified. I'll add the link to the doc here as well!

FWIW I actually got as far as prototyping this approach before realizing we could do it without fully duplicating the table, so I even have the trigger I _would_ have proposed:

```pgsql
CREATE OR REPLACE FUNCTION refresh_copy()
  RETURNS TRIGGER
  LANGUAGE plpgsql
AS $$
  BEGIN
    IF (TG_OP = 'DELETE') THEN
        DELETE FROM image_copy WHERE OLD.identifier;
     ELSE
        DELETE FROM image_copy where identifier = NEW.identifier;
        INSERT INTO image_copy
          SELECT *
          FROM image
          WHERE identifier = NEW.identifier;
     END IF;
     RETURN NULL;
  END;
$$;

CREATE CONSTRAINT TRIGGER refresh_copy_trigger AFTER INSERT OR DELETE OR UPDATE ON image FOR EACH ROW EXECUTE PROCEDURE refresh_copy();
```

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Good point, I'll make sure to specify that!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+It should be noted that conventionally, storing calculated fields would be
+considered a violation of database normalization principles. However, doing so
+in this case will give us much better performance of the data refresh and also
+save a significant amount of space, as it will enable us to drop the
+materialized views.
+
+### Update provider DAGs to calculate standardized popularity
+
+Using the `ProviderDataIngester`, we can easily add standardized popularity
+calculation to all of our providers at once [^1].
```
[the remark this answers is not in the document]
stacimc: Always super pleased when we find a way to use the base class for something 😄

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+- insert newly ingested images
+- update previously ingested images (for example, by updating raw popularity
+  data)
+
+#### `image_popularity_metrics`
+
+A table which contains known 'metrics': that is, the name of the meta*data field
+which contains raw popularity data for each provider. \_Not all providers have a
+configured popularity metric.*
```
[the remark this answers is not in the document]
stacimc: Oh thank you, no idea what happened there 😮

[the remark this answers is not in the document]
stacimc: Weird, this seems to be something strange that the linter is adding. I'll take a closer look...

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+<!-- A brief one or two sentence overview of the implementation being described. -->
+
+This document describes a method for separating the popularity calculation steps
+from the data refresh process, allowing the data refresh to be re-enabled
+independently of popularity calculations.
+
+## Background
+
+To improve search relevancy, we boost records in our search results relative to
+a normalized popularity score. Because the process of calculating these scores
```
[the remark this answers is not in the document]
sarayourfriend: Doesn't this have a different meaning? "relative to" seems precise in that individual document popularity scores are "relative" to the normalised. It's not clear what "using" would mean here. Using _how_, I suppose would be the clarification I would need, but "relative" already implies it, assuming a general understanding of statistics (weighted/relative scores are common enough, I think?).

Not trying to be picky, I just want to make sure that if the wording here changes that "using" is disambiguated because it isn't obvious on its own in a way that I do think "relative" to is, at least somewhat.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+For simplicity, I will use `image` as an example, but note that equivalent
+structures exist for `audio` and will exist for any media types added in the
+future. Descriptions are not exhaustive; some details that are not relevant have
+been omitted for brevity.
+
+#### `image`
+
+The 'main' image table. Rows in this table may contain **raw** popularity data
+in the `meta_data` column, such as the number of times this image has been
+viewed or downloaded.
```
sarayourfriend, line 44: Maybe worth disambiguating here (or elsewhere) that this popularity data is sourced in the provider as opposed to Openverse-specific popularity data we could derive in the future since adding analytics to the frontend (an explicit goal of adding those).

The project proposal also does not disambiguate this.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+To improve search relevancy, we boost records in our search results relative to
+a normalized popularity score. Because the process of calculating these scores
+and pushing them into the API DB (referred to as "data refresh") is complex, it
+may be beneficial to briefly outline the tables and views currently used.
+
+For simplicity, I will use `image` as an example, but note that equivalent
+structures exist for `audio` and will exist for any media types added in the
+future. Descriptions are not exhaustive; some details that are not relevant have
+been omitted for brevity.
```
sarayourfriend, line 38: Can you add a note that these tables are in the catalogue database rather than the API database? The names being 1:1 the same makes it ambiguous.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+   1. We may consider running the `refresh_<provider>_scores` tasks in parallel
+      to speed up the update.
+   1. Optionally, we can hard code a `SKIPLIST` of providers that are present in
+      the `<media>_popularity_constants` view, but for which we do not want to
+      create a refresh task. We currently have some providers (Nappy, Rawpixel,
+      Stocksnap) that support popularity data but are not dated, meaning scores
+      for **all** of their records will be updated the next time the DAG runs.
+      _Note that some of these DAGs are on a `@monthly` schedule however, which
+      means skipping them in this DAG could result in delayed recalculation
+      time._
```
sarayourfriend, line 225: Exciting use case for [dynamic task mapping](https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html) :tada:

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+
+We have now avoided deadlock but we still have a problem with our example: the
+Flickr reingestion DAG must still wait for `refresh_flickr_scores` to release
+the lock on record B. Our upsert task will hang and likely time out, as the
+popularity refresh task will take a considerably long time to complete.
+
+To avoid this, we must break up the popularity refresh into smaller
+transactions, committing after every 10,000 rows are updated (this number can be
+tweaked as necessary). This ensures that a concurrently running provider DAG
+will only ever have to wait for a second or two before locked rows are released.
```
sarayourfriend, line 275: Nice :rocket:

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,486 @@
...
+### Separate `image_popularity` view
+
+We considered keeping the `image` table as is, and an `image_popularity`
+materialized view equivalent to the current `image_view`. A new,
+non-materialized `image_view` would join `image` and `image_popularity` and be
+used for the data refresh. The advantage is that the `image_popularity` view
+could be refreshed independently of the data refresh; this approach would also
+be very quick and easy to implement. The disadvantages:
+
+- Would require an expensive `join` at the time of data refresh
```
sarayourfriend, line 404: What is expensive about the join? Wouldn't the identifiers be indexed?

documentation/projects/proposals/popularity_optimizations/index.md
```diff
@@ -0,0 +1,5 @@
+# Popularity Optimizations
+
+```{include} /_toc_templates/flat.md
+
+```
```
sarayourfriend, line 5: Note: this may need updating if #1868 is merged before this PR.

sarayourfriend (commented):
So cool! I reviewed because I'm excited about this work and curious to learn more about our data infrastructure and potential improvements going forward. To clarify, however, I do not plan to participate in the decision round, so no need to wait for me. FWIW this plan makes a lot of sense to me.

stacimc:
Something I realized that I'll need to add: we'll want to update the `recreate_<media>_popularity_calculation` DAGs to remove the steps for dropping and recreating the materialized view. The rest of the DAG is still useful for dropping and recreating the popularity functions, metrics, and constants tables.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Madison also [pointed this out](https://github.com/WordPress/openverse/pull/1894#discussion_r1178274871) and I totally agree -- actually, I suppose that's an editorial change so I'll update it now without waiting for the round to end.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Totally!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: It was a tough balance to decide how much of the existing structure to document while aiming for brevity, and a lot is left out here if it is not affected or changed by the proposal, including almost everything about the scores are actually computed. We should add much more extensive documentation of the calculation as part of this project, though.

I think since I used an example of raw popularity data here it's definitely worth disambiguating that I mean `the number of times this image has been viewed or downloaded at the source provider`, _not_ views on Openverse. Good call.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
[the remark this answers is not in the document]
stacimc: That's what I was aiming for with the wording, but I do also see how it could be confusing, especially without getting too deep into explaining the configuration. I tried looking up how the Elasticsearch docs phrase this. Maybe the easiest thing to do is to borrow their phrasing and link directly to the elasticsearch docs, something like:

```To improve search relevancy, we use a [rank feature query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-rank-feature-query.html) to calculate relevance scores based (in part) on a `standardized_popularity` score.```

File: documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md (hunk 17 above)
Comment (stacimc) [reply documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md:406]: I should probably remove the word 'expensive' since it does imply there's something _particularly_ expensive about this join, which I don't think is true. It is more that there is a preference to avoid any `join`s on the image table when possible.

File: documentation/projects/proposals/popularity_optimizations/index.md (hunk 18 above)
Comment (stacimc) [reply documentation/projects/proposals/popularity_optimizations/index.md:5]: Good to know, thank you!

File: documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md [hunk 19]
```diff
@@ -0,0 +1,488 @@
...
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+We should not need any additional dependencies.
+
+### Other projects or work
+
+<!-- Note any projects this plan is dependent on. -->
+
+This does not depend on any existing projects.
```
stacimc, line 366: Thinking about this more, it might be a good idea to link to the project for [Documenting Media Properties](https://github.com/WordPress/openverse/issues/412) since we'll want to make sure that we document changes (like the addition of the `standardized_popularity` column) in accordance with that work.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,488 @@
...
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+The basic strategy proposed here for decoupling popularity calculations from the
+data refresh is:
+
+1. Add `standardized_popularity` as a `nullable` column on the `image` and
+   `audio` tables. ([link](#add-standardized-popularity-to-the-media-tables))
+1. Update the `provider_dag_factory` to calculate `standardized_popularity` **at
```
rwidom, line 101: I'm curious about the reference to `provider_dag_factory` here, but not `ProviderDataIngester`. Is that just because the metadata is standardized enough across providers that it'll be DRYer to do it in the dag factory instead?

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,488 @@
...
+### Separate `image_popularity` view
+
+We considered keeping the `image` table as is, and an `image_popularity`
+materialized view equivalent to the current `image_view`. A new,
+non-materialized `image_view` would join `image` and `image_popularity` and be
+used for the data refresh. The advantage is that the `image_popularity` view
+could be refreshed independently of the data refresh; this approach would also
+be very quick and easy to implement. The disadvantages:
+
+- Would require an expensive `join` at the time of data refresh
```
AetherUnbound, line 406: I was thinking on this more, and I actually wonder if it might be possible to combine the proposed at-ingestion-calculation approach with having a separate table for popularity data. We could have a fully separate table for `image_popularity` (_not_ a matview, but full table) that gets updated on ingestion and also updated by the `popularity_refresh` DAG. This has the advantage of preventing our image table from growing wider, and the join can happen by way of a non-materialized `image_view` which joins the two. That join would only be done when the table is queried, i.e. during the data refresh. It might also mean that updates to the table by both the ingestion and the popularity refresh could be quicker, since the table is only two columns wide. And we'd only be storing popularity data for records we have, rather than having a null column for all our data without popularity info. On the flip side, it'd bring some additional complexity at ingestion because we'd need a way to write to both tables (and I believe the `identifier` which is used as our ID for a record is generated on insert into the primary media table) 🤔 I'm also not sure it gets us out of the concurrent write issue you've described. What do you think? Are there other drawbacks or benefits that I might have missed in thinking about this?

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Good point, it should probably read ```Update the `provider_dag_factory` AND the `ProviderDataIngester`...```! The actual calculation will happen in the `ProviderDataIngester`, but there are also updates needed to the dag factory to pass the necessary constants into the ingester class. There's more details in the implementation section.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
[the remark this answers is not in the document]
sarayourfriend: That's far more precise :100:

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: > It might also mean that updates to the table by both the ingestion and the popularity refresh could be quicker, since the table is only two columns wide.

😮 Is an _update_ of a single column necessarily faster if the table it's in is smaller? Wow!

>  And we'd only be storing popularity data for records we have, rather than having a null column for all our data without popularity info.

This definitely strikes me as an advantage, although it would be a theoretically diminishing one if we're able to add popularity data to a larger proportion of our providers.

On the other hand, each row would also have to duplicate the `identifier` in order to join the tables, so if the proportion of records that do have pop data increases, we might end up using _more_ space. The proportion of records with pop data is actually already quite high (I believe >50%), because all the Flickr records have it.

> On the flip side, it'd bring some additional complexity at ingestion because we'd need a way to write to both tables (and I believe the identifier which is used as our ID for a record is generated on insert into the primary media table) 🤔 I'm also not sure it gets us out of the concurrent write issue you've described.

I'll have to double check but I'm actually fairly certain that `identifier` is generated and present in the tsvs, so I _think_ we have it before insert in the primary media table. We would probably still need to add a separate task for writing the scores to a separate table though, which would add a little bit of complexity if we still want all the values in the same tsv.

I don't think it prevents the concurrent write issues, since we'd still have multiple DAGs writing to the same table. Thinking off the cuff, if the popularity writes are in a separate task in the provider DAG, we could add a sensor that skips that task if the `refresh_<provider>_popularity` task is running. But I don't think that will work, because the provider DAG might be adding _new_ records. So unfortunately I don't think the popularity refresh tasks would be any more performant, unless the individual `update`s run faster on smaller tables.

> Are there other drawbacks or benefits that I might have missed in thinking about this?

I think it would be interesting to benchmark copying directly from the image table vs copying from a non-materialized view with the `join`. Surely there must be some performance impact, but it might be totally insignificant. In either case I don't see how it could be _faster._

I think my gut feeling is that this would not be easier to implement (and might add extra complexity), and likely wouldn't improve the performance of the popularity refresh, but _might_ make the data refresh a little slower. I'm really intrigued by the idea that the smaller table might make popularity updates a little bit faster, but the speed of the popularity _refresh_ is theoretically the least important aspect of the system.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
obulat: I was thinking we could make the SQL file parser configurable to allow for parsing `view` if it's available, and only use `image` if not - of course if Documenting Media Properties is implemented before this project :)

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
obulat: > I'll have to double check but I'm actually fairly certain that identifier is generated and present in the tsvs, so I think we have it before insert in the primary media table. We would probably still need to add a separate task for writing the scores to a separate table though, which would add a little bit of complexity if we still want all the values in the same tsv.

Unfortunately we don't have `identifier` before inserting the media from the loading table into the main table:
https://github.com/WordPress/openverse/blob/59a4411e4363207627e54467f7cf6ad40b1a4b0b/catalog/dags/common/storage/db_columns.py#L10-L11

>😮 Is an update of a single column necessarily faster if the table it's in is smaller? Wow!
When doing updates, Postgres writes an entirely new row, unless it is a [HOT update](https://www.cybertec-postgresql.com/en/hot-updates-in-postgresql-for-better-performance/). I learned this when trying to update some value in the upstream database and it would take more than 40 days :)

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,488 @@
...
+
+We can simply remove the popularity tasks from the relevant data refresh DAG
+factories. See the `Parallelizable Streams` section for more information on the
+timing of this change.
+
+### Run the data refresh from the media tables
+
+Once the `image` and `audio` tables contain up-to-date standardized popularity
+data, we can update the data refresh to read from these tables rather than their
+respective materialized views. Because the schema of the media tables now
```
obulat, line 346: Now we use the materialized views for selecting data during the data refresh.
Would it have any effect on the ingestion or data refresh performance if we use `image` tables both for refreshing data and for ongoing writing of the data from provider API scripts? My guess is no, considering that we do use the `image` table for creating the `materialized view` anyways.

obulat (commented):
I was really wondering what the IP would have, and love the suggested approach!

I had only one clarification request about the implication of using image table for both selecting data for refresh and for writing the provider API script data.

Madison's suggestion about creating a separate table for popularity data is interesting, but I don't know now the lack of `identifier` could be handled.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Oof, thanks for double checking that Olga. That would be an additional complication, then :/ We could possibly change where the `identifier`s are generated so we _do_ have them in the tsvs. We'd definitely want to avoid extra queries.

obulat: If we create identifiers before writing to TSVs, we will have inconsistencies with the database. If the identifier is created during ingestion and written to TSV, then during the data refresh, for items that we already have in the database, we would replace the identifier from the TSV with the identifier from the database :/

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,488 @@
...
+popularity data. Because this configuration exactly matches the schema of the
+current `image_view` and `audio_view` matviews, we will be able to begin running
+data refreshes directly off these tables with very few changes in the ingestion
+server.
+
+It should be noted that conventionally, storing calculated fields would be
+considered a violation of database normalization principles. However, doing so
+in this case will give us much better performance of the data refresh and also
+save a significant amount of space, as it will enable us to drop the
+materialized views.
```
rwidom, line 136: I started writing a whole philosophy of non-dogmatic appreciation for normalization, kimball, and one big table approaches to data modelling. But at the end of the day, we're just trying to provide as timely and accurate data as possible, as fast as we can, right? So to me, where we store it is less important than when we calculate it, and there is lots on that below. :) 

I also looked into whether another data type might be better in terms of table width for records without popularity data, but at the end of the day, I feel like 8 bytes won't break the bank. It's more the jsons and super long text fields that need some consideration.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,488 @@
...
+- The DAG will initially run on a `None` schedule, while we get a sense of how
+  long it will take to complete. Thereafter we will enable it on a `@monthly`
+  schedule if possible, or quarterly if it proves too long-running.
+- This DAG can run completely independently of the data refresh. The data
+  refresh will always have access to the most recently ingested records, with
+  standardized popularity. They will at worst be calculated using slightly
+  outdated constants, which is the case today in production.
+
+#### Special Considerations: Avoiding Deadlocks and Timeouts
+
```
rwidom, line 240: I wonder if it would be possible to handle conflicts at a higher level. Just thinking out loud... would it be possible to separate out old records from dated DAGs with popularity data, from the new/current run only data? Or is it that the whole point of dated DAGs is to still do a full refresh, just using dates to break it into manageable chunks? The smaller commit strategy is very cool, but the switching costs could be significant, so I guess I'm wondering if there is a way to limit the scope of Step 5 below.

rwidom (commented):
This is really awesome, and I hope that I'm on point enough here to ask some useful questions!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
sarayourfriend: Is it possible to use Python's UUID module to generate the identifier and only have Postgres generate one when one isn't provided in the TSV? Apologies if this misses an obvious piece of the puzzle of how ingested records get written. Basically though my question is: can the UUID be generated in Python when the record is retrieved? Is there an issue with needing to confirm that records that are being ingested aren't duplicates of existing records, in which case they'd already have a UUID (queryable via the URL)? Would a Redis cache of URL -> UUID for existing records help with that?

obulat: Currently, we collect all the data from the provider using the providerDataIngester, and write it all into a TSV. Then we load the TSV into a loading table and clean it up (remove duplicates on foreign_id and provider pair, and records with null for required fields). Then we upsert these rows into the main `image` or `audio` table. During the upsert, if there's already an item with the same `url`, we update it. Otherwise, we create a insert a new row and create a new UUID then.

We could do something like `SELECT identifier FROM image where url='image_url'` during ingestion, but I have no idea how that would affect the speed or performance. Also, if we create a UUID during the ingestion, we will have to make sure that there is no collision of UUIDs created by different provider scripts running at the same time.

sarayourfriend: > Also, if we create a UUID during the ingestion, we will have to make sure that there is no collision of UUIDs created by different provider scripts running at the same time.

This is theoretically guaranteed to be a non-issue with UUIDs right? [Wikipedia claims that you need to generate 2.71 quintillion UUIDs before the chance of collision is 50%](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions). Granted we'd need to make sure collisions don't happen, but it's sure to be unlikely. Even with our several hundred million records, the chances are miniscule. From the article linked above:

> the probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion.

I'd venture to say that the chances are so small we should feel comfortable deferring even considering how to solve that to a later project and date.

> We could do something like SELECT identifier FROM image where url='image_url' during ingestion, but I have no idea how that would affect the speed or performance.

That's along the lines of what I was thinking should be possible. [If `url` is indexed](https://github.com/WordPress/openverse/blob/697f62f01a32cb7fcf2f4a7627650a113cba40da/catalog/dags/common/loader/sql.py#L105) it should be very fast, right?

obulat: Theoretically I agree with you, @sarayourfriend. But I'd still like to hear what others think, because I don't have enough experience with it.

stacimc: @sarayourfriend's proposal is what I had in mind, yep. I think that would make it possible to generate the identifiers in advance, so we could keep everything in one tsv and eliminate the need for additional queries. In the incredibly unlikely event of a collision the provider DAG would fail on the load step, which can be easily retried, so there is no risk as far as I can see.

That being said, this would still add complexity, and I'm not sure I see any equivalent _advantage_ to this approach over the proposed one. Arguably storing the calculated popularity scores in another table still violates normalization principles. I also think it would end up taking up more space and potentially make the data refresh take longer.

I'd like to move this IP into the revision stage, but I want to give @AetherUnbound especially a chance to respond to the thread here.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: I went down some similar lines of thought while working on this, but unfortunately I didn't see a way. We want to accommodate both the insertion of new records and the updating of old records by provider DAGs. Those updates of old records are crucial in part because they include updates to raw popularity scores, so we need this information ASAP. Moreover, it's not just dated DAGs that have popularity data -- some non-dated DAGs, which reingest all records on each run, have popularity scores. 

Since we don't want to prevent the provider DAGs from inserting/updating any of their data, I don't see any way to prevent them from getting stuck awaiting the popularity refresh without batching it (although I'm very open to ideas!). There is definitely a significant cost to the batching, but since we're tolerant to the popularity refresh being long-running as long as it does not block our other processes, I think this is acceptable.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
AetherUnbound: Apologies for the delay in getting back to this - I've been mulling this over and I don't have a strong argument for why it should be in a separate table. My _gut_ says that having it in a separate table is "the right way" to do it and would have a positive impact on various optimizations, but I can't put much more behind that than a feeling of not wanting our primary tables to get larger. All that to say, I don't have a good reason for this being a blocking objection, so we can disregard it and stick with the plan as proposed 🙂 No other blockers on moving into the objection stage on my end!

stacimc:
I'm moving the IP into the Revision Round and will be drafting the PR while I make updates based on feedback.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
rwidom: Hmmm... Sorry if this is too late, please feel free to ignore. But I guess my suggestion was about limiting the records included in the popularity refresh so that there isn't a conflict. For providers with non-dated DAGs, every ingestion is a full refresh, right? So their popularity stuff would be handled at ingestion and (I think?) there wouldn't be a need for a popularity refresh at all. (Unless they run so rarely that the ingestion couldn't take care of it, but in that case, a conflict would be pretty unlikely, right?) I guess maybe I was assuming that it would be possible in the dated dags, to divide records into two groups -- those where we need to update (or insert) _both_ popularity calculations _and_ other data (handled by the regular provider dags), and those where we _only_ want to update the popularity calculation (handled by the refresh). And that maybe some combination of the last update field, DAG type / schedules / statuses from airflow, could create a useful filter for the popularity refresh that would allow it to complete more quickly while avoiding records likely to be handled by the regular dag.

rwidom: Inserts would definitely be handled by the regular provider dag, and wouldn't lock the whole table, so they don't seem like a concern here. Right? And I still think doing one provider at a time would be useful, maybe building a temp table and [merging](https://www.postgresql.org/docs/current/sql-merge.html) it in could be more performant for the popularity only updates.

rwidom: OK, yeah, I just started to look in airflow, and made [this super basic spreadsheet](https://docs.google.com/spreadsheets/d/1yzmFbs2L4yiu3yCqzWNNIPb_uUadryetMaNo_67mTOg/edit#gid=864570427). So, I think this is where we are, right?

- Most providers don't have popularity data, so it's a non-issue for them.
- Of those that do, four of them (Nappy, Jamendo, Rawpixel, and Stocksnap) are relatively small/fast, non-dated, _monthly_ dags. So it would be pretty straightforward to just run the refresh for each of those providers individually by building a provider temp table with just the UUID and popularity calculation using a select from the main table, and then doing a merge back into the main table for the update. And to avoid conflicts, we'd just need to pick a day when the regular dag wasn't running.
- That leaves flickr and wikimedia. And they're tricky, because they have daily, dated, dags, that processed <50,000 records in <15 minutes the last time they ran, but I guess we don't know how many of those were inserts and how many were updates. The inserts don't (or shouldn't) matter for refresh purposes. But let's say that there were 1,000 updates in each run, and that we have no way of knowing where those 1,000 will fall in the (guessing here) 300,000,000 records we have from each of those providers. So, could we instead, do a similar process as for the smaller providers described above, but further divide these providers by `updated_on`, taking a month at a time or something? If that won't work because of the recency of implementing dated dags (i.e. the vast majority have the date of  last run prior to dated dags), then maybe we'd use `created_on` in addition to get a reasonably sized temp table. Start at the beginning and work forward. We could pause the dag when the regular daily run is active, but we wouldn't have to if there was a filter in the merge statement to say something like "if this has been updated since we started the refresh, skip it". That's assuming that the refresh starts after the ongoing provider runs are already calculating popularity, which I think was the order, right?

rwidom: If `updated_on` + `created_on`, still leaves too big of a partition, maybe there is a further partition using the first n characters of the uuid or something?

stacimc: Hm, I'm not sure I'm following. I think I generally agree but I'm not sure what changes you mean to the proposed solution 🤔 

**Non-dated DAGs**
I pointed out [here](https://github.com/WordPress/openverse/pull/1894/files#diff-9d0dc40138273796366c2301a730b7da43513f37364fdc77bc2835b7aff4aa14R220) that non-dated DAGs will get their popularity scores updated the next time their DAG runs, as you mentioned. I suggested having a `SKIPLIST` for those providers and letting them refresh ‘naturally’, as one option. I’m not convinced that we should, though: my reasoning for not doing so is that the popularity refresh could feasibly complete just moments after the provider DAG, and then popularity data for that provider will be ‘stale’ for an entire month. If we really want to avoid doing popularity refreshes for them we could get away with it but I don’t see any real benefit to not? Particularly if the providers refreshes happen in parallel.

> So it would be pretty straightforward to just run the refresh for each of those providers individually by building a provider temp table with just the UUID and popularity calculation using a select from the main table, and then doing a merge back into the main table for the update. And to avoid conflicts, we'd just need to pick a day when the regular dag wasn't running.

If we’re for sure running it when the regular DAG is not running I don’t think you need a temp table at all, you could just do the update. But I also would not _rely_ on offset default schedules to guarantee that these DAGs don’t run at the same time, especially since we need to make sure this works for any new providers that are added in the future, which may have schedules and runtimes we haven’t considered. We could add in sensors to make sure the provider DAG isn't running, but that's additional complexity -- and I'd prefer for the provider DAGs to _never_ be blocked on popularity refreshes.

Using the batched update is slow, but incremental updates are committed ASAP (which is not the case if we wait for the regular DAG run to naturally refresh data) and we’re guaranteed to have no conflicts, no matter what the DAG’s schedule is, or if a backfill is running or a DAG is being retried on an atypical day/some other unforeseen problem. Plus, we don't have to maintain a separate process for updating dated vs non-dated DAGs.

**Dated DAGs**
I pointed out [here](https://github.com/WordPress/openverse/pull/1894/files#diff-9d0dc40138273796366c2301a730b7da43513f37364fdc77bc2835b7aff4aa14R198) that our existing reingestion workflows do, in theory, eventually refresh all records for our dated DAGs, so if we really wanted to we could also rely on these to naturally refresh popularity data. Unfortunately at the moment:
* these are in many cases flaky
* Flickr’s is actually disabled due to rate limiting concerns
* many are on sparser schedules than we would like (running only `weekly` instead of `daily`) due to the flakiness, which means they are not reingesting enough data to reliably and fully refresh all records

Reingestion DAGs are always, 100% of the time doing updates rather than inserts that risk conflict with the popularity refresh. We don’t want to turn off or interfere with the reignestion workflows, because they update more than just popularity scores. At the moment you could maybe schedule around them because they are on weekly schedules, but we intend to change them to daily schedules, which will make that impossible. Regular dated workflows do a mix of inserts and updates and as you point out, we don’t have any way of guessing which rows are going to be updated.

>  If that won't work because of the recency of implementing dated dags (i.e. the vast majority have the date of last run prior to dated dags), then maybe we'd use created_on in addition to get a reasonably sized temp table. Start at the beginning and work forward. We could pause the dag when the regular daily run is active, but we wouldn't have to if there was a filter in the merge statement to say something like "if this has been updated since we started the refresh, skip it".

This sounds really similar to me to the proposed solution. “If this has been updated since we started the refresh, skip it” is essentially what I’m accomplishing by using `SKIP LOCKED`. The batching is to prevent the provider DAG from hanging/waiting on the refresh, and I still do not see any way around that at all. I think you're saying theh same thing when you talk about partitions, is that right? Or do you see a way to avoid batches altogether?

When I build the temp table that is used to generate batches, I already [only attempt to update records](https://github.com/WordPress/openverse/pull/1894/files#diff-9d0dc40138273796366c2301a730b7da43513f37364fdc77bc2835b7aff4aa14R302) that were last updated before the constants were recalculated. It is true that some of those records will end up being updated by the reingestion workflows in the background, after that temp table is built, so the `last_updated` filter isn't perfect.

One nice optimization would be to add one more check of the `last_updated` date to the update for each batch. Ie:

```pgsql
  -- Loop over `rows_to_update` 10k rows at a time, updating scores.
  -- Avoid deadlock using SKIP LOCKED
  FOR i IN min_row..max_row BY rows_per_loop LOOP
   UPDATE image
   SET standardized_popularity = standardized_image_popularity(
       image.provider,
       image.meta_data
   )
   WHERE identifier in (
       SELECT identifier FROM rows_to_update
       WHERE 
            row_id >= i
            AND row_id < i+rows_per_loop
            AND updated_on < max_timestamp
       FOR UPDATE SKIP LOCKED
   );
   -- COMMIT this batch
   COMMIT;
```

Is that all that you're suggesting? I'm sorry if I'm totally off mark 😅

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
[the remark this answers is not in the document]
AetherUnbound: Oh good call, agreed on all fronts!

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,513 @@
+# Implementation Plan: Decoupling Popularity Calculations from the Data Refresh
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @AetherUnbound
```
AetherUnbound, line 10:
```suggestion
- [x] @AetherUnbound
```

AetherUnbound (approved):
Woohoo, I'm excited about this! The revisions look great - we could add a note about the process we mentioned for updating the existing image table with the existing popularity data in the materialized view (by provider). That's not a blocker though, awesome work! 🚀

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
rwidom: Here's what I think is different, and I haven't tested it, but I guess I'm thinking about doing the calculation in the table build, so it could be something more like this:

```
create temp_table as (
    select 
        identifier, 
        now() as cut_off_timestamp, 
        standardized_image_popularity( image.provider, image.meta_data ) as standardized_popularity
    from image
    where provider='flickr' and [some constants calculated in a loop in python to divide the table into ~100,000 record chunks, using updated_on, created_on, and maybe a substring of identifier]
);
commit;
merge into image using temp_table on (identifier)
when matched and (image.updated_on < temp_table.cut_off_timestamp)
set image.standardized_popularity = temp_table.standardized_popularity
;
commit;
```

Doing it this way would avoid the issue of updating records that had more recently been updated (by whatever job, with the date filter), might be faster than using row_id (at least that's the lesson learned from improving iNaturalist performance, it would depend on indexing on the image table, but I don't think it would be worse than this), and hopefully makes the update freeze itself a lot faster by moving the calculation out of the update step and into the table build. And because the chunks are meaningfully defined / sorted, there is less risk of traffic jams for new records, right?

stacimc:
> we could add a note about the process we mentioned for updating the existing image table with the existing popularity data in the materialized view (by provider)

@AetherUnbound I asked [here](https://github.com/WordPress/openverse/pull/1894#discussion_r1178415765), but wouldn't trying to copy existing popularity data into the image table at the onset run into the same deadlocking/timeout issues I described in the IP? Even if we do it one provider at a time, this would require disabling the providers while the initial load happens, which might still be pretty time consuming, at least for Flickr -- although I admit I don't have a good estimate of how long.

https://github.com/WordPress/openverse/pull/2006 should unblock us from running data refreshes in the interim though! So fortunately we're no longer under pressure to get this working immediately.

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
stacimc: Thanks for writing out the queries, that really helps! Sorry to be long-winded, but I’m going to re-state my understanding of things to make sure I understand, and you can correct me where I’m wrong please 😄 

So a problem with the approach lined out in the IP is that we build the whole temp table of `records_to_update` at the beginning and then split it into batches. By the time we get around to processing any given batch, many of the records in that batch might already have been updated by another DAG, so it is a waste of time to update it again.

I [tried to address this](https://github.com/WordPress/openverse/pull/1894#discussion_r1187900145) by adding an additional filter to the `update` query for each batch, to filter out any records that have since been updated. So while each batch will have 10k candidate records in it, many of them might get filtered out and it will process potentially much faster than expected.

The alternative you’ve proposed, as I understand it (please correct me), is to instead first programmatically identify a series of constants that can be used in a query to divide the records into rough batches. The simplest version of this would be a series of timestamps `[a, b, c, … n]` such that there are about 10k records whose `updated_on` date is between each interval `(a, b)`, `(b, c)`, and so on. Then for each of these intervals, we create a temp table for just that batch, and we calculate the popularity score while building this table. This is then merged into the `image` table, rather than doing an update.

So a comparison/contrast:
* preventing updates to records that have been more recently updated by a provider DAG:
    * Both ideas use a filter on some `cut_off_timestamp` in the update/merge query to prevent unnecessary updates
* length of time the table is locked
    * Since your idea does the calculation during the table build, the `image` table will not be locked for as long. But since we’re forced to process in batches anyway, I don’t think we need to be particularly concerned: provider DAGs will never have to wait on it long
* “because the chunks are meaningfully defined / sorted, there is less risk of traffic jams for new records”
    * This part I’m not sure I understand. In both cases we try to avoid updating records with a recent `updated_on` date, which is really the only factor in collisions. I still think we need to use `SKIP LOCKED`, but that should work fine in either case.

You mentioned that it might be faster though than using `row_id`, which I’m interested in investigating. It would require a lot more work and complexity to actually create the batches (if I’m understanding what you meant about using “constants calculated in a loop in python to divide the table”), right? I planned on using `row_id` because I thought that was supposed to be much faster than sorting and selecting with very large `OFFSET`s, which is the first thing I thought of for calculating those constants — but I could be totally wrong, or else maybe you had a different idea for how to calculate them?

I don't have a good sense of how we would reliably make those partitions. It _sounds_ like it would be much more complex and potentially tricky (again, please correct me if this assumption is wrong!), so I'm a little hesitant to do so if we're not absolutely certain that it will be _significantly_ faster, especially since the popularity refresh is a process that we actually don't mind being very slow. We may find that the `row_id` approach takes an extremely long time (>month), in which case it would definitely make sense to prioritize further optimizations.

AetherUnbound:
BTW @stacimc it looks like there's a warning that's being encountered when building the docs that will need to be resolved before we merge 😮

documentation/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.md
```diff
@@ -0,0 +1,513 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [x] @AetherUnbound
+- [ ] @obulat
```
obulat, line 11:
```suggestion
- [x] @obulat
```

obulat (approved):
This is a very thorough plan. Thank you for so much documentation both on the tables that we have, and prevention of deadlocks in Postgres, it's really useful and interesting to learn :)

PR #1897, opened 2023-04-25
Add checks for required parameters to Provider scripts (S-Z)

obulat:
## Fixes

Related to #1883 by @obulat

## Description

This PR was originally created to replace the None checks for required values with `if <value>` checks because falsy values such as empty string or empty list should also short-circuit the provider API scripts.

After @AetherUnbound's suggestion to add tests, I split the PR into 3 to make the changes easier to review. This PR contains changes for the last 5 providers (from "S").

I wasn't aware about the recent addition of an Exception when one of the required param is missing. After this, we have to make sure that the provider API scripts check for truthiness of the required parameter values before returning the data in `get_record_data` function.

## Testing Instructions

The tests should pass, and the additional tests should test the changes.
Check that all of the possible "falsy" values are tested in the provider tests.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/licenses/licenses.py
+++ b/catalog/dags/common/licenses/licenses.py
@@ -36,12 +36,6 @@ special_cases = {
     "https://creativecommons.org/share-your-work/public-domain/pdm/": "https://creativecommons.org/publicdomain/mark/1.0/",
 }
 
-# SMK sometimes uses incorrect URL for licenses
-special_cases = {
-    "https://creativecommons.org/share-your-work/public-domain/cc0/": "https://creativecommons.org/publicdomain/zero/1.0/",
-    "https://creativecommons.org/share-your-work/public-domain/pdm/": "https://creativecommons.org/publicdomain/mark/1.0/",
-}
-
 
 @lru_cache(maxsize=1024)
 def get_license_info(
--- a/catalog/dags/common/licenses/licenses.py
+++ b/catalog/dags/common/licenses/licenses.py
@@ -36,6 +36,12 @@ special_cases = {
     "https://creativecommons.org/share-your-work/public-domain/pdm/": "https://creativecommons.org/publicdomain/mark/1.0/",
 }
 
+# SMK sometimes uses incorrect URL for licenses
+special_cases = {
+    "https://creativecommons.org/share-your-work/public-domain/cc0/": "https://creativecommons.org/publicdomain/zero/1.0/",
+    "https://creativecommons.org/share-your-work/public-domain/pdm/": "https://creativecommons.org/publicdomain/mark/1.0/",
+}
+
 
 @lru_cache(maxsize=1024)
 def get_license_info(
--- a/catalog/dags/providers/provider_api_scripts/smk.py
+++ b/catalog/dags/providers/provider_api_scripts/smk.py
@@ -49,14 +49,13 @@ class SmkDataIngester(ProviderDataIngester):
     @staticmethod
     def _get_foreign_landing_url(item) -> str | None:
         """Use the English site instead of the original link."""
-        object_num = item.get("object_number")
-        if not object_num:
+        if not (object_num := item.get("object_number")):
             logger.info(
                 f"Image with (foreign) id {item.get('id')} does not have "
                 "`object_number`! Therefore we cannot build the "
                 "foreign_landing_url."
             )
-            return
+            return None
         # Occasionally, the object number will have a space in it. for these cases we
         # need to urlencode it.
         return f"https://open.smk.dk/en/artwork/image/{urllib.parse.quote(object_num)}"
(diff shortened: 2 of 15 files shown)


catalog/dags/common/licenses/licenses.py
```diff
@@ -25,6 +25,12 @@ class InvalidLicenseURLException(Exception):
 
 LicenseInfo = namedtuple("LicenseInfo", ["license", "version", "url", "raw_url"])
 
+# SMK sometimes uses incorrect URL for licenses
```
obulat, line 28: Added in here because the example SMK used in test uses this ("share-your-work") URL, and the tests fail  because of this. Also added in the `LicenseURL` PR because that's where this change actually belongs. 

I do think that we should treat the works with this URL as marked with CC0 or PDM. If others think we shouldn't, I'll update the test object to use a valid CC license URL.

catalog/dags/providers/provider_api_scripts/smk.py
```diff
@@ -87,41 +86,39 @@ def _get_creator(item: dict) -> str | None:
...
+
+        thumbnail_url = item.get("image_thumbnail")
+        height = item.get("image_height")
+        width = item.get("image_width")
+        filesize = item.get("image_size") or item.get("size")
 
         # We used to get additional images from the `alternative_images` field,
         # but we found these to be either redundant, duplicates, or lower quality
         # versions. See https://github.com/WordPress/openverse-catalog/issues/875
         # for the full investigation & discussion
```
stacimc, line 110: The method signature here confused me at first, but this comment explains it; this used to build a list including related images. Now that we only ever return one image, I think we should get rid of this method and pull this logic into `get_record_data`.

This isn't related to this refactor so we can do that in another PR, but it stood out to me so worth mentioning!

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
```diff
@@ -107,6 +107,56 @@ def test_get_query_param_offset_page_number():
...
+@pytest.mark.parametrize(
+    "record",
+    [
+        # missing foreign_landing_url
+        {"links": {}},
+        {"links": {"self": ""}},
+        {"links": {"self": "link"}, "attributes": {"multimedia": []}},
+    ],
+)
+def test_get_record_data_returns_none_for_falsy_fid_and_landing_url(record):
```
stacimc, line 119: I didn't immediately understand why these cases are pulled out from the test case below. It's because when the foreign landing url or id are missing, we return `None` right away, but otherwise we attempt to build and return an image list -- which will be empty if the records were all invalid (missing image_url, etc). Could you add a comment to that effect? Or alternatively, I suppose they could all be in one test with `assert not actual_record_data`.

stacimc, line 119:
```suggestion
def test_get_record_data_returns_none_for_falsy_landing_url_and_multimedia(record):
```

Falsy fid is not tested in this case, right?

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
```diff
@@ -107,6 +107,56 @@ def test_get_query_param_offset_page_number():
...
+        # missing foreign_identifier
+        {"links": {"self": "link"}, "attributes": {"multimedia": [{"admin": {}}]}},
+        {
+            "links": {"self": "link"},
+            "attributes": {"multimedia": [{"admin": {"uid": ""}}]},
+        },
+        {
+            "links": {"self": "link"},
+            "attributes": {"multimedia": [{"admin": {"uid": "fid"}}]},
+        },
```
stacimc, line 136: This test case is an example of missing image_url, right, not missing foreign_identifier?

catalog/tests/dags/providers/provider_api_scripts/test_smk.py
```diff
@@ -132,3 +141,27 @@ def test_get_record_data_returns_main_image():
...
+@pytest.mark.parametrize(
+    "property_name",
+    [
+        pytest.param("object_number", id="falsy foreign_landing_url"),
+        pytest.param("rights", id="falsy license_info"),
+    ],
+)
+def test_get_record_data_returns_none_with_falsy_required_property(property_name):
+    item = _get_resource_json("item.json")
+    if property_name == "id":
```
stacimc, line 155: What is this check doing?

catalog/tests/dags/providers/provider_api_scripts/test_wordpress.py
```diff
@@ -61,11 +61,11 @@ def test_get_should_continue_checks_total_pages(
 
 
 @pytest.mark.parametrize("missing_field", ["slug", "link"])
-def test_get_record_data_returns_none_when_missing_necessary_data(
+def test_get_record_data_returns_none_when_necessary_data_is_falsy(
     ingester, missing_field
 ):
     image_data = _get_resource_json("full_item.json")
-    image_data.pop(missing_field, None)
+    image_data[missing_field] = ""
```
stacimc, line 68: Could we add a check for when the field is missing, in addition to when it is falsy?

stacimc (commented):
Thanks so much for all of this, @obulat :) I appreciate the split into smaller PRs as well, it has been very helpful for reviewing! I left some comments on test coverage but no real blocking concerns other than adding an `image_url` check for SMK.

catalog/dags/providers/provider_api_scripts/smk.py
```diff
@@ -87,41 +86,39 @@ def _get_creator(item: dict) -> str | None:
...
-            )
+
+        if not (foreign_identifier := iiif_id or item.get("id")):
+            return []
+
+        if iiif_id:
+            image_url = SmkDataIngester._get_image_url(iiif_id)
+        else:
+            # Legacy images do not have IIIF links.
+            image_url = item.get("image_native")
```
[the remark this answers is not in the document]
obulat: Definitely! Thank you for catching :)

catalog/tests/dags/providers/provider_api_scripts/test_smk.py
obulat: It's left out from copy-pasting. I removed it, and improved tests

stacimc (approved):
Looks great! Thanks for cleaning up SMK as well :)

catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
```diff
@@ -249,6 +249,33 @@ def test_extract_title_gets_cleaned_title(wmc):
     assert actual_title == expected_title
 
 
+def test_get_record_data_returns_none_when_missing_foreign_identifier(wmc):
+    media_data = _get_resource_json("image_data_example.json")
+    media_data["pageid"] = ""
+
+    record_data = wmc.get_record_data(media_data)
+
+    assert record_data is None
```
krysal, line 258: Can't this be included as a parameter in the test below?

krysal (approved):
LGTM! I left a few comments but not consider those blocking. Great job with the None <> Falsy check changes for all the DAGs! 👏

catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
obulat: It's separate because the level of nesting for the parameters is different: `pageid` is the top-level property, and others are nested in the first item of `imageinfo`. I could combine them, but that would complicate the test functions.

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
```diff
@@ -107,6 +107,63 @@ def test_get_query_param_offset_page_number():
...
+        {"links": {}},
+        {"links": {"self": ""}},
+        {"links": {"self": "link"}, "attributes": {"multimedia": []}},
+    ],
+)
+def test_get_record_data_returns_none_for_falsy_foreign_landing_url_and_multimedia(
+    record,
+):
+    actual_record_data = sm.get_record_data(record)
+    assert actual_record_data is None
```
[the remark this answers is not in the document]
obulat: It's more difficult to parametrize these tests because of the different nesting levels: we would have to somehow get the correct nested property and set it to null. I've used the order in which we extract the properties here, so that `foreign_landing_url` is the first thing that is checked in the provider script, and the test will return `None` right away. For other tests, `foreign_landing_url` is always available because they first check for it before checking other parameters.

We could create a separate issue to improve the tests here, what do you think?

PR #1903, opened 2023-04-26
Query the filtered indexes

### LINKED ISSUE #1198: Query the filtered index from the API
## Problem

After https://github.com/WordPress/openverse/issues/1196 is complete, we can now query the filtered index from the API.

This is **not** blocked by #1197.

## Description

Please see the implementation plan for technical implementation details: https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html#query-the-filtered-index-overview-step-3

Steps to complete:

1. Add the new unstable query parameter to the search serializer and wire up the `mature` deprecation.
2. Add a new Django setting `ENABLE_FILTERED_INDEX_QUERIES` and default it to `False`
3. Query the filtered index as described in the [implementation plan section for this step](https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html#query-the-filtered-index-overview-step-3)

sarayourfriend:
## Fixes

Fixes #1198 by @sarayourfriend 

## Description

Add a new setting `ENABLE_FILTERED_INDEX_QUERIES` and a new search parameter `unstable__include_sensitive_results`. Query the filtered index when the setting is enabled and the query parameter is false. Finally, modifies the abstract mature and deleted media classes to accommodate the filtered indexes. This was an unexpected need that was missed during implementation planning.

## Testing Instructions

Checkout this branch and test the default behaviour which does not query the filtered indexes. Run `just api/up` to get the API running.

The following requests verify one of the mock sensitive terms used for the local filtered index creation, "bird". Please reference `ingestion_server/static/mock_sensitive_terms.txt` for other terms to test with.

These requests should include results that reference birds:

- http://localhost:50280/v1/images/?q=perching
- http://localhost:50280/v1/images/?q=bird

Now test this PR with the new setting enabled. Add the new setting to your `api/.env` file:

```
$ echo "ENABLE_FILTERED_INDEX_QUERIES=True" >> api/.env
```

If you already have the API running, you'll need to restart it for the setting to take effect:

```
$ just down
$ just api/up
```

The following requests should now all differ. Requests without the `include_sensitive_results` params should not include any results matching "bird":

- http://localhost:50280/v1/images/?q=perching
- http://localhost:50280/v1/images/?q=perching&unstable__include_sensitive_results=True
- http://localhost:50280/v1/images/?q=bird
    - **These results should be entirely empty**
- http://localhost:50280/v1/images/?q=bird&unstable__include_sensitive_results=True

We must also test that the existing mature filter for documents with `mature=true` in Elasticsearch still works.

From your local Django admin (http://localhost:50280/admin, username and password both are deploy), create an image report with the reason set to mature. I used the identifier `6d6d8c7d-9f86-4bb4-b332-cb039f798fbf` because it is the first result in the `perching` query so is easy to verify. Set the decision to "mature_filtered" so that the result gets update in Elasticsearch. Now run the query again and confirm that the identifier you made the report for does not appear in the filtered results. Confirm that the result does appear when the `include_sensitive_results` parameter is true.

Now do the same with another result but deindex it this time instead of marking it as mature. Confirm that it does not appear in _any_ searches, regardless of whether sensitive results are included.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Looks like I need to update the unit tests to accommodate the filtered indexes. @stacimc @krysal What's the best way to do that, do y'all think? On the one hand, we expect this to become the default behaviour, so it would make sense to update the tests to assume this is the default behaviour. In that case, I'd like to duplicate some subset of those tests to also run with the setting turned off, to ensure both still work.

On the other hand, if we assume that this might get turned off or need to be reverted, then making those changes would also "lock in" the tests and any revert would not only need to remove that assumption from the tests but also backport any subsequent changes made to those tests. In that case, I could see an argument for keeping the tests as is but running them with the filtered index setting turned _off_ or by passing the new query parameter. Running them with the setting turned off would require making the default `env.template` entry for the new setting False (or removing it) or devising a way to have separate defaults for testing. Making the setting off by default in the env.template seems OK-ish, but it would also be good to dog-food this feature locally if we can. To clarify, the setting is and will stay off by default in the settings.py configuration so that it won't be turned on in staging until the new filtered indexes actually exist (once #1833 is merged and staging ingestion server is updated and the DAG is run against staging [does the DAG need additional provision to be able to run against staging? :open_mouth:]).

sarayourfriend:
Blocking this PR to wait for advice on how to update the tests. Either approach would take a non-trivial amount of work so I'd rather do neither until I have confirmation from at least one other person of which makes the most sense to them.

stacimc:
Interesting question 🤔 

> Making the setting off by default in the env.template seems OK-ish, but it would also be good to dog-food this feature locally if we can.

Agreed with this. I'd rather keep it True in the env.template. I think my preference is your first suggestion -- I see what you mean about a potentially tangled revert if the feature gets fully reverted, but it should be fine to just turn it _off_ as far as I can see, so I don't foresee those changes needing to be made with urgency. I think a little bit of potential churn in the event of a revert is preferable.

sarayourfriend:
Alright, I had to make some additional changes to the tests. First, the model factories were not generating an pk for the instance, meaning that the tests couldn't rely on the `id` or `pk` property.

Second, I hadn't realised that there were pre-existing tests for the media reports. I've combined these tests and made the minimal changes to the existing ones so that they use the model factory and account for the filtered index. The tests still approach mocking ES differently, but I chose not to make further changes to the tests to reduce the load on reviewers.

Finally, I found different queries to use for the quoted search tests for images and audio that avoid the filtered index altogether and therefore do not need to be changed regardless of whether the feature is rolled back eventually.

api/catalog/api/serializers/media_serializers.py
```diff
@@ -243,6 +252,19 @@ def validate_unstable__sort_dir(self, value):
...
+        value,
+    ):
+        exclusive_fields = ("mature", "unstable__include_sensitive_results")
+        if all(f in self.initial_data for f in exclusive_fields):
+            raise serializers.ValidationError(
+                "`mature` and `unstable__include_sensitive_results` "
+                "must not both be defined."
+            )
+
+        return self.initial_data.get("mature") or value
```
krysal, line 266: What is doing `mature` at this point if they are exclusive?

api/catalog/api/models/media.py
```diff
@@ -242,7 +247,51 @@ def save(self, *args, **kwargs):
...
+                    index=index,
+                    id=document_id,
+                    refresh=True,
+                    **es_method_args,
+                )
+            except TransportError as e:
+                if e.status_code == 404:
+                    # This is expected for the filtered index, but we should still
+                    # log, just in case.
+                    logger.warning(
```
krysal, line 286: If this is expected should it be at debug level then?
```suggestion
                    logger.debug(
```

krysal (commented):
I'll try this out in more detail tomorrow when I get a chance to read the context, but at first glance it looks quite good!

api/catalog/api/serializers/media_serializers.py
sarayourfriend: `mature` will work as a deprecated alias for `include_sensitive_results` so that API consumers currently relying on the `mature` query will not have their results break but will still get more expansive information about sensitive content generally. This was discussed in the implementation plan and in the project proposal:

https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230330-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html#query-the-filtered-index-overview-step-3

and 

https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#filtering

> Results with sensitive terms in their textual content are only included in results when include_sensitive_results=True. This parameter will supplant the mature=True as a more comprehensively and descriptively named parameter. mature=True should still work, but it should just be an alias for include_sensitive_results and should be marked as deprecated in or altogether excluded from the API documentation.

api/catalog/api/models/media.py
```diff
@@ -242,7 +247,51 @@ def save(self, *args, **kwargs):
             same_reports.update(status=self.status)
 
 
-class AbstractDeletedMedia(OpenLedgerModel):
+class PerformIndexUpdateMixin:
```
stacimc, line 250: Very nice 💯

stacimc (approved):
This looks really great -- tests all worked perfectly for me, and I see the 400 when both `mature` and `unstable__include_sensitive_results` are both defined. Thanks for the excellent documentation :)

The mixin implementation is slick, and I'm glad you were able to find queries that avoid the filtered index for the quoted search tests. Nice!

api/catalog/api/models/media.py
sarayourfriend: It's only expected for the filtered index and only then when the filtered index excluded it. I could refine this a bit to register an issue with Sentry when the 404 occurred on the origin index (documents shouldn't be missing there), but I'd still like to have a visible, historically accessible log (i.e., one we don't need to turn on to have appear in the logs) so we can investigate if documents aren't being excluded.

This code does not run often, so this new warning log isn't going to cause an big or even noticeable increase in the number of logs stored.

obulat (approved):
I spend quite a while trying to understand why the "We must also test that the existing mature filter for documents with mature=true in Elasticsearch still works." step doesn't work for me. I added the report, but the first image result I see is still "Butterfly Insect Photo". Apparently, there are several images with the same title! 🤦 

I tested this locally with your excellent instructions, and it all worked as described.

Should we update the docs to add that `mature` is deprecated in this PR, or later?

sarayourfriend:
> Should we update the docs to add that mature is deprecated in this PR, or later?

I was thinking we would deprecate `mature` and note its deprecation in the docs once we promote `include_sensitive_results`. What do you think?

krysal (approved):
This is awesome! 🌟 Thank you for the detailed instructions for tests. Everything works perfectly. Sorry it took so long to review. I thought I'd need more context since I haven't reviewed the implementation plan but everything was quite direct and understandable 😄

zackkrida:
> "Butterfly Insect Photo". Apparently, there are several images with the same title! 🤦

Small note @obulat, I suspect this would come from StockSnap, where titles are generated from the primary tags of an image.

Issue #448, opened 2023-02-13
Accessibility with Windows screen readers tracking issue for home/search results pages


alexstine:
## Description

Tracking issue for accessibility fixes for Windows screen readers on the home/search results pages.

## Issues

- [x] WordPress/openverse#449
- [x] Home link is not in the navigation landmark or list.
- [x] It would be nice to have a logo with the name of the site or similar site title to keep context for screen reader users. This could be the second element in the banner landmark after skip to main content link. #788 
- [x] #1010 - #1920
- [x] #794 - #1979
- [ ] After submitting a search, I would expect to have a heading 1 with wording such as "Search results for: XXX, 10 results returned". This should be the first item in the main content, above the search form.
- [ ] #790
- [ ] #452
- [x] #793 - #1888
- [ ] The text in quote 2 is not placed in a very discoverable area for users. I would delete it and rely on the heading 1 proposed above.
- [ ] The button with text "Select a content type: Audio" should appear above the "Search" button.
- [x] WordPress/openverse-frontend#2176
- [x] WordPress/openverse-frontend#2179
- [x] WordPress/openverse-frontend#2183
- [x] WordPress/openverse-frontend#2182

### Quote 1

> link    graphic    Impasto four-handled ovoid olla with lid, from S. Marzano sul Sarno
> link    graphic    Stem Cup with Buddhist Emblems: Shufu Ware
> link    graphic    Female Figurine Rattle
> link    graphic    Sake Flask
> link    graphic    Female Holding A Bowl Figurine
> link    graphic    Double-gourd Shaped Bottle
> link    graphic    The Pottery Seller
> link    graphic    Large cup or bowl with slightly
> link    graphic    Pottery Whistle
> link    graphic    Cincinnati Pottery Club Incised Pitcher

### Quote 2

> Over 10,000 results

### Code snippet 1

Problem:

```
[...]

PR #1920, opened 2023-04-27
Add media type context to start of result link titles

### LINKED ISSUE #1010: Add context to links for screen readers to read the type of media the link goes to.
This issue is derived from the fourth point in this issue by @alexstine https://github.com/WordPress/openverse/issues/448

## Problem

When using a screen reader, there isn't sufficient context about the type of media for a given result. Screen readers will read out "link graphic <media title>", but this doesn't give an indication to the user for what type of media the result is.

## Description

Add the media type before the media title to the link title: `{media_type}: {media_title}`.

sarayourfriend:
## Fixes

Fixes #1010 by @sarayourfriend 

## Description

Adds SR accessible context as per the issue. There are two "todos" that I'll ping folks on to clarify what to do with them.

I also made some changes to the playwright configuration to make it easier to use `start` instead of `prod` to quickly re-run the tests inside the container. I can revert these changes if they're not wanted, but it is merely a codification of what was present in the comment I moved and updated.

## Testing Instructions

Confirm CI runs. I changed a test to exercise the accessible title.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VAudioTrack/VAudioTrack.vue
```diff
@@ -470,7 +470,10 @@ export default defineComponent({
     )
     const ariaLabel = computed(() =>
       isComposite.value
-        ? i18n.t("audio-track.aria-label-interactive", {
+        ? // TODO: Confirm with @dhruvk when this _shouldn't_ be the "seekable" variant
+          // `seekable` variant for the string wasn't used before and changing this makes
+          // non-seekable variant unused: can we remove one or the other and just use one?
+          i18n.t("audio-track.aria-label-interactive-seekable", {
```
sarayourfriend, line 476: @dhruvkb Can you confirm what the correct label should be here? I haven't gone digging for when these labels were introduced or if the seekable one _used_ to be used but was accidentally removed at some point. We don't use both, so we can remove the one we don't use if this change doesn't require additional work to choose the correct one.

In what cases is the player interactive but not seekable? At first I thought it might be the box layout (used for all results grid) but that one is seekable when I tested :thinking:

frontend/src/components/VContentLink/VContentLink.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <!-- We 'disable' the link when there are 0 results by removing the href and setting aria-disabled. -->
+  <!-- TODO: Confirm with @obulat and @dhruvkb that the above comment was inaccurate but that aria-disabled should indeed be set below -->
   <VLink
     :href="hasResults ? to : undefined"
+    :aria-disabled="!hasResults"
```
sarayourfriend, line 6: I need to remove this change and update the comment. It seems that this is out of date with some changes that moved the disabling behaviour into VLink itself.

sarayourfriend:
After much debugging and fighting with Microsoft's servers to actually successfully download the Playwright Docker image from their registry and then the browser binary from their distribution endpoint, both of which were impossibly slow on my connection, I finally found the issue.

The audio cells did not have `title` on the elements, just aria-label. This meant that when the tests tried to walk the DOM via keyboard to find the audio element, it never finds it, span forever, and eventually crashed the browser, failing all the tests in that module.

I tried to add `title` to the audio links, but for some reason it was not possible. I'm not sure if it's because of our use of `component` in `VAudioTrack` to render the link, but nothing I did seemed to pass the title through to the anchor element eventually created.

Instead of trying to continue fighting with it, I opted for the navigation code to fall back to aria-label if title is empty. This avoids the issue entirely and still exercises the context that a screen-reader user would have for the link.

obulat:
> After much debugging and fighting with Microsoft's servers to actually successfully download the Playwright Docker image from their registry and then the browser binary from their distribution endpoint, both of which were impossibly slow on my connection, I finally found the issue.
> 
> The audio cells did not have `title` on the elements, just aria-label. This meant that when the tests tried to walk the DOM via keyboard to find the audio element, it never finds it, span forever, and eventually crashed the browser, failing all the tests in that module.
> 
> I tried to add `title` to the audio links, but for some reason it was not possible. I'm not sure if it's because of our use of `component` in `VAudioTrack` to render the link, but nothing I did seemed to pass the title through to the anchor element eventually created.
> 
> Instead of trying to continue fighting with it, I opted for the navigation code to fall back to aria-label if title is empty. This avoids the issue entirely and still exercises the context that a screen-reader user would have for the link.

In Vue 2, if a parent component passes a parameter to the child component, and it is not declared as the child component's prop, it becomes a part of the child component's`attrs` object. There are 2 options for it:

- if the child component is an HTML Element, then all `attrs` are applied as the element's attributes.
- if the child component is a Vue Component, the `id`, `class` and `style` attrs "fall through" to the HTML element that is a child of this Vue Component, but other attrs are discarded.

For `VLink`, `title` is not a prop. So, it becomes an `attr` that is attached to the `VLink`. `VLink`, in turn, is a Vue Component that is a wrapper for `NuxtLink` component (which, in turn, is a wrapper for `VueRouter`). So, the non-prop parameter that you pass to `VLink` (except for `id`, `class` and `style`) are discarded.

Behavior for internal links: 

- `VLink` (parameters: `id`, `class`, `style`, `title`)
  - `NuxtLink` - `id`, `class`, `style` fall through to child HTML element, `title` is discarded
    - `a` - "catches" the fall-through attrs (`id`, `class`, `style`)

Behavior for external links:

- `VLink` (parameters: `id`, `class`, `style`, `title`)
    - `a` -  all parameters passed by the parent (`id`, `class`, `style` and `title`) are applied to the HTML element 

To make sure that all `VLink` parameters are passed to both external and internal links, we should add `v-bind="$attrs"` to `NuxtLink` inside `VLink`.

```
diff --git a/frontend/src/components/VLink.vue b/frontend/src/components/VLink.vue
index f96dcfc1b..237485597 100644
--- a/frontend/src/components/VLink.vue
+++ b/frontend/src/components/VLink.vue
@@ -5,6 +5,7 @@
     :class="{ 'inline-flex flex-row items-center gap-2': showExternalIcon }"
     :to="linkTo"
     v-on="$listeners"
+    v-bind="$attrs"
     @click.native="$emit('click', $event)"

```

frontend/src/components/VSearchResultsGrid/VImageCell.vue
```diff
@@ -130,12 +132,19 @@ export default defineComponent({
       imgWidth.value = element.naturalWidth
     }
 
+    const contextSensitiveTitle = computed(() => {
+      return i18n.t("browse-page.aria.image-title", {
+        title: props.image.title,
```
obulat, line 137: Do you think it would be clearer for screen reader users if we added more info, something like the full attribution sentence (Image: Cat by creator licensed under CC BY licensed, maybe even adding the provider)?

frontend/src/locales/scripts/en.json5
```diff
@@ -595,9 +597,9 @@
     alt: 'Cover art for "{title}" by {creator}',
   },
   "audio-track": {
-    "aria-label": "{title} - Audio Player",
-    "aria-label-interactive": "{title} - audio player - press the spacebar to play and pause a preview of the audio",
-    "aria-label-interactive-seekable": "{title} - audio player - press the spacebar to play and pause a preview of the audio; use the left and right arrow keys to seek through the track.",
+    "aria-label": "Audio: {title}",
+    "aria-label-interactive": "Audio: {title} - interactive player - press the spacebar to play and pause a preview of the audio",
```
obulat, line 601:
```suggestion
    "aria-label-interactive": "Audio: {title} - interactive player - press the space bar to play and pause a preview of the audio",
```
On Voiceover on Mac, the no space variant is read as "spuh-seh-bar".

frontend/src/locales/scripts/en.json5
```diff
@@ -595,9 +597,9 @@
...
+    "aria-label-interactive-seekable": "Audio: {title} - interactive player - press the spacebar to play and pause a preview of the audio; use the left and right arrow keys to seek through the track.",
```
obulat, line 602:
```suggestion
    "aria-label-interactive-seekable": "Audio: {title} - interactive player - press the space bar to play and pause a preview of the audio; use the left and right arrow keys to seek through the track.",
```

obulat (approved):
This is an improvement for the Image results, so I'm approving to unblock it, but I think we need to make the experience for audio and image results more unified.

For images screen reader reads "Link Image Cat", and for audio it reads "Cat audio player, press spacebar to play" - and has no indication that the element is actually a link. I don't know what the best solution here is, though: would adding something like "Press enter to go to the Audio details" to the end of the audio player instructions be enough?

Also, I asked a question in an inline comment: would it improve the screen reader user experience if we add more information about the images? Because currently, they don't get much info about the image beyond its title and media type.

sarayourfriend:
> For images screen reader reads "Link Image Cat", and for audio it reads "Cat audio player, press spacebar to play" - and has no indication that the element is actually a link. I don't know what the best solution here is, though: would adding something like "Press enter to go to the Audio details" to the end of the audio player instructions be enough?

Thanks for testing this. I haven't had access to a screen reader and need to fix that soon; realistically I probably shouldn't have taken on this issue due to that.

I agree audio should read similarly to images. I don't know if our current approach might be annoying though: it's quite wordy. For a page full of audio results, moving through them would be pretty chatty already, with the instructions repeated on each one.

Could we implement something like the [Gutenberg's `a11y` utility](https://github.com/WordPress/gutenberg/tree/trunk/packages/a11y#api) to announce the toast with instructions and only do so once per page load, and then unify the aria label on all media type links?

I'm also not sure how best to approach this but do agree that it is suboptimal. This issue was stemmed from https://github.com/WordPress/openverse/issues/448 which has a host of other already-identified improvements. My inclination is that we should prioritise fixing those issues, especially because the issue describes specific fixes for almost everything listed.

obulat:
> I'm also not sure how best to approach this but do agree that it is suboptimal. This issue was stemmed from https://github.com/WordPress/openverse/issues/448 which has a host of other already-identified improvements. My inclination is that we should prioritise fixing those issues, especially because the issue describes specific fixes for almost everything listed.

I agree with prioritizing those issues. I think this PR adds a good incremental change for the Image results, and could be merged because it's useful as is. We should also open an issue for improving the overall experience for the search results, and exploring the a11y utility you mentioned.

frontend/src/components/VAudioTrack/VAudioTrack.vue
zackkrida: I recall the box layout _not_ being seekable previously. It isn't via mouse, due to the link, therefore I'm not sure if it should be by keyboard 🤔

zackkrida: @dhruvkb pinging you here again for visibility, thank you! I don't believe the box layout should be seekable.

frontend/src/components/VSearchResultsGrid/VImageCell.vue
zackkrida: I support this idea. Perhaps it's better-suited to a follow-up PR though, as I'd want to make sure, and check with a11y folks, that the titles wouldn't be too long and hinder the search browsing experience.

frontend/src/locales/scripts/en.json5
zackkrida: @obulat this might depend on the default system voice—"spacebar" was read correctly for me.

zackkrida (approved):
LGTM. Olga's suggestions to expand the titles (with license info, creator, and so on) seems like a good follow-up issue.

frontend/src/components/VAudioTrack/VAudioTrack.vue
dhruvkb: Yes, the box layout should not be seekable. It was considered bad UX to allow it to be seekable when the seeking area is super tiny and lies inside a link.

zackkrida: Got it, thanks! So currently the box layout _is_ seekable, so as-written this code will work fine. I'll remove the comment and create a new issue to make the box layout not-seekable, at which point the labels should be updated.

zackkrida:
I'm going to merge this and create a follow-up issue for some of the further identified accessibility improvements. I've also created one to fix the box audio layout: #2124

Issue #1959, opened 2023-05-01
Flaky VR test: visual-regression/pages/pages-single-result.spec.ts:32:15 › image rtl single-result page snapshots


obulat:
## Description

~~This test failed on main, as well as in other PRs (https://github.com/WordPress/openverse/actions/runs/4850898382/jobs/8644256809), so maybe it's not flaky but the snapshot is incorrect?~~
The following tests are flaky in  visual-regression/pages/pages-single-result.spec.ts:

`ltr`/`rlt` - `lg`/`2xl`
The problem is in the top margin of the "licenses" section: 

![image](https://user-images.githubusercontent.com/15233243/235461013-d43d7b3a-abd4-40d2-b761-9512b3307a4b.png)

## Additional context

Marking this as critical to fix the Playwright CI as soon as possible.

I've rebased all my open PRs onto `main` to make sure that they use the updated Playwright navigation utilities. A couple of PRs passed the CI:
https://github.com/WordPress/openverse/pull/1889
https://github.com/WordPress/openverse/pull/1158
https://github.com/WordPress/openverse/pull/1888

Several others failed. I'll list all the failed PRs to make sure that we fix all of them.

## Failing tests
https://github.com/WordPress/openverse/actions/runs/4851053078/jobs/8644722405?pr=1004

playwright_1  |     visual-regression/pages/pages-single-result.spec.ts:32:15 › image ltr single-result page snapshots › screen at breakpoint 2xl with width 1536 › from search results 

https://github.com/WordPress/openverse/actions/runs/4851412123/jobs/8645267500?pr=1141

playwright_1  |     visual-regression/pages/pages-single-result.spec.ts:32:15 › image rtl single-result page snapshots › screen at breakpoint 2xl with width 1536 › from search results 
playwright_1  |     visual-regression/pages/pages-single-result.spec.ts:32:15 › image rtl single-result page snapshots › screen at breakpoint lg with width 1024 › from search results 

https://github.com/WordPress/openverse/actions/runs/4850933584/jobs/8644347536?pr=1188
playwright_1  |     visual-regression/pages/pages-single-result.spec.ts:32:15 › image ltr single-result page snapshots › screen at breakpoint 2xl with width 1536 › from search results

obulat:
I'm closing this issue as `critical` because #1960 fixed failing CI. I will, however, open an issue for exploring how we can fix the test without `maxDiffPixelRatio`.

PR #1132, opened 2023-04-04
Update the VFilterButton to use the new variants

obulat:
## Fixes

Related to #1020 by @obulat

## Description

This PR uses the new variants for the VFilterButton. This standardizes the border behavior (no dark ring in the focus ring, no transitions for border color).

The Storybook tests are also updated: better order, added `focused` state test, and the hot-fix for focus-ring (wait) is removed.

~~The `xl` and `2xl` snapshots are updated because now the button has no border, so the width for larger buttons becomes 2px more.~~ The buttons were updated to always have a border, so no snapshot changes were necessary.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #1960, opened 2023-05-01
Correctly pass `maxDiffPixelRatio` to add tolerance for pages-single-result VR test

obulat:
## Fixes

Related to #1959 by @obulat

## Description

In https://github.com/WordPress/openverse/pull/1132, I've added `maxDiffPixelRatio` screenshot parameter set to 0.02 to fix the flakiness of the `pages-single-result` test. However, I passed it incorrectly, so it wasn't actually used. This PR passes it using the last parameter of `expectSnapshot` function.

I've also added this change to all of the failing PRs from #1959 description, and they all pass now.
This does not fully fix the root cause of the VR failure, however, I think test is still useful even if the license icons section is off. I would welcome any solution for the root cause, too.

@sarayourfriend, considering that this seems to fix all of the `pages-single-result` flakiness, do you think we should still `skip` the failing test, or would merging this PR as is be fine?

## Testing Instructions

The CI should pass. However, because the test is flaky, it could pass even though the flakiness is fixed. I will re-run the CI a couple of times to make sure it doesn't fail.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/playwright/utils/breakpoints.ts
+++ b/frontend/test/playwright/utils/breakpoints.ts
@@ -51,17 +51,10 @@ interface Options {
    * @defaultValue true
    */
   uaMocking?: boolean
-  /**
-   * Adjust the error tolerance for a single test to avoid
-   * flakiness.
-   * Do not set to a value higher than 0.02.
-   */
-  maxDiffPixelRatio?: number
 }
 
 const defaultOptions = Object.freeze({
   uaMocking: true,
-  maxDiffPixelRatio: 0,
 })
 
 const makeBreakpointDescribe =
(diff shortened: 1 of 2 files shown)


sarayourfriend (approved):
This LGTM. Let's skip the tests if they fail in the future rather than proactively. Better to confirm that they are still flaky rather than skip them. If you think that we can remove the maxDiffPixelRatio option here, then an issue to explore that would be appropriate, I think.

AetherUnbound (approved):
I think this looks good! Tests are passing, which seems positive 🙂

PR #2023, opened 2023-05-04
Simplify and reorder parameters to `catalog/pgcli` recipe

AetherUnbound:
## Description

This PR adjust the recipe for `catalog/pgcli` to reduce the number of overrides necessary to define when switching between the Airflow and upstream databases. Additionally, I've reordered the parameters (since with `just` you have to override _all_ parameters in order if you want to change one) so that the fewest need to be changed. Our username and passwords are the same for both databases so I just combined the field.

Lastly, the current config actually didn't work because it was using the airflow creds to connect to the upstream database (which it didn't have access to). I can only speak from personal experience, but almost every time I use the `pgcli` command for the catalog it's to access the upstream database, so I changed the default to be that.

Update: Based on feedback below, I've combined the recipes into a `_pgcli` recipe at the root level, which both the API and catalog now call. 

## Testing Instructions

1. On `main`, running `j catalog/pgcli` and then `select * from image limit 10` should raise an error
1. Repeating the above on this branch should return results (if they exist)!
1. Try running `j catalog/pgcli airflow airflow` and run `\d` and various selects to ensure you can see the airflow schema
2. Run `j api/pgcli` and confirm that this correctly connects to the API's database and can be used to query the API as well.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/justfile
+++ b/api/justfile
@@ -79,8 +79,8 @@ stats media="images" host="localhost:50280":
     just _curl-get "{{ media }}/stats/" {{ host }}
 
 # Launch a `pgcli` shell in the web container
-pgcli db_host="db" db_port="5432" db_name="openledger" db_user="deploy" db_password="deploy": up
-    env DC_USER="opener" just ../exec web pgcli postgresql://{{ db_user }}:{{ db_password }}@{{ db_host }}:{{ db_port }}/{{ db_name }}
+pgcli db_user_pass="deploy" db_name="openledger": up
+    env DC_USER="opener" just ../_pgcli web {{ db_user_pass }} {{ db_name }} db
 
 #########################
 # Django administration #
--- a/catalog/justfile
+++ b/catalog/justfile
@@ -85,8 +85,8 @@ ipython: up-deps
         bash -c \'ipython\'
 
 # Launch a `pgcli` shell in the PostgreSQL container
-pgcli db_host="upstream_db" db_port="5432" db_name="openledger" db_user="airflow" db_password="airflow": up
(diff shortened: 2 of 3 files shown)


sarayourfriend (approved):
LGTM.

It may be best to even move this recipe up to the root file considering both databases are shared across multiple apps. `just api/dbshell` will get you the API database.

Something like `just dbshell catalog` and `just dbshell api` would be pretty handy!

dhruvkb (approved):
I was under the impression that the host and port were the parameters most often to change, but I haven't used this recipe enough to know. 
Since the `api` also has a `pgcli` recipe could you copy the same changes over there?

@sarayourfriend's suggestion to have a `pgcli` recipe in the root makes sense, considering they have the same basic structure (and they both ultimately use `../exec` recipe from the root `justfile` anyway). 

But I'd still prefer to retain the recipes `api/pgcli` and `catalog/pgcli` (that call the root `pgcli` with the right defaults) to keep with the `<project>/<recipe>` convention instead of the `<recipe> <project>` convention (that I've seen in the infra repo).

Screenshot 2023-05-04 at 10 49 28 AM

dhruvkb (approved):

AetherUnbound:
@dhruvkb and @sarayourfriend - I've combined recipes where possible, how does this look?

sarayourfriend (approved):
Right on!

Issue #2026, opened 2023-05-04
PR reminder DAG skips urgent PRs where all requested reviewers have left comment-only reviews


sarayourfriend:
## Problem

Because the DAG relies on the "requested reviews" and because a comment-only review will remove the request from that array on the response, we need to probably be better about re-requesting reviews from people once their comments have been addressed.

## Description

I've been trying to think of the best way to handle this. This is what I came up with, but it's open for discussion: if a PR is outside the urgency window, has comment only reviews, but has been updated since the comment only reviews were left, all reviewers that left a comment who were part of the original requested reviewers should be pinged in addition to people who've not yet left a review.

## Additional context

To derive the list of "original requested reviewers" as distinct from the list of people who've left reviews (which could include people who've reviewed ad-hoc, without being requested via the UI) we can look for "review requested" events in the events list for the PR.

I marked this only as medium instead of high because there is a workaround. If PR authors are diligent about re-requesting reviews in the GitHub UI from people who've left comment-only reviews, the DAG will ping those reviewers.

PR #2027, opened 2023-05-04
Calculate PR urgency based on ready for review time

### LINKED ISSUE #1984: PR review reminders skips PRs if they have any update within the urgency period
## Description

If a PR is low urgency and unreviewed for 4 days and then the author pushes a rebase, the PR's "urgency" clock will get reset back to 0.

https://github.com/WordPress/openverse/blob/2710483a50bec736909eac4a73ef189842e07e01/catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py#L80-L86

## Additional context

I think we should change this to ping since the last request for review. If a PR isn't ready for review yet, the author should draft it.

I've had a lot of PRs get skipped over for reminders because I am diligent in keeping PRs rebased and updated while waiting for reviews. The reminders are effectively useless in the current configuration except for PRs that are totally ignored. With our recent discussions about PR review timeliness and having too many PRs open, I think this change will help make the reminders more useful for our current goals.

Marked high because these reminders are meant to help us review PRs in a timely manner and the current approach is not achieving that. If it's too noisy after this change, we can make further refinements, perhaps looking for specific types of updates to index against, rather than the undraft/open date.

sarayourfriend:
## Fixes

Fixes #1984 by @sarayourfriend 

## Description

This turned out to be a good deal more complicated than I anticipated, primarily due to the need to find the date when previously drafted PRs were marked ready for the review. Accounting for this required creating whole new fixtures.

Along the way, I also noticed we had a small issue with our test fixtures that caused an incorrect test! We had the PR `number` and `id` set to the same thing. However, the `id` is GitHub's unique identifier for the PR across all of GitHub. It is used for some specific endpoints while the `number` is used for others. Changing the fixture to create IDs that are not the same as the PR "number" revealed a test that was accidentally using the "id" instead of the "number" to confirm a particular end state.

In testing this, I noticed another edge case that we do not handle in which PRs are skipped for reminders when they probably should not be: https://github.com/WordPress/openverse/issues/2026

That's a sufficiently complex issue that I didn't think it was wise to solve it in this already rather complicated PR.

## Testing Instructions

Check out the new unit tests and confirm they cover the different cases a PR could be in. The "artificial" test case is unfortunate in that the scenario it is testing isn't possible, but it's the easiest way to prove that PR urgency is deferred
to the "ready_for_review" event when it exists rather than the PR's creation date. Another way to test this is to call the `get_urgency_if_urgent` function directly but all the other tests are "integration" style rather than unit style and I didn't want to break that pattern if I could avoid it. However, if reviewers think it would make more sense to avoid creating an artificial scenario to prove the case, that is a viable alternative to testing the same concept.

Test this locally by setting `AIRFLOW_VAR_GITHUB_API_KEY` in you env to a github PAT and `AIRFLOW_VAR_PR_REVIEW_REMINDER_DRY_RUN=true`. The second one is critical if you want to avoid actually sending out pings!

When I ran it locally, I received these pings in the logs of the DAG:

```
[2023-05-04, 05:46:28 UTC] {pr_review_reminders.py:257} INFO - Pinging @zackkrida to review Add media type context to start of result link titles
[2023-05-04, 05:46:28 UTC] {pr_review_reminders.py:257} INFO - Pinging @obulat to review Project Proposal: Additional search views
[2023-05-04, 05:46:28 UTC] {pr_review_reminders.py:257} INFO - Pinging @dhruvkb to review Handle more than one value for a url query parameter
```

All three of these PRs, as of writing, are in the same boat. They were all created beyond the urgency window but have received updates within the window. They would not receive a ping under the current DAG but will correctly receive a ping under this new approach.

We don't have any real PRs at the moment that exercise the events list scenario, so hopefully the unit tests cover those cases sufficiently!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Opened an issue for the flaky DAG doc generator: https://github.com/WordPress/openverse/issues/2056

catalog/tests/dags/maintenance/test_pr_review_reminders.py
```diff
@@ -315,30 +427,104 @@ def test_does_not_ping_if_pr_has_min_required_approvals(github, urgency, base_br
...
     post_reminders("not_set", dry_run=False)
 
     assert past_due_pull["number"] not in github["posted_comments"]
     assert not_due_pull["number"] not in github["posted_comments"]
+    assert old_but_not_due_pull["number"] not in github["posted_comments"]
+
+
+@parametrize_urgency
+@pytest.mark.parametrize(
+    "events",
```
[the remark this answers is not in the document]
sarayourfriend: We can't because the test case is only relevant for PRs that have a ready for review event. `possible_pingable_events` includes the one case where there wouldn't be a ready for review event on a PR that could be eligible: a PR that was never drafted. I'll add a comment explaining that to the code.

catalog/tests/dags/maintenance/test_pr_review_reminders.py
```diff
@@ -154,38 +164,110 @@ def _setup_branch_protection_for_branch(
...
+                "labeled",
+                "review_requested",
+                "convert_to_draft",
+                ("ready_for_review", 1),
+            ],
+            id="opened_then_drafted_finally_ready",
+        ),
+        # A PR opened as a draft does not have a "convert_to_draft" event but does still have "ready_for_review"
+        pytest.param(
+            ["labeled", "review_requested", ("ready_for_review", 2)],
```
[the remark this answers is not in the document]
sarayourfriend: I'll create a named tuple for it, sure thing :+1:

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
```diff
@@ -28,33 +28,35 @@ class Urgency:
     MEDIUM = Urgency("medium", 4)
     LOW = Urgency("low", 5)
 
+    @classmethod
+    def for_pr(cls, pr: dict) -> Urgency:
```
AetherUnbound, line 32: Given L40, should this be?
```suggestion
    def for_pr(cls, pr: dict) -> Urgency | None:
```

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
```diff
@@ -28,33 +28,35 @@ class Urgency:
...
+        priority_label = priority_labels[0]
+
+        if "critical" in priority_label:
+            return cls.CRITICAL
+        elif "high" in priority_label:
+            return cls.HIGH
+        elif "medium" in priority_label:
+            return cls.MEDIUM
+        elif "low" in priority_label:
+            return cls.LOW
```
AetherUnbound, line 51: I was curious what this might look like using Python's [structural pattern matching syntax](https://peps.python.org/pep-0622/) - in case you're interested as well!

```suggestion
        match priority_label.split()[-1]:
            case "critical":
                return cls.CRITICAL
            case "high":
                return cls.HIGH
            case "medium":
                return cls.MEDIUM
            case "low":
                return cls.LOW
```

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
```diff
@@ -76,16 +78,31 @@ def parse_gh_date(d) -> datetime.datetime:
...
+        # We don't need to check for or stop if we
+        # find a draft event because drafted PRs are
+        # already filtered out at this point. Any draft
+        # events present in the events list would
+        # be further back in time than the latest
+        # (and therefore relevant) ready_for_review
+        # event will appear before we reach any
+        # draft events.
+        if event["event"] == "ready_for_review":
+            ready_for_review_date = event["created_at"]
```
AetherUnbound, line 95: If there are multiple `ready_for_review` events in a PR's event timeline, this seems like it'll get the oldest event - does it make sense to `break` on the case where a ready for review event is found?

catalog/tests/dags/maintenance/test_pr_review_reminders.py
```diff
@@ -315,30 +428,121 @@ def test_does_not_ping_if_pr_has_min_required_approvals(github, urgency, base_br
...
         _setup_branch_protection(github, pr)
 
     post_reminders("not_set", dry_run=False)
 
     assert past_due_pull["number"] not in github["posted_comments"]
     assert not_due_pull["number"] not in github["posted_comments"]
+    assert old_but_not_due_pull["number"] not in github["posted_comments"]
+
+
+parametrize_possible_pingable_events.args
```
AetherUnbound, line 471: Should this be here?

AetherUnbound (approved):
Dang, I'm thankful we have so many thorough tests for this. There is a surprising amount of complexity in all this 😱 

I have a few notes, nothing blocking though! Nice work on this! I was also able to confirm that I got the new reminders when running locally.

catalog/tests/dags/maintenance/test_pr_review_reminders.py
sarayourfriend: It should not, whoops!

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
sarayourfriend: It 100000% should have a break, thank you for catching that!

AetherUnbound:
Okay, the [multiple ready-for-review events](https://github.com/WordPress/openverse/pull/2027#discussion_r1187950952) case was a _really_ tricky one to adequately test for. In trying to create a specific test that failed for the existing implementation but passed as soon as I added a `break`, I identified an issue with the `make_events` fixture: it was creating the events starting at the offset and walking back in time _in the order the event was received_. This meant that a test set like:

```python
events = [
    "labeled",
    "review_requested",
    EventsConfig("ready_for_review", 2),
    EventsConfig("convert_to_draft", 1),
    EventsConfig("ready_for_review", 1),
]
```

Was actually creating these events with `labeled` as the most recent (this is with some additional printing I added, you can see the statements in 2a0055fdf52fb9374192ea31b569291da6b8af72):

```
origin_days_ago=3
event_name='labeled', days_since_previous_event=0
days_ago=3, created_at=datetime.datetime(2022, 6, 14, 0, 0)
event_name='review_requested', days_since_previous_event=0
days_ago=3, created_at=datetime.datetime(2022, 6, 14, 0, 0)
event_name='ready_for_review', days_since_previous_event=2
days_ago=5, created_at=datetime.datetime(2022, 6, 10, 0, 0)
event_name='convert_to_draft', days_since_previous_event=1
days_ago=6, created_at=datetime.datetime(2022, 6, 9, 0, 0)
event_name='ready_for_review', days_since_previous_event=1
days_ago=7, created_at=datetime.datetime(2022, 6, 8, 0, 0)
```

To fix this, I reversed the list of events when performing the walk back, then reverse the returned results so they were given in the right order. One side effect of this was that some of the parameterizations had an `EventConfig` as their final (i.e. most recent) event, with an offset of >1 days. That meant that all critical and some high urgency tests would always fail, because the entire event set was made beyond their urgency boundary. This was addressed by changing those parameter sets and removing the bit of logic in the fixture which was intended to prevent events from being made in the future (which was no longer necessary due to how the events were processed).

Finally, I was able to confirm that the `break` added in 9ae6fd30f016c7fe329c885121310595b3c97e00 addressed a bug by running the tests without it. If you check out 6556943a930aaf0d6daee761b48b52fac6a8d22e and run the tests, you'll notice all of the `test_multiple_drafts_only_uses_most_recent_ready_for_review_for_urgency` cases should fail (along with a number of other instances). After checking out the tip of this branch, all the tests should now pass!

That was a very difficult issue to find due to the complexity of these fixtures :sweat_smile:

AetherUnbound:
@stacimc I've asked for your review again due to the significance of the changes I made. I'm hoping they all make sense, but please let me know if there's more I can explain or more documentation I can add!

catalog/tests/factories/github/__init__.py
```diff
@@ -152,3 +179,92 @@ def make_branch_protection(required_reviewers: int = 2) -> dict:
...
+    """
+    result = []
+    previous_event_days_ago = origin_days_ago
+    for event_name in reversed(events):
+        days_since_previous_event = 0
+        if isinstance(event_name, tuple):
+            days_since_previous_event = event_name[1]
+            event_name = event_name[0]
+
+        days_ago = previous_event_days_ago + days_since_previous_event
```
stacimc, line 209: I think the actual bug with this function was with this line:

```suggestion
                # If the previous event was 2 days ago, and it has been one day since
                # the previous event, then this event was *1 day ago*
                days_ago = previous_event_days_ago - days_since_previous_event
```

Note that should be subtraction, not addition.

Reversing and unreversing the list does make the events happen in the right order chronologically, but completely messes up the actual dates (which is why it was causing failures for critical urgency). So for example here's an event list:

```
events_list = [
  "labeled",
  "review_requested",
  "convert_to_draft",
  ("ready_for_review", 1),
  "convert_to_draft",
  "ready_for_review"
]
```

The way this should read is: "The PR was opened (labeled/review requested) then converted to draft on the same day. One day later it was marked ready for review, converted to a draft, and marked ready for review again".

If I call `make_events` with this event list and `origin_days_ago` set to 2, it should make the list of events with the first event starting 2 days ago. If today is 2023-05-10, I expect results like this:

```
[{'name': 'labeled', 'created_at': '2023-05-08'},  # First event is 2 days ago
 {'name': 'review_requested', 'created_at': '2023-05-08'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-08'},
 {'name': 'ready_for_review', 'created_at': '2023-05-09'}, # 1 day passed between the last event and this one
 {'name': 'convert_to_draft', 'created_at': '2023-05-09'},
 {'name': 'ready_for_review', 'created_at': '2023-05-09'}
]
```

If we don't do the list reversing and instead change that `days_ago` calculation, that is indeed what we get! 

With the list reversing instead, we end up with _the last issue being `origin_days_ago`, rather than the first issue_, and the days between events gets borked:
```
[{'name': 'labeled', 'created_at': '2023-05-05'}, # First event is now *five days ago*
 {'name': 'review_requested', 'created_at': '2023-05-05'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-05'},
 {'name': 'ready_for_review', 'created_at': '2023-05-05'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-08'}, # 3 days have passed before this event
 {'name': 'ready_for_review', 'created_at': '2023-05-08'}]
```

So this breaks the logic for testing the urgency because everything is too far in the past.

I think what we can undo all the changes to reverse/unreverse the list, change this calculation to subtraction, and then we should be able to undo all of the changes to the test parameters and it should just work!

stacimc, line 209: I think the actual bug with this function was with this line:

```suggestion
                # If the previous event was 2 days ago, and it has been one day since
                # the previous event, then this event was *1 day ago*
                days_ago = previous_event_days_ago - days_since_previous_event
```

Note that should be subtraction, not addition.

Reversing and unreversing the list does make the events happen in the right order chronologically, but completely messes up the actual dates (which is why it was causing failures for critical urgency). So for example here's an event list:

```
events_list = [
  "labeled",
  "review_requested",
  "convert_to_draft",
  ("ready_for_review", 1),
  "convert_to_draft",
  "ready_for_review"
]
```

The way this should read is: "The PR was opened (labeled/review requested) then converted to draft on the same day. One day later it was marked ready for review, converted to a draft, and marked ready for review again".

If I call `make_events` with this event list and `origin_days_ago` set to 2, it should make the list of events with the first event starting 2 days ago. If today is 2023-05-10, I expect results like this:

```
[{'name': 'labeled', 'created_at': '2023-05-08'},  # First event is 2 days ago
 {'name': 'review_requested', 'created_at': '2023-05-08'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-08'},
 {'name': 'ready_for_review', 'created_at': '2023-05-09'}, # 1 day passed between the last event and this one
 {'name': 'convert_to_draft', 'created_at': '2023-05-09'},
 {'name': 'ready_for_review', 'created_at': '2023-05-09'}
]
```

If we don't do the list reversing and instead change that `days_ago` calculation, that is indeed what we get! 

With the list reversing instead, we end up with _the last issue being `origin_days_ago`, rather than the first issue_, and the days between events gets borked:
```
[{'name': 'labeled', 'created_at': '2023-05-05'}, # First event is now *five days ago*
 {'name': 'review_requested', 'created_at': '2023-05-05'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-05'},
 {'name': 'ready_for_review', 'created_at': '2023-05-05'},
 {'name': 'convert_to_draft', 'created_at': '2023-05-08'}, # 3 days have passed before this event
 {'name': 'ready_for_review', 'created_at': '2023-05-08'}]
```

So this breaks the logic for testing the urgency because everything is too far in the past.

I think what we can undo all the changes to reverse/unreverse the list, change this calculation to subtraction, and then we should be able to undo all of the changes to the test parameters and it should just work!

stacimc (changes requested):

catalog/tests/factories/github/__init__.py
```diff
@@ -152,3 +179,109 @@ def make_branch_protection(required_reviewers: int = 2) -> dict:
...
+    Create an events list where all events happened today.
+
+    By setting ``origin_days_ago`` to 0 we force the entire
+    events list to happen _now_ which necessarily means
+    none of the events could be considered urgent.
+    """
+    return make_events(0, events)
+
+
+def make_non_urgent_reviewable_events(events: list[str | EventsConfig]) -> list[dict]:
```
stacimc, line 243: Previously this was hard-coding `origin_days_ago` to 0, which forces all the events to happen "today". This does have the effect of ensuring the events list is non-urgent, but it collapsed the list unnecessarily. All we need to really do is ensure that the _last `ready_for_review` event _ happened today.

By updating this to do so, we can add an `events_list` to the parametrized list that has an initial `ready_for_review` event a very long time ago, so that _all_ of our tests will effectively test that only the date of the latest ready_for_review event is considered when checking PRs. (You can verify this: if you remove the [break](https://github.com/WordPress/openverse/pull/2027/files#diff-586c0dab45307447dd65e29600c995a9f4b85cad75f1c241b6f67f8e31d99e3bR98) and run the tests, many will fail.)

stacimc (approved):
I pushed the changes I talked about, so re-requesting a review from @AetherUnbound :)

catalog/tests/factories/github/__init__.py
AetherUnbound: Ahhhh, I got it!! Amazing find!

AetherUnbound (approved):
Tests do break when that  is removed, and your changes make sense! Thanks so much for the help @stacimc!

PR #1973, opened 2023-05-02
Fix documentation rate limiting and formatting issues; fail docs build in CI if docs build has warnings

sarayourfriend:
## Fixes

Fixes #1972 by @sarayourfriend
Fixes #1971 by @sarayourfriend 
Fixes #1860 by @sarayourfriend
Fixes #1861 by @sarayourfriend

## Description

- Fails the documentation build in CI if warnings exist during the build
- Fixes all toctree issues by manually writing all toc trees. Combined with the previous item, there's no need to have general glob usage to ensure documents are included. If they're not included, the build will fail in CI.
- Adds a `LINK_ISSUE_GITHUB_TOKEN` environment variable to prevent rate limiting from breaking the issue references during build. Disables issue link resolution locally to speed up local build. Adds a documentation page for the link issues plugin and the rationale for the changes.
- Update release drafter templates to use static links for changelog PR links going forward. This prevents the number of issues that `link_issues` needs to process each time the docs are built from growing over time. One can foresee a circumstance where eventually we have so many changelogs that even with the GitHub token we get rate limited. Besides the rate limiting, each new issue reference requires a new API request to GitHub and slows down the documentation build process. Issue to retroactively apply the template changes to existing changelogs is here: https://github.com/WordPress/openverse/issues/1974
- Fixes all cross-reference issues
- Normalise project planning document titles
- Adds checkmarks for approved reviewers and updates the reviewer list when needed
- Add all documents that were previously incorrectly excluded from the documentation

## Testing Instructions

Check out the docs preview generated for this PR. To test the changes to the `link_issues` plugin, the best place to look are on the changelogs. In the currently published version of the docs, several changelog pages do not have properly resolved issue links (see issue for an example). There should no longer be any issues that do not have links properly resolved.

Run `just documentation/build` locally and confirm there are no more errors to resolve.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [N/A] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

[...]

PR #2036, opened 2023-05-04
Add GitHub username link to changelog, unify & simplify format

### LINKED ISSUE #1974: Retroactively apply release-drafter template changes to previous doc site changelogs to reduce docs build times in CI
## Problem

Each GitHub issue reference using the `#` format requires a new request to GitHub (they are cached, but they rarely overlap). The changelogs are the biggest source of this. In https://github.com/WordPress/openverse/pull/1973 we introduced a new changelog template that uses static links to PRs. However, that change does not affect previous changelogs, so they're all processed from scratch each time CI builds the documentation.

## Description

Retroactively apply the template changes to the changelogs added before the changes in the linked PR.

AetherUnbound:
## Fixes

Fixes #1974 by @sarayourfriend
(among a few other changes)

## Description

This PR makes a few changes to the changelog format and release draft rendering to unify the various templates, simplify the rendering, and add more context:

- PR authors are now linked to their GitHub profile in the sphinx rendered view
- Removed the "Credits" section, which was inaccurate (see [this comment](https://github.com/WordPress/openverse/pull/1980#discussion_r1183032150) for more info) and made redundant by both the individual `@` references to PR authors and GitHub's own "Contributors" section ([example](https://github.com/WordPress/openverse/releases/tag/catalog-2023.04.27.02.43.13))
- Changed the link format for pull requests to have the link only consist of the actual number, rather than the full PR title. This is consistent with some other changelogs I've seen ([example](https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#new-features)), along with how GitHub renders the links in release format too ([example](https://github.com/WordPress/openverse/releases/tag/catalog-2023.04.27.02.43.13)). It also makes the lines a bit easier to read since the entire line isn't the bright blue of a hyperlink (see below for an example).
- ~~Removed the `link_issues` extension - this was providing a tooltip for the link but that info was the exact same as the PR title which is supplied immediately before it, which makes that redundant. This also prevents having to use any GitHub API requests (described more in #1974) since we're no longer using GitHub to fill some of that information out.~~ No longer correct, see discussion below
- Retroactively applies all of the above to existing changelog files.

There might be more that the `link_issues` extension is doing that I'm unaware of, but we were already no longer leveraging it with the changes in #1973 so I think we're safe to remove it (which is one less thing to vendor and maintain!).

**Before**
image

**After**
image

## Testing Instructions

1. Run `just documentation/live`
1. Go through each existing changelog (or spot check) and ensure that usernames link to GitHub profiles, and PR numbers link to the appropriate PRs.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
> There might be more that the link_issues extension is doing that I'm unaware of, but we were already no longer leveraging it with the changes in https://github.com/WordPress/openverse/pull/1973 so I think we're safe to remove it (which is one less thing to vendor and maintain!).

The link issues extension applies to every other document in the documentation site as well. I don't know if we relied on it there but we did want it, I think? Does it hurt to keep it around, just so documentation behaves more or less like it does on GitHub in that respect?

Will review more closely later.

sarayourfriend (changes requested):
Requesting changes because if we do remove the `link_issues` extension, then we can remove the GITHUB_TOKEN passed to the build-docs action as it is only used by that extension.

As far as removing `link_issues`, it looks like there is one usage of a `#NUM` style issue link in an older project planning document : https://github.com/WordPress/openverse/blob/2e82df0105a94f7b85c48321fc3e05c23edc49bd/documentation/projects/proposals/3d_model_support/20220221-project_proposal.md#L38

Whether that warrants keeping the extension around or not, I'm not sure. Certainly I like the idea that issues and PRs will be automatically linked if they're referenced in that format in our documentation or project planning documents, though we often use more descriptive links in the style of this one from the catalog documentation:

> Direct link to the media file. Note that until [issue #784 is addressed](https://github.com/WordPress/openverse-catalog/issues/784) the field name differs depending on media type.

Anyway, if it's only to catch stragglers, I can see both sides of the argument for keeping it or removing it. Either we remove it because the cases it catches are negligible and keeping the code around incurs more maintenance burden than it needs (and slightly, though not dramatically, slows the docs build); or we keep it because there are indeed cases it catches and while the code is present it doesn't need very many if any modifications for it to continue working into the future (assuming the Sphinx API remains the same).

Anyway, I won't block on that part as I have no strong opinion either way, just slightly leaning towards keeping it for the convenience until it truly becomes a maintenance burden.

The only requested change is to remove the secrets passed to `build-docs`. Otherwise, this LGTM.

AetherUnbound:
Oh right! I thought it was only for the changelog and so was made fully redundant - if it's being used elsewhere (and furthermore, _can_ be used elsewhere without us having to think much about it), then we can keep it. I'll modify the PR so it's no longer removed. I think the only thing I want to double check is that it isn't trying to do anything with the `#<num>` references that are already made into links. If it's still trying to manipulate those, then I think we should get rid of it because it will always be contacting the GitHub API no matter what we do. I'll verify whether that's the case before moving forward either way.

sarayourfriend:
>  I think the only thing I want to double check is that it isn't trying to do anything with the #<num> references that are already made into links.

I already looked into this in the PR where I added the GitHub token, I had the same question. It skips anything that is a "reference" node: 

```py
            if isinstance(parent, nodes.reference):
                continue
```

Links are considered references, so if the pattern is inside something that's already a link, it just skips it.

AetherUnbound:
Excellent!! Thanks, I'll undo that removal tomorrow then 😄

sarayourfriend (approved):
LGTM :+1: If you used a script for converting the old changelogs, can you share it, just in case it's useful again in the future, for whatever reason?

AetherUnbound:
> If you used a script for converting the old changelogs, can you share it, just in case it's useful again in the future, for whatever reason?

I did not, I used a series of regex find-and-replace patterns in PyCharm, several of which I had to correct by hand or redo/run another replace to correct 😅 I am hoping we only have to do this once 😶

dhruvkb (approved):
I like the changes. Thanks!

PR #2041, opened 2023-05-05
Fix false positive test and remove all usages of fraught `shallowMount` utility

sarayourfriend:
## Fixes

Some tests using `shallowMount` (via `test-utils/render`) would always pass despite not being correctly configured.

## Description

This PR fixes that by updating the tests that I could easily identify and then removing all usages of `shallowMount` completely, replacing them with the much harder to misuse (or abuse) testing-library `render`.

## Testing Instructions

Unit tests should pass.

To test the new ESLint rule, checkout the branch and try to import `shallowMount` from `@vue/test-utils`. You should get a new error telling you to use testing-library's render instead.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/unit/specs/components/ImageDetails/v-copy-license.spec.js
```diff
@@ -20,16 +28,22 @@ describe("VCopyLicense", () => {
...
     }
   })
 
-  it("should contain the correct contents", () => {
-    const wrapper = render(VCopyLicense, options)
-    expect(wrapper.find(".copy-license")).toBeDefined()
+  it("should contain the correct contents", async () => {
+    render(VCopyLicense, options)
+    expect(
+      screen.queryAllByText("media-details.reuse.copy-license.copy-text")
```
obulat, line 46: Why do we not need an `await` here?

obulat:
I think re-writing the `render` utilities to use `testing-library/vue`'s render function instead of completely removing it would be a better solution. This would allow us to pass the `i18n` and `pinia` to all of the tests in a central location, instead of repeating it in every test.

I'm working on a new PR that does this, and re-uses all the fixes from this PR.

sarayourfriend:
That sounds good, but if we can get rid of the false-positives in the meantime, this PR should go ahead, unless yours will be ready before this one can get in anyway :slightly_smiling_face:

obulat:
> That sounds good, but if we can get rid of the false-positives in the meantime, this PR should go ahead, unless yours will be ready before this one can get in anyway 🙂

https://github.com/WordPress/openverse/pull/2044 :)

frontend/test/unit/specs/components/ImageDetails/v-copy-license.spec.js
```diff
@@ -20,16 +28,22 @@ describe("VCopyLicense", () => {
...
+    ).toHaveLength(3)
```
AetherUnbound, line 47: Where is this length coming from? Is there something more specific we can check it against?

frontend/test/unit/specs/components/VMediaTag/v-media-tag.spec.js
```diff
@@ -14,9 +13,8 @@ describe("VMediaTag", () => {
   })
 
   it("should render an span tag by default", () => {
-    const wrapper = render(VMediaTag, options)
-    expect(wrapper.vm.$props.tag).toEqual("span")
-    expect(wrapper.vm.$el).toBeInstanceOf(HTMLSpanElement)
+    const { container } = render(VMediaTag, options)
+    expect(container.firstChild.tagName).toEqual("SPAN")
```
AetherUnbound, line 17: Does `SPAN` vs `span` matter here?

frontend/test/unit/specs/components/VMediaTag/v-media-tag.spec.js
```diff
@@ -26,10 +24,9 @@ describe("VMediaTag", () => {
       href: "https://example.com/",
     }
 
-    const wrapper = render(VMediaTag, options)
-    expect(wrapper.vm.$props.tag).toEqual("a")
-    expect(wrapper.vm.$el).toBeInstanceOf(HTMLAnchorElement)
```
AetherUnbound, line 31: Is this check no longer worth having?

AetherUnbound (commented):
I have a few questions, and I'm not sure the ESLint rule is working. Here's my diff:

```diff
diff --git a/frontend/test/unit/specs/components/v-related-audios.spec.js b/frontend/test/unit/specs/components/v-related-audios.spec.js
index 3737f34e7..bf5165a4e 100644
--- a/frontend/test/unit/specs/components/v-related-audios.spec.js
+++ b/frontend/test/unit/specs/components/v-related-audios.spec.js
@@ -1,5 +1,7 @@
 import { screen } from "@testing-library/vue"
 
+import { shallowMount } from "@vue/test-utils"
+
 import { render } from "~~/test/unit/test-utils/render"
 import { getAudioObj } from "~~/test/unit/fixtures/audio"
 
@@ -9,6 +11,7 @@ const audioResults = [getAudioObj(), getAudioObj()]
 
 describe("RelatedAudios", () => {
   it("should render content when finished loading related audios", async () => {
+    shallowMount()
     await render(VRelatedAudio, {
       propsData: {
         media: audioResults,
```

When I run `just lint eslint`, it passes successfully 🤔 am I setting up the expected failure incorrectly?

AetherUnbound (approved):
Oops, I was on the wrong branch 🤦🏼‍♀️ that ESLint warning now comes up for me! And tests pass locally as well 👍🏼

frontend/test/unit/specs/components/ImageDetails/v-copy-license.spec.js
sarayourfriend: `queryAllByText` is synchronous, as far as I can tell :thinking: I think only the "find" methods are async.

frontend/test/unit/specs/components/ImageDetails/v-copy-license.spec.js
sarayourfriend: There are three tabs, so the copy text appears three times. We could check for the presence of the individual tab titles as well if that would be clearer.

frontend/test/unit/specs/components/VMediaTag/v-media-tag.spec.js
sarayourfriend: It does. The previous `render` method returns Vue-dom JavaScript objects whereas the testing-library `screen` query functions return actual (simulated via jsdom) DOM elements. There are situations when the source-casing is relevant (XHTML) but in regular HTML, the element's name is in all caps: https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

I think that's only true for regular HTML elements, though. Custom elements (ones with names that are not part of the HTML spec), remain in their original casing, _I think_. Don't quote me on it :slightly_smiling_face:

frontend/test/unit/specs/components/VMediaTag/v-media-tag.spec.js
sarayourfriend: It's fully redundant with checking the tag name. If the tag name is "A" it will be an anchor element unless we don't think the DOM implementation works correctly.

sarayourfriend:
Closing as @obulat has incorporated these changes into #2044

PR #2043, opened 2023-05-05
Disallow unexpect console output during tests and fix all errors

sarayourfriend:
## Fixes

Similar to #2041 (which this PR builds upon), this PR prevents any unexpected console errors from appearing in tests via the `jest-fail-on-console` utility. This prevents people observing test runs from being confused as to why errors are appearing in the test logs. It also forces us to actually fix these errors instead of glossing over them or ignoring them.

## Description

Introduce `jest-fail-on-console` to prevent unexpected console logging in tests. Please note that with this utility you can still enabled _expected_ console logging of any level and the utility gives clear instructions for how to do so when tests fail due to unexpected console logs.

This PR also then fixes all the instances where tests were logging errors. Mostly this was through fixing a variety of fixtures and test configurations.

## Testing Instructions

Run the unit tests and confirm that they pass (you can rely on CI for this). Also check the CI output to confirm that there are no logged warnings or errors of any kind in the tests (indeed, they should fail if there are).

Check out the branch and add a console error to a test locally to test that the `jest-fail-on-console` utility is configured correctly.

Finally, because I had to make changes to the focus trap, test some dialogues in the running application to make sure they still work as expected. Specifically test that when opening them via keyboard navigation that the correct element is focused when the dialogue is opened.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/unit/setup-after-env.js
+++ b/frontend/test/unit/setup-after-env.js
@@ -1,5 +1,8 @@
 import Vue from "vue"
 import "@testing-library/jest-dom"
+import failOnConsole from "jest-fail-on-console"
+
+failOnConsole()
 
 Vue.prototype.$nuxt = {
   context: {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -150,6 +150,7 @@
     "eslint-plugin-vue": "^9.9.0",
     "eslint-plugin-vuejs-accessibility": "^1.1.1",
     "jest": "^26.6.3",
+    "jest-fail-on-console": "^3.1.1",
     "jest-transform-stub": "^2.0.0",
     "module-alias": "^2.2.2",
     "npm-run-all": "^4.1.5",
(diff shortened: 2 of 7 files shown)


frontend/test/unit/setup-after-env.js
```diff
@@ -1,5 +1,8 @@
 import Vue from "vue"
 import "@testing-library/jest-dom"
+import failOnConsole from "jest-fail-on-console"
+
+failOnConsole()
```
dhruvkb, line 5: Will this also catch output from our custom console utils? They indirectly refer to the inbuilt `console` functions but I'm asking just to be sure.

dhruvkb (approved):
Thanks for fixing a bunch of these errors. LGTM, except one doubt.

frontend/test/unit/setup-after-env.js
sarayourfriend: Yup! The library is implemented by modifying the global console object so unless we were doing something like writing directly to `stdout`, we wouldn't be able to easily _accidentally_ bypass this utility.

https://github.com/ValentinH/jest-fail-on-console/blob/main/index.js#L102-L115

sarayourfriend:
Closing as @obulat has incorporated these changes into #2044

PR #2046, opened 2023-05-06
Determine labels for checks using labels.yml, factor out common workflow utils

### LINKED ISSUE #1195: Determine labels for check using labels.yml
## Description

Currently, any time we add a new option for a required label category, we have to update the label check PR manually:

https://github.com/WordPress/openverse/blob/bc60edc64fe34e59bbda5b880db76f4b1b41b9da/.github/workflows/pr_label_check.yml#L17-L55

It would be ideal if we could use the values from [`labels.yml`](https://github.com/WordPress/openverse/blame/4cb78373557c4619ec67fb2ac570d2e28d9e0994/automations/data/labels.yml) in order to determine acceptable values.

## Additional context

Related to #1194

AetherUnbound:
## Fixes

Fixes #1195 by @AetherUnbound

## Description

This PR is primarily intended to modify the `pr_label_check.yml` GitHub Action workflow so that it reads from the `labels.yml` definition rather then requiring the label groups be explicitly redefined. In order to do this, I factored out some shared functions which read in the labels and convert them to Python objects. I also factored out code in several of our workflow python scripts which would print data to both `stdout` and the `GITHUB_OUTPUTS` destinations. I had to add the `PYTHONPATH` environment variable to the places where we call python scripts in order to have those shared modules load properly.

The effect of this is now any labels we add to the required categories in `labels.yml` will automatically be added to the `pr_label_check` workflow!

## Testing Instructions

The fact that CI passes here is sign that things are working! See the output from the adjusted actions to see that they are still outputting what they were before.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/pr_label_check.yml
+++ b/.github/workflows/pr_label_check.yml
@@ -14,46 +14,74 @@ on:
       - unlabeled
 
 jobs:
+  # Extract the current label list by group
+  get_label_groups:
+    name: Get label groups
+    runs-on: ubuntu-latest
+    outputs:
+      aspect: ${{ steps.get_groups.outputs.aspect }}
+      goal: ${{ steps.get_groups.outputs.goal }}
+      priority: ${{ steps.get_groups.outputs.priority }}
+      stack: ${{ steps.get_groups.outputs.stack }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Get label groups
+        id: get_groups
+        env:
+          PYTHONPATH: ${{ github.workspace }}/automations/python
+        working-directory: automations/python/workflows
+        run: python get_label_groups.py
+
   # List of aspect labels: https://github.com/WordPress/openverse/labels?q=aspect
   check_aspect_label:
     name: Check aspect label
     runs-on: ubuntu-latest
+    needs:
+      - get_label_groups
     steps:
       - name: Check aspect label
         uses: docker://agilepathway/pull-request-label-checker:v1.4.26
         with:
-          one_of: "💻 aspect: code,📄 aspect: text,🤖 aspect: dx,🕹 aspect: interface,♿️ aspect: a11y"
+          one_of: ${{ needs.get_label_groups.outputs.aspect }}
           repo_token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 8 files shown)


automations/python/shared/actions.py
```diff
@@ -0,0 +1,15 @@
+import os
+import sys
+
+
+def write_to_github_output(lines: list[str]) -> None:
+    """
+    Write a list of strings to both stdout and the GITHUB_OUTPUT environment variable.
```
obulat, line 7: Why do we need both? What is `stdout` output used for?

.github/workflows/pr_label_check.yml
```diff
@@ -14,46 +14,74 @@ on:
...
     name: Check stack label
     runs-on: ubuntu-latest
+    needs:
+      - get_label_groups
     steps:
       - name: Check stack label
         uses: docker://agilepathway/pull-request-label-checker:v1.4.26
         with:
-          any_of: "🧱 stack: backend,🧱 stack: analytics,🧱 stack: documentation,🧱 stack: frontend,🧱 stack: mgmt,🧱 stack: api,🧱 stack: catalog,🧱 stack: ingestion server,🧱 stack: infra"
+          any_of: ${{ needs.get_label_groups.outputs.stack }}
```
dhruvkb, line 86: One way this can be simplified. Although I won't mind if you decided against this recommended and favoured consistency instead.

```suggestion
          prefix_mode: true
          any_of: "🧱 stack:"
```

automations/python/shared/actions.py
dhruvkb: I like writing to `stdout` so that it's printed in the workflow logs and helps with debugging.

dhruvkb (approved):
LGTM, this is great. Now only if there was a way to merge `.github/filters.yml` into the labels file as well.

sarayourfriend (approved):
:cool: LGTM!

.github/workflows/pr_label_check.yml
AetherUnbound: Oh right! Yea I think it might be best if we stick to the `labels.yml` definition - that way this will fail if we add a new stack label manually without updating `labels.yml`.

PR #2048, opened 2023-05-06
Use additional build contexts to build frontend Docker image

### LINKED ISSUE #2040: Pass `-f` to frontend/Dockerfile `rm` run so that missing files do not cause errors
## Problem

It isn't possible to run the frontend Dockerfile build locally any more because it assumes that it is running in a context where the CI/CD workflow has copied certain files into the `frontend` directory. Because `rm` fails if files do not exist, when you run the build locally (assuming you didn't also needlessly copy the files in question), the build will fail.

## Description

Add `-f` to this line in `frontend/Dockerfile`:

```dockerfile
RUN rm ./frontend/.npmrc ./frontend/.pnpmfile.cjs ./frontend/pnpm-lock.yaml
```

should read:

```dockerfile
RUN rm -f ./frontend/.npmrc ./frontend/.pnpmfile.cjs ./frontend/pnpm-lock.yaml
```

## Alternatives

Change the way the CI/CD workflow runs so that it doesn't rely on these files being moved around at all. Then we can remove this CI implementation detail from the Dockerfile entirely, which ideally wouldn't know anything about it.

To do this, we probably need to more heavily leverage `pnpm`'s workspace stuff, but I'm not sure. It's almost certainly a lot more work to untangle the assumptions there.

dhruvkb:
## Fixes

Fixes #2040 by @sarayourfriend 

## Description

This PR
- makes the repo root an additional build context (named `repo_root`)
- uses this build context in the building of the `openverse-frontend` image
- eliminates the need to first move files into the `frontend/` directory and then delete them inside the `Dockerfile`

## Testing Instructions

The images should build and CI should pass. Additonally,
- Check out the PR.
- Build the frontend image locally.

  ```bash
  $ just frontend/build-img
  ```

- Run the frontend.

  ```bash
  $ docker run --rm --port 8443:8443 openverse-frontend:local
  ```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/general/general_setup.md
+++ b/documentation/general/general_setup.md
@@ -184,6 +184,19 @@ If not, you can use one of the many
 [installation methods](https://docs.docker.com/engine/install/) officially
 supported by Docker.
 
+Note that the building images for the frontend needs an advanced build feature,
+namely additional build contexts, that is provided by
+[Buildx](https://docs.docker.com/build/architecture/#buildx). Buildx is included
+in Docker Desktop and all
+[installation methods](https://docs.docker.com/engine/install/) for Docker
+Engine include steps to also install the Buildx plugin `docker-buildx-plugin`.
+
+```{tip}
+If it is not already, you should
+[make Buildx the default](https://github.com/docker/buildx#set-buildx-as-the-default-builder)
+builder.
+```
+
 #### Docker Compose
 
 Docker Compose makes it easy to orchestrate multiple services which we put to
--- a/documentation/general/general_setup.md
+++ b/documentation/general/general_setup.md
@@ -185,17 +185,20 @@ If not, you can use one of the many
 supported by Docker.
 
 Note that the building images for the frontend needs an advanced build feature,
-namely additional build contexts, that is provided by
-[Buildx](https://docs.docker.com/build/architecture/#buildx). Buildx is included
-in Docker Desktop and all
+namely additional build contexts, that is provided by the
+[Buildx plugin](https://docs.docker.com/build/architecture/#buildx). The Buildx
+plugin is included in Docker Desktop and all
 [installation methods](https://docs.docker.com/engine/install/) for Docker
-Engine include steps to also install the Buildx plugin `docker-buildx-plugin`.
+Engine include steps to also install the Buildx plugin.
 
-```{tip}
-If it is not already, you should
-[make Buildx the default](https://github.com/docker/buildx#set-buildx-as-the-default-builder)
-builder.
+To know if your Docker setup is using Buildx, run the following command.
+
+```console
+$ docker build --help
 ```
+
+If you see `docker buildx build` under 'Aliases', you have Buildx set up as the
+build client.
 
 #### Docker Compose
 
--- a/frontend/justfile
+++ b/frontend/justfile
@@ -25,6 +25,23 @@ export FRONTEND_PNPM_VERSION := `grep 'packageManager' ../package.json | awk -F'
 @pnpm-version:
     echo $FRONTEND_PNPM_VERSION
 
+##########
+# Docker #
+##########
+
+# This section exists because the frontend is not a part of the Docker-based setup
+# and cannot be built easily like `just dc build <service>`.
+
+# Build the frontend Docker image
+build-img tag="openverse-frontend:local" target="app":
+    docker buildx build \
+    --target {{ target }} \
+    --build-context 'repo_root=..' \
+    --build-arg FRONTEND_NODE_VERSION=$(just node-version) \
+    --build-arg FRONTEND_PNPM_VERSION=$(just pnpm-version) \
+    --tag {{ tag }} \
+    .
+
 ######
 # Up #
 ######
--- a/frontend/justfile
+++ b/frontend/justfile
@@ -35,12 +35,19 @@ export FRONTEND_PNPM_VERSION := `grep 'packageManager' ../package.json | awk -F'
 # Build the frontend Docker image
 build-img tag="openverse-frontend:local" target="app":
     docker build \
-    --target {{ target }} \
-    --build-context 'repo_root=..' \
-    --build-arg FRONTEND_NODE_VERSION=$(just node-version) \
-    --build-arg FRONTEND_PNPM_VERSION=$(just pnpm-version) \
-    --tag {{ tag }} \
-    .
+      --target {{ target }} \
+      --build-context 'repo_root=..' \
+      --build-arg FRONTEND_NODE_VERSION=$(just node-version) \
+      --build-arg FRONTEND_PNPM_VERSION=$(just pnpm-version) \
+      --tag {{ tag }} \
+      .
+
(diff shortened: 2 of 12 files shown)


sarayourfriend:
@dhruvkb I added skip-changelog to this issue because the stack labels were acting strangely; this change seems like it should be _just_ mgmt anyway and wouldn't appear in the changelogs in that case.

frontend/justfile
```diff
@@ -25,6 +25,23 @@ export FRONTEND_PNPM_VERSION := `grep 'packageManager' ../package.json | awk -F'
...
+##########
+# Docker #
+##########
+
+# This section exists because the frontend is not a part of the Docker-based setup
+# and cannot be built easily like `just dc build <service>`.
+
+# Build the frontend Docker image
+build-img tag="openverse-frontend:local" target="app":
+    docker buildx build \
```
sarayourfriend, line 37: Can you add a note in the frontend documentation about needing to install (and potentially enable) `docker buildx`? [It's included by default in Docker Desktop but requires a separate step for Docker Engine](https://docs.docker.com/build/architecture/#install-buildx):

> Docker Engine version 23.0 and later requires that you install Buildx from a separate package

The error message you get if you do _not_ install buildx is nasty and unhelpful too:

```
➜  openverse git:(frontend_no_mv) just frontend/build-img
docker buildx build --target app --build-context 'repo_root=..' --build-arg FRONTEND_NODE_VERSION=$(just node-version) --build-arg FRONTEND_PNPM_VERSION=$(just pnpm-version) --tag openverse-frontend:local .
unknown flag: --target
See 'docker --help'.
```

Nothing to indicate (directly) that you don't have buildx installed.

sarayourfriend (changes requested):
LGTM! The build ran fine locally, the container serves the frontend as expected, and CI passes. Good enough for me!

I left one request to add documentation noting the buildx requirement somewhere in the frontend documentation. It's not a normal dev-lifecycle requirement so a small aside somewhere would be enough, in my opinion.

frontend/justfile
dhruvkb: @sarayourfriend it's installed as a separate package, yes, but it's included in all the [installation methods](https://docs.docker.com/engine/install/). If you're on Docker Engine v23 and above, and followed the instructions verbatim, you have Buildx and `docker build` is an alias for `docker buildx build`.

I've added a note about needing Buildx to the docs in bfb6f83.

documentation/general/general_setup.md
```diff
@@ -184,6 +184,19 @@ If not, you can use one of the many
...
+[Buildx](https://docs.docker.com/build/architecture/#buildx). Buildx is included
+in Docker Desktop and all
+[installation methods](https://docs.docker.com/engine/install/) for Docker
+Engine include steps to also install the Buildx plugin `docker-buildx-plugin`.
+
+```{tip}
+If it is not already, you should
+[make Buildx the default](https://github.com/docker/buildx#set-buildx-as-the-default-builder)
+builder.
+```
```
sarayourfriend, line 198: Why is this?

documentation/general/general_setup.md
```diff
@@ -184,6 +184,19 @@ If not, you can use one of the many
 [installation methods](https://docs.docker.com/engine/install/) officially
 supported by Docker.
 
+Note that the building images for the frontend needs an advanced build feature,
+namely additional build contexts, that is provided by
+[Buildx](https://docs.docker.com/build/architecture/#buildx). Buildx is included
+in Docker Desktop and all
+[installation methods](https://docs.docker.com/engine/install/) for Docker
+Engine include steps to also install the Buildx plugin `docker-buildx-plugin`.
```
sarayourfriend, line 192: Is `docker-buildx-plugin` the package name? If so, this differs by distro and I think the link to the documentation is better suited for that, or a generic instruction to search for your distro's installation instructions.

documentation/general/general_setup.md
dhruvkb: Because Buildx has more features compared to the legacy build system. Case in point: additional build contexts.

documentation/general/general_setup.md
dhruvkb: I had made sure that the name was the same for all distros before referring to it in the docs.

Fedora:
Screenshot 2023-05-08 at 8 01 55 AM

CentOS/RHEL:
Screenshot 2023-05-08 at 8 01 03 AM

Debian/Ubuntu:
Screenshot 2023-05-08 at 8 01 24 AM

SLES:
Screenshot 2023-05-08 at 8 02 49 AM

sarayourfriend: For Arch it's `docker-buildx`. But of course, anyone not using one of those distros probably already knows to look elsewhere. In any case, repeating the implementation detail here is strange, and it could easily go out of date and cause confusion.

documentation/general/general_setup.md
sarayourfriend: But why should an Openverse contributor do this? I haven't done it and haven't run into any issues. Is it a requirement of the project? This tip has strong language "should" but isn't clear as to the _reason_, and considering it has system-wide implications (setting a different default), it should have a good reason to back it up. Other projects the person reading contributes to may not work with buildx, for example.

dhruvkb: Buildx is backwards compatible so using it as the default will not negatively affect any existing workflows.

dhruvkb: Anyway, Buildx is the default on engine v23+ and so I've removed that tip from the docs.

sarayourfriend (approved):
LGTM, thanks for the docs updates!

frontend/justfile
```diff
@@ -25,6 +25,23 @@ export FRONTEND_PNPM_VERSION := `grep 'packageManager' ../package.json | awk -F'
...
 
+##########
+# Docker #
+##########
+
+# This section exists because the frontend is not a part of the Docker-based setup
+# and cannot be built easily like `just dc build <service>`.
+
+# Build the frontend Docker image
+build-img tag="openverse-frontend:local" target="app":
```
AetherUnbound, line 36: Could this be just `build`, for consistency with the top-level `just build` command (even if we can't incorporate it in `just build` easily)?

AetherUnbound (approved):
The changes look good and I was able to build and run everything locally! I have a few non-blocking changes. Would it also be possible to add the `docker run` command you supply as a `just` recipe for the frontend? Perhaps `just frontend/docker-run`? That way we could reference it directly in documentation & test instructions! Also it looks like it should be `--publish` and not `--port`

frontend/justfile
dhruvkb: I kept the name different in case we adopted the Docker model for the frontend. I recalled that someone on the team (maybe @sarayourfriend?) had suggested something along those lines earlier.

PR #2049, opened 2023-05-06
Fix i18n setup in the unit tests

### LINKED ISSUE #2050: The i18n setup in unit tests does not work
## Description

We mock i18n in the unit tests, however, it does not work correctly, so Vue-i18n falls back to the key value instead of translation. 
Another problem is that we mock i18n in a several slightly-different ways in different tests, and we repeat the set up in many tests.

## Additional context

There are several ways we use i18n throughout the app: using Nuxt context inside the components' `setup` function and the Vue's `$t` in components. Both needs to be mocked.

obulat:
## Fixes

Fixes #2050 by @obulat

## Description

This PR correctly sets up i18n in unit tests so that the keys are actually translated, both in the `setup` function that uses `useContext().i18n` from `Nuxt`, and in the templates' `$t` function.

Over the time, we have used several ways of mocking i18n in unit tests, and i18n is  set up slightly differently for different tests. This PR simplifies the set up by moving all of it into 3 points:
- `i18n` instance is set up in `test-utils/i18n.js`, and imported from there
- this instance is used in the `$nuxt` context mock - to be used everywhere where we use `useI18n()` in the components' `setup` functions.
- this instance is also used in the `render` function that mounts the component for tests. This is passed to `Vue` and is used in the templates' `$t` function.

All of the `i18n` set up and mocking from the components was removed.
This PR also uses `test-utils/render` for all components to correctly pass the `i18n` to all of the components. Using `render` from `testing-library` would require us to set up `i18n` separately (which is not necessary, and would needlessly complicate the set up).

## Testing Instructions

All the unit tests should pass. They should use the translated strings instead of the keys for selecting elements (see how the element selectors in the component tests have changed from e.g. "audio-details.related-audios" key to the value we can see in the UI ("Related audio")).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VFilters/VSearchGridFilter.vue
+++ b/frontend/src/components/VFilters/VSearchGridFilter.vue
@@ -33,7 +33,7 @@
 <script lang="ts">
 import { computed, defineComponent, ref } from "vue"
 
-import { useContext, useRouter } from "@nuxtjs/composition-api"
+import { useRouter } from "@nuxtjs/composition-api"
 import { kebab } from "case"
 
 import { watchDebounced } from "@vueuse/core"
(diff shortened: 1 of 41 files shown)


sarayourfriend (approved):
Great!

zackkrida (approved):
This is a significant improvement. One question: should we mention in our documentation somewhere that folks should use plain text strings in tests instead of translation keys, now?

PR #2057, opened 2023-05-08
Derive document sensitivity markings from filtered index

### LINKED ISSUE #1199: Derive the list of sensitive document IDs based on the filtered index
## Problem

Blocked by #1198

We need to add the `sensitivity` field to the media serializers and populate it based on additional context returned by `search_controller.search`.

## Description

Follow the instructions in [the implementation plan section for this step](https://docs.openverse.org/projects/proposals/detecting_sensitive_textual_content/20230308-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html#derive-the-list-of-sensitive-document-ids-based-on-the-filtered-index-overview-step-4) to derive the list of user reported sensitive content IDs, provider supplied sensitive content IDs, and the IDs for documents that we've detected have sensitive textual content.

sarayourfriend:
## Fixes

Fixes #1199 by @sarayourfriend 

## Description

This PR includes significant additions to the test factories and fixtures in order to facilitate Elasticsearch document creation. I chose this approach over ElasticMock for the following reasons:

1. It allows us to avoid mocking altogether (great for tests generally)
2. It exercises our _actual_ indexes, not fake ones
3. It still follows the general test-data hygiene rules of removing test data after each test case

The last one is the source of some abstraction that is not necessarily easy to follow (in that the parts aren't explicitly connected), but at least somewhat tied together via the `CREATED_BY_FIXTURE_MARKER`. The usage of pytest's `request` fixture is also potentially obscure but is quite common in pytest plugins of any significant complexity. I learned that specific usage (how to look for a specific mark) reading `pytest-django`'s code.

I also added a new `media_type_config` fixture that provides a dataclass containing most of the relevant disparate bits for different media types necessary for many of our tests. This made it easier to update tests that were previously using _very_ naive and shallow stubs which broke due to the new requirements. Luckily, the changes to those tests are easier to understand and will continue to benefit from additional improvements to the fixture classes. We'll be able to continue cleaning up tests with this fixture in the future in separate PRs, but not this one. In this one I tried to touch only the tests that actually needed changes.

Anyway, with that in mind, here is a summary of the significant changes:

- Enable Elasticsearch document creation in model factories
  - I needed to fix the Postgres ID creation for this because it is also used to set the document ID in Elasticsearch. The query for the latest ID needs to go to Elasticsearch because in tests the models are not able to query the existing data (it seems).
- Add `unstable__sensitivity` to the result serializer
  - This is marked as `unstable` and will be stabilised alongside the new query parameter after testing
- Remove ``request`` parameter from `search_controller.search` and `search_controller.related_media`. This parameter was not being used and made tests slightly more annoying to write as you had to either pass a real request or `None`. The first requires calling a new fixture and the other is obscure and unclear.
- Update many unit tests to accommodate the new `search_controller.search` interface (made easier by the improved fixtures and factories).

This PR includes a deviation from the original plan. Specifically, in the course of writing this PR and testing it, I discovered that we do not actually have _any_ documents marked mature in our production dataset that are not marked mature due to confirmed user reports. That's to say: we have no provider-supplied maturity. I left the constant in there along with a big comment in the new serializer method explaining why that is. Trying to accommodate that in this PR would significantly increase the complexity because we need to be able to efficiently (read: no duplicate queries) read from the database _and_ ES, regardless of the serialization context. It turns out that's not that hard for our actual runtime, but many many many many many tests assume the ability to bypass Elasticsearch (probably because we did not have easy Elasticsearch fixture creation). We need to put off that feature and luckily we don't need it because simply put, there are no results that fall into that case.

## Testing Instructions

Review the new unit tests. There are significant changes to the test factories to support these tests and they integrate directly with the live Elasticsearch instance.

Additionally, test by making queries. You must enable `ENABLE_FILTERED_INDEX_QUERIES` for the new parameter to work. Refer to `ingestion_server/static/mock_sensitive_terms.txt` for the terms used to generate the local filtered indexes (these terms are _not_ sensitive and are only for testing).

You should observe that requests that do not include sensitive results should always have `[]` for `unstable__sensitivity`.

You must create a confirmed mature report for results to test the "user reported sensitive" results.

You should query for one of the mock sensitive terms directly while including sensitive results and confirm that the documents appropriately include the sensitive text marking in `unstable__sensitivity`. Try a combination of the two as well, (i.e., create a confirmed report for a "bird" result locally). The new field should have both text and user-reported sensitivity indications.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/search_context.py
+++ b/api/api/utils/search_context.py
@@ -1,0 +1,78 @@
+from dataclasses import asdict, dataclass
+from typing import Self
+
+from django.db.models import CharField
+from django.db.models.functions import Cast
+
+from elasticsearch_dsl import Q, Search
+from elasticsearch_dsl.response import Hit
+
+from api import models
+from api.constants.media_types import OriginIndex
+
+
+@dataclass
+class SearchContext:
+    all_result_identifiers: set[str]
+    """All the result identifiers gathered for the search."""
+
+    user_reported_sensitive_result_identifiers: set[str]
+    """Subset of result identifiers for results with confirmed sensitive reports."""
+
+    sensitive_text_result_identifiers: set[str]
+    """Subset of result identifiers for results with sensitive textual content."""
+
+    @classmethod
+    def get_mature_media_class_for_index(
+        cls, origin_index: OriginIndex
+    ) -> models.MatureImage | models.MatureAudio:
+        match origin_index:
+            case "image":
+                return models.MatureImage
+            case "audio":
+                return models.MatureAudio
+            case _:
+                raise ValueError(f"Unknown `origin_index` value '{origin_index}'")
+
+    @classmethod
+    def build(cls, results: list[Hit], origin_index: OriginIndex) -> Self:
+        if not results:
+            return cls(set(), set(), set())
+
+        all_result_identifiers = {r.identifier for r in results}
+
+        filtered_index_search = Search(index=f"{origin_index}-filtered")
+        filtered_index_search = filtered_index_search.query(
+            Q("terms", id=[result.id for result in results])
+        )
+
+        results_in_filtered_index = filtered_index_search.execute()
+        filtered_index_identifiers = {
+            result.identifier for result in results_in_filtered_index
+        }
+        sensitive_text_result_identifiers = {
+            identifier
+            for identifier in all_result_identifiers
+            if identifier not in filtered_index_identifiers
+        }
+
+        mature_media_class = cls.get_mature_media_class_for_index(origin_index)
+
+        user_reported_sensitive_result_identifiers = set(
+            mature_media_class.objects
+            # ``identifier`` comes back as a UUID by default, so it needs to be
+            # cast to string to match the interface of the sets expected
+            # by ``SearchContext``
+            .annotate(identifier_string=Cast("media_obj_id", output_field=CharField()))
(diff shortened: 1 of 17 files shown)


obulat (approved):
I tested locally using the instructions, and everything worked perfectly. Thank you for great instructions! The testing changes look good, too.
Note: There are some merge conflicts that need to be resolved.

api/api/utils/search_context.py
```diff
@@ -0,0 +1,53 @@
...
+            if identifier not in filtered_index_identifiers
+        }
+
+        return cls(
+            all_result_identifiers=all_result_identifiers,
+            sensitive_text_result_identifiers=sensitive_text_result_identifiers,
+        )
+
+    def asdict(self):
+        return asdict(self)
```
dhruvkb, line 53: Nice. I initially (and incorrectly) thought it would infinitely recurse. Also a comment explaining why this is needed would be nice, particularly why it's better than invoking `dataclasses.asdict` directly.

dhruvkb (approved):
LGTM. I have a question but nothing serious enough to be a blocker. Thanks for the detailed PR description and testing instructions.

PR #2061, opened 2023-05-08
Add throttle classess for health check endpoint

### LINKED ISSUE #1924: Healthcheck endpoint throttles load balancer health checks
## Description

The API healthcheck endpoint uses the default anon rate limits. This causes the load balancers to be throttled when making healthcheck requests.

## Reproduction

This is evident in deployment logs (non-public) but is reproduced locally by turning on throttling and making healthcheck requests.

## Additional context

We should increase the throttle rate for this to 2/minute to match the load balancer's healthcheck interval of 30 seconds.

I do not think we should increase the load balancers healthcheck interval as it would increase deployment times.

krysal:
## Fixes

Fixes #1924 by @sarayourfriend 

## Description

Increase the throttle rate for this endpoint as per suggested in the issue (albeit giving it a leeway). I believe this is the first use of the `ExemptOAuth2IdRateThrottle` class as well, seems suitable in this case though I don't think we make many of these requests authenticated.

## Testing Instructions

Run the django app and try to make several requests to http://localhost:50280/healthcheck/ See

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/throttle.py
+++ b/api/api/utils/throttle.py
@@ -82,6 +82,10 @@ class SustainedRateThrottle(AbstractAnonRateThrottle):
     scope = "anon_sustained"
 
 
+class HealthcheckAnonRateThrottle(AbstractAnonRateThrottle):
+    scope = "anon_healthcheck"
+
+
 class AnonThumbnailRateThrottle(AbstractAnonRateThrottle):
     scope = "anon_thumbnail"
 
--- a/api/api/views/health_views.py
+++ b/api/api/views/health_views.py
@@ -6,6 +6,8 @@ from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+from api.utils.throttle import ExemptOAuth2IdRateThrottle, HealthcheckAnonRateThrottle
+
 
 class ElasticsearchHealthcheckException(APIException):
     status_code = status.HTTP_503_SERVICE_UNAVAILABLE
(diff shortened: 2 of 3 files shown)


dhruvkb (approved):
Could you add some unit tests, please? Other than that, LGTM! Approving to unblock.

sarayourfriend (approved):
LGTM. I don't know exactly what kind of unit test would be useful for this as all the ones I can think of would mostly be testing DRF rather than our code. This PR seems more like configuration than anything.

By the way, `ExemptOAuth2IdRateThrottle` is in the default throttle classes and the frontend relies on it already, so it is well tested in production.

krysal:
I thought the same as Sara first but as per Dhruvk request, I checked test locally with `DISABLE_GLOBAL_THROTTLING=False` and health check tests failed. This situation is a bit complicated as I can't find a way to disable throttling _only for certain tests_. So I'll merge this for now as that would require more investigation but this is necessary for the following deployments.

sarayourfriend:
> I thought the same as Sara first but as per Dhruvk request, I checked test locally with DISABLE_GLOBAL_THROTTLING=False and health check tests failed. This situation is a bit complicated as I can't find a way to disable throttling only for certain tests. So I'll merge this for now as that would require more investigation but this is necessary for the following deployments.

We may need an issue to do this as it would be helpful for some tests we already have that will fail if you run them multiple times in succession due to rate limiting.

To do it we'd need to update our rate limiting class to check settings at runtime. As it stands, our approach to disabling throttling is to just remove all the throttling configuration in the settings at start-up so any runtime changes to the `DISABLE_GLOBAL_THROTTLING` doesn't cause a difference unless you force the settings module to be re-evaluated.

PR #2073, opened 2023-05-10
Add `internal__index` param, walled behind auth

### LINKED ISSUE #2070: Allow choosing preferred ES index via an internal query param
## Problem

To ensure that a given index is working without problems, allow choosing the preferred index when performing a search using an `internal__index` parameter. This is intended to be used by the ingestion server to make sure the index works before promoting it as the default index for a given media type.

## Additional context

This is similar to the following existing features.

- The feature is hidden behind auth and not documented, similar to the `unstable__` parameters.

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L137

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L144

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L152

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L159

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L168

- The feature changes the ES index used by the search, similar to the `qs` parameter. Once implemented, the `qs` parameter would be mutually exclusive with it and would be a shorthand for `internal__index=search-qa-<media_type>`.

https://github.com/WordPress/openverse/blob/b2bcc39f291273d860d4753b6286c93b14d88999/api/api/serializers/media_serializers.py#L124

dhruvkb:
## Fixes

Fixes #2070 by @dhruvkb

## Description

This PR
- adds a new query param, called `internal__index`
- this param, only accessible behind authentication, allows choosing a different index to use when searching
- this param conflicts with the `qa` param, and only one of the two can be set at a time

## Testing Instructions

1. Using the ingestion server API, create a new index for image and audio, name it something like `index-foo` and `audio-bar`.
2. Search for something. Note the top few results. They should be the same when searching the default index and from the new one.
3. Randomly delete some of the top few results from the new index. You can do this in Elasticvue.
4. Observe no change when searching the default index but new results when searching the new one.
5. Use can also keep an eye on the logs because it will print which index is being used for the search.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -25,13 +25,15 @@ from api.utils.url import add_protocol
 
 
 @extend_schema_serializer(
-    # Hide unstable fields from documentation, also see `field_names` below.
+    # Hide unstable and internal fields from documentation.
+    # Also see `field_names` below.
     exclude_fields=[
         "unstable__sort_by",
         "unstable__sort_dir",
         "unstable__authority",
         "unstable__authority_boost",
         "unstable__include_sensitive_results",
+        "internal__index",
     ],
 )
 class MediaSearchRequestSerializer(serializers.Serializer):
@@ -173,6 +175,16 @@ class MediaSearchRequestSerializer(serializers.Serializer):
         default=False,
     )
 
+    # The ``internal__`` prefix is used in the query params.
+    # If you rename these fields, update the following references:
+    #   - ``field_names`` in ``MediaSearchRequestSerializer``
+    #   - validators for these fields in ``MediaSearchRequestSerializer``
+    internal__index = serializers.CharField(
+        source="index",
+        help_text="The index against which to perform the search.",
+        required=False,
+    )
+
     page_size = serializers.IntegerField(
         label="page_size",
         help_text="Number of results to return per page.",
@@ -260,6 +272,23 @@ class MediaSearchRequestSerializer(serializers.Serializer):
 
         return self.initial_data.get("mature") or value
 
+    def validate_internal__index(self, value):
+        """
+        Check whether the given index name is a valid index or alias. However,
+        for unauthenticated requests, no check is performed and ``None`` is
+        returned immediately.
+
+        :param value: the provided index name to check
+        :return: ``None`` if request is anonymous, the provided name if it is valid
+        :raise: ``serializers.ValidationError`` if not anonymous and invalid index name
+        """
+
+        if self.is_request_anonymous():
+            return None
+        if not settings.ES.indices.exists(value):  # ``exists`` includes aliases.
+            raise serializers.ValidationError(f"Invalid index name `{value}`.")
+        return value
+
     def validate_page_size(self, value):
         is_anonymous = self.is_request_anonymous()
         max_value = (
@@ -303,6 +332,12 @@ class MediaSearchRequestSerializer(serializers.Serializer):
                 )
         if errors:
             raise serializers.ValidationError(errors)
+
+        if "qa" in self.initial_data and "internal__index" in self.initial_data:
+            raise serializers.ValidationError(
+                "Cannot set both 'qa' and 'internal__index'. "
+                "Use exactly one of these."
+            )
         return data
 
     @property
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -88,17 +88,21 @@ class MediaViewSet(ReadOnlyModelViewSet):
         if qa:
             logger.info("Using QA index for media.")
             search_index = self.qa_index
+            exact_index = False
         elif pref_index := params.validated_data.get("index"):
             logger.info(f"Using preferred index {pref_index} for media.")
             search_index = pref_index
+            exact_index = True
         else:
             logger.info("Using default index for media.")
             search_index = self.default_index
+            exact_index = False
 
         try:
             results, num_pages, num_results = search_controller.search(
                 params,
                 search_index,
+                exact_index,
                 page_size,
                 hashed_ip,
                 request,
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -13,7 +13,7 @@ from api.utils import photon
 from api.utils.pagination import StandardPagination
 
 
-parent_logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
 
 class MediaViewSet(ReadOnlyModelViewSet):
@@ -85,7 +85,16 @@ class MediaViewSet(ReadOnlyModelViewSet):
         qa = params.validated_data["qa"]
         filter_dead = params.validated_data["filter_dead"]
 
-        search_index = self.qa_index if qa else self.default_index
+        if qa:
+            logger.info("Using QA index for media.")
+            search_index = self.qa_index
+        elif pref_index := params.validated_data.get("index"):
+            logger.info(f"Using preferred index {pref_index} for media.")
+            search_index = pref_index
+        else:
+            logger.info("Using default index for media.")
+            search_index = self.default_index
+
         try:
             results, num_pages, num_results = search_controller.search(
                 params,
--- a/api/test/unit/serializers/test_media_serializers.py
+++ b/api/test/unit/serializers/test_media_serializers.py
@@ -1,6 +1,6 @@
 import uuid
 from test.factory.models.oauth2 import AccessTokenFactory
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 from django.conf import settings
 from rest_framework.exceptions import NotAuthenticated, ValidationError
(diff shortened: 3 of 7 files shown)


zackkrida:
@krysal given the priority of this PR and @sarayourfriend's AFK, I've added you as a reviewer to this PR. Thank you!

stacimc (changes requested):
If `ENABLE_FILTERED_INDEX_QUERIES` is enabled (and neither `mature` or `unstable__include_sensitive_results` are set to `True`), [we append `-filtered`](https://github.com/WordPress/openverse/pull/2073/files#diff-8429482fb10e7f976ad2295c754632b960e3edc0dae9edfa29504cbabf24d967L281) to the index name before searching. If we're using `internal__index` to point to an index which does not have a corresponding filtered index, this will raise an `index_not_found_exception`.

I think we could just update `_resolve_index` to not use the filtered index if an `internal__index` param was supplied - what do you think?

______________________________________
Other than that, I think this looks good! Some notes for other testers as I had some trouble with a few of the steps:

I created a new `image-foo` index using:
```
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "INGEST_UPSTREAM", "index_suffix": "foo"}' 'http://localhost:50281/task'
```

I was note sure how to delete documents in Elasticvue. I got the document id of one of my search results from elasticvue (`1879` in my case) and then deleted it on the image-foo index by:
```
curl -X DELETE "localhost:50292/image-foo/_doc/1879"
```

It still was not working so I had to refresh the `image-foo` index in Elasticvue. Then I was able to see that the deleted document was absent from my query against the `image_foo` index, but not against the default.

api/api/views/media_views.py
```diff
@@ -88,17 +88,21 @@ def list(self, request, *_, **__):
         if qa:
             logger.info("Using QA index for media.")
             search_index = self.qa_index
+            exact_index = False
```
stacimc, line 91: I'm not familiar with what the `qa_index` is for so I could be wrong, but I don't think we have a `-filtered` version of this either.

dhruvkb: The `qa` index is mostly defunct, it has like 4 handwritten entries. We should probably deprecate it entirely and just remove all references pointing to it.

stacimc: Sounds reasonable to me!

stacimc (approved):

api/api/serializers/media_serializers.py
```diff
@@ -260,6 +272,23 @@ def validate_unstable__include_sensitive_results(
...
+        for unauthenticated requests, no check is performed and ``None`` is
+        returned immediately.
+
+        :param value: the provided index name to check
+        :return: ``None`` if request is anonymous, the provided name if it is valid
+        :raise: ``serializers.ValidationError`` if not anonymous and invalid index name
+        """
+
+        if self.is_request_anonymous():
+            return None
```
sarayourfriend, line 287: Should this be restricted to a specific OAuth scope, one that people cannot get by merely registering a plain API key?

sarayourfriend (changes requested):
I haven't run this locally but the code LGTM (I'll test it now), but I think we should scope this feature to a specific OAuth scope rather than _any_ authed request.

sarayourfriend:
Code works well for me, so the only requested change is to either confirm that we absolutely want to allow anyone with any API key to access this feature (I am sceptical that that is what we want).

@stacimc here are some tips based on my testing of this PR that might be useful for the future, especially for Elasticvue:

- `just ingestion_server/ingest-upstream` can be used to make the call to the ingestion server easier. I used `just ingestion_server/ingest-upstream image index_param_test` for example.
- You can use Elasticsearch's [delete by query API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html) to delete a document based on the Openverse identifier to avoid needing to dig up the document ID
- Elasticvue has a "REST" tab at the top that makes it easier to send HTTP requests to Elasticsearch without needing to juggle JSON formatting in cURL. I sent this request, for example, to delete the first result on a queryless image request locally:
```py
"""
POST /image-index_param_test/_delete-by-query

{
	"query": {
		"match": {
			"identifier.keyword": "379c8a05-1aca-4256-8ab3-59c5d801e394"
		}
	}
}
```

dhruvkb:
Drafting to address review changes.

dhruvkb:
@sarayourfriend I need a bit more information about your comment regarding scopes. I don't understand how the throttling-scopes could be used for authentication and limiting access to a parameter.

sarayourfriend:
> I don't understand how the throttling-scopes could be used for authentication and limiting access to a parameter.

I'm not referring to the throttle scopes (per-se) but the ThrottledApplication rate-limit models: 

https://github.com//WordPress/openverse/blob/098925afa90b26881b4b4d3e2baffa69049692c0/api/api/models/oauth.py#L37-L39

We could reuse those (which would unfortunately overload the meaning). Though after thinking about it more, I think we should use User permissions on the Django User model, that's the "correct" way to do these sorts of things.

`is_request_anonymous` will return `False` as long as any token is included on the request, but for kind of a weird reason. If the request is authenticated, there is no `user` on the request. `request.user` is None. If the request has no token, `request.user` is `AnonymousUser`. That was surprising to me. It turns out we don't have our OAuth clients wired up to Django Users, which now that I think about it makes sense. If we used the Django permissions approach, we'd need to wire them up, which would take extra time, even though it is the "correct" way to manage access control for individual features in Django (theoretically). Or we could do the quick and dirty approach and add a new rate limit model for them, but again, that significantly overloads what "rate limit model" is for.

My suggestion would be to put this PR on hold and for us to have a high priority issue to resolve what we mean by "walled behind auth" and to either confirm this is the behaviour we expect (anyone with any token) or whether we want to be able to restrict it further. If it's anyone with any token then I am kind of sceptical for why it's even needed in the first place. Anyone who wanted to abuse these parameters would be able to do so trivially. What are we getting by hiding them behind authentication at this level?

However, it's apparent to me that this isn't a question that affects only this PR. `is_request_anonymous` was added as part of other work and other features exist behind it that have the same problem. With that in mind, we could also merge this as is but with the stipulation that we'd need to urgently re-evaluate our "walled behind auth" idea. As I said above, either we should confirm that we really want this effectively useless guard on the feature (or else define why this is useful, I can't come up with a reason other than to make using it tedious) or decide on a more specific approach. I'm not familiar with the urgency of this PR though to say whether it's worth merging this before addressing what we mean by "walled behind auth". If we need this tomorrow for some reason, then I won't block on this. Otherwise, if isn't needed ASAP and we can take ~1 or 2 weeks to confirm or take a better approach, it's worth it.

Basically, all this boils down to:

- What do we mean by "walled behind auth"?
- Does that match the current behaviour (any request with any token)?
- If so, what is the point of that restriction? At best, as far as I can tell, it makes using the feature tedious for us as maintainers and does nothing to prevent it from being misused by a malicious actor or abused otherwise.
- If not, then we should change how we manage access to these features in a high or even critical priority issue (considering there are features live in production using this same approach).
- Likewise (if we need to make changes to the "walled behind auth" approach), do we need this feature ASAP? If not, it can be put on hold for the fix and never be introduced in a way that could cause someone to attack us by using it maliciously to cause errors if there are any unexpected behaviours caused by this.

dhruvkb:
@sarayourfriend this PR (or more accurately, the issue this PR addresses) was discussed as something to implement in the data refresh process when we did the postmortem for the outage. The milestone has the overall priority of "🟧 high".

As for auth, I do feel it's worth discussing a proper mechanism for authentication rather than trying to rush this PR out the door.

When working on the PR, the main intention for auth, in my mind, was only to rate-limit and have a log of users accessing this endpoint. I did not intend for this endpoint to only be something that staff can access, especially because the exact name of any yet-to-be-promoted index would not be known to any party other than the ingestion server and the data-refresh DAG.

sarayourfriend (approved):
In that case then I don't think there's any issue. If the goal is merely to have a log of who is using it and how, then the code covers that sufficiently. It would be worth documenting that somewhere. If in the future we want further levels of restriction, we can implement it then.

[The content moderation process project](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#volunteer-moderator-django-admin-tools-and-access-control-requirement-2) includes work to wire up certain Django permissions stuff, so we'll naturally have at least some groundwork covered by then. It will make it easier to solve new questions of access control when they arise.

api/test/unit/serializers/test_media_serializers.py
Saschaym, line 1: index_param

Saschaym (commented):
index_param

PR #2086, opened 2023-05-12
Check the file type of the thumbnail before passing it to Photon

### LINKED ISSUE #2067: Do not send `svg` images to photon for thumbnails
## Problem

Photon, the service we use to create thumbnails, only processes gif, png, webp, jpeg, and jpg images[^1]. When we send the svg images to it, it returns responds with 400 status.

Try accessing the thumbnail for an svg file from wikimedia:
https://i0.wp.com/upload.wikimedia.org/wikipedia/commons/a/a8/20210420_Gun_control_survey_by_political_party_-_Pew_Research.svg?w=600&quality=80&ssl=true
https://upload.wikimedia.org/wikipedia/commons/a/a8/20210420_Gun_control_survey_by_political_party_-_Pew_Research.svg

## Description

We should use the svg files directly as thumbnails instead of passing them through photon.

## Alternatives

If we do not use a proxy for svg thumbnails, we don't get them cached. Should we have some kind of caching mechanism specifically for svg files?

## Additional context

There shouldn't be a need to send svg files anyways as they are usually smaller and don't need to be compressed.

[^1]: https://developer.wordpress.com/docs/photon/

krysal:
## Fixes

Fixes #2067 by @krysal

## Description

To prevent known errors from Photon due to unsupported file types, this PR adds the following check before making the call to Photon:
1. Get the thumbnail extension from the URL. We can't assume it'd be the same of the main image (in the case of `images/`), so try to guess it first from the URL. I believe it is worth trying this route, since this is where most of the cases will fall.
2. If can't guess the extension from the URL then make a HEAD request to get the content-type of the file, caching the result
3. If the media type is not supported by Photon then raise an exception: [415 Unsupported Media Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415).

## Testing Instructions

If you want to manually check, try replacing the first row (after the header) of the `sample_data/sample_image.csv` file with the following:

```
0e3315c5-3328-4a99-80ab-567ac32f685f,2022-12-21 17:29:54.000000+00,2022-12-21 17:29:54.000000+00,provider_api,wikimedia,wikimedia,,https://www.flickr.com/photos/54633257@N04/51745822704,https://upload.wikimedia.org/wikipedia/commons/9/92/Flower_sewing_pattern_outline.svg,,433,1024,97150,by-sa,2.0,Jooja,https://commons.wikimedia.org/wiki/User:Jooja,Flower sewing pattern outline,"{""views"": ""5991"", ""pub_date"": ""1639443671"", ""date_taken"": ""2021-07-20 18:19:25"", ""description"": ""dressed"", ""license_url"": ""https://creativecommons.org/licenses/by-sa/3.0/"", ""raw_license_url"": null}",,f,2021-12-15 20:49:19.976193+00,f,svg,illustration
```

The run `just api/recreate` and visit http://localhost:50280/v1/images/0e3315c5-3328-4a99-80ab-567ac32f685f/thumb/

You should see the new response message 'Unsupported media type \"svg\" in request.'

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/photon.py
+++ b/api/api/utils/photon.py
@@ -14,6 +14,8 @@ import sentry_sdk
 
 parent_logger = logging.getLogger(__name__)
 
+cache = django_redis.get_redis_connection("default")
+
 
 class UpstreamThumbnailException(APIException):
     status_code = status.HTTP_424_FAILED_DEPENDENCY
@@ -39,7 +41,6 @@ def _get_file_extension_from_url(image_url: str) -> str:
 
 def check_image_type(image_url: str, media_obj) -> None:
     key = f"media:{media_obj.identifier}:thumb_type"
-    cache = django_redis.get_redis_connection("default")
 
     ext = _get_file_extension_from_url(image_url)
 
@@ -128,7 +129,6 @@ def get(
     except requests.ReadTimeout as exc:
         # Count the incident so that we can identify providers with most timeouts.
         key = f"{settings.THUMBNAIL_TIMEOUT_PREFIX}{domain}"
-        cache = django_redis.get_redis_connection("default")
         try:
             cache.incr(key)
         except ValueError:  # Key does not exist.
@@ -138,6 +138,19 @@ def get(
         raise UpstreamThumbnailException(
             f"Failed to render thumbnail due to timeout: {exc}"
         )
+    except requests.HTTPError as exc:
+        # Save occurrences to redis and only sent to Sentry when reached a
+        # threshold to prevent providers downtime consuming all the events
+        # quota.
+        key = f"{settings.THUMBNAIL_HTTP_ERROR_PREFIX}{domain}"
+        errors = 1
+        try:
+            errors = cache.incr(key)
+        except ValueError:  # Key does not exist.
+            cache.set(key, 1)
+        if errors >= settings.THUMBNAIL_HTTP_ERROR_THRESHOLD_TO_NOTIFY:
+            sentry_sdk.capture_exception(exc)
+        raise UpstreamThumbnailException(f"Failed to render thumbnail. {exc}")
     except requests.RequestException as exc:
         sentry_sdk.capture_exception(exc)
         raise UpstreamThumbnailException(f"Failed to render thumbnail. {exc}")
--- a/api/api/utils/photon.py
+++ b/api/api/utils/photon.py
@@ -144,9 +144,9 @@ def get(
             f"Failed to render thumbnail due to timeout: {exc}"
         )
     except requests.HTTPError as exc:
-        # Save number of occurrences to redis. Only raise a Sentry error when the number of
-        # occurrences reaches the configured threshold, to prevent provider API downtime from
-        # causing excessive Sentry events.
+        # Save number of occurrences to redis. Only raise a Sentry error when
+        # the number of occurrences reaches the configured threshold, to
+        # prevent provider API downtime from causing excessive Sentry events.
         key = f"{settings.THUMBNAIL_HTTP_ERROR_PREFIX}{domain}"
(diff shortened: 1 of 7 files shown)


api/api/utils/photon.py
```diff
@@ -20,13 +21,51 @@ class UpstreamThumbnailException(APIException):
...
+            response = requests.head(image_url)
+            response.raise_for_status()
+        except Exception as exc:
+            sentry_sdk.capture_exception(exc)
+            raise UpstreamThumbnailException(
+                "Failed to render thumbnail due to inability to check media "
+                f"type. {exc}"
+            )
+        else:
+            ext = response.headers["Content-Type"].split("/")[1]
```
AetherUnbound, line 62: Would there ever be a case where `Content-Type` would not be in the `headers` of the response? Would we want to add some resiliency around that case?

api/api/utils/photon.py
```diff
@@ -99,6 +139,19 @@ def get(
...
+        # quota.
+        key = f"{settings.THUMBNAIL_HTTP_ERROR_PREFIX}{domain}"
+        errors = 1
+        try:
+            errors = cache.incr(key)
+        except ValueError:  # Key does not exist.
+            cache.set(key, 1)
+        if errors >= settings.THUMBNAIL_HTTP_ERROR_THRESHOLD_TO_NOTIFY:
+            sentry_sdk.capture_exception(exc)
+        raise UpstreamThumbnailException(f"Failed to render thumbnail. {exc}")
```
AetherUnbound, line 154: Great idea planning for sentry throttling in this manner! Do you think it would make sense to add a test to simulate this case and assert that `sentry_sdk.capture_exception` isn't called until the threshold is hit?

api/api/utils/photon.py
```diff
@@ -99,6 +139,19 @@ def get(
         raise UpstreamThumbnailException(
             f"Failed to render thumbnail due to timeout: {exc}"
         )
+    except requests.HTTPError as exc:
+        # Save occurrences to redis and only sent to Sentry when reached a
+        # threshold to prevent providers downtime consuming all the events
+        # quota.
```
stacimc, line 145:
```suggestion
        # Save number of occurrences to redis. Only raise a Sentry error when the number of
        # occurrences reaches the configured threshold, to prevent provider API downtime from
        # causing excessive Sentry events.
```

AetherUnbound (approved):
The code and tests look good, great call on adding caching at various levels! I have a few suggestions/questions but nothing blocking 🚀 🖼️

stacimc (approved):
LGTM and tested well! I appreciated the pro-active work to reduce Sentry messages 🚀

AetherUnbound:
Oh, one other thing I was curious about! I know we do the dead link filtering with using `async` to parallelize requests: https://github.com/WordPress/openverse/blob/098925afa90b26881b4b4d3e2baffa69049692c0/api/api/utils/check_dead_links/__init__.py#L46

I _think_ that's because they're all happening as part of one request, whereas queries for the thumbnail are each their own request, right? That to say, do we need to try and parallelize this or is it okay to make the `HEAD` request as part of the request since all thumbnail requests will be issued separately, at the same time?

api/api/utils/photon.py
krysal: Added. Thanks to @sarayourfriend that made all the setup a breeze to built this test ⭐️

AetherUnbound:
Got it, just got myself mixed up and had to talk through it to understand - thanks @krysal!

PR #2106, opened 2023-05-15
Skip the create filtered index steps that expect an index to already exist

### LINKED ISSUE #2105: Filtered index `get_current_index` receives a `None` value
## Airflow log link

> **Note**: _Airflow is currently only accessible to maintainers & those given
> access. If you would like access to Airflow, please reach out to a member of
> @WordPress/openverse-maintainers_.

https://airflow.openverse.engineering/log?execution_date=2023-05-15T18%3A42%3A21.244399%2B00%3A00&task_id=get_current_index&dag_id=create_filtered_audio_index&map_index=-1

## Description

```
[2023-05-15, 18:52:28 UTC] {taskinstance.py:1776} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/providers/http/operators/http.py", line 134, in execute
    return self.response_filter(response, **kwargs)
  File "/opt/airflow/catalog/dags/common/ingestion_server.py", line 30, in response_filter_stat
    _, index_suffix = index_name.split("-", maxsplit=1)
AttributeError: 'NoneType' object has no attribute 'split'
```

## Reproduction

This occurred across several retries.

## DAG status

I have paused the DAG while we create a fix.

AetherUnbound:
## Fixes

Fixes #2105 by @AetherUnbound

## Description

This PR fixes an issue with the `create_filtered_<media>_index` DAG wherein it expects that the filtered index will already exist when it is first run. I've changed the configuration of the DAG so that the `get_current_index` step can skip (along with skipping the `trigger_delete_index` step) without skipping any of the rest of the DAG. I did this by using an `EmptyOperator` to manage the control flow - due to the way this DAG is constructed and how the functions are reused elsewhere, I didn't want to have to add `trigger_rule` at every level within the various function calls.

I also made a slight tweak to one of the `ingestion_server.py` logic to raise a `ValueError` instead of an `AirflowException`, since the latter has some airflow-specific meaning that we want to avoid overloading.

**Before**
image

**After**

image

<details> <summary> Previous iteration </summary>

image

</details>

## Testing Instructions

1. `just down -v`
2. Apply the following diff to simulate the lack of a filtered audio index:

```diff
diff --git a/load_sample_data.sh b/load_sample_data.sh
index 07a157382..60c74faa5 100755
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -74,10 +74,10 @@ just ingestion_server/ingest-upstream "audio" "init"
 just docker/es/wait-for-index "audio-init"
 just ingestion_server/promote "audio" "init" "audio"
 just docker/es/wait-for-index "audio"
-just ingestion_server/create-and-populate-filtered-index "audio" "init"
-just docker/es/wait-for-index "audio-init-filtered"
-just ingestion_server/point-alias "audio" "init-filtered" "audio-filtered"
-just docker/es/wait-for-index "audio-filtered" "audio-init-filtered"
+#just ingestion_server/create-and-populate-filtered-index "audio" "init"
+#just docker/es/wait-for-index "audio-init-filtered"
+#just ingestion_server/point-alias "audio" "init-filtered" "audio-filtered"
+#just docker/es/wait-for-index "audio-filtered" "audio-init-filtered"
```

3. `j api/init`
4. `j down && j ingestion_server/up && j catalog/up`
5. Trigger the `create_filtered_audio_index` DAG locally **with config** with `init` as the origin and destination index.
6. Observe that the `get_current_index` and `trigger_delete_index` steps skip, but the rest of the DAG processes as expected.
7. Run another audio data refresh and observe that in a second run of the `create_filtered_audio_index` which gets triggered by the data refresh, it runs all steps.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -3,7 +3,7 @@ import os
 from datetime import timedelta
 from urllib.parse import urlparse
 
-from airflow.exceptions import AirflowException
+from airflow.exceptions import AirflowSkipException
 from airflow.providers.http.operators.http import SimpleHttpOperator
 from airflow.providers.http.sensors.http import HttpSensor
 from requests import Response
@@ -24,7 +24,10 @@ def response_filter_stat(response: Response) -> str:
     This is used to extract the name of the current index that the concerned alias
     points to. This index name will be available via XCom in the downstream tasks.
     """
-    index_name = response.json()["alt_names"]
+    data = response.json()
+    if not (data["exists"]):
+        raise AirflowSkipException("Index does not exist.")
+    index_name = data["alt_names"]
     # Indices are named as '<media type>-<suffix>', so everything after the first
     # hyphen '-' is the suffix.
     _, index_suffix = index_name.split("-", maxsplit=1)
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -56,9 +56,7 @@ def response_check_wait_for_completion(response: Response) -> bool:
         return False
 
     if data["error"]:
-        raise AirflowException(
-            "Ingestion server encountered an error during data refresh."
-        )
+        raise ValueError("Ingestion server encountered an error during data refresh.")
 
     logger.info(f"Data refresh done with {data['progress']}% completed.")
     return True
--- a/catalog/dags/data_refresh/create_filtered_index_dag.py
+++ b/catalog/dags/data_refresh/create_filtered_index_dag.py
@@ -57,7 +57,9 @@ from airflow import DAG
 from airflow.decorators import task
 from airflow.exceptions import AirflowSensorTimeout
 from airflow.models.param import Param
+from airflow.operators.empty import EmptyOperator
 from airflow.sensors.external_task import ExternalTaskSensor
+from airflow.utils.trigger_rule import TriggerRule
 
 from common import ingestion_server
 from common.constants import DAG_DEFAULT_ARGS, XCOM_PULL_TEMPLATE
@@ -199,7 +201,7 @@ def filtered_index_creation_dag_factory(data_refresh: DataRefresh):
 
         get_current_index = ingestion_server.get_current_index(target_alias)
 
-        do_create = create_and_populate_filtered_index(
+        do_create, await_create = create_and_populate_filtered_index(
             origin_index_suffix="{{ params.origin_index_suffix }}",
             destination_index_suffix=destination_index_suffix,
         )
@@ -218,12 +220,18 @@ def filtered_index_creation_dag_factory(data_refresh: DataRefresh):
 
         (
             prevent_concurrency
-            >> destination_index_suffix
-            >> get_current_index
+            >> [destination_index_suffix, get_current_index]
+            # Gather step which will run even if the get current index step is skipped
+            >> EmptyOperator(
+                task_id="pre_create_check",
+                trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS,
+            )
             >> do_create
-            >> do_point_alias
-            >> delete_old_index
         )
+
+        await_create >> do_point_alias
+
+        [get_current_index, do_point_alias] >> delete_old_index
 
     return dag
 
--- a/catalog/dags/data_refresh/create_filtered_index_dag.py
+++ b/catalog/dags/data_refresh/create_filtered_index_dag.py
@@ -199,7 +199,7 @@ def filtered_index_creation_dag_factory(data_refresh: DataRefresh):
             "{{ params.destination_index_suffix }}"
         )
 
-        get_current_index = ingestion_server.get_current_index(target_alias)
+        get_current_index_if_exists = ingestion_server.get_current_index(target_alias)
 
         do_create, await_create = create_and_populate_filtered_index(
             origin_index_suffix="{{ params.origin_index_suffix }}",
@@ -213,7 +213,7 @@ def filtered_index_creation_dag_factory(data_refresh: DataRefresh):
             model=data_refresh.media_type,
             data={
                 "index_suffix": XCOM_PULL_TEMPLATE.format(
-                    get_current_index.task_id, "return_value"
+                    get_current_index_if_exists.task_id, "return_value"
                 ),
             },
         )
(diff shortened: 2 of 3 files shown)


catalog/dags/common/ingestion_server.py
```diff
@@ -56,9 +59,7 @@ def response_check_wait_for_completion(response: Response) -> bool:
         return False
 
     if data["error"]:
-        raise AirflowException(
-            "Ingestion server encountered an error during data refresh."
-        )
+        raise ValueError("Ingestion server encountered an error during data refresh.")
```
stacimc, line 62: Good call 👍

stacimc (approved):
This looks great! I tested with audio using the test instructions, and then also tested with image by:
* Dropping the filtered indices for image
* Running a DAG locally to get some image data
* Enabling the `create_filtered_image_index` DAG locally
* Running an `image_data_refresh`, and then verifying that the `create_filtered_image_index` DAG skipped `get_current_index` and `trigger_delete_index`, and that a filtered index was created correctly
* Running a second `image_data_refresh` and ensuring that this also worked, and the original filtered index was deleted

The control flow of the DAG is even easier to understand here. LGTM 👍

AetherUnbound:
(doing some auto-label testing apologies for the timeline spam)

obulat (commented):
I tried testing locally, but I think my local setup is incorrect. The `get_current_index` task fails with `requests.exceptions.ConnectionError: HTTPConnectionPool(host='172.17.0.1', port=8001): Max retries exceeded with url: /stat/audio-filtered (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xffff95774700>: Failed to establish a new connection: [Errno 111] Connection refused'))`
I tried running it per the testing instructions, and also without the `config`, it always fails with this error.

AetherUnbound:
@obulat do you mind running `just build` and trying again? Also are there any logs from `just logs ingestion_server` that might indicate what's going wrong?

catalog/dags/common/ingestion_server.py
```diff
@@ -24,7 +24,10 @@ def response_filter_stat(response: Response) -> str:
     This is used to extract the name of the current index that the concerned alias
     points to. This index name will be available via XCom in the downstream tasks.
     """
-    index_name = response.json()["alt_names"]
+    data = response.json()
+    if not (data["exists"]):
+        raise AirflowSkipException("Index does not exist.")
```
sarayourfriend, line 29: Does this exception cause the entire DAG to skip or are DAGs able to handle skipped tasks individually? If it cannot be individually handled by DAGs (in order to continue if there are expected cases where the index does not exist) then returning None might be better? I don't know if that makes sense in an Airflow context though, so not blocking on this.

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -218,13 +220,19 @@ def point_alias(destination_index_suffix: str):
...
         (
             prevent_concurrency
-            >> destination_index_suffix
-            >> get_current_index
+            >> [destination_index_suffix, get_current_index]
+            # Gather step which will run even if the get current index step is skipped
+            >> EmptyOperator(
+                task_id="pre_create_check",
+                trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS,
+            )
```
sarayourfriend, line 228: Because these two lines are kind of their own unit, would you mind moving them into a helper function or their own variable with a clearer explanatory comment? Something along the lines of "`get_current_index` can fail if the filtered index does not exist (i.e., on first runs) so we'll only require `destination_index_suffix` to succeed in order to continue with the DAG" would be very helpful. It might just be due to my lack of Airflow experience but putting the `EmptyOperator` inline doesn't make the relationship clear between it and the previous group. A change to the task ID to relate it to the task group in question (rather than the subsequent task) would also help with clarity.

sarayourfriend (changes requested):
Just one small requested change for clarity, but please feel free to dismiss and re-request a review if you think it isn't necessary (i.e., if it is perfectly clear as-is but just not to me due to lack of Airflow experience).

catalog/dags/common/ingestion_server.py
AetherUnbound: Normally, skips do propagate downstream! However, Airflow has a few [trigger rules](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#concepts-trigger-rules) which allow you to control the flow of skips and determine how/if some steps should run due to upstream failures or skips. You can see from the "**After**" screenshot in the description what a skipped run looks like (only the `get_current_index` and `trigger_delete_index` get skipped).

A skip is preferable here because:
- It's more visible from the UI that this and affected steps _did not run_ when they get skipped (since they're a different color). That makes it clear at-a-glance what steps actually did run. We had a case earlier (I looked but couldn't find the exact change) where loading steps were completing successfully even if they didn't load anything. It was really hard to tell when those cases where occurring amidst dozens of runs when all of them were green in the grid UI. We changed that to skip and it was immediately much clearer!
- Skipped tasks do not run, so in this case we wouldn't need to add any extra logic to the downstream tasks for how to appropriately handle a `None` value. We can run on the assumption that if a task gets run, it should be receiving a valid value.

catalog/dags/data_refresh/create_filtered_index_dag.py
AetherUnbound: Sure! I'll see if there's a better way I can aggregate/separate them so it's clearer what's going on here.

AetherUnbound: Okay, I've reworked this a bit in 9d62ac9030e700d7435069d0397f9450c8908d6c, let me know if that's clearer! Also, if you have any suggestions for the task ID or the name for improved clarity, I'm open to them 🙂 I've also updated the PR description to reflect the new DAG.

catalog/dags/common/ingestion_server.py
sarayourfriend: Thanks for the explanation. Skip makes sense, especially after learning about the "gather" step that comes after this one.

sarayourfriend (approved):
LGTM. The changes actually do not ultimately improve the clarity to me but I am struggling to find suggestions for better names. I think `get_current_index` might be part of the issue. Potentially this would be clearer?

```
get_current_index_if_exists >> continue_if_no_current_index >> etc...
```

"ensure_continue" is essentially meaningless to me, primarily because there's no indication for what we're ensuring a continuation in the face of.

In any case, I don't want to hold this up over variable names so feel free to make changes to them if you like my suggestion or have different ideas that would be similarly clarifying, or just merge as is.

AetherUnbound:
Those are WAY better names, thank you! I'll make one more change, then merge this 🚀

Issue #1965, opened 2023-05-02
Implementation Plan: Copy updates `mature` -> `sensitive`


sarayourfriend:
## Description

Project proposal: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#update-copy-to-use-the-more-general-sensitive-language-requirement-1

Write an implementation plan to update copy from `mature` to `sensitive` in the frontend and Django API. To reiterate what is said in the project proposal, there should be no design changes or new features as part of this plan, only changes to copy. Field and model names should also be updated when possible, but without the need for database migrations (i.e., use [Django's built in tools for specifying different underlying names for fields and tables](https://docs.djangoproject.com/en/4.2/ref/models/options/#table-names)).

PR #2118, opened 2023-05-17
Implementation plan: Blurring sensitive content in the frontend

dhruvkb:
## Fixes

Fixes #938 by @sarayourfriend 

## Description

This PR adds an implementation plan for blurring sensitive content in the frontend. You can read it here: https://docs.openverse.org/_preview/2118/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230506-implementation_plan_frontend_blurring_sensitive.html

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #2126, opened 2023-05-18
Implementation Plan: Copy updates `mature` -> `sensitive`

AetherUnbound:
## Due date:

2023-06-08

## Assigned reviewers

- [ ] @krysal (for API expertise)
- [ ] @dhruvkb (for frontend expertise)

## Description

Resolves: #1965

This is the implementation plan for copy/code changes necessary for updating references from `mature` to `sensitive` for various aspects of the frontend and API.

Admittedly I have very little familiarity with the frontend and translations specifically, so my understanding there may be inaccurate. Any guidance in that area would be welcome!

## Current round

The deadline for review of this round is **2023-06-08**

### DIFF
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
@@ -27,9 +27,13 @@ This implementation plan concerns the copy changes necessary for updating
 references of `mature` content to `sensitive` content. These changes will be
 made to the frontend and Django API, both in code and in web-facing copy. There
 should be no design changes or new features as part of this plan, only changes
-to copy/text itself. Field and model names for the API should also be updated
-when possible, but without the need for database migrations (i.e., using
+to copy/text itself. The initial approach for this effort was to only update
+field and model names for the API without the need for database migrations
+(i.e., using
 [Django's built-in tools for specifying different underlying names for fields and tables](https://docs.djangoproject.com/en/4.2/ref/models/options/#table-names)).
+After some discussion however, we decided that now would be an ideal time to
+perform the migrations since the process will only become more difficult and
+take longer as the tables/columns grow.
 
 ## Expected Outcomes
 
@@ -52,9 +56,7 @@ are broken into two aspects (code and copy).
 
 #### Code
 
-The following models will need to be updated. Since we want to avoid a migration
-with this work, they will need to
-[explicitly reference the old table names using `Meta::db_table`](https://docs.djangoproject.com/en/4.2/ref/models/options/#table-names).
+The following model names will need to be updated:
 
 - [`AbstractMatureMedia`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/media.py#L334-L333)
 - [`MatureImage`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/image.py#L83)
@@ -65,9 +67,52 @@ The
 (and
 [`mature_filtered`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/media.py#L18-L17)
 value) in `AbstractMediaReport` definition will also need to be changed to
-`sensitive` and `sensitive_filtered` respectively. In order to remain backwards
-compatible with the frontend, incoming references on reports for `mature` will
-need to be mapped to the new value.
+`sensitive` and `sensitive_filtered` respectively. The migration necessary for
+changing the data is described below, however the API will need to continue to
+accept `mature` as a report reason and convert it internally to `sensitive` (at
+least until we decide to make a version change to the API). The media report
+tables already use table aliases as well
+([`ImageReport`](https://github.com/WordPress/openverse/blob/7b95a4c8eaa9804f53b4be7ac969e04ca437695a/api/api/models/image.py#L121-L122)
+and
+[`AudioReport`](https://github.com/WordPress/openverse/blob/7b95a4c8eaa9804f53b4be7ac969e04ca437695a/api/api/models/audio.py#L304-L305)) -
+since we're already performing table renames as part of this IP, it makes sense
+to rename these tables for the same reason.
+
+In order to perform both of these changes in a
+[zero-downtime manner](https://docs.openverse.org/general/zero_downtime_database_management.html),
+the following steps will need to be taken:
+
+1. Add new models with updated names and fields for `SensitiveImage` and
+   `SensitiveAudio`. Also add new models for `ImageReport` and `AudioReport`.
+   This will require renaming code references from the existing `ImageReport`
+   and `AudioReport` models to `NsfwReport` and `NsfwReportAudio` temporarily
+   while maintaining the reference to the existing table name. Said another way:
+   - `ImageReport` -> `NsfwReport` _(code only, `db_table` should remain the
+     same)_
+   - `AudioReport` -> `NsfwReportAudio` _(code only, `db_table` should remain
+     the same)_
+   - Add `ImageReport` _(new model, `db_table` will not need to be overridden)_
+   - Add `AudioReport` _(new model, `db_table` will not need to be overridden)_
+2. Modify the API to write to both the old and new tables for each of the above
+   instances. On the new report tables, the `mature_filtered` reason should be
+   written as `sensitive_filtered`. Data should continue to be read from the
+   original tables during this time.
+3. Deploy this version of the API.
+4. Create a
+   [data management command](https://docs.openverse.org/general/zero_downtime_database_management.html#django-management-command-based-data-transformations)
+   which copies all data from the original tables to the new tables. The
+   `identifier` field can be used for determining which rows have already been
+   copied.
+5. Deploy this version of the API and run the data management command until no
+   more rows need to be copied for each of the columns.
+6. Remove all code references to the old model from the API, but leave the
+   models in place for the time being. Data should now only be written to and
+   read from the new tables.
+7. Deploy this version of the API so that the only versions currently deployed
+   are writing to the new tables.
+8. Remove the old models (`MatureImage`, `MatureAudio`, `NsfwReport`,
+   `NsfwReportAudio`) from the API codebase and deploy this version of the API.
+   This will remove the old tables from the database.
 
 Per the
 [detecting sensitive textual content project plan](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#designation-of-results-with-sensitive-terms),
@@ -78,7 +123,7 @@ purview of this IP.
 
 The Elasticsearch model property
 [`mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/ingestion_server/ingestion_server/elasticsearch_models.py#L118)
-will also not be affected.
+will also not be affected as part of this effort.
 
 #### Copy
 
@@ -92,6 +137,8 @@ updated are the `help_text` values in the following places:
 - [`AbstractMatureMedia::media_obj`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/media.py#L356)
 - [`MediaSerializer::mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/serializers/media_serializers.py#L433)
 - [`MediaSearchRequestSerializer::mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/serializers/media_serializers.py#L122)
+
+Changes to `help_text` will produce no-op migrations within Django.
 
 ### Frontend
 
@@ -147,7 +194,7 @@ parameter will be changed to `include_sensitive_results` in the API
 #### Copy
 
 Almost all copy changes necessary can be made in the
-[`en.json5` base translations file](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/frontend/src/locales/scripts/en.json5#L1-L0).
+[`en.json5` base translations file](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/frontend/src/locales/scripts/en.json5).
 The `filters.mature` key can be removed entirely, as it references a filter
 which no longer exists.
 
@@ -157,8 +204,9 @@ which no longer exists.
 
 <!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
 
-No infrastructure changes are necessary for this copy change. A migration may be
-generated, but it should be a no-op.
+This work will include several migrations for the API, which will need to be
+deployed one after another in production as described in
+[the API section](#api).
 
 ### Tools & packages
 
@@ -215,9 +263,12 @@ No accessibility changes should be necessary, as this should only affect copy.
 <!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
 
 In the case that we needed to roll back, the code and copy changes could be
-easily undone. Since we're not changing the underlying table names, a rollback
-should similarly not necessitate a migration (or at least one with SQL
-operations associated with it).
+easily undone.
+
+The API changes would be more difficult to roll back, as they would require the
+reverse of the multi-step migrations described in [the API section](#api). This
+would still be feasible, but we should be clear before moving forward that the
+changes to the table names are ones we wish to make at this juncture.
 
 ## Localization
 
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
@@ -1,0 +1,214 @@
+# 2023-05-17 Implementation Plan: Copy updates `mature` -> `sensitive`
+
+**Author**: @aetherunbound
+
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,235 @@
...
+Per the
+[detecting sensitive textual content project plan](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#designation-of-results-with-sensitive-terms),
+the `mature` field of the media serializer will remain, and will eventually have
+a `sensitivity` field added to it to encapsulate specific information about the
+reason behind a media being marked sensitive. Those changes are not under the
+purview of this IP.
+
+The Elasticsearch model property
+[`mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/ingestion_server/ingestion_server/elasticsearch_models.py#L118)
+will also not be affected.
```
krysal, line 81: Will the ES model property also change later in conjunction with the field addition to the serializer?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,235 @@
...
+The work is broken up into two services (API and frontend) and those services
+are broken into two aspects (code and copy).
+
+### API
+
+#### Code
+
+The following models will need to be updated. Since we want to avoid a migration
+with this work, they will need to
+[explicitly reference the old table names using `Meta::db_table`](https://docs.djangoproject.com/en/4.2/ref/models/options/#table-names).
```
krysal, line 57: I'll also support a version of this IP with migrations. No need to do it now but it would be a plus to keep the DB model updated too.

krysal (commented):
Thank you for writing this proposal. It looks quite detailed, so I just have a question on this round.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,235 @@
...
+- [`MatureAudio`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/audio.py#L263)
+
+The
+[`mature` reason](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/media.py#L22-L21)
+(and
+[`mature_filtered`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/api/api/models/media.py#L18-L17)
+value) in `AbstractMediaReport` definition will also need to be changed to
+`sensitive` and `sensitive_filtered` respectively. In order to remain backwards
+compatible with the frontend, incoming references on reports for `mature` will
+need to be mapped to the new value.
```
sarayourfriend, line 70: If I'm understanding the approach suggested here, this would have us update all the previous report values, right? I think that will require multiple steps including Django migrations (Django insists on encoding "choices" into migrations even though Postgres doesn't know anything about it) and (even if Django migrations aren't needed) a management command to update the old values (as per [our documentation on data transformations](https://docs.openverse.org/general/zero_downtime_database_management.html#how-to-achieve-zero-downtime-deployments)). Is my understanding of the suggestion correct? If so I think it would be worth outlining the steps for that explicitly.

FWIW, I don't think this violates the no-migrations stipulation from the project proposal as that was aimed primarily at table and column name changes rather than _data_ changes. Table and column name changes are just a lot harder to juggle and can be masked in code and UI using Django's tools. The data inside the columns, however, is easier to update and much peskier if we don't (easy to accidentally mess up queries, etc).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
sarayourfriend: It would probably be better to do sooner than later, if it's decided that that is desirable. It'll only get more tedious as the tables grow in size, right? Specifically with respect to the fact that we'd need to keep two tables for the rename to work with zero-downtime and copying the data will take longer as the tables grow. They're all pretty small (relatively speaking) at the moment, however.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: I'm not sure! I think `sensitive` is actually something we might end up deriving from a few different places in Elasticsearch, and might not be something we store on the indexed records directly. @sarayourfriend did you have a different impression?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: Agreed, if we were going to change them now would be the time. I can rework this a bit to include instructions for how to approach that.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: Oh that's very true, thank you for mentioning it! I'll add that to this IP 🙂

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
sarayourfriend: `mature` is already a derived property. It theoretically can encode if a provider supplies the fact but also gets flipped to true for confirmed user reports. If we can update it to sensitive that would be ideal. It again would take more careful planning and potentially several PRs to do it in a backwards compatible way (similar to a database column rename, really).

sarayourfriend: Although, even more ideally would be to rename it to `has_confirmed_sensitivity_report` and then when we ingest sensitivity from providers in the future to use yet another field for that. If we can make the name more explicit and prevent overloading the meaning in the future, that would be 100% better than the current ambiguous name.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,235 @@
...
+
+- [Search query transform utilities](https://github.com/WordPress/openverse/blob/92ea64c6d4ac1801785eadf99a3782628eadad06/frontend/src/utils/search-query-transform.ts#L31)
+  _(query parameter)_
+- [`Media` interface attributes](https://github.com/WordPress/openverse/blob/930a2ec4583d53e2f17e7aa397dc9b3b32cc8ebb/frontend/src/types/media.ts#L48)
+  _(data model)_
+
+#### Copy
+
+Almost all copy changes necessary can be made in the
+[`en.json5` base translations file](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/frontend/src/locales/scripts/en.json5#L1-L0).
```
dhruvkb, line 150:
```suggestion
[`en.json5` base translations file](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/frontend/src/locales/scripts/en.json5#L487-L491).
```

dhruvkb (approved):
LGTM. Apart from what seems like a typo, I have nothing more to add after what @krysal and @sarayourfriend have already covered.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,235 @@
...
+The Elasticsearch model property
+[`mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/ingestion_server/ingestion_server/elasticsearch_models.py#L118)
+will also not be affected.
+
+#### Copy
+
+Most changes necessary for updating the copy in the Django Admin UI will be
+reflected by the code changes that are made above (since the Admin UI
+introspects the code directly). The only other references that will need to be
+updated are the `help_text` values in the following places:
```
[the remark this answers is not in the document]
AetherUnbound: Right, but I think they will be migrations without SQL?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: Ah, I meant to link the whole file here, will fix!

AetherUnbound:
Thanks for the review folks! I'm moving this into the **Revision round** while I make updates to the document 🚀

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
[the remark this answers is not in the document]
sarayourfriend: Yup, they'll be noop migrations. Same with changing the choices list for a field.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: That makes sense! It seems like it might be outside the scope of this IP though, so I'll leave this as-is for now unless there are any objections.

sarayourfriend: Totally, I did not think it was something we would handle in this. Totally fine by me to continue excluding it from the scope.

AetherUnbound:
@krysal and @sarayourfriend - I've added _what I believe are_ the necessary steps for modifying the underlying table names. This seemed like a good opportunity to modify the report table names as well, since we'd also be changing the `reason` value in the table anyway (and doing this via a different table seemed easiest anyway). Clearly that breaks our intention to not have this IP include migrations :smiling: But I think now is a good time as we have so few records (<700 reports on the image table for instance)! Please let me know if any of these steps could use more clarity; it was a lot to wrap my head around for what needed to be done in which order and get it all into the IP.

@dhruvkb do you mind also taking one more look at the new proposed approach to make sure it seems okay?

Given the revisions I'm moving this into the **Decision Round**.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,286 @@
...
+No design work should be necessary.
+
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+The frontend copy work and the API copy & code work can be done simultaneously.
+The frontend code work should wait until the API work is complete so the content
+reporting using `sensitive` rather than `mature` is available on the API prior
+to the frontend code changes.
```
sarayourfriend, line 237: This probably needs to take into account the changes that will happen in #2118, primarily to note that we'll need to coordinate around that.

I don't think the frontend copy changes need to be blocked on the API work though, which seems to be what the first sentence in this section says.

Not a blocker, just clarification. I think this coordination will happen naturally as the work on the frontend copy gets done. Maybe just adding a note to the frontend issues to be aware that the other project will also be making related changes and to be aware of potential overlaps/merge conflicts that might arise.

sarayourfriend (changes requested):
One blocker, but just to add a critical step/detail to the migration to cover the ingestion server changes.

Really excited about this work, thanks for the careful planning, Madison!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,286 @@
...
+Per the
+[detecting sensitive textual content project plan](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#designation-of-results-with-sensitive-terms),
+the `mature` field of the media serializer will remain, and will eventually have
+a `sensitivity` field added to it to encapsulate specific information about the
+reason behind a media being marked sensitive. Those changes are not under the
+purview of this IP.
+
+The Elasticsearch model property
+[`mature`](https://github.com/WordPress/openverse/blob/2041c5df1e9d5d1f9f37e7c177f2e70f61ea5dba/ingestion_server/ingestion_server/elasticsearch_models.py#L118)
+will also not be affected as part of this effort.
```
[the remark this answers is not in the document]
AetherUnbound: Oh that is an extremely good call and great find, I'll be sure to add this!

AetherUnbound:
@sarayourfriend I believe I've addressed your concerns but let me know if there's anything else I'm missing! Thanks for your thorough review 🙌🏼

sarayourfriend (approved):
LGTM :rocket:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,294 @@
...
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @krysal
+- [ ] @dhruvkb
+
```
sarayourfriend, line 14:
```suggestion
- [x] @sarayourfriend

```

Might as well add my approval for the record.

sarayourfriend (approved):

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,294 @@
...
+value) in `AbstractMediaReport` definition will also need to be changed to
+`sensitive` and `sensitive_filtered` respectively. The migration necessary for
+changing the data is described below, however the API will need to continue to
+accept `mature` as a report reason and convert it internally to `sensitive` (at
+least until we decide to make a version change to the API). The media report
+tables already use table aliases as well
+([`ImageReport`](https://github.com/WordPress/openverse/blob/7b95a4c8eaa9804f53b4be7ac969e04ca437695a/api/api/models/image.py#L121-L122)
+and
+[`AudioReport`](https://github.com/WordPress/openverse/blob/7b95a4c8eaa9804f53b4be7ac969e04ca437695a/api/api/models/audio.py#L304-L305)) -
+since we're already performing table renames as part of this IP, it makes sense
```
krysal, line 78:
```suggestion
since we're already performing table renames as part of this implementation plan, it makes sense
```

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,294 @@
...
+   `identifier` field can be used for determining which rows have already been
+   copied.
+5. Deploy this version of the API and run the data management command until no
+   more rows need to be copied for each of the columns.
+6. Modify the ingestion server to use the new tables as reference for
+   determining mature status. Specifically the
+   [`exists_in_mature` table check](https://github.com/WordPress/openverse/blob/765a24028baa922b27fc0c38b8ae4c69902eec40/ingestion_server/ingestion_server/queries.py#L33-L37)
+   (this will necessarily involve updating tests and initialization SQL as
+   well).
+7. Remove all code references to the old model from the API, but leave the
```
krysal, line 113: Which model exactly are we speaking of here? Is it in the code from the Ingestion server?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
```diff
@@ -0,0 +1,294 @@
...
+9. Remove the old models (`MatureImage`, `MatureAudio`, `NsfwReport`,
+   `NsfwReportAudio`) from the API codebase and deploy this version of the API.
+   This will remove the old tables from the database.
+
+Per the
+[detecting sensitive textual content project plan](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#designation-of-results-with-sensitive-terms),
+the `mature` field of the media serializer will remain, and will eventually have
+a `sensitivity` field added to it to encapsulate specific information about the
+reason behind a media being marked sensitive. Those changes are not under the
+purview of this IP.
```
krysal, line 127:
```suggestion
purview of this implementation plan.
```

krysal (approved):
It's a pretty comprehensive plan and looks good!

I only have a question about one of the steps for the API, but it shouldn't be a blocker.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.md
AetherUnbound: This refers to the `MatureImage` and `MatureAudio` models, which are replaced by `SensitiveImage` and SensitiveAudio`. I'll make that explicit here. All the steps here deal with the API, unless explicitly stated like in step 6.

PR #2128, opened 2023-05-18
Fix sidebar height and margin calculation when banners are shown

### LINKED ISSUE #1159: The banners height needs to be considered in height calculations
## Problem

The site layout is calculated as 100vh. Then, we calculate the height of the main content as 100vh - height of the header. This is correct when the banners are not visible. However, when translation and migration banners are shown, the height overflows.

## Description

We can use a CSS variable in the banners to set their height, and then calculate the main content height using `100vh - header - banner heights` formula. The use of CSS variables with fallback to 0 will allow us to avoid using comparisons in JS.

To see the problem, go to https://openverse.org/bn/search/image?q=cat&referrer=creativecommons.org
You would see 2 banners, and if the filters sidebar is open, the top part of it is cut off.
Screenshot of the filters sidebar cut off by the banners

## Additional notes
While solving this issue, it would also be good to check the issue with banner height reported by @zackkrida in #988:
> @obulat in testing compared to production, I noticed the translation notification bar appears taller:

CleanShot 2023-04-10 at 09 55 20@2x
CleanShot 2023-04-10 at 09 54 58@2x

However, this may be related to a production issue with Arabic in particular, as I'm seeing much more English here than locally:

https://openverse.org/ar

I believe the PR is correct though and this is some unrelated production issue.

_Originally posted by @zackkrida in https://github.com/WordPress/openverse/pull/988#pullrequestreview-1377570683_

obulat:
## Fixes

Fixes #1159 by @obulat

## Description

This PR adds a `--banner-height` CSS variable that holds the height of the shown banners. It is used to set the top margin and the height of the filters sidebar to make sure that it is not covered by the header.

This PR also sets the header height to a fixed value of 5rem (`h-20`). This simplifies the calculations from the current value of 81px (because it's 80px + 1px of the border), and also makes sure that we consistently use relative values in `rem` instead of mixed `px` and `rem` values.
This change means that there are a lot of snapshot changes. I could extract them to a different PR if the reviewers find it helpful.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/layouts/content-layout.vue
```diff
@@ -121,6 +122,10 @@ export default defineComponent({
         isDesktopLayout.value
     )
 
+    const bannerHeight = computed(() => {
+      return "--banner-height: " + uiStore.shownBannersCount * 4 + "rem;"
+    })
```
dhruvkb, line 127: Styles can take an object, it's more readable.

```suggestion
    const appStyles = computed(() => {{
      "--banner-height": `${uiStore.shownBannersCount * 4}rem`
    }})
```

dhruvkb (changes requested):
I'm not 100% sure about the correctness of a hardcoded `4rem` per banner. Why not compute the height of the `VBanners` element on mounting and use that instead, that way it can continue to work with the updated designs and also allow for taller banners with text wrapper more than one lines?

obulat:
> I'm not 100% sure about the correctness of a hardcoded `4rem` per banner. Why not compute the height of the `VBanners` element on mounting and use that instead, that way it can continue to work with the updated designs and also allow for taller banners with text wrapper more than one lines?

Drafted to add the variable height.

Issue #1987, opened 2023-05-02
Implementation Plan: Staging Elasticsearch reindex DAGs for both potential index types


AetherUnbound:
## Description

This issue is for tracking the drafting of an implementation plan in correspondence with the one noted in the search relevancy sandbox project proposal: Staging Elasticsearch reindex DAGs for both potential index types (these will be subsets of the full data refresh).

From the project plan: 

> This plan will describe the DAG or DAGs which will be used to create/update both the proportional-by-provider and production-data-volume indices.

> It will also describe the mechanism by which maintainers can rapidly switch index the staging API uses. This could be done in two separate ways: a DAG which allows changing the primary index alias or a set of changes to the API which would allow queries to specify which index they use. The implementation plan should explore and describe both options.

There was some discussion on this in #1107 ([link](https://github.com/WordPress/openverse/pull/1107#pullrequestreview-1385957729)), from @zackkrida and @sarayourfriend:

> _(Zack)_
> Feels quite important to me and like it may be a bit under-explained by this proposal. As written, the proposal seems to suggest that the DAGs would be used to allow for creating a new staging index that replaces the default, "live" index the staging API is pointed at. Is that correct?
> 
> Alternatively, I can see a lot of value in having DAGs that allow for creating and updating staging indices with different qualities, but having a mechanism that allows developers to switch between them much more rapidly, perhaps even a query param-based solution (which would allow, for example, to switch indices via a checkbox on the staging.openverse.org/preferences page). 
> 
> To put this all another way, I'm wondering if requirement #5 should be a bit more specific in terms of what "easily" means.
> 
> ...
> 
> _(Madison)_
> That's a great point, but I actually think this would be a good thing to determine in the implementation plan! It was my intention to have a single alias (like production, at least at the current moment) and allow maintainers to swap between these aliases. However, I do quite like the idea of an API setting which might allow dynamically changing which index is used at query time! That would bring increased complexity to the plan and the project that I wasn't initially considering, but if we scope it out as part of the plan, potentially starting with the rapid swapping but laying out a plan for what changing the index at query time might look like, we could even add that capability down the line outside of the bounds of this specific effort!
> 
> ...
> 
> _(Sara)_
[...]

PR #2133, opened 2023-05-18
Implementation Plan: Rapid iteration on ingestion server index configuration

### LINKED ISSUE #1985: Implementation Plan: Rapid iteration on ingestion server index configuration
## Description

This issue is for tracking the drafting of an implementation plan in correspondence with the one noted in the [search relevancy sandbox project proposal](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.html#required-implementation-plans): Rapid iteration on ingestion server index configuration.

Some initial discussion on this happened in #1107 ([link](https://github.com/WordPress/openverse/pull/1107#discussion_r1155399508)), from @sarayourfriend:

> I also think we should leverage the re-index API for this. I could see an Airflow DAG meant for iterating on index configuration that works something like this:
> 
> 1. Trigger the DAG with a new index configuration body.
>     - The existing configuration can be generated using the `index_settings` function from the ingestion server project. This function runs fine in a Python shell with zero setup inside the ingestion_server project.
> 1. The DAG creates the index via the create index endpoint, either using a name provided when the DAG was triggered or by generating a UUID or date-based suffix. This should happen by hitting ES directly, not through the ingestion server.
> 1. The DAG then calls the reindex API to copy the data from the existing (periodically refreshed?) "production data" index into the new index. If we want to incorporate "proportional data" or have a subset of data, we could also have the DAG take in a query. If we just need to test a stemmer change, for example, we can query for documents that include the terms in question. For broader text analysis changes we'd want to use the full production data (see my comment above). If we do use the proportional-by-provider data, we can allow changing the origin index between the full production data and the proportional-by-provider index.
> 1. After reindex is done, refresh the new index.
> 
> Done! We've just completed one index configuration iteration.
> 
> This bypasses the complexity of the ingestion server and wouldn't require any changes to the ingestion server _at all_.
> 
> ...
> 
> A small note in the implementation plan section that we can use the reindex API directly from Airflow to ES rather than passing through the ingestion server would be sufficient. Right now the suggested approach feels heavily skewed towards needing to iterate on the ingestion server code itself and maybe even moving it to ECS, which is a lot more work that we could avoid. I think it would at least triple the scope of the implementation plan if it included moving the ingestion server to ECS vs just hitting ES directly from Airflow and using the reindex API.

AetherUnbound:
## Due date:

2023-06-07

Sara is AFK for the first few days of this review process so I've extended the clarification round to account for that.

Resolves #1985

## Assigned reviewers

- [x] @sarayourfriend (for Elasticsearch expertise and familiarity with the proposed implementation)
- [x] @stacimc (for catalog/Airflow expertise)

## Description

This PR is the implementation plan for rapid iteration on ingestion server configurations.

## Current round

[on the Openverse documentation site](https://docs.openverse.org/reference/decision_making/index.html).

The deadline for review of this round is **2023-06-07**

### DIFF
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
@@ -45,8 +45,6 @@ The following are prerequisites for the DAG:
 - Installation of the
   [`elasticsearch` Airflow provider package](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/)
 - Airflow Connections for each Elasticsearch cluster
-- Sharing of the
-  [`es_mappings.py` configuration](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
 
 ### Elasticsearch provider
 
@@ -70,26 +68,6 @@ named `elasticsearch_production` and `elasticsearch_staging` and reference each
 respective environment. These will be created by hand in the Airflow UI. For
 local development, defaults will be added to the
 [`env.template` file](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/catalog/env.template).
-
-### Sharing `es_mappings.py`
-
-The
-[`es_mappings.py` configuration](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
-in the `ingestion_server` module defines, in python, the default index
-configuration we use for all media types. This configuration does not require
-any additional Python dependencies, but does not exist in a form that can be
-imported by the DAG. To make this configuration available to the DAG, we will
-add the `es_mapping.py` file to the
-[`sync.yml` workflow configuration](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/.github/sync.yml)
-to keep the file up-to-date across both services. The destination for this sync
-could be `catalog/dags/common/elasticsearch/es_mapping.py`.
-
-It would be trivial for local development to use a symlink, but all of our
-services run within Docker containers. Unfortunately, symlinks are generally
-difficult (or impractical) to set up within Docker[^1].
-
-[^1]:
-    [Stack Overflow: Mount host directory with a symbolic link inside in docker container](https://stackoverflow.com/a/40322275/3277713)
 
 ### Infrastructure
 
@@ -154,15 +132,58 @@ Once the parameters are provided, the DAG will execute the following steps:
 
 1. Create the index configuration. If `override_config` is supplied, the
    `index_config` parameter will be used as the entire configuration. If not,
-   the `index_config` parameter will be merged with the configuration generated
-   by `es_mapping::index_settings`. See [Merging policy](#merging-policy) for
-   how this merge will be performed.
+   the `index_config` parameter will be merged with the existing index settings.
+   See [Merging policy](#merging-policy) for how this merge will be performed.
+   Below are the steps for gathering the current index settings:
+   1. Get the current index information. This will be done using the
+      [`ElasticsearchPythonHook`](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/_api/airflow/providers/elasticsearch/hooks/elasticsearch/index.html#airflow.providers.elasticsearch.hooks.elasticsearch.ElasticsearchPythonHook),
+      specifically the
+      [`index.get` function](https://elasticsearch-py.readthedocs.io/en/v8.8.0/api.html#elasticsearch.client.IndicesClient.get).
+   2. Extract the relevant settings from the response. They come back in a form
+      that differs from what is rendered by the
+      [`es_mapping.py`](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/es_mapping.py)
+      file which generates the index settings on the ingestion server.
+      `es_mapping::index_settings` returns data of the form:
+      ```json
+      {
+        "settings": {
+          "index": {
+            "number_of_shards": 18,
+            "number_of_replicas": 0,
+            "refresh_interval": "-1"
+          },
+          "analysis": {...},
+        },
+        "mapping": {...}
+      }
+      ```
+      However, the `index.get` function returns data of the form:
+      ```json
+      {
+        "<index-name>": {
+          "settings": {
+            "index": {
+              "number_of_shards": "18",
+              "number_of_replicas": 0,
+              "refresh_interval": "-1",
+              ...
+              "analysis": {
+                ...
+              }
+            }
+          },
+          "mappings": {...}
+        }
+      }
+      ```
+      The returned information will need to be extracted and converted into the
+      form returned by `es_mapping::index_settings` before it is merged with the
+      new configuration.
 2. Create the new index with the configuration generated in step 1. The index
    name will either be a combination of the `media_type` and `index_suffix`
    parameters (`{media_type}-{index_suffix}`, e.g. `image-my-special-suffix`),
    or the `media_type` and a timestamp (`{media_type}-{timestamp}`, e.g.
-   `image-20200102030405`). This will be done using the
-   [`ElasticsearchPythonHook`](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/_api/airflow/providers/elasticsearch/hooks/elasticsearch/index.html#airflow.providers.elasticsearch.hooks.elasticsearch.ElasticsearchPythonHook)
+   `image-20200102030405`).
    ([example from the existing ingestion server code](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L518-L521)).
 3. Initiate a reindex using the `source_index` as the source
    ([example from the ingestion server](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L525-L547)).
@@ -308,8 +329,7 @@ No new design work is required for this plan.
 <!-- What, if any, work within this plan can be parallelized? -->
 
 The addition of the Elasticsearch provider is a blocker for the rest of the work
-described here, sans the `es_mapping.py` sync. The sync change can occur at any
-time, all other work must be done serially.
+described here; all work must be done serially.
 
 ## Blockers
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
@@ -9,7 +9,7 @@
 
 <!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
 
-- [ ] @sarayourfriend
+- [x] @sarayourfriend
 - [ ] @stacimc
 
 ## Project links
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
@@ -23,7 +23,7 @@
 
 <!-- An overview of the implementation plan, if necessary. Save any specific steps for the section(s) below. -->
 
-This document describes a DAG (and associated machinery) which can be used to
+This document describes a DAG (and associated dependencies) which can be used to
 generate a new index in Elasticsearch without requiring a data refresh,
 ingestion server deployment, or any interaction with the ingestion server.
 
@@ -31,8 +31,8 @@ ingestion server deployment, or any interaction with the ingestion server.
 
 <!-- List any succinct expected products from this implementation plan. -->
 
-Once this implementation is complete, a maintainer should be able to run this
-DAG to create a new index on either elasticsearch environment with an altered
+Once the implementation is complete, a maintainer should be able to run this DAG
+to create a new index on either elasticsearch environment with an altered
 configuration. The maintainer should only need to specify the pieces of the
 configuration that are different from the default configuration, although they
 can supply a full configuration if they wish. The DAG will report when it is
@@ -44,7 +44,7 @@ The following are prerequisites for the DAG:
 
 - Installation of the
   [`elasticsearch` Airflow provider package](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/)
-- Airflow Connections to each Elasticsearch cluster
+- Airflow Connections for each Elasticsearch cluster
 - Sharing of the
   [`es_mappings.py` configuration](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
 
@@ -52,7 +52,7 @@ The following are prerequisites for the DAG:
 
 The
 [`elasticsearch` provider](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/)
-will need to be added to the list of dependencies for our catalog container.
+will need to be added to the list of dependencies for our catalog Docker image.
 This will be done first and deployed on its own, as the connections will be
 needed for the rest of the work.
 
@@ -81,12 +81,12 @@ any additional Python dependencies, but does not exist in a form that can be
 imported by the DAG. To make this configuration available to the DAG, we will
 add the `es_mapping.py` file to the
 [`sync.yml` workflow configuration](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/.github/sync.yml)
-to keep the file up-to-date across both services.
+to keep the file up-to-date across both services. The destination for this sync
+could be `catalog/dags/common/elasticsearch/es_mapping.py`.
 
 It would be trivial for local development to use a symlink, but all of our
-services run both locally and in production within Docker containers.
-Unfortunately, symlinks are generally difficult (or impractical) to set up
-within Docker[^1].
+services run within Docker containers. Unfortunately, symlinks are generally
+difficult (or impractical) to set up within Docker[^1].
 
 [^1]:
     [Stack Overflow: Mount host directory with a symbolic link inside in docker container](https://stackoverflow.com/a/40322275/3277713)
@@ -112,21 +112,23 @@ discussed in [issue #2070](https://github.com/WordPress/openverse/issues/2070).
 <!-- Describe the implementation step necessary for completion. -->
 
 Once the prerequisites above have been filled, the DAG can be created. This DAG
-will be named `create_new_es_index`, and will have the following parameters and
-steps.
+will be named `create_new_es_index`, and will have the parameters and steps
+described below. It will have a schedule of `None` so that it is only run when
+triggered.
 
 ### Parameters
 
 1. `media_type`: The media type for which the index is being created. Presently
    this would only be `image` or `audio`.
 2. `environment`: The Elasticsearch environment to operate against. Options are
-   `production` and `staging`.
+   `production` and `staging` (and will be mapped to the appropriate
+   Elasticsearch connection).
 3. `index_config`: A JSON object containing the configuration for the new index.
    The values in this object will be merged with the existing configuration,
    where the value specified at a leaf key in the object will override the
-   existing value. This can also be the entire index configuration, in which
-   case the existing configuration will be replaced entirely (see
-   `override_config` parameter below).
+   existing value (see [Merging policy](#merging-policy) below). This can also
+   be the entire index configuration, in which case the existing configuration
+   will be replaced entirely (see `override_config` parameter below).
 4. `index_suffix`: (Optional) The name suffix of the new index to create. This
    will be a string, and will be used to name the index in Elasticsearch of the
    form `{media_type}-{index_suffix}`. If not provided, the suffix will be a
@@ -280,7 +282,7 @@ extra dependency for this step is possible.
 
 Instead of using a self-referential sync operation, we could read or pull the
 `es_mapping.py` file directly from GitHub (e.g.
-https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/es_mapping.py).
+[https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/es_mapping.py](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/es_mapping.py)).
 This would require that GitHub be available, and (in order to ensure the most
 up-to-date version is being used) could cause issues if the file is moved to a
 different location. Since this would create unnecessary dependencies on an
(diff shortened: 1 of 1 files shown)


sarayourfriend (commented):
Looking great so far. I left a couple clarification requests and one significant alternative approach to the index settings.

Something that comes to mind is that it would be pretty nice to remove `create_and_populate_filtered_index` from the ingestion server and to re-implement it using the Elasticsearch provider after this project is done. It would get one more thing off the plate of the ingestion server responsibility (assuming we want to rely on the ingestion server less and less with the goal of eventually getting rid of it completely).

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+
+[^1]:
+    [Stack Overflow: Mount host directory with a symbolic link inside in docker container](https://stackoverflow.com/a/40322275/3277713)
+
+### Infrastructure
+
+<!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
+
+No infrastructure changes should be necessary for this implementation, beyond
+that described above.
```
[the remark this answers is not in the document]
stacimc: I assume adding the ES provider and adding the mapping to the sync. What counts as "infrastructure changes" in this section has been somewhat vaguely/broadly defined across most of the IPs, I think. Maybe the description could be updated.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+The
+[`es_mappings.py` configuration](https://github.com/WordPress/openverse/blob/0a5f4ab2ce5d80a48bd1c57d2a2dbcca14fcbedc/ingestion_server/ingestion_server/es_mapping.py)
+in the `ingestion_server` module defines, in python, the default index
+configuration we use for all media types. This configuration does not require
+any additional Python dependencies, but does not exist in a form that can be
+imported by the DAG. To make this configuration available to the DAG, we will
+add the `es_mapping.py` file to the
+[`sync.yml` workflow configuration](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/.github/sync.yml)
+to keep the file up-to-date across both services. The destination for this sync
+could be `catalog/dags/common/elasticsearch/es_mapping.py`.
```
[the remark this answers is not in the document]
stacimc: > Would we be able to avoid this by using the [get index settings API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-settings.html) to retrieve the settings on the current origin index for the media type being copied and base the new settings on that instead?

That would be very cool! And avoids the concerns described [here](https://github.com/WordPress/openverse/pull/2133/files#diff-344cb6ad9993d7e365933d71faa76a18fbbb062a333a5ae4f73b9dee861b46d8R281) as well.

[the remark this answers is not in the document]
AetherUnbound: Great idea, I'll look into this!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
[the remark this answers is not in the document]
AetherUnbound: Yes, the ES provider requires a deployment (rather than a simple DAG sync, and thus necessarily deals with the infrastructure in that way) and I consider adding the connection info a change to infrastructure too in this case. I'll clarify this section!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+described below. It will have a schedule of `None` so that it is only run when
+triggered.
+
+### Parameters
+
+1. `media_type`: The media type for which the index is being created. Presently
+   this would only be `image` or `audio`.
+2. `environment`: The Elasticsearch environment to operate against. Options are
+   `production` and `staging` (and will be mapped to the appropriate
+   Elasticsearch connection).
```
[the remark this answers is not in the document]
AetherUnbound: Managing concurrency within a single DAG is pretty straight forward, as we'd just need to set `max_active_runs` for the DAG to `1`. It wouldn't be too much extra lift to make separate staging & production DAGs, and it would probably remove an operator error vector. I'll change that!

The issue with `pools` is that pool slots are only consumed when a task on a DAG is running. The way this DAG is set up, it minimizes how long a task runs (via Sensors) to make way for other tasks in other DAGs running on Airflow. The concurrency question is definitely an important one though, I'll ponder some ways to incorporate/enforce that.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+   the existing index configuration entirely with the new configuration. If
+   `True`, the `index_config` parameter will be used as the entire
+   configuration. If `False`, the `index_config` parameter will be merged with
+   the existing configuration. Defaults to `False`.
+7. `query`: (Optional) An
+   [Elasticsearch query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html)
+   to use to filter the documents to be copied to the new index. If not
+   provided, all documents will be copied. See
+   [`create_and_populate_filtered_index` on the ingestion server](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L529)
+   for an example of how this is used.
```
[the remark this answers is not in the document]
AetherUnbound: Sure! I find the reference to our own code helpful because it comes with context I'm already familiar with. I'll go ahead and link both 🙂

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+   until the reindex is complete, this step will set `wait_for_completion=False`
+   and continue to the next step. The
+   [task ID returned from Elasticsearch for the reindex operation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-task-api)
+   will be stored in an XCom variable. This step will have `refresh=True` on the
+   request to ensure the reindex will immediately refresh after completion to
+   make the data available to searches.
+4. Use a sensor and the previously emitted task ID to wait for the reindex to
+   complete. This will check the task using the
+   [Task API](https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html).
+   This could be done either using a subsequent `wait_for_completion` call with
```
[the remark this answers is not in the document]
AetherUnbound: Ah, I was thinking a simple Python sensor since we can use the ES hook to run this query, I'll clarify this.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+   be the entire index configuration, in which case the existing configuration
+   will be replaced entirely (see `override_config` parameter below).
+4. `index_suffix`: (Optional) The name suffix of the new index to create. This
+   will be a string, and will be used to name the index in Elasticsearch of the
+   form `{media_type}-{index_suffix}`. If not provided, the suffix will be a
+   timestamp of the form `YYYYMMDDHHMMSS`.
+5. `source_index`: (Optional) The existing index on Elasticsearch to use as the
+   basis for the new index. If not provided, the index aliased to `media_type`
+   will be used (e.g. `image` for the `image` media type).
+6. `override_config`: (Optional) A boolean value which can be toggled to replace
```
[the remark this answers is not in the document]
AetherUnbound: Yea that's essentially the case I was thinking of: maybe there's a _much simpler_ config that we're trying out that we don't want to attempt any merging of. I also want to try and make sure our merging code is simple and straightforward, I could see some cases where it becomes a bulk of the complexity of this work (given there's at least one if not several packages that accomplish that specific piece available on PyPI) 😅

AetherUnbound:
Thank you @stacimc and @sarayourfriend for your review! I'm moving this into the **Revision round** while I make updates to the document 🙂

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,357 @@
...
+   how this merge will be performed.
+2. Create the new index with the configuration generated in step 1. The index
+   name will either be a combination of the `media_type` and `index_suffix`
+   parameters (`{media_type}-{index_suffix}`, e.g. `image-my-special-suffix`),
+   or the `media_type` and a timestamp (`{media_type}-{timestamp}`, e.g.
+   `image-20200102030405`). This will be done using the
+   [`ElasticsearchPythonHook`](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/_api/airflow/providers/elasticsearch/hooks/elasticsearch/index.html#airflow.providers.elasticsearch.hooks.elasticsearch.ElasticsearchPythonHook)
+   ([example from the existing ingestion server code](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L518-L521)).
+3. Initiate a reindex using the `source_index` as the source
+   ([example from the ingestion server](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L525-L547)).
```
krysal, line 168: I have a question regarding this step. While thinking about the plan for #1987, I wondered if these plans are not more closely related and dependent on each other. Since the DAG will be using the [`_reindex` API of ES](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-reindex.html), it assumes _there exists at least one index_ already, presumably the result of a data refresh. Am I correct? If so, it's worth pointing this out explicitly in the dependencies. The first time I read it, I overlooked it.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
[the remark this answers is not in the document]
AetherUnbound: Note for myself here - we can definitely use the existing API rather than a repo sync, but it will require some massaging of the response data. For instance, Current form of the `index_setting` function spits out:

```json
{
  "settings": {
    "index": {
      "number_of_shards": 18,
      "number_of_replicas": 0,
      "refresh_interval": "-1"
    },
    "analysis": {...},
  },
  "mapping": {...}
}
```

However, the response returned from `GET /<index-or-alias>/`:

```json
{
  "<index-name>": {
    "settings": {
      "index": {
        "number_of_shards": "18",
        "number_of_replicas": 0,
        "refresh_interval": "-1",
        ...
        "analysis": {
          ...
        }
      }
    },
    "mappings": {...}
  }
}
```

I'll add this to the IP, just didn't want to forget it!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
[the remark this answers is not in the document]
AetherUnbound: After spending a few hours trying to find the most appropriate solution for concurrency management (waiting on other jobs like we do for the existing reindex DAGs, querying Elasticsearch, etc), I opted for a simple solution: check if the DAGs are running and fail if they are.

AetherUnbound:
I believe I have addressed the feedback you provided @stacimc and @sarayourfriend! I've moved this into the **Decision round** as I believe it should now be ready, but if there are any blockers please feel free to mention them!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
krysal: I agree the DAG would be indifferent to which index is iterated on –although it will be quite relevant to the person triggering– still, it requires _one_ index to derive from, as I said before, due to the use of the Reindex API of ES. Which will be this very base index now that the staging DB is constantly being replaced? Perhaps this is what I should cover in the IP for the issue mentioned above but I having a hard time figuring out this part 😅

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+The
+[`elasticsearch` provider](https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/)
+will need to be added to the list of dependencies for our catalog Docker image.
+This will be done first and deployed on its own, as the connections will be
+needed for the rest of the work.
+
+The
+[production requirements](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/catalog/requirements_prod.txt#L5)
+will need to be changed from `apache-airflow[amazon,postgres,http]` to
+`apache-airflow[amazon,postgres,http,elasticsearch]`.
```
sarayourfriend, line 60: Are these two discrete steps? Can they be done at the same time? Or is the second line the directions for implementing the first bit? Sorry if this is obvious, the wording just leaves it a bit unclear :sweat_smile:

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Other projects or work
+
+<!-- Note any projects this plan is dependent on. -->
+
+This work does not depend on any other projects and can be implemented at any
+time. In order to be utilized however, a query parameter on the API will need to
+be used for determining which index to query. This has been implemented in
+specifying which index to use for a given query. This has been implemented in
+#2073.
```
sarayourfriend, line 88:
```suggestion
be used for determining which index to query. This has been implemented in
#2073.
```

Small editorial change, looks like the sentence was accidentally semi-duplicated.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+              }
+            }
+          },
+          "mappings": {...}
+        }
+      }
+      ```
+      The returned information will need to be extracted and converted into the
+      form returned by `es_mapping::index_settings` before it is merged with the
+      new configuration.
```
sarayourfriend, line 190: Just a nit, but the focus on the API response being different from `es_mapping::index_settings` rather than the expected input of the Elasticsearch create index API feels strange and potentially misleading as to the reason for it. It feels like just a holdover from when this IP originally used `es_mapping::index_settings` instead of requesting the existing settings from the API that distracts from the actual goal of sending those settings to Elasticsearch. `es_mapping::index_settings`'s return format is only relevant insofar as it currently happens to return a dict exactly compatible with ES's create index API. That's not a necessary relationship, however, so it feels like a distraction from what goal the transformation of the index settings API response needs to be (to match ES create index input API).

Linking to the ES create index API request body documentation would be more direct and clearer, especially as that `es_mapping::index_settings` function is pretty much irrelevant for any of the other stuff this DAG does.

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-create-index.html#indices-create-api-request-body

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+3. Create the new index with the configuration generated in step 1. The index
+   name will either be a combination of the `media_type` and `index_suffix`
+   parameters (`{media_type}-{index_suffix}`, e.g. `image-my-special-suffix`),
+   or the `media_type` and a timestamp (`{media_type}-{timestamp}`, e.g.
+   `image-20200102030405`).
+   ([example from the existing ingestion server code](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L518-L521)).
```
sarayourfriend, line 196: Just to confirm (maybe this should be explicit in this IP), this will be done by making the create index request directly to ES, right? In other words, we're not creating the index by making a request to the ingestion server. The links to the ingestion server make this slightly confusing, even though I think I understand they're just meant to be there as examples of how to make the requests using the Python ES client :confused:.

Does the Airflow provider give the same ES client that the ingestion server uses? Otherwise the code samples will be fully irrelevant. Maybe somewhere should specify that we're going to use the `ElasticsearchPythonHook` specifically rather than the `ElasticsearchSQLHook` and then draw the specific connection to the fact that the Python hook returns the same ES client that is used in the ingestion server. That would help explain the use of the ingestion server as code examples.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+4. Initiate a reindex using the `source_index` as the source
+   ([example from the ingestion server](https://github.com/WordPress/openverse/blob/3fcce5ade2165955db5bbcb4f679257b3260547b/ingestion_server/ingestion_server/indexer.py#L525-L547)).
+   Rather than setting `wait_for_completion=True` on the ES reindex call and
+   having this step halt until the reindex is complete, this step will set
+   `wait_for_completion=False` and continue to the next step. The
+   [task ID returned from Elasticsearch for the reindex operation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-task-api)
+   will be stored in an XCom variable. This step will have `refresh=True` on the
```
sarayourfriend, line 203: Just a small nit, if you pass `refresh=True` to Elasticsearch's REST API instead of `refresh=true`, it will complain about invalid inputs. If this is meant to refer to the specific Python call rather than the REST API call and if we're using the Python hook, then it could be useful to call that out here.

To expand on what I said in the other comment, IP doesn't make it clear that we're using a Python API to interact with Elasticsearch rather than using a utility connection to send REST requests directly. Clarifying that would clear up this potential source of confusion for implementers.

sarayourfriend (approved):
LGTM. The plan makes sense, there are no blockers from my perspective, and I believe it can be successfully implemented. It also sufficiently leaves room open for iteration and flexibility during implementation. I've only left a couple comments on places where I think it could be clarified, either through more explicit expectations about how the Elasticsearch provider will be used or through changing the perspective of the documentation to reorient towards the ES API rather than our ingestion server implementation details (which don't have a necessary relationship).

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,380 @@
...
+   existing value (see [Merging policy](#merging-policy) below). This can also
+   be the entire index configuration, in which case the existing configuration
+   will be replaced entirely (see `override_config` parameter below).
+3. `index_suffix`: (Optional) The name suffix of the new index to create. This
+   will be a string, and will be used to name the index in Elasticsearch of the
+   form `{media_type}-{index_suffix}`. If not provided, the suffix will be a
+   timestamp of the form `YYYYMMDDHHMMSS`.
+4. `source_index`: (Optional) The existing index on Elasticsearch to use as the
+   basis for the new index. If not provided, the index aliased to `media_type`
+   will be used (e.g. `image` for the `image` media type).
```
krysal, line 118: Requesting changes for clarifications around the `source_index`, if the default index is assumed to come from the data refresh process I believe is worth pointing it out.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: Ahhh, I see what you mean! Implicit to my thinking was that one of those indices, be it the proportional-by-provider index (my pick) or the production-data-volume, would be aliased to `image` (or `audio` depending on the media type). Right now we do have those aliases, and they point to indices that were hand-made: 
image

I do think that's a piece to be hashed out as part of the final IP for this project (i.e. how to create those indices and which should be used as the "primary" alias). I'm realizing now though that "Staging Elasticsearch reindex DAGs for both potential index types" is maybe not the best name for that effort, potentially "Staging Elasticsearch index creation DAGs for both potential index types" is better? This is maybe even a discussion to have on that issue instead 😅

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: I answered this a bit above but I'll be sure to add some extra context for where this might be coming from in staging!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: The second paragraph contains the implementation directions for the first - I'll reword this to make that a bit more clear!

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: You're correct that we **will not** be using the ingestion server for this operation. The code example that is shared here is merely an example of how the ingestion server does this using the Elasticsearch Python API - I find examples within our own codebase, which have context I'm personally familiar with, more helpful than exclusively linking the Elasticsearch API documentation. 

I also do mention that we will be using the `ElasticsearchPythonHook` for these operations above this: https://github.com/WordPress/openverse/pull/2133/files/65d4e84f8b6570856d8976bee1a364d99c07b82a#diff-344cb6ad9993d7e365933d71faa76a18fbbb062a333a5ae4f73b9dee861b46d8R148 (Line 148 in the current iteration, or step 2.1). I will also add a link to the accompanying Elasticsearch documentation for the `index.create` call here, like I had done in the `index.get` call above.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: This was intended to reflect the parameters required for the Python hook, which does use the Elasticsearch Python API. I will try to make this more explicit throughout the document.

AetherUnbound:
@krysal @sarayourfriend thank you both for your additional thoughts! I believe I've addressed the points which required clarification, but let me know if there's anything further I can do to improve the IP.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,390 @@
...
+**Author**: @aetherunbound
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend
```
sarayourfriend, line 12:
```suggestion
- [x] @sarayourfriend
```

sarayourfriend (approved):
LGTM, excellent work, Madison. Thanks for clarifying the hook usage stuff :+1:

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
```diff
@@ -0,0 +1,390 @@
...
+
+### DAG
+
+Once the parameters are provided, the DAG will execute the following steps:
+
+1. (For `production` only) Check that there are no other reindexing jobs
+   currently underway. This can be done by checking whether the
+   `{media_type}_data_refresh` and `create_filtered_{media_type}_index` DAGs are
+   currently running. If either of them are, this DAG should fail immediately
+   with an appropriate error message.
```
stacimc, line 143: Should we add a note that we need to do this check in the `{media_type}_data_refresh` and `create_filtered_{media_type}_index` DAGs as well? Would it not be an issue if one of those DAGs started while this was running?

stacimc (approved):
Updates look fantastic :) No blockers, just a note for clarification.

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
AetherUnbound: Ah, I hadn't thought of the case where this was started before those 😖 It felt easy enough to fail this DAG in the first case, but for the other DAGs we'll need to probably have a more robust mechanism for handling this...I'll try to add some clarification/thoughts on this before this gets merged.

krysal (approved):
Thanks for clarifying! Looks great ✨

documentation/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.md
sarayourfriend: The data refresh DAG already waits for the create_filtered_index DAGs to finish. We could do the same for these new ones. So long as the new DAGs are only allowed to run 1 at a time, they wouldn't be able to indefinitely block the data refresh DAG from running, especially if the new DAGs themselves fail when the data refresh is running.

AetherUnbound: That's true! And that should be easy enough to add to the list of conditions for the data refresh. I'll do that!

PR #2160, opened 2023-05-22
Fix the full-height layout

### LINKED ISSUE #504: Filter sidebar is not synced correctly with page navigation causing page jittering when navigating between search and a single result
## Description

The filter sidebar pops out in too early when navigating back to search from a single result or it pops out too early when navigating from search to a single result.

This causes the search and single result pages to dramatically layout shift just before the navigation actually happens.

## Reproduction

1.  Go to https://search.openverse.engineering/search/?q=galah and lave the filter sidebar open
2.  Click on a result and note that the filter sidebar disappears before the single result page ever starts to render
3.  Click on the "go back to search results" link at the top of the single result page and notice that a) the link immediately disappears and b) the filter sidebar pops in causing everything on the page to shift left before actually navigating back to the search results page.

## Screenshots

https://user-images.githubusercontent.com/24264157/191226936-85353331-e7a9-49c5-b6ea-d3c5f7a54e97.mp4

## Resolution

### LINKED ISSUE #594: Scroll to top button is cut off on mobile

### LINKED ISSUE #1159: The banners height needs to be considered in height calculations

### LINKED ISSUE #2062: Filter apply button cut off in Firefox mobile if address bar is on the bottom
## Description

It appears that the filter tab's "Apply" button gets cut off on Firefox mobile if the address bar is set to be on the bottom.

## Reproduction

1. Visit openverse.org 
2. Make a search
3. Open the filter and observe the screenshot

## Screenshots

signal-2023-05-07-100019_002

## Environment

- Device: Google Pixel 7
- OS: Android 13
- Browser: Firefox

obulat:
## Fixes

Fixes #1159 by @obulat
Fixes #504 by @sarayourfriend 
Fixes #2062 by @AetherUnbound 
Fixes #594 by @obulat

## Description

This PR adds a new layout file for search pages.

This layout is different from content pages because needs the sidebar to be fixed, as well as the header. We were using `fixed` for its position, but it required adding a top margin to account for the sticky header.

This implementation uses the following CSS grid:
```
+--------------+---------+
|            header      |
+--------------+---------+
| main content | sidebar |
| footer       |         |
+--------------+---------+
```
This means that we no longer have to make the header or the sidebar sticky, and they are placed correctly vertically and not covered with other elements (see the changes in `search-with-banners/page-with-all-banners` snapshots for how it fixes the issue).

This also fixes the cut-off of the modal's bottom part and the scroll to top button on mobile screens.

This layout changes the scrollable element from the document to "main-content" (the part where the results are displayed, not the header or sidebar). This means that Playwright cannot create the full-page snapshots using `fullPage` prop, and this PR adds a utility that increases the viewport for no results page so that all of its content fits into the page to balance it.

## Testing Instructions

Run the app and check that the search page works correctly both on desktop and mobile screens.
On desktop:
- Try opening all banners by going to localhost:8443/af/search/?q=cat&referrer=creativecommons.org and opening the filters sidebar: its headings should be visible (unlike on https://openverse.org/af/search/?q=cat&referrer=creativecommons.org)
- Try searching - the recent searches should be visible above the page content
- If you scroll down (both with filters open and closed), the scroll to top button should be visible

On mobile:
- The content settings mobile should be fully visible, and the buttons should not be cut off.
- If you scroll down, the scroll to top button should be visible and not cut off.

The content layout (the pages like `/about` and single result pages) should also work correctly. They did not have a scroll to top button - do we want it there?

When you navigate from the single result view to the search results on desktop, there should be no flickering described in #504.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VScrollButton.vue
```diff
@@ -44,8 +44,13 @@ export default defineComponent({
...
     )
-    const scrollToTop = () => {
-      window.scrollTo({ top: 0, left: 0, behavior: "smooth" })
+    const scrollToTop = (e: MouseEvent) => {
+      const mainPage = e.currentTarget?.closest("#main-page")
+      if (mainPage) {
+        mainPage.scrollTo({ top: 0, left: 0, behavior: "smooth" })
+      } else {
+        window.scrollTo({ top: 0, left: 0, behavior: "smooth" })
+      }
```
AetherUnbound, line 53: Nit: is this any better/clearer? Or is there a better way to consolidate this?
```suggestion
      const element = mainPage ?? window
      element.scrollTo({ top: 0, left: 0, behavior: "smooth" })
```

AetherUnbound (approved):
This ran great locally and all the steps provided in the testing instructions behaved as expected!

frontend/src/components/VScrollButton.vue
obulat: That's definitely clearer! I'll check if TS complains about it locally :)

obulat: It works :)

frontend/src/components/VModal/VModalContent.vue
```diff
@@ -180,8 +180,7 @@ export default defineComponent({
   @apply bg-tx;
 }
 .modal-backdrop {
-  /* linting plugin changes the order if we use non-custom value of h-screen for 100vh */
-  @apply h-[100dvh] h-[100vh];
```
sarayourfriend, line 184: Is this comment no longer applicable?

sarayourfriend (commented):
Changes test fine to me, but it looks like the VR tests need another update, both Storybook and Playwright.

frontend/src/components/VModal/VModalContent.vue
obulat: I think the comment is unclear, but it's still applicable.
Normally, I'd prefer to use `h-[100dvh] h-screen` so that if `dvh` is not supported, tailwind uses the `vh` value which has more support. However, then eslint reorders them to `h-screen h-[100dvh]`, which defeats the purpose. That's why I use the custom values for both 100dvh and 100vh.

sarayourfriend: Should it be added back, but fixed, in that case? I'm not familiar enough with this to understand the implications either way, but it sounds like we're using this non-standard approach for a reason that could be documented (and prevent someone from thinking we should switch to the more standard approach)?

sarayourfriend (approved):
LGTM :slightly_smiling_face:

PR #2174, opened 2023-05-23
Make analytics switchable in production

zackkrida:
Makes analytics switchable in production, defaulting to off.

## Testing Instructions

`just frontend/run dev` and make sure the toggle on the `/privacy` page is disabled by default, but also that if you click it it stays "on" after page refresh.


dhruvkb (approved):
LGTM!

PR #2180, opened 2023-05-24
Include `env` as a dependency for `up` just recipe

### LINKED ISSUE #964: Quickstart guide doesn't mention `just env`
The quickstart guide as written will error upon the first `just up` because there's no `.env` file.

We can either ask users to `just env`, or we can consider including it elsewhere as part of `just install`, or again some alternative approach.

AetherUnbound:
## Fixes

Fixes #964 by @zackkrida

## Description

~~This PR adds the `env` just recipe as a dependency of both the `py-install` and `node-install` recipes, then changes the syntax for the `install` recipe to use just's implicit dependency management. This ensures `env` is run once rather than for each install command when `just install` is run.~~ 

~~Since `up` is the more common recipe, adding it as a dependency to `install` seems reasonable since it shouldn't be necessary to run that frequently.~~

Per @sarayourfriend's comment below, I have change this so `env` is now a dependency of `up`.

## Testing Instructions

I tested this by cloning a fresh repo (not a quick process!) and ensuring that the steps in the Quickstart guide (i.e. `just install` -> `just up`) did indeed succeed now because the `.env` files are created.

Running `just install` should be sufficient for testing though, as that will show the env steps running.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/justfile
+++ b/justfile
@@ -96,7 +96,7 @@ lint hook="" *files="": precommit
 ########
 
 # Create .env files from templates
-env:
+@env:
     # Root
     ([ ! -f .env ] && cp env.template .env) || true
     # Docker
@@ -156,7 +156,7 @@ build *args:
 
 # Also see `up` recipe in sub-justfiles
 # Bring all Docker services up, in all profiles
-up *flags: && ps
+up *flags: env && ps
     #!/usr/bin/env bash
     set -eo pipefail
     while true; do


sarayourfriend (commented):
I don't really understand why not adding it to `up`. `just env` takes less than a tenth of a second to run (maybe less than a 20th of a second) and, as it is the more common one to run, will catch _every_ case where it is necessary to have those .env files rather than just the initial installation (which doesn't directly depend on it and is itself not even necessary for just running the application locally without development tools).

```
➜  openverse git:(fix/region-name-template-rendering) ✗ time just env
# Root
([ ! -f .env ] && cp env.template .env) || true
# Docker
([ ! -f docker/minio/.env ] && cp docker/minio/env.template docker/minio/.env) || true
# First-party services
([ ! -f catalog/.env ] && cp catalog/env.template catalog/.env) || true
([ ! -f ingestion_server/.env ] && cp ingestion_server/env.template ingestion_server/.env) || true
([ ! -f api/.env ] && cp api/env.template api/.env) || true
just env  0.07s user 0.03s system 122% cpu 0.084 total
```

`env` should be made into a silent recipe though, it certainly doesn't need to print all of that every time _anything_ is run.

justfile
```diff
@@ -45,20 +45,18 @@ ENABLE_DC_OVERRIDES := env_var_or_default("OPENVERSE_ENABLE_DC_OVERRIDES", "true
...
-py-install:
+py-install: env
     just automations/python/install
     just documentation/install
 
 # Install all dependencies
-install:
-    just node-install
-    just py-install
+install: node-install py-install
```
dhruvkb, line 59: This is not aligned with the general convention in our `justfile`s. Alternatively we could adopt this and then there's a lot more instances of recipes just being a collection of other `just` recipe calls.

dhruvkb (approved):
LGTM, except a minor nit.

AetherUnbound:
I have no strong feelings for which recipe it should be attached to. I'll revert the changes to the `install` recipe and `env` a dependency of `up` instead.

sarayourfriend (approved):
LGTM :+1:

PR #2194, opened 2023-05-25
Remove focused playwright test so all tests run; fix broken tests

obulat:
## Fixes

Fixes #2191 by @obulat 🤞 

## Description

~~This PR reverts the changes to the Playwright config introduced in #1920 in an attempt to fix Playwright e2e tests.~~

~~I do not understand what in these changes makes Playwright not find all of the tests. @sarayourfriend, do you have an idea?~~

We missed the `.only` added in one of the tests, which caused Playwright to only run that one test...

## Testing Instructions

Check the CI Playwright e2e tests log - it should have around 155 tests (instead of 1).

Locally, you can use `just frontend/run test:playwright -- e2e` to run the tests.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
[...]

PR #2196, opened 2023-05-25
Update Playwright to use a custom test fixture with global mocks

zackkrida:
Creates a `test` fixture which overrides the base Playwright test fixture. This allows for easily setting global route mocks and other settings. Most crucially it removes the thousands of logs cluttering the output about the failed proxying of the analytics server.

This PR also makes our provider mocks global and adds a global mock for the analytics event endpoint, to prevent that from outputting hundreds of proxy-related error logs during the run of the test suite.

`frontend/test/playwright/utils/test-fixture.ts` is the main thing to review here. The rest is just implementations of the new fixture (switching the import from playwright to our custom test fixture).

~This will be cleaned up and undrafted once #2194 is merged.~

## Testing instructions

- Observe tests are still working and logs are no longer cluttered with analytics event warnings
- Ensure provider APIs are still mocked correctly in all relevant tests
- Try to `import { test } from "@playwright/test"` and you should be able to observe a lint error.


documentation/frontend/reference/playwright_tests.md
```diff
@@ -67,7 +67,7 @@ just frontend/run test:playwright visual-regression -u
 The above will run only test files with `visual-regression` in the path and will
 update any snapshot tests due to the `-u` flag.
 
-## Visual Regression tests
+## Writing Visual Regression tests
 
 When writing visual regression tests, it is good practice to write tests for
```
zackkrida, line 72: TODO: Add something about the new fixture here.

obulat (approved):
Nice cleanup!

Is there a way to add an eslint rule to stop anyone from importing `test` from `@playwright/test`?

zackkrida:
@obulat indeed, there is! I had already written it and forgot to push 🤦

.eslintrc.js
```diff
@@ -208,6 +208,12 @@ module.exports = {
             message:
               "Do not use @vue/test-utils' `shallowMount`. Use `~~/test/unit/test-utils/render` instead which includes helpful context setup or @testing-library/vue's `render` directly.",
           },
+          {
+            selector:
+              "ImportDeclaration[source.value='@playwright/test']:has(ImportSpecifier[local.name='test'])",
+            message:
+              "Do not import 'test' from '@playwright/test'. Use 'import { test } from '~~/test/playwright/utils/test-fixture' instead.",
```
zackkrida, line 215:
```suggestion
              "Do not import 'test' from '@playwright/test'. Use 'import { test } from '~~/test/playwright/utils/test-fixture' instead, which mocks Openverse provider APIs and the analytics server.",
```

frontend/test/playwright/utils/test-fixture.ts
```diff
@@ -0,0 +1,18 @@
...
+ * This file overwrites the default playwright test fixture
+ * to apply some global settings and route handling.
+ */
+export const test = base.extend({
+  page: async ({ page }, use) => {
+    // Intercept /api/event requests
+    await page.route("/api/event", (route) => route.abort())
+    // Intercept  provider APIs
+    await page.route("**.jamendo.com**", (route) => route.abort())
+    await page.route("**.freesound.**", (route) => route.abort())
```
AetherUnbound, line 14: Just curious, why only these two?

AetherUnbound (changes requested):
Hmm, the `eslint` rule doesn't appear to be working for me 😮 I made the following change:

```diff
diff --git a/frontend/test/storybook/visual-regression/v-select-field.spec.ts b/frontend/test/storybook/visual-regression/v-select-field.spec.ts
index c759c64af..b0265a671 100644
--- a/frontend/test/storybook/visual-regression/v-select-field.spec.ts
+++ b/frontend/test/storybook/visual-regression/v-select-field.spec.ts
@@ -1,6 +1,4 @@
-import { Page, expect } from "@playwright/test"
-
-import { test } from "~~/test/playwright/utils/test-fixture"
+import { Page, expect, test } from "@playwright/test"
 
 import { makeGotoWithArgs } from "~~/test/storybook/utils/args"
```

But `just lint eslint` is still passing. Is that expected?

frontend/test/playwright/utils/test-fixture.ts
zackkrida: These were the only two that existed, my guess would be that the tests only use sample audio data from these providers but perhaps someone else on @WordPress/openverse-frontend can confirm.

zackkrida:
@AetherUnbound I will look into the eslint rule! It _might_ be that you need to `just install` prior to testing, but I'll verify myself.

sarayourfriend:
@zackkrida The ESLint rule you added is in the overrides block scoped exclusively to the unit tests (see the `files` setting). You'll need to move it to a different block that targets the playwright tests (`frontend/test/{playwright,storybook}/**/*.spec.{ts,js}` would do it, I think).

zackkrida:
@sarayourfriend, ah, thanks so much! That's definitely it

frontend/test/playwright/utils/test-fixture.ts
zackkrida: That's probably an excellent idea, I'll create a separate issue.

zackkrida:
@AetherUnbound thanks to Sara's excellent suggestion this is ready for review 👍 the rule should definitely work now.

.eslintrc.js
```diff
@@ -209,6 +210,27 @@ module.exports = {
             message:
               "Do not use @vue/test-utils' `shallowMount`. Use `~~/test/unit/test-utils/render` instead which includes helpful context setup or @testing-library/vue's `render` directly.",
           },
+          {
+            selector:
+              "ImportDeclaration[source.value='@playwright/test']:has(ImportSpecifier[local.name='test'])",
+            message:
+              "Do not import 'test' from '@playwright/test'. Use 'import { test } from '~~/test/playwright/utils/test-fixture' instead, which mocks Openverse provider APIs and the analytics server.",
+          },
```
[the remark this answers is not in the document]
zackkrida: Yup, oops! Just forgot to delete.

sarayourfriend (approved):

sarayourfriend:
LGTM once the tests pass :+1:

AetherUnbound (approved):
Confirmed the lint rule now works (read: breaks correctly) for me! 😄

zackkrida:
Looks like there might be some test flake here 😢 I'll confirm on the next run and create the necessary critical issues.

PR #1130, opened 2023-04-04
Create `@openverse/eslint-plugin` package with inagural rule for analytics config

sarayourfriend:
## Fixes

Fixes the same issue as https://github.com/WordPress/openverse/pull/1125 but using ESLint instead of TypeScript, which allows for much more descriptive and helpful error messages.

## Description

- Create new `@openverse/eslint-plugin` package and setup scaffolding necessary for local packages including testing, typescript, etc.
- Add `@openverse/eslint-plugin` to the base eslint configuration
- Create `@openverse/analytics-configuration` rule to lint the events configuration

TODO:
- Get frontend unit tests running locally. For some reason I can't run the frontend unit tests locally successfully at all, even after recloning the repository, clearing my pnpm cache to redownload all packages, running the tests on main.
- Setup scripts for continuously building packages to make development easier. This should be possible via tsc's watch: https://www.typescriptlang.org/docs/handbook/configuring-watch.html

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like
      `Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or
      a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
[...]

PR #2199, opened 2023-05-26
Prevent accidentally committing focused test selectors

sarayourfriend:
## Fixes

Follow up to https://github.com/WordPress/openverse/pull/2194 by @obulat 

## Description

The related PR removes an accidentally committed `.only` focused test selector. We can easily prevent accidental commits like that in the future with the Jest ESLint plugin and with Playwright's own `forbidOnly` setting.

This PR adds the Jest plugin and enables Playwright's `forbidOnly` setting in CI.

## Testing Instructions

Check out the changes made to appease the new ESLint plugin. I think they are good rules. I opened a discussion about whether relying on testing-library's `get*` function's error handling behaviour as an implicit assertion is a good idea long-term: https://github.com/WordPress/openverse/discussions/2198

For now I've just selectively disabled the `expect-expect` check or expanded the assert function list for modules that heavily rely on that.

The plugin caught a handful of tests that truly were incorrect, either because they had no assertions at all (!!) or because they were using `expect` incorrectly (like `expect(predicate)` rather than `expect(someValue).predicate()`, the former does nothing).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (approved):
LGTM. Also what's up with #1130?

sarayourfriend:
@dhruvkb I started rebasing that PR earlier this week and need to get back to it. I basically just need to start back from scratch rather than trying to apply it directly here. I think the changes in here (hoisting the eslint config for tests up) will actually make it easier. Something I ran into and kept trying to work around in that PR was weird dependency installation issues for nested configurations.

If we run eslint in a local hook (as discussed in https://github.com/WordPress/openverse/discussions/2197) that would also help as installing a local package also took a good bit of fiddling and in the end didn't really work correctly all the time.

Hopefully next week I can get that other eslint PR working fully and we can write some more abstract custom rules :crossed_fingers:

sarayourfriend:
Shoot, I forgot to push :facepalm: Changes are up now, apologies for the confusion.

.eslintrc.js
```diff
@@ -95,16 +95,7 @@ module.exports = {
...
     ],
-    "no-restricted-syntax": [
-      "error",
-      ...i18nDestructureRules,
-      {
-        selector:
-          "ImportDeclaration[source.value='@vue/test-utils']:has(ImportSpecifier[local.name='shallowMount'])",
-        message:
-          "Do not use @vue/test-utils' `shallowMount`. Use @testing-library/vue's `render` instead.",
-      },
```
sarayourfriend, line 106: This rule is moved to the unit test scope in overrides further down in this file.

.eslintrc.js
```diff
@@ -194,9 +185,30 @@ module.exports = {
...
+        "no-restricted-syntax": [
+          "error",
+          {
+            selector:
+              "ImportDeclaration[source.value='@vue/test-utils']:has(ImportSpecifier[local.name='shallowMount'])",
+            message:
+              "Do not use @vue/test-utils' `shallowMount`. Use `~~/test/unit/test-utils/render` instead which includes helpful context setup or @testing-library/vue's `render` directly.",
+          },
+        ],
       },
```
sarayourfriend, line 212: Using a single ESLint configuration with `overrides` to configure sub-folder behaviour is ultimately easier to understand, I think, than nested configurations. We also had incorrectly configured the nested configuration with an unnecessary `extends` setting, (which causes a new configuration to be created rather than a merged one), showing that it's harder to maintain. It also makes it easier to reason about which rules apply where and which ones are shared without having to jump between multiple files. We can move the other ESLint configuration files into overrides in separate PRs.

frontend/test/playwright/playwright.config.ts
```diff
@@ -26,6 +26,7 @@ const pwCommand =
   process.env.FASTSTART !== "false" ? "start:playwright" : "prod:playwright"
 
 const config: PlaywrightTestConfig = {
+  forbidOnly: !!process.env.CI,
```
sarayourfriend, line 29: This allows it to pass locally but fails CI if `.only` is present.

frontend/test/unit/specs/components/AudioTrack/v-audio-track.spec.js
```diff
@@ -98,9 +100,9 @@ describe("AudioTrack", () => {
     const { getByRole, getByText } = render(VAudioTrack, options, configureVue)
 
     await fireEvent.click(getByRole("button"))
-    await expect(playStub).toHaveBeenCalledTimes(1)
-    await expect(pauseStub).toHaveBeenCalledTimes(1)
-    await expect(getByText(/Reproduction not allowed./i)).toBeVisible()
+    expect(playStub).toHaveBeenCalledTimes(1)
+    expect(pauseStub).toHaveBeenCalledTimes(1)
+    expect(getByText(/Reproduction not allowed./i)).toBeVisible()
```
sarayourfriend, line 105: None of these expect calls returned promises.

frontend/test/unit/specs/components/AudioTrack/v-box-layout.spec.js
```diff
@@ -18,15 +20,17 @@ describe("VBoxLayout", () => {
...
-    const screen = render(VBoxLayout, options)
-    screen.getByText(props.audio.title)
-    screen.getByLabelText("Attribution-NonCommercial-Share-Alike")
-    screen.getByText("Music")
+    render(VBoxLayout, options)
+    expect(() => {
+      screen.getByText(props.audio.title)
+      screen.getByLabelText("Attribution-NonCommercial-Share-Alike")
+      screen.getByText("Music")
+    }).not.toThrow()
```
sarayourfriend, line 28: This is a workaround related to https://github.com/WordPress/openverse/discussions/2198. I think it's better than setting each of these as assert functions using the comment approach in other modules.

A potentially better alternative is to change them to `query*` functions and assert `.not.toBeNull()` on them, but I'll let the discussion I linked play out before we make concerted refactors to these. We may decide on a different approach altogether.

frontend/test/unit/specs/components/v-tabs.spec.js
```diff
@@ -15,7 +17,7 @@ import VTabPanel from "~/components/VTabs/VTabPanel.vue"
  * @param {string} text
  * @param {boolean?} visibility
  */
-const expectVisibitility = (screen, text, visibility = true) => {
```
sarayourfriend, line 18: This function name was misspelled.

frontend/test/unit/specs/stores/search-store.spec.js
```diff
@@ -249,14 +249,14 @@ describe("Search Store", () => {
...
         if (supportedSearchTypes.includes(searchType)) {
+          // eslint-disable-next-line jest/no-conditional-expect
           expect(searchStore.query).not.toEqual(expectedQueryParams)
         }
         searchStore.clearFilters()
         expect(searchStore.searchQueryParams).toEqual(expectedQueryParams)
       }
     )
-  })
-  describe("actions", () => {
```
sarayourfriend, line 259: This describe block had the same name as the previous block so I merged them.

frontend/test/unit/specs/stores/ui-store.spec.js
```diff
@@ -101,11 +101,9 @@ describe("Ui Store", () => {
...
         // isMobileUa is set to true only if we explicitly get a mobile UA
         // from cookie or the browser request
-        if (key === "isMobileUa") {
-          expect(uiStore[key]).toEqual(false)
-        } else {
-          expect(uiStore[key]).toEqual(initialState[key])
-        }
+        const exepectedStoreValue =
+          key === "isMobileUi" ? false : initialState[key]
+        expect(uiStore[key]).toEqual(exepectedStoreValue)
```
sarayourfriend, line 106: This code is equivalent to the original but prevents the `no-conditional-expect` rule from needing to be disabled twice (or at all).

frontend/test/unit/specs/utils/attribution-html.spec.ts
```diff
@@ -64,11 +64,11 @@ describe("getAttribution", () => {
 
   it("does not use anchors in plain-text mode", () => {
     document.body.innerHTML = getAttribution(mediaItem, i18n)
-    expect(document.getElementsByTagName("a").length !== 0)
+    expect(document.getElementsByTagName("a")).not.toHaveLength(0)
     document.body.innerHTML = getAttribution(mediaItem, i18n, {
       isPlaintext: true,
     })
-    expect(document.getElementsByTagName("a").length === 0)
+    expect(document.getElementsByTagName("a")).toHaveLength(0)
```
sarayourfriend, line 71: The original version of these assertions were noops. `expect()` by itself does nothing. Here's an excellent example of the Jest plugin catching incorrect, false positive (and therefore dangerous because they create false confidence) tests.

.eslintrc.js
```diff
@@ -194,9 +185,30 @@ module.exports = {
...
+        ],
+        "no-restricted-syntax": [
+          "error",
+          {
+            selector:
+              "ImportDeclaration[source.value='@vue/test-utils']:has(ImportSpecifier[local.name='shallowMount'])",
+            message:
+              "Do not use @vue/test-utils' `shallowMount`. Use `~~/test/unit/test-utils/render` instead which includes helpful context setup or @testing-library/vue's `render` directly.",
+          },
+        ],
```
zackkrida, line 211: OOOH this is perfect, I need this in #2196

zackkrida (changes requested):
According to CI there are two remaining unit test failures but the overall approach here and restructuring of the eslint configurations is excellent!

frontend/test/unit/specs/components/v-popover.spec.js
```diff
@@ -147,20 +143,26 @@ describe("VPopover", () => {
...
     })
   })
 
   describe("hideOnClickOutside", () => {
-    it("should hide the popover if a click happens outside the popover by default", async () => {
+    // This test is broken (for some reason clickOutside does not appear to actually cause a click
+    // to happen in this case).
+    // https://github.com/WordPress/openverse/issues/2220
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip("should hide the popover if a click happens outside the popover by default", async () => {
```
obulat, line 158:
```suggestion
    it("should hide the popover if a click happens outside the popover by default", async () => {
```
The fix has been merged.

obulat (approved):
Thank you for all your work on improving our confidence in tests and code changes, @sarayourfriend . All the changes look correct

frontend/test/playwright/playwright.config.ts
AetherUnbound: Why would we allow it to pass locally if `.only` is present but fail on CI? Is it time consuming to have this check run locally? And do you think it makes sense to add a comment as to why this was added here?

frontend/test/storybook/playwright.config.ts
```diff
@@ -10,6 +10,7 @@ addAliases({
 })
 
 const config: PlaywrightTestConfig = {
+  forbidOnly: !!process.env.CI,
```
AetherUnbound, line 13: Same note here as above

zackkrida:
@sarayourfriend I'm going to merge this in the interest of getting #2196 undrafted and so I can finish it while it's fresh in my memory.

AetherUnbound (approved):
All of the test changes make sense to me! I have a few questions but nothing to block a merge. I'm impressed by how specific the ESLint rules can get!

frontend/test/storybook/playwright.config.ts
sarayourfriend: `.only` is useful to have locally if you're debugging a single, specific test, for example, in debug mode, where it runs in a "headed" browser that you can poke around the context with: https://playwright.dev/docs/debug

In that setting, if you haven't selected a single test to run, the runner is going to run _everything_ in the module, requiring you to try check which test is running each time and pass it until you get to the one you want.

That's to say: `only` is useful locally, but harmful if committed and forbidding it in CI is an easy way to catch when it is committed. Additionally, something that isn't necessarily obvious, is that Playwright and Jest are distinct test runners. Even though they look very similar in API, they have certain differences that will make the `eslint-plugin-jest` incompatible (or at least flaky) with Playwright. It would be better if our configuration didn't let you commit the test selection, though, and luckily there is a Playwright-specific ESLint plugin with the same `no-focused-test` rule that the Jest plugin provides. I've opened an issue to add it here: https://github.com/WordPress/openverse/issues/2312. That would give us a whole host of helpful checks, but also prevent `.only` from being committed at all in our Playwright tests (rather than just failing in CI).

AetherUnbound: Woah, okay that makes sense! Thanks for the helpful context!

PR #2209, opened 2023-05-27
Add index readiness check to data refresh

### LINKED ISSUE #2071: Check API queries against the new index before promotion
## Problem

The data refresh DAG makes the ingestion server promote the new index once it is ready. This can be an issue because as we recently experienced, a ready index does not always imply results will be properly returned. Instead the promotion should take place only when the API calls against the index are successful.

## Description

Before promotion, use the `internal__index` param (#2070) to make API calls against the new index and only promote if results are returned and the number of hits is within expected ranges.

Eventually this can be expanded to include further checks regarding performance, quality, relevance and content safety.

## Additional context

I'm unsure about the best place to check for this but I think it should be at the data-refresh DAG level because Airflow can handle the retries in case the checks fail for a short while and then start passing after many hours.

stacimc:
## Fixes

Fixes #2071 by @dhruvkb  

## Description

Adds an `index_readiness_check` task to the data refresh DAGs (including the `create_filtered_<media>_index` DAGs), used to detect issues where the index is ready, but the API is not connecting properly and returning results. The task is an HttpSensor which runs after `ingest_upstream` completes (creating the new es index), and ~queries the API using the new `internal__index` param to target the new index~ directly queries the new ES index, passing only when we retrieve a threshold number of `hits`. Once it passes, the DAG is able to move on to the `promote` steps and actually promote the indices in production.

Screen Shot 2023-06-15 at 4 31 32 PM

_Only the data refresh TaskGroup is shown._

Screen Shot 2023-06-20 at 3 01 07 PM

The Sensor times out in 24 hours, so the entire DAG will fail (without promoting the new index) and notify in Slack if the ES index does not become ready within that timeframe. After manual intervention, the DAG can be safely manually resumed from this point.

### Setup

* Add `ES_INDEX_READINESS_RECORD_COUNT=1000` to your `catalog/.env`. This is used in local testing to lower the threshold record count, since local test data does not have as many records.

* Set up the  ES connection by adding the following to `catalog/.env`: `AIRFLOW_CONN_ELASTICSEARCH_HTTP_PRODUCTION=http://es:9200`

* Run `just down && just up`

* Enable the `create_filtered_audio_index` and `create_filtered_image_index` DAGs

### Tests

1. For each media type, run the associated data_refresh DAG and ensure that the whole DAG passes.
2. Test "unhealthy" results: the easiest way to do this is to manually edit the `ES_INDEX_READINESS_RECORD_COUNT` record to a larger number (`100_000`). Then run the DAG and observe that the `index_readiness_check` continuously goes `up_for_reschedule` and does not pass.
3. Test 404s for invalid index: the easiest way to simulate this is to hardcode the `index_readiness_check` task to query with an incorrect index name. Manually edit the task [here](https://github.com/WordPress/openverse/pull/2209/files#diff-571622cb9ebb289fda3e250897a511865723ab7f6d0691bd6a4ec4e527be38e2R159):
```
...
endpoint=f"{media_type}-foo/_search",  # This index does not exist
method="GET",
...
```
Re-run the DAG and observe that the task continuously goes `up_for_reschedule` and does not pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/constants.py
+++ b/catalog/dags/common/constants.py
@@ -37,3 +37,4 @@ POSTGRES_API_STAGING_CONN_ID = os.getenv(
 API_CONN_ID = os.getenv("API_CONN_ID", "api")
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
+ES_PROD_CONN_ID = "elasticsearch_production"
--- a/catalog/dags/common/constants.py
+++ b/catalog/dags/common/constants.py
@@ -34,5 +34,6 @@ OPENLEDGER_API_CONN_ID = os.getenv("OPENLEDGER_API_CONN_ID", "postgres_openledge
 POSTGRES_API_STAGING_CONN_ID = os.getenv(
     "POSTGRES_API_STAGING_CONN_ID", "postgres_openledger_api_staging"
 )
+API_CONN_ID = os.getenv("API_CONN_ID", "api")
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
--- a/catalog/dags/common/constants.py
+++ b/catalog/dags/common/constants.py
@@ -36,4 +36,4 @@ POSTGRES_API_STAGING_CONN_ID = os.getenv(
 )
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
-ES_PROD_CONN_ID = "elasticsearch_production"
+ES_PROD_HTTP_CONN_ID = "elasticsearch_http_production"
--- a/catalog/dags/common/constants.py
+++ b/catalog/dags/common/constants.py
@@ -34,7 +34,6 @@ OPENLEDGER_API_CONN_ID = os.getenv("OPENLEDGER_API_CONN_ID", "postgres_openledge
 POSTGRES_API_STAGING_CONN_ID = os.getenv(
     "POSTGRES_API_STAGING_CONN_ID", "postgres_openledger_api_staging"
 )
-API_CONN_ID = os.getenv("API_CONN_ID", "api")
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
 ES_PROD_CONN_ID = "elasticsearch_production"
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -148,7 +148,7 @@ def trigger_and_wait_for_task(
 def index_readiness_check(
     media_type: str,
     index_suffix: str,
-    timeout: timedelta,
+    timeout: timedelta = timedelta(days=1),
     poke_interval: int = POKE_INTERVAL,
 ) -> HttpSensor:
     """
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -4,11 +4,12 @@ from datetime import timedelta
 from urllib.parse import urlparse
 
 from airflow.exceptions import AirflowSkipException
+from airflow.models import Variable
 from airflow.providers.http.operators.http import SimpleHttpOperator
 from airflow.providers.http.sensors.http import HttpSensor
 from requests import Response
 
-from common.constants import API_CONN_ID, XCOM_PULL_TEMPLATE
+from common.constants import ES_PROD_CONN_ID, XCOM_PULL_TEMPLATE
 
 
 logger = logging.getLogger(__name__)
@@ -68,17 +69,19 @@ def response_check_wait_for_completion(response: Response) -> bool:
     return True
 
 
-def response_check_api_health_check(response: Response) -> bool:
+def response_check_index_readiness_check(response: Response) -> bool:
     """
-    Handle the response for `api_health_check` Sensor.
-
-    Waits for an API query using the chosen index to return results.
+    Handle the response for `index_readiness_check` Sensor, to await a
+    healthy Elasticsearch cluster.
     """
     data = response.json()
+    status = data.get("status", None)
 
-    result_count = data.get("result_count", 0)
-    logger.info(f"Retrieved {result_count} records from the API using the new index.")
-    return result_count > THRESHOLD_RESULT_COUNT
+    if Variable.get("ENVIRONMENT", default_var="dev") == "prod":
+        return status == "green"
+
+    # If we are not in production, we may proceed with a yellow cluster status.
+    return status in ["green", "yellow"]
 
 
 def get_current_index(target_alias: str) -> SimpleHttpOperator:
@@ -143,35 +146,23 @@ def trigger_and_wait_for_task(
     return trigger, waiter
 
 
-def api_health_check(
+def index_readiness_check(
     media_type: str,
     index_suffix: str,
     timeout: timedelta,
     poke_interval: int = POKE_INTERVAL,
-    retries: int = 8,
 ) -> HttpSensor:
     """
-    Poll the API for the given media endpoint, targeting the given index, and return
-    successfully when healthy results are returned.
-
-    If no results are returned, it will reschedule and poll again. If a 400 is received
-    because the API is not able to connect to the given index, the task will fail and
-    retry. The task is set to retry with exponential backoff, such that the retry delay
-    doubles between each attempt.
+    Poll the Elasticsearch cluster health endpoint, targeting the given index, and
+    return successfully when a healthy status is identified.
     """
     return HttpSensor(
-        task_id="api_health_check",
-        http_conn_id=API_CONN_ID,
-        endpoint=f"{media_type}",
-        request_params={"internal__index": f"{media_type}-{index_suffix}"},
-        headers={
-            "Authorization": "Bearer {{ var.value.get('API_ACCESS_TOKEN', 'not_set') }}"
-        },
+        task_id="index_readiness_check",
+        http_conn_id=ES_PROD_CONN_ID,
+        endpoint=f"_cluster/health/{media_type}-{index_suffix}",
         method="GET",
-        response_check=response_check_api_health_check,
+        response_check=response_check_index_readiness_check,
         mode="reschedule",
         poke_interval=poke_interval,
         timeout=timeout.total_seconds(),
-        retries=retries,
-        retry_exponential_backoff=True,
     )
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -160,7 +160,7 @@ def api_health_check(
     retry. The task is set to retry with exponential backoff, such that the retry delay
     doubles between each attempt.
     """
-    access_token = Variable.get("API_ACCESS_TOKEN")
+    access_token = Variable.get("API_ACCESS_TOKEN", "not_set")
     return HttpSensor(
         task_id="api_health_check",
         http_conn_id=API_CONN_ID,
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -8,7 +8,7 @@ from airflow.providers.http.operators.http import SimpleHttpOperator
 from airflow.providers.http.sensors.http import HttpSensor
 from requests import Response
 
-from common.constants import ES_PROD_CONN_ID, XCOM_PULL_TEMPLATE
+from common.constants import ES_PROD_HTTP_CONN_ID, XCOM_PULL_TEMPLATE
 
 
 logger = logging.getLogger(__name__)
@@ -154,7 +154,7 @@ def index_readiness_check(
     """
     return HttpSensor(
         task_id="index_readiness_check",
-        http_conn_id=ES_PROD_CONN_ID,
+        http_conn_id=ES_PROD_HTTP_CONN_ID,
         endpoint=f"{media_type}-{index_suffix}/_search",
         method="GET",
         response_check=response_check_index_readiness_check,
--- a/catalog/dags/common/ingestion_server.py
+++ b/catalog/dags/common/ingestion_server.py
@@ -8,13 +8,15 @@ from airflow.providers.http.operators.http import SimpleHttpOperator
 from airflow.providers.http.sensors.http import HttpSensor
 from requests import Response
 
-from common.constants import XCOM_PULL_TEMPLATE
+from common.constants import API_CONN_ID, XCOM_PULL_TEMPLATE
 
 
 logger = logging.getLogger(__name__)
 
 
 POKE_INTERVAL = int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60 * 15))
+# Minimum number of records we expect to get back from the API when querying an index.
+THRESHOLD_RESULT_COUNT = int(os.getenv("API_HEALTHCHECK_RECORD_COUNT", 10_000))
 
 
 def response_filter_stat(response: Response) -> str:
@@ -31,6 +33,7 @@ def response_filter_stat(response: Response) -> str:
     # Indices are named as '<media type>-<suffix>', so everything after the first
     # hyphen '-' is the suffix.
     _, index_suffix = index_name.split("-", maxsplit=1)
+    logger.info(index_suffix)
     return index_suffix
 
 
@@ -63,6 +66,19 @@ def response_check_wait_for_completion(response: Response) -> bool:
 
     logger.info(f"Data refresh done with {data['progress']}% completed.")
     return True
+
+
+def response_check_api_health_check(response: Response) -> bool:
+    """
+    Handle the response for `api_health_check` Sensor.
+
+    Waits for an API query using the chosen index to return results.
+    """
+    data = response.json()
+
+    result_count = data.get("result_count", 0)
+    logger.info(f"Retrieved {result_count} records from the API using the new index.")
+    return result_count > THRESHOLD_RESULT_COUNT
 
 
 def get_current_index(target_alias: str) -> SimpleHttpOperator:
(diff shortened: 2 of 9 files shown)


AetherUnbound:
I'm testing this out locally right now, but wanted to surface something - the Oauth tokens we receive from the API aren't indefinite, right? I think they have an expiration time (my local one was `43200s`), which means that the token will only be valid for a short time. We may need to dust off the [Oauth DAGs](https://github.com/WordPress/openverse/blob/a5fb3ea5f5671085f0239641ddc1c637a6193763/catalog/dags/oauth2/README.md) in order to get this working without intervention 😮

catalog/dags/common/constants.py
```diff
@@ -31,5 +31,6 @@
 
 POSTGRES_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
 OPENLEDGER_API_CONN_ID = os.getenv("OPENLEDGER_API_CONN_ID", "postgres_openledger_api")
+API_CONN_ID = os.getenv("API_CONN_ID", "api")
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
```
AetherUnbound, line 36: Just making a note here for myself - I'm going to make an issue to remove the `conn_id` configuration - we really don't need to allow the connection ID itself to be configurable, we should pick something and stick with it 😅 This PR follows the paradigm though so no need to change here!

catalog/dags/common/ingestion_server.py
```diff
@@ -125,3 +142,36 @@ def trigger_and_wait_for_task(
...
+    retry. The task is set to retry with exponential backoff, such that the retry delay
+    doubles between each attempt.
+    """
+    access_token = Variable.get("API_ACCESS_TOKEN", "not_set")
+    return HttpSensor(
+        task_id="api_health_check",
+        http_conn_id=API_CONN_ID,
+        endpoint=f"{media_type}",
+        request_params={"internal__index": f"{media_type}-{index_suffix}"},
+        headers={"Authorization": f"Bearer {access_token}"},
```
AetherUnbound, line 169: The `headers` param of the HTTP sensor is templated, so we should be able to just use the value directly!

```suggestion
    return HttpSensor(
        task_id="api_health_check",
        http_conn_id=API_CONN_ID,
        endpoint=f"{media_type}",
        request_params={"internal__index": f"{media_type}-{index_suffix}"},
        headers={"Authorization": "Bearer {{ var.value.API_ACCESS_TOKEN }}"},
```

catalog/dags/common/ingestion_server.py
```diff
@@ -125,3 +142,36 @@ def trigger_and_wait_for_task(
...
+        method="GET",
+        response_check=response_check_api_health_check,
+        mode="reschedule",
+        poke_interval=poke_interval,
+        timeout=timeout.total_seconds(),
+        retries=retries,
+        retry_exponential_backoff=True,
```
AetherUnbound, line 176: TIL, very neat!

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -152,25 +152,49 @@ def create_data_refresh_task_group(
...
+            index_suffix=XCOM_PULL_TEMPLATE.format(
+                generate_index_suffix.task_id, "return_value"
+            ),
+            timeout=data_refresh.api_healthcheck_timeout,
+        )
+        tasks.append(api_health_check)
+
+        # Trigger the `promote` task on the ingestion server and await its completion.
+        # This task promotes the newly created API DB table and elasticsearch index.
+        with TaskGroup(group_id="promote") as promote_tasks:
```
AetherUnbound, line 185: Rather than duplicating this and splitting it out, is it maybe better to move it into a generation function that takes an action and returns the task group?

catalog/tests/dags/common/test_ingestion_server.py
```diff
@@ -1,11 +1,49 @@
...
+
+@pytest.fixture()
+def healthcheck_dag():
+    index_suffix = "my-test-suffix"
+
+    # Create a DAG that just has an api_health_check task
+    with DAG(dag_id=TEST_DAG_ID, schedule=None, start_date=TEST_START_DATE) as dag:
+        ingestion_server.api_health_check(
+            media_type="image", index_suffix=index_suffix, timeout=timedelta(days=1)
+        )
```
AetherUnbound, line 42: Nit: it seems like `index_suffix` can just be defined in-line here at the `api_health_check` call

AetherUnbound (approved):
This LGTM! I have a few comments but nothing blocking. I do, however, think that the token piece will require a follow up PR for managing the Oauth registration.

> If an access token is invalid or expired querying the media endpoints using `internal__index` will not throw an error: it will silently query against the live table. This could obviously be misleading in the health check.

This seems like behavior we'd want to change on the API; returning a 400 would probably be best in the case that a token is invalid for that particular parameter I think, especially given that the token we have will eventually expire.

catalog/dags/common/constants.py
AetherUnbound: Issue made! https://github.com/WordPress/openverse/issues/2294

catalog/dags/data_refresh/data_refresh_task_factory.py
stacimc: Yeah, I went back and forth on this during implementation as well and I'm not sure! We could add another `trigger_and_wait_for_task_group` function that just wraps `trigger_and_wait_for_task` in a TaskGroup, but there's almost no code duplication saved because we still need to pass through all the same params. We can create the TaskGroups in a loop as was done originally, but the api health check needs to be between them in the task control flow. We _can_ set up the dependencies afterward but I thought doing it this way much improved readability, especially because it allowed me to more easily document the individual task groups in comments. I would definitely be more inclined to use the action map if we end up adding more than two trigger-and-wait taskgroups.

AetherUnbound: Totally fair! We can keep it as is 😄

catalog/dags/common/constants.py
stacimc: Totally agree! 😅

sarayourfriend (changes requested):
I may be wrong about this, but I think the current implementation incorrectly uses the API token/misses an important implementation detail.

To make authenticated requests to the API, a new token must be retrieved using the API credentials. These tokens expire after a set period of time. See the response body documented in https://api.openverse.engineering/v1/#tag/auth/operation/token which includes `expires_in`. The token _must_ be refreshed by making a new API token request each time. Even our own frontend does this token refresh dance: 

https://github.com//WordPress/openverse/blob/697f62f01a32cb7fcf2f4a7627650a113cba40da/frontend/src/plugins/api-token.server.ts#L71-L102

Based on what I've seen of how the AwsHook works from the AWS provider, I think a new Hook is the right way to manage that kind of thing. The hook would need the client ID and secret in order to make the token refresh request. Ideally the newest token would be stored somewhere in the database (or shared memory? Redis?) and reused if possible or refreshed before a request is sent.

sarayourfriend (commented):
Two questions:

- Is it possible to derive index "readiness" from the Elasticsearch API? If so, it could be a more reliable and technically "correct" indicator than a rather generic threshold of responses to expect. This approach works, but it also required complicating the API by adding the `internal__index` param, which, if not needed, would be nice to remove and reduce the API complexity. Though, I don't know if that parameter is only needed for this, or if this is just one of several use cases for it.
- At a glance, I think these changes don't make https://github.com/WordPress/openverse/issues/2314 more complicated. Can you double-check to make sure that is the case? These changes, if anything, probably make it slightly easier, but I could be wrong about that and just want to have it in mind so as not to create and inadvertent snags for the other issue down the line.

sarayourfriend:
The workers should ping `/worker_finished` when they're done, which triggers the refresh:

https://github.com//WordPress/openverse/blob/4d6e995d2268dcd71381bce61b51b8eb558d7adf/ingestion_server/ingestion_server/api.py#L264-L313

We shouldn't refresh during the bulk insert because the cost of each refresh can be offset until all documents are inserted and indexed (i.e., when `/worker_finished` is called by the last worker).

zackkrida:
A small question here: is health_check appropriate or does it cause too much confusion with our existing service-level healthchecks for the API, frontend, etc? Would "index readiness check" or something potentially be more appropriate?

catalog/dags/common/constants.py
```diff
@@ -36,3 +36,4 @@
 )
 AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
 AWS_RDS_CONN_ID = os.environ.get("AWS_RDS_CONN_ID", AWS_CONN_ID)
+ES_PROD_CONN_ID = "elasticsearch_production"
```
stacimc, line 39: Since we're moving away from the pattern of defining a separate `CONN_ID`, I did not do so here.

catalog/dags/common/ingestion_server.py
```diff
@@ -65,6 +69,21 @@ def response_check_wait_for_completion(response: Response) -> bool:
...
+    healthy Elasticsearch cluster.
+    """
+    data = response.json()
+    status = data.get("status", None)
+
+    if Variable.get("ENVIRONMENT", default_var="dev") == "prod":
+        return status == "green"
+
+    # If we are not in production, we may proceed with a yellow cluster status.
+    return status in ["green", "yellow"]
```
sarayourfriend, line 84: This confused me for a moment because the staging ES cluster should also always be green, but then I remembered that the catalog's environment is always either `prod` or `dev` ("local", unfortunately named considering our infrastructure configuration), and it would still be `prod` even if this DAG was pointed at the staging cluster.

I don't have a suggestion for how to clarify this, it seems like just a funky bit of domain knowledge for our Airflow instance.

catalog/dags/common/ingestion_server.py
```diff
@@ -4,17 +4,20 @@
...
-from common.constants import XCOM_PULL_TEMPLATE
+from common.constants import ES_PROD_CONN_ID, XCOM_PULL_TEMPLATE
 
 
 logger = logging.getLogger(__name__)
 
 
 POKE_INTERVAL = int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60 * 15))
+# Minimum number of records we expect to get back from the API when querying an index.
+THRESHOLD_RESULT_COUNT = int(os.getenv("API_HEALTHCHECK_RECORD_COUNT", 10_000))
```
sarayourfriend, line 20: This is no longer used in this PR, right? Can it be removed?

catalog/tests/dags/common/test_ingestion_server.py
```diff
@@ -19,7 +55,82 @@
...
     actual = ingestion_server.response_filter_stat(response)
     assert actual == expected
+
+
+@pytest.mark.parametrize(
+    "response_code, response_json, environment, expected_status",
+    [
+        # Production
+        # Healthy cluster
+        (200, {"status": "green"}, "prod", TaskInstanceState.SUCCESS),
```
sarayourfriend, line 69: Just a nit/tip, you can give parametrize param groups a custom id. This helps make it easier to identify which ones fail in the test output as well because pytest uses the id rather than concatenating each param value's `repr` to name the parametrized instance.

```suggestion
        pytest.param(
        200,
        {"status": "green"},
        "prod",
        TaskInstanceState.SUCCESS,
        id="healthy-cluster"
       ),
```

If this test failed, the test name would be `test_index_readiness_check[healthy-cluster]` instead of `test_index_readiness_check[200-response_json0-prod-SUCCESS]`

sarayourfriend (commented):
The API related environment variables can be removed now, right?

sarayourfriend (commented):
I've tested this locally and it works great, including the failure states.

One question: are we confident that index "readiness" can be sufficiently deduced from cluster health? I can't remember from the original incident whether the cluster was reporting green health even while it wasn't returning any search results.

I was also curious if the cluster would go yellow at the expected time, but to test that I had to have a green cluster locally. I adapted this docker-compose ES configuration to get a three node cluster running locally with the diff below (collapsed to save space):

<details>
<summary>Multi-node local ES cluster with green health</summary>

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 3304ac042..36d0b0bc1 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -177,30 +177,65 @@ services:
     ports:
       - "50263:6379"
 
-  es:
-    profiles:
-      - ingestion_server
-      - api
+  es01:
+    networks:
+      default:
+        aliases:
+          - es
     image: docker.elastic.co/elasticsearch/elasticsearch:7.12.0
+    container_name: es01
+    environment:
+      - node.name=es01
+      - cluster.name=openverse-local
+      - discovery.seed_hosts=es02,es03
+      - cluster.initial_master_nodes=es01,es02,es03
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    depends_on:
+      - es02
+      - es03
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - es-data01:/usr/share/elasticsearch/data
     ports:
       - "50292:9200"
-    env_file:
-      - docker/es/env.docker
-    healthcheck:
-      test:
-        [
-          "CMD-SHELL",
-          "curl -si -XGET 'localhost:9200/_cluster/health?pretty' | grep -qE 'yellow|green'",
-        ]
-      interval: 10s
-      timeout: 60s
-      retries: 10
+    
+  es02:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.12.0
+    container_name: es02
+    environment:
+      - node.name=es02
+      - cluster.name=openverse-local
+      - discovery.seed_hosts=es01,es03
+      - cluster.initial_master_nodes=es01,es02,es03
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
     ulimits:
-      nofile:
-        soft: 65536
-        hard: 65536
+      memlock:
+        soft: -1
+        hard: -1
+    volumes:
+      - es-data02:/usr/share/elasticsearch/data
+    
+  es03:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.12.0
+    container_name: es03
+    environment:
+      - node.name=es03
+      - cluster.name=openverse-local
+      - discovery.seed_hosts=es01,es02
+      - cluster.initial_master_nodes=es01,es02,es03
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    ulimits:
+      memlock:
+        soft: -1
+        hard: -1
     volumes:
-      - es-data:/usr/share/elasticsearch/data
+      - es-data03:/usr/share/elasticsearch/data
 
   web:
     profiles:
@@ -219,7 +254,7 @@ services:
       - "50230:3000" # Sphinx (unused by default; see `sphinx-live` recipe)
     depends_on:
       - db
-      - es
+      - es01
       - cache
     env_file:
       - api/env.docker
@@ -246,7 +281,7 @@ services:
     depends_on:
       - db
       - upstream_db
-      - es
+      - es01
       - indexer_worker
     volumes:
       - ./ingestion_server:/ingestion_server
@@ -272,7 +307,7 @@ services:
     depends_on:
       - db
       - upstream_db
-      - es
+      - es01
     volumes:
       - ./ingestion_server:/ingestion_server
     env_file:
@@ -300,6 +335,8 @@ volumes:
   catalog-postgres:
   plausible-postgres:
   plausible-clickhouse:
-  es-data:
+  es-data01:
+  es-data02:
+  es-data03:
   minio:
   catalog-cache:
```
</details>

When I run the reindex, it does go yellow at the time we expect here.

So my only question is whether it's still worth querying the new index directly to see if it reports hits matching/exceeding the threshold.

Also, as a follow-up to this change, are we able to remove `exact_index` after this or do we need to keep it for anything else?

I'll approve the questions are cleared up and unneeded environment variables etc are cleaned up (if we don't need them anymore).

catalog/dags/common/ingestion_server.py
stacimc: Ah yes, thanks -- missed that one.

stacimc:
> One question: are we confident that index "readiness" can be sufficiently deduced from cluster health? I can't remember from the original incident whether the cluster was reporting green health even while it wasn't returning any search results.
> ...
> When I run the reindex, it does go yellow at the time we expect here.

I _was_ under this impression from the incident report and this seems like it _should_ be the case from the elasticsearch documentation. However I went back through Slack just now to triple check and saw a comment suggesting the cluster was green when it was returning no documents 😱

I've updated this to check for hits from the index instead. I don't think it's necessary to keep the cluster health check as well, if it indeed cannot be relied upon.

> Also, as a follow-up to this change, are we able to remove exact_index after this or do we need to keep it for anything else?

I believe we can and I intend to open that as a follow-up PR, yes.

AetherUnbound:
It seems this needs the `elasticsearch_production` connection, do you mind adding the steps for defining that in the testing instructions?

sarayourfriend:
> It seems this needs the elasticsearch_production connection, do you mind adding the steps for defining that in the testing instructions? Additionally, we're going to have an Elasticsearch-type (rather than HTTP-type) connection added as part of https://github.com/WordPress/openverse/issues/2371, should we distinguish this connection type in some way? Possibly elasticsearch_http_production?

I figured this HTTP based one would be replaced with the full ES provider package one :grimacing: I guess that would need to be added as an addendum to the IP though.

obulat (commented):
I think I'm missing something in my local setup, because I can't get the data_refresh dag to run locally.
"wait_for_data_refresh" step is always brown ("scheduled"). If I manually set it "successful", then the next step, , fails with 
```
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPConnection object at 0xffffaa81d930>: Failed to establish a new connection: [Errno 111] Connection refused
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='172.17.0.1', port=8001): Max retries exceeded with url: /stat/audio (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xffffaa81d930>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

sarayourfriend (approved):
LGTM! Works perfectly locally.

@obulat Can you check that you recreated the `catalog/.env` file (`rm catalog/.env && cp catalog/env.template catalog/.env && just down && just up`) and created the data refresh pool? I always get weird behaviour when I forget either of those things even try to rerun.

I opened #2434 to prevent the data refresh pool from missing locally.

stacimc:
> @obulat Can you check that you recreated the catalog/.env file (rm catalog/.env && cp catalog/env.template catalog/.env && just down && just up) and created the data refresh pool? I always get weird behaviour when I forget either of those things even try to rerun.
>
> I opened https://github.com/WordPress/openverse/issues/2434 to prevent the data refresh pool from missing locally.

@sarayourfriend FWIW #2352 made it so that you don't need to set up the `data_refresh` pool in local development, which I rebased this branch specifically to pick up because it's so annoying :)

@obulat Some of the env variable names changed during development -- the instructions are up to date but maybe you had an outdated connection id? If recreating the catalog/.env file and `just down && just up` doesn't work, I'd also recommend running `just init` to ensure you have some data in the catalog & api dbs.

obulat (approved):
> @obulat Some of the env variable names changed during development -- the instructions are up to date but maybe you had an outdated connection id? If recreating the catalog/.env file and `just down && just up` doesn't work, I'd also recommend running `just init` to ensure you have some data in the catalog & api dbs.

I checked and re-checked the env variables and the `data_refresh` pool, but apparently missed something. Ran the commands @sarayourfriend listed to recreate the `.env` file, and everything ran very well! Excited to start running refreshes again 🚀

sarayourfriend:
> @sarayourfriend FWIW https://github.com/WordPress/openverse/pull/2352 made it so that you don't need to set up the data_refresh pool in local development, which I rebased this branch specifically to pick up because it's so annoying :)

!! Thanks for letting me know, I guess that makes the other issue moot :slightly_smiling_face:

sarayourfriend:
@stacimc I just realised that we probably need the same check in filtered index creation for newly created filtered indexes, right?

stacimc:
> @stacimc I just realised that we probably need the same check in filtered index creation for newly created filtered indexes, right?

Very good call! Done.

sarayourfriend (approved):
Nice!

PR #2224, opened 2023-05-30
Use local hook for ESLint pre-commit

### LINKED ISSUE #2200: Run eslint as a local hook in pre-commit configuration
### Discussed in https://github.com/WordPress/openverse/discussions/2197

<div type='discussions-op-text'>

<sup>Originally posted by **sarayourfriend** May 26, 2023</sup>
Right now we use pre-commit to configure ESLint. This requires us to duplicate our ESLint dependencies (and update the versions in two places) by hand. I believe we originally did this so that it would be easy to run ESLint on our several repositories with some shared baseline configuration. However, now that we have all our JavaScript in the monorepo, does this still make sense? I can see that it is probably nice that ESLint can run without needing to run `pnpm install` locally. However, the hook only runs when JavaScript files are changed, so I think it would be reasonable to expect the JavaScript development dependencies to exist locally.

If that is the case, then we could switch the ESLint hook in our pre-commit configuration to a local hook and run ESLint directly. This would remove the need to manually duplicate the dependencies in two places.

Pinging @dhruvkb specifically for advice regarding the monorepo stuff and because you were foundational in getting the nice pre-commit configuration we have now. Does anything about this stand out to you? Can it be done or would it cause other issues that are better avoided?</div>

To do this, move all eslint dependencies into the root package.json of the workspace and hoist all nested eslint configurations into the root eslintrc.js file via `overrides`.

sarayourfriend:
## Fixes

Fixes #2200 by @sarayourfriend 

cc @dhruvkb who was involved in the original discussion that prompted that issue.

## Description

Includes moving ESLint dependencies to the workspace root and hositing all configurations (except the test/unit one, as it is covered in a separate PR) into the root configuration. This also prepares for ESLint "flat config" changes as it unifies the configuration in a way that more closely mirrors how the flat config works. Hositing ESLint configuration to the root along with its dependencies removes the burden of ESLint dependencies on each individual package (if we add additional packages, but also in automations, which does not define any eslint dependencies despite having a "root" config that extends the parent root configuration [which works separately from how ESLint merges cascading configurations]) and makes it possible to share a single ESLint configuration for all JavaScript in all our projects without ever needing to duplicate dependencies or configuration variables.

## Testing Instructions

Run `just lint eslint` locally. CI should pass. ESLint rules should continue to be picked up by your editor if you have them configured and show correctly in the expected places (i.e., overrides should be working).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
This should probably get merged (and rebased) after #2199 so that the ESLint configuration can be fully merged once this PR is applied. I'll leave this as a draft until then.

.eslintrc.js
```diff
@@ -209,6 +210,46 @@ module.exports = {
               "Do not use @vue/test-utils' `shallowMount`. Use `~~/test/unit/test-utils/render` instead which includes helpful context setup or @testing-library/vue's `render` directly.",
           },
         ],
+        "@intlify/vue-i18n/no-raw-text": "off",
```
zackkrida, line 213:
```suggestion
```

Duplicated, see the second line of the "rules" block

.pre-commit-config.yaml
```diff
@@ -145,6 +116,13 @@ repos:
         entry: bash -c 'pnpm exec vue-tsc -p frontend --noEmit'
         language: system
         pass_filenames: false
+      - id: eslint
```
obulat, line 119: Can we do the same for prettier?

obulat (approved):
Thank you for working on this! To be honest, I find the current configuration very complicated and difficult to understand, so this is a great improvement.

Would it make sense to do the same for prettier? If so, would it be better to do it in a separate PR, or this one?
I have a question that is probably not related to the changes in this PR: is there a way to run `lint --fix` locally, without running the pre-commit?

.eslintrc.js
sarayourfriend: Oops, thanks!

.pre-commit-config.yaml
sarayourfriend: I think prettier is a different case. If you check out the original discussion and Dhruv and I had that led to this change, we focused on the fact that ESLint only runs if you're making changes to JS files. And if you're making changes to JS files, the thinking is, then you should have run `pnpm install` (to run unit tests, try the changes out locally, etc). Therefore, ESLint can reliably run as a local hook because the times when it does run are times when the local environment is expected to have `pnpm install` run before committing. Types follow the same logic, I believe, and also wouldn't work without `pnpm install` having been run beforehand.

Prettier, on the other hand, we rely on to format our Markdown files, which includes, for example, the catalog DAG docs, the documentation site, etc. So while the dependencies duplication is annoying for prettier as well, I don't know that there's much we can do to fix it, at least not the same way we can for ESLint, because the conditions under which prettier is depended on do not necessitate `pnpm install`.

We could make `pnpm install`, at least for the workspace root. If there's a way to tell `pnpm` to only install workspace root dependencies, maybe that is a workable assumption, especially if it simplifies our configuration.

In any case, it should be a separate issue. The considerations that went into making this possible for ESLint are different from the ones we need to think about for prettier.

AetherUnbound (approved):
`just lint eslint` runs locally for me, and the changes make sense! Definitely appreciate the more condensed configuration :)

.eslintignore
```diff
@@ -1,13 +1,8 @@
...
-nuxt-template-overrides/
-test/tapes
-coverage/
 node_modules
-.nuxt
-.nuxt-storybook
-!.github
-storybook-static
-.idea
+coverage
```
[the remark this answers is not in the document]
sarayourfriend: Is there a particular `dist` directory that should be excluded? None of the scripts appear to output to a `dist` directory. `pnpm build` outputs to `.nuxt`. I tried to keep the list small and precise. It was getting confusing trying to identify which ignores were unnecessary, which ones were actually valid, etc.

ESLint automatically ignores hidden directories and files, for example, so we don't need to exclude those. And then we only need to exclude directories or files that are inside the particular folders we care about.

I can't find a dist directory so I didn't keep it, but if there is one, I'll add it back.

sarayourfriend:
> Thank you for working on this! To be honest, I find the current configuration very complicated and difficult to understand, so this is a great improvement.

I think the FlatConfig will be a big improvement. Needing to juggle between passing explicit files to include in the command line, plus overrides, plus the displaced eslintignore file is all very confusing! With FlatConfig I think we'd be able to more easily just apply rules to specific directories we actually want ESLint running in rather than having to use three different methods to specify where we want what rules to apply.

I'm pushing one more change to this PR to add some of the root JS configuration files, but otherwise I will merge as is.

.pre-commit-config.yaml
obulat: I see, thank you for explanation!

PR #2141, opened 2023-05-19
Make additional search types flag clearer and add external sources for additional types

obulat:
## Fixes

Fixes #2138 by @obulat

## Description

This PR renames the "external_sources" feature flag to "additional_search_types" as the flag enables showing additional search types in the content switcher.
It also enables showing the additional search type external sources if the flag is enabled, instead of falling back to image on "all_content", "video" and "model_3d" pages.

## Testing Instructions

Go to "/preferences" and turn on the "additional_search_types" flag.
Then select additional search types (video, 3d models) in the content switchers (in the header, in the search bar on the homepage). When you are on a video page, you should see the video source links in the external sources popover, not the image sources.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
[...]

PR #2226, opened 2023-05-30
Hide popover on click outside

### LINKED ISSUE #2220: v-popover.spec test "hideOnClickOutside -> should hide the popover if a click happens outside the popover by default" is broken
## Description

Noticed in https://github.com/WordPress/openverse/pull/2199 

The test is broken. The previously used `expectClosed` assertion did nothing (fixed in the linked PR) and the fixed assertions (also in the linked PR) that try to show that the popover is closed do not work. In fact, the popover does not close in the test. In my debugging, I added a click handler to the external area element and I wasn't able to get that handler to fire. This seems to only effect this particular test case, however, as other test cases with almost identical set up had no issue with the external click not registering. I wonder if it's something related to the issue described in this review comment by @AetherUnbound: https://github.com/WordPress/openverse/pull/2141#pullrequestreview-1442655633

## Reproduction

1.  Run the test mentioned after 2199 is merged (so that the assertions are actually fixed and doing something, rather than nothing) with the test un-skipped and see that it fails.

obulat:
## Fixes

Fixes #2220 by @sarayourfriend

## Description

This PR sets `clickOutsideDeactivates` for the popover to make sure that it is closed when you click outside.

This PR also fixes the issue reported by @AetherUnbound in https://github.com/WordPress/openverse/pull/2141#pullrequestreview-1442655633: when the external sources popover cannot be closed in any way but clicking on the "Close" button.

I also tried rebasing this PR on top of #2199 and removing `.skip` from the test - and the test passes.

## Testing Instructions

Run the app, open the search page, scroll to the bottom of the results and open "External sources" popover. Click outside of the popover - it should close.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/composables/use-focus-on-show.ts
+++ b/frontend/src/composables/use-focus-on-show.ts
@@ -18,6 +18,7 @@ type Props = {
   visibleRef: Ref<boolean>
   autoFocusOnShowRef: Ref<boolean>
   trapFocusRef: Ref<boolean>
+  hideOnClickOutsideRef: Ref<boolean>
   initialFocusElementRef: Ref<HTMLElement | null>
 }
 
@@ -29,6 +30,7 @@ export const useFocusOnShow = ({
   visibleRef,
   autoFocusOnShowRef,
   trapFocusRef,
+  hideOnClickOutsideRef,
   initialFocusElementRef,
 }: Props) => {
   let activateFocusTrap = () => {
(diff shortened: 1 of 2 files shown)


frontend/src/composables/use-focus-on-show.ts
```diff
@@ -48,7 +50,9 @@ export const useFocusOnShow = ({
       // We already do that in a more flexible, adaptive way in our Dialog composables.
       initialFocus: false,
       // if set to true, focus-trap prevents the default for the keyboard event, and we cannot handle it in our composables.
-      escapeDeactivates: false, // Even though we pass `initialFocus` as `false` above, `focus-trap` still
+      escapeDeactivates: false,
+      clickOutsideDeactivates: () => hideOnClickOutsideRef.value,
+      // Even though we pass `initialFocus` as `false` above, `focus-trap` still
```
sarayourfriend, line 55: Should we eventually re-evaluate whether using focus-trap in this way makes sense? Can we either remove it entirely somehow or else rely on it exclusively for focus management of dialogs?

sarayourfriend (approved):
LGTM, tested fine locally.

frontend/src/composables/use-focus-on-show.ts
sarayourfriend: @obulat any advice here? Is there anything we should reconsider here and if so, can it be turned into an issue? If not, I'd just like confirmation for my own understanding, though documenting it in the code would also be helpful and could be its own issue.

PR #2228, opened 2023-05-30
Fix content report form unit test

### LINKED ISSUE #2221: v-content-report-form "should not send other report if description is short" has no assertions
## Description

The test mentioned has no assertions. Caught in #2199. All tests should have assertions.

obulat:
## Fixes

Fixes #2221 by @sarayourfriend

## Description

Changes the test to check if the button is disabled when the text in the form is shorter than the required minimum 20 characters, since it is impossible to send the report anyways when the button is disabled.

## Testing Instructions

The test changes should make sense.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/unit/specs/components/v-content-report-form.spec.js
+++ b/frontend/test/unit/specs/components/v-content-report-form.spec.js
@@ -156,15 +156,20 @@ describe("VContentReportForm", () => {
     })
   })
 
-  // https://github.com/WordPress/openverse/issues/2221
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip("should not send other report if description is short", async () => {
+  it("submit button on other form should only be enabled if input is longer than 20 characters", async () => {
     ReportService.sendReport = jest.fn()
 
     render(VContentReportForm, options)
     await fireEvent.click(getOtherInput())
 
-    const description = "less than 20 chars"
+    // The minimum length for report description is 20 characters. This is invalid
+    const description = "1234567890123456789"
     await fireEvent.update(getDescriptionTextarea(), description)
+
+    // The button is not fully disabled, it uses `aria-disabled` attribute for a11y
+    expect(await getReportButton()).toHaveAttribute("aria-disabled", "true")
+
+    await fireEvent.update(getDescriptionTextarea(), description + "0") // Valid, 20 characters
+    expect(await getReportButton()).not.toHaveAttribute("aria-disabled")
   })
 })
--- a/frontend/test/unit/specs/components/v-content-report-form.spec.js
+++ b/frontend/test/unit/specs/components/v-content-report-form.spec.js
@@ -170,6 +170,6 @@ describe("VContentReportForm", () => {
     expect(await getReportButton()).toHaveAttribute("aria-disabled", "true")
 
     await fireEvent.update(getDescriptionTextarea(), description + "0") // Valid, 20 characters
-    expect(await getReportButton()).not.toHaveAttribute("aria-disabled")
+    expect(getReportButton()).toBeEnabled()
   })
 })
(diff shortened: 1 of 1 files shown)


frontend/test/unit/specs/components/v-content-report-form.spec.js
```diff
@@ -154,13 +154,18 @@ describe("VContentReportForm", () => {
...
     ReportService.sendReport = jest.fn()
 
     render(VContentReportForm, options)
     await fireEvent.click(getOtherInput())
 
-    const description = "less than 20 chars"
+    // The minimum length for report description is 20 characters. This is invalid
+    const description = "1234567890123456789"
     await fireEvent.update(getDescriptionTextarea(), description)
+    expect(await getReportButton()).toHaveAttribute("aria-disabled", "true")
```
sarayourfriend, line 166: `get*` methods are not async, so await isn't necessary here. `@testing-library/jest-dom` also has a matcher for this case: https://github.com/testing-library/jest-dom#tobedisabled

```suggestion
    expect(getReportButton()).toBeDisabled()
```

frontend/test/unit/specs/components/v-content-report-form.spec.js
```diff
@@ -154,13 +154,18 @@ describe("VContentReportForm", () => {
...
+
+    await fireEvent.update(getDescriptionTextarea(), description + "0") // Valid, 20 characters
+    expect(await getReportButton()).not.toHaveAttribute("aria-disabled")
```
sarayourfriend, line 169: Ditto here:

```suggestion
    expect(getReportButton()).toBeEnabled()
```

sarayourfriend (changes requested):
LGTM aside from two small changes to remove the unnecessary `await`s and to use the testing-library matchers. The TL matches cover and check more cases than just the aria attributes and will match the actual state more accurately in general.

AetherUnbound (approved):
Test changes LGTM!

sarayourfriend (changes requested):
@obulat I'm glad I requested those changes. From the latest test run in CI that uses the comprehensive matcher from testing library that takes all a11y contexts into account, the button isn't fully disabled (it should have a `disabled=true`, attribute, I believe).

obulat:
> @obulat I'm glad I requested those changes. From the latest test run in CI that uses the comprehensive matcher from testing library that takes all a11y contexts into account, the button isn't fully disabled (it should have a `disabled=true`, attribute, I believe).

@sarayourfriend, this button is not fully disabled, it only uses `aria-disabled="true"`. Testing library only has a matcher for fully disabled buttons. Based on a discussion within a testing library issue[^1], I believe we should use the previous implementation (`toHaveAttribute("aria-disabled", "true")`) for an `aria-disabled` button. I'll update the code accordingly.

[^1]: https://github.com/testing-library/jest-dom/issues/144

sarayourfriend (commented):
Right! I forgot that there was the case where the button should report as disabled but still be focusable.

Can you add a comment to the test that indicates that? Otherwise it really isn't clear why the test expects `aria-disabled` to be `true` (or not present) but not matchable with `toBeDisabled`. Not blocking on this but I do believe it clarifies the test and makes it easier for future folks to understand why the test is implemented in this way.

frontend/test/unit/specs/components/v-content-report-form.spec.js
```diff
@@ -156,15 +156,20 @@ describe("VContentReportForm", () => {
...
 
     render(VContentReportForm, options)
     await fireEvent.click(getOtherInput())
 
-    const description = "less than 20 chars"
+    // The minimum length for report description is 20 characters. This is invalid
+    const description = "1234567890123456789"
     await fireEvent.update(getDescriptionTextarea(), description)
+
+    // The button is not fully disabled, it uses `aria-disabled` attribute for a11y
```
sarayourfriend, line 169: Nit: can you add the _reason_ for this? This comment doesn't describe _why_ we do that, which is so that the button remains focusable for screen readers to access the context.

Issue #1938, opened 2023-04-30
Update `test_pr_review_reminders` suite to use `pook` instead of bespoke GitHub API request mocking


sarayourfriend:
## Current Situation

`test_pr_review_reminders` mocks the GitHub API class used by the DAG to return specific responses (and save requests made). This makes it difficult to exercise HTTP request fail states like 404s without needing to manually raise a 404 error in the mock code.

## Suggested Improvement

Use `pook` (as used by some Django tests) to mock the actual requests with success or failure responses as appropriate. Pook allows asserting against requests that are made as well, so we can replace the aspects of the tests that assert against a saved list of requests.

## Benefit

Easier to exercise all aspects of the HTTP request lifecycle in the DAG and reduces the amount of scaffolding code we need to maintain to mock those requests.

PR #2243, opened 2023-05-31
Simplify and consolidate photon log handling, improve tallying

### LINKED ISSUE #2225: Photon error handling makes it impossible to fully debug connection issues with Photon
## Description

The error handling strategy applied to the Photon request for thumbnails makes it impossible to get relevant information about the types of errors that occur while communicating with Photon.

https://github.com//WordPress/openverse/blob/08837a8b39b0bb195eac490882b59d92b8da2b00/api/api/utils/photon.py#L146-L181

There are the following problems:

1. We send every thousandth error to Sentry. However, we do this regardless of how often that particular error happened. That means if a new type of catastrophic error starts happening for a particular provider or something like that, it would have to happen at exactly the _thousandth_ request of all requests to ever make its way to Sentry. This means that it is highly likely that we would never see certain exceptional types of problems because the chance of them getting to be the thousandth error that we do send is slim to none.
2. We should disambiguate the error tally by status code or error type. Currently, we just record that an error happened, but not the type of error. A 400 is pretty different from a communication error. We need to be able to know the difference between these and how frequently they are happening to identify whether something is critically wrong or if there is a minor fluke going on. This builds off the first point in that a catastrophic error could be sent as the thousandth error by chance, but we'd have no way of knowing whether it was a common error at the moment because all errors get flattened into a 424 and tallied the same exact way.

## Fix

We should count in Redis the following:

1. Status code responses from Photon (successes and errors, so that we can easily get percentages)
2. If status codes overlap various messages from Photon, disambiguate them with an additional token (i.e., `400:invalid_params` and `400:ssl` etc).
3. For errors that are not HTTPErrors or similar, tally based on the Exception type name

Additionally, send the first 100 and subsequent thousandth _of each distinct type of error_ to Sentry, not just _any_ errors. We shouldn't wait to receive 1000 of a single error type before alerting Sentry, especially not if it is a new type of error we haven't seen before, or a rare one that hasn't happened in a while.

Finally, timestamp these tallies by month so that regressions can be caught and sent to Sentry.

## Additional context

Marked critical because this makes it impossible to properly investigate the issues with our Photon integration: issues so grave they entirely consumed our Sentry quota.

AetherUnbound:
## Fixes

Fixes #2225 by @sarayourfriend

## Description

This PR simplifies and consolidates the photon exception handling by making a few changes:

- If a status code is returned, it is always tallied under `thumbnail_respone_code:{month}:{response_code}` (e.g. `thumbnail_response_code:2023-05:200`, for global response code counts) and `thumbnail_response_code_by_domain:{domain}:{month}:{response_code}` (e.g. `thumbnail_response_code_by_domain:sample.subdomain.org:2023-05:200`, for per-domain counts)
- If an exception is raised, it is always tallied under `thumbnail_error:{exception_name}:{domain}:{month}` (e.g. `thumbnail_error:requests.SSLError:sample.subdomain.org:2023-05`)
- If an HTTP error specifically is raised via the `raise_for_status`, it is _additionally_ tallied under `thumbnail_http_error:{domain}:{month}:{status_code}:{error_text}` (e.g. `thumbnail_http_error:sample.subdomain.org:2023-05:400:Cannot process request`)
- Every exception raised will return a `UpstreamThumbnailException` of the same form
- All initial errors for a particular generic exception (per domain) will be sent to sentry up to the initial threshold (default 100), then every repeated threshold (default 1000) after that.

This has a few upshots:
- The exception handling code is more straightforward
- We are tallying multiple pieces of information per-request, regardless of status
- In practice, an issue that continues to be alerted will alert at higher frequency at the top of the month when a new key is generated (and thus the threshold resets)

I believe this addresses all of the concerns raised in #2225, save for `If status codes overlap various messages from Photon, disambiguate them with an additional token (i.e., 400:invalid_params and 400:ssl etc).` - I could not easily find different error messages which photon might emit besides `We cannot complete this request, remote data could not be fetched`, so I'm hoping that the addition of `error_text` in the redis key will help with differentiating this. I didn't want to presume/prescribe a mapping in case that missed other cases, so that seemed like the easiest catch-all.

## Testing Instructions

I don't know that there's a way to test this locally with the error cases we're seeing in production, since that would require hooking the local API up to the staging/production database. Tests have changed drastically so they should pass. You can also add an intentional exception within the Photon "get" function and check that redis has the values, for example:

1. Add a `raise ValueError("Foo")` after the `requests.get` call and other increments but before the `raise_for_status` call
2. Launch the API and run a search, then open a thumbnail result
3. Run `redis-cli -p 50263` to connect to redis
4. Observe keys like the following: 

```
127.0.0.1:50263> keys thumbnail_*
1) "thumbnail_response_code:2023-05:200"
2) "thumbnail_response_code_by_domain:cdn.stocksnap.io:2023-05:200"
3) "thumbnail_error:builtins.ValueError:cdn.stocksnap.io:2023-05"
```
5. Query for the key results and see that they are counting occurrences:

```
127.0.0.1:50263> get "thumbnail_response_code_by_domain:cdn.stocksnap.io:2023-05:200"
"1"
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/utils/photon.py
```diff
@@ -142,53 +145,34 @@ def get(
...
 
-        return HttpResponse(
-            upstream_response.content,
-            status=res_status,
-            content_type=content_type,
-        )
+    res_status = upstream_response.status_code
+    content_type = upstream_response.headers.get("Content-Type")
+    logger.debug(
+        "Image proxy response " f"status: {res_status}, content-type: {content_type}"
```
dhruvkb, line 171:
```suggestion
        f"Image proxy response status: {res_status}, content-type: {content_type}"
```

api/conf/settings/thumbnails.py
```diff
@@ -16,18 +16,10 @@
...
-    "THUMBNAIL_HTTP_ERROR_PREFIX", default="thumbnail_http_error:"
-)
-
-THUMBNAIL_HTTP_ERROR_THRESHOLD_TO_NOTIFY = config(
-    "THUMBNAIL_HTTP_ERROR_THRESHOLD_TO_NOTIFY", default=1000, cast=int
-)
-
-THUMBNAIL_HTTP_ERROR_THRESHOLD_FOR_REPEATED_NOTIFICATIONS = config(
-    "THUMBNAIL_HTTP_ERROR_THRESHOLD_FOR_REPEATED_NOTIFICATIONS", default=1000, cast=int
+THUMBNAIL_ERROR_REPEATED_ALERT_THRESHOLD = config(
```
dhruvkb, line 23: Nit: `_THRESHOLD` is not the right name for this since it is more of a frequency/interval.

api/test/unit/utils/test_photon.py
```diff
@@ -216,76 +218,133 @@ def raise_exc(*args, **kwargs):
...
+        )
+        == b"1"
+    )
 
-    capture_exception.assert_called_once_with(exc)
 
-    key = f"{settings.THUMBNAIL_TIMEOUT_PREFIX}subdomain.example.com"
-    assert redis.get(key) == b"1"
+alert_count_params = pytest.mark.parametrize(
+    "count_start, should_alert",
```
dhruvkb, line 250: Would count alert be more appropriate? Seeing how the alert count maps 1-1 with the event count till 100, stays there for 900 more, then adds 1 for every thousand?
```suggestion
    "count_start, count_alert",
```

api/test/unit/utils/test_photon.py
```diff
@@ -216,76 +218,133 @@ def raise_exc(*args, **kwargs):
...
+    [
+        (0, True),
+        (1, True),
+        (99, True),
+        (100, False),
+        (999, True),
+        (1000, False),
```
dhruvkb, line 257: Please also include a few tests for values between 1k and 2k, over 2k to count the actual number of alerts emitted.

dhruvkb (approved):
Thanks for taking over on this so quickly. I only have superficial recommendations the core logic LGTM!

obulat (approved):
The changes and the added tests look good.

api/api/utils/photon.py
```diff
@@ -142,53 +145,34 @@ def get(
...
-        if errors >= settings.THUMBNAIL_HTTP_ERROR_THRESHOLD_TO_NOTIFY and (
-            errors % settings.THUMBNAIL_HTTP_ERROR_THRESHOLD_FOR_REPEATED_NOTIFICATIONS
-            == 0
+        upstream_response.raise_for_status()
+    except Exception as exc:
+        exception_name = f"{exc.__class__.__module__}.{exc.__class__.__name__}"
+        key = f"thumbnail_error:{exception_name}:{domain}:{month}"
+        count = cache.incr(key)
+        if count <= settings.THUMBNAIL_ERROR_INITIAL_ALERT_THRESHOLD or (
+            count % settings.THUMBNAIL_ERROR_REPEATED_ALERT_THRESHOLD == 0
```
krysal, line 159: A misunderstanding of the issue description is that we send _every thousandth error to Sentry_ but in reality, this was limiting only HTTP errors, which were the main source of our Sentry events consumption. Every other error type was sent and I think this was the main motivation for the issue. Should we keep the same limit narrowed here?

AetherUnbound: I think this strategy is worth having in place for all exceptions - it's possible that some other issue might arise which is _not_ an HTTP error but still gets raised in the thousands and could make us hit our quota. By specifying a minimum threshold we're still allowing those initial 100 events through (which currently almost all of the alerts in question have <100 besides the HTTP error one) while ensuring Sentry doesn't get overwhelmed.

Especially given that this essentially resets every month and is per domain, I think it's reasonable to keep the threshold logic here for everything.

api/conf/settings/thumbnails.py
AetherUnbound: True, I'll adjust!

api/test/unit/utils/test_photon.py
AetherUnbound: Hmm, I think I find `count_alert` a bit more confusing to use here 😅 The second value describes whether or not the next exception should be alerted to Sentry, rather than an actual count of how many alerts are made.

api/test/unit/utils/test_photon.py
AetherUnbound: I can include higher number cases here as well, but just to clarify, are you also asking for a test which raises several thousand exceptions to check that the Sentry alert is only called a few times after the initial 100? I feel like that behavior is captured by the test range supplied here, but let me know if you feel that a test which goes through those exceptions is more what you were thinking.

api/api/utils/photon.py
```diff
@@ -11,6 +11,8 @@
 import requests
 import sentry_sdk
 
+from api.utils.tallies import get_monthly_timestamp
+
```
sarayourfriend, line 15: Just a small nit/future improvement/consideration (depending on how general we want to think about "tallying", if it's an idea we'll continue to expand and do elsewhere) I wonder if somehow this error handling code could be moved as a separate unit into the tallies module. It would at least make it simpler to read the general Photon request lifecycle without getting muddled down in the error handling.

api/api/utils/photon.py
```diff
@@ -120,6 +122,7 @@ def get(
 ) -> HttpResponse:
     logger = parent_logger.getChild("get")
     cache = django_redis.get_redis_connection("default")
```
sarayourfriend, line 124: Because this and the `tallies` module are operating on the same conceptual level, should this code also use the tallies connection? This would go hand-in-hand with moving the error handling code into the tallies module itself, though, not part of this change.

I think it makes sense to group these conceptually similar observability utilities.

api/test/unit/utils/test_photon.py
```diff
@@ -216,76 +218,136 @@ def raise_exc(*args, **kwargs):
...
+    capture_exception,
+    setup_requests_get_exception,
+    redis,
 ):
-    setup_requests_get_exception(exception)
+    mock_response = MagicMock(spec=requests.Response)
+    mock_response.status_code = status_code
+    mock_response.text = text
+    exc = requests.HTTPError(response=mock_response)
+    setup_requests_get_exception(exc)
```
sarayourfriend, line 327: Nit: these tests don't need to use `setup_requests_get_exception` and can use `pook` without issue. If they used pook, they would be more explicit/easier to understand and precisely match the actual behaviour of requests for these status codes (without encoding/relying on specific behaviour).

IIRC the only reason the tests above this one use `setup_requests_get_exception` is because the timeout testing in `pook` doesn't work exactly as I expected it to... or I ran into an issue there... I can't remember.

However, there are a number of tests that should be converted to use `pook` rather than mocking requests directly and we should have an issue (similar to https://github.com/WordPress/openverse/issues/1938) for refactoring these. Long-term it produces a more stable test suite, I think.

sarayourfriend (approved):
LGTM. Nothing I've mentioned is blocking. Given the nature of the issue I'm going to merge and deploy this. I have also confirmed with @AetherUnbound that she's okay with me doing so.

api/api/utils/photon.py
AetherUnbound: I was thinking about that as well but opted to hold off on that change for now since that's where the previous error counts were going. However, now might be the time to move them to the tallies DB because I've changed the prefixes/key format anyway. Maybe we fast-follow this?

PR #2244, opened 2023-05-31
Convert route guards to middleware

### LINKED ISSUE #2234: Replace route guards with route middleware
## Description
In Nuxt 3, we don't have access to Pinia stores anymore, so we should move code from `onBeforeRouteEnter` to the `middleware`, for example, in 
https://github.com/WordPress/openverse/blob/20949acc9063d8dde629cad1c9941b5ad2ca4050/frontend/src/pages/search.vue#L55-L65

obulat:
## Fixes

Fixes #2234 by @obulat

## Description

This PR converts the route guards (`beforeRouteEnter` etc) to named and typed middleware. This middleware sets the "back to search" link URL on the single result pages, and the search term on the search pages. The "back to search" link is moved to the search store because `route.meta` does not behave the same way in the `middleware` as it did in the route guards.

## Testing Instructions

There are some navigation tests in the CI that should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/middleware/search.ts
+++ b/frontend/src/middleware/search.ts
@@ -1,0 +1,25 @@
+import { useSearchStore } from "~/stores/search"
+
+import type { Middleware } from "@nuxt/types"
+
+export const searchMiddleware: Middleware = ({ redirect, route, $pinia }) => {
+  const {
+    query: { q, searchBy },
+  } = route
+  /**
+   * This anonymous middleware redirects any search without a query to the homepage.
+   * This is meant to block direct access to /search and all sub-routes, with
+   * an exception for the 'creator' filter. The creator filter doesn't send
+   * the search query to the API.
+   */
+  if (!q && !searchBy) return redirect("/")
+  /**
+   * We need to make sure that query `q` exists before checking if it matches
+   * the store searchTerm.
+   */
+  const searchStore = useSearchStore($pinia)
+  const querySearchTerm = Array.isArray(q) ? q[0] : q
+  if (querySearchTerm !== searchStore.searchTerm) {
+    searchStore.setSearchTerm(querySearchTerm)
+  }
+}
--- a/frontend/src/middleware/search.ts
+++ b/frontend/src/middleware/search.ts
@@ -4,15 +4,21 @@ import type { Middleware } from "@nuxt/types"
 
 export const searchMiddleware: Middleware = ({ redirect, route, $pinia }) => {
   const {
-    query: { q, searchBy },
+    query: { q },
   } = route
   /**
-   * This anonymous middleware redirects any search without a query to the homepage.
-   * This is meant to block direct access to /search and all sub-routes, with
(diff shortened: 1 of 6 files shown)


frontend/src/middleware/search.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const searchMiddleware: Middleware = ({ redirect, route, $pinia }) => {
+  const {
+    query: { q, searchBy },
+  } = route
+  /**
+   * This anonymous middleware redirects any search without a query to the homepage.
+   * This is meant to block direct access to /search and all sub-routes, with
+   * an exception for the 'creator' filter. The creator filter doesn't send
+   * the search query to the API.
```
sarayourfriend, line 13: > The creator filter doesn't send the search query to the API.

Is this true? How does it work, then? :open_mouth:

frontend/src/middleware/search.ts
```diff
@@ -0,0 +1,25 @@
+import { useSearchStore } from "~/stores/search"
+
+import type { Middleware } from "@nuxt/types"
+
+export const searchMiddleware: Middleware = ({ redirect, route, $pinia }) => {
+  const {
+    query: { q, searchBy },
+  } = route
+  /**
+   * This anonymous middleware redirects any search without a query to the homepage.
```
sarayourfriend, line 10: What does "anonymous" mean here? Is that a Nuxt distinction?

sarayourfriend (commented):
I am unable to finish reviewing this today, I still need to test it locally. However, the changes look fine to me. I left some questions for clarification of the comments.

frontend/src/middleware/search.ts
obulat: It _was_ anonymous before this change, and now that I've moved it to another file, the comment needs to be updated :) You can see in the changes that previously, `middleware` was a property of the `search.vue` page component.

sarayourfriend (approved):
LGTM, tested locally and search navigation tests, at least the bits I think would be effected by this change. Specifically I tested going back and forth from results to a single result, using the back to results button and the browser's back button. I also tried to reach search without a query and it correctly redirected to the homepage.

Updating the comments would be helpful but not necessarily blocking.

zackkrida (approved):
LGTM once the anonymous middleware comment is updated.

frontend/src/middleware/search.ts
obulat: Thank you for noticing this. I realized that this was implemented incorrectly: even if we search by creator (which we currently can only do by manually updating the URL, the UI for searchBy creator was removed), the frontend URL query parameter is `q`: `?q=cat&searchBy=creator` is converted to `?searchBy=creator&creator=cat` API request query. I've removed the check for `searchBy` parameter, and updated the comment accordingly.

PR #2271, opened 2023-06-01
Use camelCase for i18n strings

### LINKED ISSUE #2059: Replace `kebap-case` within brackets in locale files with `camelCase`
## Problem

We use `kebap-case` in the locale string interpolations that are replaced by `VueI18n`. These should be replaced with `camelCase` to prevent errors in `Nuxt3`'s version of i18n. 

## Description

The keys that use `kebap-case` should be replaced with keys with `camelCase`. Care should be taken to check if they are correctly interpolated in the components.
Before
`Shown because {feat-name} is {feat-state}`
After
`Shown because {featName} is {feat-State}`

## Alternatives

There might be a way to fix the handling of dashes by @nuxjs/i18n 8, but it's probably easier to just replace the strings (there are about 13 of them).

## Additional context

If you update to the version of nuxt i18n that's compatible with Nuxt 3 (8), you'll get the following errors:
```
 ERROR  [unplugin-vue-i18n]  Unterminated closing brace (error code: 7) in /src/locales/en.json                                                                   2:40:38 PM
  target message: Shown because {feat-name} is {feat-state}
  target message path: pref-page.explanation

  1  |  Shown because {feat-name} is {feat-state}
```

obulat:
_Note: The #polyglots team in the Making WP Chat was [notified](https://wordpress.slack.com/archives/C02RP50LK/p1686541009983419?thread_ts=1685545286.270339&cid=C02RP50LK) of this change before the PR was merged._

## Fixes

Fixes #2059 by @obulat

Replaces
#2241 
#2240 
#2231
#2235

## Description

This PR replaces kebap-cased strings with camelCased strings to facilitate the migration of i18n to Nuxt 3. To minimize disturbances to the translators, this PR is combined from the 4 PRs that were split earlier. The strings that change their context in the `.po` file (and we save the Vue i18n key as message context, while the string itself is used as the `id`) are marked as fuzzy in GlotPress.

## Testing Instructions

Run the app. There should be no i18n warnings in the console about missing keys.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
@obulat What do you think about adding `eslint-plugin-json5` and enabling the [camelcase rule](https://eslint.org/docs/latest/rules/camelcase) for `en.json5` and the testing locales? Once we're able to write custom rules we could even enforce that stuff between curly braces (string variables) are also camel cased.

obulat:
> eslint-plugin-json5



> @obulat What do you think about adding `eslint-plugin-json5` and enabling the [camelcase rule](https://eslint.org/docs/latest/rules/camelcase) for `en.json5` and the testing locales? Once we're able to write custom rules we could even enforce that stuff between curly braces (string variables) are also camel cased.

Great idea, @sarayourfriend! I have never added an eslint plugin, so I would have to research for this. Do you think it should be a part of this PR, or a follow-up?

sarayourfriend:
I think it depends on the risk of accidentally adding back non-camel cased strings. If it's pretty bad to accidentally commit a new one after this change goes on, I'd suggest adding it to this PR. If it's not that bad and we can clean it up no problem, then a separate issue and PR would be fine.

To add an ESLint plugin:

1. `pnpm install <plugin name>`
2. Add the plugin to the list of plugins in `eslintrc.js`

Done :)

We'd need to add `json5` to the eslint files filter in the pre-commit configuration and then exclude `test/tapes` as well. In the eslintrc file, we'll also need to add an override that enables the camelcase rule for the `en.json5` and the test locale json files.

obulat:
> We'd need to add `json5` to the eslint files filter in the pre-commit configuration and then exclude `test/tapes` as well. In the eslintrc file, we'll also need to add an override that enables the camelcase rule for the `en.json5` and the test locale json files.

I added the plugin, but need help with this step. I committed the rule in `overrides` in `eslintrc` file, but it doesn't seem to work.

sarayourfriend:
ESLint doesn't include JSON or JSON5 files by default, so they needed to be explicitly passed in `--ext`. Our pre-commit configuration also doesn't include them in the ESLint files filter. The diff below got it working for me locally:


```
diff --git a/.eslintignore b/.eslintignore
index 404560f5e..97de17a7c 100644
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,3 +11,4 @@ node_modules
 !.github
 storybook-static
 .idea
+catalog
\ No newline at end of file
diff --git a/.eslintrc.js b/.eslintrc.js
index d8d9b2a7d..98d7a7e9a 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -201,7 +201,7 @@ module.exports = {
       },
     },
     {
-      files: ["**/*.json5", "test/**/*.json"],
+      files: ["frontend/src/locales/scripts/en.json5", "frontend/test/locales/*.json"],
       rules: {
         "camelcase/properties": "error",
       },
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index b32c1ee5d..97a22265a 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,13 +87,14 @@ repos:
     rev: v8.37.0
     hooks:
       - id: eslint
-        files: \.(js|ts|vue)$
-        exclude: ^utilities/load_testing/.*|frontend/(nuxt-template-overrides/.*|.remake/.*)$ # ESLint raises warnings for ignored files.
+        files: (frontend|automations).*?\.(js|ts|vue|json5|json)$
+        exclude: ^frontend/(nuxt-template-overrides/.*|.remake/.*|test/tapes/.*)$ # ESLint raises warnings for ignored files.
         "types": [file] # ESLint only accepts [javascript] by default.
         args:
           - --ignore-path=.gitignore
           - --ignore-path=.eslintignore
           - --max-warnings=0
+          - --ext=.js,.ts,.vue,.json5,.json
         additional_dependencies:
           - eslint@8.28.0
           - eslint-config-prettier@8.6.0
@@ -110,6 +111,7 @@ repos:
           - eslint-plugin-unicorn@42.0.0
           - eslint-plugin-vue@9.9.0
           - eslint-plugin-vuejs-accessibility@1.1.1
+          - eslint-plugin-json5@0.1.4
 
   # Use the `.prettierignore` and `.prettier.config.js` files to configure project-specific requirements.
```

This will be a lot easier once #2224 is merged. I'll undraft that PR later today now that its parent PR is merged. After that the configuration to include these files will be a bit clearer and won't require duplicating the dependency.

I also changed the targeting of the JSON files to only include the local files in the camel case property names. We don't want our test fixtures to have that rule applied as they need to match the API response.

Other JSON files are returning errors about their file names not being kebab case. We could either disable the kebab case rule for those files by adding another override for JSON that disables that rule (definitely the easier option, the one I would go with) or update those files to match the case expected by our linter rules (way more work, should be a separate issue if we want it, imo).

obulat:
Thank you so much for the diff and suggestions, @sarayourfriend !
I've rebased this PR onto `main` now that #2224 is merged, and added the config to ignore `camelCase` rule for the `json` files as in the first option of your suggestions.
I also realized that camelCase is not the rule we need here, because the documentation says that "This rule looks for any underscores (_) located within the source code. " - and it does not look for `-` dashes. I looked for other ways of adding linting for `json5` property name casing, and found `[eslint-plugin-jsonc](https://ota-meshi.github.io/eslint-plugin-jsonc/)` . It allows us to set the rules for `json` and `json5` files and their properties. There is a conflict between jsonc and prettier json quoting rules, so I disabled the `jsonc` rules.

.pre-commit-config.yaml
```diff
@@ -118,11 +118,11 @@ repos:
         pass_filenames: false
       - id: eslint
         name: eslint
-        files: \.(js|ts|vue)$
+        files: (frontend|automations).*?\.(js|ts|vue|json5|json)$
```
obulat, line 121: Do we need to remove the (frontend|automations) from here?

AetherUnbound (approved):
Changes make sense, I was able to run the frontend locally without any i18n errors!

.pre-commit-config.yaml
sarayourfriend: The `files` setting is used to tell pre-commit when to run ESLint. We only want it to run when there are changes to files that actually fall within ESLint's purview. But this change to add frontend and automations will exclude the repo-root configuration files (prettier, ESLint, and pnpmfile configs).

However, adding `json` and `json5` will start it to match files we don't want the ESLint hook to run for. Off the top of my head I know there are JSON fixtures in both the catalog and API unit tests. So we do need to restrict it at least somewhat more than just any JS file.

Perhaps something along these lines would work: https://regexr.com/7f1i4

It's annoyingly complicated to use a regex for this rather than an array of glob patterns, which would be a little cleaner:

```
[
  "{prettier.config,.eslintrc,pnpmfile}.{cjs,js}",
  "{frontend,automations}/**/*.{js,ts,vue,json,json5}",
]
```

They're effectively the same settings, but the glob is way easier to understand, imo :confused: I may also have overcomplicated that regex. It's hard to figure out how best to express run on any JS/TS/Vue file change in any directory and any JSON file changed only in these directories. Maybe you have some better ideas than me :crossed_fingers:

sarayourfriend:
`eslint-plugin-jsonc` looks awesome, nice find!

.pre-commit-config.yaml
sarayourfriend: ESLint FlatConfig will make this easier only because it will allow us to remove the duplicated file targeting from the ESLint command. Right now the base config for ESLint will apply to _every file it finds_ so we need to tell it only to look in valid places. This forces us to duplicate this configuration when we also want to tell pre-commit to only run ESLint when those particular files are changing.

zackkrida (approved):
LGTM. I had to manually run the `i18n` script to fix the en.json keys, which I think should happen automatically as part of the dev script.

Issue #1566, opened 2022-05-17
Remove duplicated tags


stacimc:
## Problem

Here's an example on the frontend of an image with duplicated tags: https://search-staging.openverse.engineering/image/9fc28cae-ec9b-437a-a960-98f48db2cac8 (`coffee`, `lol`, `lol`).

## Description

We should remove duplicated tags as part of the cleaning steps when loading ingested data.

For backfilling: Non-dated DAGs consume all the data on each run and ought to update old records on the next DAG run. For dated DAGs, we may need to write a backfill query.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

AetherUnbound:
Here's another example: https://openverse.org/image/9ab8329c-8037-4fd4-8d55-8a52eb0dca90 (with a lot more duplicates).

obulat:
@AetherUnbound, the catalog database for the item you shared contains duplicate tags:

```
[...]

PR #2302, opened 2023-06-02
Add DAG to remove Flickr thumbnails

### LINKED ISSUE #1816: Delete unacceptable thumbnails from catalog DB after the image data refresh is finished
## Description

In #1811 the need arose to delete the thumbnails of the following providers: 'rijksmuseum', 'sketchfab', 'sciencemuseum', 'thingiverse' and 'flickr'.

## Description

Make a backup (just in case something goes wrong) and apply the update to nullify those providers' thumbnails:

```sql
UPDATE image SET thumbnail = NULL WHERE provider IN 
	('rijksmuseum', 'sketchfab', 'sciencemuseum', 'thingiverse')

-- Flickr will presumably take much longer so it's suggested to apply it separately
UPDATE image SET thumbnail = NULL WHERE provider='flickr';
```

## Additional context

Related to #675.

krysal:
## Fixes

Fixes #1816 by @krysal

## Description

Flickr is the last and main provider retaining thumbnails that do not fit our requirements for showing in the Openverse UI (mostly on desktop), so here is a DAG to remove them progressively in batches. This should allow other tasks while running and advance steadily. It uses the new TaskFlow API which comes really handy for a DAG like this.

After the DAG has runs successfully, this will allow us to revert #1812 on the API side.

In minor related changes, I also exposed the port of the upstream_db for being able to use UI software like DataGrip or TablePlus, and fixed a shadowing name in the Flickr DAG.

## Testing Instructions

1. Run the normal flicker DAG to ingest some rows.
2. Modify those rows to have some fake thumbnails URLs
```sql
UPDATE image SET thumbnail='https://flickr.com/fake_thumb.JPG' WHERE identifier IN (
	SELECT identifier FROM image WHERE provider='flickr' AND thumbnail IS NULL
	FETCH FIRST 50000 ROWS ONLY
);
```
3. Run the new `flickr_thumbnails_removal` DAG
4. Check in the DB the Flickr rows
```sql
-- This query should return 0
SELECT count(*) FROM image WHERE provider='flickr' AND thumbnail IS NOT NULL;
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/flickr_thumbs_removal.py
+++ b/catalog/dags/flickr_thumbs_removal.py
@@ -50,7 +50,7 @@ def flickr_thumbnails_removal():
             query = dedent(
                 f"""
                 UPDATE image SET thumbnail = NULL WHERE identifier IN
-                (SELECT identifier {select_conditions} FETCH FIRST 10000 ROWS ONLY)
+                (SELECT identifier {select_conditions} FETCH FIRST 10000 ROWS ONLY FOR UPDATE SKIP LOCKED)
                 """
             )
             pg.run(query)
--- a/catalog/dags/flickr_thumbs_removal.py
+++ b/catalog/dags/flickr_thumbs_removal.py
@@ -1,3 +1,7 @@
+"""
+One-time run DAG to remove progressively all the old Flickr thumbnails,
+as they were determined to be unsuitable for the Openverse UI requirements.
+"""
 import logging
 from datetime import timedelta
 from textwrap import dedent
--- a/catalog/dags/flickr_thumbs_removal.py
+++ b/catalog/dags/flickr_thumbs_removal.py
@@ -43,21 +43,25 @@ def flickr_thumbnails_removal():
 
     @task()
     def delete(num_thumbs):
+        log_sql = True
         if num_thumbs == 0:
             logger.info("No Flickr thumbnails found.")
 
         while num_thumbs > 0:
             query = dedent(
                 f"""
-                UPDATE image SET thumbnail = NULL WHERE identifier IN
-                (SELECT identifier {select_conditions} FETCH FIRST 10000 ROWS ONLY FOR UPDATE SKIP LOCKED)
+                UPDATE image SET thumbnail = NULL WHERE identifier IN (
+                    SELECT identifier {select_conditions}
+                    FETCH FIRST 10000 ROWS ONLY FOR UPDATE SKIP LOCKED
+                )
                 """
             )
-            pg.run(query)
+            pg.run(query, log_sql=log_sql)
             num_thumbs -= 10000
             logger.info(
                 f"Flickr thumbnails left: {num_thumbs if num_thumbs > 0 else 0}."
             )
+            log_sql = False
 
     @task()
     def report():
--- a/catalog/dags/flickr_thumbs_removal.py
+++ b/catalog/dags/flickr_thumbs_removal.py
@@ -1,0 +1,72 @@
+import logging
+from datetime import timedelta
+from textwrap import dedent
+
+from airflow.decorators import dag, task
+
+from common.constants import DAG_DEFAULT_ARGS, POSTGRES_CONN_ID
+from common.slack import send_message
+from common.sql import PostgresHook
+
+
+logger = logging.getLogger(__name__)
+
+
+DAG_ID = "flickr_thumbnails_removal"
+
+
+@dag(
+    dag_id=DAG_ID,
+    default_args={
+        **DAG_DEFAULT_ARGS,
+        "retries": 0,
+        "execution_timeout": timedelta(days=7),
+    },
+    schedule=None,
+    catchup=False,
+    doc_md=__doc__,
+)
+def flickr_thumbnails_removal():
+    pg = PostgresHook(postgres_conn_id=POSTGRES_CONN_ID)
+    select_conditions = "FROM image WHERE provider = 'flickr' AND thumbnail IS NOT NULL"
+
+    @task()
+    def count():
+        num_thumbs = pg.get_first(f"SELECT COUNT(*) {select_conditions}")[0]
+        logger.info(f"Flickr thumbnails found: {num_thumbs}.")
+
+        return num_thumbs
+
+    @task()
+    def delete(num_thumbs):
+        if num_thumbs == 0:
+            logger.info("No Flickr thumbnails found.")
+
+        while num_thumbs > 0:
+            query = dedent(
+                f"""
+                UPDATE image SET thumbnail = NULL WHERE identifier IN
+                (SELECT identifier {select_conditions} FETCH FIRST 10000 ROWS ONLY)
+                """
+            )
+            pg.run(query)
+            num_thumbs -= 10000
+            logger.info(
+                f"Flickr thumbnails left: {num_thumbs if num_thumbs > 0 else 0}."
+            )
+
+    @task()
+    def report():
+        msg = (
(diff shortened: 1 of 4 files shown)


stacimc:
I mentioned this on the [popularity refresh project thread](https://github.com/WordPress/openverse/issues/433) but should have also done so on #1816 -- I'm currently working on a reusable `batched_update` DAG that will be used by the popularity refresh DAGs, but can also be run manually for these sorts of backfills without having to spin up temporary DAGs. I plan on having that up by early next week.

The DAG I'm working on uses a slightly different approach for the batched update, which I think is slightly more optimized (about 1.3x as fast during some tests I did on a DB snapshot of production data, although I was only able to run a few tests). I suspect this update is going to be **quite** slow either way 😞 so any performance improvement might really add up. A _full_ update of Flickr needs almost 50k batches, although I actually don't know how many have null thumbnails. Do you have a sense of how many need to be updated?

krysal:
@stacimc By [March 11](https://github.com/WordPress/openverse/issues/675#issuecomment-1464969665), I commented on the following ratio of thumbnails availability for Flickr. Today there could be fewer since that was before my attempt to run the `UPDATE` manually.
```
+---------------+------------------+----------------------+
| provider      | thumbs_available | thumbs_not_available |
|---------------+------------------+----------------------|
| flickr        | 497009314        | 911578               |
+---------------+------------------+----------------------+
```

I'm eager to see the optimizations for this task. However, since it's expected to take some time either way, I would like to start as soon as possible. The DAG I created here can be started and stopped at any time without harm. We can move forward while you prepare the other DAG and it is reviewed without pressure. What do you think?

catalog/dags/flickr_thumbs_removal.py
```diff
@@ -0,0 +1,76 @@
...
+    @task()
+    def delete(num_thumbs):
+        if num_thumbs == 0:
+            logger.info("No Flickr thumbnails found.")
+
+        while num_thumbs > 0:
+            query = dedent(
+                f"""
+                UPDATE image SET thumbnail = NULL WHERE identifier IN
+                (SELECT identifier {select_conditions} FETCH FIRST 10000 ROWS ONLY)
```
dhruvkb, line 53: This was new to me! I only new `LIMIT` till now.

catalog/dags/providers/provider_api_scripts/flickr.py
```diff
@@ -120,9 +120,9 @@ def ingest_records(self, **kwargs):
         for start_ts, end_ts in self.large_batches:
             # For each large batch, ingest records for that interval one license
             # type at a time.
-            for license in LICENSE_INFO.keys():
+            for license_ in LICENSE_INFO.keys():
```
dhruvkb, line 123: Just for my own edification, what is the reason behind this rename?

dhruvkb (approved):
My limited understanding of the catalog notwithstanding, this LGTM!

catalog/dags/providers/provider_api_scripts/flickr.py
krysal: Pycharm was complaining it was shadowing [the built-in name](https://docs.python.org/3/library/constants.html#license).

krysal:
@AetherUnbound I don't think any of those reasons are strong enough to block this particular task. 

> We have the perfect test case for using this new process against production data with the Flickr thumbnails.

Isn't the popularity calculations backfill the main case for @stacimc's DAG?

> I don't know when we'll have another opportunity to run that DAG and see how it works/behaves/formalize its process.

We need to [clean the tags](https://github.com/WordPress/openverse/issues/1566) so that is another opportunity to try the DAG, despite being a bit more complicated due to the type of data.

> It's potentially more optimized, which means the generic DAG should complete sooner.

As I said before, since it's expected to take some time either way would be better to start sooner and make progress, on what we can achieve with this DAG.

> If we wait and use the generic DAG, we don't have to worry about retiring or removing this temporary DAG down the line.

It's a code change so small and easy that it shouldn't count weighing the benefit of gaining time. Tasks related to thumbnails have been delayed for too long.

However, I'll grant on waiting if you both want to use Flickr's thumbnails as a test case.

catalog/dags/flickr_thumbs_removal.py
```diff
@@ -0,0 +1,76 @@
...
+                """
+            )
+            pg.run(query)
```
stacimc, line 56: This will spam the logs with many repeats of the same SQL query being executed over and over. You can pass `log_sql=False` to `PostgresHook` to prevent this, although we should log the query on the first run.

stacimc (changes requested):
> Isn't the popularity calculations backfill the main case for @stacimc's DAG?

I'm writing a `batched_update` DAG that will be used by the popularity refresh DAGs, but which can also be run manually to do arbitrary SQL updates. The goal was for this to remove the need for one-off temporary DAGs for backfills. Its implementation was unfortunately delayed because of the catalog performance investigations.

The DAG is working (and definitely works for this use case), I am just writing tests. It is my priority and should be up by the end of the day tomorrow.

> As I said before, since it's expected to take some time either way would be better to start sooner and make progress, on what we can achieve with this DAG.

I definitely see your point -- at least what I was getting at was that if this DAG is going to be _very_ long running, which may very well be the case, then a delay of a day or two on the PR _might_ actually still be faster. There is certainly no harm in starting this DAG, though.

I was hoping to use the thumbnail update to test out the new DAG, if urgency allows. The v1 implementation is not especially complex to review because of some limitations on dynamic task mapping; the primary difference is in the use of indexed temp tables for updating (which speeds up the inner SELECT per batch), and the configurability of the DAG itself. Of course, I'm not sure how long review will take however.

All that said, if you feel strongly that this should be started today we can go ahead. My only blocking request is the addition of `SKIP LOCKED`.

catalog/dags/flickr_thumbs_removal.py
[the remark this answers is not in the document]
krysal: Oh, that seems perfect for this case! 💯

catalog/dags/flickr_thumbs_removal.py
krysal: Nice

krysal:
Another good reason to try this DAG is to have hard data for effective comparison. So far we have been talking about hypothetical efficiency, but no numbers have been shared. I can't tell where that is coming from until the other DAG is up for review and tried.

stacimc (approved):
> Another good reason to try this DAG is to have hard data for effective comparison. So far we have been talking about hypothetical efficiency, but no numbers have been shared. I can't tell where that is coming from until the other DAG is up for review and tried.

Approving, but for what it's worth we will not be able to compare this in production once the update is finished because this exact update can't be reasonably performed twice. Based on what we have seen in other tests, it will not be especially meaningful to compare it to different updates -- there are too many confounding factors.

The performance test I mentioned earlier was on production data on a test DB instance restored from a production snapshot. The number I gave was 1.3x as fast. I was comparing the performance of updating a single batch. If you are curious it took a little more than an hour (1hr 2min 22sec) with this approach, and 45min 12 sec with the approach in #2331. I gave the relative performance rather than the exact times because in our testing we've seen that production is generally faster than these test instances, so the absolute run time in the tests isn't necessarily predictive, just the relative performance. I am very hopeful that the batches will be faster on prod but we'll have to see; this is one of the reasons I'm very eager to test #2331 soon.

PR #2313, opened 2023-06-04
Use stable identifier for filtered index presence

### LINKED ISSUE #2306: Staggering of ES alias pointing may cause significant discrepancies in sensitivity marker
## Description

I was looking into monitoring concurrency for #2133 when I realized that we may have a non-trivial time delay between when the alias for main document index gets changed and when the filtered index gets changed. As it stands currently, the main index alias for a media type (e.g. `image`) gets pointed from the old index to the new one and then the `create_filtered_{media}_index` DAG is triggered after that. This creates the new filtered index from the new main index, and then points the alias (e.g. `image-filtered`) to the new index.

As I understand it, we use the Elasticsearch `id` value when running a search query to determine the sensitivity based on this: 

https://github.com/WordPress/openverse/blob/91bb41c0190839edb54644ee2222fc12e32329e9/api/api/utils/search_context.py#L27-L32

If I'm understanding this correctly, this would mean that we use the presence of an Elasticsearch `id` value in the `image-filtered` index to determine sensitivity. Given that there is a time gap between when each index alias gets changed, this could mean that the `id` values returned from the new main index will map to completely different documents in the old filtered index (or documents that don't exist at all). This could return incorrect sensitivity information for those results.

I think it might be possible to make one (or both) of the following changes to mitigate this:

1. Wait to remap the main alias until after the filtered index is built & the alias for it is changed. There would still be a time when old and new exist alongside each other, but that gap could be reduced.
1. Use the `identifier` rather than the Elasticsearch `id` for comparisons when determining sensitivity. This may have a performance impact on the query since it's not the `id` value (although it might not, I'm uncertain what the effect would be).

@sarayourfriend is my understanding of the implementation of #2057 and its interaction with the refresh process correct? CC also @WordPress/openverse-api for any additional context on this.

## Fix

This issue will just address the API aspect. The fix for that is to use `identifier` rather than `id` to query the filtered index, therefore using a document identifier that is guaranteed to be stable. Even if we change how the data refresh order goes, this will be a benefit and its the easiest way to quickly fix the issue without needing to change anything in the data refresh or ingestion server, which will be more complicated.

sarayourfriend:
## Fixes

Fixes #2306 by @AetherUnbound 

## Description

Use `identifier.keyword` instead of `id` to query the filtered index.

## Testing Instructions

Unit tests should pass. Try the testing instructions from #2057 as well.

Luckily the test data factories are written in such a way that the unit tests themselves don't need to change at all and still properly test the functionality :tada: 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/search_context.py
+++ b/api/api/utils/search_context.py
@@ -28,7 +28,16 @@ class SearchContext:
 
         filtered_index_search = Search(index=f"{origin_index}-filtered")
         filtered_index_search = filtered_index_search.query(
-            Q("terms", id=[result.id for result in results])
+            # Use `identifier` rather than the document `id` due to
+            # `id` instability between refreshes:
+            # https://github.com/WordPress/openverse/issues/2306
+            # `identifier` is mapped as `text` which will match fuzzily.
+            # Use `identifier.keyword` to match _exactly_
+            # cf: https://github.com/WordPress/openverse/issues/2154
+            Q(
+                "terms",
+                **{"identifier.keyword": [result.identifier for result in results]},
+            )
         )
 
         # The default query size is 10, so we need to slice the query


obulat (approved):
The change makes sense, the CI passes, and everything works locally. Great docs!

stacimc (approved):
LGTM! Nice the tests don't need changes :)

Issue #2089, opened 2023-05-12
Create `popularity_refresh` DAG factory


stacimc:
## Description

Create a `popularity_refresh_dag_factory` similar to the DAG factories for provider and data refresh DAGs. For each `media_type`, it should generate a `<media_type>_popularity_refresh` DAG which does the following:

1. Update the `<media>_popularity_metrics` table to include any newly added
   metrics.
1. Refresh the `<media>_popularity_constants` view to recalculate the popularity
   constants.
   1. This is done `CONCURRENTLY` so that provider DAGs can continue reading
      from the view while it updates.
1. For each unique `provider` in the `<media>_popularity_constants` view,
   generate a `refresh_<provider>_scores` task. The task will run an `UPDATE` of
   the `standardized_popularity` on all records matching that provider which
   were last updated before the task began.
   1. We may consider running the `refresh_<provider>_scores` tasks in parallel
      to speed up the update.
   1. Optionally, we can hard code a `SKIPLIST` of providers that are present in
      the `<media>_popularity_constants` view, but for which we do not want to
      create a refresh task. We currently have some providers (Nappy, Rawpixel,
      Stocksnap) that support popularity data but are not dated, meaning scores
      for **all** of their records will be updated the next time the DAG runs.
      _Note that some of these DAGs are on a `@monthly` schedule however, which
      means skipping them in this DAG could result in delayed recalculation
      time._
1. Report to Slack when the scores have finished updating.

Refer to [this section in the IP](https://docs.openverse.org/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.html#special-considerations-avoiding-deadlocks-and-timeouts) to ensure that the refresh tasks avoid issues with deadlocking and timeouts.

PR #2331, opened 2023-06-06
Add batched_update dag

stacimc:
## Fixes

Related to #2089 by @stacimc 

## Description

This is the first step for #2089. This PR creates a `batched_update` DAG, a reusable DAG which can be used to perform an arbitrary batched update on a Catalog media table, while handling deadlocking and timeout concerns. This DAG will eventually be used by the popularity refresh DAGs, but should also be run manually to handle any Catalog data migration.

Screen Shot 2023-06-07 at 12 02 13 PM

More details about how the update works/prevents deadlocks and timeouts can be found in the [implementation plan](https://docs.openverse.org/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.html#special-considerations-avoiding-deadlocks-and-timeouts
).

To run the DAG, select `Trigger DAG with config...` from the DAG run menu and fill out the DAG conf parameters:

Screen Shot 2023-06-07 at 1 36 33 PM

The parameters are described in help text and an example is provided in the DAG docs.

In order to test this DAG, I pulled some common SQL testing utilities out of `common/loader/test_sql.py` and into a shared `test_utils/sql.py`. This accounts for many of the changes, so the line count is misleading.

### Notes
_Some context about alternative approaches I did not take: feel free to skip this section_

Initially I hoped to use dynamic task mapping to generate a separate task for each batch. Unfortunately Airflow by default [only allows a maximum of 1024 dynamic tasks](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#max-map-length) at a time. As this number gets higher, the scheduler apparently becomes clogged and [Airflow gets extremely slow](https://www.reddit.com/r/dataengineering/comments/zrv612/is_there_an_alternative_for_airflow_for_running/). This is not significant for many of our updates, but for a full Flickr update of >500 million records and with a batch size of ~10k records we would need **50,000 batches**.

A few other disadvantages:
* With dynamic task mapping we also are unable to fail the entire DAG as soon as a single batch fails; potentially many hundreds batches will be attempted, each failing and notifying Slack
* There is overhead in scheduling and triggering tasks which will make this slower

In this v1 approach, the batches are instead updated in a single long-running task. The task is given an extremely long execution timeout, and the practical timeout is configurable by the user through setting the `update_timeout` for any single batch. The biggest cons of this approach are:
* You cannot restart a single batch
* It's not possible (or not trivial, at least) to parallelize in the future (you *can* do this by setting up multiple DagRuns that select a subset of what you want to update, though)

However, the `batch_start` and `resume_update` params (described below with examples) make it trivial to resume a DAG from an arbitrary failure point.

## Testing Instructions

**Test happy path**
Test conf:
```
{
	"query_id": "my_query",
  	"table_name": "image",
  	"select_query": "WHERE provider='flickr'",
  	"update_query": "SET standardized_popularity=standardized_image_popularity(image.provider, image.meta_data)",
	"batch_size": 100,
	"dry_run": false
}
```

* `just init` to load sample data into the catalog.
* `just catalog/pgcli` to open a pgcli session
  * Run `SELECT COUNT(*) FROM image WHERE provider='flickr' AND standardized_image_popularity IS NULL` to verify that you have 2500 such records
  * Run `REFRESH MATERIALIZED VIEW image_popularity_constants;` -- (this is necessary to calculate a popularity constant for flickr, so that the popularity calculation in the next step will work)
* Run the batched_update DAG using the test conf above
* Ensure that all tasks pass, except `notify_failure` which is skipped
  * Verify the logs make sense for each task. `select_rows_to_update` and `notify_before_update` should report 2500 rows, `update_batches` should log the update query once and thereafter log the batch indices as each is processed...
* Verify the actual update worked in pgcli by rerunning `SELECT COUNT(*) FROM image WHERE provider='flickr' AND standardized_image_popularity IS NULL` and seeing that it is now 0

**Test dry run**
Test conf:
```
{
	"query_id": "my_query",
  	"table_name": "image",
  	"select_query": "WHERE provider='flickr'",
  	"update_query": "SET standardized_popularity=null",
	"batch_size": 100,
	"dry_run": true
}
```
Run the DAG with the above conf. This update _would_ reset standardized popularity back to null, but because `dry_run` is True it should not actually perform the update.

After running the DAG, check the logs on the `select_rows_to_update` and `update_batches` steps and verify that you see the `This is a dry run: no SQL will be executed` log. You can also verify manually that the records were not updated.

**Test a select with 0 results**
Test conf:
```
{
	"query_id": "my_query",
  	"table_name": "image",
  	"select_query": "WHERE provider='foo'",
  	"update_query": "SET standardized_popularity=null",
	"batch_size": 100,
	"dry_run": false
}
```

Verify that the DAG passes, updating 0 records.

**Test a bad update**
Test conf:
```
{
	"query_id": "my_query",
  	"table_name": "image",
  	"select_query": "WHERE provider='flickr'",
  	"update_query": "SET foo=null",
	"batch_size": 100,
	"dry_run": false
}
```

Verify that the `update_batches` task fails, the `notify_failure` task reports to Slack, and `drop_temp_table` is NOT run. This is because creating the temp table is a potentially expensive operation, so we do not want to automatically drop it if there's an issue during update; that allows the DAG to be cleared at the `update` step and retried.

Screen Shot 2023-06-07 at 1 37 24 PM

**Test resuming a failed DagRun without recreating temp table**
Test conf:
```
{
	"query_id": "my_query",
  	"table_name": "image",
  	"select_query": "WHERE provider='flickr'",
  	"update_query": "SET title='foo",
	"batch_size": 100,
        "batch_start": 100,
        "resume_update": true,
	"dry_run": false
}
```
Running this configuration has the effect of re-using the temp table created with query id `my_query` from the previous test, and starting from index 100. Since there are 2500 sample records for Flickr, we should end up with 2400 updated records.

Run this configuration and verify that the `select_rows_to_update` and `create_index` rows are skipped, the DAG executes successfully, and that it updates 2400 flickr records before finally dropping the table.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/tests/test_utils/sql.py
```diff
@@ -0,0 +1,92 @@
+import os
```
stacimc, line 1: These were all pulled directly out of `common/loader/test_sql.py`.

obulat (commented):
Thank you for excellent PR description. It's really interesting to read how this change was created.

I tested locally using the provided instructions, and almost all steps worked as expected.

I ran the "bad query" using `dry-run=true` as described in the example but got unexpected results:
```
[2023-06-07, 16:57:08 +03] {batched_update.py:75} INFO - Updating rows with id 0 through 100.
[2023-06-07, 16:57:08 +03] {batched_update.py:38} INFO - This is a dry run: no SQL will be executed. To perform the updates, rerun the DAG with the conf option `'dry_run': false`.
[2023-06-07, 16:57:08 +03] {batched_update.py:42} INFO - 
    UPDATE image
    SET foo=null
    WHERE identifier in (
        SELECT identifier FROM bad_update_rows_to_update
        WHERE row_id > 0 AND row_id <= 100
        FOR UPDATE SKIP LOCKED
    );
    
[2023-06-07, 16:57:08 +03] {batched_update.py:93} INFO - Updated 0 rows. 0 remaining.
[2023-06-07, 16:57:08 +03] {python.py:183} INFO - Done. Returned value was: 0
[2023-06-07, 16:57:08 +03] {taskinstance.py:1368} INFO - Marking task as SUCCESS. dag_id=batched_update, task_id=update_batches, execution_date=20230607T135622, start_date=20230607T135708, end_date=20230607T135708
[2023-06-07, 16:57:08 +03] {local_task_job_runner.py:232} INFO - Task exited with return code 0
[2023-06-07, 16:57:09 +03] {taskinstance.py:2674} INFO - 2 downstream tasks scheduled from follow-on schedule check
```
When I changed the dry_run to be false, the result was the same as described in the instructions.
What is the purpose of dry run? It seems that it cannot catch the errors when running a SQL query.

obulat (approved):

stacimc:
> When I changed the dry_run to be false, the result was the same as described in the instructions.
What is the purpose of dry run? It seems that it cannot catch the errors when running a SQL query.

That's my fault, sorry! I copy-pasted the test params and should have said to test those with `"dry_run": false`. I have updated the instructions.

The purpose of the dry run is to be able to view the generated SQL to see what _will_ be executed, before actually running the queries. This is intended as a fail-safe to let the user visually confirm that the SQL they're about to run is what they intended. If there are SQL errors that aren't caught by visual inspection, those will be handled by the DAG when it is run with `dry_run` disabled, but this gives you an opportunity to double check what you're about to run.

stacimc:
@obulat I had the v2 changes ready and pushed them to this branch since it was still awaiting approval. I re-requested a review since the changes affect the flow of the DAG, but I can move the last commits into a separate PR if you'd prefer :)

I have just added some quick options to allow resuming a DAG from an arbitrary point without recreating the temp table. Description is updated.

krysal (commented):
This is exciting! The form to add the configuration params is neat 👌 I'll test further tomorrow but my first two tries failed. I used the conf from the happy path this trace error:

```python
2023-06-07, 21:33:20 UTC] {taskinstance.py:1847} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/decorators/base.py", line 220, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 181, in execute
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 198, in execute_callable
  File "/opt/airflow/catalog/dags/database/batched_update/batched_update.py", line 108, in update_batches
    while batch_start <= expected_row_count:
TypeError: '<=' not supported between instances of 'NoneType' and 'int'
```

Then on the second try, I used 0 for the `batch_start` (maybe it should be 1?) and it failed in the second task.

```python
[2023-06-07, 21:40:33 UTC] {taskinstance.py:1847} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/decorators/base.py", line 220, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 181, in execute
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 198, in execute_callable
  File "/opt/airflow/catalog/dags/database/batched_update/batched_update.py", line 89, in run_sql
    return postgres.run(query, handler=handler)
  File "/opt/airflow/catalog/dags/common/sql.py", line 79, in run
    return super().run(sql, autocommit, parameters, handler)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/providers/common/sql/hooks/sql.py", line 349, in run
    self._run_command(cur, sql_statement, parameters)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/providers/common/sql/hooks/sql.py", line 380, in _run_command
    cur.execute(sql_statement)
psycopg2.errors.DuplicateTable: relation "test_query_rows_to_update" already exists
```

stacimc:
> I used the conf from the happy path this trace error:

Fixed! Apologies, after adding the new param I did not manually re-test the happy path. The unit tests did not explicitly pass `None` for the batch_start so did not cover this case -- updated as well.

> Then on the second try, I used 0 for the batch_start (maybe it should be 1?) and it failed in the second task.

This is expected if the previous DagRun failed, because we deliberately do not drop the table automatically on failure (since calculating it may be relatively expensive). You can either:
* clear the `update_batches` task on the first, failed DagRun to resume that run where it left off
* start a new dagrun with the same `query_id` and enable the `resume_update` param. This will make the dagrun skip the table creation steps.

obulat:
> @obulat I had the v2 changes ready and pushed them to this branch since it was still awaiting approval. I re-requested a review since the changes affect the flow of the DAG, but I can move the last commits into a separate PR if you'd prefer :)

Thank you for the update! There's no need to change the PR :) I'll test it locally later today

krysal:
@stacimc Could you rebase or merge this branch with `main`? I cannot connect to the upstream DB with a tool other than pgcli because the port is not exposed here.

krysal (approved):
Looks great! Tested the mentioned paths and everything worked as expected. Fantastic idea ⭐️

I only left a pair of non-blocking comments.

PR #2336, opened 2023-06-06
Add a unit test for RIGHT_CLICK_IMAGE

obulat:
## Fixes

Fixes #[issue number] by @[issue author]

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/types/analytics.ts
```diff
@@ -47,7 +47,7 @@ export type Events = {
    *     when not paired with a `GET_MEDIA` event?
    */
   RIGHT_CLICK_IMAGE: {
-    identifier: string
+    id: string
```
sarayourfriend, line 50: I know this isn't something introduced in this PR and that this change is to the match the other events, but it would be good to keep in mind the potential ambiguity here introduced by the fact that `id` and `identifier` are very different for results on the API, ingestion server, and Elasticsearch side. Especially considering that analysis on these events would likely intersect with making queries against Elasticsearch, the API database, or the catalog database, all of which use `identifier` for the information stored in this payload property.

frontend/test/unit/specs/pages/single-image-result-page.spec.js
```diff
@@ -0,0 +1,46 @@
...
+import { useAnalytics } from "~/composables/use-analytics"
+
+import SingleImageResultPage from "~/pages/image/_id/index.vue"
+
+jest.mock("~/composables/use-analytics", () => ({
+  useAnalytics: jest.fn(),
+}))
+describe("SingleImageResultPage", () => {
+  let options
+  let singleImageStore
```
sarayourfriend, line 16: Why is this initialised outside of the test function?

frontend/test/unit/specs/pages/single-image-result-page.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  const id = "123"
+  const sendCustomEventMock = jest.fn()
+  useAnalytics.mockImplementation(() => ({
+    sendCustomEvent: sendCustomEventMock,
+  }))
```
sarayourfriend, line 21: Ditto here. Is it better to do this and then add a `beforeEach` that resets all mocks or to initialise mocks in a `beforeEach`? I always went back and forth but was never sure which was best :thinking: Initialising it once was always confusing to me though, as the mock history leaked between test cases.

sarayourfriend (approved):
LGTM. The id vs identifier comment feels important to me but is a broader consideration for the analytics events configuration. I'd prefer consistency in the name of that particular piece of data across the entire stack to avoid confusion as much as possible, as well as within the events themselves. Nothing to change in this PR, but if it's something you think worth considering a wider change to the events payload data for, let's start a discussion about that.

frontend/src/types/analytics.ts
obulat: >  keep in mind the potential ambiguity here introduced by the fact that id and identifier are very different for results on the API
I'll open a discussion because this is really an important issue.

For the API _results_ and the frontend, it's simple - there's only the `id` field in the frontend that comes from the API responses. The API uses Openverse UUID that is created by the catalog as the response's `id` field (not the Elasticsearch `id`). One place where the `identifier` property is used is in the homepage images json - which I'll open an issue to rename with `id`.

I tried to find how the `identifier` field becomes the API response's `id` field - and couldn't 😮

frontend/test/unit/specs/pages/single-image-result-page.spec.js
obulat: I just wrote what worked and did not go into the details of the order - I always go back and forth on this, too.

PR #2347, opened 2023-06-07
Remove string reference to old s3 bucket

zackkrida:
Removes a string reference to an old s3 bucket in the retired ETL pipeline script. 

This bucket has been reclaimed by another user and was flagged in a security review of the repo. This PR really only exists to prevent further warnings, even though this dead, unused code poses no actual security risk.

### DIFF
--- a/catalog/dags/retired/common/etl/scripts/ExtractCCLinks.py
+++ b/catalog/dags/retired/common/etl/scripts/ExtractCCLinks.py
@@ -22,7 +22,7 @@ from warcio.archiveiterator import ArchiveIterator
 
 
 BUCKET = os.environ.get("COMMONCRAWL_BUCKET", "not_set")
-COMMONSMAPPER_BUCKET = os.environ.get("COMMONSMAPPER_BUCKET", "ov-commonsmapper")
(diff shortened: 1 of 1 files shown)


catalog/dags/retired/common/etl/scripts/ExtractCCLinks.py
```diff
@@ -22,7 +22,7 @@
 
 
 BUCKET = os.environ.get("COMMONCRAWL_BUCKET", "not_set")
-
+COMMONSMAPPER_BUCKET = os.environ.get("COMMONSMAPPER_BUCKET", "ov-commonsmapper")
```
sarayourfriend, line 25: Won't this cause further issues if `ov-commonsmapper` gets claimed? Or is that a bucket we still control? It might be easier to just leave the default out entirely.

zackkrida, line 25:
```suggestion
COMMONSMAPPER_BUCKET = os.environ.get("COMMONSMAPPER_BUCKET", "not_set")
```

stacimc (approved):
LGTM 👍

sarayourfriend (approved):

AetherUnbound (approved):
LGTM!

PR #2350, opened 2023-06-08
Implementation Plan: Nuxt 3 migration

### LINKED ISSUE #2324: Implementation Plan - Nuxt 3 Migration
## Description

This is an issue to create a single Implementation plan for #411.

obulat:
## Fixes

Resolves #2324 by @obulat

## Reviewers
- [ ] @sarayourfriend - for experience with the previous attempts at migrating to Nuxt 3
- [ ] @zackkrida - for experience with initial conversion from Vue SSR to Nuxt

## Description

This is a plan of how I propose to approach Nuxt 3 migration. I am especially curious about ideas on "Nuxt 3 / Vue 3 features in Openverse: discussion" section.  It must be noted that there will be changes to this plan during its implementation due to the unknowns of this process.

Rendered plan: https://docs.openverse.org/_preview/2350/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.html

## Current round

This discussion is currently in the Clarification round.

The deadline for review of this round is TBD.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
+++ b/documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
@@ -167,10 +167,6 @@ There should be no changes to infrastructure.
 
 <!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
 
-### Tools & packages
-
-<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
-
 Nuxt uses `module`s for adding functionality. The table below describes the
 changes that need to be made to the modules we use.
 
--- a/documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
+++ b/documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
@@ -1,0 +1,278 @@
+# 2023-06-01 Implementation Plan: Nuxt 3 Migration
+
+**Author**: @obulat
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend - for extensive experience with JS
+- [ ] @zackkrida - for extensive frontend experience and the previous migration
+      to Nuxt
+
+## Project links
+
+<!-- Enumerate any references to other documents/pages, including milestones and other plans -->
+
+- [Project Thread](https://github.com/WordPress/openverse/issues/411)
+- [Project Proposal](https://docs.openverse.org/projects/proposals/nuxt_3_migration/20230604-project_proposal_nuxt_3_migration.html)
+
+## Overview
+
+<!-- An overview of the implementation plan, if necessary. Save any specific steps for the section(s) below. -->
+
+Nuxt team has created Nuxt bridge to facilitate the migration from Nuxt 2.
+However, there seems to be
+[no support for Nuxt Bridge in Nuxt i18n](https://github.com/nuxt-modules/i18n/discussions/1334#discussioncomment-1967220),
+so the best way forward is to migrate directly to Nuxt 3.
+
+Nuxt 3 migration will require a lot of changes split into several PRs.
+
+One set of changes is required for Nuxt 3, but can also work in the current Nuxt
+2 app. An example of such change is the conversion of `onBeforeRouteEnter` to
+`middleware`: both work well in Nuxt 2, but only `middleware` can access `store`
+in Nuxt 3. These changes are described in the "Changes to main" section below,
+should be merged directly into `main` branch.
+
+The other changes like actually updating the `nuxt` version would break the
+current app. To make sure that we can split these changes into several distinct
+steps without breaking the production app, we should use a separate branch,
+`nuxt3`, that would gradually get more and more features re-enabled.
+
+### Nuxt 3 / Vue 3 features in Openverse: discussion
+
+#### Autoimports
+
+Nuxt 3 auto imports composables and components. Unlike the usual global imports,
+these imports are handled in such a way that only the imports that are actually
+used are left in the production files. The Nuxt team have optimized the way they
+work both in the working app and in the IDEs, so I think we _should_ use the
+auto import feature and not go against the framework and disable it. In the
+current setup, our test suite throws warnings when a child component is not
+imported. We would need to make sure that this does not happen with Nuxt 3.
+
+#### `script setup`
+
+The new way of writing components in Vue 3 is using `script setup` which reduces
+the number of code lines by automatically exporting the variables for use in the
+components. While we can eventually move to using `script setup` throughout the
+app, this conversion is out of scope for this project. We can, however, convert
+some components if they need extensive code changes anyways, or if the behavior
+within them changes significantly and the `script setup` simplifies the
+component. The decision wether to convert components to `script setup` would be
+at the discretion of the PR implementer.
+
+#### Testing
+
+Vue 3 recommends to use Vitest for better performance and integration with
+`vite` (which will replace `webpack` in our setup). I have not had an
+opportunity to test it out, but I think it would make sense to convert our unit
+tests to `Vitest`. The decision on whether to keep using the current
+dependencies or convert to `Vitest` should be based on the ease of conversion.
+
+## Expected Outcomes
+
+<!-- List any succinct expected products from this implementation plan. -->
+
+Openverse frontend uses Nuxt 3, and all current features (including i18n,
+sitemaps, SEO) work.
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+This section describes the known steps that need to be taken to migrate to
+Nuxt 3. However, it is possible that there will be some unknown steps that will
+be discovered during the implementation.
+
+- [ ] Add changes that can be done in Nuxt 2 (see "Changes to main")
+- [ ] Create a `nuxt3` branch with the first PR as described in the "Nuxt 3
+      branch"
+- [ ] Add all other Nuxt 3 changes, as described in the "Nuxt 3 branch"
+- [ ] Re-enable Playwright tests
+- [ ] Make all unit tests pass
+
+### Changes to main
+
+- Deprecation of CJS modules (the ones that use `require` and `module.exports`):
+  - `case` package that we use for `kebab-case`, `title` case and `camelCase`
+    function is CJS. We can either find replacements that use `ESM` modules,
+    vendor in these functions, or use a workaround that `vite` suggests (instead
+    of `const { kebab } = require "case"` use
+    `import casePkg; const { kebab } = casePkg;`)
+  - home gallery uses `require` when importing the images dynamically. There is
+    a small number of them, so we should just move them to `static` folder (or
+    `public` in Nuxt 3) and use static `import`s.
+  - [ ] #2342 Tailwind config should be converted to TypeScript.
+- Moving functionality inside the `setup` function would make it easier to
+  handle with Vue 3/Nuxt 3:
+  - Refactor `search.vue` page: move the route watcher and `fetchMedia`
+    functions to `setup`
+- Refactor image and audio single result pages to show the skeleton page when
+  the "image"/"audio" is `null`. This is required because the way data is
+  fetched changes, and _is_ the correct way it _should_ be set up (instead of
+  not handling the case of `image` being `null` and instead relying on the error
+  page redirect)
+- [x] #2234 Replace `onBeforeRouteEnter` router guards that don't have access to
+      `pinia` stores with router `middleware` that does
+- [x] #2233 Remove `@nuxtjs/style-resources` - this module was used to
+      auto-import `scss` variables into all components. We don't use `scss`
+      anymore
+
+### Changes in Nuxt 3 branch
+
+- The first PR should update the bare minimum to get the app working with Nuxt
+  3:
+  - update `nuxt` itself and `i18n`, `tailwind` and `svg-sprite` modules (and
+    the way they are set up). Remove/comment out other modules, plugins,
+    middleware.
+  - update the Typescript config and node version.
+  - Move the functions that are called at the app set up from
+    `middleware`/`layout`s to `app.vue`.
+  - Convert the layouts and pages to use `NuxtPage` instead of `NuxtChild` and
+    `slot` instead of `Nuxt`. For other necessary changes, look up the Nuxt 3
+    docs.
+  - Use the new way of declaring `async components`
+  - Remove `Vue.set`
+  - Move `key` from items inside a `v-for` to the surrounding `template`. This
+    PR would produce an app version that can be run with the `pnpm dev` or
+    `pnpm build && pnpm start` commands, but would have significantly less
+    functionality and many console errors/warnings, so it would probably be
+    pushed using `--no-verify` git flag.
+- Set up unit tests and add `.skip` to all the tests that are failing.
+  Temporarily remove the setting that disallows error logs in the test console.
+- Correctly set up the head using `useHead` and `i18n`
+- Remove the `@nuxtjs/composition-api` imports, replacing them with the
+  corresponding `useNuxtApp` or `i18n` import. If not yet possible, comment out
+  the import and the functionality that uses it.
+- Remove all `v-on="$listeners"`, making sure that the `on` handlers are still
+  passed to the correct component/element.
+- Remove the `.native` modifiers, and convert `VLink` to use `NuxtLink`. The
+  `VButton` that has `as="VLink"` can be tricky here, so we would need to make
+  sure it works.
+- Rewrite data fetching to use the new `useAsyncFetch` composable.
+- Re-add the modules listed in the "Modules" section.
+
+## Dependencies
+
+### Infrastructure
+
+<!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
+
+There should be no changes to infrastructure.
+
+### Tools & packages
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+### Tools & packages
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+Nuxt uses `module`s for adding functionality. The table below describes the
+changes that need to be made to the modules we use.
+
+### Nuxt module changes
+
+| Module name                        | Change needed                                                                                        | Links                                                                                                                                                    |
+| ---------------------------------- | ---------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `"@nuxt/typescript-build"`         | Delete as TypeScript is natively supported by Nuxt 3                                                 |                                                                                                                                                          |
+| `"@nuxtjs/composition-api/module"` | Delete - natively supported by Nuxt 3                                                                |                                                                                                                                                          |
+| `"portal-vue/nuxt"`                | Delete - natively supported by Vue 3 / Nuxt 3                                                        |                                                                                                                                                          |
+| `"cookie-universal-nuxt"`          | Delete - natively supported by Nuxt 3, [useCookie](https://nuxt.com/docs/api/composables/use-cookie) |                                                                                                                                                          |
+| `"@nuxtjs/svg-sprite"`             | Update to current that supports Nuxt 3                                                               |                                                                                                                                                          |
+| `"@nuxtjs/eslint-module"`          | Update to current that supports Nuxt 3                                                               | https://github.com/nuxt-modules/eslint                                                                                                                   |
+| `"@pinia/nuxt"`                    | Update to current version that supports Nuxt 3                                                       |                                                                                                                                                          |
+| `"@nuxtjs/proxy"`                  | Replace with Nitro routeRules or use `proxyRequest`                                                  | https://stackoverflow.com/questions/76120894/how-do-i-set-proxy-with-baseurl-with-nuxt3-and-nitro-config, https://github.com/nuxt/nuxt/discussions/15907 |
+| `"@nuxtjs/redirect-module"`        | Define Nitro routeRules in `nuxt.config`                                                             | https://nitro.unjs.io/config/#routerules                                                                                                                 |
+| `"@nuxtjs/sentry"`                 | No official support                                                                                  | Workaround in https://github.com/nuxt-community/sentry-module/issues/530                                                                                 |
+| `"vue-plausible"`                  | Update to current version that supports Nuxt 3 or https://github.com/nuxt-modules/plausible          | Note: enabling outbound tracking breaks links with target blank in `vue-plausible`                                                                       |
+| `"@nuxtjs/sitemap"`                | Replace with `nuxt-simple-sitemap`                                                                   | https://github.com/harlan-zw/nuxt-simple-sitemap                                                                                                         |
+
+We should probably add [`nuxt-vitest`](https://github.com/danielroe/nuxt-vitest)
+for testing in Nuxt. The package is under active development, so we should pin
+to the patch version
+
+The migration to Nuxt 3 also means migration from Vue 2 (with added composition
+API) to Vue 3. Below is a list of breaking changes that need to be addressed in
+Openverse.
+
+#### Vue 3 breaking changes
+
+- [ ] `v-model`
+      [changes](https://v3-migration.vuejs.org/breaking-changes/v-model.html)
+- [ ] `key` usage on `template v-for` instead of the components inside that
+      `template`
+- [ ] [`v-on:event.native` removed](https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed.html) -
+      after removing `.native`, check that links work as expected, sending the
+      analytics events
+- [ ] [`$listeners` has been removed / merged into `$attrs`](https://v3-migration.vuejs.org/breaking-changes/listeners-removed)-
+      remove the `$listeners` everywhere and make sure that we have
+      `v-bind="$attrs"` everywhere we had `$listeners` before
+- [ ] [`$attrs` includes `class` & `style`](https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html) -
+      remove `v-bind="$attrs"` from the root element in components where
+      `inheritAttrs="false"`
+- [ ] `Vue.set` is not necessary anymore to keep array/object reactivity
+- [ ] [async components change of syntax](https://v3-migration.vuejs.org/breaking-changes/async-components.html)
+      (for components that use `Comp = () => import("component.vue")`)
+- [ ] [`emits` option](https://v3-migration.vuejs.org/breaking-changes/emits-option.html) -
+      replace the `defineEvents` custom implementation
+
+#### Nuxt 3 breaking changes
+
+- [ ] `NuxtLink` is now a wrapper around both external and internal links, so we
+      can replace our custom `VLink` component with it.
+- [ ] `app.vue` is a single place where we can put the code that needs to run
+      once when the app starts up. Currently, we have some layout code that is
+      duplicated, and `middleware` code that is run before the first load. We
+      should move it to `app.vue`
+- [ ] [Pages and layouts changes](https://nuxt.com/docs/migration/pages-and-layouts)
+
+## Alternatives
+
+<!-- Describe any alternatives considered and why they were not chosen or recommended. -->
+
+Migrate to a non-Nuxt SSR solution for Vue. This would require significant
+changes to code for an unclear benefit.
+
+## Design
+
+<!-- Note any design requirements for this plan. -->
+
+The designs should stay exactly the same.
+
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+The Nuxt 2 changes can be worked on immediately. The changes to Nuxt 3 branch
+can be parallellized after the initial PR.
+
+## Blockers
+
+<!-- What hard blockers exist which might prevent further work on this project? -->
+
+## Accessibility
+
+<!-- Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+
+## Rollback
+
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,278 @@
...
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend - for extensive experience with JS
+- [ ] @zackkrida - for extensive frontend experience and the previous migration
+      to Nuxt
```
obulat, line 14:
```suggestion
- [ ] @sarayourfriend
- [ ] @zackkrida
```

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+| `"@nuxtjs/proxy"`                  | Replace with Nitro routeRules or use `proxyRequest`                                                  | https://stackoverflow.com/questions/76120894/how-do-i-set-proxy-with-baseurl-with-nuxt3-and-nitro-config, https://github.com/nuxt/nuxt/discussions/15907 |
+| `"@nuxtjs/redirect-module"`        | Define Nitro routeRules in `nuxt.config`                                                             | https://nitro.unjs.io/config/#routerules                                                                                                                 |
+| `"@nuxtjs/sentry"`                 | No official support                                                                                  | Workaround in https://github.com/nuxt-community/sentry-module/issues/530                                                                                 |
+| `"vue-plausible"`                  | Update to current version that supports Nuxt 3 or https://github.com/nuxt-modules/plausible          | Note: enabling outbound tracking breaks links with target blank in `vue-plausible`                                                                       |
```
sarayourfriend, line 191: Would each of these that need workarounds be separate issues? I've only looked at the proxy one but it looks certainly complex enough to be its own issue. It also looks like we could implement the workaround for it in Nuxt 2, FWIW, perhaps in conjunction with #2392 which would complicate the current configuration anyway.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+      remove the `$listeners` everywhere and make sure that we have
+      `v-bind="$attrs"` everywhere we had `$listeners` before
+- [ ] [`$attrs` includes `class` & `style`](https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html) -
+      remove `v-bind="$attrs"` from the root element in components where
+      `inheritAttrs="false"`
+- [ ] `Vue.set` is not necessary anymore to keep array/object reactivity
+- [ ] [async components change of syntax](https://v3-migration.vuejs.org/breaking-changes/async-components.html)
+      (for components that use `Comp = () => import("component.vue")`)
+- [ ] [`emits` option](https://v3-migration.vuejs.org/breaking-changes/emits-option.html) -
+      replace the `defineEvents` custom implementation
```
sarayourfriend, line 221: These changes do not replace `defineEvent`, as far as I can tell. The `emits` feature described in the link is actually required for `defineEvent` to work in the first place, the docstring for the function specifies that it's a workaround for TS in Composition API: https://github.com//WordPress/openverse/blob/697f62f01a32cb7fcf2f4a7627650a113cba40da/frontend/src/types/emits.ts#L9-L18

I'm surprised that Vue doesn't have a built-in solution for this though, it's very silly to need to write an empty validation function (maybe one that even gets included in your final bundle, I'm not sure if Nuxt shakes those out in build) just to get types for events.

In any case, the `defineEvent` utility was already developed for the feature linked on the documentation page and I don't see anything there or on the Vue 3 `emits` documentation for ways to define the event types without the strange empty validator function workaround.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+  fetched changes, and _is_ the correct way it _should_ be set up (instead of
+  not handling the case of `image` being `null` and instead relying on the error
+  page redirect)
+- [x] #2234 Replace `onBeforeRouteEnter` router guards that don't have access to
+      `pinia` stores with router `middleware` that does
+- [x] #2233 Remove `@nuxtjs/style-resources` - this module was used to
+      auto-import `scss` variables into all components. We don't use `scss`
+      anymore
+
+### Changes in Nuxt 3 branch
```
sarayourfriend, line 124: We need a step in this to remove the `nuxt-templates-overrides` too, right?

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+- [ ] [`v-on:event.native` removed](https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed.html) -
+      after removing `.native`, check that links work as expected, sending the
+      analytics events
+- [ ] [`$listeners` has been removed / merged into `$attrs`](https://v3-migration.vuejs.org/breaking-changes/listeners-removed)-
+      remove the `$listeners` everywhere and make sure that we have
+      `v-bind="$attrs"` everywhere we had `$listeners` before
+- [ ] [`$attrs` includes `class` & `style`](https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html) -
+      remove `v-bind="$attrs"` from the root element in components where
+      `inheritAttrs="false"`
+- [ ] `Vue.set` is not necessary anymore to keep array/object reactivity
```
sarayourfriend, line 217: Is the current usage of `Vue.set` (only in the feature flags store, as far as I see) able to be removed in Nuxt 2 if we changed the `Vue.set(flag, 'preferredState', ...)` to `this.flags[flagName].preferredState = ...`, etc.?

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+      `template`
+- [ ] [`v-on:event.native` removed](https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed.html) -
+      after removing `.native`, check that links work as expected, sending the
+      analytics events
+- [ ] [`$listeners` has been removed / merged into `$attrs`](https://v3-migration.vuejs.org/breaking-changes/listeners-removed)-
+      remove the `$listeners` everywhere and make sure that we have
+      `v-bind="$attrs"` everywhere we had `$listeners` before
+- [ ] [`$attrs` includes `class` & `style`](https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html) -
+      remove `v-bind="$attrs"` from the root element in components where
+      `inheritAttrs="false"`
```
sarayourfriend, line 216: Can you give more details for why this is the modification required to handle this breaking change? I'm not making the connection between `inheritAttrs="false"` and the class and style attributes being included in `$attrs`.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+  3:
+  - update `nuxt` itself and `i18n`, `tailwind` and `svg-sprite` modules (and
+    the way they are set up). Remove/comment out other modules, plugins,
+    middleware.
+  - update the Typescript config and node version.
+  - Move the functions that are called at the app set up from
+    `middleware`/`layout`s to `app.vue`.
+  - Convert the layouts and pages to use `NuxtPage` instead of `NuxtChild` and
+    `slot` instead of `Nuxt`. For other necessary changes, look up the Nuxt 3
+    docs.
```
sarayourfriend, line 136: Would it be possible to describe the necessary Nuxt 3 changes here the way the Vue 3 changes are described?

sarayourfriend (commented):
This looks good to me so far, and it gives a decent sense of how much we have to do for this. I've left a handful of questions for clarification, but I would also like to request a clearer/more explicit list of the required changes, ordered as though could be converted to issues and perhaps with links to individual sections for ones that need more explanation. Splitting the changes between the modules section, the Vue 3 breaking changes section, and the implementation stage sections makes it hard to get a general overview of how the implementation would go. I appreciate the two implementation stages being explicit, though. How much benefit is there to pushing as much as we can forward into the pre-`nuxt3` branch phase? Seems like it would prevent a lot more merge conflicts, right? Also, should we put a freeze on frontend feature development during the Nuxt 3 conversion phase, with all frontend development focused exclusively on Nuxt 3 conversion? It would be frustrating to have the Nuxt 3 branch almost perfectly working only to run into big feature related merge conflicts (like #2118  and #2126, which haven't started implementation yet).

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### Autoimports
+
+Nuxt 3 auto imports composables and components. Unlike the usual global imports,
+these imports are handled in such a way that only the imports that are actually
+used are left in the production files. The Nuxt team have optimized the way they
+work both in the working app and in the IDEs, so I think we _should_ use the
+auto import feature and not go against the framework and disable it. In the
+current setup, our test suite throws warnings when a child component is not
+imported. We would need to make sure that this does not happen with Nuxt 3.
```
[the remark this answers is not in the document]
obulat: I anticipated a similar response :) Thank you for describing the reasons in such a detail, it's really helpful.

On the first point, I looked at the Nuxt 3 draft branch locally to see if the IDE support helps with it. And even though I can click "Go to definition" on a non-imported component and open the component file, I can't do the same for a composable. Going to definition there sends you to the file that lists all of the auto-imported items, so while it _is_ possible to find the source of the composable, it's not as straightforward as when we import it. 
> Variable name collision and shadowing is also much harder to understand or anticipate side effects for.
This is also a serious consideration.

The second point - I guess you can't understand the extent of how bad it can be until that happens to you :)
I agree with the third point, too.

What I wanted to avoid was going against the framework. This would cause us to have to add work-arounds around the magic, which might need a lot of work. However, seeing that the IDE support is not as great as I was expecting, and considering all of the drawbacks of using auto-imports, I'm open to _not_ using this feature.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+component. The decision wether to convert components to `script setup` would be
+at the discretion of the PR implementer.
+
+#### Testing
+
+Vue 3 recommends to use Vitest for better performance and integration with
+`vite` (which will replace `webpack` in our setup). I have not had an
+opportunity to test it out, but I think it would make sense to convert our unit
+tests to `Vitest`. The decision on whether to keep using the current
+dependencies or convert to `Vitest` should be based on the ease of conversion.
```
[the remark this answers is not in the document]
obulat: I've found [`eslint-plugin-vitest`](https://github.com/veritem/eslint-plugin-vitest#credits) which says that most of its rules are essentially the ports of `eslint-plugin-jest` with necessary modifications.
I'll add it here.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+  - update `nuxt` itself and `i18n`, `tailwind` and `svg-sprite` modules (and
+    the way they are set up). Remove/comment out other modules, plugins,
+    middleware.
+  - update the Typescript config and node version.
+  - Move the functions that are called at the app set up from
+    `middleware`/`layout`s to `app.vue`.
+  - Convert the layouts and pages to use `NuxtPage` instead of `NuxtChild` and
+    `slot` instead of `Nuxt`. For other necessary changes, look up the Nuxt 3
+    docs.
+  - Use the new way of declaring `async components`
```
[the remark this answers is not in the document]
obulat: The banners inside the `VBanner`, and headers in the `search-layout`.
https://github.com/WordPress/openverse/blob/41b0cdb37d616b5e53ee45f77f0448b07747e04f/frontend/src/components/VBanner/VBanners.vue#L34-L38

https://github.com/WordPress/openverse/blob/41b0cdb37d616b5e53ee45f77f0448b07747e04f/frontend/src/layouts/search-layout.vue#L73-L75
I'll add this to the plan

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+    pushed using `--no-verify` git flag.
+- Set up unit tests and add `.skip` to all the tests that are failing.
+  Temporarily remove the setting that disallows error logs in the test console.
+- Correctly set up the head using `useHead` and `i18n`
+- Remove the `@nuxtjs/composition-api` imports, replacing them with the
+  corresponding `useNuxtApp` or `i18n` import. If not yet possible, comment out
+  the import and the functionality that uses it.
+- Remove all `v-on="$listeners"`, making sure that the `on` handlers are still
+  passed to the correct component/element.
+- Remove the `.native` modifiers, and convert `VLink` to use `NuxtLink`. The
```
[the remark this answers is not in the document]
obulat: New `NuxtLink` is basically what our `VLink` is: it handles both internal and external links: https://nuxt.com/docs/api/components/nuxt-link , so we could actually replace `VLink` with `NuxtLink`.

When trying to migrate to Nuxt 3 locally, I had some problems with the buttons that use `as="VLink"`, they wouldn't navigate correctly. I'm not sure if it's the problem with `NuxtLink` or the overall pages setup, though.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
obulat: Vue 2 cannot detect property addition or deletion in objects. In the feature flags store, we add a "preferredState" property to the object loaded from JSON. So, to make it reactive, we have to use `Vue.set`[^1] in Vue 2 (and Nuxt 2, respectively).

`this.flags[flagName].preferredState` - would be adding a new property to `this.flags[flagName]`, so this new property would not be reactive.

[^1]: https://v2.vuejs.org/v2/guide/reactivity.html#For-Objects

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,277 @@
...
+- Rewrite data fetching to use the new `useAsyncFetch` composable.
+- Re-add the modules listed in the "Modules" section.
+
+## Dependencies
+
+### Infrastructure
+
+<!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
+
+There should be no changes to infrastructure.
```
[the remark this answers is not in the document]
obulat: I don't really know.

Here are two things I can say when looking through the Docker file:
The `static` will need to be changed to `/` here:
```
RUN echo "{\"release\":\"${SEMANTIC_VERSION}\"}" > /home/node/frontend/src/static/version.json
```

PORT should be set correctly for the production build (I'm not yet sure what exactly "correctly" means here)

I'll add a sentence "Test that the Docker build works as expected".

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
obulat: Yes, separating the workarounds into separate issues is a good idea. Adding modules can also be worked on in parallel.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
obulat: I was actually thinking of `defineEmits`: https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits, more on typing emits: https://vuejs.org/guide/typescript/composition-api.html#typing-component-emits

However, they are only available in components that use `script setup`. 

I think the best course here would be to keep using `defineEvent` with `emits`, and refactor the components to `script setup` and `defineEmits` later, after this project is done. It could be a separate milestone for converting all of the components that emit events to `script setup`.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
obulat: Right!

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
sarayourfriend: Ahhhh, I see, I didn't realise there was a difference between Nuxt 3 `NuxtLink` and was confused because we already used `NuxtLink` :sweat_smile: The difference might be worth clarifying, especially "so we could actually replace VLink with NuxtLink", which is awesome.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
sarayourfriend: > I think the best course here would be to keep using defineEvent with emits, and refactor the components to script setup and defineEmits later, after this project is done. It could be a separate milestone for converting all of the components that emit events to script setup.

Sounds good to me! I'm now remembering that I wrote `defineEvent` in frustration that `defineEmits` didn't work in traditional components :sweat_smile:

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
sarayourfriend: That's fantastic :rocket:

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
obulat: After writing this comment, I decided to check if we would have problems with the imports and tests. In the current Nuxt 2 app, the components are auto-imported, so the app can run even if we don't add all component imports. However, the tests show warnings for components that weren't imported.

In my first try, this was a big problem with explicitly adding all of the imports. The biggest problem is the imports from `.nuxt` and the imports that start with `#`, especially deep within modules (`nuxtjs/svg-sprite` being the one I tried to work around the most). Not saying that this is a blocker, but just noting here that getting all of the imports to work both in the prod app and the tests would require a lot of debugging.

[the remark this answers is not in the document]
sarayourfriend: > And even though I can click "Go to definition" on a non-imported component and open the component file, I can't do the same for a composable. Going to definition there sends you to the file that lists all of the auto-imported items, so while it is possible to find the source of the composable, it's not as straightforward as when we import it.

This reminds me actually, that Nuxt 3's auto-imports I believe is the main reason why running TypeScript _requires_ a build. Nuxt 3's build process pulls the auto-imported composables out and generates the tsconfig.json to include all the ambient types that you're meant to extend in your Nuxt app's tsconfig.json.

I wonder how much all those ambient types slow down TypeScript (or perhaps increase its memory footprint).

In either case, we'd need to change CI to run the types check only after a Nuxt 3 build is run, at least if that process is still the same, which sounds to be the case based on the IDE behaviour of go-to-definition for the composables.

> Not saying that this is a blocker, but just noting here that getting all of the imports to work both in the prod app and the tests would require a lot of debugging

That's a shame :slightly_frowning_face:. When I tried to convert Nuxt using bridge, the option to disable auto-imports didn't even work. Does turning off auto-imports stop component auto-importing in the app build? The Nuxt 3 only reference auto-importing for _composables_, both in the composables documentation and in the auto-imports documentation :thinking:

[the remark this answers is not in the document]
obulat: > In either case, we'd need to change CI to run the types check only after a Nuxt 3 build is run, at least if that process is still the same, which sounds to be the case based on the IDE behaviour of go-to-definition for the composables.

I will add this to the proposal, thanks! (Check if we need to change CI to run the types check after a Nuxt 3 build is run - change if necessary)

> The Nuxt 3 only reference auto-importing for composables, both in the composables documentation and in the auto-imports documentation 🤔
I'm not sure if I understand this sentence correctly, but Nuxt 3 docs do say that they auto import components, composables and even utils (although the way it's worded is not very straightforward): https://nuxt.com/docs/guide/concepts/auto-imports#directory-based-auto-imports

By the way, I didn't realize that the recommended way to explicitly add imports is by using `#imports`, found while re-reading the docs for this comment.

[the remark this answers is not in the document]
sarayourfriend: > Check if we need to change CI to run the types check after a Nuxt 3 build is run - change if necessary

This would also need to change in the pre-commit, which is probably the place it will hurt the DevEx the most.

zackkrida:
As I shared with Olga, I'll review this today.

obulat:
I've updated the plan to add the suggestions based on @sarayourfriend's comments. I removed the "Breaking changes" section for Vue and Nuxt. Instead, I added the steps of what we need to update with the note if it's a "Vue 3 breaking change". I also tried to add links to documentation to each step.
Some items will need to be edited for clarity, I'll try to do it tomorrow.

sarayourfriend:
@zackkrida Will you be able to review this still or are you planning to wait until after the revisions?

zackkrida:
I'll be reviewing this prior to any additional revisions. I did want until the _previous_ revisions were made, after chatting about them with Olga.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Changes to main
+
+- Deprecation of CJS modules (the ones that use `require` and `module.exports`):
+
+  - [ ] `case` package that we use for `kebab-case`, `title` case and
+        `camelCase` function is CJS. We can either find replacements that use
+        `ESM` modules, vendor in these functions, or use a workaround that
+        `vite` suggests (instead of `const { kebab } = require "case"` use
+        `import casePkg; const { kebab } = casePkg;`)
```
zackkrida, line 105: I think we could just vendor this and recommend that, rather than listing multiple potential options here.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,338 @@
...
+| `"@nuxtjs/sentry"`                 | No official support                                                                                     | Workaround in https://github.com/nuxt-community/sentry-module/issues/530                                                                                 |
+| `"vue-plausible"`                  | ~~Update to current version that supports Nuxt 3 or~~ Move to https://github.com/nuxt-modules/plausible | Note: enabling outbound tracking breaks links with target blank in `vue-plausible`                                                                       |
+| `"@nuxtjs/sitemap"`                | Replace with `nuxt-simple-sitemap`                                                                      | https://github.com/harlan-zw/nuxt-simple-sitemap                                                                                                         |
+
+We should probably add [`nuxt-vitest`](https://github.com/danielroe/nuxt-vitest)
+for testing in Nuxt. The package is under active development, so we should pin
+to the patch version
```
zackkrida, line 291:
```suggestion
to the patch version.
```

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
zackkrida: I definitely understand the desire to do what's idiomatic for the framework _and_ absolutely hating this automagical import stuff. 

> Does turning off auto-imports stop component auto-importing in the app build? 

I just tested this with a fresh Nuxt 3 install. Sadly, components still appear to be auto imported even with this config:

```ts
export default defineNuxtConfig({
  imports: {
    autoImport: false
  }
})
```

It _does_ work for composeables, but not components. 

I'm really conflicted on how to proceed for this one.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
zackkrida: Vitest maintains a migration doc that might be useful to link here:

https://vitest.dev/guide/migration.html#migrating-from-jest

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,338 @@
...
+      when the "image"/"audio" is `null`. This is required because the way data
+      is fetched changes, and _is_ the correct way it _should_ be set up
+      (instead of not handling the case of `image` being `null` and instead
+      relying on the error page redirect)
+- [ ] Refactor skip-to-content functionality. Nuxt 3 only supports teleports in
+      ClientOnly, so the current implementation is causing server-client
+      mismatch and rehydration. We should move to a simple HTML anchor tag and
+      an `id` on the content.
+- [ ] Refactor the functionality that is called on app load and route
+      navigation.
```
zackkrida, line 121: Could you elaborate on what this entails? It feels really vague and I'm not sure what it refers to.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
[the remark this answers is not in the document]
sarayourfriend: Actually, if it auto-imports components, I think that's fine. That's the current behaviour of Nuxt 2, so it would present effectively zero change to our workflow.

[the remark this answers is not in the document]
zackkrida: Oh, true!

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,338 @@
...
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+The Nuxt 2 changes can be worked on immediately. The changes to Nuxt 3 branch
+can be parallellized after the initial PR.
+
+## Blockers
+
+<!-- What hard blockers exist which might prevent further work on this project? -->
```
zackkrida, line 316: Team capacity may be a legitimate concern here. At the very least, we would have to pull maintainers from other parts of the project if we plan to do this in Q3-Q4 of this year.

documentation/projects/proposals/nuxt_3_migration/20230608-implementation_plan_nuxt_3_migration.md
```diff
@@ -0,0 +1,338 @@
...
+<!-- Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+
+## Rollback
+
+<!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
+
+Since Nuxt 3 specific features will be in a separate branch, if there is a need
+to roll back after merging `nuxt3` branch, we can revert the merge commit.
+
+## Risks
```
zackkrida, line 329: Some other risks I can think of:

1. Drift between the `nuxt3` and `main` branches, or difficulty keeping them in sync
2. Encountering an unexpected dependency incompatability that halts the migration, further exacerbating pt.1

zackkrida:
This is an excellent breakdown of the sheer mass and complexity of work required to execute this migration. I have to admit I find it terrifying. Particularly the lack of stability and support in many of the ecosystem packages: i18n, nuxt-vitest, storybook (there's a $500 bounty in [this issue](https://github.com/nuxt-community/storybook/issues/330), sentry, and I imagine many more. It seems the only thing that might bring some stability to these dependencies is more time, as many of them are in active development. I worry that the list of changes here represents our known issues, that the list of known issues is quite sizable, and that the list of _unknown_ issues may be even larger. 

Given that this is no longer time sensitive, as #2430 adds Node 18.x support which will be maintained far into 2025, I wonder if an ideal course of action may be to focus on the prerequisite changes, but also any work to audit and clean up the frontend as much as possible beyond what is necessary for Nuxt 3. And to explicitly revisit the Nuxt 3 part of the project several months from now when some of these libraries might have better support.

It's a fine line between delaying the inevitable and waiting for dependencies to make improvements. 

It's also worth considering that two frontend contributors, in fact the reviewers of this proposal, will either be AFK entierely or have reduced capacity at the time of year we might consider implementing this work. 

I could also see the Nuxt 3 steps and branch dragging on for a really long time. I think the workflow to fix these problems could benefit from some pair programming sessions. So often this work involves untangling cryptic dependency errors.

I think I'd like to wait and give this a final approval once any remaining revisions are in place, but the overall approach and breakdown of the tasks makes a lot of sense to me. I find myself really distracted by how gargantuan of an undertaking this will be.

sarayourfriend:
> Given that this is no longer time sensitive, as https://github.com/WordPress/openverse/pull/2430 adds Node 18.x support which will be maintained far into 2025, I wonder if an ideal course of action may be to focus on the prerequisite changes, but also any work to audit and clean up the frontend as much as possible beyond what is necessary for Nuxt 3. And to explicitly revisit the Nuxt 3 part of the project several months from now when some of these libraries might have better support.

I strongly support this option now that Node 16 EOL is no longer a factor, especially when [Elasticsearch 17 EOL](https://www.elastic.co/support/eol) is also approaching at the same time we were originally planning to implement this, as @obulat (IIRC, might have been Krystle) pointed out elsewhere.

krysal:
@obulat Should this be marked as blocked until #2490 is merged? The project thread says it depends on some TypeScript updates. Could you leave a comment on what is needed to complete this plan?

obulat:
Closing this PR until the Additional search views project is done. Then, I'll re-open it with the updated status and plan.

PR #2358, opened 2023-06-08
Implementation Plan: Staging Elasticsearch Reindex DAGs

### LINKED ISSUE #1987: Implementation Plan: Staging Elasticsearch reindex DAGs for both potential index types

krysal:
## Fixes
Resolves #1987 by @AetherUnbound 

## Assigned reviewers

- [x] @AetherUnbound (Project lead)
- [x] @sarayourfriend (for Elasticsearch expertise and familiarity with the proposed implementation)

## Current round

> **Note**
> This discussion is following the Openverse decision-making process. Information about this process can be found
on the [Openverse documentation site](https://docs.openverse.org/meta/decision_making/index.html). Requested reviewers or participants will be following this process. If you are being asked to give input on a specific detail, you do not need to familiarise yourself with the process and follow it.

The discussion is currently in the [Revision round](https://docs.openverse.org/meta/decision_making/process_description.html#revision-round).

The deadline for review of this round is **2023-06-27**.

### DIFF
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -93,7 +93,7 @@ create the index.
    step.
 5. If the index is aliased then the DAG checks if `delete_old_if_aliased=True`
    and proceeds to run [`indices.delete`][es_py_delete] with the old index name.
-   Otherwise the DAG ends at the previous step.
+   Otherwise, the DAG ends at the previous step.
 
 [reindex]:
   https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L282
@@ -126,55 +126,62 @@ Reindex][es_reindex_api] API.
 
 #### Parameters
 
-2. `source_index`: (Optional) The existing index on Elasticsearch to use as the
+1. `source_index`: (Optional) The existing index on Elasticsearch to use as the
    basis for the new index. If not provided, the index aliased to
    `<media_type>-full` will be used.
-3. `percentage_of_production`: The proportion of items to take from each
-   provider from the total amount existing in production.
+2. `percentage_of_prod`: A float indicating the proportion of items to take from
+   each provider from the total amount existing in production. E.g. `0.25` for a
+   quarter of the production documents.
 
 #### Outlined Steps
 
 1. Get the list of media count by sources from the production Openverse API
-   [`https://api.openverse.engineering/v1/<media_type>/stats/`](https://api.openverse.engineering/v1/<media_type>/stats/)
-2. Calculate the total media adding up all the counts by provider
-3. Calculate the name of the new index with the following format:
+   `https://api.openverse.engineering/v1/<media_type>/stats/`
+2. Calculate the `total_media` adding up all the counts by provider
+3. Build the `dest_index` name with the following format:
 
 ```python
- f"{media_type}-{percentage_of_production}-percent-of-providers-{current_datetime}"
+ f"{media_type}-{percentage_of_prod}-percent-of-providers-{current_datetime}"
 ```
 
-4. Make a dictionary mapping all the providers with the required amount of items
-   for the new index based on the provided `percentage_of_production` param.
-5. Iterate over the items of the resulting dictionary to index the subset of
-   each provider.
+4. Make a a list of dictionaries mapping all the providers with the required
+   amount of items for the new index based on the provided `percentage_of_prod`
+   param.
+5. Using the [Dynamic Task Mapping][airflow_dtm] feature of Airflow, expand the
+   providers list to dispatch [`reindex`][es_py_reindex] tasks with the
+   following params setting to index the subset of each provider in parallel.
 
-```
-POST _reindex?wait_for_completion=false
-```
-
-```json
-{
-  "max_docs": num_items,
-  "source": {
-    "index": "image-full",
-    "query": {
-      "term": {
-        "source.keyword": "stocksnap"
+   - `wait_for_completion=False`
+   - `max_docs=docs_num`
+   - `dest=dest_index`
+   - ```python
+      source={
+         "index": source_index,
+         "query": {
+            "term": {
+               "source.keyword": provider
+            }
+         }
       }
-    }
-  },
-  "dest": {
-    "index": "image-50-percent-of-providers-20230608183000"
-  }
-}
-```
+     ```
 
-6. Make the alias `<media_type>-subset-by-provider` point to the new index.
-7. Optionally. Query the stats of the resulting index and print the results.
+6. Avoiding to `wait_for_completion` will make the previous step return records
+   of these tasks as documents in an aggregated form. Then here use Sensors and
+   the previously emitted task IDs to wait for reindex tasks to complete.
+7. Once all tasks are finished, trigger an [`indices.refresh`][es_py_refresh] to
+   make the index queyrable.
+8. Make the alias `<media_type>-subset-by-provider` point to the new index.
+   Follow the same procedure to that of `<media_type>-full` alias of the
+   previous DAG.
+9. Optionally. Query the [stats][es_py_stats] of the resulting index and print
+   the results.
 
-```
-GET /image-reindexed-by-provider/_stats
-```
+[es_py_reindex]:
+  https://elasticsearch-py.readthedocs.io/en/v8.8.0/api.html#elasticsearch.Elasticsearch.reindex
+[es_py_refresh]:
+  https://elasticsearch-py.readthedocs.io/en/v8.8.0/api.html#elasticsearch.client.IndicesClient.refresh
+[es_py_stats]:
+  https://elasticsearch-py.readthedocs.io/en/v8.8.0/api.html#elasticsearch.client.IndicesClient.stats
 
 ## Alternatives
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -178,6 +178,8 @@ Reindex][es_reindex_api] API.
 9. Optionally. Query the [stats][es_py_stats] of the resulting index and print
    the results.
 
+[airflow_dtm]:
+  https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dynamic-task-mapping.html
 [es_py_reindex]:
   https://elasticsearch-py.readthedocs.io/en/v8.8.0/api.html#elasticsearch.Elasticsearch.reindex
 [es_py_refresh]:
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -10,7 +10,7 @@
 ## Project links
 
 - [Project Thread](https://github.com/WordPress/openverse/issues/392)
-- [Project Proposal](./20230331-project_proposal_search_relevancy_sandbox.md)
+- [Project Proposal](/projects/proposals/search_relevancy_sandbox/20230331-project_proposal_search_relevancy_sandbox.md)
 
 ## Overview
 
@@ -19,7 +19,7 @@ creation ––full and proportional-by-provider–– which will allow us to de
 the process from the long Ingestion server's data refresh process and experiment
 with smaller indices. Also includes the adoption of two new index aliases for
 ease of handling and querying the new index types from the API with the
-[`internal__index`](api_ii_param) param.
+[`internal__index`][api_ii_param] param.
 
 [api_ii_param]: https://github.com/WordPress/openverse/pull/2073
 
@@ -33,11 +33,11 @@ ease of handling and querying the new index types from the API with the
 ## Dependencies
 
 Same as for
-[Implementation Plan: Update Staging Database](./20230406-implementation_plan_update_staging_database.md).
+[Implementation Plan: Update Staging Database](/projects/proposals/search_relevancy_sandbox/20230406-implementation_plan_update_staging_database.md).
 
-This work is related to the
-[Staging database recreation DAG](staging_db_recreation) plan for having
-production volumes in the staging DB, which is expected to finish soon.
+This work is related to the [Staging database recreation
+DAG][staging_db_recreation] plan for having production volumes in the staging
+DB, which is expected to finish soon.
 
 [staging_db_recreation]: https://github.com/WordPress/openverse/issues/1989
 
@@ -59,7 +59,7 @@ production volumes in the staging DB, which is expected to finish soon.
 #### Outlined Steps
 
 This DAG will leverage the **Ingestion server's API** and use the existing
-[`REINDEX` task](REINDEX), which is the same used by the data refresh process to
+[`REINDEX` task][reindex], which is the same used by the data refresh process to
 create the index.
 
 1. Get the current timestamp in order to create the index suffix with the form
@@ -70,7 +70,7 @@ create the index.
 3. If `promoted=True` is passed, then inmediatly make the `media` alias point to
    the new index. A
 4. If the index is promoted then the DAG checks if `remove_old_if_promoted=True`
-   and proceeds to trigger the [DELETE_INDEX](delete_index) task in the
+   and proceeds to trigger the [DELETE_INDEX][delete_index] task in the
    Ingestion server. Otherwise the DAG ends at the previous step.
 
 [reindex]:
@@ -85,7 +85,7 @@ create the index.
 This DAG is intented to be used most likely with the index resulting from the
 previous DAG or from the data refresh process, that is, an index with the
 database fully indexed, as the `source_index` for the ES
-[Reindex](es_reindex_api) API.
+[Reindex][es_reindex_api] API.
 
 [es_reindex_api]:
   https://www.elastic.co/guide/en/elasticsearch/reference/7.12/docs-reindex.html
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -1,4 +1,4 @@
-# 2023-06-08 Implementation Plan: Staging Elasticsearch Reindexing DAGs
+# 2023-06-08 Implementation Plan: Staging Elasticsearch Reindex DAGs
 
 **Author**: @krysal
 
@@ -14,10 +14,11 @@
 
 ## Overview
 
-This document describes the addition of two DAGs for Elasticsearch index
+This document describes the addition of two DAGs for Elasticsearch (ES) index
 creation ––full and proportional-by-provider–– which will allow us to decouple
-the process from the extended Ingestion server's data refresh process and
-iterate faster experimenting with smaller indices.
+the process from the long Ingestion server's data refresh process and experiment
+with smaller indices. Also includes the adoption of two new index aliases for
+ease of handling the new index types.
 
 ## Expected Outcomes
 
@@ -31,10 +32,13 @@ iterate faster experimenting with smaller indices.
 Same as for
 [Implementation Plan: Update Staging Database](./20230406-implementation_plan_update_staging_database.md).
 
-This work does not depend on any other projects and can be implemented at any
-time.
+This work is related to the
+[Staging database recreation DAG](staging_db_recreation) plan for having
+production volumes in the staging DB, which is expected to finish soon.
 
-## Outlined Steps
+[staging_db_recreation]: https://github.com/WordPress/openverse/issues/1989
+
+## DAGs
 
 ### `recreate_full_<media>_index` DAG
 
@@ -49,9 +53,9 @@ time.
    index pointed by the `<media>` alias should be deleted after replacement.
    Defaults to `False`.
 
-#### Description
+#### Outlined Steps
 
-This DAG will leverage the Ingestion server's API and use the existing
+This DAG will leverage the **Ingestion server's API** and use the existing
 [`REINDEX` task](REINDEX), which is the same used by the data refresh process to
 create the index.
 
@@ -71,33 +75,93 @@ create the index.
 [delete_index]:
   https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/catalog/dags/data_refresh/data_refresh_task_factory.py#L222-L239
 
+<!--------------------------------------------------------------------------->
+
 ### `create_proportional_by_provider_<media>_index` DAG
 
+This DAG is intented to be used most likely with the index resulting from the
+previous DAG or from the data refresh process, that is, an index with the
+database fully indexed, as the `source_index` for the ES
+[Reindex](es_reindex_api) API.
+
+[es_reindex_api]:
+  https://www.elastic.co/guide/en/elasticsearch/reference/7.12/docs-reindex.html
+
 #### Parameters
+
+1. `media_type`: The media type for which the index is being created. Presently
+   this would only be `image` or `audio`.
+2. `source_index`: (Optional) The existing index on Elasticsearch to use as the
+   basis for the new index. If not provided, the index aliased to
+   `<media_type>-full` will be used.
+3. `percentage_of_production`: The proportion of items to take from each
+   provider from the total amount existing in production.
+
+#### Outlined Steps
+
+1. Get the list of media count by sources from the production Openverse API
+   [`https://api.openverse.engineering/v1/<media>/stats/`](https://api.openverse.engineering/v1/<media>/stats/)
+2. Calculate the total media adding up all the counts by provider
+3. Calculate the name of the new index with the following format:
+
+```python
+ f"{media_type}-{percentage_of_production}-percent-of-providers-{current_datetime}"
+```
+
+4. Make a dictionary mapping all the providers with the required amount of items
+   for the new index based on the provided `percentage_of_production` param.
+5. Iterate over the items of the resulting dictionary to index the subset of
+   each provider.
+
+```json
+POST _reindex?wait_for_completion=false
+{
+  "max_docs": num_items,
+  "source": {
+    "index": "image-full",
+    "query": {
+      "term": {
+        "source.keyword": "stocksnap"
+      }
+    }
+  },
+  "dest": {
+    "index": "image-50-percent-of-providers-20230608183000"
+  }
+}
+```
+
+6. Make the alias `<media>-subset-by-provider` point to the new index.
+7. Optionally. Query the stats of the resulting infex and print the results.
+
+```
+GET /image-reindexed-by-provider/_stats
+```
 
 ## Alternatives
 
 <!-- Describe any alternatives considered and why they were not chosen or recommended. -->
 
-## Parallelizable streams
+💭
 
-<!-- What, if any, work within this plan can be parallelized? -->
+## Parallelizable streams
 
 Both DAGs can be developed in parallel.
 
 ## Blockers
 
-<!-- What hard blockers exist which might prevent further work on this project? -->
-
 There is nothing currently blocking the implementation of this proposal.
 
+<!--
 ## Accessibility
 
-<!-- Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
+ Are there specific accessibility concerns relevant to this plan? Do you expect new UI elements that would need particular care to ensure they're implemented in an accessible way? Consider also low-spec device and slow internet accessibility, if relevant. -->
 
 ## Rollback
 
 <!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
+
+🤔
 
 ## Risks
 
@@ -111,3 +175,5 @@ where this impact the cluster performance.
 ## Prior art
 
 <!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
+
+- [Script which added 100 records per provider into the testing ECS database](https://github.com/WordPress/openverse-infrastructure/pull/314)
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -4,8 +4,8 @@
 
 ## Reviewers
 
-- [ ] @AetherUnbound
-- [ ] @sarayourfriend
+- [x] @AetherUnbound
+- [x] @sarayourfriend
 
 ## Project links
 
@@ -136,6 +136,10 @@ version for maintainers safety.
 2. `percentage_of_prod`: A float indicating the proportion of items to take from
    each provider from the total amount existing in production. E.g. `0.25` for a
    quarter of the production documents.
+
+> **Note**: Until defining an automated mechanism for creating the filtered
+> index on staging, it will need to be recreated often manually or default the
+> `source_index` to the `<media_type>-full` alias.
 
 #### Outlined Steps
 
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -27,7 +27,7 @@ handling and querying the new index types from the API with the
 
 ## Expected Outcomes
 
-- DAGs for fully recreation of the index that the API uses which includes all
+- DAGs for full recreation of the index that the API uses which includes all
   the database contents, by media type.
 - DAGs to create indexes of reduced size, proportional-to-production-by-provider
   index, by media type.
@@ -87,7 +87,7 @@ create the index.
       with a body including both actions, analogous to the [ingest server's
       task][ing_point_alias].
 
-4. If `point_alias=True` is passed, then inmediatly make the `<media_type>`
+4. If `point_alias=True` is passed, then immediately make the `<media_type>`
    alias point to the new index, detaching any other following the same
    procedure as indicated above. If `False` then the DAG ends at the previous
    step.
--- a/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
+++ b/documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
@@ -18,7 +18,10 @@ This document describes the addition of two DAGs for Elasticsearch (ES) index
 creation ––full and proportional-by-provider–– which will allow us to decouple
 the process from the long Ingestion server's data refresh process and experiment
 with smaller indices. Also includes the adoption of two new index aliases for
-ease of handling the new index types.
+ease of handling and querying the new index types from the API with the
+[`internal__index`](api_ii_param) param.
+
+[api_ii_param]: https://github.com/WordPress/openverse/pull/2073
 
 ## Expected Outcomes
 
@@ -67,8 +70,8 @@ create the index.
 3. If `promoted=True` is passed, then inmediatly make the `media` alias point to
    the new index. A
 4. If the index is promoted then the DAG checks if `remove_old_if_promoted=True`
-   and proceeds to [trigger the task in the ingestion server](DELETE_INDEX).
-   Otherwise the DAG ends at the previous step.
+   and proceeds to trigger the [DELETE_INDEX](delete_index) task in the
+   Ingestion server. Otherwise the DAG ends at the previous step.
 
 [reindex]:
   https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L282
@@ -113,8 +116,11 @@ database fully indexed, as the `source_index` for the ES
 5. Iterate over the items of the resulting dictionary to index the subset of
    each provider.
 
-```json
+```
 POST _reindex?wait_for_completion=false
+```
+
+```json
 {
   "max_docs": num_items,
   "source": {
@@ -132,7 +138,7 @@ POST _reindex?wait_for_completion=false
 ```
 
 6. Make the alias `<media>-subset-by-provider` point to the new index.
-7. Optionally. Query the stats of the resulting infex and print the results.
+7. Optionally. Query the stats of the resulting index and print the results.
 
 ```
 GET /image-reindexed-by-provider/_stats
@@ -140,9 +146,16 @@ GET /image-reindexed-by-provider/_stats
 
 ## Alternatives
 
-<!-- Describe any alternatives considered and why they were not chosen or recommended. -->
+### Combining both DAGs into one
 
-💭
+One alternative to creating two different indices by separate is to create the
+proportional by provider index using the Ingestion server. This would require
+modifying the REINDEX task of the ingestion server or creating a new one that
+takes only a subset of the providers by the indicated proportion.
+
+However, I discarded this option in favor of the one explained above because
+having both DAGs is much simpler and provides more possibilities for the
+creation of different indexes, which is the end goal of the project.
 
 ## Parallelizable streams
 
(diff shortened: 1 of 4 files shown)


documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+5. Iterate over the items of the resulting dictionary to index the subset of
+   each provider.
+
+```
+POST _reindex?wait_for_completion=false
+```
+
+```json
+{
+  "max_docs": num_items,
```
sarayourfriend, line 125: Nice! I didn't know about `max_docs`, that makes this a lot easier than I thought it would be.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+  "source": {
+    "index": "image-full",
+    "query": {
+      "term": {
+        "source.keyword": "stocksnap"
+      }
+    }
```
sarayourfriend, line 132: Have you tried this locally to see how the resulting query orders the documents? For these proportional indexes, I wonder if using [the random score function](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) would help us get a statistically significant sample :thinking:

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+2. Calculate the total media adding up all the counts by provider
+3. Calculate the name of the new index with the following format:
+
+```python
+ f"{media_type}-{percentage_of_production}-percent-of-providers-{current_datetime}"
+```
+
+4. Make a dictionary mapping all the providers with the required amount of items
+   for the new index based on the provided `percentage_of_production` param.
+5. Iterate over the items of the resulting dictionary to index the subset of
```
sarayourfriend, line 116: Do you know if these can run in parallel or serially? If they can run in parallel, would it make sense to use [dynamic task mapping](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dynamic-task-mapping.html) to create a new task per provider and then create individual sensors for each task ID? I've added a note below about needing the sensor.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+        "source.keyword": "stocksnap"
+      }
+    }
+  },
+  "dest": {
+    "index": "image-50-percent-of-providers-20230608183000"
+  }
+}
+```
+
```
sarayourfriend, line 139: Because we're passing `wait_for_completion=false`, ES will return a task ID for each provider request. Before we repoint aliases (or execute the next reindex request, if they must run serially), we should query for the task status and wait until it is finished. We'll also need a new step to refresh the index for it to be queryable after all the reindex jobs are finished.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+### Combining both DAGs into one
+
+One alternative to creating two different indices by separate is to create the
+proportional by provider index using the Ingestion server. This would require
+modifying the REINDEX task of the ingestion server or creating a new one that
+takes only a subset of the providers by the indicated proportion.
+
+However, I discarded this option in favor of the one explained above because
+having both DAGs is much simpler and provides more possibilities for the
+creation of different indexes, which is the end goal of the project.
```
sarayourfriend, line 158: :+1: 

Do these DAGs relate to the [index settings iteration DAG](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230518-implementation_plan_staging_index_rapid_iteration.html) at all? Do you think it's worth adding the ability for us to create a proportional index that also has modified settings? From my perspective these proportional indexes are primarily useful for testing different index settings, because it would allow us to "test drive" the modified settings without needing to wait for the entire reindex to happen. I almost wonder if the proportional index creation should be added to the settings iteration DAG as an option :thinking:

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+We can discard the DAGs if the results are not as expected.
+
+## Risks
+
+<!-- What risks are we taking with this solution? Are there risks that once taken can’t be undone?-->
+
+Elasticsearch does not impose any limit on the amount of indices one can create
+but naturally they come with a cost. We don't have policies for creating or
+deleting indices for the time being so we should monitor if we reach a point
+where having many indexes impact the cluster performance.
```
sarayourfriend, line 186: We could add a disk usage monitor for this.

sarayourfriend (commented):
LGTM so far, I left some clarifying questions, some suggestions, and some potential alternatives. But this looks great so far :slightly_smiling_face:

AetherUnbound:
I will be looking at this IP this week! Sorry for the delay due to my AFK.

stacimc:
@krysal -- I'm happy to review, but just clarifying whether I'm a required reviewer on this, as I'm not named in the PR?

krysal:
I'll address @sarayourfriend's comments today! Sorry for the delay, other urgent matters took precedence in my calendar these days.

@AetherUnbound No worries at all!

@stacimc Yours was an automated assignation (probably because Sara isn't in the @WordPress/openverse-catalog group) but you're welcome to comment as well if you want. I am sure your ideas will improve the proposed plan :)

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+[`REINDEX` task](REINDEX), which is the same used by the data refresh process to
+create the index.
+
+1. Get the current timestamp in order to create the index suffix with the form
+   of `full-<timestamp>`.
+2. Use the `ingestion_server.trigger_and_wait_for_task()` utility to send the
+   `REINDEX` call to ingestion server, passing the previously generated
+   `index_suffix` in the data payload.
+3. If `promoted=True` is passed, then inmediatly make the `media` alias point to
+   the new index. A
```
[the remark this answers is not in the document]
AetherUnbound: Agreed, I think we might want to use `POINT_ALIAS` terminology here instead.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+4. If the index is promoted then the DAG checks if `remove_old_if_promoted=True`
+   and proceeds to trigger the [DELETE_INDEX](delete_index) task in the
+   Ingestion server. Otherwise the DAG ends at the previous step.
+
+[reindex]:
+  https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L282
+[delete_index]:
+  https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/catalog/dags/data_refresh/data_refresh_task_factory.py#L222-L239
```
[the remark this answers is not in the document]
AetherUnbound: (note that it looks like this comment was cut off Sara)

That's an interesting suggestion! I think that the additional complexity around preventing deletion & managing concurrency might not make this approach worth it, even if it means relying on the ingestion server less.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+[es_reindex_api]:
+  https://www.elastic.co/guide/en/elasticsearch/reference/7.12/docs-reindex.html
+
+#### Parameters
+
+1. `media_type`: The media type for which the index is being created. Presently
+   this would only be `image` or `audio`.
+2. `source_index`: (Optional) The existing index on Elasticsearch to use as the
+   basis for the new index. If not provided, the index aliased to
+   `<media_type>-full` will be used.
```
AetherUnbound, line 99: This new alias, `<media_type>-full` is mentioned, but there isn't a note in the previous DAG definition about creating it. Do you mind noting that we'd add `<media_type>-full` as an alias explicitly to the steps of the above DAG?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
AetherUnbound: Definitely, this rocks!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
AetherUnbound: +1, we should make this step(s) explicit in the plan.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
AetherUnbound: That's a good idea! I think one approach might be using the already-created proportional index as a baseline for creating a new index with new settings. That _should_ be possible with the DAGs as-defined already I believe!

AetherUnbound (commented):
Thanks for drafting this Krystle, and apologies it took me so long to review! I'm in agreement with Sara on a few points of clarification, along with some additional explicit steps being added.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
sarayourfriend: >  I think one approach might be using the already-created proportional index as a baseline for creating a new index with new settings

But why have separate steps for them? It requires _two_ reindexes rather than one. If the DAGs were combined then a single reindex can run, pushing documents into the index with new settings. To get only one or the other behaviour, just leave out one of the arguments.

I don't see why we would need to split this into two distinct steps, is all, unless you're talking about creating a new index without any documents in it. But then the DAG would need to check if we need to create a new index and also be sure that we're not somehow accidentally reindexing into a pre-existing index _accidentally_ instead of intentionally.

Combining them reduces the number of DAGs and the complexity of all of this. The DAG would just be:

1. Pull the source index settings
2. Get modified settings (noop if there are no settings changes passed to the DAG config)
3. Create the new index with the new settings
4. Reindex, either fully or proportionally depending on the DAG config

This appears to be much simpler than two separate DAGs, especially if it can skip a full reindex that we don't always need _especially when testing new settings_. Like I said in the project proposal, I still don't see a very good use case for these proportional indexes _except for testing new index settings_. The features appear to be fundamentally related such that having separate DAGs for them makes the intention of them extremely unclear.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
sarayourfriend: For what it's worth we already have complex concurrency rules for the same reasons in DAGs that use the ingestion server, so it wouldn't necessarily increase overall complexity of managing these things. Staging ingestion server only has two workers, so it'll likely go a _lot_ slower than production which already takes a while with 5 workers.

Reindexing from a remote cluster would probably be faster because our staging and production Elasticsearch clusters are provisioned with the same capacity.

The cut off part of my comment, IIRC, was going to say that the approach to just use the data refresh process for staging is probably faster to implement and could be simplified if we repurposed the existing data refresh DAG to run against multiple environments.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+## Overview
+
+This document describes the addition of two DAGs for Elasticsearch (ES) index
+creation ––full and proportional-by-provider–– which will allow us to decouple
+the process from the long Ingestion server's data refresh process and experiment
+with smaller indices. Also includes the adoption of two new index aliases for
+ease of handling and querying the new index types from the API with the
+[`internal__index`](api_ii_param) param.
+
+[api_ii_param]: https://github.com/WordPress/openverse/pull/2073
```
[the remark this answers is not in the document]
krysal: It meant to be a [Reference-style link](https://www.markdownguide.org/basic-syntax#reference-style-links) but apparently I got it wrong 😅 Which is strange, I had always used it like this here in GH but maybe it's not an standard form 🤔

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+[staging_db_recreation]: https://github.com/WordPress/openverse/issues/1989
+
+## DAGs
+
+### `recreate_full_<media>_index` DAG
+
+#### Parameters
+
+1. `media_type`: The media type for which the index is being created. Presently
+   this would only be `image` or `audio`.
```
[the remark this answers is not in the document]
krysal: Right, I was indecisive myself as to which approach would be best. I was thinking having the same DAG repeated just for different media types could make the list unnecessarily long but now thinking again, the benefit of having the history of runs separated by media type sounds appealing (and potentially easier for management as well?).

I'll change to skip this param and go with DAG factorys unless there are reasons keep the one DAG approach!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
sarayourfriend: I didn't know about that feature! It turns out MyST might just be using a different syntax: https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#markdown-link-syntax

Rather than `[`internal__index`](api_ii_param)` it would be `[`internal__index`][api_ii_param]`.

TIL about this feature!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
krysal: Good point! I saw that option of reindex from a remote cluster but I think generally, I'm wary of this kind of data mixing between environments. You make it sound it simple but I cannot see how that would _not_ cause additional load to production. It still has to transfer _all_ the documents from one cluster to another. It may not impede its operation but we certainly do not know.

I also like that the `<media>-full` index product is the result of the same process used for production but just with staging data. If something goes wrong on the way there would be less uncertainty as to what could have gone wrong. I can add this remote cluster approach to the alternatives, maybe we could try as well to evaluate the performance. If it several times faster than using the Ingestion server without degrading the production cluster that would be a big plus to consider leaving it.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
krysal: Hm, I thouht it would be better differentiate it from the `point_alias` task of the Ingestion server because it involves deleting the old index, which is not the case here. However, the promote task (from Ingestion server) also involves that so makes sense to use the more obvious name for this param 👍 Noted.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: I try it locally but didn't pay attention to the order, I didn't think it was important. How would you define a "statistically significant sample" for this case?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: I was thinking more about a scenario where having many indices degrades the overall cluster response time. I read something around these lines in the Elastic forum. But that is a good suggestion as well. I believe something similar is included in the ECS alarms proposal, right?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: Right! I'll add that step explicitly.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
sarayourfriend: > You make it sound it simple but I cannot see how that would not cause additional load to production. It still has to transfer all the documents from one cluster to another. It may not impede its operation but we certainly do not know.

To add some additional nuance (but not necessarily arguing strongly in favour of the remote reindex approach, just adding more clarity). The only load I think it could possible cause to production is a full scan through the index documents, which is trivial compared to indexing and even querying, I would think. We can confirm this during a period when production isn't indexing and use something like `max_docs` set to 1 million just to see what happens in production. Reindex jobs are cancellable so it if resource consumption is too high in production for some reason, we can recover very quickly. That's to say: I suspect it will not cause significant load and we can pretty safely find out exactly what that would be. If the main reason to table this alternative (which has the benefits shared before) is because of unknown effects in production, we can confirm whether that's a good reason to table it for sure.

Additionally, now that the staging database is just a restored snapshot of production, there shouldn't be a difference between using the ingestion server worker approach vs reindexing from a remote cluster. The set of documents would be the same and we wouldn't be bottlenecked by the data volume/staging ingestion worker mis-match. That's to say, staging data is just production data. If the intention of this DAG is really to exercise the ingestion server worker approach, though, then that's a different story. That, however, would probably be better suited by retrofitting the existing data refresh DAG to be able to be manually run against staging (and therefore exercising the same data refresh routine that would be used in production) rather than a new DAG that operates slightly differently.

Anyway, I strongly agree with this, of course:

> I can add this remote cluster approach to the alternatives, maybe we could try as well to evaluate the performance. If it several times faster than using the Ingestion server without degrading the production cluster that would be a big plus to consider leaving it.

I'd be very happy to buddy with you on testing this if you'd like.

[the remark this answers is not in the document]
krysal: > We can confirm this during a period when production isn't indexing and use something like max_docs set to 1 million just to see what happens in production. Reindex jobs are cancellable so it if resource consumption is too high in production for some reason, we can recover very quickly.

That will be quite an appropriate experiment! Let's give it a shot 🚀

> If the intention of this DAG is really to exercise the ingestion server worker approach, though, then that's a different story. That, however, would probably be better suited by retrofitting the existing data refresh DAG to be able to be manually run against staging (and therefore exercising the same data refresh routine that would be used in production) rather than a new DAG that operates slightly differently.

The thing is that we don't want to run the whole data refresh process, just the part of ES reindexing. That is what the DAG defined here aims for.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
sarayourfriend: EC2 instances need the same treatment that the catalogue got to add the CloudWatch agent to get disk space reporting: https://github.com/WordPress/openverse-infrastructure/pull/455

The bigger question, I suppose, is how we would measure/notice degraded staging performance, or if we would need to at all. The worst case is: we notice performance degrades through regular usage and then delete some indexes. Do we need any special monitoring at all, in that case?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
sarayourfriend: > That will be quite an appropriate experiment! Let's give it a shot 🚀

Pinged in Slack to coordinate :slightly_smiling_face:

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,192 @@
...
+
+This work is related to the
+[Staging database recreation DAG](staging_db_recreation) plan for having
+production volumes in the staging DB, which is expected to finish soon.
+
+[staging_db_recreation]: https://github.com/WordPress/openverse/issues/1989
+
+## DAGs
+
+### `recreate_full_<media>_index` DAG
```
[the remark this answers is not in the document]
krysal: Done!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
sarayourfriend: > statistically significant sample

This isn't the term I should have used. More accurately would be "more diverse sample". If an individual provider's document's scores are consistent, then we'd be testing the _same_ 10% (or whatever proportion we use) each time the reindex happens. Using a random score would make it more likely to have a diverse set of document configurations (file types, sizes, etc). It may not be an issue if the default scores are already sufficiently random and distribute attribute variation proportionally. Random score would theoretically make it more likely for the subset of documents to be representative of the whole if the default scores are uniform (which I don't know).

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: I also thought about combining both DAGs into only one. The result of having two comes from that, at first, it seemed like a lot of complexity due to all the moving parts, but now that we know more about ES's APIs and the DAGs are better defined, I would say there is a possibility to combine them.

I still decided to write the proposal as two extra DAGs in case the separation was preferred for flexibility/simplicity and because the `create_new_es_index_{environment}` DAG is still not implemented, so, in theory, this shouldn't depend on that to be implemented. We can also merge them after both are implemented and tested separately.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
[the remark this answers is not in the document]
krysal: **Update**: We tried reindexing from staging to a local ES instance but it wasn't possible to access the staging ES cluster. It required access via HTTPS. The error received was the following:
```json
{
	...
	"error": {
        "type": "illegal_argument_exception",
        "reason": "[localhost:9200] not whitelisted in reindex.remote.whitelist"
    }
}
```

This would require modifications to the cluster settings so we will not pursue this option this time.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: It's possible to [run them in parallel](https://www.elastic.co/guide/en/elasticsearch/reference/7.12/docs-reindex.html#docs-reindex-from-multiple-sources)! Great suggestion ✨ I'll add the note to use dynamic task mapping.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
AetherUnbound: Your perspective makes sense Sara. I'm fine with separate for now with the idea that we can combine them down the line to simplify once things are implemented and tested!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
AetherUnbound: I would assume not, given that "regular usage" for staging is primarily our own interactions with it rather than direct user interactions.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+create a DAG per media type supported, which currently consists `image` and
+`audio`. Also includes the adoption of two new index aliases for ease of
+handling and querying the new index types from the API with the
+[`internal__index`][api_ii_param] param.
+
+[api_ii_param]: https://github.com/WordPress/openverse/pull/2073
+
+## Expected Outcomes
+
+- DAGs for fully recreation of the index that the API uses which includes all
```
AetherUnbound, line 30:
```suggestion
- DAGs for full recreation of the index that the API uses which includes all
```

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+      [indices.exists_alias][es_py_exists_alias] function.
+
+   2. If the alias doesn't exist, then it can be created and assigned in one
+      step using the [`indices.put_alias`][es_py_put_alias] function.
+
+   3. If the alias exists, send the request to add the new index and remove the
+      old one(s). Get the current index pointed by `<media_type>-full` alias.
+      The [indices.resolves_index][es_py_resolves_index] function can provide
+      this information. An alias can be related to multiple indexes but it will
+      most likely be only one in this case. Then use `indices.update_aliases()`
```
AetherUnbound, line 86: Do you mind adding a documentation link to this function call as well?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+      with a body including both actions, analogous to the [ingest server's
+      task][ing_point_alias].
+
+4. If `point_alias=True` is passed, then inmediatly make the `<media_type>`
```
AetherUnbound, line 90:
```suggestion
4. If `point_alias=True` is passed, then immediately make the `<media_type>`
```

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+5. If the index is aliased then the DAG checks if `delete_old_if_aliased=True`
+   and proceeds to run [`indices.delete`][es_py_delete] with the old index name.
+   Otherwise, the DAG ends at the previous step.
+
+[reindex]:
+  https://github.com/WordPress/openverse/blob/7427bbd4a8178d05a27e6fef07d70905ec7ef16b/ingestion_server/ingestion_server/indexer.py#L282
+[resolve]:
+  https://www.elastic.co/guide/en/elasticsearch/reference/7.12/indices-resolve-index-api.html
+[es_python_hook]:
+  https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/_api/airflow/providers/elasticsearch/hooks/elasticsearch/index.html#airflow.providers.elasticsearch.hooks.elasticsearch.ElasticsearchPythonHook
```
AetherUnbound, line 103: This notation is great! I'll definitely be using it in the future 😄

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+3. Once the index is created, make the alias `<media_type>-full` point to it.
+
+   1. Check if the alias exists. Use the
+      [`ElasticsearchPythonHook`][es_python_hook] with the
+      [indices.exists_alias][es_py_exists_alias] function.
+
+   2. If the alias doesn't exist, then it can be created and assigned in one
+      step using the [`indices.put_alias`][es_py_put_alias] function.
+
+   3. If the alias exists, send the request to add the new index and remove the
```
AetherUnbound, line 82: In the case we have a previous index pointing to `<media_type>-full` that isn't aliased to `<media_type>` as well, should we also delete the old index once the new index is pointed to `<media_type>-full`? (Since the old index will essentially be dangling)?

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -0,0 +1,231 @@
...
+3. Build the `dest_index` name with the following format:
+
+```python
+ f"{media_type}-{percentage_of_prod}-percent-of-providers-{current_datetime}"
+```
+
+4. Make a a list of dictionaries mapping all the providers with the required
+   amount of items for the new index based on the provided `percentage_of_prod`
+   param.
+5. Using the [Dynamic Task Mapping][airflow_dtm] feature of Airflow, expand the
```
AetherUnbound, line 150: This `airflow_dtm` doesn't appear to have a reference.

AetherUnbound (commented):
This is great! I have just a few minor notes with one outstanding clarification around deleting indices.

sarayourfriend:
@krysal I will re-review this during my evening today (before your day tomorrow).

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: I was thinking of keeping all the indexes (unless specified otherwise) for benchmarking and testing purposes and deleting them manually. They could be switched manually in the API as well, but maybe I'm wrong in this assumption and we want to avoid dangling indexes as much as possible. What do you think? @AetherUnbound @sarayourfriend

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: Oh wow, it's weird that the check didn't catch that 😮 Thanks!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: Done.

sarayourfriend (changes requested):
No huge blockers on what is already covered. I shared my opinion about the aliases and automatically deleting previous indexes but have no opposition to whatever we land on there. My concerns are super small and whatever we do now could be changed with little effort in the future if we decided that the process was not working. In other words, no lasting harm to the project or its contributors.

I am requesting changes only to ask for clarification about how filtered indexes would apply here. If index settings are changing, for example, would the associated filtered index need those same settings applied? Or/additionally: should the filtered index be the default source index, to proactively exclude content we've determined is sensitive?

I also would like to advocate once more for combining the index settings DAG with these ones, not as part of this IP, but to have that explicitly documented either as an end note on this document and/or in the project thread. Pinging @AetherUnbound as the project lead on this one in particular. As I said, the proportional-by-provider indexes almost only seem useful if you can change the index settings, and considering it's already using almost all the same steps as the index settings DAG, just with a difference on how it runs the reindex (whole index or by provider), these feel fundamentally related. It seems weird not to explicitly mention this _somewhere_ in the project documentation. Apologies if I have missed that or if it got lost in the conversation here, I've had a hard time keeping track of this :sweat_smile: 

Anyway, the only blocker here is as I said, to request explicit clarification about the filtered indexes, just to make sure that either they are irrelevant to this work or if the source index should default to the filtered indexes for maintainer quality of life/safety.

krysal:
@sarayourfriend 
> I am requesting changes only to ask for clarification about how filtered indexes would apply here. If index settings are changing, for example, would the associated filtered index need those same settings applied?

That's a good question! It looks more pertinent to have been discussed in the previous plan, but since that one is already merged let's do it here 😄 I'd say it depends on the new settings but probably yes, it'd need the same settings applied. Since these DAGs are planned for experimentation initially, how to apply a new configuration permanently will require a separate conversation, evaluating the results obtained.

> Or/additionally: should the filtered index be the default source index, to proactively exclude content we've determined is sensitive?

I don't have a strong opinion here. If people want to default to filtered indexes for safety reasons, I'll change to that. I chose the usual media indices because they are our base and have the most data. The DAG runner can also choose a filtered index as `source_index` for the -proportional-by-provider DAG in the plan's current form.

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
sarayourfriend: > Their use is totally optional. Should I highlight this more clearly in the proposal?

:+1: 

> [T]he proposal allows for multiple -full or -subset-by-provider indexes ... unless we keep telling the DAGs to delete the old ones

Just a heads up (I'm not entirely sure if this is the situation you're talking about), but if there are multiple indexes under the same alias, then ES will query _all_ those indexes with the alias, not just one. If `logs-2023` and `logs-2022` both have a `logs` alias applied to them, when querying with `logs` as the index name, both indexes will be queried and results from both will be included (not sure how scores are compared though). Not sure if that's the situation you were talking about by multiple indexes with the suffix aliases, or if you were talking about the aliases at all :sweat_smile:

sarayourfriend (approved):
LGTM.

I think defaulting to the filtered index is the right thing to do for maintainer safety. "Default to the safe option" is a good guiding principle. It's easy enough to change the default one way or another regardless, so approving.

Looking forward to seeing how these DAGs get used!

AetherUnbound:
Ah, that's a really important point you've raised @sarayourfriend. I don't think any of our current plans cover the _filtered_ indices in staging actually 😮 That might be because both this project and the filtering of results on the API were being proposed alongside each other, so the latter hadn't been finalized while we were planning out the former. As it stands, I don't believe we have an explicit mechanism defined for creating the filtered index on staging. Given that this is not production though, perhaps it would make sense to **filter the results while building the `<media_type>-full` index from the database**? That would mean all downstream results based on that index are always filtered, and we don't have to worry about managing both the unfiltered & filtered indices on staging. What do you think @krysal?

We've had a lot of discussion about this IP, I wonder if it might make sense to merge it _as-is_ given it's quite close, with a fast-follow clarification about how we'd manage the filtering (given we won't be actually working on the implementation of this for a little bit since the project is on hold). Would you prefer that Krystle?

And as for the automatic deletion, those are great points. I don't think we need to incorporate that aspect of it in the proposal here - we can leave it as an MSR task or something similar!

krysal:
> Given that this is not production though, perhaps it would make sense to filter the results while building the <media_type>-full index from the database? That would mean all downstream results based on that index are always filtered, and we don't have to worry about managing both the unfiltered & filtered indices on staging. What do you think @krysal?

@AetherUnbound The current way of building the filtered index is using the ES's Reindex API, it's not exactly a direct process form from the database. That would require creating a new process but I think we can just trigger the DAG for creating the filtered index as an additional step for building the `<media_type>-full` index. I'd prefer to have both indexes, the `<media_type>-full` and `<media_type>-filtered`, to provide more flexibility and work with the same indexes we have for production. What I am not sure about is if the DAG to create the filtered index is enabled for both environments or if it only applies to production. In the latter case, it would require some modifications. But it would be a really nice addition and actually closer to the part of the data refresh process that we're extracting here 😄

> We've had a lot of discussion about this IP, I wonder if it might make sense to merge it as-is given it's quite close, with a fast-follow clarification about how we'd manage the filtering (given we won't be actually working on the implementation of this for a little bit since the project is on hold). Would you prefer that Krystle?

Agree on merging this, as the filtered index is a requirement that was not initially included. I'll change the default source to the filtered index for the subset-by-provider DAG, as it's an easy change. Thank you both for the excellent suggestions.

AetherUnbound:
Sounds good! Yes unfortunately the filtered index DAG is only enabled for production, so we'd need to either have a similar DAG factory for it or set the environment as a DAG parameter (though I think our team preference is for the former to prevent accidental production operations).

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
```diff
@@ -131,7 +131,7 @@ Reindex][es_reindex_api] API.
 
 1. `source_index`: (Optional) The existing index on Elasticsearch to use as the
    basis for the new index. If not provided, the index aliased to
-   `<media_type>-full` will be used.
+   `<media_type>-filtered` will be used.
```
AetherUnbound, line 134: Do you mind adding one note somewhere in the document that we'll need to ensure `<media_type>-filtered` can be created regularly as well, so we don't forget it when we go to tackle the implementation steps?

AetherUnbound (approved):
Amazing, thanks for all our hard work on this!

documentation/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.md
krysal: Done in https://github.com/WordPress/openverse/pull/2358/commits/f2224a5cbc2eb6d54ef419b62514fa7ed54ac8a1 .

krysal:
Thank you all!

PR #2380, opened 2023-06-11
Unified logging approach api and ingestion server

ashiramin:
## Fixes

Fixes #[2031] by @dhruvkb

## Description

Unified logging approach between api and ingestion server by logging container logs in ci/cd instead of inside python code. Now both `test-api` and `test-ing` follow a similar pattern of storing test logs. 

1. **`.github/workflows/ci_cd.yml`:**
   - Added a step to store the ingestion-server test logs an

2. **`ingestion_server/justfile`:**
   - Added a new command to display integration test logs.

3. **`ingestion_server/test/integration_test.py`:**
   - Removed code that stopped running containers and deleted data in volumes. This now happens in CI/CD pipleline
   - Also removed step to store test logs in a txt file

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -326,7 +326,7 @@ jobs:
 
       - name: Print ingestion-server test logs
         run: |
-          just ingestion_server/logs > ingestion_server/test/ingestion_logs.txt
+          just ingestion_server/test-logs > ingestion_server/test/ingestion_logs.txt
           cat ingestion_server/test/ingestion_logs.txt
 
       - name: Upload ingestion test logs
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -324,6 +324,11 @@ jobs:
       - name: Run ingestion-server tests
         run: just ingestion_server/test-local
 
+      - name: Print ingestion-server test logs
+        if: success() || failure()
+        run: |
+          just ingestion_server/logs > ingestion_server/test/ingestion_logs.txt
+
       - name: Upload ingestion test logs
         if: success() || failure()
         uses: actions/upload-artifact@v3
@@ -334,6 +339,9 @@ jobs:
       - name: Print ingestion test logs
         if: success() || failure()
         run: cat ingestion_server/test/ingestion_logs.txt
+
+      - name: Cleanup ingestion-server test
+        run: just ingestion_server/test-clean-dc
 
   #######
   # API #
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -324,7 +324,7 @@ jobs:
       - name: Run ingestion-server tests
         run: just ingestion_server/test-local
 
-      - name: Print ingestion-server test logs
+      - name: Output ingestion-server test logs
         run: |
           just ingestion_server/logs > ingestion_server/test/ingestion_logs.txt
 
@@ -334,10 +334,6 @@ jobs:
         with:
           name: ing_logs
           path: ingestion_server/test/ingestion_logs.txt
-
-      - name: Print ingestion test logs
-        if: success() || failure()
-        run: cat ingestion_server/test/ingestion_logs.txt
 
       - name: Cleanup ingestion-server test
         run: just ingestion_server/test-clean-dc
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -325,7 +325,6 @@ jobs:
         run: just ingestion_server/test-local
 
       - name: Print ingestion-server test logs
-        if: success() || failure()
         run: |
           just ingestion_server/logs > ingestion_server/test/ingestion_logs.txt
 
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -324,9 +324,10 @@ jobs:
       - name: Run ingestion-server tests
         run: just ingestion_server/test-local
 
-      - name: Output ingestion-server test logs
+      - name: Print ingestion-server test logs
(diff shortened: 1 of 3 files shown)


sarayourfriend:
@ashiramin Can you add testing instructions for this change? It makes it much easier for maintainers to review the changes leading to reviews coming in sooner :slightly_smiling_face:

ashiramin:
> @ashiramin Can you add testing instructions for this change? It makes it much easier for maintainers to review the changes leading to reviews coming in sooner 🙂

@sarayourfriend Thank you for letting me know. I updated the testing instruction section in the PR description :)

.github/workflows/ci_cd.yml
```diff
@@ -324,16 +324,20 @@ jobs:
...
         uses: actions/upload-artifact@v3
         with:
           name: ing_logs
           path: ingestion_server/test/ingestion_logs.txt
 
-      - name: Print ingestion test logs
-        if: success() || failure()
-        run: cat ingestion_server/test/ingestion_logs.txt
+      - name: Cleanup ingestion-server test
+        run: just ingestion_server/test-clean-dc
```
sarayourfriend, line 340: I'm not sure, does this need an `if: success() | failure()` as well so that it always runs? @dhruvkb would know best.

ingestion_server/test/integration_test.py
```diff
@@ -431,18 +432,9 @@ def tearDownClass(cls) -> None:
...
-            cls._compose_cmd(
-                ["logs", "--no-color"], stderr=subprocess.STDOUT, stdout=file
-            )
-
-        cls._compose_cmd(
-            ["down", "-v"],
-            capture_output=True,
-        )
+        # Tear down the Docker Compose stack locally
+        if "GITHUB_ACTIONS" not in os.environ:
```
sarayourfriend, line 436: Just a nit: we usually check for the `CI` variable in these cases. This isn't a blocker, but using a consistent variable makes it more intuitive for us to test the differences between a CI run and local run by setting `CI=true` before running the command (vs needing to dig up what specific variable(s) the particular thing we're running checks).

```suggestion
        if "CI" not in os.environ:
```

Here are examples of our usage of the CI variable:

https://github.com//WordPress/openverse/blob/bdb8f0ad1dd0b2ae504f708b4105fdadc10e8988/justfile#L10

https://github.com//WordPress/openverse/blob/fae1dc0ef5f49c6f424db5c018e328e8d8e91f15/frontend/nuxt.config.ts#L71

https://github.com//WordPress/openverse/blob/9471728b4686d054493cd4528554e957dfee4350/frontend/test/playwright/playwright.config.ts#L29

I think CI is a bit more standardised, if we switched CI runners away from GitHub actions (what a mess that would be :sweat_smile:) those checks would continue working without modification.

sarayourfriend (commented):
Looks good to me! I'll be happy to approve once @dhruvkb confirms whether the if condition is needed on "Cleanup ingestion-server test".

ingestion_server/test/integration_test.py
ashiramin: Thank you for this. Made the update :)

.github/workflows/ci_cd.yml
ashiramin: Yes I think so. I'm not entirely sure if we need the cleanup step at the end. I'm guessing since its the end of step, github actions should delete the docker containers by itself. I added it in because the CI was doing that previously. the only difference was that it was happening as part of the teardown step in pytest

sarayourfriend: I think it's not strictly necessary to do any clean up in github actions for the reason you said, the container running it will get thrown out anyway. Maybe we can remove this step altogether?

ashiramin: Removed that step

sarayourfriend (changes requested):
Apologies for not noticing this before, but I'd like to request that this change does not affect the local development experience by making the logs inaccessible locally. Before this change the logs would output to a file before the stack is torn down. I believe the easiest way to preserve this is to remove tearing down the stack from the Python script entirely. Instead, we can add `test-logs test-clean-dc` to `test-local`.

It would look something like this:

```
test-local:
    pipenv run pytest {{ args }}
    if [ -z "$CI" ]; then just test-logs && just test-clean-dc; fi
```

Please also update the `test-clean-dc` doc-comment as it'll now be regularly used.

ashiramin:
> Apologies for not noticing this before, but I'd like to request that this change does not affect the local development experience by making the logs inaccessible locally. Before this change the logs would output to a file before the stack is torn down. I believe the easiest way to preserve this is to remove tearing down the stack from the Python script entirely. Instead, we can add `test-logs test-clean-dc` to `test-local`.
> 
> It would look something like this:
> 
> ```
> test-local:
>     pipenv run pytest {{ args }}
>     if [ -z "$CI" ]; then just test-logs && just test-clean-dc; fi
> ```
> 
> Please also update the `test-clean-dc` doc-comment as it'll now be regularly used.

@sarayourfriend made the change :) sorry it took me a bit longer. Please

sarayourfriend (approved):
Fantastic, thanks very much for your patience in the review process. I've tested this locally by running `just ingestion_server/test-local` with and without `CI=1` and the behaviour is exactly as expected.

Thanks again very much for your contribution @ashiramin :pray:

stacimc (approved):
Thanks for your contribution and especially for your patience @ashiramin :)

Pinging @dhruvkb again for awareness, but I was able to test this locally and this looks good to me.

PR #2386, opened 2023-06-12
Removed connection ID variables in favour of simply defining

### LINKED ISSUE #2294: Remove connection ID environment variables in favor of simply defining them
## Description

We currently have a pattern of defining the connection ID for an Airflow Connection via an environment variable (with a default) in addition to the environment variable for the connection itself. This creates two new configuration values for one connection ID. Below are all of the current examples:

https://github.com/WordPress/openverse/blob/a5fb3ea5f5671085f0239641ddc1c637a6193763/catalog/dags/common/constants.py#L32-L36

https://github.com/WordPress/openverse/blob/a5fb3ea5f5671085f0239641ddc1c637a6193763/catalog/tests/dags/common/popularity/test_sql.py#L17

https://github.com/WordPress/openverse/blob/a5fb3ea5f5671085f0239641ddc1c637a6193763/catalog/tests/dags/common/loader/test_sql.py#L26

https://github.com/WordPress/openverse/blob/a5fb3ea5f5671085f0239641ddc1c637a6193763/catalog/tests/dags/common/test_sql.py#L16

These values rarely change (if ever), and do not need to be configured via environment variable.

Instead, we should define these values outright as constants and remove the environment variables from the catalog's `env.template` file.

duwarakan:
## Fixes
removed connection ID variables in favour of simply defining

Fixes #2294

## Description
Removed environment variables and defined them as constants: 
1. POSTGRES_CONN_ID, OPENLEDGER_API_CONN_ID and  AWS_CONN_ID in contants.py
2. POSTGRES_CONN_ID in loader\test_sql.py, popularity\test_sql.py
3. POSTGRES_TEST_CONN_ID in common\test_sql.py

Removed OPENLEDGER_CONN_ID OPENLEDGER_API_CONN_ID AWS_CONN_ID in catlog\env.template
POSTGRES_API_STAGING_CONN_ID and AWS_RDS_CONN_ID in constants.py were not changed.


AetherUnbound (commented):
Thanks for this change @duwarakan! Would you also mind making the same change to the `POSTGRES_API_STAGING_CONN_ID` value? We can leave `AWS_RDS_CONN_ID` as-is since we _do_ want that to be overridden locally using environment variables.

AetherUnbound (dismissed):
Looks excellent, thanks!!

sarayourfriend (changes requested):
@duwarakan Madison noted that `AWS_CONN_ID` should be left as is:

> We can leave AWS_RDS_CONN_ID as-is since we do want that to be overridden locally using environment variables.

@AetherUnbound Can you confirm this is still needed? Just want to clarify as there doesn't appear to be conversation between your request for the change, it not being applied, and your approval on the PR.

catalog/dags/common/constants.py
```diff
@@ -29,10 +29,8 @@
 }
 XCOM_PULL_TEMPLATE = "{{{{ ti.xcom_pull(task_ids='{}', key='{}') }}}}"
 
-POSTGRES_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
-OPENLEDGER_API_CONN_ID = os.getenv("OPENLEDGER_API_CONN_ID", "postgres_openledger_api")
-POSTGRES_API_STAGING_CONN_ID = os.getenv(
-    "POSTGRES_API_STAGING_CONN_ID", "postgres_openledger_api_staging"
-)
-AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
+POSTGRES_CONN_ID = "postgres_openledger_testing"
```
stacimc, line 32: Should this not actually be `postgres_openledger_upstream`?

catalog/dags/common/constants.py
```diff
@@ -29,10 +29,8 @@
...
+OPENLEDGER_API_CONN_ID = "postgres_openledger_api"
+POSTGRES_API_STAGING_CONN_ID = "postgres_openledger_api_staging"
+AWS_CONN_ID = "aws_conn_id"
```
stacimc, line 35: Shouldn't this be `aws_default`?

stacimc (changes requested):
> Madison noted that AWS_CONN_ID should be left as is

I believe that's `AWS_RDS_CONN_ID`, not `AWS_CONN_ID`, so I think that's fine.

I do have a question about `POSTGRES_CONN_ID` and `AWS_CONN_ID`, however. It looks like those values changed from the previous implementation.

sarayourfriend:
@duwarakan I'm going to draft this PR to prevent getting new notifications while changes aren't made. Please let us know if you're still interested in working on this PR and applying the changes identified by Staci.

AetherUnbound:
I've rebased and updated the PR, and it should be ready for review once more.

AetherUnbound (approved):

catalog/dags/common/constants.py
```diff
@@ -29,11 +29,9 @@
 }
 XCOM_PULL_TEMPLATE = "{{{{ ti.xcom_pull(task_ids='{}', key='{}') }}}}"
 
-POSTGRES_CONN_ID = os.getenv("OPENLEDGER_CONN_ID", "postgres_openledger_testing")
-OPENLEDGER_API_CONN_ID = os.getenv("OPENLEDGER_API_CONN_ID", "postgres_openledger_api")
-POSTGRES_API_STAGING_CONN_ID = os.getenv(
-    "POSTGRES_API_STAGING_CONN_ID", "postgres_openledger_api_staging"
-)
-AWS_CONN_ID = os.getenv("AWS_CONN_ID", "aws_conn_id")
+POSTGRES_CONN_ID = "postgres_openledger_upstream"
```
obulat, line 32: Should we add a follow-up issue to standardize and document these constants?

obulat (approved):
Thank you for your contribution, @duwarakan! 🎉

catalog/dags/common/constants.py
AetherUnbound: An issue for documenting them would be good! Standardizing them would be a bit harder for _existing_ ones because it would require some extra work in production so we don't break things when the code is deployed. I can make an issue for it and make note of that there 🙂

PR #2395, opened 2023-06-13
Add Slack functionality inside logger

### LINKED ISSUE #698: Add Slack functionality inside logger
## Problem
In the `ingestion_server` codebase, calls to `log` functions like `log.info` and `log.error` are generally followed by sending the same information to Slack using the wrapper. This can be solved in a more elegant way by making Slack itself a logging destination within the Python logging framework.

## Description

Integrating Slack with the logging configuration means that
- we won't have to invoke Slack everywhere manually
- Slack can make use of Pythonic log levels and other utilities like formatters

## Alternatives

We can either roll with a homemade solution or use something like https://github.com/mathiasose/slacker_log_handler.

## Implementation

Kerfred:
## Fixes

Fixes #698 by @dhruvkb

## Description
The slack logger is now handling the standard logging. Nomore need to add log.info(), log.error(), log.debug() when we are logging to slack.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/ingestion_server/slack.py
```diff
@@ -61,14 +61,17 @@ def _message(text: str, summary: str = None, level: Level = Level.INFO) -> None:
...
 
 
 def info(text: str, summary: str = None) -> None:
+    log.info(text)
     _message(text, summary, level=Level.INFO)
 
 
 def error(text: str, summary: str = None) -> None:
+    log.error(text)
     _message(text, summary, level=Level.ERROR)
```
AetherUnbound, line 77: I can't add a change to the line directly but I think we should add a similar `log.info` line to the `status` method just below this!

ingestion_server/ingestion_server/slack.py
```diff
@@ -61,14 +61,17 @@ def _message(text: str, summary: str = None, level: Level = Level.INFO) -> None:
 
 
 def verbose(text: str, summary: str = None) -> None:
+    log.debug(text)
```
AetherUnbound, line 64: This (and all other `log` calls in made by this PR) should have [`stacklevel=2`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) added to them:

```suggestion
    log.debug(text, stacklevel=2)
```

This will ensure that when a log line is emitted from this module, it appears in the logs as if it came from the place where `slack.verbose` (or the other logging functions) were called. So rather than something like

```
DEBUG slack.py:63 - <the real message>
```

it'd say something like

```
DEBUG - indexer.py:102 - <the real message>
```

AetherUnbound (commented):
Thanks so much for your contribution @Kerfred! This PR just needs a few small changes then it should be ready 🙂

ingestion_server/ingestion_server/slack.py
Kerfred: OK

ingestion_server/ingestion_server/slack.py
Kerfred: Doing that would duplicate the log because the status() method calls the info() method that already does a log.info().
Do you agree?

AetherUnbound: OH, right! We should be set then 😄

Kerfred:
I have added stacklevel=2, commited and pushed again on the branch.

AetherUnbound:
As I was testing this, I realized that _because_ we're calling `slack.info` within `slack.status`, the stack level for `stack.status` calls will always show the log line within it. I think it might make the  most sense to add `stacklevel` as a keyword argument to the other 3 `slack` module logging functions, with a default of 2. Then within `slack.status`, we can call `info(text, None, stacklevel=3)` so the logs show the line that `slack.status` was called from instead. Something like this:

```diff
diff --git a/ingestion_server/ingestion_server/slack.py b/ingestion_server/ingestion_server/slack.py
index 717f5dbda..f5246f353 100644
--- a/ingestion_server/ingestion_server/slack.py
+++ b/ingestion_server/ingestion_server/slack.py
@@ -65,8 +65,8 @@ def verbose(text: str, summary: str = None) -> None:
     _message(text, summary, level=Level.VERBOSE)
 
 
-def info(text: str, summary: str = None) -> None:
-    log.info(text, stacklevel=2)
+def info(text: str, summary: str = None, stacklevel: int = 2) -> None:
+    log.info(text, stacklevel=stacklevel)
     _message(text, summary, level=Level.INFO)
 
 
@@ -82,4 +82,4 @@ def status(model: str, text: str) -> None:
     Model is required an all messages get prepended with the model.
     """
     text = f"`{model}`: {text}"
-    info(text, None)
+    info(text, None, stacklevel=3)
```

What do you think?

Kerfred:
> As I was testing this, I realized that _because_ we're calling `slack.info` within `slack.status`, the stack level for `stack.status` calls will always show the log line within it. I think it might make the most sense to add `stacklevel` as a keyword argument to the other 3 `slack` module logging functions, with a default of 2. Then within `slack.status`, we can call `info(text, None, stacklevel=3)` so the logs show the line that `slack.status` was called from instead. Something like this:
> 
> ```diff
> diff --git a/ingestion_server/ingestion_server/slack.py b/ingestion_server/ingestion_server/slack.py
> index 717f5dbda..f5246f353 100644
> --- a/ingestion_server/ingestion_server/slack.py
> +++ b/ingestion_server/ingestion_server/slack.py
> @@ -65,8 +65,8 @@ def verbose(text: str, summary: str = None) -> None:
>      _message(text, summary, level=Level.VERBOSE)
>  
>  
> -def info(text: str, summary: str = None) -> None:
> -    log.info(text, stacklevel=2)
> +def info(text: str, summary: str = None, stacklevel: int = 2) -> None:
> +    log.info(text, stacklevel=stacklevel)
>      _message(text, summary, level=Level.INFO)
>  
>  
> @@ -82,4 +82,4 @@ def status(model: str, text: str) -> None:
>      Model is required an all messages get prepended with the model.
>      """
>      text = f"`{model}`: {text}"
> -    info(text, None)
> +    info(text, None, stacklevel=3)
> ```
> 
> What do you think?

You are totally right. I have done the change.

stacimc (approved):
Thanks for this @Kerfred! This looks great.

AetherUnbound (approved):
Tested this all locally and it worked perfectly! This is such an improvement, thanks a ton @Kerfred 😄

PR #2418, opened 2023-06-16
Prevent unnecessary post-process results recursion

### LINKED ISSUE #2415: `_post_process_results` does not check if any hits exist before making additional pagination requests
## Description

`_post_process_results` attempts to "fill the requested page". However, it never checks if the page can be filled based on the total number of hits available on the query reported by Elasticsearch.

## Reproduction

1.  Run the server locally on `main` and observe unnecessary recursive calls to `_post_process_results` when making this request: `http://0.0.0.0:50280/v1/images/?q=bird%20perched&page_size=20&page=3&filter_dead=True`

That particular request works because it has a half-filled page on the last page of live links. To get the recursive calls, `filter_dead` must be enabled and you must have a results set that does not fill the page size.

## Fix

In addition to checking that `start + end > ELASTICSEARCH_MAX_RESULT_WINDOW`, also check `start + end > search_results.hits.total.value`.

This diff fixed it for me locally, but I think there are some edge cases we need to consider for when `query_size == hits` :thinking: 

```diff
diff --git a/api/api/controllers/search_controller.py b/api/api/controllers/search_controller.py
index def66bb34..e18bf388e 100644
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -206,7 +206,12 @@ def _post_process_results(
             ```
             """
             end += int(end / 2)
-            if start + end > ELASTICSEARCH_MAX_RESULT_WINDOW:
+            query_size = start + end
+            cannot_fulfill_query_size = (
+                query_size > ELASTICSEARCH_MAX_RESULT_WINDOW or
+                query_size > search_results.hits.total.value
+            )
+            if cannot_fulfill_query_size:
                 return results
 
             s = s[start:end]
```

After applying the fix, we should see a single call to `_post_process_results` for the example query.

sarayourfriend:
## Fixes

Fixes #2415 by @sarayourfriend 

## Description

Stops `_post_process_results` recursion once the total available hits have been exhausted. Previously, if queries were unable to fill the page_size with live links, `_post_process_results` would continue to iterate until the ES max page depth was met.

For some reason the `Hit` `__repr__` was suddenly failing for me and causing tests to incorrectly fail (call argument mocks use `repr`, apparently). The fix for this was to ensure the document `_id` is a string. Here is the error I was seeing:

```
self = <[TypeError('sequence item 1: expected str instance, int found') raised in repr()] Hit object at 0x7f34bb326c90>

    def __repr__(self):
        return "<Hit({}): {}>".format(
>           "/".join(
                getattr(self.meta, key) for key in ("index", "id") if key in self.meta
            ),
            super(Hit, self).__repr__(),
        )
E       TypeError: sequence item 1: expected str instance, int found
```

`self.meta.id` was an `int`. Changing the factory to pass a string here (which does match ES's actual response type) fixed the issue.

## Testing Instructions

Make sure you understand how to reproduce the error on `main`, follow the reproduction instructions in the issue.

Once you've done that, check out this branch and try to reproduce the issue. You should not be able to. With the example in the issue, you should see a single call to `post_process_results`.

The unit test attempts to cover the case where additional calls to `post_process_results` are needed to fill the first or last page of a query. It becomes increasingly complex to test the more calls to `post_process_results` we want to assert, but I can try to add more unit tests if we want examples of recursion depth greater than 2 in the unit tests. The unit tests are also fairly complex, if reviewers have any ideas for how to simplify (and especially how to _clarify_ them), please share.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -205,9 +205,25 @@ def _post_process_results(
             end = 90 + 45
             ```
             """
-            end += int(end / 2)
-            if start + end > ELASTICSEARCH_MAX_RESULT_WINDOW:
+            if end >= search_results.hits.total.value:
+                # Total available hits already exhausted in previous iteration
                 return results
+
+            end += int(end / 2)
+            query_size = start + end
+            if query_size > ELASTICSEARCH_MAX_RESULT_WINDOW:
+                return results
+
+            # subtract start to account for the records skipped
+            # and which should not count towards the total
+            # available hits for the query
+            total_available_hits = search_results.hits.total.value - start
+            if query_size > total_available_hits:
+                # Clamp the query size to last available hit. On the next
+                # iteration, if results are still insufficient, the check
+                # to compare previous_query_size and total_available_hits
+                # will prevent further query attempts
+                end = search_results.hits.total.value
 
             s = s[start:end]
             search_response = s.execute()
--- a/api/api/utils/dead_link_mask.py
+++ b/api/api/utils/dead_link_mask.py
@@ -1,5 +1,5 @@
+import django_redis
 from deepdiff import DeepHash
-from django_redis import get_redis_connection
 from elasticsearch_dsl import Search
 
 
@@ -32,7 +32,7 @@ def get_query_mask(query_hash: str) -> list[int]:
     :param query_hash: Unique value for a particular query.
     :return: Boolean mask as a list of integers (0 or 1).
     """
-    redis = get_redis_connection("default")
+    redis = django_redis.get_redis_connection("default")
     key = f"{query_hash}:dead_link_mask"
     return list(map(int, redis.lrange(key, 0, -1)))
 
@@ -44,7 +44,7 @@ def save_query_mask(query_hash: str, mask: list):
     :param mask: Boolean mask as a list of integers (0 or 1).
     :param query_hash: Unique value to be used as key.
     """
-    redis_pipe = get_redis_connection("default").pipeline()
+    redis_pipe = django_redis.get_redis_connection("default").pipeline()
     key = f"{query_hash}:dead_link_mask"
 
     redis_pipe.delete(key)
--- a/api/test/factory/es_http.py
+++ b/api/test/factory/es_http.py
@@ -1,0 +1,79 @@
+from uuid import uuid4
+
+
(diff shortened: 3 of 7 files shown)


sarayourfriend:
I'm not sure why the test fails in CI but passes locally. I'll try adding a more strict matcher on the request body in case something else is making the request and causing pook to use the wrong response, somehow. I cannot reproduce it locally, though :cry:

sarayourfriend:
Re-running just to see if the test is flaky or always wrong. It's failed several times now so I suspect CI is doing something slightly different that I'm not expecting... just not sure how to anticipate it.

sarayourfriend:
Aha! I'm able to reproduce it locally now by adding the redis fixture to the first test. Debugging now :slightly_smiling_face:

stacimc (commented):
Noting that if you test with search terms including `bird` as suggested, you'll need to pass `mature=true` as well to get results.

I was able to reproduce the bug on main and verify the fix here -- I can see that there are no more recursive calls 💯  I wanted to verify that dead links are still filtered but was not able to get a test working yet as I've gone down a few ES rabbit holes.

I do not get the errors that I see in CI when running the tests locally, but I *do* get multiple errors of the form:

```
E           pook.exceptions.PookNoMatches: pook error!
E
E           => Cannot match any mock for the following request:
E           ==================================================
E           Method: POST
E           URL: http://es:9200/image-filtered/_search
```

I'm having trouble following the tests. Do you mind commenting them more extensively?

sarayourfriend:
> I'm having trouble following the tests. Do you mind commenting them more extensively?

Can you share where you're losing track specifically? It's hard to know exactly what to comment without potentially over-complicating the explanation. Unfortunately this part of the code is complicated and the test complexity mirrors that. T

obulat (approved):
This is a really welcome change! From 11 to 1 requests 👍

stacimc (approved):
> I'm going to mock out SearchContext to remove at least that part of the test implementation details.

This honestly goes a long way to simplify the test setup and helps considerably -- thanks for humoring me :)

PR #2423, opened 2023-06-16
Add APPLY_FILTER analytics event

### LINKED ISSUE #1070: Analytics event: `APPLY_FILTER`
## Description

The goal is to 

- add the type for event `APPLY_FILTER` to the `Events` type.

## Reference

https://github.com/WordPress/openverse/blob/839418fabb0972ec3c3ccf48c72fd64cbba03cb9/rfcs/analytics/20221006-implementation_plan_frontend.md?plain=1#L170

## Additional information

- [RFC](https://docs.openverse.org/projects/proposals/analytics/20221006-implementation_plan_frontend.html#events)

obulat:
## Fixes

Fixes #1070 by @dhruvkb

## Description

This PR adds `APPLY_FILTER` event sent when the user selects a filter.
We might want to debounce it, I'm not sure if I should add it here?

Some unrelated changes were also added to clean up the unused ref, and use `storeToRefs` pinia helper to reduce the number of computeds used in the file.

## Testing Instructions

Try clicking on the filters (both in the modal on mobile screen, and the sidebar on the desktop). You should see the event logged in the console, and tracked in Plausible at 0.0.0.0:50288 (if you run `just frontend/up` and `just frontend/init` as described in https://docs.openverse.org/frontend/guides/analytics.html).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VFilters/VSearchGridFilter.vue
+++ b/frontend/src/components/VFilters/VSearchGridFilter.vue
@@ -121,10 +121,11 @@ export default defineComponent({
       filterType: NonMatureFilterCategory
       code: string
     }) => {
-      searchStore.toggleFilter({ filterType, code })
+      const checked = searchStore.toggleFilter({ filterType, code })
       sendCustomEvent("APPLY_FILTER", {
-        key: filterType,
-        value: code,
+        category: filterType,
+        key: code,
+        checked,
         searchType: searchType.value,
         query: searchTerm.value,
       })
--- a/frontend/src/components/VFilters/VSearchGridFilter.vue
+++ b/frontend/src/components/VFilters/VSearchGridFilter.vue
@@ -18,7 +18,7 @@
         {{ $t("filterList.clear") }}
       </VButton>
     </header>
-    <form ref="filtersFormRef" class="filters-form">
+    <form class="filters-form">
       <VFilterChecklist
         v-for="filterType in filterTypes"
         :key="filterType"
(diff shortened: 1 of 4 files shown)


sarayourfriend (commented):
If we want to debounce this it should be added here, otherwise there will be a time span where the event happens _far more often_ than otherwise, making it harder to understand the rate of the event being sent per session.

However, if we debounced it, what would we send? In debouncing, would we still send individual events per filter value changed? In that case, don't bother debouncing. We only debounce the filter changes being sent to search to prevent unnecessary search requests. If we debounced and combine all the updated filters into a single event, that might be worth it, but would slightly change the way this event would be queried. It would probably warrant a slightly different API to make it easier to tell both what filters are applied and how often a particular filter is applied then unapplied (which might imply the filter isn't as useful as others).

zackkrida:
Olga and I chatted a bit yesterday and we concluded we should try without debouncing these. The only reason to debounce would be perceived performance for the user or performance of the `/api/event` endpoint and we don't suspect either of those will be an issue here.

zackkrida (changes requested):
This works perfectly. However, I think we're missing a property for if the filter is being enabled or disabled in the payload. This was the original intention of the "value" property, but the way it's implemented here makes sense to me so I think we just need an additional property.

dhruvkb (approved):
LGTM! Works well.

zackkrida (approved):

PR #2427, opened 2023-06-16
Remove internal__index param

stacimc:
## Fixes

Fast follow to #2209

## Description

The `internal__index` query param was added in #2073 with the intention of using it in an index readiness check during the data refresh. #2209 instead implemented this check by querying the ES index directly, instead of going through the API (which would have required more complexity to handle authentication).

Consequently the `internal__index` param is no longer needed and can be removed. This PR reverts #2073. Pinging @dhruvkb and @zackkrida to confirm that there are definitely no other intended use cases for this param.

## Testing Instructions

Tests should all pass. You should also try making a few authenticated and unauthenticated API requests just to smoke test.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -296,7 +296,6 @@ def _resolve_index(
 def search(
     search_params: media_serializers.MediaSearchRequestSerializer,
     origin_index: OriginIndex,
-    exact_index: bool,
     page_size: int,
     ip: int,
     filter_dead: bool,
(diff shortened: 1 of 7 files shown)


krysal (changes requested):
The param will still be used for the [Search Relevancy Sandbox](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/index.html) project. Although we need to define better security policies for it.

stacimc:
Thanks for letting me know, @krysal! I'll go ahead and close the PR.

dhruvkb:
@statimc another reason for going via the API instead of directly checking ES was that we could validate the optimal end result of getting search results from the API. Checking ES directly leaves the possibility of some snag after the ES layer preventing results from being delivered.

zackkrida:
@dhruvkb since the implementation as-written in #2209 will catch the incident which inspired it, I think we could move ahead with that PR as-is. We have many other ways of checking if the API and ES are communicating appropriately. 

@stacimc has had to reimplement that PR a few times and I think the current set of tradeoffs are preferable since they allow us to bypass all the authentication work connecting directly to the API would require.

PR #2447, opened 2023-06-19
Add a popover storybook test

### LINKED ISSUE #2215: VPopover story is broken
## Description

https://docs.openverse.org/storybook/?path=/story/components-vpopover--default-story is broken with the error message below:

```
state.rects is undefined

detectOverflow@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:98616
detectMaxHeight@https://docs.openverse.org/storybook/main.97df8a96.iframe.bundle.js:2:1543821
./src/components/VPopover/VPopoverContent.vue/usePopper/<@https://docs.openverse.org/storybook/main.97df8a96.iframe.bundle.js:2:1543993
invokeWithErrorHandling@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2576284
call@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2572334
../node_modules/.pnpm/vue@2.7.14/node_modules/vue/dist/vue.js/</factory/doWatch/watcher.run@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2573785
flushSchedulerQueue@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2569913
../node_modules/.pnpm/vue@2.7.14/node_modules/vue/dist/vue.js/</factory/nextTick/<@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2577873
flushCallbacks@https://docs.openverse.org/storybook/vendors~main.2636c955.iframe.bundle.js:2:2577061
```

### LINKED ISSUE #2220: v-popover.spec test "hideOnClickOutside -> should hide the popover if a click happens outside the popover by default" is broken

### LINKED ISSUE #2300: "VPopover › accessibility › autoFocusOnShow › should focus the first tabbable element in the popover by default and not warn" test broken
## Description

The test `VPopover › accessibility › autoFocusOnShow › should focus the first tabbable element in the popover by default and not warn` in `v-popover.spec.js` is broken.

This was caught in https://github.com/WordPress/openverse/pull/2199

The test currently passes a callback to `nextTick`. However, `nextTick` takes no arguments. The assertions present in the test are never run. The PR linked above changes this and pulls them out of `nextTick` and finds that assertions do not pass.

This may be solved by #2226 so cc @obulat who opened that PR.

### LINKED ISSUE #2316: Remove implicit `@testing-library` `get*` assertions: `v-popover.spec.js`
## Current Situation

See https://github.com/WordPress/openverse/discussions/2198 for description of issue.

## Suggested Improvement

Remove the `get*` functions from the `eslint` comment from the top of `v-popover.spec.js` and update all tests to have explicit assertions. `expectOpen` can be left so long as it uses an explicit assertion. `doOpen` and `expectClosed` should be removed along with the `get*` functions as they no longer exist in the module.

Example transformations:

```js
screen.getByRole("button") // throws if not present
```

should be

```js
expect(screen.findByRole("button")).toBeInTheDocument()
```

## Benefit

See https://github.com/WordPress/openverse/discussions/2198

obulat:
## Fixes

Fixes #2300 by @sarayourfriend
Fixes #2215 by @sarayourfriend 
Fixes #2316 by @sarayourfriend 
~~Fixes #2220 by @sarayourfriend~~ Removes the doc comment with a link to #2220 because the test had actually been fixed

## Description

This PR adds a Storybook test for the `VPopover` component. It tests that when a popover opens, by default, the first tabbable element inside the popover is focused.
~~This is not a visual regression test, so I put it into `storybook/e2e` folder to match the structure of `playwright` folder, but maybe `storybook/functional` would be a better name?~~
I've put this test into `storybook/functional` folder as it is not a vr test.

This PR also adds the required `variant` and `size` parameters to `VButton`s in the broken stories.

_Edited to add_: I've also removed the `.skip` annotations on tests. The "should hide the popover if a click happens outside the popover by default" test was fixed.
I did not replace the `getBy` with `queryBy` functions in the popover test because they are not used for `expect`s, but are actually used to get the element. Because they only return an element (or throw an error) instead of <element|null>, we won't have to do the `null` checking in the tests, and the tests would be much simpler.

## Testing Instructions

The CI should pass. The Storybook in the docs preview link in the comment below should have working stories for `VPopover`, `VItemGroup`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
>  This is not a visual regression test, so I put it into storybook/e2e folder to match the structure of playwright folder, but maybe storybook/functional would be a better name?

I like "functional" as well. "Behavioural" might also be appropriate to match the BDD-ish nature of these kinds of tests. Personally functional makes more sense to me without being to jargon-y :slightly_smiling_face: 

Can we remove the skipped test in this PR as well then? Should other popover tests be moved to the storybook approach, if it is more stable? :thinking:

obulat:
> Can we remove the skipped test in this PR as well then? Should other popover tests be moved to the storybook approach, if it is more stable? 🤔

We might want to do that, but I'd prefer not to do it in this PR.
I unskipped the tests (and one that was fixed but was still skipped 😨 )

sarayourfriend (approved):
LGTM!

> We might want to do that, but I'd prefer not to do it in this PR.

With respect to this, I suppose it would be worth us having documented guidelines/thought-process for when a storybook functional test is appropriate instead of a testing library unit test. Are these only a fallback option? The unit tests start up much faster, so it's nicer to debug them. Is that good enough reason to preference them over this approach when possible?

dhruvkb (approved):
Thanks for fixing so many issues. It feels a bit wrong to be [running a functional test in the `storybook_vr` job](https://github.com/WordPress/openverse/actions/runs/5321297412/jobs/9636730154?pr=2447#step:4:261) so we should either

- split functional tests into their own job (if there will be many more such functional tests to follow)
- rename the job to indicate that it's not just VR tests (if this is an exceptional case)
 
That's not a blocker though, so approving. Thanks again, this is very nice!

obulat:
> With respect to this, I suppose it would be worth us having documented guidelines/thought-process for when a storybook functional test is appropriate instead of a testing library unit test. Are these only a fallback option? The unit tests start up much faster, so it's nicer to debug them. Is that good enough reason to preference them over this approach when possible?

You've actually written the guidelines before, @sarayourfriend  :) I've added more details to the docs in a follow-up PR: #2460. 

@dhruvkb, I've renamed the storybook step from `storybook_vr` to `storybook` in #2460. I think it would make sense to split the steps only when we have a critical mass of functional storybook tests.

PR #2449, opened 2023-06-19
Analytics Event: `LOAD_MORE_RESULTS`

zackkrida:
## Fixes

Fixes #1074 by @dhruvkb 
Fixes #1992 by @sarayourfriend 

## Description

This PR adds a `LOAD_MORE_RESULTS` analytics event to the load more button. ~Crucially, I did _not_ include the original page number property in the event payload, as it requires major additions to the media and search stores.~ Thanks, @obulat.

## Testing Instructions

CI should pass, and then review the code. Let me know what you think of this API surface! Should we even be testing analytics events in e2e tests? I suspect it could be a nice additional assurance. If so we might want to add it in other places, like for pageviews and other custom events. That gets into larger issues about our frontend test consistency and coverage though.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [x] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
[...]

PR #2454, opened 2023-06-20
Analytics: Add REACH_RESULT_END event

### LINKED ISSUE #1072: Analytics event: `REACH_RESULT_END`
## Description

The goal is to 

- add the type for event `REACH_RESULT_END` to the `Events` type.

## Reference

https://github.com/WordPress/openverse/blob/839418fabb0972ec3c3ccf48c72fd64cbba03cb9/rfcs/analytics/20221006-implementation_plan_frontend.md?plain=1#L207

## Additional information

- [RFC](https://docs.openverse.org/projects/proposals/analytics/20221006-implementation_plan_frontend.html#events)

obulat:
## Fixes

Fixes #1072 by @dhruvkb

## Description

The main goal of this PR was to add the `REACH_RESULT_END` event when the user scrolls down the search results page and sees the "Load More" button.

To add this event, I used the `useElementVisibility` composable from Vueuese. However, because the page was not scrolled to top when the user changes the search type, when the user opened a second page of images, for example, and then switched search type, the `REACH_RESULT_END` event would be sent right after the "Load More" button was clicked, before the next page was even loaded.

This is why I also added scroll to top functionality to the search result page so now when the user changes the search type, the results page scrolls to top.
This change surfaced a problem where the search type changed, and there were still results from the previous search type, and the change of the media and search title/content links was visibly inconsistent after the page scrolled to top. To fix this, I added `clearMedia` call to the function that changes the search type so that the page displays a loading skeleton once it scrolls up. For all content page, when the fetchState's `isFetching` was false before the fetching started, this wouldn't show the loading skeleton, and the external search form was visible right below the content links. This is why I had to update the condition for showing the loading skeleton to include `!hasStarted`.

To add the `resultPage` to the event payload, I updated the `mediaStore` to keep track of the current page. @zackkrida, this might be useful for #2449 (I found a relatively simple way of adding the page, without major changes to the media and search stores)

I also did some cleanup in the `search.vue`: moved the `fetchMedia` and `routeWatcher` inside the `setup` function, and used `storeToRefs` to simplify the store values.

All of this is to say that seemingly small addition required a lot of changes in the code, this is why the PR is large.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VLicense/VLicense.vue
```diff
@@ -62,10 +62,10 @@ export default defineComponent({
 
     const iconNames = computed(() => getElements(props.license))
     const licenseName = computed(() => {
+      const licenseKey =
+        props.license === "sampling+" ? props.license : camel(props.license)
       return {
-        readable: i18n
-          .t(`licenseReadableNames.${camel(props.license)}`)
-          .toString(),
+        readable: i18n.t(`licenseReadableNames.${licenseKey}`).toString(),
```
zackkrida, line 68: Assuming this is an unrelated bug? Seems fine to include here if it's non-urgent.

frontend/src/components/VLoadMore.vue
```diff
@@ -26,11 +30,18 @@ export default defineComponent({
...
+    const loadMoreButtonRef = ref(null)
     const i18n = useI18n()
     const mediaStore = useMediaStore()
     const searchStore = useSearchStore()
 
-    const isFetching = computed(() => mediaStore.fetchState.isFetching)
+    const { sendCustomEvent } = useAnalytics()
+
+    // Use the `_searchType` from mediaStore because it falls back to ALL_MEDIA
+    // for unsupported search types.
```
zackkrida, line 41: Does this matter for the load more button? We don't show it on unsuppoorted media types, right? So perhaps it'd be better to get the searchType from the search store.

obulat: This appeases Typescript 😆

sarayourfriend (approved):
LGTM, tested fine locally.

frontend/test/playwright/e2e/load-more.spec.ts
```diff
@@ -125,6 +130,32 @@ test.describe("Load more button", () => {
         })
         await expect(page.locator(loadMoreButton)).not.toBeVisible()
       })
+
+      /**
+       * Checks that an analytics event is posted to /api/event and has the correct
+       * payload for the REACH_RESULT_END event.
+       */
+      test(`Sends a valid analytics event when user reaches the load more page`, async ({
```
AetherUnbound, line 138: Should we have a second test or a second piece to this test to ensure that the event is fired off again after load more is clicked and that the event appropriately has page 2 in the payload?

AetherUnbound (approved):
This looks good and tested as expected locally!

frontend/test/playwright/e2e/load-more.spec.ts
zackkrida: I'll look into adding one, I think that would be wise.

zackkrida: Done 👍

zackkrida:
I made two small changes here:

- Fix the code to work after changing the search query when the user has already pressed "load more" once. There was a bug.
- Fix the code to work on the last page, when the "load more" button is hidden. This was accomplished by watching a wrapping div for visibility instead of the load more button itself. This way we still track when the final page of a search is observed, and searches with only a single page of results (and no load more button) are tracked correctly.

AetherUnbound:
Nice catches!

Issue #2472, opened 2023-06-22
Add `dnsmasq` and cache DNS queries in API docker image


sarayourfriend:
## Problem

The deployed API docker image is based on the `-slim` Python image, which is Debian stripped of anything not strictly required to run Python. That includes the removal of `dnsmasq` which means no DNS queries are being stashed and every outbound request to a URL requires a new DNS query.

## Description

Install `dnsmasq` in the Dockerfile and configure it according to the Debian documentation: https://wiki.debian.org/dnsmasq?action=show&redirect=HowTo%2Fdnsmasq#Local_Caching

## Additional context

Related to DNS resolution sentry issues (I can't log into Sentry at the moment to share a link and there's no issue created for it for me to link to yet).

sarayourfriend:
This is actually impossible with the described solution. It's not possible to get dns caching inside a docker container without severely interfering with Docker's networking (which is complex enough as it is). Additionally, it's not possible to use a sidecar container as a DNS cache because [FARGATE launch type does not support the `dnsServers` configuration setting required](https://github.com/aws/containers-roadmap/issues/1123).

I think we could manage to wrangle something by using Route 53 Resolvers, but that's a paid service :cry: I'll look around and see if there's any way to at least cache the database DNS.

sarayourfriend:
One alternative that comes to bind is to use [PgBouncer](https://www.pgbouncer.org/) on a small EC2 instance that we could reference by VPC IP rather than domain and then setup DNS caching _there_ instead. The PgBouncer box would cache the DNS queries for the RDS database.

Doing so would still require us to cache with a max TTL of 30 seconds, [as per the RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_BestPractices.html#CHAP_BestPractices.DiskPerformance).

sarayourfriend:
Just to add more colour to the problem: https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-dns-limits

The quota is 1024 requests per second per network interface. That's pretty massive, and it's wild that we're hitting it. I wonder if there's something misconfigured in our API service that could cause this?

sarayourfriend:
[This Django plugin might be an easier (and definitely cheaper) alternative to PgBouncer](https://github.com/lcd1232/django-postgrespool2).

zackkrida:
[...]

PR #2475, opened 2023-06-23
Add CONN_MAX_AGE setting to API

sarayourfriend:
## Fixes

(Maybe) fixes https://github.com/WordPress/openverse/issues/2472 by @sarayourfriend 

## Description

[`CONN_MAX_AGE`](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-CONN_MAX_AGE) defaults to 0 which means Django is opening a new connection for every single database request. This may help lower the number of DNS requests coming from our API to avoid exceeding the quota. Even if it doesn't this setting would remove a small amount of overhead required for each database connection that can presently be avoided.

## Testing Instructions

Integration tests should pass. Run the API locally and make some search requests and other things that exercise the database. There should be no change in perceivable behaviour.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/conf/settings/databases.py
+++ b/api/conf/settings/databases.py
@@ -12,6 +12,8 @@ DATABASES = {
         "USER": config("DJANGO_DATABASE_USER", default="deploy"),
         "PASSWORD": config("DJANGO_DATABASE_PASSWORD", default="deploy"),
         "NAME": config("DJANGO_DATABASE_NAME", default="openledger"),
+        # Default of 5 matches RDS documentation's advised max DNS caching time
+        "CONN_MAX_AGE": config("DJANGO_CONN_MAX_AGE", default=5),
         "OPTIONS": {
             "application_name": config(
                 "DJANGO_DATABASE_APPLICATION_NAME", default="openverse-api"


api/conf/settings/databases.py
```diff
@@ -12,6 +12,8 @@
         "USER": config("DJANGO_DATABASE_USER", default="deploy"),
         "PASSWORD": config("DJANGO_DATABASE_PASSWORD", default="deploy"),
         "NAME": config("DJANGO_DATABASE_NAME", default="openledger"),
+        # Default of 5 matches RDS documentation's advised max DNS caching time
```
zackkrida, line 15: Could you help me understand using this DNS caching time to determine how long we should leave open DB connections? My initial suspicion would be that this number could be a lot higher and I'm not seeing how the DNS cache time relates.

krysal: I'm curious too. Is this in seconds?

zackkrida: It is in seconds, @krysal 👍

sarayourfriend: This probably comes down to my lack of understanding of TCP/IP. If the IP address of one of the ends of the connection changes, would the connection break? The reason RDS says to only cache DNS requests to 5 seconds is because the IP address of the database can change (I assume with some frequency and overlap that 5 seconds wouldn't cause an issue). I made an assumption that this would have an effect on long connections, but I could be wrong about that. Do you know for certain either way?

We could try it at a higher number and reduce it, if you like. It's just an environment variable though so we don't necessarily need to decide that at this point and can change it as we please in production using the environment variable.

sarayourfriend: Did a quick search around and confirmed that my assumption is correct. A [TCP connection cannot be maintained if the IP of either side changes during the connection](https://networkengineering.stackexchange.com/questions/12565/how-is-a-tcp-connection-maintained-over-a-changing-ip-address). I think this implies that we cannot reliably keep connections alive for longer than RDS recommends caching the DNS response.

zackkrida: Ah, I see. You're thinking is that matching the DNS cache time will ensure that connections aren't left open, resolving to an IP address that's no longer valid, like in the case of a failover or any other reason the database IP might change. I hadn't considered that.

Where are you seeing the 5 second documentation? In this ["Amazon RDS basic operational guidelines"](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_BestPractices.html) guide I _think_ I see 30 seconds recommended:

> If your client application is caching the Domain Name Service (DNS) data of your DB instances, set a time-to-live (TTL) value of less than 30 seconds. The underlying IP address of a DB instance can change after a failover. Caching the DNS data for an extended time can thus lead to connection failures. Your application might try to connect to an IP address that's no longer in service.

One more thing. I see a new-ish Django setting added in 4.1 which will run a healthcheck on persistient connections before reusing them https://docs.djangoproject.com/en/4.2/ref/settings/#conn-health-checks. Perhaps that + a `CONN_MAX_AGE` of 60 would be reasonable? 

I'm also content to start with 5 seconds and experiment with larger values in the future. Thanks for explaining 😄

sarayourfriend: I must have been reading https://docs.aws.amazon.com/whitepapers/latest/amazon-aurora-mysql-db-admin-handbook/dns-caching.html which is Aurora specific, I now realise. We can raise this to 30 seconds, good catch.

`CONN_HEALTH_CHECKS` looks good to enable as well, will do!

zackkrida (approved):
LGTM. I think this will be an awesome starting point and great improvement.

api/conf/settings/databases.py
zackkrida: Awesome, PR approved. I learned a lot here about DNS caching, thanks again for your patience!

AetherUnbound (approved):
Changes look great, hopefully this is a sufficient fix! 🤞🏼

api/conf/settings/databases.py
```diff
@@ -12,6 +12,12 @@
         "USER": config("DJANGO_DATABASE_USER", default="deploy"),
         "PASSWORD": config("DJANGO_DATABASE_PASSWORD", default="deploy"),
         "NAME": config("DJANGO_DATABASE_NAME", default="openledger"),
+        # Default of 30 matches RDS documentation's advised max DNS caching time
```
krysal, line 15:
```suggestion
        # Default of 30s matches RDS documentation's advised max DNS caching time
```

dhruvkb (approved):
I think having 'DATABASE' in the env var names would be helpful and more consistent. But nothing so serious as to block.

krysal (approved):
LGTM

PR #2479, opened 2023-06-24
Make `load_sample_data` more resilient

### LINKED ISSUE #2359: `TableIndexer` has no attribute `load_test_data`
## Description

#2322 removed the `load_test_data` method from `TableIndexer`. This will be fixed in a PR I will put up soon but a separate critical issue needs to evaluate why this didn't cause CI to fail. With this error locally, all local indexes end up with 0 results.

## Reproduction

1.  Checkout `main` and run `just down -v`, `just api/up` then `just api/init`. See that `LOAD_TEST_DATA` actions both fail, but the rest of init proceeds. Check your Elasticsearch cluster locally and see that there are zero documents.

```
$ curl http://localhost:50292/_cat/indices/image\?v\=true
health status index      uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   image-init 0jnENyXiTrCy-NeNWB6Siw  18   0          0            0      3.6kb          3.6kb
```

## Environment

- Device: 
- OS: 
- Browser: 
- Version: 
- Other info:

dhruvkb:
## Fixes

Addresses #2359 by @sarayourfriend

## Description

This PR makes these changes:
- ensures that invalid API calls to the ingestion server using cURL fail locally for non-2xx response codes.
- checks for sample data to be loaded in SQL after issuing `DELETE` and `\copy` to the database.

This PR does not rewrite the script in a different language, like Python, for two reasons:
- It has been tried before (WordPress/openvere-api#830) but it was turned down as it was too much rewriting for the little value gained.
- The current structure of the script is largely a collection of `just` recipe invocations. Switching to Python will not improve any inherent fragility in those recipes.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/justfile
+++ b/ingestion_server/justfile
@@ -62,12 +62,17 @@ wait-up: up
 
 # Make a cURL POST request to the service with the given data
 _curl-post data host="localhost:50281":
-    curl \
+    STATUS_CODE=$(curl \
       -X POST \
       -H 'Content-Type: application/json' \
       -d '{{ data }}' \
-      -w "\n" \
-      'http://{{ host }}/task'
+      -o /dev/stderr \
+      -w "%{http_code}" \
+      'http://{{ host }}/task'); \
+    if [ $STATUS_CODE -lt 200 ] || [ $STATUS_CODE -ge 300 ]; then \
+      echo "Status: $STATUS_CODE"; \
+      exit 1; \
+    fi
 
 # Load sample data into temp table in API and new index in Elasticsearch
 ingest-upstream model="image" suffix="init":
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -12,7 +12,7 @@ DB_SERVICE_NAME="${DB_SERVICE_NAME:-db}"
 # Load sample data
 function load_sample_data {
   docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" bash -c "psql <<-EOF
-    DELETE FROM $1;
+		DELETE FROM $1;
 		\copy $1 \
 			from './sample_data/sample_$1.csv' \
 			with (FORMAT csv, HEADER true);
@@ -20,8 +20,20 @@ function load_sample_data {
 		EOF"
 }
 
+function verify_loaded_data {
+  COUNT=$(docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" bash -c "psql -AXqt <<-EOF
+		SELECT COUNT(*) FROM $1;
+		EOF")
+  if [ "$COUNT" -ne 5000 ]; then
+    echo "Error: table $1 count differs from expected."
+    exit 1
+  fi
+}
+
 load_sample_data "image"
+verify_loaded_data "image"
 load_sample_data "audio"
+verify_loaded_data "image"
 
 #######
 # API #
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -33,7 +33,7 @@ function verify_loaded_data {
 load_sample_data "image"
 verify_loaded_data "image"
 load_sample_data "audio"
-verify_loaded_data "image"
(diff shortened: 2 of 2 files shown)


load_sample_data.sh
```diff
@@ -12,16 +12,28 @@ DB_SERVICE_NAME="${DB_SERVICE_NAME:-db}"
...
+  if [ "$COUNT" -ne 5000 ]; then
+    echo "Error: table $1 count differs from expected."
+    exit 1
+  fi
+}
+
 load_sample_data "image"
+verify_loaded_data "image"
 load_sample_data "audio"
+verify_loaded_data "image"
```
zackkrida, line 36:
```suggestion
verify_loaded_data "audio"
```

zackkrida (approved):
One major error resulting from a typo, but it's such a quick fix that I'll commit it and give this the approval. 

Edit: typo fixed in b53e266d3ce5c6b95ebce04e9e5678d811d04c26

ingestion_server/justfile
```diff
@@ -62,12 +62,17 @@ wait-up: up
...
       -d '{{ data }}' \
-      -w "\n" \
-      'http://{{ host }}/task'
+      -o /dev/stderr \
+      -w "%{http_code}" \
+      'http://{{ host }}/task'); \
+    if [ $STATUS_CODE -lt 200 ] || [ $STATUS_CODE -ge 300 ]; then \
+      echo "Status: $STATUS_CODE"; \
+      exit 1; \
+    fi
```
sarayourfriend, line 75: I was surprised by this because I assumed cURL supported some kind of method for changing the exit code if the status was bad. It turns out there's a relatively new feature (2021) that does this: [`--fail-with-body`](https://curl.se/docs/manpage.html#--fail-with-body). It has the added benefit of still printing the response body with the failure without needing to add any more flags (or checks).

sarayourfriend (approved):
This is a nice change, but I don't think it resolves the linked issue. Calls to the ingestion server return a task ID. If that task fails the script will continue on without issue. The call to ingest the test data, for example, only checks if the index is created afterwards before moving on, not that the task actually finished or succeeded.

The reason to convert to Python is that it would be way easier to parse the request body for the task ID and make repeated requests to the task endpoint of the ingestion server. [The PR you mentioned](https://github.com/WordPress/openverse-api/pull/830) didn't do that either, to be fair, and I think overcomplicated things by reusing the just recipes and adding new functionality, rather than converting the script as is. I don't think that previous PR is necessarily a good comparison or an indication that re-writing to Python doesn't make certain aspects of it easier. The linked PR did a _lot_ more than re-write to Python and add checks on the task IDs.

Anyway, this is a change that we should merge, but I don't think it fully resolves the issue as the tasks can still fail and the script would go on anyway.

ingestion_server/justfile
dhruvkb: I didn't use that because a lot of us are developing on macOS and might have an older version of cURL built-in. Installing a new version of cURL is technically an option but I experienced SSL issues when I last did that and I didn't want to force an additional system-level dependency.

dhruvkb:
I changed "Fixes ..." to "Addresses ..." to leave the linked issue open after merging this PR.

ingestion_server/justfile
sarayourfriend: Ah, TIL that macOS is on an old version :+1:

dhruvkb: Not only cURL, macOS comes bundled with plenty of old utils. 

I also had problems with Git when I tried to use some new feature (can't recall what it was), didn't have it, installed Git with Brew which then refused to work with SSH proto. It was a whole thing to reset everything 😢.

PR #951, opened 2023-03-20
Update TS configuration to use composite projects and fix VSCode integration

sarayourfriend:
## Description

I was going to work on #504 but ran into VSCode spitting a hundred TypeScript errors. Since the monorepo change, we need to update the configuration to make it work with Volar (the Vue plugin for VSCode that also handles `vue-tsc` integration).

1. Needed to make the project composite. This was done by adding a root level tsconfig.json that only acts as a reference to the single existing TS project in the monorepo, the frontend. `files: []` is present to prevent the root project from ever trying to compile or typecheck files.
2. Update pre-commit hook to compile from the root project using `-b` or `--build`. This tells tsc to build all the projects part of the composite project. Because this requires emitting declarations (`-b --noEmit` throws an error), I've added an output directory for the generated type definitions that is ignored by git.
3. After Volar finally worked, it complained about a handful of errors. There were some errors about `vueCompilerOptions` that to resolve required removing the deprecated options. There was also an error about needing to remove `@vue/runtime-dom` so I removed it. This had a cascading effect of needing to fix types that were being imported from it including CSSProperties. See the commit description for that commit for more information.
4. This raised a handful of errors with files that were already being included in the TypeScript project. For whatever reason, they now failed to compile when they didn't before. It makes perfect sense that they failed because, indeed, they relied on other files that were not typed. I fixed these by removing them from the tsconfig.json for the vue files and fixed the .ts files directly as it was a simple fix (replacing Breakpoint with RealBreakpoint to match the actual usage).
5. Finally, Olga mentioned to me that the SVG import types were broken. I believe this is a simple fix of just renaming the `typings/openverse` directory to `typings/svg`. I'm still not 100% clear on how resolution works there but it seems to have done the trick. If you import an SVG into a file included in the tsconfig, it should be typed as a string now (just like the declaration instructs it to).

I have an outstanding question, however. There are lots of files with `lang="ts"` set on the script tag but that are not included in tsconfig.json. Why is this? In VSCode, because the files are not included in any tsconfig.json, TypeScript uses the vanilla default configuration and all our aliased imports fail. It's very annoying to work in these files in VSCode sometimes the entire file can be underlined if types from typings are missing or all the aliased imports always.

**Update**: I talked to Olga about this and it sounds like WebStorm handles these cases differently from VSCode in that it ignores the errors. There might be some way to tell VSCode to ignore typescript files that don't have an associated tsconfig, but I haven't been able to find it in my brief search.

[...]

PR #2482, opened 2023-06-26
Add ESLint plugin package

### LINKED ISSUE #2366: Add custom ESLint plugin with inaugural rule to check frontend configuration correctness
## Current Situation

We have no way of creating complex ESLint rules. However, we need them for certain things like checking that the Events configuration on the frontend is correct (no overlapping payload data, correct names, etc).

## Suggested Improvement

Create a new `@openverse/eslint-plugin` package in a new private pnpm workspace in our monorepo that exports a custom ESLint plugin that checks the events configuration.

Hint: the plugin has already been written and tested in https://github.com/WordPress/openverse/pull/1130, we just need to re-write that PR since the frontend unit tests have been fixed and in light of ESLint configuration changes in #2200.

## Benefit

We can write complex ESLint rules. This also sets up our monorepo package strategy which would allows us to atomise parts of our codebase and share code more easily between packages.

## Additional context

Marked medium as it will help our analytics events maintain a consistent structure, which is important.

sarayourfriend:
## Fixes

Fixes #2366 by @sarayourfriend

## Description

This PR adds a new `packages` directory with a single new package, `@openverse/eslint-plugin`. It configures various things to support the packages:

- `tsconfig.json` is configured to be a monolithic configuration. This is necessary because `vue-tsc` does not support composite projects and even if it did, it would require emitting definition files which we don't need. Using a monolithic configuration is faster in IDEs and in type checking. This required making changes to places that assumed the tsconfig lived in the frontend directory. I ran into difficulties here because I made erroneous changes to the tsconfig (because I originally _tried_ to use a composite configuration, which vue-tsc doesn't support). These led to a lot of debugging and back and forth in this PR that you'll see in the comments below. However! Do not let that make it seem like this change is more complex than it is. Both the frontend unit tests breaking and the bundle size increases were caused by me incorrectly (and unintentionally) changing tsconfig settings that led to changes in how babel and webpack were transpiling files. 
- Adds new CI/CD checks to run package build and unit tests
- Adds `eslint` as a script to the root `package.json` to facilitate running the script with and without `--fix`

The diff stats look huge, but `pnpm-lock.yaml` accounts for ~ 2,653 changes: 816 additions & 1,837 deletions (i.e., the _vast_ majority of them). In reality there is very little _changing_ here other than the TypeScript configuration changes (described above). Almost all of this PR is aimed at adding the packages directory, CI/CD checks for those packages, and the new eslint plugin package along with tests and documentation.

The new eslint plugin has a single rule to check the analytics configuration and encodes the following assumptions:

1. Payload values must be numbers, boolean values, or strings
2. Events with no payload must still be configured on the `Events` type with a `never` payload type
3. Events must be named in screaming snake case
4. Events must not include payload members with keys overlapping the default keys included for all events

> **Note**
>
> I tried to think of ways to automatically configure the default props but couldn't think of any that made much sense. If you have any ideas, let me know!

> **Note**
>
> There are limitations to what the rule can do because it does not have access to type information. Therefore, it allows any type alias/reference, even if it could potentially be to an invalid type. For example, if a type `type Whatever = { foo: 'bar' }` existed, this would not be a valid payload prop value in practice, but `EVENT: { whatever: Whatever }` would pass the linting check. The TypeScript ESLint plugin does have a feature that allows you to write rules that consume type information and it would be nice to enable it one day, but it adds significant time to linting (TS needs to build the project first) and our project has heaps of errors generated by the recommended configuration of it that would need to be fixed first. These limitations mean we still need to be thoughtful when reviewing PRs that add analytics events. The rule aims to catch the vast majority of accidental misconfiguration.

## Testing Instructions

Run `pnpm --filter eslint-plugin run test:unit` to confirm the unit tests pass (though they should also pass explicitly in the new CI/CD check).

Run `pnpm run eslint` to run ESLint with the new rule configured. This will automatically run the build for the eslint-plugin package (required because ESLint only supports rules in plain JS and the package is written in TypeScript).

Make changes to `types/analytics.ts`'s `Events` type that you believe should be valid and confirm that there are no false positive errors. Make changes to the type that should not be valid and confirm that those do indeed raise an error.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


packages/eslint-plugin/src/configs/recommended.ts
```diff
@@ -0,0 +1,24 @@
...
+          "os",
+          "platform",
+          "browser",
+          "version",
+          "origin",
+          "pathname",
+          "referrer",
+          "width",
+          "height",
+        ],
```
zackkrida, line 20: Great!

packages/eslint-plugin/src/rules/analytics-configuration.ts
```diff
@@ -0,0 +1,217 @@
...
+
+type Options = readonly [
+  {
+    reservedPropNames: string[]
+  }
+]
+
+const messages = {
+  reservedPayloadPropNames: "{{ propName }} is a reserved payload prop name.",
+
```
zackkrida, line 11:
```suggestion
```

packages/eslint-plugin/src/rules/analytics-configuration.ts
```diff
@@ -0,0 +1,217 @@
...
+    return node.key.value?.toString()
+  }
+
+  throw new Error(
+    `Could not identify event name on node type: ${node.key.type}`
+  )
+}
+
+// Use without docs for now... should we add a section in the reference
+// documentation explaining custom rules?
```
zackkrida, line 67: Assuming this refers to external docs urls, like https://eslint.org/docs/latest/rules/no-const-assign, for example? I think it's okay to go without for now, but yes this would be a nice addition to the docs site, even if it was a single page with jump links for each rule.

zackkrida (approved):
LGTM. It's. really cool to see the test for the rules, and everything works locally. The analytics events rule works perfectly.

One nit with the test instructions in the PR, for other reviewers. The first step `pnpm --filter eslint-plugin run test:unit` doesn't built the eslint rule, so you should run `pnpm run eslint` first.

zackkrida:
Really confused by the actionlint CI failures here 🤔

packages/eslint-plugin/src/rules/analytics-configuration.ts
sarayourfriend: Yup, that's exactly it.

sarayourfriend:
> The first step pnpm --filter eslint-plugin run test:unit doesn't built the eslint rule, so you should run pnpm run eslint first

Ah! I didn't think the tests needed the build to run, good call. I should update the test script to run the build as well then.

packages/eslint-plugin/src/rules/analytics-configuration.ts
sarayourfriend: Actually, it's easy enough to add a doc site page for these, I'll go ahead and do so.

sarayourfriend:
> Really confused by the actionlint CI failures here 🤔

I had left out the `name:` block on the matrix and needed to add a new output variable in the `get-changes` step. actionlint was very helpful!

Changes are up now. I've added a documentation section for the packages generally with some broad documentation about the setup and how to add new packages. Additionally, I added a section for the `eslint-plugin` package specifically with documentation on the new rule.

sarayourfriend:
Now the same test issues that occurred the last time I tried this are happening :cry: Trying to debug it locally, maybe I can figure it out this time. It makes no sense to me though because there are zero changes, as far as I can tell, to the way the frontend gets built. Nothing in its configuration should have changed _at all_ but maybe because we shamefully-hoist or something stuff gets wonky.

sarayourfriend:
Finding that jest transformation cache was a god send as now I actually have something to debug against. It seems to come down to something about how the file is transformed. On `main`, VCopyButton imports clipboard like this:

```
  (cov_2rrewfj21m().s[3]++, _interopRequireDefault(require("clipboard")));
```

And on this branch it looks like this:

```
  (cov_2rrewfj21m().s[2]++, require("clipboard"));
```

Notice how on `main`, where the unit tests work, it is wrapped with `_interopRequireDefault`. Time to dig into where that comes from.

sarayourfriend:
After spending tons of time convinced it was an issue with babel, I finally found [this SO answer](https://stackoverflow.com/questions/42652423/error-default-is-not-a-constructor) that immediately fixed the issue :sob: Thank goodness. Hopefully everything works now...

I'm logging off for the day but if CI fails I will look into it more tomorrow.

.github/workflows/bundle_size.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           pattern: "./frontend/.nuxt/dist/client/**/*.{modern.js,js,css}"
           exclude: "{./frontend/.nuxt/dist/client/lang-*.json.js,./frontend/.nuxt/dist/client/lang-*.json.modern.js}"
           minimum-change-threshold: 10
+          build-script: --filter frontend build
```
sarayourfriend, line 57: This removes the need for the slightly confusing (in a now multi-buildable-package repo) root package.json `build` script.

sarayourfriend:
Whoa, those are huge increases in file size :cold_sweat: That is _not_ expected.

sarayourfriend:
Based on the bundle analysis on `main` and on this branch (`pnpm exec nuxt build --analyze` in the `frontend` directory will produce the analysis view for you) I think tree shaking isn't working properly on this branch. There's heaps of stuff present in this branch that isn't on main (`@sentry/replay`) and libraries that are in both are much larger on this branch, for some reason (`@vueuse/core` is 10kB parsed on `main` and 103 on this branch).

Gonna dig into what I did to cause this.

sarayourfriend:
Alright, it did indeed come down to module resolution issues (yet more of them!). After digging through the webpack bundle analyser stats I saw that `ts-loader` had heaps of issues. That led me to look back through the Nuxt typescript configuration and I noticed that I'd set module type to commonjs rather than ESNext [as suggested by the Nuxt TS plugin](https://typescript.nuxtjs.org/guide/setup). That fixes the bundle issue completely.

package.json
```diff
@@ -20,19 +21,24 @@
     "node": "18.16.0"
   },
   "devDependencies": {
-    "@babel/core": "7.20.12",
-    "@babel/eslint-parser": "7.19.1",
```
sarayourfriend, line 24: Neither of these dependencies are needed in the repository root. `@babel/eslint-parser` isn't needed at all (we use the typescript parser anyway).

package.json
```diff
@@ -20,27 +21,25 @@
...
+    "eslint-import-resolver-typescript": "3.5.5",
     "eslint-plugin-eslint-comments": "3.2.0",
-    "eslint-plugin-jest": "27.2.1",
     "eslint-plugin-import": "2.27.5",
+    "eslint-plugin-jest": "27.2.1",
     "eslint-plugin-prettier": "4.2.1",
     "eslint-plugin-tsdoc": "0.2.17",
     "eslint-plugin-unicorn": "42.0.0",
     "eslint-plugin-vue": "9.9.0",
     "eslint-plugin-vuejs-accessibility": "1.1.1",
```
sarayourfriend, line 40: Theoretically (potentially _ideally_) we could move our entire ESLint configuration into the plugin with our root `eslintrc` merely importing the plugin. That would allow us to get rid of all the ESLint dependencies (except ESLint and the openverse eslint plugin) in the root package.json and make things a bit cleaner here.

That's something for a future issue, however. Not worth adding the complexity to this PR.

frontend/test/playwright/playwright.config.ts
```diff
@@ -1,14 +1,5 @@
...
-
-import { addAliases } from "module-alias"
-
 import type { PlaywrightTestConfig } from "@playwright/test"
 
-addAliases({
-  "~": path.resolve(process.cwd(), "src"),
-  "~~": process.cwd(),
-})
-
```
sarayourfriend, line 11: This (and the same in the storybook playwright config) is no longer needed now that the directory structure of the playwright container matches the real structure, which in turn allows Playwright to use our regular tsconfig's paths configuration to resolve aliases.

frontend/package.json
```diff
@@ -40,7 +40,6 @@
     "test:storybook:local": "playwright test -c test/storybook",
     "test:storybook:debug": "PWDEBUG=1 pnpm test:storybook:local",
     "test:storybook:gen": "playwright codegen localhost:54000/",
-    "types": "vue-tsc --noEmit",
```
sarayourfriend, line 43: TypeScript config and runtime all lives in the repo root now.

frontend/package.json
```diff
@@ -100,14 +100,13 @@
     "vue-plausible": "^1.3.2"
   },
   "devDependencies": {
-    "@babel/core": "^7.20.12",
-    "@babel/parser": "^7.20.13",
-    "@babel/preset-env": "^7.20.2",
-    "@babel/preset-typescript": "^7.21.0",
-    "@babel/runtime-corejs3": "^7.21.0",
```
sarayourfriend, line 107: This dependency didn't need to be explicitly defined. We do not directly reference it anywhere and it is included anyway as a transient dependency of `preset-env`.

frontend/package.json
```diff
@@ -135,7 +132,6 @@
     "jest": "^26.6.3",
     "jest-fail-on-console": "^3.1.1",
     "jest-transform-stub": "^2.0.0",
-    "module-alias": "^2.2.2",
```
sarayourfriend, line 138: See the note on `playwright.config.ts` for why this is removed (no longer necessary)

tsconfig.json
```diff
@@ -1,8 +1,113 @@
...
+     * Some files from this section cannot be converted to TypeScript because they are
+     * used in non-TS contexts (like Tailwind configuration). Therefore they're left as
+     * "vanilla JavaScript" with JSDoc based type annotations. This allows us to check
+     * them with TypeScript without introducing syntax changes that are not compatible
+     * with other tools we rely on.
+     */
+    "frontend/src/constants/**/*",
+    "frontend/src/server-middleware/healthcheck.js",
+    /** End of non TypeScript files section */
+    "frontend/**/*.json"
```
sarayourfriend, line 107: I'm not sure why we include json files in our tsconfig :thinking:

tsconfig.json
```diff
@@ -1,8 +1,113 @@
...
+    "importsNotUsedAsValues": "error",
+    /* Module Resolution Options */
+    "moduleResolution": "Node",
+    "resolveJsonModule": true,
+    "types": ["@types/jest", "@nuxt/types", "@nuxtjs/i18n", "@pinia/nuxt"],
+    "typeRoots": ["./frontend/typings", "./node_modules/@types"],
+    "paths": {
+      "~/*": ["./frontend/src/*"],
+      "~~/*": ["./frontend/*"],
+      "@openverse/*": ["./packages/*/src"]
```
sarayourfriend, line 32: Directing this to the `src` directory instead of the package root bypasses the `main` and `types` declarations in the package's `package.json`, allowing us to avoid needing to build types for packages that are only used internally.

obulat (approved):
Tried running `just lint eslint`, `just lint types`, `just lint` locally, changing some event names in the `analytics.ts` and in components. It all worked well, and caught the linting errors.
Thank you for explaining the changes in dependencies and the fixes you added for the build size!

zackkrida:
I'll re-review this today 👍

tsconfig.json
zackkrida: Not sure either, but I looks like you added it in this PR:

https://github.com/WordPress/openverse/pull/951

I tried removing it locally and playing around with modifying json files, I didn't notice anything 🤷

zackkrida (approved):
LGTM. Everything I've tested locally seemingly works great.

PR #2487, opened 2023-06-26
Filter single media views for hidden providers

### LINKED ISSUE #2483: Single results are directly visible for removed sources
## Description

Single results for disabled sources in the Openverse API are still directly available. The search controller filters to exclude providers with `filter_content` set to `true`, but our single result controller does not. 

I believe we should. We use this "Hide content" checkbox for providers in the Django admin with the assumption that it excludes content from Openverse in all forms.

## Reproduction

500px is removed from Openverse, yet https://api.openverse.engineering/v1/images/87113165-b45d-4c9c-b20b-34232e2dccfd/ still works.

## Environment

- Device: 
- OS: 
- Browser: 
- Version: 
- Other info:

sarayourfriend:
## Fixes

Fixes #2483 by @zackkrida 

## Description

Adds a filter to exclude works for which the provider has `filter_content=True`. Please see the comment for why we must use a subquery rather than a join. Note that this **does not** increase the number of queries on these endpoints, as proven by the query counting tests in `test_media_views`.

## Testing Instructions

Reproduce the problem in the issue locally by running `just api/up` on `main`. Visit `/v1/images` and confirm you see results from flickr. Click on one of the single result views from the flickr provider and leave the tab open. Go to `/admin` in another tab (deploy/deploy are username/password) and set filter content to true on the flickr provider from the ContentProvider view. In the single result tab, refresh the page and confirm you can still access the result. Confirm that you cannot see the flickr result at `/v1/images` any more. Additionally, to reproduce search returning results for works that do not have an associated ContentProvider entry, delete the flickr view and confirm that you can now see the flickr results at `/v1/images` again.

Now checkout this branch and do the same test. You will need to add the flickr provider back again or run `just api/init` (potentially with `just down -v`) to get it back before running the manual test. Confirm that you cannot access the single result when the provider's filter content is enabled but that you can still access it if you delete the provider.

Look over the new unit tests and confirm they cover the case. Note that I did not opt to test all endpoints even though each endpoint does need to correctly use `get_object` for this approach to work. I could add a `parametrize` for it but wanted to avoid overkill. Just let me know if you think it's worth doing.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -56,7 +56,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
             provider__in=ContentProvider.objects.filter(
                 filter_content=True
             ).values_list("provider_identifier")
-        ).all()
+        )
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -8,6 +8,7 @@ from rest_framework.viewsets import ReadOnlyModelViewSet
 
 from api.controllers import search_controller
 from api.models import ContentProvider
+from api.models.media import AbstractMedia
 from api.serializers.provider_serializers import ProviderSerializer
 from api.utils import image_proxy
 from api.utils.pagination import StandardPagination
@@ -26,7 +27,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
     pagination_class = StandardPagination
 
     # Populate these in the corresponding subclass
-    model_class = None
+    model_class: AbstractMedia = None
     query_serializer_class = None
     default_index = None
 
@@ -42,7 +43,20 @@ class MediaViewSet(ReadOnlyModelViewSet):
             raise ValueError(msg)
 
     def get_queryset(self):
-        return self.model_class.objects.all()
+        # The alternative to a sub-query would be using `extra` to do a join
+        # to the content provider table and filtering `filter_content`. However,
+        # that assumes that a content provider entry exists, which is not necessarily
+        # the case. We often don't add a content provider until after works from
+        # new providers are available in the API, and sometimes not even then.
+        # Search returns results with providers that do not have a ContentProvider
+        # table entry. Therefore, to maintain that assumption, a subquery is the only
+        # workable approach, as Django's `extra` does not provide any facility for
+        # handling null relations on the join.
+        return self.model_class.objects.exclude(
+            provider__in=ContentProvider.objects.filter(
+                filter_content=True
+            ).values_list("provider_identifier")
+        ).all()
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -27,7 +27,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
     pagination_class = StandardPagination
 
     # Populate these in the corresponding subclass
-    model_class: AbstractMedia = None
+    model_class: type[AbstractMedia] = None
     query_serializer_class = None
     default_index = None
 
--- a/api/test/unit/views/test_media_views.py
+++ b/api/test/unit/views/test_media_views.py
@@ -1,7 +1,11 @@
+from datetime import datetime, timezone
 from unittest.mock import MagicMock, patch
+from uuid import uuid4
 
 import pytest
 import pytest_django.asserts
+
+from api.models.models import ContentProvider
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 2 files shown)


AetherUnbound (approved):
Tests make perfect sense, and behaved as expected! Thank you for the great testing instructions! I was able to replicate this locally, all looks good.

krysal (approved):
Perfect ✅

api/api/views/media_views.py
```diff
@@ -26,7 +27,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
     pagination_class = StandardPagination
 
     # Populate these in the corresponding subclass
-    model_class = None
+    model_class: AbstractMedia = None
```
dhruvkb, line 30: This takes a subclass of `AbstractMedia`, not an instance. I think this is the right type.

```suggestion
    model_class: Type[AbstractMedia] = None
```

api/api/views/media_views.py
```diff
@@ -42,7 +43,20 @@ def __init__(self, *args, **kwargs):
...
+        # new providers are available in the API, and sometimes not even then.
+        # Search returns results with providers that do not have a ContentProvider
+        # table entry. Therefore, to maintain that assumption, a subquery is the only
+        # workable approach, as Django's `extra` does not provide any facility for
+        # handling null relations on the join.
+        return self.model_class.objects.exclude(
+            provider__in=ContentProvider.objects.filter(
+                filter_content=True
+            ).values_list("provider_identifier")
+        ).all()
```
dhruvkb, line 59: The `.all()` is not needed here as `.exclude` returns a `QuerySet` too.

api/api/views/media_views.py
```diff
@@ -42,7 +43,20 @@ def __init__(self, *args, **kwargs):
...
+        # the case. We often don't add a content provider until after works from
+        # new providers are available in the API, and sometimes not even then.
+        # Search returns results with providers that do not have a ContentProvider
+        # table entry. Therefore, to maintain that assumption, a subquery is the only
+        # workable approach, as Django's `extra` does not provide any facility for
+        # handling null relations on the join.
+        return self.model_class.objects.exclude(
+            provider__in=ContentProvider.objects.filter(
+                filter_content=True
+            ).values_list("provider_identifier")
```
dhruvkb, line 58: I am not 100% sure of the performance implications but would it be faster if we get this list outside and use `.all` instead of `.exclude` if the list is empty?

dhruvkb (approved):
LGTM, the tests are great and quite comprehensive. I have some small suggestions and one performance-related question.

api/api/views/media_views.py
sarayourfriend: Good question! For it to matter the list would need to be empty. In production it currently is not and I don't necessarily foresee it being empty unless we do some significant work to remove certain providers.

In the interest of checking for it anyway, though, I ran explain locally on both scenarios.

- Here's the explain for when the subquery is empty: https://explain.depesz.com/s/WD8Y
- Here's the explain for when the subquery is not empty: https://explain.depesz.com/s/5Q4O
- Here's the explain with no subquery at all: https://explain.depesz.com/s/FPWt

I ran these locally and I currently have no providers with `content_filter=true`, which is why the first query would have an empty subquery (you can see the filter on the subquery removed 4 rows, i.e., all four content providers I have locally).

Comparing both, you can see that the empty subquery results in a faster query, which makes sense, Postgres has nothing to compare the rows to. However, it is indeed slower than no subquery at all, which also makes sense as it isn't obvious to me that Postgres is _skipping_ the "in" check (though it sure seems like it should be smart enough to do so :shrug:).

Additionally, it is also simpler and faster (it seems) to use a literal list for the right side of "in" vs a subquery, though I'm not sure why. At this point we'd also have to factor in Django ORM overhead. Does the overhead of executing a separate query (meaning the new request to the database) then serializing it into Python objects, then passing those to Django ORM for it to then serialize back into the string format for Postgres (which includes escaping the input) compensate for the better query time performance in Postgres? I'd certainly be curious to know, but to get meaningful data I think we would need to run both in production and compare how they go. It's hard to tell locally what the holistic performance difference would be, especially when we know that in production the case where we can remove the provider check simply doesn't exist because we do have filtered providers in production.

api/api/views/media_views.py
sarayourfriend: Oops, missed this, thank you!

sarayourfriend: Fixed in the last commit, also TIL that the regular `type` builtin works for `Type` in this context as well :tada: Ruff automatically changed that for me :sweat_smile:

zackkrida:
@sarayourfriend given this PR's critical nature I am going to merge it, test on staging, and deploy.

PR #2510, opened 2023-06-29
Fix frontend Sentry configuration

### LINKED ISSUE #2406: Frontend sentry not reporting errors causing 500s
## Description

The production frontend service has a regular but a very small number of 500s, less than 30 per day and often enough, none at all in a given day, typically. However, whatever errors cause these 500s are never reported in Sentry.

## Additional context

Graph of Openverse Nuxt 500 errors daily for the last 4 weeks showing that often there are no errors but sometimes some do exist

As you can see, we do have 500s, yet these never appear in Sentry.

It may be helpful to look through Nuxt logs around the time that 500 happens to see if anything reveals itself as far as reproduction or why Sentry might not be receiving these errors.

cc @WordPress/openverse-frontend I've marked this high because it could mean that Sentry is not working, which means our main method of observability for frontend stability isn't reliable right now.

sarayourfriend:
## Fixes

Fixes #2406 by @sarayourfriend 

## Description

Some of the changes in this PR are not strictly necessary. Specifically, removing `DISABLE_SENTRY` and upgrading the sentry dependency. However, `DISABLE_SENTRY` doesn't appear to be necessary (it was disabling itself _just fine_ without us even trying to in production (:crying_cat_face:)), Sentry disables itself if no DSN exists and this was only used for Playwright, which runs its own Nuxt build, so there shouldn't be any leakage from the base Dockerfile that has the DSN now. Upgrading the Sentry module also isn't necessary but why not, if we're already in here anyway :shrug:

The actual fixes in this PR:

1. Adding `SENTRY_DSN` as a build-time environment variable in the frontend's Dockerfile. This is the most important one as it is what actually configures Sentry to be enabled. The Nuxt Sentry module depends on the fact that it can configure Sentry at build time. There is no option to leave this to a run-time configuration, not if we also want client reporting, unless we did some special shenanigans to inject the environment variable at SSR time. Using public runtime config does _not_ solve this for the _client_, only the server.
2. Adding `config.environment` to the public runtime config for Sentry. This aspect of Sentry's configuration is _not_ bypassed at build time if SENTRY_DSN is not available to the module, but it won't automatically pull `SENTRY_ENVIRONMENT` from the runtime environment _for the client side_, so we need to manually pass that here (so that the client-side Sentry knows if it is staging or production). We add it to `config` because annoyingly, the Sentry module hard-codes `environment` to `NODE_ENV` at build time, meaning it'll always be hard-coded to `production`, even in the staging runtime, and the Sentry SDK will skip checking `SENTRY_ENVIRONMENT` if `environment` is explicitly passed to Sentry's `init` function (which the Nuxt module handles for us).

Basically: Nuxt module hidden complexity and assumptions bite us again :confused:

## Testing Instructions

Confirm that the fix works in theory by running `SENTRY_DSN=<valid-dsn> pnpm exec nuxt build` in the `frontend` directory followed by `SENTRY_ENVIRONMENT=local pnpm exec nuxt start`. You'll get CORS errors because Sentry isn't configured to accept requests from localhost, so it's probably just easiest for us to check this in staging by manually causing an error on the frontend staging after deployment.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
I thought this was working but it is not. Adding a comment to the issue.

zackkrida (approved):
So glad you found a fix here 🥳 Hardcoding the DSN and removing the annoying `disable` setting sounds good.

AetherUnbound (approved):
I was not able to run the testing instructions locally and see the CORS errors you mention, but the changes look good nonetheless!

zackkrida:
Accidentally pushed and reverted an unrelated commit to this branch. I'll merge this now if tests pass.

PR #2538, opened 2023-07-03
Simplify Skip-to-content links

### LINKED ISSUE #2474: Skip to content button hard to access using Firefox
## Description

When using Firefox, the Skip to content button seems to not be visible on the page until a focus event makes it appear. I am not sure why screen readers cannot find it as it works fine in Google Chrome, always visible and accessible.

## Reproduction

1. Enable a screen reader on Windows such as NVDA or JAWS.
2. Open https://openverse.org/ in Firefox.
3. Notice how the screen reader comes to the analytics banner as the first element, not the Skip to content button.
4. Compare with Google Chrome. Notice how opening https://openverse.org/ results in Skip to content being the first element on the page.

## Environment

- Device: Laptop.
- OS: Windows 10 Professional 64-bit
- Browser: Firefox.
- Version: 114.02.
- Other info: NVDA or JAWS screen reader.

## Additional context

I have a feeling this is related to Vue portals. React portals have varying support with assistive tech and I am sure Vue suffers from the same problem.

```
<div force-destroy="true" class="vue-portal-target" data-v-6fb81374=""><button type="button" class="group/button flex appearance-none items-center justify-center rounded-sm no-underline z-50 ms-2 mt-2 focus:fixed focus:absolute button_8qIVo filled-pink_TINkH size-medium_faRVV focus-slim-filled border skipButton_FKScN">Skip to content</button></div>
```

obulat:
## Fixes

Fixes #2474 by @alexstine

## Description

This PR simplifies the "Skip to content" links, making them work without JavaScript.

Previously, the links used a Teleport to move the button to the top of the page. However, the teleports are really finicky, especially in SSR. When I tried migrating to Nuxt 3, teleports were causing client/server mismatch.
Besides, Openverse tries to use progressive enhancing, making  the server-rendered page without Javascript as much as possible work. With this change, "Skip to content" links should work even before the page is "hydrated", and JS starts working.

This PR adds the button at the top of the page in all layouts. This duplication should be removed after the migration to Nuxt 3, when we use `app.vue` as the top-level component.

`VLink.vue` is updated to not use a localized path for anchor links that start with `#` and link to an element on the same page as the link. Some `computed`s were unnecessary and removed.

## Testing Instructions

Open different pages in the app, and navigate to the "Skip-to-content" button using <kbd>Tab</kbd>. It should be the first button on the page. On button click, you should be taken to the "main content" of the page. The screen reader should start reading the main content.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (dismissed):
A much simpler approach indeed and glad that it works without JS being hydrated. I was going to approve because it works, but I noticed one small issue that I'm unsure of whether it should be a blocker (but requesting changes just to be as clear as possible). I also have a question/request for documentation below as well.

## Documentation request

Would it be okay to use a more explicit ID than "content"? That ID is vague enough that I can imagine many would not think twice about removing it. WordPress.org uses `wp--skip-link--target`, for example. That's a fair bit more self-documenting and expressive as to the intention behind that _specific_ ID, considering it needs to be that ID _exactly_. It would be tedious, but even refactoring it out into a constant that itself is documented would be an even better approach. If we did that I'd have no qualms about an innocuous ID like `content`, at least then its purpose would be clearly documented everywhere it is used.

## Potential blocker

With this PR, the skip to content link causes the page to scroll to the `content` anchor. On pages like about, the floating header covers the heading when scrolled "to the anchor": http://localhost:8443/about#content

This page describes the phenomenon and how to fix it (a mix of padding and margin CSS on the heading): https://perishablepress.com/margin-offset-anchor-targets/

Whatever ID we land on, we could add the appropriate CSS to ensure the header never covers it when it is scrolled to as an anchor.

I'm 100% confident that this is a blocker, but for sighted users who use keyboard navigation, this would be a regression.

obulat:
> ## Documentation request
> 
> Would it be okay to use a more explicit ID than "content"? That ID is vague enough that I can imagine many would not think twice about removing it. WordPress.org uses `wp--skip-link--target`, for example. That's a fair bit more self-documenting and expressive as to the intention behind that _specific_ ID, considering it needs to be that ID _exactly_. It would be tedious, but even refactoring it out into a constant that itself is documented would be an even better approach. If we did that I'd have no qualms about an innocuous ID like `content`, at least then its purpose would be clearly documented everywhere it is used.

I used your second suggestion and added a constant for the `content` id that is used for all of the skip-to-content link target elements. 

> ## Potential blocker
> 
> Whatever ID we land on, we could add the appropriate CSS to ensure the header never covers it when it is scrolled to as an anchor.
> 
> I'm 100% confident that this is a blocker, but for sighted users who use keyboard navigation, this would be a regression.

I've added `scroll-mt-20` class to the skip to content target element in `tailwind.css`. This might not work well with the banners, but works with

zackkrida:
This is looking really good. One thing I notice is that the focused area after clicking the ling no longer has a visible focus state as it does is production. Is that intentional?

obulat:
> This is looking really good. One thing I notice is that the focused area after clicking the ling no longer has a visible focus state as it does is production. Is that intentional?

I don't know whether the correct behavior is to have a visible focus state or not, but I added it in the latest commit. So, now it matches the prod.

sarayourfriend:
I'll review this again tomorrow, @obulat

sarayourfriend (approved):
LGTM. The changes to prevent the header overlap work when there are no banners. With one banner at least part of the word is still visible, so while not ideal, there's the visual queue to scroll up a little to see the heading. Ideally we'd be able to correct the amount of margin for the number of banners present. A CSS custom property that gets updated by the banner components when they render is the only way I can think to do that, but it would need to be a manual approach rather than using the tailwind utilities for this.

In any case, not a blocker, LGTM. If you decide not to address the banner issue in this PR, can you open an issue or let me know and I'll open one?

zackkrida (approved):
LGTM, ➕ to Sara's suggestion concerning the banner handling.

PR #2544, opened 2023-07-04
Set UA string for Openverse API requests from the frontend

zackkrida:
## Description

Forgot I wrote this earlier in June. Adds a UA string identifying the frontend to the API in all requests. Useful for monitoring API traffic from the frontend over time.

## Testing Instructions

Run locally, make a client side search and check for the user agent set in the request in devtools.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/data/api-service.ts
```diff
@@ -90,10 +92,15 @@ export const createApiService = ({
   const axiosParams: AxiosRequestConfig = {
     baseURL: isVersioned ? `${baseUrl}v1/` : baseUrl,
     timeout: DEFAULT_REQUEST_TIMEOUT,
+    headers: { "User-Agent": userAgent },
   }
+
   if (accessToken) {
-    axiosParams.headers = { Authorization: `Bearer ${accessToken}` }
+    if (axiosParams.headers) {
```
dhruvkb, line 99: Why the `if`? Isn't `headers` hardcoded to be present on L95?

frontend/src/utils/user-agent.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * A user agent string by which Openverse.org identifies itself to services
+ */
+const userAgent =
+  "Openverse/0.1 (https://openverse.org; openverse@wordpress.org)"
```
dhruvkb, line 5: Nit, potentially non-relevant: this version is rather arbitrary.

dhruvkb (approved):
LGTM!

frontend/src/utils/user-agent.js
zackkrida: Agreed, I copied this from our usage somewhere else, but now I don't remember.  I'll try to find it and reassess an alternative

obulat: Nit: should this file go to the `/constants` folder instead?

obulat (commented):
LGTM!

frontend/src/data/api-service.ts
sarayourfriend: The if is pretty confusing. When I first read it, it made me think that somehow `headers` could be undefined, even though it's _always_ on the axios params above.

There are two ways of doing this. One is purely in the type space and the other would be a slight change in the code to remove the if:

1. Change `const axiosParams: AxiosRequestConfig` to `const axiosParams: Required<Pick<AxiosRequestConfig, "headers">> & AxiosRequestConfig`. That tells TypeScript to remove `undefined` from the `headers` property, removing the need to check the `if` below. It's a bit wordy, but removes the need for unnecessary code. It also doesn't introduce the risk of a cast because if `headers` is _not_ defined on the right-hand side of the assignment expression, the type check will fail because the object won't fulfil the requirements of the type (`headers` being a headers object).
2. Cast headers below in the assignment to remove undefined: ``(axiosParams.headers as Exclude<typeof axiosParams["headers"], undefined>)["Authorization"] = `Bearer ${accessToken}` ``. To me this is objectively worse because if `headers` was removed from the `axiosParams` assignment above, there would be no type error.

I have a strong preference for the first option. To reduce the wordiness of it, you could create a new type:

```ts
/**
 * Openverse Axios request config with adjusted types for our use-case.
 */
type OpenverseAxiosRequestConfig = Required<Pick<AxiosRequestConfig, "headers">> & AxiosRequestConfig
```

Then use `OpenverseAxiosRequestConfig` in place of `AxiosRequestConfig` in the `axiosParams` assignment above.

frontend/src/utils/user-agent.js
AetherUnbound: We use `0.1` for the catalog as well it seems: 

https://github.com/WordPress/openverse/blob/e1f48221938b4cc9838a85578b47bb4f6bfa33f6/catalog/dags/common/loader/provider_details.py#L121

AetherUnbound (commented):
LGTM! I'm not seeing the user agent changed in the request headers when making a search locally though 🤔 it's the `/v1/images?q=<term>` request that the headers should be set, right?

sarayourfriend:
@AetherUnbound IIRC Chrome does not allow user agent to be changed. Firefox also might not allow it. In that case, this change would only be evidenced for server side requests, not browser based ones. For Openverse.org requests from the frontend, the `referrer` header is a reliable identifier of source rather than the UA string.

AetherUnbound:
Oh I see! Does that mean that `Referrer` here should be the value that was changed? Or is it as-expected?
image

zackkrida:
@AetherUnbound Referrer is set by the browser for client-side requests, it's the url of whatever the site is, in this case it'll be https://openverse.org in production.

AetherUnbound (approved):
Thanks for the clarification!

obulat:
> @AetherUnbound IIRC Chrome does not allow user agent to be changed. Firefox also might not allow it. In that case, this change would only be evidenced for server side requests, not browser based ones. For Openverse.org requests from the frontend, the `referrer` header is a reliable identifier of source rather than the UA string.

Is there a way to suppress the warnings? On staging, we get `instrument.js:289 Refused to set unsafe header "User-Agent"` on every client request now.

zackkrida:
Ah, we should probably only set the UA server-side then. I'll make an issue today.

PR #2570, opened 2023-07-06
Make batched_update DAG resistent to mid-update failures

### LINKED ISSUE #2507: Make batched_update DAG resilient to task failures
## Current Situation

We plan to use the `batched_update` DAG to run popularity refreshes for each provider. In the case of very large providers like Flickr, the `update_batches` task may be very long-running (more than a week long). This means that it is highly likely that we will need to run catalog deployments while the task is still in progress.

If the catalog is deployed while the task is running, the task will fail and then retry (assuming it has remaining retries) once Airflow is back up. **Currently, this means that the `update_batches` task will resume doing its update from the BEGINNING of the rows_to_update table.** Thus while it looks like the DAG is still running fine, the batched update has actually started over from the beginning and is re-updating rows that were already updated.

We need a way to ensure that when the task retries, it picks up where it left off instead of starting over.

Note: A developer **can** do this **manually**, by failing the DAG, noting in the logs where it left off (what was the last `row_id` processed), and then manually triggering a new DagRun with the same query parameters but with `batch_start` set to this row_id, and `resume_update` enabled to prevent the update table from being recreated. However this is a tedious manual process that may need to be done multiple times, for multiple providers, and which could easily be forgotten.

## Suggested Improvement

There are a couple ways we could implement this:

### Deleting records from the `rows_to_update` table as we go

The simplest fix would be to delete records from the `rows_to_update` table after their corresponding records are updated on the media table. Then when the task fails and is retried, anything that was already updated will be missing from the table.

Cons:
* Adds an extra query to each batch update. It should be a very quick query, but adding any overhead to this update is undesirable.
* We would need to update the `batch_start` logic. Currently we assume that unless a custom start is provided, we start at `row_id = 0`, but that would no longer be the case.

### Store the last processed row_id somewhere

We could persist a variable keeping track of the last row_id processed, which we update each time a batch is processed. Because this must persist through a task failure and retry, we'd need to store this in the database. When the `update_batches` task runs, it chooses where to begin the update in this order:
* `batch_start` if supplied is always preferred
* `last_processed_row_id` if it exists
* defaults to 0, beginning from the start of the update table.

The simplest way might be to use an Airflow variable using the `query_id` to uniquely name it (`last_processed_row_id_{{query_id}}`), which we can easily set in the task. The only hitch is cleaning up/deleting the variable when the DagRun finishes, but I believe we can do that with the cli and a bash operator ([example](https://stackoverflow.com/a/76336288)).

### Dynamically generated batch update tasks, instead of a single update task

The current approach has a single `update_batches` task which iteratively runs the updates. We could instead use dynamic task mapping to generate a separate Airflow task per batch. Each task succeeds when it is done, so only the currently running batches will be retried if they are halted during a deployment.

This was not implemented originally because large providers may require huge numbers of batches (50k batches for Flickr, for example). Airflow by default [only allows a maximum of 1024 dynamic tasks per DAG](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#max-map-length) at a time. As this number gets higher, the scheduler apparently becomes clogged and [Airflow gets extremely slow](https://www.reddit.com/r/dataengineering/comments/zrv612/is_there_an_alternative_for_airflow_for_running/).

I did have some ideas for handling this, but it's not really how Airflow is intended to be used. For example, we could have the batched_update DAG schedule the first 1024 batches and then, if not all batches were processed, use the TriggerDagRunOperator to [trigger another DagRun of itself](https://wang-kuanchih.medium.com/apache-airflow-self-trigger-looping-back-in-dag-6bef69818fe8) using the existing `batch_start` and `resume_update` params to get the next batches. 

Cons:
* Even with my solution to the task-number problem, this is introducing a lot of Airflow overhead and a big strain on the scheduler -- especially if multiple batched updates are running at the same time. This may considerably slow down an already slow process.
* Dynamically mapped tasks are meant to run concurrently. We can use `max_active_tasks` to restrict how many run at one time, but critically there is no way to make the entire DAG fail as soon as one fails. If there's a problem with the SQL, we could expect a bombardment of 1024 task failures and Slack notifications. (It is possible to prevent the Slack notifications, but this is still unfortunate).

## Benefit

Allows very long-running batched updates to continue through deployments and to resume from failure without manual intervention.

## Additional context

#2331

stacimc:
## Fixes

Fixes #2507 by @stacimc 

## Description

More context in #2507, but briefly:

Because the `batched_update` DAG is going to be used to perform potentially very long-running (multiple weeks long) updates, it needs to be able to handle the update task failing and retrying due to Catalog deployments. In the current implementation, when the update task retries it would resume _from the beginning_.

This PR takes the following approach:
* In each DagRun, we create an Airflow variable suffixed with the `query_id` that we use to keep track of the progress of the update, updating it after each batch.
* When the `update_batches` task runs, it checks to see if this variable exists and if so, resumes the update from that point. Otherwise it starts at the beginning.
* An additional clean up task is added at the end to delete the Airflow variable.

Screen Shot 2023-07-05 at 4 43 23 PM

Now, if the `update_batches` task fails and retries (or is cleared manually), it will resume from the point it left off.

Notably, I also removed the `batch_start` DagRun conf option that I had previously, as it should no longer be necessary to manually choose the starting point. **If a DAG fails, you can manually find and edit this Airflow variable before retrying the DAG _if you want to_, but it should never be necessary.**

## Additional context

I chose this approach over deleting rows from the temp table as we go because it reduces the amount of error handling needed, but more importantly because I am concerned about adding an additional DELETE query to each batch when performance is already a major concern. Details about the cons to the dynamic task mapping alternative are available in the issue #2507.

I did _not_ remove the `resume_update` DagRun conf option in this PR, because this is still useful in the following scenario:
* User starts a DagRun which builds a very large temp table
* The DagRun builds the temp table successfully
* The `update_batches` task fails due to a syntax error in the user-provided `update_query`. Because this is a part of the DagRun conf, you cannot simply clear the task. You have to start a new DagRun with an updated conf, but you don't want to have to rebuild the temp table.

## Testing Instructions

Setup:
Run `just init` to make sure you're starting out with our sample data in your local catalog db.

**Test a 'regular' flow**
Trigger the `batched_update` DAG with the following conf:
```
{
	"query_id": "my_query",
  	"table_name": "audio",
  	"select_query": "WHERE provider='wikimedia_audio'",
  	"update_query": "SET duration=0",
	"batch_size": 100,
	"dry_run": false
}
```
All tasks should pass (except for `notify_failure`, which is skipped).

* Run `just catalog/pgcli` to open a pgcli session
* Run `SELECT COUNT(*) FROM audio WHERE provider='wikimedia_audio' AND duration=0` to verify that you have 3992 such records after running the update

**Test clearing a DagRun**
For this we need to simulate a DAG failure. I did this by modifying the `update_batches` task to raise an Error on the 3rd batch, by adding the following code on L118[here](https://github.com/WordPress/openverse/pull/2570/files#diff-39e2cb653ee54d2523d84930ee904d9cd865f64ac4d85069aeb6da2040fe6203R118):

```
if batch_start == 300:
    raise Exception("Oops")
```

Now trigger a DagRun with this conf:

```
{
	"query_id": "my_query",
  	"table_name": "audio",
  	"select_query": "WHERE provider='wikimedia_audio'",
  	"update_query": "SET duration=100",
	"batch_size": 100,
	"dry_run": false
}
```

The `update_batches` task should fail and go `up_for_retry`. In pgcli, run `SELECT COUNT(*) FROM audio WHERE provider='wikimedia_audio' AND duration=100` and you should be able to see that 300 records were updated.

You should also be able to go to Admin > Variables in the UI and see an Airflow variable named `batch_start_my_query` which is set to 300.

Now remove the code you added to raise an Exception and clear the task. This time it should succeed. Verify in the logs that it resumed starting at row 300 and that it only updated 3692 records. (Search for the logs: `Starting at 300` and finally, `Updated 3692 rows. 0 remaining.`.)

Run the SELECT COUNT(*) command again in pgcli to verify that all records were updated.

Go to Admin > Variables and verify the Airflow variable is gone.

**Test a new DagRun using `resume_update`**

Trigger a new DagRun with this conf (note the typo in `update_query`):

```
{
	"query_id": "my_query",
  	"table_name": "audio",
  	"select_query": "WHERE provider='wikimedia_audio'",
  	"update_query": "SET durationn=500",
	"batch_size": 100,
	"dry_run": false
}
```

When the `update_batches` task fails and goes up for retry, manually mark the task as `failed` (to avoid having to wait for the retries). The whole DAG should fail. The temp table should not be dropped.

Now trigger a new DagRun with this updated conf, fixing the typo and enabling `resume_update`:

```
{
	"query_id": "my_query",
  	"table_name": "audio",
  	"select_query": "WHERE provider='wikimedia_audio'",
  	"update_query": "SET duration=500",
	"batch_size": 100,
	"dry_run": false,
        "resume_update": true
}
```

Verify that the `select_rows_to_update` and `create_index` tasks are skipped. `update_batches` should succeed, and the temp table and airflow variable should be dropped.

In pgcli, run `SELECT COUNT(*) FROM audio WHERE provider='wikimedia_audio' AND duration=500` to verify that all 3992 records were updated.
## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/database/batched_update/batched_update.py
+++ b/catalog/dags/database/batched_update/batched_update.py
@@ -120,6 +120,7 @@ def update_batches(
             dry_run=dry_run,
             sql_template=constants.UPDATE_BATCH_QUERY,
             query_id=query_id,
+            # Only log the query the first time, so as not to spam the logs
             log_sql=batch_start == initial_batch_start,
             postgres_conn_id=postgres_conn_id,
             task=task,
--- a/catalog/dags/database/batched_update/batched_update.py
+++ b/catalog/dags/database/batched_update/batched_update.py
@@ -2,6 +2,7 @@ import logging
 from datetime import timedelta
 
 from airflow.decorators import task
+from airflow.models import Variable
 from airflow.models.abstractoperator import AbstractOperator
 
 from common import slack
@@ -36,21 +37,23 @@ def resume_update(
 
 
 @task
-def get_expected_update_count(query_id: str, batch_start: int | None, dry_run: bool):
-    """
-    Get the number of records left to update, when resuming an update
-    on an existing temp table.
-    """
-    total_count = run_sql.function(
+def get_expected_update_count(
+    query_id: str, total_row_count: int | None, dry_run: bool
+):
+    """Get the number of records in the temp table."""
+    # If we created the temp table as part of this DagRun, the total row count is
+    # passed in through XComs and is simply forwarded.
+    if total_row_count is not None:
+        return total_row_count
+
+    # If `resume_update` is True, we skip the table creation task and the count will
+    # not be available in XComs. We instead explicitly query the table for the count.
+    return run_sql.function(
         dry_run=dry_run,
-        sql_template=constants.SELECT_TEMP_TABLE_QUERY,
+        sql_template=constants.SELECT_TEMP_TABLE_COUNT_QUERY,
         query_id=query_id,
         handler=_single_value,
     )
-
-    if batch_start:
-        total_count -= batch_start
-    return max(total_count, 0)
 
 
 @task
@@ -89,22 +92,27 @@ def run_sql(
 
 @task
 def update_batches(
-    expected_row_count: int,
+    total_row_count: int,
     batch_size: int,
     dry_run: bool,
     table_name: str,
     query_id: str,
     update_query: str,
     update_timeout: int,
-    batch_start: int = 0,
+    batch_start_var: str,
     postgres_conn_id: str = POSTGRES_CONN_ID,
     task: AbstractOperator = None,
     **kwargs,
 ):
-    updated_count = 0
-    if batch_start is None:
-        batch_start = 0
+    # Progress is tracked in an Airflow variable. When the task run starts, we resume
+    # from the start point set by this variable (defaulted to 0). This prevents the
+    # task from starting over at the beginning on retries.
+    initial_batch_start = Variable.get(batch_start_var, 0, deserialize_json=True)
+    expected_row_count = max(total_row_count - initial_batch_start, 0)
+    logger.info(f"Starting at {initial_batch_start}")
 
+    updated_count = 0
+    batch_start = initial_batch_start
     while batch_start <= expected_row_count:
         batch_end = batch_start + batch_size
 
@@ -113,8 +121,7 @@ def update_batches(
             dry_run=dry_run,
             sql_template=constants.UPDATE_BATCH_QUERY,
             query_id=query_id,
-            # Only log the query the first time, so as not to spam the logs
-            log_sql=batch_start == 1,
+            log_sql=batch_start == initial_batch_start,
             postgres_conn_id=postgres_conn_id,
             task=task,
             timeout=update_timeout,
@@ -126,6 +133,9 @@ def update_batches(
 
         updated_count += count
         batch_start = batch_end
+
+        # Update the Airflow variable to the next value of batch_start.
+        Variable.set(batch_start_var, batch_end)
         logger.info(
             f"Updated {updated_count} rows. {expected_row_count - updated_count}"
             " remaining."
--- a/catalog/dags/database/batched_update/batched_update.py
+++ b/catalog/dags/database/batched_update/batched_update.py
@@ -144,6 +144,11 @@ def update_batches(
 
 
 @task
+def drop_temp_airflow_variable(airflow_var: str):
+    Variable.delete(airflow_var)
+
+
+@task
 def notify_slack(text: str, dry_run: bool) -> None:
     if not dry_run:
         slack.send_message(
--- a/catalog/dags/database/batched_update/batched_update.py
+++ b/catalog/dags/database/batched_update/batched_update.py
@@ -108,12 +108,11 @@ def update_batches(
     # from the start point set by this variable (defaulted to 0). This prevents the
     # task from starting over at the beginning on retries.
     initial_batch_start = Variable.get(batch_start_var, 0, deserialize_json=True)
-    expected_row_count = max(total_row_count - initial_batch_start, 0)
     logger.info(f"Starting at {initial_batch_start}")
 
     updated_count = 0
     batch_start = initial_batch_start
-    while batch_start <= expected_row_count:
+    while batch_start <= total_row_count:
         batch_end = batch_start + batch_size
 
         logger.info(f"Updating rows with id {batch_start} through {batch_end}.")
@@ -136,10 +135,9 @@ def update_batches(
 
         # Update the Airflow variable to the next value of batch_start.
         Variable.set(batch_start_var, batch_end)
-        logger.info(
-            f"Updated {updated_count} rows. {expected_row_count - updated_count}"
-            " remaining."
-        )
+
+        remaining_count = max(total_row_count - initial_batch_start - updated_count, 0)
+        logger.info(f"Updated {updated_count} rows. {remaining_count} remaining.")
 
     return updated_count
 
--- a/catalog/dags/database/batched_update/batched_update_dag.py
+++ b/catalog/dags/database/batched_update/batched_update_dag.py
@@ -31,8 +31,6 @@ Optional params:
 * batch_size: int number of records to process in each batch. By default, 10_000
 * update_timeout: int number of seconds to run an individual batch update before timing
                   out. By default, 3600 (or one hour)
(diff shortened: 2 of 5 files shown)


krysal (approved):
Looks great! The tests returned the expected results 👌

catalog/dags/database/batched_update/batched_update.py
```diff
@@ -114,7 +121,7 @@ def update_batches(
             sql_template=constants.UPDATE_BATCH_QUERY,
             query_id=query_id,
             # Only log the query the first time, so as not to spam the logs
-            log_sql=batch_start == 1,
+            log_sql=batch_start == initial_batch_start,
```
AetherUnbound, line 124: It looks like this was `1` before but with the new default it will be `0`. This seems like the correct value now (and maybe was the wrong value before?) but I just wanted to double check!

catalog/dags/database/batched_update/batched_update_dag.py
```diff
@@ -181,6 +162,9 @@
     },
 )
 def batched_update():
+    # Unique Airflow variable name for tracking the batch_start for this query
+    BATCH_START_VAR = "batch_start_{{ params.query_id }}"
```
AetherUnbound, line 166: Would it be possible to have this match the DAG ID so it's potentially a bit easier to identify? Maybe:
```suggestion
    BATCH_START_VAR = "batched_update_start_{{ params.query_id }}"
```

catalog/dags/database/batched_update/batched_update_dag.py
```diff
@@ -253,6 +237,12 @@ def batched_update():
         query_id="{{ params.query_id }}",
     )
 
+    # Clean up the variable we used for tracking the last row
+    drop_variable = BashOperator(
+        task_id="drop_temp_airflow_variables",
+        bash_command=f"airflow variables delete {BATCH_START_VAR}",
+    )
```
AetherUnbound, line 244: Would it make sense to potentially use [`Variable.delete`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/variable/index.html#airflow.models.variable.Variable.delete) here rather than a `BashOperator`? I guess I worry about environment/access issues that might appear on the CLI which we know wouldn't occur within Python 😅 No strong opinion here, just curious!

AetherUnbound (approved):
This is great, the use of the Variable and automatic cleanup is excellent! I have a few comments/questions (and echo what Krystle mentioned) but nothing to block a merge.

catalog/dags/database/batched_update/batched_update.py
stacimc: Correct! I had a very small commit ready to fix it to 0, which ended up being superseded by this PR.

catalog/dags/database/batched_update/batched_update_dag.py
stacimc: Done!

stacimc:
There are a lot of formatting changes to `DAGs.md` after rebasing and regenerating the dag docs 🤔 Any concerns with that @AetherUnbound?

AetherUnbound:
> There are a lot of formatting changes to `DAGs.md` after rebasing and regenerating the dag docs 🤔  Any concerns with that @AetherUnbound?

Yea that definitely is O_o it looks like linting is no longer being run on the DAGs.md file after it's generated. I'll try to look into why that's happening.

catalog/dags/database/batched_update/batched_update_dag.py
```diff
@@ -51,23 +49,18 @@
...
+off. Manually managing this Airflow variable should not be necessary.
+
+It is also possible to start an entirely new DagRun using an existing temp table,
+by setting the `resume_update` param to True. With this option enabled, the
+DAG will skip creating the temp table and instead attempt to run an update with an
+existing temp table matching the `query_id`. This option should only be used when
+the DagRun configuration needs to be changed after the table was already created:
+for example, if there was a problem with the `update_query` which caused DAG
+failures during the `update_batches` step.
 ```
```
AetherUnbound, line 64: Oh! @stacimc it looks like this erroneous triple backtick is throwing the linting off because it thinks it's in a code block 😅

stacimc: Ah! Good find, that explains a lot 😄

Issue #2553, opened 2023-07-05
Fix types in the single result pages


obulat:
## Problem

The single result pages ([`frontend/src/pages/image/_id/index.vue`](https://github.com/WordPress/openverse/blob/08bb0317e1110694ca4d51058bebbc1dafb4fc13/frontend/src/pages/image/_id/index.vue), [`frontend/src/pages/audio/_id/index.vue`](https://github.com/WordPress/openverse/blob/08bb0317e1110694ca4d51058bebbc1dafb4fc13/frontend/src/pages/audio/_id/index.vue)) and the single image report page [`frontend/src/pages/image/_id/report.vue`](https://github.com/WordPress/openverse/blob/08bb0317e1110694ca4d51058bebbc1dafb4fc13/frontend/src/pages/image/_id/report.vue)) fail TypeScript checks. The `image` or `audio` can be `null` when loading, but the components expect it always to be non-null.

`head()` references properties on `this.audio` and `this.image` respectively.

https://github.com/WordPress/openverse/blob/ba2081e6cd8b042da3aaf985e26653839bc001d2/frontend/src/pages/audio/_id/index.vue#L112

https://github.com/WordPress/openverse/blob/ba2081e6cd8b042da3aaf985e26653839bc001d2/frontend/src/pages/image/_id/index.vue#L276

This field can be null when the request to get the audio or image has failed, as it recently did due to 429 errors. In such cases, the following error is raised:

```
TypeError: Cannot read properties of null
```

## Description

We should add these pages to `tsconfig`, and refactor them so that the child components can handle `null` in `image`/`audio`. We could add a Skeleton loader to the image and audio track to show when the main media item is being loaded.

## Alternatives

We could also set the image/audio in `asyncData` which loads the image before navigating, blocking the page load. This will have a negative effect in the perceived page loading speed. The page will be in loading state until the data has been loaded.

## Additional information

There are several Sentry errors that will be solved when this is implemented:
https://openverse.sentry.io/share/issue/2e7b05dae0f043c1b0bb519c8f7ebc4b/
https://openverse.sentry.io/share/issue/8c877cb6654d488a82269423b3329c30/
https://openverse.sentry.io/share/issue/bace3d16306e43e88d590608b2f87d2b/
https://openverse.sentry.io/share/issue/317debf6f9cd4c25971efdde9c2ad14d/
https://openverse.sentry.io/share/issue/c8556bbba1c34124a2d1267537210046/
[...]

PR #2584, opened 2023-07-07
Fix types in the templates

obulat:
~~Note: This PR is based on #2583, and should be reviewed after it.~~ [^1]

## Fixes

Related to #2553 by @obulat

## Description

This PR adds conditional rendering to the single result page templates to account for the fact that the main media (`image`/`audio`) can be `null`.

This PR was split from the bigger fix to simplify review.

## Testing Instructions

[^1]: I realized that this PR doesn't really need the changes from 2583, so I rebased it on main.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/pages/audio/_id/index.vue
+++ b/frontend/src/pages/audio/_id/index.vue
@@ -1,27 +1,32 @@
 <template>
   <VSkipToContentContainer as="main">
     <div v-if="backToSearchPath" class="w-full px-2 py-2 md:px-6">
-      <VBackToSearchResultsLink :id="audio.id" :href="backToSearchPath" />
-    </div>
-
-    <VAudioTrack
-      layout="full"
-      :audio="audio"
-      class="main-track"
-      @interacted="sendAudioEvent($event, 'AudioDetailPage')"
-    />
-    <div
-      class="mx-auto mt-10 flex flex-col gap-10 px-6 lg:mt-16 lg:max-w-5xl lg:gap-16"
-    >
-      <VMediaReuse data-testid="audio-attribution" :media="audio" />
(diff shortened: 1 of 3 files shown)


zackkrida (approved):
👍 When we switch to Nuxt 3 we should consider using error boundaries for client-side errors like these:

https://nuxt.com/docs/api/components/nuxt-error-boundary

Currently, if a frontend request to a single media result from the API fails, the page will render blank like this:

CleanShot 2023-07-07 at 10 48 36@2x

I do wonder if we should be throwing a 404 error or displaying some "{media} not found or temporarily unavailable" message here. Perhaps we could show a specific message if the API 429s. All of those would be follow-up work though and require design from @fcoveram.

obulat:
> 👍 When we switch to Nuxt 3 we should consider using error boundaries for client-side errors like these:

Nice!

> Currently, if a frontend request to a single media result from the API fails, the page will render blank like this:
> 
> CleanShot 2023-07-07 at 10 48 36@2x
> 
> I do wonder if we should be throwing a 404 error or displaying some "{media} not found or temporarily unavailable" message here. Perhaps we could show a specific message if the API 429s. All of those would be follow-up work though and require design from @fcoveram.

This wasn't the intended behavior 😨 I did try to redirect to the error page (our generic FourOhFour) if the request fails.
I'm trying to fix it in the follow-up PR.

I had the same idea for 429, here's the issue: https://github.com/WordPress/openverse/issues/2586

dhruvkb (approved):
The changes make sense, LGTM!

PR #2585, opened 2023-07-07
Fix fetching and types on single result pages

### LINKED ISSUE #2553: Fix types in the single result pages

obulat:
~~This PR is based on #2584 and should be reviewed after it~~ 2584 is already merged

## Fixes

Fixes #2553 by @obulat

## Description

This is a large PR that fixes fetching logic for single result pages.

The changes here are based on the template changes added in #2584.

### Refactor fetching
The `singleResult` middleware sets the media `id` before navigating to the route. If the media item is available in another media store (so, if it was a part of the search results, the related media list or it is the global audio), `middleware` also sets the media as the `singleResultStore`'s `mainMedia`.

_Note: Our Nuxt app has two rendering modes. SSR: **Server** renders the single result page if you open the single result page directly, by going to the direct URL like https://openverse.org/image/f9384235-b72e-4f1e-9b05-e1b116262a29.  CSR: **Client** renders the page if you navigate to the single result page from another page, such as when you click on a search result, or one of the related media items, or even on the global audio track._

On the server, we fetch the media item and serve the fully-rendered single result page if the fetching is successful, or an error page if there's a fetching error.

On client navigation, we fetch the media item asynchronously, in the `useFetch` composable in the single result page component. Most of the time we will be only fetching the related media because the media data is already set in the `middleware`. The only exception to this is when you click on the homepage images: those do not set the media data, and we have to fetch both the media and the related media data to render the page.

### Single result store
I added getters for `audio`/`image` that return media detail of the correct type.
The `fetch` method now returns the media of the correct type or throws an error. ~~The error that we get when fetching is sent to Sentry as is. After sending the error to Sentry, we save the error to the store to use it in the error pages (which should be updated to show different messages on errors other than 404).~~ This pattern caused sending the errors to Sentry twice: in the store's `fetchMedia`'s `catch` clause and after the error is re-thrown. So, I refactored it: The error is caught in the `fetchMedia` method in the store. We extract the statusCode and the message to set as the `fetchState.fetchingError` value, and then re-throw the error. This error is caught in the `middleware` (on SSR) or in the component's `useFetch` method, and _then_ sent to Sentry. After that, we redirect to the Nuxt error page.

While working on #2618 I realized that we can avoid capturing error in Sentry in a better way. The store was refactored to return either the media item or null. Null is returned if there is an error, which is caught in the `fetch` function, which updates the store `fetchingError` and sends the error to Sentry. Then, the components handle a return of `null` as the error to show the error page.

### Templates
In addition to changes in #2584, this PR adds a skeleton loader until the item's thumbnail is loaded.
The related images component had both a Loading icon and a Loading skeleton when fetching - I removed the loading icon.
The server-rendered image report page is also simplified. It only needs to be rendered on the server and does not need to be navigated to client-side, and it should not be rendered if there is an error loading the `image`, so I used the `asyncData` there to ensure that `image` is always non-null.

## Testing Instructions

Test loading the image/audio single result pages: the Server-side rendered version (loading directly) and the client-side rendered version (navigating from the search results, from the related media section, global audio or the homepage image links).
If you throttle your browser, you should see the loading skeleton when you click on the homepage images because we need to fetch the image data. In other cases of client-side navigation, we already have the data and the thumbnail, so the loading should be instant.
You can also update the default timeout set in the `api-service` to cause timeout errors.
Another way I used to test errors on the single result page is to change the search result link's href using devtools, making it invalid. Then, if you click this link you should be redirected to the 404 page. If you open the Vue devtools on this page, you should also see the error status code and the error message as the error page prop. 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/pages/audio/_id/index.vue
```diff
@@ -67,19 +73,29 @@ export default defineComponent({
...
+    const audio = ref<AudioDetail | null>(singleResultStore.audio)
+
+    useFetch(async () => {
+      const audioId = route.value.params.id
+      await singleResultStore.fetch(AUDIO, audioId)
+
+      const fetchedAudio = singleResultStore.audio
+
+      if (!fetchedAudio) {
+        useContext().error({ statusCode: 404 })
```
sarayourfriend, line 87: Is it safe to call `useContext()` inside the `useFetch` callback? Just wondering if it could cause the same ref-based memory leak we had in https://github.com/WordPress/openverse-frontend/issues/1722

frontend/src/stores/media/single-result.ts
```diff
@@ -66,99 +84,136 @@ export const useSingleResultStore = defineStore("single-result", {
...
-            : `Request failed with status ${
-                error.response?.status ?? "unknown"
-              }`
-      } else {
-        errorMessage =
-          error instanceof Error ? error.message : "Oops! Something went wrong"
+    async fetchMediaItem<T extends SupportedMediaType>(
+      type: SupportedMediaType,
+      id: string
+    ): Promise<DetailFromMediaType<typeof type>> {
```
sarayourfriend, line 201: The generic `T` type can be used to remove the need to cast at call sites:

```suggestion
    async fetchMediaItem<MediaType extends SupportedMediaType>(
      type: MediaType,
      id: string
    ): Promise<DetailFromMediaType<MediaType>> {
```

Changing it from `T` to `MediaType` helps document what the type is.

The reason the cast was necessary at call sites is because `typeof type` resolved to broad `SupportedMediaType` rather than the specific type passed to `T` because `type` was typed as `SupportedMeidaType` rather than the generic type

frontend/src/stores/media/single-result.ts
```diff
@@ -66,99 +84,136 @@ export const useSingleResultStore = defineStore("single-result", {
...
+      this._updateFetchState("start")
+      const accessToken = this.$nuxt.$openverseApiToken
+      const service = initServices[type](accessToken)
+      const item = this._addProviderName(await service.getMediaDetail(id))
+
+      // Fake ~50% of results as mature. This leaves actual mature results unchanged.
+      const featureFlagStore = useFeatureFlagStore()
+      if (featureFlagStore.isOn("fake_sensitive")) {
+        markFakeSensitive(item)
       }
```
sarayourfriend, line 211: Should this PR wait for #2613? Otherwise the cleanup from that PR would need to be redone whereas here it can just be left out. I guess this PR could just be updated to reflect the changes in #2613 by removing the handling here entirely?

I might be getting my wires crossed between these PRs though, apologies if there's nothing overlapping with them.

frontend/src/types/fetch-state.ts
```diff
@@ -1,6 +1,21 @@
...
+
+/**
+ * The fetching error has the props required for Nuxt's
+ * error page.
+ * TODO: replace FetchState with FetchStateWithNuxtError
+ * in the provider, main media and related media stores.
+ */
+export interface FetchStateWithNuxtError extends BaseFetchState {
+  fetchingError: null | NuxtError
+}
```
sarayourfriend, line 21: To avoid the verbose name, we could make `FetchState` generic:

```suggestion
export interface FetchState<ErrorType = string> extends BaseFetchState {
  fetchingError: null | ErrorType
}
```

Then everywhere using `FetchStateWithNuxtError` can just use `FetchState<NuxtError>`. Once all usages have been converted to pass `NuxtError` we could change the default to `NuxtError` instead of `string` and remove the explicit generic parameters _or_ just remove the generic capability altogether.

frontend/src/types/media.ts
```diff
@@ -103,10 +103,17 @@ export interface ImageDimensions {
...
+    return !!media && media.frontendMediaType === "image"
   },
 }
+
+export const isMediaDetail = <T extends SupportedMediaType>(
+  media: AudioDetail | ImageDetail | null,
+  mediaType: T
+): media is DetailFromMediaType<T> => {
+  return !!media && media.frontendMediaType === mediaType
+}
```
sarayourfriend, line 119: Why do we need to use `AudioDetail | ImageDetail | null` rather than `Media | null`? `Media` inherently covers `AudioDetail` and `ImageDetail`.

Also, does the `isMediaDetail` function duplicate the `isDetail.*` functions? If we need the separated ones, the `isDetail` functions can be implemented with `isMediaDetail`:

```typescript
const isDetail = {
  audio: (media: Media | null) => isMediaDetail(media, "audio"),
  image: (media: Media | null) => isMediaDetail(media, "image"),
}
```

frontend/test/unit/specs/pages/single-image-result-page.spec.js
```diff
@@ -12,48 +13,48 @@ jest.mock("~/composables/use-analytics", () => ({
...
-          id,
-          url: "http://example.com/image.jpg",
-          width: 100,
-          height: 100,
-          providerName: "provider",
-        }
-      }
-    )
+    const screen = render(SingleImageResultPage, options, storeSetupCb)
+    screen.debug()
```
sarayourfriend, line 39: Nit: leftover debug statement

```suggestion
```

sarayourfriend (commented):
I am getting heaps of type errors due to files from the catalog's Airflow javascript getting included in. These might be resolved by rebasing and fixing the merge conflicts? I started it locally but it looks like it might be a doozy :sweat_smile:

frontend/src/pages/audio/_id/index.vue
obulat: It's not safe, thank you for catching this! It was a left-over from previous attempts.

frontend/src/stores/media/single-result.ts
obulat: Nice! Thank you!

frontend/src/stores/media/single-result.ts
obulat: Right. That PR was merged, and I rebased this one on it.

frontend/src/types/media.ts
obulat: I tried applying this suggestion, but then I get an error in the `VAllResultsGrid` component, which uses `isDetal.audio`/`isDetail.image` in the template. 
```
frontend/src/components/VSearchResultsGrid/VAllResultsGrid.vue(43,11): error TS2322: Type 'Media' is not assignable to type 'ImageDetail'.
  Types of property 'frontendMediaType' are incompatible.
    Type '"audio" | "image"' is not assignable to type '"image"'.
      Type '"audio"' is not assignable to type '"image"'.
frontend/src/stores/media/single-result.ts(47,9): error TS2322: Type '{ frontendMediaType: "image"; height?: number | undefined; width?: number | undefined; id: string; title: string; originalTitle: string; creator?: string | undefined; creator_url?: string | undefined; ... 21 more ...; isSensitive: boolean; } | { ...; } | null' is not assignable to type 'ImageDetail | null'.
  Type '{ frontendMediaType: "audio"; audio_set?: { title: string; foreign_landing_url: string; creator?: string | undefined; creator_url?: string | undefined; url?: string | undefined; filesize?: number | undefined; filetype?: string | undefined; } | undefined; ... 35 more ...; isSensitive: boolean; }' is not assignable to type 'ImageDetail'.
    Types of property 'frontendMediaType' are incompatible.
      Type '"audio"' is not assignable to type '"image"'.
```

frontend/src/types/media.ts
```diff
@@ -109,10 +110,15 @@ export interface ImageDimensions {
...
-  audio: (media: Media): media is AudioDetail => {
-    return media.frontendMediaType === "audio"
-  },
-  image: (media: Media): media is ImageDetail => {
-    return media.frontendMediaType === "image"
-  },
+  audio: (media: Media | null): media is AudioDetail =>
+    isMediaDetail<typeof AUDIO>(media, AUDIO),
+  image: (media: Media | null): media is ImageDetail =>
+    isMediaDetail<typeof IMAGE>(media, IMAGE),
```
sarayourfriend, line 116: FWIW (just a nit), TypeScript can infer `T` based on the argument passed for `mediaType`, so we don't need to explicitly pass the type parameter:

```suggestion
  audio: (media: Media | null): media is AudioDetail =>
    isMediaDetail(media, AUDIO),
  image: (media: Media | null): media is ImageDetail =>
    isMediaDetail(media, IMAGE),
```

It's a shame that the return type gets flattened to `boolean` though :cry: TS doesn't transfer the `is` narrowing to the return type of `audio` :shrug:

sarayourfriend (changes requested):
This LGTM aside from a strange edge-case bug. I don't see any errors in the SSR console for this, which also signals to me that perhaps things are failing in a way that we would want to surface :thinking: 

single-result-server-side-edge-case

My hunch is that it has something to do with rate limiting, I can reproduce it reliably once I get it to happen once.

sarayourfriend:
FWIW this might be somewhat existing behaviour, but because we do not currently show an _error_ page for it, you sort of expectantly sit through the page finally loading the single result if it failed server side (which I appear to be able to reproduce in production). But now that it (correctly) shows an error page, but then replaces the page with the actual result after a few seconds of waiting, it feels _very_ strange, like something is badly disconnected and the site doesn't know what is actually going on.

obulat:
@sarayourfriend, could you provide more details on this issue reproduction? I couldn't reproduce it locally.

sarayourfriend:
Sure:

1. Run `just frontend/run dev`
2. Search all media
3. Click into a single image result and refresh
4. If you do not see the issue, use browser back to go to the search results and select a new single image result
5. Repeat 3-4 until you see the issue

The issue is that the yellow error screen appears after refreshing the single result page (to SSR render it), but only _sometimes_. If a particular single result doesn't reproduce the issue, I just need to click through 3~5 more results and it'll eventually happen. The first time it happened on the second result I clicked on, but other times have taken clicking into and refreshing 4 or 5 individual image results.

obulat:
> The issue is that the yellow error screen appears after refreshing the single result page (to SSR render it), but only _sometimes_. If a particular single result doesn't reproduce the issue, I just need to click through 3~5 more results and it'll eventually happen. The first time it happened on the second result I clicked on, but other times have taken clicking into and refreshing 4 or 5 individual image results.

Thank you for the instructions! I managed to replicate it just a couple of times.
I pushed the fix, setting `fetchOnServer` to `false` in the single result pages. Without it, server-rendered pages would call `useFetch` some time after mounting the page, and if the request is successful, would load a page after showing the error, just like you describe.
Now, with `fetchOnServer` set to false, `useFetch` is not called on the server-rendered pages. The user would have to refresh the page to retry. This makes it even more important to display the actual error so that the users can see if it's a timeout error that needs to be retried, a 429 error which would show when the user would need to retry, or a 404 error which should mean that no re-tries are necessary.

frontend/src/middleware/single-result.ts
```diff
@@ -1,16 +1,36 @@
...
+    singleResultStore.setMediaById(mediaType, route.params.id)
+    if (from && from.path.includes("/search/")) {
+      const searchStore = useSearchStore($pinia)
+      searchStore.setBackToSearchPath(from.fullPath)
+      const searchTerm = Array.isArray(route.query.q)
+        ? route.query.q[0]
+        : route.query.q
+      if (searchTerm) {
+        searchStore.setSearchTerm(searchTerm)
+      }
```
sarayourfriend, line 33: Just a nit but this code is _very_ dense. Refactoring it to have some early returns or call to separate, clearly named functions for different scenarios to reduce the nested conditionals would go a long way to make it easier to read and understand. Whitespace would also help, which I've left a suggestion for


```suggestion
  const mediaType = route.fullPath.includes("/image/") ? IMAGE : AUDIO
  const singleResultStore = useSingleResultStore($pinia)

  if (process.server) {
    const media = await singleResultStore.fetch(mediaType, route.params.id)
    await useRelatedMediaStore($pinia).fetchMedia(mediaType, route.params.id)
    
    if (!media) {
      error(singleResultStore.fetchState.fetchingError ?? {})
    }
  } else {
    // Client-side rendering
    singleResultStore.setMediaById(mediaType, route.params.id)

    if (from && from.path.includes("/search/")) {
      const searchStore = useSearchStore($pinia)
      searchStore.setBackToSearchPath(from.fullPath)
      
      const searchTerm = Array.isArray(route.query.q)
        ? route.query.q[0]
        : route.query.q
    
      if (searchTerm) {
        searchStore.setSearchTerm(searchTerm)
      }
```

sarayourfriend (approved):
LGTM! Interesting fix using `fetchOnServer` in that way. I didn't understand at first what your comment read, but after reading the code it makes a lot of sense. Hopefully that would make things more flexible if it was eventually refactored into a `<script setup>` style component after the eventual Nuxt 3 change :thinking: 

I have to admit that I did not very carefully review all the code changes because there's simply too many to do a close reading of everything. I've left one nit on a place that was particularly dense, otherwise it's a lot of sprawling changes to fix this bug. Kudos for being able to keep track of it all :slightly_smiling_face:. The changes work perfectly for me locally and as long as our e2e and unit tests are passing I'm fine approving this.

obulat:
> LGTM! Interesting fix using `fetchOnServer` in that way. I didn't understand at first what your comment read, but after reading the code it makes a lot of sense. Hopefully that would make things more flexible if it was eventually refactored into a `<script setup>` style component after the eventual Nuxt 3 change 🤔

I started with the assumption that **on the server**, we want to block rendering before all the data is fetched, and **on the client**, we want to render the page skeleton and then fetch the data and render the details after the fetching is complete.  This way, for Server-rendered pages, we don't get the sudden transition from the skeleton to the error page when the data is finally fetched (and the Nuxt client-server mismatch errors in the console). And on the client, we get a much faster loading experience, when instead of waiting for the full data to load, we get the feeling that page loading is almost instant (and it mostly is, because for most search -> single detail transitions, we already have the data in our media stores, and the thumbnail is loaded). 

Nuxt has two options for fetching the data.
`fetch` loads the data asynchronously, before rendering the page on the server, and after mounting on the client. Which means that if the server received an error, the client will retry and if the request is successful, show the actual page (the quirk that @sarayourfriend you showed in the recordings above). While this could be useful in other scenarios, we want the server-rendered page to be fully rendered instead of being the skeleton that might be filled up on the client while hydrating.
_Note: I just realized that I don't know what will happen with cache. Do we cache the full server responses? If a user gets a 429 page from the server, will that be cached? @zackkrida, if we do cache the responses, do 429s get cached, too?_

`asyncData` blocks the page rendering until the data is fetched. Which is what we want on the server, but not on the client.

The solution in this PR became a combination of these two approaches: it acts as `asyncData` on the server and as `fetch` on the client. The middleware blocks rendering of the page on the server until all of the necessary data is fetched, and redirects to the error page if necessary. The `fetch` hook loads the data sometime after the page was mounted, which results in faster navigation/transitions. To prevent fetch from also fetching on the server (as we already have the middleware handling server-side fetching), we add `fetchOnServer: false` to the page.

obulat:
> I have to admit that I did not very carefully review all the code changes because there's simply too many to do a close reading of everything. I've left one nit on a place that was particularly dense, otherwise it's a lot of sprawling changes to fix this bug. Kudos for being able to keep track of it all 🙂. The changes work perfectly for me locally and as long as our e2e and unit tests are passing I'm fine approving this.

Thank you for your help cleaning this up. I had to change the approach several times in this PR, and all of the changes are really intertwined :)

sarayourfriend:
> Do we cache the full server responses? If a user gets a 429 page from the server, will that be cached? @zackkrida, if we do cache the responses, do 429s get cached, too?

Statuses > 400 are ignored from the cache :+1: 

> Which means that if the server received an error, the client will retry and if the request is successful, show the actual page

Thank you for explaining in more depth. Can you clarify something, though: do you know _what_ the server error was? Was the server getting throttled in a way that the client was not? I didn't see any errors in the server console logs or anything, so I wasn't understanding what the error _was_, aside from the behaviour of it needing to be fixed.

In other words: while this change fixes the issue where the server fails to render the page but the client might be able to, are we inadvertently covering up the fact that the server failed without recording _why_ that happened?

obulat:
> Thank you for explaining in more depth. Can you clarify something, though: do you know _what_ the server error was? Was the server getting throttled in a way that the client was not? I didn't see any errors in the server console logs or anything, so I wasn't understanding what the error _was_, aside from the behaviour of it needing to be fixed.

Normally, you do not see the server errors in the console. I am not 100% certain, but I added server logging when testing, and saw 429 server errors. Also, we do send those errors to Sentry, and set the `fetchingError` based on the server errors. We can monitor Sentry logs, and update the `fetchingError` values to closely match the actual server error.

> In other words: while this change fixes the issue where the server fails to render the page but the client might be able to, are we inadvertently covering up the fact that the server failed without recording why that happened?

We can set the `fetchingError` status code and use that to show the error page message accordingly. So, if the server got a 429, we would show a "Too many requests" pages, and the user would now that they can retry. The same would be true for 500. Currently, the default is 404, but I guess we should use a different error status for the default error to show the users that they _can_ retry the request, and use 404 only for pages that shouldn't be retried because the page does not exist.

frontend/src/components/VImageDetails/VRelatedImages.vue
```diff
@@ -14,26 +13,49 @@
...
+    )
+
+    const fetchState = computed(() => relatedMediaStore.fetchState)
+
+    return {
+      showRelated,
+
+      media,
+
+      fetchState,
```
zackkrida, line 57:
```suggestion
      showRelated,
      media,
      fetchState,
```

A little formatting oddness here.

zackkrida (approved):
LGTM. I spent a lot of time looking at the code and _think_ I'm grasping the changes 😁 Really nice work!

sarayourfriend:
> So, if the server got a 429, we would show a "Too many requests" pages, and the user would now that they can retry

This is really tricky though. If the _server_ gets a 429 then something is _severely_ wrong. In production this should never be the case because we use the rate-limit exempted API key on the server side. If the server gets a 429, it's essentially a catastrophic error that we need to immediately address.

This is triky!

obulat:
> > So, if the server got a 429, we would show a "Too many requests" pages, and the user would now that they can retry
> 
> This is really tricky though. If the _server_ gets a 429 then something is _severely_ wrong. In production this should never be the case because we use the rate-limit exempted API key on the server side. If the server gets a 429, it's essentially a catastrophic error that we need to immediately address.
> 
> This is tricky!

I didn't realize that we use a rate-limit exempted API key on the server side! Then we probably don't need to test for 429 exceptions on the server :) Thank you for noting this, it's really helpful, @sarayourfriend !

PR #2592, opened 2023-07-07
Add popularity refresh DAGs

### LINKED ISSUE #2089: Create `popularity_refresh` DAG factory

stacimc:
## Fixes

Fixes #2089 by @stacimc 

## Description

Adds a popularity refresh DAG factory that generates `audio_popularity_refresh` and `image_popularity_refresh` DAGs. Each DAG:
* refreshes popularity metrics and constants (recalculating the popularity constants)
* then, triggers a `batched_update` DagRun for each of the providers of this media type that support popularity data
* waits for the batched updates to complete and notifies Slack

Screen Shot 2023-07-19 at 5 13 21 PM

* It is currently on a `None` schedule, so must be triggered manually, and has very generous timeouts. These will be revisited in follow-up PRs after testing is done in production.

## Testing Instructions

The `TriggerDagRunOperator` we're using to trigger the `batched_update`s is put in `deferrable` mode, in order to free the worker slot while waiting for the batched updates to complete. This requires a Triggerer to be running. ~Locally, I did this by running `just catalog/shell` and then `airflow triggerer` to start the Triggerer.~ Run `just down -v && just up` to start the triggerer.

You'll also want to make sure you have `DATA_REFRESH_POKE_INTERVAL=5` in your `catalog/.env` so that you don't have to wait 30 minutes for the TriggerDagRunOperator to re-run.

Then run `just init` to get sample data in your local environment. The sample data all have null popularity scores, which you can verify by running `just catalog/pgcli` to open a pgcli session and then running:

```
SELECT COUNT(*) FROM image WHERE standardized_popularity is null;

SELECT COUNT(*) FROM audio WHERE standardized_popularity is null;
```

In both cases the result should be `5000`.

Now, simply run the `audio_popularity_refresh` and `image_popularity_refresh` DAGs locally. Both should pass with no task failures. You should go to `http://localhost:9090/dags/batched_update/grid` and verify that you see a separate DagRun for each provider that supports popularity data. You can inspect the logs for the `notify_updated_count` for each, and should see:
* INFO - Updated 3992 records for update: wikimedia_audio_popularity_refresh_20230707
* INFO - Updated 828 records for update: freesound_popularity_refresh_20230707
* INFO - Updated 180 records for update: jamendo_popularity_refresh_20230707
* INFO - Updated 0 records for update: nappy_popularity_refresh_20230707
* INFO - Updated 0 records for update: rawpixel_popularity_refresh_20230707
* INFO - Updated 0 records for update: wikimedia_popularity_refresh_20230707
* INFO - Updated 2500 records for update: flickr_popularity_refresh_20230707
* INFO - Updated 2500 records for update: stocksnap_popularity_refresh_20230707

Where 0 records were updated, this is because there are no records for that provider in our sample data. Your `query_id`s will differ in the date suffix.

____

Because this PR also moves some popularity task factories around, also run both data refreshes to ensure they pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/popularity/sql.py
+++ b/catalog/dags/common/popularity/sql.py
@@ -439,6 +439,54 @@ def create_media_view(
         postgres.run(create_audioset_view_query())
 
 
+def get_providers_with_popularity_data_for_media_type(
+    postgres_conn_id: str,
+    media_type: str = IMAGE,
+    constants_view: str = IMAGE_POPULARITY_CONSTANTS_VIEW,
+    pg_timeout: float = timedelta(minutes=10).total_seconds(),
+):
+    """
+    Return a list of distinct `provider`s that support popularity data,
+    for the given media type.
+    """
+    if media_type == AUDIO:
+        constants_view = AUDIO_POPULARITY_CONSTANTS_VIEW
+
+    postgres = PostgresHook(
+        postgres_conn_id=postgres_conn_id, default_statement_timeout=pg_timeout
+    )
+    providers = postgres.get_records(
+        f"SELECT DISTINCT provider FROM public.{constants_view};"
+    )
+
+    return [x[0] for x in providers]
+
+
+def format_update_standardized_popularity_query(
+    media_type=IMAGE,
+    standardized_popularity_func=STANDARDIZED_IMAGE_POPULARITY_FUNCTION,
+    table_name=TABLE_NAMES[IMAGE],
+    db_columns=IMAGE_TABLE_COLUMNS,
+    db_view_name=IMAGE_VIEW_NAME,
+    db_view_id_idx=IMAGE_VIEW_ID_IDX,
+    db_view_provider_fid_idx=IMAGE_VIEW_PROVIDER_FID_IDX,
+    task: AbstractOperator = None,
+):
+    """
+    Create a SQL query for updating the standardized popularity for the given
+    media type. Only the `SET ...` portion of the query is returned, to be used
+    by a `batched_update` DagRun.
+    """
+    if media_type == AUDIO:
+        table_name = TABLE_NAMES[AUDIO]
+        standardized_popularity_func = STANDARDIZED_AUDIO_POPULARITY_FUNCTION
+
+    return (
+        f"SET {col.STANDARDIZED_POPULARITY.db_name} = {standardized_popularity_func}"
+        f"({table_name}.{PARTITION}, {table_name}.{METADATA_COLUMN})"
+    )
+
+
 def update_db_view(
     postgres_conn_id, media_type=IMAGE, db_view_name=IMAGE_VIEW_NAME, task=None
 ):
--- a/catalog/dags/popularity/dag_factory.py
+++ b/catalog/dags/popularity/dag_factory.py
@@ -1,0 +1,179 @@
+"""
+# Popularity Refresh DAG Factory
+
+This file generates our popularity refresh DAGs using a factory function.
+
+For the given media type these DAGs will first update the popularity metrics table,
+adding any new metrics and updating the percentile that is used in calculating the
+popularity constants. It then refreshes the popularity constants view, which
+recalculates the popularity constant for each provider.
+
+Once the constants have been updated, the DAG will trigger a `batched_update`
+DagRun for each provider of this media_type that is configured to support popularity
+data. The batched update recalculates standardized popularity scores for all
+records, using the new constant. When the updates are complete, all records have
+up-to-date popularity data. This DAG can be run concurrently with data refreshes
+and regular ingestion.
+
+
+You can find more background information on this process in the following
+implementation plan:
+
+- [[Implementation Plan] Decoupling Popularity Calculations from the Data Refresh](
+https://docs.openverse.org/projects/proposals/popularity_optimizations/20230420-implementation_plan_popularity_optimizations.html)
+"""
+import datetime
+import logging
+import os
+
+from airflow import DAG
+from airflow.decorators import task
+from airflow.operators.trigger_dagrun import TriggerDagRunOperator
+from popularity.popularity_refresh_types import (
+    POPULARITY_REFRESH_CONFIGS,
+    PopularityRefresh,
+)
+from popularity.refresh_popularity_metrics_task_factory import (
+    create_refresh_popularity_metrics_task_group,
+)
+
+from common import slack
+from common.constants import DAG_DEFAULT_ARGS, POSTGRES_CONN_ID
+from common.popularity import sql
+from database.batched_update.constants import DAG_ID as BATCHED_UPDATE_DAG_ID
+
+
+logger = logging.getLogger(__name__)
+
+
+@task
+def notify_slack(text: str, media_type: str, dag_id: str) -> None:
+    slack.send_message(
+        text,
+        username=f"{media_type} Popularity Refresh",
+        icon_emoji=":database:",
+        dag_id=dag_id,
+    )
+
+
+@task
+def get_providers_update_confs(
+    postgres_conn_id: str, popularity_refresh: PopularityRefresh
+):
+    """
+    Build a list of DagRun confs for each provider of this media type. The confs will
+    be used by the `batched_update` DAG to perform a batched update of all existing
+    records, to recalculate their standardized_popularity with the new popularity
+    constant. Providers that do not support popularity data are omitted.
+    """
+    # For the media type, get a list of the providers who support popularity data
+    providers = sql.get_providers_with_popularity_data_for_media_type(
+        postgres_conn_id, popularity_refresh.media_type
+    )
+
+    # Once the new popularity constants are calculated, they will automatically be used
+    # to calculate popularity scores at ingestion. The popularity refresh need only
+    # update the scores for records which were last updated before this time.
+    last_updated_time = datetime.datetime.utcnow()
+
+    # For each provider, create a conf that will be used by the batched_update to
+    # refresh standardized popularity scores.
+    return [
+        {
+            # Uniquely identify the query
+            "query_id": (
+                f"{provider}_popularity_refresh_{last_updated_time.strftime('%Y%m%d')}"
+            ),
+            "table_name": popularity_refresh.media_type,
+            # Query used to select records that should be refreshed
+            "select_query": (
+                f"WHERE provider='{provider}' AND updated_on <"
+                f"'{last_updated_time.strftime('%Y-%m-%d %H:%M:%S')}'"
+            ),
+            # Query used to update the standardized_popularity
+            "update_query": sql.format_update_standardized_popularity_query(
+                popularity_refresh.media_type
+            ),
+            "batch_size": 10_000,
+            "update_timeout": popularity_refresh.refresh_popularity_batch_timeout,
+            "dry_run": False,
+            "resume_update": False,
+        }
+        for provider in providers
+    ]
+
+
+def create_popularity_refresh_dag(popularity_refresh: PopularityRefresh):
+    """
+    Instantiate a DAG for a popularity refresh.
+
+    For the given media type, this DAG will recalculate the popularity constants and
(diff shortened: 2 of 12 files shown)


catalog/dags/common/popularity/sql.py
```diff
@@ -439,6 +439,54 @@ def create_media_view(
...
+    task: AbstractOperator = None,
+):
+    """
+    Create a SQL query for updating the standardized popularity for the given
+    media type. Only the `SET ...` portion of the query is returned, to be used
+    by a `batched_update` DagRun.
+    """
+    if media_type == AUDIO:
+        table_name = TABLE_NAMES[AUDIO]
+        standardized_popularity_func = STANDARDIZED_AUDIO_POPULARITY_FUNCTION
```
zackkrida, line 482: Outside of the scope of this PR,  maybe we could have some kind of `MEDIA_TYPE_CONFIG` dictionary in the future? Then we could only pass the media type to `format_update_standardized_popularity_query` and retrieve all the db columns, table names and so on from the config.

zackkrida (approved):
THIS IS SO COOL. I had the pleasure of forgetting to enable the `batched_update` DAG in the Airflow UI and got to see all of the `refresh_popularity` mapped instances be marked as `deferred` until I enabled it. 

Works as described, and I don't see any suggestions for the code itself, so LGTM! Excited to use this and tune the timing intervals after we observe it in production.

AetherUnbound:
> The `TriggerDagRunOperator` we're using to trigger the `batched_update`s is put in `deferrable` mode, in order to free the worker slot while waiting for the batched updates to complete. This requires a Triggerer to be running. Locally, I did this by running `just catalog/shell` and then `airflow triggerer` to start the Triggerer.

Woah interesting! Will we need to change the deployment steps in production to start a triggerer there as well?

catalog/dags/common/popularity/sql.py
```diff
@@ -439,6 +439,54 @@ def create_media_view(
         postgres.run(create_audioset_view_query())
 
 
+def get_providers_with_popularity_data_for_media_type(
+    postgres_conn_id: str,
+    media_type: str = IMAGE,
+    constants_view: str = IMAGE_POPULARITY_CONSTANTS_VIEW,
```
AetherUnbound, line 445: Should we maybe just make these required variables, so we don't have to add additional logic for future media types?

Or if the concern is that this is a nested function and we don't want to hand in the pop constants view through the entire stack, maybe we could use a global mapping from media type to constants view and look up from there? Or use the media type to format the name of the constants view?

catalog/dags/common/popularity/sql.py
AetherUnbound: Oh, jinx with my own comments 😄

catalog/dags/popularity/dag_factory.py
```diff
@@ -0,0 +1,181 @@
...
+    # update the scores for records which were last updated before this time.
+    last_updated_time = datetime.datetime.utcnow()
+
+    # For each provider, create a conf that will be used by the batched_update to
+    # refresh standardized popularity scores.
+    return [
+        {
+            # Uniquely identify the query
+            "query_id": (
+                f"{provider}_popularity_refresh_{last_updated_time.strftime('%Y%m%d')}"
```
AetherUnbound, line 85: Do you think this is fine enough resolution for this DAG? There wouldn't be a case where we'd run it twice in one day potentially?

catalog/dags/popularity/dag_factory.py
```diff
@@ -0,0 +1,181 @@
...
+
+    dag = DAG(
+        dag_id=popularity_refresh.dag_id,
+        default_args=default_args,
+        start_date=popularity_refresh.start_date,
+        schedule=popularity_refresh.schedule,
+        max_active_runs=1,
+        catchup=False,
+        doc_md=__doc__,
+        tags=["popularity_refresh"],
```
AetherUnbound, line 132: Do you think it would make sense for this to also have the `"data_refresh"` tag? Or should we leave it off because it's independent?

catalog/dags/popularity/dag_factory.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    with dag:
+        poke_interval = int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60 * 30))
```
AetherUnbound, line 136: We define this in 3 places currently (albeit this is a different default), could we move it to a constants file potentially?

catalog/dags/popularity/dag_factory.py
```diff
@@ -0,0 +1,181 @@
...
+            # Wait for all the dagruns to finish
+            wait_for_completion=True,
+            # Release the worker slot while waiting
+            deferrable=True,
+            poke_interval=poke_interval,
+            retries=0,
+        ).expand(
+            # Build the conf for each provider
+            conf=get_providers_update_confs(POSTGRES_CONN_ID, popularity_refresh)
+        )
```
AetherUnbound, line 160: So cool that we're using dynamic tasks again!!

catalog/dags/popularity/popularity_refresh_types.py
```diff
@@ -0,0 +1,63 @@
...
+    schedule:                          string giving the schedule on which the DAG
+                                       should be run.  Passed to the
+                                       airflow.dag.DAG __init__ method.
+    refresh_popularity_batch_timeout:  timedelta expressing the amount of time
+                                       refreshing popularity scores for an individual
+                                       batch of records may take.
+    refresh_metrics_timeout:           timedelta expressing amount of time the
+                                       refresh popularity metrics and constants
+                                       may take.
+    doc_md:                            str used for the DAG's documentation markdown
```
AetherUnbound, line 36: Is this still accurate?

catalog/dags/popularity/popularity_refresh_types.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    dag_id: str = field(init=False)
+    media_type: str  # TODO
```
AetherUnbound, line 40: Should the `TODO` here be removed?

catalog/dags/popularity/refresh_popularity_metrics_task_factory.py
```diff
@@ -22,7 +23,9 @@
 UPDATE_MEDIA_POPULARITY_CONSTANTS_TASK_ID = "update_media_popularity_constants_view"
 
 
-def create_refresh_popularity_metrics_task_group(data_refresh: DataRefresh):
+def create_refresh_popularity_metrics_task_group(
+    refresh_config: DataRefresh | PopularityRefresh,
```
AetherUnbound, line 27: Very cool!!

AetherUnbound (changes requested):
This is fantastic! I'm so happy to see all of the pieces coming together here 😄 I was able to run the testing instructions locally and everything ran as expected. I have a few questions and notes, and want to mention again about the `airflow triggerer` as we'll likely need to make an adjustment for deployments prior to kicking this off.

Additionally, would you be willing to add some tests for the `get_providers_update_confs` function?

krysal:
I'll put this on draft while Staci catches up.

catalog/dags/common/popularity/sql.py
stacimc: Re this and also [this comment](https://github.com/WordPress/openverse/pull/2592#discussion_r1264269559) -- 100% yes :) In fact I started doing that in this PR but removed it because it created a much larger changeset. I kept to the current convention for this work, but I'll create an issue for refactoring this separately (unless @AetherUnbound do you think this should be updated in this PR?)

AetherUnbound: Ah right, cause it's in `sql.py` there could be a lot more changes present. Yes, I think a separate issue is good!

catalog/dags/popularity/popularity_refresh_types.py
stacimc: Good call -- I don't think we have a use case for overriding this anyway. Removed.

catalog/dags/popularity/dag_factory.py
stacimc: The query_id is used for building the temp table in the triggered `batched_update` runs, which is dropped when the update is successful. If it's not successful, we should be managing that by clearing tasks or by manually triggering a new one with the `resume_update` param to use the existing temp table.

This DAG has `max_active_runs=1`, so I think the only way to get a collision would be to start a popularity refresh, fail it and at least one of the triggered `batched_update`s, and then retry the `popularity_refresh` DAG on the same day. In that case a failure seems fine, since that's not the intended way to handle errors for batched updates anyway.

catalog/dags/popularity/dag_factory.py
stacimc: When the project is complete, the data and popularity refreshes should be totally decoupled. There's definitely a case to be made that they're _related_ but I think it might be more confusing to use the `data_refresh` tag here, in case it implies a relationship that no longer exists. But that's not a strongly held opinion :)

catalog/dags/common/popularity/sql.py
stacimc: I opened #2678 to refactor all of `popularity/sql.py` to do this 💯

stacimc:
I think that's all feedback addressed. I also modified the DAG slightly to pull `get_last_updated_time` out into its own task. This (1) ensures that the cutoff `updated_on` time is taken _after completing the popularity constant refresh_ and (2) makes testing `get_providers_update_confs` much easier :)

stacimc:
Actually, drafting again for a moment while I look into changes necessary to start the `triggerer` in production.

stacimc:
I think we'll need to update the catalog's `docker-compose` in the infrastructure repo as well in order to start the triggerer when deploying production, although I'm not 100% sure what changes we'll need or how to test them. This PR has been updated so you should now be able to `just down -v && just up` and the triggerer will start automatically in local development.

krysal (approved):
Awesome! Worked as indicated. It seems that we will need a new service instance for the catalog. Looks in line with the [project recommendations](https://airflow.apache.org/docs/apache-airflow/stable/howto/docker-compose/index.html#fetching-docker-compose-yaml).

AetherUnbound (approved):
LGTM! Thanks for looking into the triggerer stuff.

Issue #2368, opened 2023-06-09
Component: Sensitive content fetch/blur sidebar toggles


dhruvkb:
## Description

Define the component that contains two toggles, one to fetch the sensitive content from the API and the other that controls whether the sensitive content fetched should be blurred in the UI.

## API

This component need not accept props or emit events, it can be directly synced with the state held in the Pinia store.

### State

| Name | Type | Description |
|-|-|-|
| shouldFetchSensitive | boolean | whether sensitive content should be fetched from the API |
| shouldBlurSensitive | boolean | whether sensitive content should be blurred in the search results and single result page |

`shouldFetchSensitive` is synced with the feature flag store, specifically the [`fetch_sensitive` feature flag](https://github.com/WordPress/openverse/blob/de7d99e70bd56df8552608a9edc0213821068089/frontend/feat/feature-flags.json#LL36C6-L36C21).

`shouldBlurSensitive` can be stored in the `ui` store but it should not be persisted across visits, so it cannot be committed to a cookie or put into `localStorage`.

> **Note**
> The second toggle should be turned off **and** disabled if the former is off.

## Screenshot

<img width="293" alt="Screenshot 2023-06-09 at 9 13 42 AM" src="https://github.com/WordPress/openverse/assets/16580576/6c348ae3-16cd-4741-8ae0-cd695ef72ea3">

## Figma 

- **Figma:** https://www.figma.com/file/bDTfGzBit8rGRPktOR1cu4/Safe-content-browsing-flow?type=design&node-id=1140-23930&t=t4b5KNE2t2viAres-4

wasimTQ:
@dhruvkb I'd like to contribute to this issue. I've looked through the code and I have to clarify how it should be implemented with you.

Should I use the feature flag store and manipulate the changes there? 

Or the other option should be to 
1. Add more fields `fetch_sensitive_content, blur_sensitive_content` in `filterCodesPerCategory.mature`
2. Add entries to the locales files under filter.{mature_keys}
3. Let the already available `searchStore` do all the work

Let me know if this is the ideal solution for this issue.

dhruvkb:
@wasimTQ this issue is mainly only about implementing this specific component and interfacing it with the UI and feature flag stores. You do not need to connect this with the API, that will be handled by the search store in a separate issue.

PR #2617, opened 2023-07-11
Remove irrelevant mature UI toggle test

### LINKED ISSUE #2572: "e2e/search-query-server.spec.ts:94:9 › search query on SSR › url mature query is set, and can be unchecked using the Safer Browsing popup" is skipped
## Problem

"e2e/search-query-server.spec.ts:94:9 › search query on SSR › url mature query is set, and can be unchecked using the Safer Browsing popup" is skipped without any explanation or associated issue that I can find. When I unskip the test it does not pass locally.

## Description

Figure out why the test does not pass. If the test is a good test to have, fix it. Otherwise, if the test is not necessary, delete it from the test suite.

## Additional context

I've marked this critical as per our flaky test guidelines but because it's already been skipped for a long time, it doesn't exactly fall into the same category. But someone did write the test for a reason, so presumably it's a good one to have. It also only fails for two breakpoints rather than all of them, so the behaviour is verifiable in _some_ conditions. Perhaps it just needs to be wrapped in a conditional based on the breakpoint being tested and excluded from the suite for certain breakpoints.

zackkrida:
## Fixes

Fixes #2572 by @sarayourfriend 

## Description

Removes an irrelevant mature UI toggle test. We will add additional test coverage for new sensntivity UI elements when they are added in #2368.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/playwright/e2e/search-query-server.spec.ts
+++ b/frontend/test/playwright/e2e/search-query-server.spec.ts
@@ -91,21 +91,13 @@ test.describe("search query on SSR", () => {
       }
     })
 
-    test.skip("url mature query is set, and can be unchecked using the Safer Browsing popup", async ({
-      page,
-    }) => {
-      await goToSearchTerm(page, "cat", {
-        searchType: IMAGE,
-        query: "mature=true",
-      })
-
-      await page.click('button:has-text("Safer Browsing")')
-
-      const matureCheckbox = await page.locator("text=Show Mature Content")
-      await expect(matureCheckbox).toBeChecked()
-
-      await page.click("text=Show Mature Content")
-      await expect(page).toHaveURL("/search/image?q=cat")
-    })
+    test.fixme(
+      "URL sensntive query is set, and can be toggled using the UI.",
+      async () => {
+        // TODO: This test should be updated as part of the
+        // blurring sensitive results project.
+        // https://github.com/WordPress/openverse/issues/377
+      }
+    )
   })
 })
--- a/frontend/test/playwright/e2e/search-query-server.spec.ts
+++ b/frontend/test/playwright/e2e/search-query-server.spec.ts
@@ -92,7 +92,7 @@ test.describe("search query on SSR", () => {
     })
 
     test.fixme(
-      "URL sensntive query is set, and can be toggled using the UI.",
+      "URL sensitive query is set, and can be toggled using the UI.",
       async () => {
         // TODO: This test should be updated as part of the
         // blurring sensitive results project.
--- a/frontend/test/playwright/e2e/search-query-server.spec.ts
+++ b/frontend/test/playwright/e2e/search-query-server.spec.ts
@@ -90,14 +90,5 @@ test.describe("search query on SSR", () => {
         await assertCheckboxStatus(page, checkbox)
       }
     })
-
-    test.fixme(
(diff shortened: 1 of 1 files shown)


frontend/test/playwright/e2e/search-query-server.spec.ts
```diff
@@ -91,21 +91,13 @@ test.describe("search query on SSR", () => {
...
-      await page.click('button:has-text("Safer Browsing")')
-
-      const matureCheckbox = await page.locator("text=Show Mature Content")
-      await expect(matureCheckbox).toBeChecked()
-
-      await page.click("text=Show Mature Content")
-      await expect(page).toHaveURL("/search/image?q=cat")
-    })
+    test.fixme(
+      "URL sensntive query is set, and can be toggled using the UI.",
```
zackkrida, line 95:
```suggestion
      "URL sensitive query is set, and can be toggled using the UI.",
```

typo

frontend/test/playwright/e2e/search-query-server.spec.ts
```diff
@@ -91,21 +91,13 @@ test.describe("search query on SSR", () => {
...
-
-      await page.click('button:has-text("Safer Browsing")')
-
-      const matureCheckbox = await page.locator("text=Show Mature Content")
-      await expect(matureCheckbox).toBeChecked()
-
-      await page.click("text=Show Mature Content")
-      await expect(page).toHaveURL("/search/image?q=cat")
-    })
+    test.fixme(
```
obulat, line 94: Now I kind of think we should just remove it. It doesn't really have any meaning now.

zackkrida: We could keep some of the implementation, if that is a more compelling reason to keep the test. I'm also fine with just deleting it.

sarayourfriend:
I also agree with removing it. If any part of this is useful, we can easily retrieve it from version control. At this point there are more than sufficient references to it that multiple maintainers have seen to know where to find it. I have frankly very little confidence that a TODO that says "this test should be implemented as part of XYZ project" is more useful. On the other hand, I think the `.skip` here is good evidence that it is very easy for us to forget about a test like this.

My question is: what is the reason (value add) to keep this test, even a stub/reference for it _in the code_ without an associated feature? If we want to add todo tests for the feature, we can do so, but lets do it for the entire feature, rather than keeping a single test with a name referencing a legacy feature that may or may not have anything to do with how the new feature ends up working. To be the downside is that this has a high likelihood of being forgotten/ignored and creating confusion in the future when someone has to dig up our conversations about this again.

zackkrida:
> If we want to add todo tests for the feature, we can do so, but lets do it for the entire feature, rather than keeping a single test with a name referencing a legacy feature that may or may not have anything to do with how the new feature ends up working. 

Agreed, I'll simply remove the code instead.

sarayourfriend (approved):
LGTM :+1: The PR description needs updating as well, just in case that wasn't on your radar :slightly_smiling_face:

AetherUnbound (approved):
Agreed with all the discussion here, removing sounds good!

Issue #2633, opened 2023-07-12
Remove ESLint webpack plugin


sarayourfriend:
## Problem

The ESLint webpack plugin prevents you from using the application if _any_ ESLint errors exist, including ones that are quite common when debugging code (like unused variables or imports). This makes quick debugging tedious because you inevitably forget that this is the case and have to then go make an additional change (which you'll then need to also set back) to resolve the error and then wait for the app to recompile.

## Description

Remove this plugin from the Nuxt/Webpack configuration.

## Additional context

IIRC we discussed this before in the old repository but I couldn't find the issue. I don't remember what the end of that discussion was.

Does anyone _like_ this plugin? pre-commit and CI catch anything anyway, I struggle to find what the utility of this plugin is for our project :thinking:

obulat:
> The ESLint webpack plugin prevents you from using the application if _any_ ESLint errors exist, including ones that are quite common when debugging code (like unused variables or imports). This makes quick debugging tedious because you inevitably forget that this is the case and have to then go make an additional change (which you'll then need to also set back) to resolve the error and then wait for the app to recompile.

I have wondered what the eslint module is used for several times, but hadn't looked into it. And have run into the debugging problems you describe countless times. I am 10000000 for removing it 😆 Thank you for opening the issue!

PR #2641, opened 2023-07-13
Update quickstart.md, added new 4th step making it 5 step documentation

### LINKED ISSUE #2616: ESLint editor integrations fail if `@openverse/eslint-plugin` isn't built
## Description

ESLint editor integrations will fail if `@openverse/eslint-plugin` isn't built. The plugin is built when the `eslint` script is run, but editors run ESLint on their own, not using the script, so the config will fail to parse correctly due to the Openverse plugin not having available code.

# Resolution

Add a `postinstall` script to the root package.json that builds all packages:

```json
{
  "scripts": {
    "postinstall": "pnpm --filter ./packages/* run build"
  }
}
```

## Additional context

- Reported: https://github.com/WordPress/openverse/pull/2601#issuecomment-1630714233
- Known fix: https://github.com/WordPress/openverse/pull/2601#issuecomment-1631030574

## Alternatives

An alternative to this that does not rely on pre/post scripts (which pnpm doesn't _want_ to run, but reluctantly does if you configure it to, which we only do now because of some dependencies that need it) is for us to export our own `eslint` binary from a new package `@openverse/eslint-bin` that correctly filters the list of folders, etc, that are evaluated (which editors currently do not have information about). Basically, transform the root package.json `eslint` script into _the_ ESLint binary that editors would use for our project.

This only works for editors that will use local ESLint binaries rather than globally installed ones or ones installed by the editor plugin. It's also potentially way more complex than just use a `postinstall` script, which, while not ideal, is still a standard Node project lifecycle feature.

Dev-Arhaan:
## Fixes

Fixes #2616 by @dhruvkb

## Description

Added pnpm filtering command to restrict commands to `@openverse/eslint-plugin` as new 4th step.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


documentation/frontend/guides/quickstart.md
```diff
@@ -38,7 +38,20 @@ you need to run this.
...
+   You might want to try Filtering to compile just the Openverse ESLint plugin.
+
+   ```console
+   $ pnpm --filter='@openverse/eslint-plugin' build
+   ```
+
+   Or
+
+   ```console
+   $ pnpm --F '@openverse/eslint-plugin' build
```
krysal, line 51: The short form of the option is with one dash only. Otherwise it looks good!
```suggestion
   $ pnpm -F '@openverse/eslint-plugin' build
```

documentation/frontend/guides/quickstart.md
```diff
@@ -38,7 +38,20 @@ you need to run this.
    $ just node-install
    ```
 
-4. To bring up the frontend, we have another `just` recipe. We have `just`
+4. Filtering allows you to restrict commands to specific subsets of packages.
+   You might want to try Filtering to compile just the Openverse ESLint plugin.
```
AetherUnbound, line 42: Slight rewording for clarity:
```suggestion
   If you encounter errors after this, you may need to compile just the Openverse ESLint plugin.
```

AetherUnbound (commented):
Thanks for your contribution! In addition to Krystle's correction, I have one additional suggestion.

Dev-Arhaan:
Your welcome! Hoping to contribute more to this project in future. Tell me if there is any other thing I can do

sarayourfriend:
@dhruvkb and @AetherUnbound @krysal who have reviewed. I think https://github.com/WordPress/openverse/issues/2633 is a better solution for this because it removes this unnecessary dependency. Alternatively, we should add building the ESLint package to `postinstall` in the frontend package, if we really want to keep the ESLint Webpack warnings plugin.

That is to say, additional instructions are not the solution here because we can easily automate the building of the package in our regular process. Which is far better, because then the plugin will actually be updated when it changes and is evaluated by the Webpack plugin.

Dev-Arhaan (commented):
This can be merged now? review it and let me know if I can improve anything else

AetherUnbound:
@sarayourfriend thanks for your perspective, I was primarily reviewing based on the context from the issue (since I don't work to directly with these tools). Do you think this is sufficiently addressed by #2643?

@Dev-Arhaan thanks for your patience while we figure out the best path forward on this!

documentation/frontend/guides/quickstart.md
```diff
@@ -38,7 +38,19 @@ you need to run this.
...
+
+   ```console
+   $ pnpm --filter='@openverse/eslint-plugin' build
+   ```
+
+   Or
+
+   ```console
+   $ pnpm -F '@openverse/eslint-plugin' build
+   ```
```
dhruvkb, line 51: This seems an unnecessary repetition of the previous command. We could omit this.

dhruvkb (approved):
While #2633 does fix the case where ESLint blocks the dev window, it doesn't address the fact that the ESLint package is not compiled when the dev server is run directly. I would prefer this to be in the docs.

sarayourfriend:
> While https://github.com/WordPress/openverse/issues/2633 does fix the case where ESLint blocks the dev window, it doesn't address the fact that the ESLint package is not compiled when the dev server is run directly. I would prefer this to be in the docs.

The eslint plugin is only a dependency of the frontend because of the ESLint Webpack plugin. The frontend Nuxt build shouldn't need the ESLint plugin _at all_ and therefore there's no need to manually run a build of it. When the package.json `eslint` script runs, it does build the plugin so that it's available when it is needed.

Is there something else that is causing a dependency between the built `@openverse/eslint-plugin` package and `dev`? There shouldn't be :thinking:

sarayourfriend:
I'll draft this until @dhruvkb responds to prevent further unnecessary pings.

dhruvkb:
Sorry, it took me a while to process your message @sarayourfriend. I think IDEs like WebStorm run ESLint constantly in the background to underline issues. They will break unless the extension is compiled. I would prefer an approach that has the package compiled even before running `dev`, so that my IDE doesn't complain, hence documenting the step seemed like the right approach.

sarayourfriend:
> I would prefer an approach that has the package compiled even before running dev, so that my IDE doesn't complain, hence documenting the step seemed like the right approach.

Could we add it as a `postinstall` step in the root package.json? The main thing I want to avoid is adding _yet another step_ for setup if there's any way at all that we can automatically handle it.

I suggest `postinstall` because ESLint won't work correctly at all without `pnpm install` anyway.

What do you think?

sarayourfriend:
Done :+1:

Dev-Arhaan:
> @sarayourfriend that seems like a good idea. Could you please update the issue to reflect this change of approach and also add some steps for how to resolve it?
> 
> @Dev-Arhaan thanks for working on this and we're sorry for not being able to utilise this PR. Once, the issue has been updated to use `postinstall`, would you be open to making a PR using that approach?

Hey, I saw this today, I was/am busy with moving to hostel, just let me know what needs to be added and I'll put a pr out by day after tomorrow

krysal:
We appreciate your help here @Dev-Arhaan. Please review the updated issue and make another PR to continue.

PR #2642, opened 2023-07-13
Check readiness against filtered index

AetherUnbound:
## Description

This PR fixes an issue wherein the filtered index creation DAG was checking for the readiness of the unfiltered index rather than the produced, filtered index. In most normal runs this would never raise an issue, because the indices are usually uniform (e.g. `image-12345` and `image-12345-filtered`, so checking against `image-12345` passes). When the create filtered index DAG is run manually, it generates a different suffix for the filtered index, one which doesn't have a matching unfiltered index (e.g. `image-98765-filtered`, which was made from `image-12345`, is trying to check for `image-98765` which does not exist).

This is rectified by running the check against the produced filtered index (suffix + `-filtered`) rather than the index just referenced with the suffix.

## Testing Instructions

1. Checkout `main`
1. Enable the `image_data_refresh` and `create_filtered_image_index` DAGs locally
1. Observe both DAGs complete successfully
1. Trigger a manual run of the latter DAG (without params)
1. Observe that it gets stuck on the `index_readiness` task with the log messages `[2023-07-13, 18:24:52 UTC] {http.py:182} ERROR - {"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index [image-ebea3ea3c04e4d369c8f5f6c614e893d]","resource.type":"index_or_alias","resource.id":"image-ebea3ea3c04e4d369c8f5f6c614e893d","index_uuid":"_na_","index":"image-ebea3ea3c04e4d369c8f5f6c614e893d"}],"type":"index_not_found_exception","reason":"no such index [image-ebea3ea3c04e4d369c8f5f6c614e893d]","resource.type":"index_or_alias","resource.id":"image-ebea3ea3c04e4d369c8f5f6c614e893d","index_uuid":"_na_","index":"image-ebea3ea3c04e4d369c8f5f6c614e893d"},"status":404}` (with the equivalent suffix that was produced during the run)
1. Mark that task as succeeded so the DAG can complete
1. Checkout this branch
1. Trigger another manual run (no params) of the `create_filtered_image_index` DAG
1. Observe that this one completes successfully without intervention

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_refresh/create_filtered_index_dag.py
+++ b/catalog/dags/data_refresh/create_filtered_index_dag.py
@@ -213,7 +213,7 @@ def filtered_index_creation_dag_factory(data_refresh: DataRefresh):
         # Await healthy results from the newly created elasticsearch index.
         index_readiness_check = ingestion_server.index_readiness_check(
             media_type=media_type,
-            index_suffix=destination_index_suffix,
+            index_suffix=f"{destination_index_suffix}-filtered",
         )
 
         do_point_alias = point_alias(destination_index_suffix=destination_index_suffix)


zackkrida (approved):
LGTM. I tried to test locally but I'm encountering an issue that I don't believe is related to this PR. After `just init` my cluster is yellow, because there's a single node and it wants to allocate replica shards for each index, but can't because replicas can't be on the same node.

krysal (approved):

AetherUnbound:
> LGTM. I tried to test locally but I'm encountering an issue that I don't believe is related to this PR. After `just init` my cluster is yellow, because there's a single node and it wants to allocate replica shards for each index, but can't because replicas can't be on the same node.

Hmm, the cluster is usually always yellow locally from what I've found, that shouldn't prevent any of the steps from running 😮

zackkrida:
Oh, got it. When I saw it was yellow I didn't bother continuing 😳

We should probably find a way to turn off replicas for local development only, so that the cluster can always be healthy. Or have a data and master node instead of just a master node.  I'll make an issue for that!

Issue #1916, opened 2023-04-27
Add unit tests for Python automation scripts


sarayourfriend:
## Problem

We need unit tests to verify that output is as expected given specific inputs, as made obvious by the various (incomplete) attempts to fix the changelog workflow generation script: https://github.com/WordPress/openverse/pull/1914#discussion_r1178576092

## Description

Add unit tests to the automation scripts. This will require re-writing them to make them unit testable (export a function that can be called, etc).

dhruvkb:
We've slowly ported almost all Python scripts over to JS. Once they're all consistently using the same structure (MJS files with `main` invoked by `actions/github-script`), we can add unit tests for them and the functions they call. It's all coming together 👍 (but slowly)!

PR #2647, opened 2023-07-14
Show individual doc differences for rendered documentation preview on PRs

### LINKED ISSUE #2646: Add a list of (changed) rendered files to the "docs emitted" PR comment
## Description

Presently we have a step in our CI/CD which builds the docs in a preview directory, then emits a message saying as much with a path to the main preview page: https://github.com/WordPress/openverse/blob/399036e699b1c0834624ba7174bd74ba1bff9f76/.github/workflows/ci_cd.yml#L769

Recently, @obulat added an edited note on the comment to help community contributors view the rendered project plan easier: https://github.com/WordPress/openverse/pull/2637#issuecomment-1632394971

It would be ideal if we could generate a list of the doc files which have been changed or differ from `main`, and then display that inside the comment as well. This might be possible using simple [directory diffing](https://stackoverflow.com/questions/2019857/diff-files-present-in-two-different-directories) at this step:

https://github.com/WordPress/openverse/blob/399036e699b1c0834624ba7174bd74ba1bff9f76/.github/workflows/ci_cd.yml#L824-L829

We'd need to account for scenarios which potentially change all (dozens at this point?) of the files in the docs directory. Perhaps we after a certain limit we can emit a message that says "too many files changed to reference each individually".

AetherUnbound:
## Fixes

Fixes #2646 by @AetherUnbound 

## Description

This PR adds some new logic to the emit docs step to determine which files have been added or changed, and notes them in the docs preview. The easiest way I found to accomplish this was with the `diff` CLI tool, but the output of that tool is not necessarily machine-readable and thus more complicated than a series of bash commands inside a GitHub Actions' `run` block. I instead put all the logic inside a python file which is called by the workflow. This emits a comment (preserved for posterity) like the one below:

image

I knowingly did not add tests for this script - we don't yet have tests (but want to add them as part of #1916), and setting up testing for the automation scripts would be its own bolus of work. I didn't want to include that in this PR, so we can tackle that as part of #1916 down the line.

## Testing Instructions

Check the emit docs comment!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


automations/python/workflows/get_folder_differences.py
```diff
@@ -0,0 +1,143 @@
...
+PR_NUMBER = os.environ.get("PR_NUMBER")
+EXCLUSIONS = {
+    "*.js",
+    "*.json",
+    "*.html",
+    ".*",
+    "CNAME",
+    "*.pickle",
+    "*.doctree",
+    "*.inv",
```
obulat, line 31:
```suggestion
    "*.inv",
    "*.txt",
    "*.map",
```

These two also need to be excluded based on the comment to this PR. I'm not so sure about the `txt`, though, would there ever be `txt` files added that we would like to review?

obulat (approved):
Awesome work! A lot of tricky edge-cases, too :) This will review the friction from reviewing the docs changes.

I wonder if we could add a change to a storybook story here to see if the link is generated.  I've added a patch that updates the story for VLicenseElements, removing the deprecated `default` parameters and replacing them with `args`.

[Remove_deprecated_defaults.patch](https://github.com/WordPress/openverse/files/12212017/Remove_deprecated_defaults.patch)

dhruvkb (approved):
Nice addition! Maybe in the future we can ship this as a standalone action, I imagine it could be useful for many other projects.

zackkrida:
@dhruvkb to your point, there seem to be a few existing actions which allow for listing changed files, for example: https://github.com/marketplace/actions/changed-files

@AetherUnbound's custom approach is still better, I reckon, since it does the link formatting we need.

sarayourfriend:
The action we currently use to determine where changes have happened in the repo can also provide the list of changed files: https://github.com/dorny/paths-filter#usage

If we add the `list_files` option, we could output that from the `get-changes` action, then we could get the files from `outputs.documentation_files`. Probably easier than parsing shell output.

I've left some other questions as well.

automations/python/workflows/get_folder_differences.py
sarayourfriend: Would it be easier to only include relevant files, instead of playing whack-a-mole with file types?

https://stackoverflow.com/a/10132592

AetherUnbound: @sarayourfriend I've tried a few different attempts based off this approach ([example](https://github.com/WordPress/openverse/actions/runs/5906329361/job/16022325010?pr=2647)) but for some reason `diff` is _still_ comparing files it should be excluding 😕 it's not ideal, but in the interest of not spending hours on this I'll add the additional exclusions and keep the current approach.

AetherUnbound: I've also added some handling in 8e5ff36e19fec512ace39beb90ad4b4737cf9763 to ensure we only display valid HTML pages in the diff.

AetherUnbound:
Pushing one last commit to remove the example file, then merging this!
image

PR #2663, opened 2023-07-17
Async Django app

### LINKED ISSUE #2566: Evaluate feasibility of Django ASGI and ADRF conversion
## Problem

Openverse needs to make a determination about how to best handle thumbnail traffic, as discussed here:

https://github.com/WordPress/openverse/discussions/2498

Converting the entire API to work asynchronously is a compelling choice, as asynchronicity benefits all parts of the API. However, we are uncertain about the compatibility of switching to ASGI and ADRF with the existing codebase.

## Description

A contributor should run a timeboxed (perhaps 4 development hours for initial experimentation, with an additional 4 if necessary to close the issue) in an attempt to determine if this conversion is _possible._

sarayourfriend:
## Fixes

Fixes #2566 by @zackkrida

## Description

This is still a WIP. I haven't done anything to unit tests at all, so I'm sure many/all of them will fail. Wanted to get this up early though, I'm excited that it was possible at all to get it working!

Adds `adrf` along with `uvicorn` and `django-async-redis`. Also removes the synchronous `wsgi-basic-auth` which we weren't using in production at all.

I'll leave notes on other specific changes inline.

I decided to take the approach of separating the async routes from the sync routes. We can move routes one-by-one this way without needing to wrap entire routes in `sync_to_async`. It duplicates a small amount of code this way and exposes how bare-bones `adrf` is, but I don't think it's a big detriment. Future work to re-implement routes in the async view can happen at leisure. They'll coexist side-by-side without issue for now.

The other big change is to convert the `image_proxy` modules to async. This was not as straight-forward as I had hoped due to some strange event loop behaviour when using async Redis for the tallies DB in `image_proxy.get`. If you change that to use `django_async_redis.get_redis_connect` (and `await` it), you'll get a strange hanging event loop and subsequent closure issue on every other thumbnail request caused by redis-py. I dug around for a solution but couldn't find anything for now. Luckily, we do not need to solve this right away. For now, using `sync_to_async` is a safe way to adapt the synchronous redis client for usage in our async views. It might be worth switching to a different approach for redis connection management interactions than django's cache module. Django channels (https://channels.readthedocs.io/en/stable/introduction.html) could be a much better way to manage these connections. It's redis support already has full async and it comes with heaps of other cool features we could use like background workers (for things like letting thumbnail generation run beyond the timeout of a single request and be cached in Photon for subsequent requesters).

If we deploy this (or something based on it), it would be in multiple stages:

1. Merge this change and continue using the WSGI app with an "async" thumbnails route (still handled by a single worker)
2. Update the deployment template to use uvicorn and the ASGI app for true async thumbnails route. Other routes automatically get wrapped with `sync_to_async` and handled in the loop.

We can use gunicorn with uvicorn workers. That way we still have multiple workers but they can each handle asynchronous request lifecycles: https://www.uvicorn.org/deployment/#deployment

Even if we have sync routes handled by those workers, we won't be any worse off for those routes than we are with the current sync handlers. As we convert more routes to async, we'll get bigger improvements.

There might be some way to tell gunicorn to direct thumbnail requests to a subset of workers with sync requests to another. I'll need to dig into it more. Ultimately it shouldn't be an issue though because the sync routes are all very fast for the most part already.

## Testing Instructions

`just api/up` should work after `just build web`. The dockerfile is configured to run uvicorn locally as a replacement for `runserver`, which doesn't properly support a full ASGI app. That means that we'll be able to fully test the ASGI app when running the app locally.

Additional testing is needed to confirm that running the app with gunicorn and uvicorn workers works correctly. This is the recommended approach from uvicorn for production deployments.

**Reviewers**: For now, please focus first on the overall approach rather than implementation details. In particular, splitting the views vs wrapping existing views in `sync_to_async` etc. If you agree with the overall approach, then feel free to dive into the implementation details. Please do test this locally, even though our test suite is pretty comprehensive.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/conf/wsgi.py
```diff
@@ -6,20 +6,11 @@
...
-
-monkey.patch_all()
-
-
 import os
 
 from django.core.wsgi import get_wsgi_application
 
-from wsgi_basic_auth import BasicAuth
-
```
sarayourfriend, line 21: This middleware was never used.

api/conf/wsgi.py
```diff
@@ -6,20 +6,11 @@
 For more information on this file, see
 https://docs.djangoproject.com/en/4.2/howto/deployment/wsgi/
 """
-
-from gevent import monkey
-
-
-monkey.patch_all()
```
sarayourfriend, line 13: We don't use geventlet anymore, so this patching was not necessary and conflicts with `asyncio`.

api/Dockerfile
```diff
@@ -101,7 +101,6 @@ COPY --chown=opener . /api/
 # Collect static assets, these are used by the next stage, `nginx`
 RUN env \
   SETUP_ES="False" \
-  STATIC_ROOT="/static" \
```
sarayourfriend, line 104: This is now correctly configured in settings for local and production deployments (which are intercepted by Nginx anyway) so there's no need to repeat the configuration here.

api/api/utils/aiohttp.py
```diff
@@ -0,0 +1,15 @@
...
+# aiohttp recommends reusing the same session for the whole application
+# https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request
+_SESSION: aiohttp.ClientSession = None
+
+
+def get_aiohttp_session():
+    global _SESSION
+
+    if _SESSION is None or _SESSION.loop.is_closed():
+        _SESSION = aiohttp.ClientSession()
```
sarayourfriend, line 13: Checking the event loop makes unit testing easier and is extra security for cases where the event loop closed unexpectedly. If the loop associated with the session is closed, we _need_ a new session or it'll try to use the closed event loop.

api/api/utils/check_dead_links/__init__.py
```diff
@@ -32,9 +33,11 @@ def _get_expiry(status, default):
     return config(f"LINK_VALIDATION_CACHE_EXPIRY__{status}", default=default, cast=int)
 
 
-async def _head(url: str, session: aiohttp.ClientSession) -> tuple[str, int]:
+async def _head(url: str, **kwargs) -> tuple[str, int]:
```
sarayourfriend, line 36: Changes in this module are just to support the reusable aiohttp session. We should have been doing that already anyway.

api/api/utils/image_proxy/__init__.py
```diff
@@ -88,24 +95,24 @@ def get(
...
-        upstream_response.raise_for_status()
+            timeout=aiohttp.ClientTimeout(total=15),
+        ) as upstream_response:
+            await incr(f"thumbnail_response_code:{month}:{upstream_response.status}")
+            await incr(
+                f"thumbnail_response_code_by_domain:{domain}:"
+                f"{month}:{upstream_response.status}"
+            )
+            upstream_response.raise_for_status()
+            body = await upstream_response.read()
```
sarayourfriend, line 111: Response must be read here, before the request context manager is closed.

api/api/views/media_views.py
```diff
@@ -17,6 +21,62 @@
...
+        # new providers are available in the API, and sometimes not even then.
+        # Search returns results with providers that do not have a ContentProvider
+        # table entry. Therefore, to maintain that assumption, a subquery is the only
+        # workable approach, as Django's `extra` does not provide any facility for
+        # handling null relations on the join.
+        return self.model_class.objects.exclude(
+            provider__in=ContentProvider.objects.filter(
+                filter_content=True
+            ).values_list("provider_identifier")
+        )
```
sarayourfriend, line 54: This is duplicated from the existing media view. We can abstract it into a mixin if we want to avoid the duplication. I'd prefer to keep them separate in case we need to modify this for async context for some reason.

api/api/views/media_views.py
```diff
@@ -17,6 +21,59 @@
...
+            media_obj.identifier,
+            accept_header=request.headers.get("Accept", "image/*"),
+            **serializer.validated_data,
+        )
+
+    @abc.abstractmethod
+    async def thumbnail(self, request):
+        raise NotImplementedError(
+            "AsyncMediaView subclasses must implement `thumbnail`"
+        )
```
sarayourfriend, line 74: I've slightly changed the design for this route to more strongly convey that the `thumbnail` method is _not_ sufficient for handling a request on its own. That should already have been clear from the method's parameter list but this makes it even more clear.

api/conf/settings/base.py
```diff
@@ -60,6 +61,11 @@
     },
 ]
 
+# Allow both during transition. We need to deploy this with zero-downtime, so we can
+# deploy the WSGI app at first and then switch the configuration over to use uvicorn
+# and the ASGI app after it is deployed. In the mean time, async routes will still be
+# "async" just inside their own event loop per-request.
+ASGI_APPLICATION = "conf.asgi.application"
```
sarayourfriend, line 68: After we deploy the API using gunicorn + uvicorn to use the ASGI application, we'll be able to delete the WSGI_APPLICATION setting and the wsgi.py file.

api/conf/settings/static.py
```diff
@@ -1,10 +1,13 @@
-from decouple import config
+"""
```
sarayourfriend, line 1: See the note in `Dockerfile` for explanation of this change.

api/conf/urls/__init__.py
```diff
@@ -35,3 +39,6 @@
     path("healthcheck/", HealthCheck.as_view(), name="health"),
     path("v1/", include(versioned_paths)),
 ]
+
+if settings.ENVIRONMENT == "local":
+    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```
sarayourfriend, line 44: This is how `runserver` serves static files. Because we're using uvicorn for the local server now, we need to replicate that functionality of runserver.

api/env.docker
```diff
@@ -5,8 +5,8 @@ DJANGO_SECRET_KEY="ny#b__$$f6ry4wy8oxre97&-68u_0lk3gw(z=d40_dxey3zw0v1"
 DJANGO_DEBUG_ENABLED=True
 
 BASE_URL=http://localhost:50280/
-ENVIRONMENT=development
-ALLOWED_HOSTS=api.openverse.engineering,api-dev.openverse.engineering,host.docker.internal
+ENVIRONMENT=local
```
sarayourfriend, line 8: We use `local` elsewhere and it's a bit clearer than `development`, which has been historically muddled with `staging`.

api/env.template
```diff
@@ -4,11 +4,6 @@ DJANGO_SETTINGS_MODULE=conf.settings
 DJANGO_SECRET_KEY="ny#b__$$f6ry4wy8oxre97&-68u_0lk3gw(z=d40_dxey3zw0v1"
 DJANGO_DEBUG_ENABLED=True
 
-BASE_URL=http://localhost:50280/
-ENVIRONMENT=development
-# List of comma-separated hosts/domain names, e.g., 127.17.0.1,local.app
-ALLOWED_HOSTS=localhost,172.17.0.1,host.docker.internal
```
sarayourfriend, line 10: No need to duplicate these between `env.docker`. These environment files are now exclusively used for local development, so we don't need to do anything to make sure it's easy to handle live vs local deployments of it.

api/test/unit/utils/test_image_proxy.py
```diff
@@ -33,6 +33,16 @@
...
 
 
+@pytest.fixture
+def photon_url():
+    return f"{settings.PHOTON_ENDPOINT}subdomain.example.com/path_part1/part2/{uuid4()}.jpg"
+
+
+@pytest.fixture
+def image_url(photon_url):
+    return photon_url.replace(settings.PHOTON_ENDPOINT, "http://")
```
sarayourfriend, line 43: Prevent duplicated image URLs.

api/test/unit/utils/test_image_proxy.py
```diff
@@ -219,30 +265,33 @@ def test_get_successful_forward_query_params(mock_image_data):
...
     def do(exc):
-        def raise_exc(*args, **kwargs):
-            raise exc
+        mock_session = MagicMock(spec=aiohttp.ClientSession)
+        mock_session.get.side_effect = exc
 
-        monkeypatch.setattr("requests.get", raise_exc)
+        monkeypatch.setattr(
+            "api.utils.image_proxy.get_aiohttp_session", lambda: mock_session
+        )
```
sarayourfriend, line 285: We don't use `requests` anymore for this, so we need to mock out the aiohttp session instead.

api/test/unit/utils/test_image_proxy.py
```diff
@@ -446,13 +499,19 @@ def test_photon_get_raises_by_not_allowed_types(image_type):
...
+    pook.on()
+    mock_head = pook.head(image_url).reply(200)
+    if headers:
+        mock_head = mock_head.headers(headers)
 
-    with mock.patch("requests.head") as mock_head, pytest.raises(UnsupportedMediaType):
-        mock_head.return_value.headers = headers
-        photon_get(image_url, image.identifier)
+    with pytest.raises(UnsupportedMediaType):
+        await photon_get(image_url, image.identifier)
```
sarayourfriend, line 514: Using pook instead of mocking aiohttp is a more reliable way to prevent the head request from going through.

sarayourfriend:
Something is happening in my implementation that causes the event loop associated with the aiohttp ClientSession to close after each thumbnail request. I can't figure out what is causing this. It can be mitigated by checking if the loop on the session is closed before reusing it (and creating a new one if needed), but the `.loop` property on the session is deprecated, so this raises a warning, and in future versions of aiohttp the property will be removed entirely.

I've tried to figure out what it is I did that is causing this behaviour. There's virtually nothing available online that describes the issue. It is only reproducible when using the production runtime configuration (gunicorn with uvicorn workers). When running directly under uvicorn, there's no issue at all.

Using httpx instead of aiohttp doesn't solve the error. I get PoolTimeouts instead of event loop closure errors. This means it isn't something to do with aiohttp, but the implementation of the thumbnail route. I _suspect_ it might have something to do with `sync_to_async` wrappers around Redis, but I just don't know.

If I switch the implementation _back_ to using `django-async-redis`, I only get event loop closure errors on every other request (as opposed to being able to serve a single request followed by only getting event loop closure errors on every subsequent request).

Anyway, I'm not able to continue working on this due to AFK. I just wanted to share where I got with this project and the feasibility.

I'm of the opinion that it is absolutely feasible and that we just need to identify the underlying issue that is causing the event loop closures. All routes other than the thumbnail continues to work perfectly fine, which leads me to believe it really is something caused by the implementation of the image proxy and small asyncio usage details that I'm not aware of. The requirement of needing to share the HTTP library session is a pain, but the implementation for it does work, theoretically!

Maybe it's worth just doing everything in FastAPI, maybe even re-writing our entire app in FastAPI. Who knows. I think the Django approach would be the best one, but getting stuck on this event loop closed issue has left me confounded. Chances are that the same implementation would have all the same problems if it was running on FastAPI instead of Django... I really don't know what the effective difference would be there!

Anyway, if someone else wants to try to debug the issue, just override the `command` in the docker-compose for `web` with `gunicorn --reload` and the gunicorn configuration will pick up the rest. It is only reproducible when running multiple workers, so maybe there's something thread-unsafe being done in the code that is causing this? :shrug:

sarayourfriend:
Also, just a note, I don't _think_ the issue is caused by splitting the views, because all other views work perfectly fine. It's just the thumbnails route that has issues. I really think it's down to the implementation of the thumbnails route, something funky going on there with the event loops.

sarayourfriend:
Okay, the core maintainers have agreed that feasibility has been demonstrated by this PR, so I'm going to close both the PR and the issue. I'll ping on the issue to get clarification on the precise next steps.

Issue #1969, opened 2023-05-02
Implementation Plan: Fine-grained API response cache management


sarayourfriend:
## Description

Project proposal: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#cache-management-requirement-5

Please refer to the project proposal section linked above for the requirements of this plan.

To reiterate what is said in the proposal, the plan must include a list of every single view that returns content with a specific determination for whether cache management can be done via Cloudflare's tools or if we need to implement the new Redis cache strategy for the view.

sarayourfriend:
This part of the implementation plan needs a significant re-thinking in light of the discussion: #2453.

The work proposed for this plan is significant and complex. If we can avoid it by reducing the amount of edge caching we rely on, it would save us a lot of process and implementation work.

Closing this _for now_ and I will revisit this plan next week with a proposed alternative for us to discuss and work though the decision making process with.

PR #2676, opened 2023-07-19
Implementation Plan: Additional search views

### LINKED ISSUE #2560: Implementation Plan: Additional search views for the frontend
## Problem

The **Additional search views for the frontend** (#410) project requires an [Implementation Plan](https://docs.openverse.org/projects/planning.html#implementation-status-in-progress) for the frontend and analytics changes. We suspect the analytics changes are lightweight enough that this can be a single plan.

## Description

For the frontend, we need to make sure we:

- Add any additional analytics events necessary for this update
- Describe how we will implement:
  - The addition of new Nuxt routes for the landing pages
  - Updates to the single result pages 
  - New icons and components (for example, the scrollable area for the author and source buttons
  - New tests which will need to be written

obulat:
## Fixes

Resolves #2560 by @zackkrida

## Assigned reviewers
- [x] @zackkrida
- [ ] @sarayourfriend

## Note
A very rough draft implementation of the additional search view pages is in branch [draft/collection_pages](https://github.com/WordPress/openverse/tree/draft/collection_pages).

## Current round

> **Note**
> This discussion is following the Openverse decision-making process. Information about this process can be found
on the [Openverse documentation site](https://docs.openverse.org/meta/decision_making/index.html). Requested reviewers or participants will be following this process. If you are being asked to give input on a specific detail, you do not need to familiarise yourself with the process and follow it.

The discussion is currently in the [Decision round](https://docs.openverse.org/meta/decision_making/process_description.html#decision-round).

The deadline for review of this round is **2023-07-25**.

### DIFF
--- a/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
+++ b/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
@@ -1,4 +1,4 @@
-2023-07-20 Implementation Plan: Additional Search Views
+# 2023-07-20 Implementation Plan: Additional Search Views
 
 **Author**: @obulat
 
@@ -34,14 +34,122 @@ Three collection pages are added to Openverse:
 
 - items with the selected tag
 - items from the selected source (provider)
-- items by the selected creator. The pages re-use the existing image grid and
-  audio collection views and can load more images or audio items on the Load
-  more button click. The single result pages are updated to add the links to
-  these pages, as seen in the Figma mockups: [link to Figma mockup]
+- items by the selected creator.
+
+The pages re-use the existing image grid and audio collection views and can load
+more images or audio items on the Load more button click. The single result
+pages are updated to add the links to these pages, as seen in the Figma mockups:
+
+- [Image single result](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-63284&mode=dev)
+- [Audio single result](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-63285&mode=dev)
+
+For the creator pages, a separate API media view is added to get the items that
+match the creator and source pair exactly.
 
 ## Outlined Steps
 
 <!-- Describe the implementation step necessary for completion. -->
+
+### API changes
+
+Currently, when filtering the search results, the API matches some query
+parameters in a fuzzy way: an item matches the query if the field value contains
+the query string as a separate word. When indexing the items, we "analyze" them,
+which means that we split the field values by whitespace and stem them. We also
+do the same to the query string. This means that the query string "bike" will
+match the field value "bikes", "biking", or "bike shop".
+
+For most of these pages, we need an exact match instead. This section will
+describe each page and the necessary changes.
+
+#### Source page
+
+To get the items from the selected source, we can use the `source` query
+parameter: https://api.openverse.engineering/v1/images/?source=met
+
+There is a limited number of sources, and we validate the `source` parameter
+against the full list. This means that the `source` parameter will be matched
+exactly.
+
+#### Creator page
+
+We could update the Elasticsearch index and the search controller to allow for
+exact matching by the `creator` field. However, I think there is an easier way
+that does not require any changes to the indexes: use the Django ORM to filter
+the items by the creator name and the provider name.
+
+We can add a media view at
+`/api/v1/<media>/creator/?q=<creator_name>&provider=<provider_name>`. It would
+use `self.get_queryset().filter(creator=creator, provider=provider)` to get all
+the items by the creator from the provider. This view would also need to filter
+out dead links, and return a paginated response.
+
+#### Tag page
+
+To get the items matching a tag, we can use the `tags` query parameter with a
+quoted value.
+https://api.openverse.engineering/v1/images/?tags=%22black%20cat%22 This query
+returns all images with tags that contain the phrase "black cat".
+
+This would not match the tags exactly, but tags page is supposed to show all
+items that are associated with the same topic or category.
+
+We could also add a media view at `/api/v1/<media>/tags/?q=<tag_name>`, but it
+is difficult to implement because the tags are stored as a list JSONField model.
+
+### Nuxt store changes
+
+We can re-use the search store as is for these pages. Currently, the store uses
+the `searchBy` filter to determine the shape of the API query. If `searchBy`
+value is set, then the `q` parameter is replaced with the
+`<searchBy>=<searchTerm>` query parameter.
+https://github.com/WordPress/openverse/blob/b4b46b903731870015c475d2c08eebef7ec6b25b/frontend/src/utils/prepare-search-query-params.ts#L22-L25
+
+We can update the `searchBy` filter to have one of the possible values:
+`{ searchBy: <null|creator|source|tag> }`. Then, this value would be used to
+construct the API query.
+
+#### Update the `searchBy` filter
+
+Currently, this filter is shaped as other filters: it is a list of objects with
+`code`, `name` and `checked` properties. This is a possible value now: `{
+searchBy: { code: "creator", name: "filters.searchBy.creator", checked: true }}.
+
+Since this filter is mutually exclusive (you cannot search by both the creator
+and the tag, for example), this shape is very difficult to update.
+
+It is better to set the `searchBy` filter to have one of the possible values:
+`{ searchBy: <null|creator|source|tag> }`.
+
+We should create a new method, `setSearchBy`, in the `search` store, that would
+allow directly set the `searchBy` filter value.
+
+This change should also update `prepare-search-query-params` to use the
+`searchBy` filter value to construct the API query:
+
+```typescript
+const prepareSearchQuery = (
+  searchParams: Record<string, string>,
+  mediaType: SupportedMediaType
+) => {
+  if (searchParams.searchBy) {
+    if (searchBy === "creator") {
+      // the name of the source is stored in the `<mediaType>Provider` filter
+      return {
+        creator: searchTerm,
+        source: searchParams[`${mediaType}Provider`],
+      }
+    } else if (searchBy === "source") {
+      return { source: searchParams[`${mediaType}Provider`] }
+    } else if (searchBy === "tag") {
+      // The parameter is plural in the API
+      return { tags: searchTerm }
+    }
+  } else {
+    // Current search query transform
+  }
+}
+```
 
 ### Page changes
 
@@ -56,16 +164,21 @@ the actual value of the tag, creator or source name.
 
 This page should use the
 [`validate` method](https://v2.nuxt.com/docs/components-glossary/validate/) to
-make sure that the `collection`, `searchType` and `term` `params` are valid.
+make sure that the `collection`, `mediaType` and `term` `params` are valid.
 
 ```
-`validate({ params, $pinia }): {
-const { collection, searchType, term } = params
+validate({ params, $pinia }): {
+    const { collection, mediaType, term } = params
+    // Check that collection is one of ["tag", "creator" or "source"],
+    // and mediaType is one of `supportedMediaTypes`.
+    // Check that `term` is correctly escaped.
+    // If the params are not valid, return `false` to show the error page.
+    return isValid ? true : false
 ```
 
 This page should also update the state (`searchType`, `searchTerm` and
-`searchBy` filter) in the `search` store and handle fetching using
-`mediaStore`'s `fetchMedia` method in the `useFetch` hook. Since it is not
+`searchBy` and `provider` filters) in the `search` store and handle fetching
+using `mediaStore`'s `fetchMedia` method in the `useFetch` hook. Since it is not
 possible to change the path or query parameters from this page client-side,
 fetching can be much simpler than on the current search page (that has to watch
 for changes in the route and fetch if necessary).
@@ -90,12 +203,19 @@ different search views.
 #### Add a `VCollectionHeader` component
 
 Figma links:
-[creator desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56323&mode=design&t=pN0PPAzlKQEKT9tJ-4),
-[creator mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56362&mode=design&t=suLIyJHNmZrM0mPH-4),
-[source desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56381&mode=design&t=suLIyJHNmZrM0mPH-4),
-[source mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56421&mode=design&t=suLIyJHNmZrM0mPH-4),
-[tag desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4),
-[tag mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)
+
+- creator
+  [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56323&mode=design&t=pN0PPAzlKQEKT9tJ-4)
+  and
+  [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56362&mode=design&t=suLIyJHNmZrM0mPH-4)
+- source
+  [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56381&mode=design&t=suLIyJHNmZrM0mPH-4)
+  and
+  [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56421&mode=design&t=suLIyJHNmZrM0mPH-4)
+- tag
+  [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4)
+  and
+  [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)
 
 The header should have an icon (tag, creator or source), the name of the
 tag/creator/source. For source and the creator, there should be an external link
@@ -105,19 +225,21 @@ The header should also display the number of results, "251 audio files with the
 selected tag", "604 images provided by this source", "37 images by this creator
 in Hirshhorn Museum and Sculpture Garden".
 
-#### Add VCreatorLink component
+Note: There are sources that only have works by one creator. In this case, we
+should probably still have two separate pages for the source and the creator,
+but we might want to add a note that the creator is the only one associated with
+this source.
+
+#### Add `VCollectionLink` component and update the title and links container
 
 The "by creator" line under the main item on the single result page should be
-replaced with a `VCreatorLink` component - a wrapper over `VButton` with a link
-to the localized "creator" collection page.
+replaced with a section that has two buttons: one for a creator link and a
+source link. These links should be `VButtons` with `as="VLink"` and should link
+to the localized page for the creator or the source pages. This section should
+be horizontally scrollable on mobile. It should implement a scroll-snap
+(example: https://play.tailwindcss.com/AbfA33Za50)
 
-#### Add VSourceLink component
-
-Both the source and the provider should be replaced with `VSourceLink` component
-that links to the `source` page. This component is also a wrapper over `VButton`
-component.
-
-#### Update VMediaTag component to be a `VButton` wrapping a `VLink`
+#### Update `VMediaTag` component to be a `VButton` wrapping a `VLink`
 
 Each `VMediaTag` should be a `VButton` with `as="VLink"` and should link to the
 localized page for the tag collection. Figma link:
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Expected Outcomes
+
+<!-- List any succinct expected products from this implementation plan. -->
+
+Three collection pages are added to Openverse:
+
+- items with the selected tag
+- items from the selected source (provider)
+- items by the selected creator. The pages re-use the existing image grid and
```
zackkrida, line 37:
```suggestion
- items by the selected creator. 

These pages re-use the existing image grid and
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+  audio collection views and can load more images or audio items on the Load
+  more button click. The single result pages are updated to add the links to
+  these pages, as seen in the Figma mockups: [link to Figma mockup]
```
zackkrida, line 40:
```suggestion
  these pages, as seen in the Figma mockups:
  
- [Image single result](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-63284&mode=dev)
- [Audio single result](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-63285&mode=dev)
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+### Page changes
+
+#### Create a page for `tag` / `creator` /`source` collections.
```
zackkrida, line 48: This section is excellent! This should be a great approach

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+#### Extract the `VAudioCollection` component
+
+Currently, it is not possible to reuse the audio collection from the audio
+search result page because it is a part of the `audio.vue` page. We should
+extract the part that shows the loading skeleton, the column of `VAudioTrack`
+rows and the Load more section. This component will be reused in the audio
+search page and on the Additional search views. One thing that I'm not sure
+about here is whether the analytics events should stay the same and be sent from
+all pages - or if we should somehow differentiate the events coming from
+different search views.
```
zackkrida, line 88: > One thing that I'm not sure
about here is whether the analytics events should stay the same and be sent from
all pages - or if we should somehow differentiate the events coming from
different search views.

These will be naturally differentiated by the built-in `pathname` property of all analytics events. They can remain the same.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+#### Add a `VCollectionHeader` component
+
+Figma links:
+[creator desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56323&mode=design&t=pN0PPAzlKQEKT9tJ-4),
+[creator mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56362&mode=design&t=suLIyJHNmZrM0mPH-4),
+[source desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56381&mode=design&t=suLIyJHNmZrM0mPH-4),
+[source mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56421&mode=design&t=suLIyJHNmZrM0mPH-4),
+[tag desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4),
+[tag mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)
```
zackkrida, line 98:
```suggestion

- [creator desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56323&mode=design&t=pN0PPAzlKQEKT9tJ-4),
- [creator mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56362&mode=design&t=suLIyJHNmZrM0mPH-4),
- [source desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56381&mode=design&t=suLIyJHNmZrM0mPH-4),
- [source mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56421&mode=design&t=suLIyJHNmZrM0mPH-4),
- [tag desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4),
- [tag mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+`<VButton variant="filled-gray" size="small" class="label-bold" as="VLink" href="href">{{tag}}</VButton>`.
+It should accept `href` as a prop.
+
+### Other interface changes
+
+#### Update the "by" area of the single result page
+
+The "by creator" line should be replaced with a section that is
+horizontally-scrollable on mobile, and contains the `VCreatorLink` and
+`VSourceLink` components.
```
zackkrida, line 135: I'd recommend we implement scroll-snap here, as a nice simple enhancement for narrow desktop viewports but also the mobile experience. Example here: 

https://play.tailwindcss.com/AbfA33Za50

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+Audio creator link should also be updated to match the image creator link. It
+should be a conditional component: `VLink` if the `creator_url` is truthy and
+`span` if the `creator_url` is falsy.
+
+Currently, the `foreign_landing_url` is linked to the "source" in the image page
+and "provider" in the audio page. The audio page should be updated to match the
+image page: the `foreign_landing_url` link should be added to the "source", not
+provider.
+[Audio provider link](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/components/VAudioDetails/VAudioDetails.vue#L61C1-L63),
+[image source link](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/components/VImageDetails/VImageDetails.vue#L26-L28)
```
zackkrida, line 154: This section just needs a bit of formatting, and cleanup it's a little hectic.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+We should also create a new method, `setSearchBy`, in the `search` store, that
+would allow directly set the `searchBy` filter value.
+
+### Analytics changes
+
+I am not sure what events we want to add, if any. Should we add `<VIEW>_CLICKED`
+events for the new views, or would page views be sufficient?
+
+Should we add a new `SELECT_COLLECTION_ITEM` event similar to
+`SELECT_SEARCH_RESULT`?
```
zackkrida, line 179: I personally think page views will be sufficient. The only way to access the pages is directly or via links on the single results, which will all be captured by standard page visits.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+The clicks on external creator link in the `VCollectionHeader` should be tracked
+as `VISIT_CREATOR_LINK` events. We don't have a special event for visiting
+source, should we add one, or should they be tracked as generic
+`VISIT_EXTERNAL_LINK`?
```
zackkrida, line 184: It makes sense to me to add an event for these because we might want to know how these links are used broadly ("how many provider link clicks are there?") rather than specifically ("how many clicks to https://flickr.com are there?").

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Tests
+
+We should add visual-regression tests for the new views. To minimize flakiness
+due to slow loading of the images, we should probably use the
+`{ filter: brightness(0%); }` trick for the images on the page.
```
zackkrida, line 190: Are we using this anywhere presently?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Tools & packages
+
+<!-- Describe any tools or packages which this work might be dependent on. If multiple options are available, try to list as many as are reasonable with your own recommendation. -->
+
+No new tools or packages are necessary.
+
+### Other projects or work
+
+<!-- Note any projects this plan is dependent on. -->
```
zackkrida, line 211:
```suggestion
### Other projects or work

<!-- Note any projects this plan is dependent on. -->
Not applicable.
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+## Parallelizable streams
+
+<!-- What, if any, work within this plan can be parallelized? -->
+
+The work on components (`VCreatorLink`, `VSourceLink`, `VMediaTag`), and on the
+search store changes can be parallelized.
+
+We should create a separate branch for this work because adding the links to the
+collection pages on the single result pages (`VCreatorLink`, `VSourceLink`,
+`VMediaTag`) before the collection pages are ready is not possible.
```
zackkrida, line 229: Should it be feature-flagged instead, Olga? I would think so. If you agree please update the plan to reflect that.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+We should make sure that the search titles are accessible, and the pages clearly
+indicate the change of context.
+
+## Rollback
+
+<!-- How do we roll back this solution in the event of failure? Are there any steps that can not easily be rolled back? -->
+
+If we work on a separate branch, it would be easy to rollback by reverting the
+branch merge commit.
```
zackkrida, line 251: Once again I'd suggest a feature flag rather than a branch-based approach.

zackkrida (changes requested):
This is excellent, I'm only requesting changes based on the volume of cleanup suggestions I've made.

Actually, one potential blocker. I believe we need to include the necessary API changes in this plan, rather than have a separate plan for the API. Since the API changes are so small.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+### Page changes
+
+#### Create a page for `tag` / `creator` /`source` collections.
+
+Nuxt allows creating nested dynamic routes like
+`/pages/_collection/_mediaType/_term`.
+
+Here, the collection can be `tag`, `creator` or `source`. `mediaType` can be one
+of the supported media types (currently, `image` and `audio`). `term` refers to
+the actual value of the tag, creator or source name.
```
sarayourfriend, line 55: That's fantastic! It could be useful for SEO as well. Speaking of which, what considerations do we need to apply to `robots.txt` for this change, if any? I don't really understand our current approach to robots.txt, is it leftover from when we were iframed? @zackkrida I believe you may know.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+
+This page should use the
+[`validate` method](https://v2.nuxt.com/docs/components-glossary/validate/) to
+make sure that the `collection`, `searchType` and `term` `params` are valid.
+
+```
+`validate({ params, $pinia }): {
```
sarayourfriend, line 62: Just removing a stray `` ` ``:

```suggestion
validate({ params, $pinia }): {
```

Is this code sample missing a portion of the intended sample?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+[tag desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4),
+[tag mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)
+
+The header should have an icon (tag, creator or source), the name of the
+tag/creator/source. For source and the creator, there should be an external link
+button.
+
+The header should also display the number of results, "251 audio files with the
+selected tag", "604 images provided by this source", "37 images by this creator
+in Hirshhorn Museum and Sculpture Garden".
```
sarayourfriend, line 106: Is it at all relevant for us to consider sources that have a single creator? Boston Public Library, as a source, for example, only has Boston Public Library as the "creator" of the work. It's weird to think that there would be two separate pages for them, but then I'm not sure how we would communicate this to the user if we automatically redirected the BPL creator search to the BPL source search :thinking:

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+searchBy: { code: "creator", name: "filters.searchBy.creator", checked: true }}.
+
+Since this filter is mutually exclusive (you cannot search by both the creator
+and the tag, for example), this shape is very difficult to update.
+
+It is better to set the `searchBy` filter to have one of the possible values:
+`{ searchBy: <null|creator|source|tag> }`.
+
+We should also create a new method, `setSearchBy`, in the `search` store, that
+would allow directly set the `searchBy` filter value.
```
sarayourfriend, line 171: Can you clarify whether this is just a DevEx/code-quality refactor or if it's necessary for the new features? I don't fully understand why this particular change is relevant to the creator/tag/source pages. Are they going to use the same search store and filters? If so, that would be helpful background to have somewhere as well (may have missed it though?).

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
sarayourfriend: Agreed. `SELECT_SEARCH_RESULT` can be narrowed to `/search` or `/tag*` (for example) to determine where the event occurred'

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
sarayourfriend: Strongly agree on feature-flagging rather than feature-branching whenever possible.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,263 @@
...
+If we work on a separate branch, it would be easy to rollback by reverting the
+branch merge commit.
+
+## Risks
+
+<!-- What risks are we taking with this solution? Are there risks that once taken can’t be undone?-->
+
+The biggest risk I see is that this project might be seen as an "invitation" to
+scraping. Hopefully, the work on providing the dataset would minimize such
+risks.
```
sarayourfriend, line 259: Frontend rate limiting will also help.

sarayourfriend (commented):
I agree with Zack, but will suggest that either the API changes should be included in this IP or they should be completed first.

Without having a deep knowledge of how the frontend stores work, and because it isn't described in this implementation plan, it's very hard to understand how the searches for these collection pages is meant to be implemented.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Yes, I actually re-used the same search store in the draft implementation that I linked in the issue description. But to get the store to handle different `searchBy` values, we should add this change.
Thank you for asking this question! I realized that I didn't write anything about the data side: what API calls will be made, how the stores will be used and how they need to be updated. I'll add this during the revision round.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
zackkrida: Our current robots.txt is blank, which allows all pages of the site to be crawled and indexed. That works exactly the same as:

```
User-agent: *
Disallow:
```


The staging site is blocked from indexing. 

I don't think we need to do anything here. These pages can be crawled and indexed.

sarayourfriend: These new pages are essentially search views with static URLs though (not query params). Should that kind of thing be indexed?

Sorry if this is already discussed somewhere. Might be helpful to document the rationale of our robots.txt in the docs site.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Link added.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Added a note to the plan. By the way, we don't have a Boston Public Library as a source filter on the frontend. What is the link to their API or collection?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Added the note on rate limiting.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,366 @@
...
+- [Audio single result](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-63285&mode=dev)
+
+For the creator pages, a separate API media view is added to get the items that
+match the creator and source pair exactly.
+
+## Outlined Steps
+
+<!-- Describe the implementation step necessary for completion. -->
+
+### API changes
```
sarayourfriend, line 53: Thank you for adding this section. It helps a lot to understand how this will all fit together.

I like the approach to add new routes on the API rather than trying to stuff everything into the existing search endpoint. That makes things a lot nicer for the creator and provider especially.

However, there are some problems with using the database rather than ES to access anything:

1. The database does not cache queries in the same way that ES does. Repeated queries will not necessarily be as efficient as from ES.
2. The database does not score documents at all, so the order will different dramatically to the way that ES would order the documents. That's an issue with respect to popularity data today already, but will become even more of an issue if we start to score documents based on other metrics as theorised by our search relevancy discussions.
3. `creator` is not indexed in the API database, so a query against it will be very slow.

Only one of these can be "fixed", the third, but it's much easier to add a `creator.raw` field to the index instead, like we've done for title, description, and tag name. Adding a new field to the index is far easier than adding a new index to the database (which would require careful application to avoid a long-running migration) and which could significantly increase the time it takes to do a data refresh.

I'd also strongly recommend against the route described as it is harder to cache in a sensible way (on the Cloudflare side of things) or to perform cache invalidation that might be required by #1969. Instead of using query strings, we can describe the resource via the path: `/<media type>/source/<source>/creator/<creator>` is very clean, easy to read and understand, and very easy to manage the cache for because it is a static path. The source page can use the same route by leaving off the creator. This removes the need to manage specific query params as well and would allow us to add querying within these routes more easily in the future behind the regular `q` parameter if we wanted.

For the tag route, I'd recommend the singular `tag` rather than plural `tags`, for legibility if we are presenting a single tag. Again, using a route-based approach instead of query parameter approach makes it easier to manage the cache if needed and could allow us to more easily feature specific tags with high quality results as example pages if we wanted, like `/images/tag/dog` or `/images/tag/flying%20birds`.

Path elements still need to be URL encoded to preserve special characters and spaces, but for most cases like `/images/source/nasa`, `/images/sources/nypl`, `/images/sources/inaturalist`, etc we've suddenly got a very good-looking URL to point people to.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,366 @@
...
+    } else if (searchBy === "source") {
+      return { source: searchParams[`${mediaType}Provider`] }
+    } else if (searchBy === "tag") {
+      // The parameter is plural in the API
+      return { tags: searchTerm }
+    }
+  } else {
+    // Current search query transform
+  }
+}
```
sarayourfriend, line 149: Just noting that regardless of whether we use my fully route based approach above, we'll still need to determine the path pattern to use. Should this function also handle that, maybe returning either an array of path segments (including holes for parameters and a final object for query parameters if still needed) or the query parameter object to send to the default `search` route?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,366 @@
...
+### Page changes
+
+#### Create a page for `tag` / `creator` /`source` collections.
+
+Nuxt allows creating nested dynamic routes like
+`/pages/_collection/_mediaType/_term`.
+
+Here, the collection can be `tag`, `creator` or `source`. `mediaType` can be one
+of the supported media types (currently, `image` and `audio`). `term` refers to
+the actual value of the tag, creator or source name.
```
sarayourfriend, line 161: Realising now that for these to work, creator needs to have source included. It could make sense to make these match the API routes: `/_mediaType/source/_source/creator/_creator`, `/_mediaType/tag/_tag` etc.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
sarayourfriend: Oops, I've just realised that BPL is just a creator we've picked up through Flickr, not an actual source. Maybe an easy source to add then!

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,366 @@
...
+
+Not applicable.
+
+## Design
+
+<!-- Note any design requirements for this plan. -->
+
+[Figma designs in the dev mode](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?node-id=22%3A13656&mode=dev)
+
+## Parallelizable streams
```
sarayourfriend, line 325: Should this be updated to include the API work in some way?

sarayourfriend (changes requested):
I have only three truly blocking issues, the rest are just additional questions or suggestions. The blockers are:

1. The various issues I've described with using Postgres to query for the creator view must be addressed. I've suggested we continue to use ES by adding new `.raw` sub-fields for creator and source.
2. The frontend routes need to disambiguate creators by source. I've suggested alternative paths that mirror my suggested path (rather than query param) based approach for the API and would solve this problem.
3. For me to fully understand _how_ the plan would be implemented, I need an ordered, high-level step-by-step implementation of all the tasks that would be converted into individual issues. I think we should clarify this in the implementation plan template to have something like `ordered steps` and `step descriptions` so that it's possible to get both a high-level view of the plan separated from the detailed descriptions and nuances of each step.

Overall the plan looks very good though. The first two issues are critical blockers for the plan because the plan will not succeed without addressing them. The last is just a personal blocker for me to get a proper understanding of the plan.

I'm starting AFK tomorrow and am confident that you and Zack will be able to resolve the blockers without me. The plan otherwise is great. Please consider this plan approved by me once those blockers are resolved :rocket:

obulat:
I've updated the plan to add the API and Elasticsearch changes. In particular, I updated the paths to match @sarayourfriend's suggestions. I also rewrote the plan to make the steps descriptions and steps sequence clearer.
@zackkrida, this is ready for re-review.

zackkrida:
I'll review this today!

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,429 @@
...
+not dependent on anything.
+
+The work on the single result pages (step 7) can be done in parallel with the
+work on the collection pages (step 6), but should follow the previous steps.
+
+## Blockers
+
+<!-- What hard blockers exist which might prevent further work on this project? -->
+
+The main blocker could be the maintainer capacity.
```
zackkrida, line 404: This is always a potential blocker, so I don't think we need to explicitly mention it.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -0,0 +1,429 @@
...
+**Author**: @obulat
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @zackkrida
```
zackkrida, line 12:
```suggestion
- [x] @zackkrida
```

zackkrida:
We're going to merge this on the basis of @sarayourfriend's blockers being resolved 🥳

PR #2695, opened 2023-07-21
Extract reusable VAudioCollection.vue

### LINKED ISSUE #2694: Extract `VAudioCollection` component
## Problem

We show a list of audio results on the search page and in the single result page Related audio section. Soon, we will also use it in the Additional search views. This means a lot of code duplication.

## Description

We should extract a reusable component that displays a list of audio results as a column of `VAudioTrack` rows. It should also show a loading skeleton when fetching the media, show the Load more button if necessary, should have accessible label for the ordered list, and should handle the analytics events.

## Alternatives

We could create a new component for the Additional search views. This would be the third place where we have similar functionality - which is means more code and prone to errors.

obulat:
## Fixes

Fixes #2694 by @obulat

## Description

This PR extracts two components: 

`VAudioCollection` handles the display of the loading skeleton, audio tracks and/or Load more button.
`VAudioList` displays the ordered list of audio tracks, adds a label for screen readers, and also handles the analytics events.

The audio page and the Related audio sessions now use `VAudioCollection` as child components.

This PR also removed the `search-props` file because it wasn't easy to see what props were passed to the pages, which also hid the fact that we weren't using the props passed by the parent component.

## Testing Instructions

The CI should pass, and the Audio track interactions should send correct events on play/pause/seek and on clicking on the Audio track.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VSearchResultsGrid/VAudioList.vue
```diff
@@ -0,0 +1,112 @@
...
+        layout="row"
+        :search-term="searchTerm"
+        @interacted="handleInteraction"
+        @mousedown="handleMousedown(audio, $event)"
+        @focus="$emit('focus')"
+      />
+    </li>
+  </ol>
+</template>
+<script lang="ts">
```
zackkrida, line 25:
```suggestion
</template>

<script lang="ts">
```

nit

frontend/src/components/VSnackbar.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div
-    role="alert"
+    :role="isVisible ? 'alert' : undefined"
```
zackkrida, line 3: unrelated, but seems fine

zackkrida (approved):
LGTM! This abstraction is better than what we have. It makes me want to rename `VImageGrid` to `VImageCollection`. And then I think about how both of _those_ components have nearly the same shape, and there's probably an even simpler abstraction for our media views out there somewhere, waiting to be discovered 😆

frontend/src/components/VSnackbar.vue
obulat: Agree that it's unrelated, but it's been bugging me every time I test using screenreader: when you search from the homepage, you hear the snackbar read outloud, even if it shouldn't be shown.

frontend/src/composables/use-audio-snackbar.ts
```diff
@@ -0,0 +1,40 @@
...
+  const snackbar = {
+    isVisible: computed(() => useUiStore().areInstructionsVisible),
+    show: showSnackbar,
+    hide: hideSnackbar,
+    handleMouseDown,
+  }
+
+  return {
+    snackbar,
+  }
```
zackkrida, line 39: Why not remove one object and just return the snackbar instead, so the usage becomes

`const snackbar = useAudioSnackbar()` or `cost { isVisible } = useSnackbar()`? 

I think that pattern better matches our existing composables.

frontend/src/types/events.ts
```diff
@@ -0,0 +1 @@
+export type AudioTrackClickEvent = { event: MouseEvent; inWaveform: boolean }
```
dhruvkb, line 1: Nice pattern.

frontend/src/components/VSearchResultsGrid/VAudioList.vue
```diff
@@ -0,0 +1,128 @@
...
+  <!-- Negative margin compensates for the `p-4` padding in row layout. -->
+  <ol
+    :aria-label="collectionLabel"
+    class="-mx-2 flex flex-col md:-mx-4"
+    :class="{ 'gap-4': isRelated }"
+  >
+    <li
+      v-for="audio in results"
+      :key="audio.id"
+      :class="{ 'mb-2 md:mb-1': !isRelated }"
```
dhruvkb, line 11: Nit: This can be accomplished with a different gap on line L6 using `isRelated ? 'gap-4' : 'gap-2 md:gap-1'`.

dhruvkb (approved):
LGTM, I have a small nit but that's extremely minor and ignorable.

Issue #460, opened 2023-02-03
Update external sources modal styles


obulat:
## Problem

The external sources section needs to be more consistent with the rest of the text styles.

### Wide desktop
<img width="1077" alt="Screenshot 2023-02-03 at 9 10 20 AM" src="https://user-images.githubusercontent.com/15233243/216525733-b9b69400-db1a-4f48-ad01-77993ea00181.png">

### Mobile
<img width="372" alt="Screenshot 2023-02-03 at 9 12 32 AM" src="https://user-images.githubusercontent.com/15233243/216525992-df3da3af-93e8-413f-b5d0-38228dea98f9.png">

## Additional context
First reported by @panchovm

fcoveram:
Here is an idea.

## `XL` breakpoint

https://github.com/WordPress/openverse/assets/895819/870c56d5-7b10-48ef-92b0-512431ac036e

## `XS` breakpoint

https://github.com/WordPress/openverse/assets/895819/4055b256-0765-43e2-850d-37cef79fb8eb

---

The button triggers a popover in a modal in all breakpoints. The design spans the following changes:

### Metasearch button
* It has the same height as the [Load more](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=5173-16609&mode=design&t=JzKg0h3PdFx8XQ1v-4) in all breakpoints
* It follows the [Button](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=4577-11868&mode=design&t=JzKg0h3PdFx8XQ1v-4) style with `border: gray` and `type: icon right`.

### Popover in modal
Overall, the popover follows the [Modal](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?type=design&node-id=28524-24040&mode=design&t=ZV4igMACfeHQZBv2-4) implemented in Gutenberg.

* The popover is divided into three sections to make it reusable: Header, description, and content.
   * Header area has the popover title and close action.
   * Close button follows [Icon](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=4658-12470&mode=design&t=JzKg0h3PdFx8XQ1v-4) component in `size: small`, `style: transparent`, `color: gray`.
* The container has a border-radius of `4px` in all breakpoints.
[...]

Issue #2602, opened 2023-07-10
New styles in Audio control component


fcoveram:
## Description

As part of design changes introduced in #2113, the [Audio control](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=5215-31484&mode=dev) component in Design Library has new styles.

<img src="https://github.com/WordPress/openverse/assets/895819/c9a374ed-935e-4c19-bb2b-ab4783dfb926" width="650" alt="New audio control component with its variants and descriptive texts">

* `Small` size is removed due to its lack of use.
* `Tiny` is renamed as `Small`.
* `Extra large` is renamed as `Large`.
* New sizes for each variant
   * Small: 32px, `w-8` `h-8`
   * Medium: 48px, `w-12` `h-12`
   * Large: 64px, `w-16` `h-16`
* New icon sizes for each variant
   * Small: 24px, `w-6` `h-6`
   * Medium: 32px, `w-8` `h-8`
   * Large: 40px, `w-10` `h-10`
* Hover and Focus styles follow the Button style
* Default style has `2px` for border radius, although this changes depending on the component context.

## Context

The new styles are based on #415 implementation

## Figma

[Component](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=5215-31484&mode=dev) in Design Library.

obulat:
This issue is blocked because the design for single result page for Audio also needs to be updated to match the new Audio control component sizes.

obulat:
Sorry, @fcoveram! I was wrong. Your list of what needs to be updated is right:

> What I understood from the chat we had https://github.com/obulat, is that this request creates unintended changes in the following components:
> 
>     Row (large, medium, and small)
>     Box (large, and medium)
>     Global

PR #2697, opened 2023-07-21
Create a new `VIconButton`

obulat:
## Fixes

Related to #460 by @fcoveram

## Description

This PR adds a new `VIconButton` component using the updated styles in the [Figma mockups](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=4658-12477&mode=design&t=3iXHyuTBcFAAWjLe-4)

The old styles for `VIconButton` were used mainly in the popover/modal close buttons, audio controls and the buttons within the search bar. They will gradually be updated to use the new `VIconButton`.

## Testing Instructions

Check out the new `VIconButton` story. It should have all of the styles from the Figma mockup. Old buttons (listed above) should also look the same.

Check the rendered Story: https://docs.openverse.org/_preview/2697/storybook/?path=/docs/components-viconbutton--default-story

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VIconButton/meta/VIconButton.stories.mdx
```diff
@@ -33,17 +38,20 @@ export const Template = (args) => ({
 export const sizesTemplate = (args) => ({
   template: `
     <div class="flex gap-x-2">
-      <div v-for="size in ['tiny', 'small', 'medium', 'large', 'extra-large']" :key="size" class="flex flex-col items-center p-2">
+      <div v-for="size in args.baseButtonSizes" :key="size" class="flex flex-col items-center p-2">
```
AetherUnbound, line 41: Nice!

AetherUnbound (approved):
LGTM! It seems like we're losing the `tiny` and `extra-large` sizes, is that intentional?

zackkrida:
@obulat could you clarify the intention of this PR? It creates the new icon button, but doesn't actually implement it anywhere, right? Does that mean that the button styles _do_ need to change in the app (which will require updating visual regression tests) and that we will do that in subsequent PR(s)?

I'm mostly curious about the approach taken and if it would make more sense to create the new VIconButton but backwards-compatible, with the old sizes included (to be deprecated later), rather than maintain two versions for a (likely brief) period.

obulat:
> LGTM! It seems like we're losing the `tiny` and `extra-large` sizes, is that intentional?

Yes! Those sizes were previously used for Audio controls. Now the designs have updated buttons with fewer sizes - that's why we are updating.

obulat:
> @obulat could you clarify the intention of this PR? It creates the new icon button, but doesn't actually implement it anywhere, right? Does that mean that the button styles _do_ need to change in the app (which will require updating visual regression tests) and that we will do that in subsequent PR(s)?

The `VIconButton` component is currently used in three ways:
- close buttons in popover/modals - the external sources popover close button is the use case that prompted this PR
- `VSearchBarButton` - these button styles should not be changed, and they can be re-factored to use the new `VIconButton` because that will simplify the sizes and focus handling.
- Audio controls - #2602 updates them, but after the designs of the audio track are updated to match them.

The change in "close" buttons will cause changes in the visual regression snapshots.

> I'm mostly curious about the approach taken and if it would make more sense to create the new VIconButton but backwards-compatible, with the old sizes included (to be deprecated later), rather than maintain two versions for a (likely brief) period.

Current component (`VOldIconButton`) has a very complicated sizes system: we have to set both the icon and the button size, and they are coupled together, so there are too many sizes and they are still not enough to match all of the buttons we use.
After the update to `VSearchBarButton`, we can audit the close buttons used throughout the app in one go (updating the snapshots), and then `VOldIconButton` will be completely removed in the PR that updates the Audio control.

fcoveram (approved):
It looks great ✨ 🚀 

I noticed two prop variants I wasn't aware of, but do not block the work made: `filled-transparent` and `transparent-tx`. The first one is also an oxymoron 😅

Issue #2462, opened 2023-06-21
Component: Content safety wall


dhruvkb:
## Description

Define the component that shows when visiting a single result page with sensitive content. 

### State

The single result page itself must have a walled/unwalled boolean state, because once the content is revealed (by clicking the 'Show content' button) it is possible to conceal it again (using the 'Hide content' button #2463).

The text that explains why the content is walled should be updated based on the value of the `sensitivity` supplied by the API. See discussion at https://github.com/WordPress/openverse/issues/791#issuecomment-1484245797.

### Events

| Name | Associated data | Description |
|-|-|-|
| `showContent` | | The user has opted to see the content, load the page content and hide the wall |
| `goBack` | | The user has opted not to see the content, go back to search results (if search term is known) or homepage (otherwise) |

`showContent` must toggle with the page's sensitive content shown/hidden state.

## Screenshots

| Desktop | Mobile |
|-|-|
| ![image](https://github.com/WordPress/openverse/assets/16580576/eae00322-018b-4961-8c65-a26986ac15a1) | ![image](https://github.com/WordPress/openverse/assets/16580576/83997399-e255-43b7-a254-679a8b0c2d6c) |

## Figma

- **Figma:** https://www.figma.com/file/bDTfGzBit8rGRPktOR1cu4/Safe-content-browsing-flow?type=design&node-id=1140-40205&t=ikf0HKUfRTFKOgA0-4

PR #2700, opened 2023-07-21
useSensitiveMedia composable

zackkrida:
## Fixes

Related to #2462 by @dhruvkb 

## Description

This PR creates a new `useSensitiveMedia` composable for working with sensitive media, and updates the blurring logic to use this composable. This composable is fully tested with a unit test, and forms the basis of the "backend" work for the single result sensitivity UI. It includes hide/reveal methods that aren't used by this PR but will be for the VSafetyWall.

This PR also adds revealedSensitiveResults, a list of UUIDs of revealed sensitive media, to the UI store.

It was extracted into a standalone PR for ease of reviewing.

## Testing Instructions

Check that blurring still works as-expected in the frontend search results, when all relevant `/preferences` toggles are enabled. Check that you like the `useSensitiveMedia` implementation and that test coverage is sufficient.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/composables/use-sensitive-media.ts
+++ b/frontend/src/composables/use-sensitive-media.ts
@@ -20,7 +20,7 @@ export function useSensitiveMedia(
   const uiStore = useUiStore()
 
   /**
-   * The current state of a single sentive media item.
+   * The current state of a single sensitive media item.
    * Respects defaults from the site's filters and is
    * updated by user interactions.
    */
--- a/frontend/src/composables/use-sensitive-media.ts
+++ b/frontend/src/composables/use-sensitive-media.ts
@@ -5,7 +5,7 @@ import type { Media } from "~/types/media"
 import { useUiStore } from "~/stores/ui"
 
 type SensitiveVisibilityState =
-  | "insensitive"
+  | "non-sensitive"
   | "sensitive-shown"
   | "sensitive-hidden"
 
@@ -26,7 +26,7 @@ export function useSensitiveMedia(
    */
   const visibility = computed<SensitiveVisibilityState>(() => {
     if (!media) {
-      return "insensitive"
+      return "non-sensitive"
     } else if (media.isSensitive) {
       if (uiStore.revealedSensitiveResults.includes(media.id)) {
         return "sensitive-shown"
@@ -36,7 +36,7 @@ export function useSensitiveMedia(
           : "sensitive-shown"
       }
     } else {
-      return "insensitive"
+      return "non-sensitive"
     }
   })
 
--- a/frontend/src/composables/use-sensitive-media.ts
+++ b/frontend/src/composables/use-sensitive-media.ts
@@ -1,0 +1,83 @@
+import { computed, watch } from "vue"
+
+import type { Media } from "~/types/media"
+
+import { useUiStore } from "~/stores/ui"
+
+type SensitiveVisibilityState =
+  | "insensitive"
+  | "sensitive-shown"
+  | "sensitive-hidden"
+
+/**
+ * A helper composable for working with sensitive media.
+ * Provides the current visibility of a sensitive media,
+ * along with toggles for hiding/revealing the media.
+ */
+export function useSensitiveMedia(
+  media: Pick<Media, "id" | "isSensitive"> | null
+) {
+  const uiStore = useUiStore()
+
+  // Clear out previously-revealed results when the toggle is switched.
+  watch(
+    () => uiStore.shouldBlurSensitive,
+    () => {
+      uiStore.revealedSensitiveResults = []
+    },
+    { deep: true }
+  )
+
+  /**
+   * The current state of a single sentive media item.
+   * Respects defaults from the site's filters and is
+   * updated by user interactions.
+   */
+  const visibility = computed<SensitiveVisibilityState>(() => {
+    if (!media) {
+      return "insensitive"
+    } else if (media.isSensitive) {
+      if (uiStore.revealedSensitiveResults.includes(media.id)) {
+        return "sensitive-shown"
+      } else {
+        return uiStore.shouldBlurSensitive
+          ? "sensitive-hidden"
+          : "sensitive-shown"
+      }
+    } else {
+      return "insensitive"
+    }
+  })
+
+  function reveal() {
+    if (media && !uiStore.revealedSensitiveResults.includes(media.id)) {
+      uiStore.revealedSensitiveResults.push(media.id)
+    }
+  }
+
+  function hide() {
+    if (!media) return
+    const index = uiStore.revealedSensitiveResults.indexOf(media.id)
+    if (index > -1) {
+      uiStore.revealedSensitiveResults.splice(index, 1)
+    }
+  }
+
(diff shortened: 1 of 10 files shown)


frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,83 @@
...
+  watch(
+    () => uiStore.shouldBlurSensitive,
+    () => {
+      uiStore.revealedSensitiveResults = []
+    },
+    { deep: true }
+  )
+
+  /**
+   * The current state of a single sentive media item.
```
obulat, line 32:
```suggestion
   * The current state of a single sensitive media item.
```

frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,83 @@
...
+  | "sensitive-shown"
+  | "sensitive-hidden"
+
+/**
+ * A helper composable for working with sensitive media.
+ * Provides the current visibility of a sensitive media,
+ * along with toggles for hiding/revealing the media.
+ */
+export function useSensitiveMedia(
+  media: Pick<Media, "id" | "isSensitive"> | null
```
obulat, line 18: Why do we want to allow `null` as a value for `media` here?
The code would be much simpler if we didn't :)

frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,83 @@
...
+) {
+  const uiStore = useUiStore()
+
+  // Clear out previously-revealed results when the toggle is switched.
+  watch(
+    () => uiStore.shouldBlurSensitive,
+    () => {
+      uiStore.revealedSensitiveResults = []
+    },
+    { deep: true }
```
obulat, line 28: I tried to see if we could get rid of this `{ deep: true }`, because Pinia docs say it's expensive, and it is usually used to watch nested objects, and here it is a boolean. Well, apparently, a non-deep watcher is fired when the property is set to a new object, and that never happens to a boolean - I guess that's why `deep` is needed.

Then I added a `console.log` here to track when this watcher is called, and it's called about 27 times on the "cat" All content search page. So, it is called by **every** component that imports this composable. And we reset `revealedSensitiveResults` resultCount times every time we toggle "blur sensitive results".

I tried re-factoring this composable to be a `useSensitiveMedia` store, but I think a better solution would be to move this watcher to the `ui` store. We can simulate a watcher there by adding a `setShouldBlurSensitive` function, which would be toggling the property _and_ also resetting the `revealedSensitiveResults`. And never set the value directly, like `uiStore.shouldBlurSensitive="true"`.

I guess using watchers that watch global properties or store properties inside composables is a bad idea because those watchers would be triggered by every component that uses the composable.

frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  function reveal() {
+    if (media && !uiStore.revealedSensitiveResults.includes(media.id)) {
+      uiStore.revealedSensitiveResults.push(media.id)
+    }
+  }
+
+  function hide() {
+    if (!media) return
+    const index = uiStore.revealedSensitiveResults.indexOf(media.id)
```
obulat, line 60: If we use a `set` for `revealedSensitiveResults`, we won't have to check whether the item is already in the array or not. 
Pinia didn't support `Set`s and `Map`s when hydrating on SSR, but the support was added in https://github.com/vuejs/pinia/issues/1608 (not that we really need SSR support since the reveal can only happen client side, right?)

obulat (changes requested):
Thank you for extracting these changes, they are really fun! They work well, but I found a problem with implementation, and one suggestion on possible simplification using `Set`.

The biggest blocker is the `uiStore` watcher that is called by every component using `use-sensitive-media` composable.

frontend/src/composables/use-sensitive-media.ts
zackkrida: I noticed this last night, too, and think moving it to the UI store would be ideal. It's not a _super_ important piece of functionality, anyway, so definitely isn't worth the performance hit to have it here

zackkrida: Resolved in e82bc7c43.

frontend/src/composables/use-sensitive-media.ts
zackkrida: Later, in the safety wall PR, we call useSensitiveMedia on the single result pages where the media might be `null`. I wasn't sure of the best way to handle that and this felt the easiest.

frontend/src/composables/use-sensitive-media.ts
zackkrida: I believe array performance is still a bit better than Set performance in browsers, so I don't have a strong preference here. if you think it's a blocker I'm fine to update it though!

obulat: First of all, let me say that it's not a blocker :) I don't think there would be a huge difference in performance, especially since we would probably have no more than 100 elements in the `revelaedSensitiveResults` anyways. 

I got curious about the browser performance of Sets, and it _seems_ that it's OK (and much faster for some operations with > 10000 items, which is not our case though)

frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,74 @@
...
+  const uiStore = useUiStore()
+
+  /**
+   * The current state of a single sensitive media item.
+   * Respects defaults from the site's filters and is
+   * updated by user interactions.
+   */
+  const visibility = computed<SensitiveVisibilityState>(() => {
+    if (!media) {
+      return "insensitive"
```
obulat, line 29: Nit: maybe "non-sensitive" is a better name for this?

obulat (approved):
Works well and has great tests :)

frontend/src/composables/use-sensitive-media.ts
zackkrida: Agreed. Updated in c8e6917b2

frontend/src/composables/use-sensitive-media.ts
```diff
@@ -0,0 +1,74 @@
+import { computed } from "vue"
+
+import type { Media } from "~/types/media"
+
+import { useUiStore } from "~/stores/ui"
+
+type SensitiveVisibilityState =
+  | "non-sensitive"
+  | "sensitive-shown"
+  | "sensitive-hidden"
```
dhruvkb, line 10: Nit: these could be individually exported from a constants file as they are being used for `==` matching below.

dhruvkb (approved):
LGTM, this is a good abstraction over the sensitivity functionality.

frontend/src/composables/use-sensitive-media.ts
zackkrida: Moved to the content-safety constants file in c69e77829; great suggestion!

PR #2727, opened 2023-07-27
Ensure data refresh can still be run if no initial index exist

### LINKED ISSUE #2725: Change trigger rule for data refresh's `generate_index_suffix` to handle case where `get_current_index` skips
## Problem

In #2106, we modified the data refresh's `get_current_index` step to skip if an index doesn't exist. While the create filtered index DAG was modified to handle this case, the primary data refresh DAG was not (since we've never encountered a case yet in production where the index did not exist!). When deploying our new ES8 cluster (https://github.com/WordPress/openverse-infrastructure/pull/569) and running the data refresh, this step skipped the entire DAG since the index did not exist and the rest of the DAG depended on it. We were able to rectify this in the moment by marking the task as succeeded and then clearing the rest of the DAG. The downstream `trigger_delete_index` step did not fail because it still passes even if the index provided does not exist.

## Fix 

We should modify the primary data refresh DAG so that the `generate_index_suffix`'s [trigger rule](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#concepts-trigger-rules) is `none_failed` (i.e. "All upstream tasks have not `failed` or `upstream_failed` - that is, all upstream tasks have succeeded or been skipped").

This change involves modifying this operator creation to include `trigger_rule=TriggerRule.NONE_FAILED` ([example](https://github.com/WordPress/openverse/blob/8ec29d723603e8210fc35951e05ef0e32e45e58e/catalog/dags/data_refresh/recreate_view_data_task_factory.py#L54)):

https://github.com/WordPress/openverse/blob/8ec29d723603e8210fc35951e05ef0e32e45e58e/catalog/dags/data_refresh/data_refresh_task_factory.py#L149-L152

sruthiv98:
## Fixes
Fixes #2725 by @AetherUnbound 

## Description 
This PR adds a `none_failed` trigger rule to the `generate_index_suffix` task in the primary data refresh DAG. It ties together the old index retrieval and old index deletion tasks so both are either run or skipped together. It also adds an `ensure_downstream_runs` task at the end of the data refresh task group to ensure that the task state from promote is reflected downstream (rather than the task state from `trigger_delete_index` which can now be skipped).

**Before**
Screenshot_2023-08-31_14-03-00

**After** _(ignore the failures, they're expected and should propagate all the way through to the end of the DAG)
Screenshot_2023-08-31_14-01-25

## Testing Instructions

See this comment below: https://github.com/WordPress/openverse/pull/2727#issuecomment-1656160714

When these steps are run on this branch, both the `get_current_index` and `trigger_delete_index` steps should be skipped, but all other data refresh steps should run as intended (in the case provided, this means that all tasks after the data refresh task group should remain "upstream failed").

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_refresh/data_refresh_task_factory.py
+++ b/catalog/dags/data_refresh/data_refresh_task_factory.py
@@ -54,6 +54,7 @@ from airflow.operators.python import PythonOperator
 from airflow.sensors.external_task import ExternalTaskSensor
 from airflow.utils.state import State
 from airflow.utils.task_group import TaskGroup
+from airflow.utils.trigger_rule import TriggerRule
 
 from common import ingestion_server
 from common.constants import REFRESH_POKE_INTERVAL, XCOM_PULL_TEMPLATE
@@ -148,6 +149,7 @@ def create_data_refresh_task_group(
         # Generate a UUID suffix that will be used by the newly created index.
         generate_index_suffix = PythonOperator(
             task_id="generate_index_suffix", python_callable=lambda: uuid.uuid4().hex
+            trigger_rule=TriggerRule.NONE_FAILED
         )
         tasks.append(generate_index_suffix)
 
--- a/catalog/dags/data_refresh/data_refresh_task_factory.py
+++ b/catalog/dags/data_refresh/data_refresh_task_factory.py
@@ -148,8 +148,9 @@ def create_data_refresh_task_group(
 
         # Generate a UUID suffix that will be used by the newly created index.
         generate_index_suffix = PythonOperator(
-            task_id="generate_index_suffix", python_callable=lambda: uuid.uuid4().hex
-            trigger_rule=TriggerRule.NONE_FAILED
+            task_id="generate_index_suffix",
+            python_callable=lambda: uuid.uuid4().hex,
+            trigger_rule=TriggerRule.NONE_FAILED,
         )
         tasks.append(generate_index_suffix)
 


AetherUnbound:
Thanks for submitting this fix @sruthiv98! You may want to look at our [general setup guide](https://docs.openverse.org/general/general_setup.html#just) for instructions on how to set up `just` and [starting the stack](https://docs.openverse.org/general/quickstart.html#starting-up). Once that's set up for you locally, you can run `just lint` and commit the changes, that will ensure the code is formatted a specific way and pass our linting check 🙂 

On testing, this particular case is pretty specific since it _technically shouldn't_ happen during normal processing. I've confirmed that you can replicate the original problem using the following steps:

1. `just down -v && just c && just i` (this clears out existing volumes, then launches both the catalog stack and the ingestion server stack)
2. Enable the audio data refresh DAG (http://localhost:9090/dags/audio_data_refresh/graph)
3. Mark Success on the `get_before_record_count` task when it fails (no data exists)
4. The `get_current_index` task should skip, and so should the rest of the DAG

After your change, the behavior of step 4 should change so that it skips, but the rest of the DAG runs (it may fail, because there's no data, but as long as it runs even if `get_current_index` is skipped we're good!

Hope this helps!

sruthiv98:
> After your change, the behavior of step 4 should change so that it skips, but the rest of the DAG runs (it may fail, because there's no data, but as long as it runs even if `get_current_index` is skipped we're good!)

Thanks for the detailed testing notes!! Super helpful. Just ran this and `get_current_index` did not automatically skip (it was stuck on `marked_for_retry`, and when I marked it as `success` the DAG moved on to the next task. Is this a behavior we were expecting or is the fact that it kinda stalled on `marked_for_retry` indicate that something isn't right?

AetherUnbound:
> Thanks for the detailed testing notes!! Super helpful. Just ran this and `get_current_index` did not automatically skip (it was stuck on `marked_for_retry`, and when I marked it as `success` the DAG moved on to the next task. Is this a behavior we were expecting or is the fact that it kinda stalled on `marked_for_retry` indicate that something isn't right?

Oh, I remember you mentioning something about running the initialization (via `just init`). Can you try once more starting from `just down -v` on the testing instructions and see if that does it?

sruthiv98:
> Can you try once more starting from `just down -v` on the testing instructions and see if that does it

Oh this is how I tested it originally! But yes,  just tried it again and the run still stalls with `up_for_retry` at the `get_current_index` task :/

AetherUnbound:
> Oh this is how I tested it originally! But yes, just tried it again and the run still stalls with `up_for_retry` at the `get_current_index` task :/

Sorry it's taken me so long to get back to this @sruthiv98! Just to confirm, did you also mark the `get_before_record_count` task as success? That one is inconsequential to the pieces we're testing but does cause the run to fail otherwise! It should be step 3 of the instructions posted above. If it still fails after that, please post the logs of the failed task!

AetherUnbound:
@sruthiv98 this required a few more changes than I anticipated, so in the interest of getting this merged I applied those changes and rebased. Thanks for your help on the initial portion!

AetherUnbound (approved):

obulat (approved):
Thank you for your contribution, @sruthiv98 🎉

PR #2734, opened 2023-07-27
Add `VCollectionHeader` component

### LINKED ISSUE #2729: Component:  `VCollectionHeader`
## Description

For the additional search views, we need a new component that will display the view's tag/creator/source and the results count.

### Props

| Name | Type | Description |
|-|-|-|
| collection | "tag", "creator", "source" | The kind of additional search view |
| title | string | The tag, creator name or the source's display name |
| url | string (optional) | The link to the creator's page or the source |
| resultsLabel | string | The results count string |

## Code samples

```
<div class="flex flex-col gap-y-6 lg:gap-y-8">
    <div class="flex flex-col gap-4 lg:flex-row lg:justify-between">
      <div class="flex items-center gap-x-2">
        <VIcon :name="iconName" :size="isDesktop ? 10 : 6" />
        <h1 class="text-3xl font-semibold leading-6 lg:text-6xl lg:leading-10">
          {{ title }}
        </h1>
      </div>
      <VButton
        v-if="url"
        as="VLink"
        variant="filled-dark"
        size="medium"
        class="label-bold"
        has-icon-end
        show-external-icon
        :external-icon-size="6"
        :href="url"
        >{{ $t(`collection.link.${collection}`) }}</VButton
      >
    </div>
    <p class="caption-regular lg:label-regular text-dark-charcoal-70">
      {{ resultsLabel }}
    </p>
  </div>
```
## Dependencies

New `svg` icons: `institution`, `person` and `tag`.

## References

- **Figma:** creator [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1200-56323&amp;mode=design&amp;t=pN0PPAzlKQEKT9tJ-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1200-56362&amp;mode=design&amp;t=suLIyJHNmZrM0mPH-4), source [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1200-56381&amp;mode=design&amp;t=suLIyJHNmZrM0mPH-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1200-56421&amp;mode=design&amp;t=suLIyJHNmZrM0mPH-4), tag [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1216-58402&amp;mode=design&amp;t=suLIyJHNmZrM0mPH-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&amp;node-id=1200-56457&amp;mode=design&amp;t=suLIyJHNmZrM0mPH-4).

obulat:
## Fixes

Fixes #2729 by @obulat

## Description

This PR adds the component that displays the collection icon, its title, external link (if available) and the number of results in the collection.

This component is not used anywhere yet, so you can only see it in the Storybook.

## Testing Instructions

Check out the rendered [Story](https://docs.openverse.org/_preview/2734/storybook/?path=/story/components-vcollectionheader--all-collections). It should match the Figma designs ( creator [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56323&mode=design&t=pN0PPAzlKQEKT9tJ-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56362&mode=design&t=suLIyJHNmZrM0mPH-4), source [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56381&mode=design&t=suLIyJHNmZrM0mPH-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56421&mode=design&t=suLIyJHNmZrM0mPH-4), tag [desktop](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1216-58402&mode=design&t=suLIyJHNmZrM0mPH-4) and [mobile](https://www.figma.com/file/niWnCgB7K0Y4e4mgxMrnRC/Additional-search-views?type=design&node-id=1200-56457&mode=design&t=suLIyJHNmZrM0mPH-4)).

Clicking on the external link button sends an analytics event, but since the build is "production", you cannot see the event details logged in the console, only the warning that Plausible was disabled.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VCollectionHeader/VCollectionHeader.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+  <div class="flex flex-col gap-y-6 lg:gap-y-8">
+    <div class="flex flex-col gap-4 lg:flex-row lg:justify-between">
+      <div class="flex items-center gap-x-2">
+        <VIcon :name="iconName" :size="isDesktop ? 10 : 6" />
+        <h1 class="text-3xl font-semibold leading-6 lg:text-6xl lg:leading-10">
+          {{ title }}
+        </h1>
```
fcoveram, line 8: In `md` I envision a white gap in the top-right area for most cases. To avoid that, I think we should try changing to `flex-row`  from `md` instead of `lg` .

fcoveram (changes requested):
I wonder if we can detect when `h1` changes to two lines and align the content (icon, text, and button) in the top. In small and big breakpoints the outcome looks bad. Besides, the label's button shrinks.

| `xs` | `lg` |
|---|---|
| Collection header in XS breakpoint | Collection header in LG breakpoint |

frontend/src/components/VCollectionHeader/VCollectionHeader.vue
```diff
@@ -0,0 +1,143 @@
...
+    url: {
+      type: String,
+    },
+    /**
+     * The label showing the result count, to display below the title.
+     */
+    resultsLabel: {
+      type: String,
+      required: true,
+    },
```
sarayourfriend, line 81: Passing the result label manually like this results in the storybook example not quite matching the designs, but only be consequence of the story configuration. The creator design has different text than the other collections:

image

If we changed the component to accept a configuration object on the `collection` prop where `collection` could be a union of the following interfaces:

```typescript
interface BaseCollection {
  type: unknown
  mediaType: MediaType
  resultCount: number
  title: string
}

interface TagCollection extends BaseCollection {
  type: 'tag'
}

interface SourceCollection extends BaseCollection {
  type: 'source'
  url: string
}

interface CreatorCollection extends BaseCollection {
  type: 'creator'
  source: string
  url: string
}
```

Then the component should select the correct template string and interpolate `title`, `resultCount`, `mediaType` and, if a creator, `source`.

There are probably some linguistic difficulties here though that might warrant passing individual source names to the translators... Some sources really need a definite article and others do not. Maybe leaving the definite article out is okay in English for some. "10000 results from Museum of Modern Art" reads clumsy to me. "10000 results from the Museum of Modern Art" would be better. But "10000 results from the NASA" is incorrect. In Portuguese the definite article is always included in the word "from" ("10000 resultados do Museum of Modern Art"). However, even then, the grammatical gender should be considered. If we had a source "Biblioteca Nacional do Brasil", the grammatical gender is feminine and it should be "10000 resultados da Biblioteca Nacional do Brasil".

I wonder if everywhere we are going to interpolate a source name into a sentence, whether we need to create a translation string for it, like automatically generating certain translations from the stats endpoints. We wouldn't translate the source names themselves, necessarily (though that may be appropriate in some cases, and transliterating source names could be useful for [languages with non-Latin script](https://ja.wikipedia.org/wiki/ニューヨーク近代美術館)), but it would still be necessary context for minimally correct translation in some languages. It would be jarring to read "10000 resultados do Public Library of New York" for an adult with English literacy, but all the more so (and potentially incomprehensible) for a child, e.g., in an educational context. We can't do anything about creator names, of course, but the list of sources is trivially obtainable and limited (unlike creator names of which there are probably hundreds of thousands and most cannot or shouldn't be translated or transliterated anyway).

If we're going down the road of creating sentences from source names, however, it does seem like something that we could consider in an entirely separate issue. I don't know exactly how adding those strings to `en.json5` would work, if we could do so reliably, etc. We could go with a naive approach now that avoids needing to do any translations... maybe "10000 images - Biblioteca Nacional do Brasil", "10000 tagged images", and "10000 images - sarayourfriend, Flickr" (not concrete suggestions).

In either case, I think passing the `resultsLabel` in this way sort of invites accidental misuse or unintentionally duplicating similar translation strings.

frontend/src/types/analytics.ts
```diff
@@ -216,6 +216,18 @@ export type Events = {
...
+   * Description: The user visits a source's link in the single result UI
+   * Questions:
+   *   - Are source links clicked much? Does Openverse increase visibility
+   *     of included sources?
+   */
+  VISIT_SOURCE_LINK: {
+    /** The unique ID of the media */
+    id: string
+    /** The permalink to the source's homepage */
+    url: string
```
sarayourfriend, line 229: How does url get used for filtering? For creator links, it will be a unique URL. Does it need to be the creator link? Could it just be the source slug? That would be easier to filter on most of the time.

frontend/src/types/analytics.ts
```diff
@@ -216,6 +216,18 @@ export type Events = {
...
   }
+  /**
+   * Description: The user visits a source's link in the single result UI
+   * Questions:
+   *   - Are source links clicked much? Does Openverse increase visibility
+   *     of included sources?
+   */
+  VISIT_SOURCE_LINK: {
+    /** The unique ID of the media */
+    id: string
```
sarayourfriend, line 227: Can we rename this to `collectionType`? The comment doesn't appear to be related to this event either.

```suggestion
    /** The type of collection, either "source" or "creator" (tags do not have external links) */
    collectionType: "source" | "creator"
```

frontend/src/components/VCollectionHeader/meta/VCollectionHeader.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+})
+
+<Meta
+  title="Components/VCollectionHeader"
+  component={VCollectionHeader}
+  argTypes={{
+    id: {
+      options: ["tag", "source", "creator", "source-with-long-name"],
+      control: { type: "select" },
+    },
```
sarayourfriend, line 49: `id` is a confusing name for this. I didn't understand what it did in the story controls until reading the code. `example-collection` would be clearer. Note: changing the name here requires updating the `Template` story implementation as well.

I wonder if we need two different stories, though? The all collections story feels entirely appropriate for the default story. It isn't a huge component that is a drag to use if all the examples are rendered at a single time.

frontend/test/storybook/visual-regression/v-collection-header.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+test.describe("VCollectionHeader", () => {
+  breakpoints.describeMobileAndDesktop(({ expectSnapshot }) => {
+    for (const languageDirection of ["ltr", "rtl"]) {
+      for (const collection of [
+        "tag",
+        "creator",
+        "source",
+        "source-with-long-name",
+      ]) {
```
sarayourfriend, line 13: If we use the all collections story for this test instead then we don't need individual tests here. Is that worse for testing, though? I guess this way if only the tag changes then we have a tag-specific test that catches that change, rather than needing to dig through the diff to identify where the change is coming from.

sarayourfriend (changes requested):
This looks excellent so far, but there are a few things that need changing and clarification before I can approve. I've requested blocking changes to the story (change the `id` control's name) and the event configuration. I've suggested non-blocking but strongly recommended changes to the props for the component. Lastly, I've added a rather long question about how best to manage translation of the subtitle for each collection type. This didn't occur to me during design reviews but is rather significant if we're going to do it correctly. I suggested a workaround that might be a good stop-gap if we want to avoid blocking on this complexity. The workaround would need to be implemented alongside the prop changes.

frontend/src/components/VCollectionHeader/VCollectionHeader.vue
fcoveram: Great catch and suggestions.

Translating source names sound interesting, we can certainly can create an improvement issue to provide better context. So far, I like the following suggestion for the sentence.

> 10000 images - sarayourfriend, Flickr

Although I would replace the dash with a period to end up with something like this: `10000 images. sarayourfriend, Flickr`

frontend/src/types/analytics.ts
obulat: The `id` property here should be removed, and `collectionType` is not necessary.

This event is fired when the user clicks on the external source link on the source collection page (the black button in the snapshots here). 
When the user clicks on the `creator`'s external link, a different event (`VISIT_CREATOR_LINK`) is fired.

fcoveram (changes requested):
It looks great. Just two comments:
* For the three variants, the horizontal layout in `md` is correct, but `h1` should follow the same style as in `xs / sm`
* In Source, the results sentence should say `10000 images provided by this source` instead of `10000 images from <source name>`. This was thought to avoid i18n conflicts.

obulat:
> * `10000 images from <source name>`

This wording also doesn't solve the problem of i18n: some sources need `the`, some don't. We should try to find a way to not have the source name part of the sentence. Something like "Source name: 10000 images"?

obulat:
> It looks great. Just two comments:
> 
> * For the three variants, the horizontal layout in `md` is correct, but `h1` should follow the same style as in `xs / sm`
> * In Source, the results sentence should say `10000 images provided by this source` instead of `10000 images from <source name>`. This was thought to avoid i18n conflicts.

I updated the `h1` style to change after the `lg` breakpoint. As for the result sentences, those should be passed as props by the parent component, so we should update them in the PR that creates the page for the additional search views.

frontend/src/types/analytics.ts
obulat: We could probably replace the url with the source slug - to make sure that if we ever change the URL we link to, the event is still the same.

frontend/src/components/VCollectionHeader/VCollectionHeader.vue
obulat: @sarayourfriend, I started to implement changes to the label here, and then realized that it needs more props for some of the collection kinds (media type to show whether it's "... images" or "...audio files", results count, etc.). It's better to create the label in the parent component (the collection page that already has all the necessary information)

frontend/test/storybook/visual-regression/v-collection-header.spec.ts
obulat: I refactored the story so that it's a single story for all 4 cases.

fcoveram (changes requested):
My suggestion about applying the horizontal layout to `md` looks great, but I didn't consider that a smaller font size would break the alignment due to a `40px` icon. The outcome looks broken.

Creator results variant of Collection header component

I tried different ideas, but the best solution is going back to the previous text style ("Heading 2") but keeping the horizontal composition. Sorry for not anticipating this.

### Button alignment

As a minor detail, the horizontal layout pushes the black button to the edge and creates a disalignment when it's close to the big text. We can solve this by adding a `margin-top: 4px` to the button. Here is my test in Chrome inspector.

Source variant of Collection header component. Margin values in Chrome inspector are highlighted

Since this is only for the horizontal layout (`md / lg / xl / 2xl`), the button margin should disappear in `xs / sm`.

### Results sentence

And finally, would [this suggestion](https://github.com/WordPress/openverse/pull/2734#discussion_r1299899890) be address in this PR?

frontend/src/components/VCollectionHeader/VCollectionHeader.vue
sarayourfriend: That's why I suggested the expanded collection type, to encode the required information for each collection type. I generally have a preference for logic being as "low" in the component tree with higher components only responsible for overall layout, but where you think it would be best to do. Anywhere we put it will include the same complexity.

obulat:
@fcoveram, I reverted the breakpoint change for the h1 style (it's `md:` now instead of `lg:`) and added the top margin for breakpoints above `md`. I also updated the snapshot tests so that you can see all of the breakpoints, not only 2.
As for the texts, I updated the creator results label according to the suggestion, but they will also be updated in the PR that fixes #2774

fcoveram (approved):
It looks great 🚀 ✨

frontend/src/types/analytics.ts
sarayourfriend: Good idea!

frontend/src/types/analytics.ts
```diff
@@ -206,15 +206,32 @@ export type Events = {
...
+   * Questions:
+   *   - Are source links clicked much? Does Openverse increase visibility
+   *     of included sources?
+   */
+  VISIT_SOURCE_LINK: {
+    /** The unique ID of the media, if the source link is on the single
+     * result page */
+    id?: string
+    /** The source's slug that identifies it */
+    slug: string
```
obulat, line 234:
```suggestion
    source: string
```

obulat: To match the `VISIT_CREATOR_LINK` event

sarayourfriend:
@obulat I was waiting for a final verdict on the sub-heading templating here: https://github.com/WordPress/openverse/pull/2734#discussion_r1302279123

Can you confirm the plan? I'll review on Monday if this PR is ready to go, but then I think we'll need a new issue to record the changes/complexities of that sub-heading.

obulat:
> Can you confirm the plan? I'll review on Monday if this PR is ready to go, but then I think we'll need a new issue to record the changes/complexities of that sub-heading.

Thank you, @sarayourfriend . The plan is to move the discussion of the changes/complexities of the sub-heading to #2774 that will create the subheading, translate it and pass as a prop to `VCollectionHeader`.

sarayourfriend (approved):
LGTM!

Just two things:

- Might be worth update the sub-heading in the story to be a generic sub-heading text like "example sub-heading" rather than approximating the eventual designs, primarily to prevent confusion about why the story does not match the eventual designs or why the implementation on the final page doesn't match the story either.
- Something strikes me as unbalanced about the mobile button @fcoveram. Some screenshots for reference below. I don't know if this is because the story has all the different types all in a row (which will never happen in real life) and each of their buttons are slightly different widths (due to different text inside), but I think I do also perceive this unbalanced issue on the Figma designs as well. Particularly with a longer title and full-width image and audio results, the button seems out of place, somehow on the xs size. On the sm size it looks weird for it to be full-width, though.

**With the story's grey background removed**
image

**Full width on xs size**
image

**Full width on sm size**
image

Not sure what the solution is, or if it needs a change at all, even. But thought I would mention it just in case.

In any case, it doesn't need any changes in this PR, it can be addressed in a follow up issue if necessary.

Good work @obulat

fcoveram:
Thanks for the suggestion @sarayourfriend; that section kept me iterating different ideas a lot. Unfortunately, one spacing solution doesn't work for horizontal and vertical layouts.

But in the future, we can try showing the view name next to the icon on `xs` to break down that imbalance. Here is a mockup.

Header area of source content view

PR #2778, opened 2023-08-03
Stop converting timeout error to an object in Nuxt API requests

### LINKED ISSUE #2746: Timeout errors are not handled correctly
## Description
The Axios error handling in the `media/index.ts` is using `axios.isAxiosError` for type safety. ~~However, this function returns false for some axios errors. In the linked Sentry issue you can see that the error has a `request` property, but doesn't have `isAxiosError` property that the `axios.isAxiosError` type checker is looking for.~~
I realized that the `isAxiosError` is fine, but the problem is that we replace a timeout error with an object that has some of the same properties here:
https://github.com/WordPress/openverse/blob/61c7abdeaed0450558859ff949846f9a1497ad13/frontend/src/data/api-service.ts#L128-L133

This is why the errors when no response was received (like the timeout errors) are not handled correctly. 

## Solution
Instead of replacing the error in those lines, we should update the `error.message`:

```
if (error.code === "ECONNABORTED") {
        error.message = `timeout of ${
            DEFAULT_REQUEST_TIMEOUT / 1000
          } seconds exceeded`
      }
```
And then throw the `error` and not an object.

## Sentry link:

Sentry Issue: [OPENVERSE-FRONTEND-19Z](https://openverse.sentry.io/issues/4350185721/?referrer=github_integration)

```
Error: Object captured as exception with keys: code, config, message, name, request
  at key (./src/stores/media/index.ts:491:28)
```

obulat:
## Fixes

Fixes #2746 by @obulat

## Description

This PR stops the conversion of the error into an object. This will allow us to handle timeout errors correctly.
A follow-up to this PR should be the fix for #632 that actually shows the timeout error page instead of the generic yellow error page as it does not.

## Testing Instructions

To easily simulate timeouts, change the default timeout to 30:
https://github.com/WordPress/openverse/blob/61c7abdeaed0450558859ff949846f9a1497ad13/frontend/src/data/api-service.ts#L8
From the logs it's not clear that the error was converted to an object because it's causing the Promise to reject, and is handled in the same `catch` block.
However, you will see the warning in the console saying "There was an error fetching media providers for image: ECONNABORTED", which means that the error is recognized as the AxiosError in the store code:
https://github.com/WordPress/openverse/blob/61c7abdeaed0450558859ff949846f9a1497ad13/frontend/src/stores/provider.ts#L138-L141 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/data/api-service.ts
+++ b/frontend/src/data/api-service.ts
@@ -125,12 +125,9 @@ export const createApiService = ({
     (response) => response,
     (error) => {
       if (error.code === "ECONNABORTED") {
-        return Promise.reject({
-          message: `timeout of ${
-            DEFAULT_REQUEST_TIMEOUT / 1000
-          } seconds exceeded`,
-          ...error,
-        })
+        error.message = `timeout of ${
+          DEFAULT_REQUEST_TIMEOUT / 1000
+        } seconds exceeded`
       }
       return Promise.reject(error)
     }


krysal (approved):
It would be _so_ nice to get rid of that error!

sarayourfriend (approved):

PR #2782, opened 2023-08-05
Collect image dimensions from Europeana

### LINKED ISSUE #1484: Update Europeana to collect image dimensions
## Problem

We'd like to make sure that all of our image provider scripts are collecting data about image dimensions.

## Description

The Europeana script does not currently do so, but the information is available in the API via the [single record endpoint](https://pro.europeana.eu/page/record). We should update the provider script to collect this info.

There is precedent for making an additional API request per record, notably [Flickr](https://github.com/WordPress/openverse-catalog/blob/7bf8ec42408b1e4c76925ce159bdeec7de691c73/openverse_catalog/dags/providers/provider_api_scripts/flickr.py#L332) but also some others like Metropolitan and NYPL. That being said we should make sure that adding it here doesn't have any dramatic effect on performance.

## Additional context/dependent PRs

~Note that Europeana is currently turned off in production, because turning it on will initiate a backfill. We should wait for WordPress/openverse-catalog#644 to make sure that we can omit the DAG from Slack notifications, to prevent flooding the channel.~ <- This has been merged.

~We also have an issue to update Europeana to use the new API endpoint, and update any necessary fields: WordPress/openverse#1727. It may be helpful or even necessary to do this work first.~ <- https://github.com/WordPress/openverse-catalog/pull/974 has been merged.

## Implementation

rwidom:
## Fixes
Fixes #1484 by @stacimc 

## Description

Adds image dimensions, filetype, and filesize from Europeana item data.

## Testing Instructions

At a high level, the goal is to run all the tests using `pytest`, then run the dag in airflow, then look at the data to confirm that it looks reasonable. 

- Run `just catalog/test-session` and then `pytest tests/dags/providers/provider_api_scripts/test_europeana.py`. 
- When I ran `pytest` alone from within the same test session, I got a bunch of errors from `test_ingestion_server.test_index_readiness_check` but that's just because the ingestion server stuff isn't started up when you're just running docker for a catalog test session.
- To get the local port for airflow (http://localhost:9090/home) working in docker, I ran `just catalog/recreate`. Then you can turn on the dag and it will go through daily runs and be sure to pick up some data. Check out the results using sql after `just catalog/pgcli`.

I temporarily replaced the dag start date with 8/8/2023 in provider_workflows.py, to get a smaller sample size for testing.  Only two days out of nine had data, but it seems like the 3 second timeout might be reasonable at least for the daily runs, if these numbers of records are comparable with production. I would need to know the total number of records in production to figure out if that is workable for a refresh.

```
*DAG*: `europeana_workflow`
*Date range*: 2023-08-08 -> 2023-08-09
*Duration of data pull tasks*: 24 mins
*Number of records upserted per media type*:
  - `image`: 474

*DAG*: `europeana_workflow`
*Date range*: 2023-08-15 -> 2023-08-16
*Duration of data pull tasks*: 56 mins, 20 secs
*Number of records upserted per media type*:
  - `image`: 1,113
```

The new fields are populated in all of the inserted records:

```
openledger> select filetype, count(*) records, count(filesize) filesize, count(width) width, count(height) height from image group by filetype order by 
 2 desc;
+----------+---------+----------+-------+--------+
| filetype | records | filesize | width | height |
|----------+---------+----------+-------+--------|
| jpg      | 1573    | 1573     | 1573  | 1573   |
| png      | 14      | 14       | 14    | 14     |
+----------+---------+----------+-------+--------+
SELECT 2
Time: 0.020s
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,52 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
...
+        # Seems like might be best to reach out to Europeana and get their preference
+        # for the best way for us to proceed.
+        if not (item_id := data.get("id")):
+            logger.warning(f"No id, cannot request additional info on {data}")
+            return {}
+        item_response = self.get_response_json(
+            query_params=self.item_params,
+            endpoint=f"https://api.europeana.eu/record/v2{item_id}.json",
+        )
+        if item_response is None or not (item_response.get("success")):
```
obulat, line 275: It is probably easier to check `if not item_response` in case there's another non-dictionary falsy value here

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,52 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
...
+        # - Switch to 3 second delay across the board and leave everything else alone,
+        #   or maybe add a max number of this kind of failure per batch or per dag run.
+        # - Separate delayed requesters 30 seconds for batch, 1 second for items, same
+        #   default batch size = 100.
+        # - Smaller batches (35), but default delay = 1 sec, like Brooklyn Museum.
+        # - Just go with defaults across the board, on the assumption that the batch
+        #   query is the expensive one, and there will be at least 30 seconds between,
+        #   given the 100 individual look-ups.
+        # Seems like might be best to reach out to Europeana and get their preference
+        # for the best way for us to proceed.
```
rwidom, line 267: @obulat any thoughts on these delay settings/timing issues? Do you think additional performance testing would be valuable here?

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,52 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
...
+        # Testing with the first webresource from the first aggregation, got dimensions
+        # for 1424 / 1589 images; so not too worried about performance implications of
+        # the loops. Limiting factor was much more the delay between requests.
+        if item_object := item_response.get("object"):
+            if aggregations := item_object.get("aggregations"):
+                for aggregation in aggregations:
+                    if webresources := aggregation.get("webResources"):
+                        for webresource in webresources:
+                            if filetype := webresource.get("ebucoreHasMimeType"):
+                                if filetype[:5] == "image":
```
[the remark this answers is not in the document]
rwidom: Ahhh, fair enough! I think that `.startswith()` is probably more readable too, and it looks to be performant as long as we can trust that the `filetype` will be a string. Seem ok?

```
(openverse) % python -m timeit -s 'filetype="image/jpeg"' '(len(filetype)>=5 and filetype[:5]=="image")'
2000000 loops, best of 5: 130 nsec per loop
(openverse) % python -m timeit -s 'filetype="image/jpeg"' 'filetype.startswith("image")'                
2000000 loops, best of 5: 102 nsec per loop
```

catalog/dags/providers/provider_api_scripts/europeana.py
obulat: I don't know, but I think @stacimc might :)

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -159,7 +176,7 @@ class EuropeanaDataIngester(ProviderDataIngester):
     providers = {"image": prov.EUROPEANA_DEFAULT_PROVIDER}
     sub_providers = prov.EUROPEANA_SUB_PROVIDERS
     endpoint = "https://api.europeana.eu/record/v2/search.json?"
-    delay = 30
+    delay = 3
```
obulat, line 179: Is this delay correct?

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,52 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
...
+        )
+        if not item_response or not item_response.get("success"):
+            logger.warning("Item request failed no response or ``success != True``")
+            return {}
+        # Assume that we just want the first info available in the item response,
+        # but consider adding checks, e.g. on landing url.
+        # Testing with the first webresource from the first aggregation, got dimensions
+        # for 1424 / 1589 images; so not too worried about performance implications of
+        # the loops. Limiting factor was much more the delay between requests.
+        if item_object := item_response.get("object"):
```
obulat, line 283: We could probably reduce nesting if we used something like
`if aggregations := item_response.get("object", {}).get("aggregations"):` and 
`for webresource in aggregation.get("webResources", [])`

obulat (approved):
I ran the script locally and the dimensions are being picked up well. I'll defer to @stacimc on the delay timing because she has more experience with it.
I added a comment on making the code a little bit less nested, but I don't want to block on it.

catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
```diff
@@ -100,9 +104,93 @@ def test_get_batch_data_gets_items_property(ingester):
...
+    ) as item_call:
+        with patch.object(image_store, "add_item"):
+            record_count = ingester.process_batch(batch_json["items"])
+            assert item_call.call_count == len(batch_json["items"])
+            assert record_count == len(batch_json["items"])
+
+
+ITEM_HAPPY_RESPONSE = _get_resource_json("item_full.json")
+ITEM_NOT_1ST_RESPONSE = _get_resource_json("item_not_first_webresource.json")
+ITEM_HAPPY_WEBRESOURCE = {
```
krysal, line 122: This looks like it could be extracted from the `item_full.json` file. Could it be the same for `ITEM_NOT_1ST_WEBRESOURCE`?

catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
```diff
@@ -180,6 +268,44 @@ def test_get_foreign_landing_url_without_edmIsShownAt(record_builder):
...
+@pytest.mark.parametrize(
+    "item_data, expected",
+    [
+        pytest.param(
+            ITEM_HAPPY_WEBRESOURCE, {"width": 381, "height": 480}, id="happy_path"
+        ),
+        pytest.param({"no": "dimensions"}, {}, id="no_dimensions"),
+    ],
+)
+def test_get_image_dimensions(item_data, expected, record_builder):
```
krysal, line 280: It would be nice to add the cases where there is only one of the dimensions.

krysal (commented):
@rwidom Thanks for adding these four fields in one PR! :) Seems like this is changing the delay of 30 seconds that we currently have. @stacimc probably has better reasons to argue about that, as Olga said, but I'll just ask: did you confirm there is no way to grab these fields from the search endpoint?

rwidom:
> @rwidom Thanks for adding these four fields in one PR! :) Seems like this is changing the delay of 30 seconds that we currently have. @stacimc probably has better reasons to argue about that, as Olga said, but I'll just ask: did you confirm there is no way to grab these fields from the search endpoint?

I looked in the search API documentation [here](https://pro.europeana.eu/page/search#metadata-sets), and it seems like it maybe should be possible (with some kind of profile magic?), but it's not clear to me how to do it. Definitely when I look at the test files we've actually downloaded, I don't see anything that looks like those fields. It would be great if we didn't have to query individual objects, because then we could leave the 30 seconds alone.

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,50 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
...
+        )
+        if not item_response or not item_response.get("success"):
+            logger.warning("Item request failed no response or ``success != True``")
+            return {}
+        # Assume that we just want the first info available in the item response,
+        # but consider adding checks, e.g. on landing url.
+        # Testing with the first webresource from the first aggregation, got dimensions
+        # for 1424 / 1589 images; so not too worried about performance implications of
+        # the loops. Limiting factor was much more the delay between requests.
+        if aggregations := item_response.get("object", {}).get("aggregations"):
```
stacimc, line 283: I'd need to double check, but I think a generator would be faster, and IMO a little easier to read (although opinions may vary 😄). Something like:

```
aggregations = item_response.get("object", {}).get("aggregations", [])
for aggregation in aggregations:
    return next(
        (resource for resource in aggregation.get("webResources", []) if resource.get("ebucoreHasMimeType", "").startswith("image")),
        {} # Returns `{}` by default
    )
```

stacimc (changes requested):
Thanks for taking this on @rwidom!

I spent some time looking at the API documentation and found the aggregations very confusing. My takeaway is that what we're getting back from the search endpoint is a collection of "metadata records", each of which may have multiple actual images associated with it. When we query the individual record endpoint, then, we may get multiple images in the `aggregations` with different filesizes etc.

So I took a look at one of the examples for record id `/2021650/memorix_0000ee69_1b9e_6823_478f_c88af61736c6`. The image url that we get from ingestion is https://images.memorix.nl/nda/thumb/fullsize/010a31d7-9316-54d7-6ad0-83b9914688d9.jpg. That url _is_ one of the webresources we can find at the record endpoint, but it's not the first one -- that's a (smaller) version of the same image, https://images.memorix.nl/nda/thumb/640x480/010a31d7-9316-54d7-6ad0-83b9914688d9.jpg.

The result is that, at least some of the time, the final record that gets ingested will have image dimensions for a different size than the one we're linking out to, which I don't think we want. Moreover based on the API documentation, I'm worried that we might also get different photos entirely -- like related photos or multiple angles of the same object, in which case it would be really strange to get the dimensions of the wrong image.

I wonder if we could try to filter the webresources for one that exactly matches the url we used for `image_url`?

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +243,50 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
         return response_json.get("items")
 
     def get_record_data(self, data: dict) -> dict:
-        return self.record_builder.get_record_data(data)
+        return self.record_builder.get_record_data(
+            data,
+            self._get_additional_item_data(data),
```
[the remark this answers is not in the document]
rwidom: I hear that. I feel nervous about it, because the overall structure seems so flexible that I worry about things getting jumbled up that way. What would you think of something like `data | {"item webresource": webresource}`?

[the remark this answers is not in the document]
stacimc: Good point, I think caution is understandable! I like that idea.

rwidom:
@stacimc , I think I made the code-only changes you asked for, except the timing one. I'm not sure the best way to reach out to Europeana, but while I was looking, I came across [this "Harvesting and Downloads" page](https://pro.europeana.eu/page/harvesting-and-downloads) which seems directly relevant to us, though I haven't looked in more detail. Hmmm...

rwidom:
Thanks for talking through the timeout issue @stacimc  and @AetherUnbound !

catalog/dags/providers/provider_api_scripts/europeana.py
stacimc: I think we can go ahead and remove this comment now!

catalog/dags/providers/provider_reingestion_workflows.py
```diff
@@ -56,15 +55,6 @@ def __post_init__(self):
 
 
 PROVIDER_REINGESTION_WORKFLOWS = [
-    ProviderReingestionWorkflow(
```
stacimc, line 59: 🥳

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -223,7 +244,61 @@ def get_batch_data(self, response_json: dict) -> None | list[dict]:
         return response_json.get("items")
 
     def get_record_data(self, data: dict) -> dict:
-        return self.record_builder.get_record_data(data)
+        return self.record_builder.get_record_data(
+            data | {"item webresource": self._get_additional_item_data(data)}
```
stacimc, line 248: Could we change the key to just `webresource` (or `item_webresource` with the underscore)?

stacimc (commented):
This looks great! Just a couple tiny clean-up requests and it looks like we need to `just catalog/generate-dag-docs`, but then this should be good to go! 🚢

stacimc (approved):
This all looks and tests great @rwidom! 🚀 Thanks for leading the discussion on this one, I'm excited for this :)

PR #2822, opened 2023-08-13
polyfill flatMap for Chrome < 69

### LINKED ISSUE #2745: Error when cleaning up vueuse event handlers on Chrome 68 TypeError: n.flatMap is not a function
## Description
We updated Vueuse version recently, and moved to using the Vueuse `useEventListener` composable instead of the one we vendored in. This is causing problems for older browsers.

Vueuse uses `flatMap` to clean up the event listeners when an element is unmounted. `flatMap` is not available in Chrome versions below 69, so the event listeners cannot be removed.
Here's a link to the issue in the Vueuse repository: https://github.com/vueuse/vueuse/issues/2642

A workaround suggested in the issue comment is to import the `core-js` `flatMap` shim in the entry file:
`import 'core-js/actual/array/flat-map'`

I'm not sure where the `entry` file in the Nuxt app would be. Best place, I think, would be a plugin, similar to the `focus-visible.client.ts` (but without `.client`). @zackkrida, do you agree?

## Sentry link
Sentry Issue: [OPENVERSE-FRONTEND-16K](https://openverse.sentry.io/issues/4329079092/?referrer=github_integration)

```
TypeError: n.flatMap is not a function
  at useFeatureFlagStore.actions.syncAnalyticsWithLocalStorage (./src/stores/feature-flag.ts:227:23)
...
(5 additional frame(s) were not displayed)
```

zackkrida:
## Fixes

Fixes #2745 by the sentry bot

## Description

Polyfills the Array.prototype.flatMap method so it works in versions of Chrome < 69. This has occurred enough times to warrant the polyfilling. 

## Testing Instructions

Build locally and confirm the presence of the polyfill in the client bundle by searching for `flatMap:` and confirming it's there. 

We should also see a minuscule increase to the bundle size.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -138,6 +138,7 @@ const config: NuxtConfig = {
     "~/plugins/ua-parse.ts",
     "~/plugins/focus-visible.client.ts",
     "~/plugins/api-token.server.ts",
+    "~/plugins/polyfills.client.ts",
     "~/plugins/sentry.ts",
   ],
   css: ["~/assets/fonts.css", "~/styles/tailwind.css", "~/styles/accent.css"],
--- a/frontend/src/plugins/polyfills.client.ts
+++ b/frontend/src/plugins/polyfills.client.ts
@@ -1,0 +1,5 @@
+// This plugin is used to load any necessary client-side browser polyfills.
+
+// Purpose: Sentry issue for issue with Chrome < 69 and our useStorage composable:
+// https:openverse.sentry.io/share/issue/2b2635c85f1c4cecbe7560df5d4892b5/
+import "core-js/actual/array/flat-map"


obulat (approved):
Thank you for fixing this! The testing instructions and comments are great.

sarayourfriend (commented):
Rather than manually adding the polyfill (and potentially playing whack-a-mole if other newer APIs get used by dependencies), would it be cleaner to expand our browserslist to accurately reflect the browsers we wish to support (in this case, Chrome < 69)? As I understand it, if we do it that way we might only see a bundle size increase for the "legacy" bundle rather than in all bundles, in addition to the benefit of not needing to manually manage polyfills.

obulat:
> Rather than manually adding the polyfill (and potentially playing whack-a-mole if other newer APIs get used by dependencies), would it be cleaner to expand our browserslist to accurately reflect the browsers we wish to support (in this case, Chrome < 69)? As I understand it, if we do it that way we might only see a bundle size increase for the "legacy" bundle rather than in all bundles, in addition to the benefit of not needing to manually manage polyfills.

Would adding something like "chrome >= 68" to the "browserslist" in `package.json` also turn into a game of wack-a-mole, though? What if another version of another browser would start causing a different issue?
I looked up the distribution of browsers that cause this error, and there is a wide range of browsers and versions. There's an argument for adding this polyfill specifically because we do get a very clear signal from Sentry that we need to polyfill the `flatMap`. I am not against your suggestion, @sarayourfriend , I'm just trying to understand the tradeoffs.

sarayourfriend (approved):
I'm going to approve and merge this now, actually, because the bundle size increase is relatively small and browserslist doesn't support complex usage of `not` ("or not" does not work the way we would think it does). But I will say that if we're adding a polyfill for this, we should consider how our browserslist target is configured. The number of actual clients that are effected by this, based on Chrome numbers on our Plausible, is probably ~200. That 200 clients of 578000. In other words, _very few people_. 

I think a preferrable alternative to always including the flat-map polyfill could be an asynchronous import of `https://polyfill.io/v3/polyfill.min.js?version=3.111.0&features=Array.prototype.flatMap%2CArray.prototype.flat`. Polyfill.io only returns the polyfills if the requesting browser needs it. We'd add `await import('https://polyfill.io/v3/polyfill.min.js?version=3.111.0&features=Array.prototype.flatMap%2CArray.prototype.flat')` before using the particular vueuse core composable that requires it.

In any case, we can go with this option for now, but we could open to change it to use a conditional polyfill. Half a kB doesn't seem like heaps when our total bundle size is 868 kB... but that rationale would have us never make bundle size improvements that incrementally add up in the end. From a carbon perspective, on the other hand, it might be better to always send that half a kB rather than creating another request to the polyfill.io CDN. Hard trade offs to make :thinking:  I haven't gone ahead and opened the issue because I don't know exactly which action to recommend myself.

sarayourfriend:
Oops, I was writing my approving review while you were writing your comment, Olga. The approach with browserslist I was hoping would work is the `not supports array-flat`, but it doesn't combine "not" in a flexible enough way. It might be possible to do but I think it would be very complex.

In any case, approved and merged. I think my comment about the actual number of people effected by this is valid though and we should consider strongly what kind of coverage we consider as standard. Our browserslist configuration is actually _narrower_ than the default browsers list config, so there isn't anything to currently suggest we have an intention of very wide client support.

PR #2826, opened 2023-08-14
Extract use-external-sources composable

### LINKED ISSUE #2825: Extract use-external-sources as a composable
## Problem

We use the list of external sources in the External sources form below the search results and on the No results page. For this, we pass this list through several layers of components, which makes it difficult to decouple components.

## Description

Extract the `use-external-sources` composable so that it can be called by the `VNoResults` and `VExternalSourceList` components directly, instead of passing the list from the `search.vue` page.

## Additional context
This will greatly simplify the external sources mobile changes and the error pages improvements.

obulat:
## Fixes

Fixes #2825 by @obulat

## Description

This PR extracts `useExternalSources` composable, which will simplify the changes for the external sources modal (#460) and the error pages (because we have external sources buttons in the No results page).

## Testing Instructions

The external sources on the image and audio results pages should be displayed correctly, have no warnings and the no results page should display correctly (you can see it if you search for "QueryWithNoResults")

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VExternalSearch/VExternalSearchForm.vue
+++ b/frontend/src/components/VExternalSearch/VExternalSearchForm.vue
@@ -1,6 +1,6 @@
 <template>
   <section
-    :key="type"
+    :key="externalSourcesType"
     ref="sectionRef"
     class="external-sources flex flex-row place-items-center justify-center py-4"
     data-testid="external-sources-form"
@@ -19,7 +19,9 @@
       class="description-regular"
     >
       <template #openverse>Openverse</template>
-      <template #type>{{ $t(`externalSources.form.types.${type}`) }}</template>
+      <template #type>{{
+        $t(`externalSources.form.types.${externalSourcesType}`)
+      }}</template>
     </i18n>
 
     <i18n
@@ -28,7 +30,9 @@
       tag="p"
       class="description-regular"
     >
-      <template #type>{{ $t(`externalSources.form.types.${type}`) }}</template>
+      <template #type>{{
+        $t(`externalSources.form.types.${externalSourcesType}`)
+      }}</template>
       <template #query>{{ searchTerm }}</template>
     </i18n>
 
@@ -64,8 +68,6 @@
       >
         <VExternalSourceList
           class="flex flex-col"
-          :external-sources="externalSources"
-          :media-type="type"
           :search-term="searchTerm"
           @close="closeDialog"
         />
@@ -75,7 +77,7 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent, PropType, ref, SetupContext } from "vue"
+import { computed, defineComponent, ref, SetupContext } from "vue"
 
 import { storeToRefs } from "pinia"
 
@@ -87,9 +89,7 @@ import { useMediaStore } from "~/stores/media"
 
 import { useDialogControl } from "~/composables/use-dialog-control"
 import { useAnalytics } from "~/composables/use-analytics"
-
-import type { MediaType } from "~/constants/media"
-import type { ExternalSource } from "~/types/external-source"
+import { useExternalSources } from "~/composables/use-external-sources"
 
 import VExternalSourceList from "~/components/VExternalSearch/VExternalSourceList.vue"
 import VButton from "~/components/VButton.vue"
@@ -107,16 +107,8 @@ export default defineComponent({
     VExternalSourceList,
   },
   props: {
-    type: {
-      type: String as PropType<MediaType>,
-      required: true,
-    },
     searchTerm: {
       type: String,
-      required: true,
-    },
-    externalSources: {
-      type: Array as PropType<ExternalSource[]>,
       required: true,
     },
     isSupported: {
@@ -133,7 +125,7 @@ export default defineComponent({
   },
   setup(_, { emit }) {
     const sectionRef = ref<HTMLElement | null>(null)
-    const triggerRef = ref<InstanceType<typeof VButton> | null>(null)
+    const triggerRef = ref<{ $el: HTMLElement } | null>(null)
     const uiStore = useUiStore()
     const searchStore = useSearchStore()
     const { sendCustomEvent } = useAnalytics()
@@ -172,7 +164,10 @@ export default defineComponent({
       onTriggerClick()
     }
 
+    const { externalSourcesType } = useExternalSources()
+
     return {
+      externalSourcesType,
       sectionRef,
       triggerRef,
       triggerElement,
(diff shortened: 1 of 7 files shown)


frontend/src/components/VExternalSearch/VExternalSearchForm.vue
```diff
@@ -133,7 +125,7 @@ export default defineComponent({
   },
   setup(_, { emit }) {
     const sectionRef = ref<HTMLElement | null>(null)
-    const triggerRef = ref<InstanceType<typeof VButton> | null>(null)
+    const triggerRef = ref<{ $el: HTMLElement } | null>(null)
```
obulat, line 128: This is not strictly necessary for this PR, but it stops WebStorm from throwing an error in detecting the type of the triggerRef.

frontend/src/composables/use-external-sources.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  IMAGE,
+  isAdditionalSearchType,
+  isSupportedMediaType,
+} from "~/constants/media"
+import { getAdditionalSources } from "~/utils/get-additional-sources"
+
+export const useExternalSources = () => {
+  /**
+   * External sources search form shows the external sources for current search type, or for images if the search type is 'All Content'.
```
obulat, line 14:
```suggestion
   * External sources search form shows the external sources for current search type.
```

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
sarayourfriend: I think that is because this file isn't included in `tsconfig.json`... if we're making changes to appease a particular IDE but ignoring the fact that VSCode, the most popular free IDE, shows heaps of errors on all the imports in these files that are marked `type="ts"` without being in the tsconfig, then I have a feeling we're not prioritising things correctly. Either we should remove the `type="ts"` from these files and not play whack-a-mole with errors like this, that actually _reduce_ type information and require manually writing types that do not need to be manually maintained or once and for all finally add everything to the tsconfig.

Critically, `type="ts"` on these files that are not included in the tsconfig are really a shadow pattern because it provides  false sense of security. It _looks_ like the file is type checked, but actually it never is. And on top of that it creates confusing issues for people setting up the project in anything that uses the normal `vue-tsc` rather than JetBrains's own implementation.

Anyway, I'm not suggesting to do anything in this PR. Just pointing out again that we're doing something very strange and I think actually quite risky (in the false-confidence sense of things) by working with TypeScript this way.

frontend/src/composables/use-external-sources.ts
```diff
@@ -0,0 +1,40 @@
...
+      return searchType
+    }
+    return IMAGE
+  })
+  const externalSources = computed(() => {
+    const searchStore = useSearchStore()
+    const query = searchStore.searchQueryParams
+    const type = externalSourcesType.value
+    return getAdditionalSources(type, query)
+  })
```
sarayourfriend, line 34: Is it safe to call `useSearchStore` etc inside the computed call-backs?

sarayourfriend (changes requested):
Requesting changes just to clarify regarding the store composable callbacks called inside `computed`. `useSearchStore` in particular is called in both and could be shared, I would think :thinking:

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
obulat: My comment was actually misleading, sorry. The problem occurs both in WebStorm and in VS Code. Here's the message you get in both:
```
The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.ts(7056)

The inferred type of 'default' cannot be named without a reference to '.pnpm/@sentry+core@7.57.0/node_modules/@sentry/core'. This is likely not portable. A type annotation is necessary.ts(2742)
```
I suppose it's having the problem with the `default` imported from `VButton` that we use for annotation, and is replacing it with the `default` imported from `sentry`, for some reason. [This issue ](https://github.com/microsoft/TypeScript/issues/47663) says that it's a bug with TypeScript and pnpm. There is a workaround in the issue, but I couldn't understand how to get it to work.

On your point of files not being in `tsconfig`, I completely agree. However, we have recently added all of the `components/` to `tsconfig`:
https://github.com/WordPress/openverse/blob/dfd98eeced8d20f5f2684b0676a58f59a4d6435f/tsconfig.json#L40
We still need to add some pages (most notably `search.vue`)

sarayourfriend: Gotcha, no worries then.

Do you know why this doesn't cause an issue when running the type check?

frontend/src/composables/use-external-sources.ts
obulat: My reasons for putting the `useStore` inside the computed were incorrect:
You cannot call `useStore` from outside of a `setup` function of a component because then `pinia` instance is not available. However, since we only call this composable from a component's `setup` function, `useStore` can be placed inside the composable. (The problem would arise if we put the `useStore` outside the `useExternalSources` function, or if we called `useExternalSources` from a middleware, for instance).

I moved the `useStore` calls to the composable itself, and added a comment on calling the composable from a `setup` function.

sarayourfriend: Just for my knowledge, should we avoid naming things that _aren't_ called during `setup` `use*` and composables? Would that be a useful disambiguator?

Also, _is_ it safe to call them inside a computed callback? I genuinely don't know and can't remember the precise conditions that cause a memory leak :slightly_smiling_face:. I remember it had to do with allocating a new ref in a callback, but I don't know if store callbacks, etc, cause that issue?

One more thing, would [something like this ESLint plugin (maybe not exactly that one as it's new and small)](https://github.com/ktsn/eslint-plugin-vue-composable) be useful to prevent some cases of mistakes? Apparently the main [Vue plugin for ESLint decided not to implement any composable rules](https://github.com/vuejs/eslint-plugin-vue/issues/2093), but only because of the naming convention issue.

sarayourfriend (approved):
LGTM!

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
obulat: > Do you know why this doesn't cause an issue when running the type check?

No, I don't. I tried adding that component to `tsconfig` manually, and type check still passed. When I also added `search.vue`, type check failed. So, I assume that type checks work, and all components **do** pass the type checks.

frontend/src/composables/use-external-sources.ts
obulat: > Also, is it safe to call them inside a computed callback? I genuinely don't know and can't remember the precise conditions that cause a memory leak 🙂. I remember it had to do with allocating a new ref in a callback, but I don't know if store callbacks, etc, cause that issue?
I also do not remember. I'll try to look it back to have a better idea (and document, if possible). An ESLint plugin sounds great!

PR #2838, opened 2023-08-16
Show timeout errors on the frontend

obulat:
## Fixes

Fixes #632 by @sarayourfriend

## Description

This PR updates the `FetchState` to always use the `FetchingError` instead of `NuxtError`|`string`|`FetchingError` for the error object. 
`FetchingError` interface provides more information about the error, including the response status code, the string code of the error from axios, the kind of request (single result, search, provider or the related media), and additional details such as the item id.

This interface makes it easier to show the timeout error pages on `SERVER_TIMEOUT` or `ECONNABORTED` errors.

`FetchState` was made a generic to allow the use of `string` and `NuxtError` for the `fetchingError` value. This PR removes this to make fetchState simpler.

## Testing Instructions

To simulate the server timeout, lower the default timeout to 30:
https://github.com/WordPress/openverse/blob/026a328a781ebebfc4b29a22b4b041b05acdb544/frontend/src/data/api-service.ts#L8
Now, you should see the timeout error page any time you search for something. You should also see the same page if you refresh this search page (so, the page will be server-rendered now).

To see the no results error page, search for something like "querywithnoresults". It should work for all search types, both on CSR and SSR, and should not try to reload the page on SSR.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
[...]

PR #2845, opened 2023-08-18
Refactor the error page

obulat:
## Fixes

Related to #632 by @sarayourfriend

## Description

_Note: this PR was split from #2838 to make the PR review easier. ~~More than 40 changed files here are the deleted snapshots~~ Snapshot changes were extracted to a separate PR_

This PR adds a new `VErrorPage` component that can show the client-side error using only the `fetchState.fetchingError` as a prop. This is simpler than the previous version that required providing the image and the error code and other necessary data separately. This will allow us in the future to add details to the error page easily. For instance, we will be able to show how much time is left until next retry on  429 error.

The layouts were updated a little bit based on the [i3 mockups](https://www.figma.com/file/H1qjjlII3PaDltPqW59Vj7/No-results-page?type=design&node-id=1001-6695&mode=dev). Those are not finalized, but I think they are the most up-to-date version that we can use until the new designs are ready.

~~The `VSearchGrid` component was also removed because it overcomplicates the layout, requires you to pass too many props, and uses a slot for `media`(audio/image) search page that breaks the context. Now `search` page handles the display of the search header above the media results and the external search form below them. It also shows the error page instead of the results when there is an error.~~ - this was moved to #2866

~~The visual regression tests were refactored: only "mobile" and "desktop" variants are checked (instead of all breakpoints) to save on the snapshot files. I believe that this is enough for testing that the error pages are displayed correctly. 
All of the error page tests were moved to `errors.spec.ts`, the names were standardized, and if some error pages are the same, the same snapshot was reused (e.g., 500, 429 use the same "generic" error, at least before we implement the new error page designs).~~

## Testing Instructions

The main changes of this PR are in the `VSearchPage.vue` and `search.vue` components, and the `VErrorPage` stories. These changes should make sense.

To test the error page, try searching for something that does not have any results ("querywithnoresults") - you should see the No results error page. 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VErrorSection/VErrorPage.vue
```diff
@@ -0,0 +1,63 @@
...
+  setup(props) {
+    const searchStore = useSearchStore()
+    const searchTerm = computed(() => searchStore.searchTerm)
+    /**
+     * The code used for the error page image.
+     * For now, NO_RESULT image is used for searches without result,
+     * and SERVER_TIMEOUT image is used as a fall-back for all other errors.
+     */
+    const errorCode = computed(() => {
+      return props.fetchingError?.message?.includes(NO_RESULT)
```
obulat, line 46: These checks will be refactored in #2838 to use the `code` property of a `fetchingError` object instead of relying on the text of an error message.

frontend/src/components/VErrorSection/meta/VErrorSection.stories.mdx
```diff
@@ -6,36 +6,31 @@ import {
...
-      </template>
-      <VNoResults :media-type="args.type" :search-term="args.searchTerm" />
-    </VErrorSection>
-    `,
-  components: { VErrorSection, VErrorImage, VNoResults },
+export const ErrorPageTemplate = (args) => ({
+  template: `<VErrorPage v-bind="args" />`,
+  components: { VErrorPage },
   setup() {
+    const searchStore = useSearchStore()
```
obulat, line 30: The searchTerm will be stored in the `fetchingError` in #2838, so we won't need to access to the store here.

frontend/test/playwright/visual-regression/components/global-audio-player.spec.ts
```diff
@@ -21,7 +22,10 @@ for (const dir of languageDirections) {
       )
       const audioRow = await audio.getNthAudioRow(page, 2)
       await audio.play(audioRow, dir)
-      await audio.pause(audioRow, dir)
+      await page
```
obulat, line 25: Clicking the "pause" button on the search results sometimes doesn't wait for the audio to load, so the focus of the global audio track is not displayed correctly.
I changed it to click the "pause" button on the global audio player instead.

krysal:
@obulat There are some conflicts in the snapshots, when do you think this can be updated and ready for review?

obulat:
> @obulat There are some conflicts in the snapshots, when do you think this can be updated and ready for review?

Thank you for the reminder, @krysal! I rebased this PR onto main, and set it as ready for review.

frontend/src/components/VErrorSection/VErrorImage.vue
```diff
@@ -71,12 +73,13 @@ export default defineComponent({
...
+  @apply mt-4 text-sr text-dark-charcoal-70;
 }
-::v-deep(a) {
+::v-deep(.attribution a) {
   @apply text-current underline;
 }
-::v-deep(img) {
-  @apply opacity-70; /* to match the text color */
+/* license icons should match the text color */
+::v-deep(.attribution img) {
```
obulat, line 82: Previous declaration was causing the large error images to have lower opacity, too. Designs only have lower opacity for the license icons in the attribution sentence.

frontend/src/components/VErrorSection/VErrorImage.vue
```diff
@@ -38,7 +38,9 @@ export default defineComponent({
      * the code of the error, used to identify and render the appropriate image
      */
     errorCode: {
-      type: String as PropType<ErrorCode>,
+      type: String as PropType<
+        Extract<ErrorCode, "NO_RESULT" | "SERVER_TIMEOUT">
+      >,
```
sarayourfriend, line 43: Why is `Exctract` needed for this? No result and server timeout are the only two error codes, right? Is there one added in a follow up PR that causes this requirement?

One kind of weird side-effect of using `Extract` is that it won't complain if something on the right-hand side isn't actually part of the left-hand side: https://www.typescriptlang.org/play?#code/C4TwDgpgBAogTnA9nAwogJtAvFARAOQHkB9AJRgGUBVAGQBVcoAfPCmUgNXeLoEkBZGISoMAUKNCQodKDhgAPYHACGAY2AAeeElQYIAGjxEylWg2at2XUjwFCRjFgUIAFGLgB8ogPTeo-gD0AflEgA

Preferably there'd be some kind of error that said "Hey, `NOPE` isn't ever part of `ErrorCode`!". Seeing as we're already defining the two we want inline anyway, I'd vote for us to import the two consts and do something like:

```suggestion
      type: String as PropType<
        typeof NO_RESULT
        | typeof SERVER_TIMEOUT
      >,
```

Or if we really want to make sure that `ErrorCode` is part of it, use `Exclude` instead of `Extract` so that we don't run into the issue where an ErrorCode is removed or changed but isn't handled.

I prefer just defining the union inline. If we really need to use something like `Extract`, a custom type would be able to cause a type error when the members don't extend the union: https://www.typescriptlang.org/play?#code/C4TwDgpgBAogTnA9nAwogJtAvFARAOQHkB9AJRgGUBVAGQBVcoAfPCmUgNXeLoEkBZGISoMAUKNCQoFAK4AjAM4RgAHioA7AJaJ1UCAA9gEdegVQFwOJvUBzADRR+EALZyIcMwaMmzG7eoA+KBwYQzgAQwBjVT8dBydXdwUA8UloOmDpeSVVeCRUDAgHAhJyanpGFlw2Tm4+QWEGZjwiAAUYXBSAei6oPoA9AH5xIA

Anyway, not really an issue, just a nit because this looks confusing and tempting for someone to remove given the 1-to-1 relationship between the left and right side of this right now. If that's changing in the future, all good, but a comment explaining why we can't define the accepted error codes inline and need to use extract instead would be helpful.

frontend/src/locales/scripts/en.json5
```diff
@@ -687,6 +687,9 @@
   serverTimeout: {
     heading: "Whoops, it looks like that request took too long to complete. Please try again.",
   },
+  unknownError: {
+    heading: "Whoops, it looks like something went wrong. Please try again. ",
```
sarayourfriend, line 691:
```suggestion
    heading: "Whoops, it looks like something went wrong. Please try again.",
```

sarayourfriend (approved):
LGTM!

dhruvkb (approved):
The changes look good to me. I only have two nits

- `VErrorPage` seems like a misleading name for the compontent because it's not technically a page like those in `pages/`
- if the component is renamed to `VErrorPage`, the folder it is in should be too

Approving because everything is functionally great 👌!

frontend/src/components/VErrorSection/VErrorImage.vue
obulat: Thank you for the detailed context! I replaced the type in this PR with the simple union, and will include the suggestions in #2838, which actually changes the `FetchingError` type.

Issue #2731, opened 2023-07-27
Update Source and Provider links in the single result page Information section


obulat:
## Problem

The links to Source and Provider on the single result page do not have an indication that they are external and open in a new tab. Moreover, they are inconsistent across the media types and link to the media foreign landing URL instead of the source or the provider (#560)

## Description

The links to creator in the [image](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/pages/image/_id/index.vue#L62-L73) and [audio](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/components/VAudioTrack/layouts/VFullLayout.vue#L32-L41) single result pages Information section should have an “external link” icon. Note: the `VLink` with an external link needs update because it doesn't wrap, and simply adding `show-external-link` to `VLink` creates single-line links, and long source names overlap other text on the page.

The audio creator link should also be updated to match the image creator link. It should be a conditional component: `VLink` if the `creator_url` is truthy and span if the `creator_url` is falsy.

Currently, the foreign_landing_url is linked to the “source” in the [image](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/components/VImageDetails/VImageDetails.vue#L26-L28) page and “provider” in the [audio page](https://github.com/WordPress/openverse/blob/35f08e5710d59f6c10f0cf54103fcce151adfefe/frontend/src/components/VAudioDetails/VAudioDetails.vue#L61C1-L63). The audio page should be updated to match the image page: the foreign_landing_url link should be added to the “source”, not provider.

## Additional context

It is possible to get the link for the `source` using the provider store, but we do not have the `provider` urls on the frontend. So, for Smithsonian museums, we only have the link to the individual museums (`source`) but not Smithsonian itself (`provider`)

PR #2846, opened 2023-08-19
Combine `VAudioDetails` and `VImageDetails` into `VMediaDetails` component

obulat:
## Fixes

Related to #2731 by @fcoveram

## Description

To make the updates to the metadata section of the single result page easier, this PR replaces the separate `VAudioDetails` and `VImageDetails` components with a single `VMediaDetails` component. The metadata (source, provider, dimensions for images and sample rates etc for audio) are prepared in a new utility function in `metadata.ts`. This function creates a list of the title i18n keys, values and urls (if available) for the media metadata such as source/creator/file size/file format etc.

This PR also moves the tag filtering (to filter out empty tags) from the components to the `decodeMediaData` function that prepares each media item from the API response to be displayed on the frontend.

Some changes to the `metadata.ts` and `VMetadata` in this PR will be improved on in #2851. Here, the effort was made to extract the components without affecting the snapshots, as much as possible

## Testing Instructions

Run the app and see that the single result metadata looks correct, and works both on the audio and image pages.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/pages/audio/_id/index.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <main :id="skipToContentTargetId" tabindex="-1" class="relative flex-grow">
-    <VSafetyWall v-if="isHidden" :media="audio" @reveal="reveal" />
-    <template v-else>
-      <VSingleResultControls v-if="audio" :media="audio" />
-      <template v-if="audio">
+    <template v-if="audio">
```
obulat, line 3: These changes are necessary for adding the pages to `tsconfig`

frontend/src/types/media.ts
```diff
@@ -122,3 +122,10 @@ export const isMediaDetail = <T extends SupportedMediaType>(
 ): media is DetailFromMediaType<T> => {
   return !!media && media.frontendMediaType === mediaType
 }
+
+export type Metadata = {
+  label: string
+  url?: string
+  value: string
+  isSource?: boolean
```
dhruvkb, line 130: This field seems unused. I couldn't find any other occurrence of `isSource` in the PR.

frontend/src/utils/metadata.ts
```diff
@@ -0,0 +1,107 @@
...
+    return audio.filetype ?? i18n.t("mediaDetails.information.unknown")
+  const altFormats = audio.alt_files.map((altFile) => altFile.filetype)
+  if (audio.filetype) {
+    altFormats.unshift(audio.filetype)
+  }
+  const uniqueFormats = new Set(altFormats)
+  return [...uniqueFormats].join(", ")
+}
+
+export const getMediaMetadata = (
```
dhruvkb, line 30: If this function is branched over image and audio it might be clearer if it called one of two separate functions depending on the media type.

frontend/src/utils/metadata.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+  return i18n.t("mediaDetails.information.unknown")
+}
+
+const getAudioType = (audio: AudioDetail, i18n: NuxtI18nInstance) => {
+  if (!audio.alt_files)
+    return audio.filetype ?? i18n.t("mediaDetails.information.unknown")
+  const altFormats = audio.alt_files.map((altFile) => altFile.filetype)
+  if (audio.filetype) {
+    altFormats.unshift(audio.filetype)
```
dhruvkb, line 24: The subsequent conversion to `Set` (unordered) and back to `Array` negates the use of `unshift` to insert at the start of the array.

frontend/src/types/media.ts
```diff
@@ -122,3 +122,10 @@ export const isMediaDetail = <T extends SupportedMediaType>(
 ): media is DetailFromMediaType<T> => {
   return !!media && media.frontendMediaType === mediaType
 }
+
+export type Metadata = {
+  label: string
```
dhruvkb, line 127: It seems as if `label` is always passed through `i18n.t()` so we could simplify the `metadata.ts` file if we move these calls to the `VMetadata.vue` component inside `v-for`.

frontend/src/components/VMediaInfo/VMediaTags.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <ul v-if="filteredTags.length" class="flex flex-wrap gap-2">
+    <VMediaTag v-for="(tag, index) in filteredTags" :key="index" tag="li">{{
+      tag.name
+    }}</VMediaTag>
+  </ul>
```
dhruvkb, line 6: I don't mind but I want to know why this is a separate component. Is there potential to reuse it in other places?

dhruvkb (approved):
I'm all for merging as many components into one as possible so I support this PR. 

I have some non-blocking suggestions and a request to add some stories for the new `VMetadata.vue` component.

frontend/src/types/media.ts
obulat: It's used in the follow-up PR, #2851, to determine what analytics event to send.

frontend/src/utils/metadata.ts
obulat: Great catch! I'll update this.

frontend/src/utils/metadata.ts
obulat: I've just converted what we were doing in the `VAudioDetails` and `VImageDetails` components to a util function here. It's improved (and synced between image and audio metadata) in the follow-up PR.

frontend/src/components/VMediaInfo/VMediaTags.vue
obulat: I think I extracted it before combining the `VAudioDetails` and `VImageDetails` when trying to add the external icon links to the source links in the metadata section. Then, this PR grew and changed, so this component might not be necessary anymore.
It does remove the `filteredTag` function from the parent component, making it a bit smaller. Although, in the future, the tags should be cleaned up in the catalog, and this function will not be necessary anymore.

obulat: I moved the tag filtering to `decodeMediaData`, where all media items from the API are cleaned for the frontend.

frontend/src/components/VMediaInfo/VMediaDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+    VMediaTags,
+    VMetadata,
+    VContentReportPopover,
+  },
+  props: {
+    media: {
+      type: Object as PropType<AudioDetail | ImageDetail>,
+      required: true,
+    },
+    imageWidth: {
```
AetherUnbound, line 49: `image` here makes me think that this might be specific to image media, although I know that the album art is likely used here for audio records. Does maybe `previewWidth` or `mediaWidth` make more sense, to make this as generic as possible? Or should we stick with the `image` prefix?

AetherUnbound (approved):
Looks good to me, and the pages render as I would expect them to! Just one note about language for an attribute, but nothing to block a merge.

frontend/src/components/VMediaInfo/VMediaDetails.vue
obulat: It's actually the property that's specific for the images. All of the audio thumbnails are square. Some images, on the other hand, don't have the values for width, height and image type in the API result. So, we send a head request to the provider, and update these values before displaying them. This is why the values are set separately here.

PR #2848, opened 2023-08-21
Fix nginx redirect for no trailing slash

### LINKED ISSUE #728: Path `/static` redirects to port 8080
## Description
Carrying out research on each of the API microservices and seeing what the installation process is for each of them, I have noticed a path called `static` which is located in `/static` (trying to understand what content it loads).

I tried to go to this address via "api.openverse.engineering/static" and found that the page redirects to https://api.openverse.engineering:8080/static/

## Reproduction
Open your terminal and run the command:

```sh
curl --location -I https://api.openverse.engineering/static
```

You'll get something like this:

```
HTTP/2 301
date: Tue, 01 Mar 2022 21:59:03 GMT
content-type: text/html
location: http://api.openverse.engineering:8080/static/
cache-control: max-age=14400
cf-cache-status: MISS
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
strict-transport-security: max-age=15552000; includeSubDomains; preload
x-content-type-options: nosniff
server: cloudflare
cf-ray: 6e552854fb8fef0a-MIA
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400

HTTP/1.1 301 Moved Permanently
Date: Tue, 01 Mar 2022 21:59:03 GMT
Connection: keep-alive
Cache-Control: max-age=3600
Expires: Tue, 01 Mar 2022 22:59:03 GMT
Location: https://api.openverse.engineering/static/
X-Content-Type-Options: nosniff
Server: cloudflare
CF-RAY: 6e552856c8249aba-MIA
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400

HTTP/2 403
date: Tue, 01 Mar 2022 21:59:03 GMT
content-type: text/html; charset=utf-8
cf-cache-status: BYPASS
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
strict-transport-security: max-age=15552000; includeSubDomains; preload
x-content-type-options: nosniff
server: cloudflare
cf-ray: 6e5528574d51ef0a-MIA
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400
```

You can noticed at the first stage, the page redirects via status code 301 to location: `http://api.openverse.engineering:8080/static/`

## Expectation
The content should load without the port 8080?

---

EDITED: I can confirm the path `/media` has the same behavior.

sarayourfriend:
## Fixes

Fixes #728 by @rbadillap 

## Description

There are three main parts to this PR: one direct change, one related change, and another change to support testing.

1. The changes to `docker-compose.yml` add a new `nginx` service that uses the `nginx` image we use in production created using `api/Dockerfile`. This is necessary for testing the changes in this PR and indeed, for easily verifying _any_ changes to `api/nginx.conf.template`.
2. The change to `location /static` to add a trailing slash fixes a small issue and potential (though unlikely) security/misconfiguration issue. Adding a trailing slash `/static/` prevents the rule from matching `/static-somethingelsewedidnotintend`. Likewise, removing the unused `/media` location also prevents potentially obscure security issues. Best to just _not allow_ anything we don't intend. I learned about this while researching this issue and reading about how the `alias` and `root` directives work. **A side effect of this change is that testing now needs to occur against `/static/admin` because `/static` won't match any configured locations and 404 instead of redirecting.**
3. Turning off `absolute_redirects` directly fixes the problem described in the issue. The behaviour in the issue was caused by nginx attempting to redirect `/static` with a missing trailing slash to `/static/`. However, because `port_in_redirect` was not off, it would inject the `:8080` port into the location on the redirect. Hence, `api.openverse.engineering/static` redirecting to `api.openverse.engineering:8080/static/`. Turning off `port_in_redirect` would fix the issue in production, but introduces a difficulty with testing locally, because it would then _strip_ the port, causing local testing requests to `0.0.0.0:50270/static` to redirect to `0.0.0.0/static/` without the original port. I couldn't find a good approach for this that didn't require heaps of additional configuration. Turning off absolute redirects does the trick perfectly in a way that is testable locally and will work exactly the same in production. [Relative redirects used to not be supported but are fine these days](https://stackoverflow.com/questions/8250259/is-a-302-redirect-to-relative-url-valid-or-invalid/25643550#25643550), which is why the default of the absolute redirect directive preserves previous behaviour to use absolute redirects.

## Testing Instructions

Checkout this branch and run `just api/up`. This will cause the nginx image to build and be accessible as a proxy to Django at `localhost:50270`. Test getting static files by going to `/admin` and logging in and poking around. Next, test the redirect: `curl --location -I http://localhost:50270/static/admin`. Confirm the redirected location is relative (does not include the port) and matches `/static/admin/` with a trailing slash. `/static/admin/` will still give you a 403 forbidden because directory navigation is not enabled in nginx (intended). You can confirm the original behaviour by commenting out the new `absolute_redirect` rule in `api/nginx.conf.template` and running `just dc build nginx && just api/up` to get the new configuration. Run the `curl` command above again and you should see a redirect to `http://0.0.0.0:8080/static/admin/` instead of just the path. You can also go ahead and try to `port_in_redirect off` approach to see the issues with it stripping all ports, rather than just not replacing the port with Nginx's (i.e., it won't preserve the port used for local requests against the docker-compose stack).

Any time you make configuration changes to nginx, you must run `just dc build nginx` to get the new configuration: it does not auto-refresh the configuration. Run `just api/up` and Docker compose will replace the running service with the new image and preserve everything else. This is generally a pretty fast process.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/nginx.conf.template
+++ b/api/nginx.conf.template
@@ -47,13 +47,10 @@ server {
     charset        utf-8;
     client_max_body_size 75M;
     error_page 500 /500.json;
+    absolute_redirect off;
 
-    location /static {
-        alias /app/static;
-    }
-
-    location /media {
-        alias /app/media;
+    location /static/ {
+        alias /app/static/;
     }
 
     location / {
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -290,6 +290,22 @@ services:
     stdin_open: true
     tty: true
 
+  nginx:
+    profiles:
+      - api
+    build:
+      context: ./api/
+      target: nginx
+      args: # Automatically inferred from env vars, unless specified
+        - SEMANTIC_VERSION=${SEMANTIC_VERSION:-v1.0.0}
+        - API_PY_VERSION
+    ports:
+      - "50270:8080"
+    environment:
+      DJANGO_NGINX_UPSTREAM_URL: web:8000
+    depends_on:
+      - web
+
   proxy:
     profiles:
       - api


dhruvkb (commented):
I'm slightly confused as to why `nginx` is a separate service and not merged into the existing `proxy` service? 

Also is this NGINX conf https://github.com/WordPress/openverse/blob/23d3972aad42aeed1e72fb604fe7e2b243f2a4c6/docker/nginx/templates/web.conf.template redundant?

sarayourfriend:
At first I was going to merge the two, but the existing `proxy` one is used for testing HTTPS setup, which we don't use in production. If we wanted to merge the two then the production configuration would need to become more complex to support a minor use-case that we rarely, if ever, actually use. I'd like to avoid that.

But now that you've mentioned it, one thing we could do is map the HTTPS config in the docker-compose, because Nginx supports concatenating multiple configuration files as long as they're in `/etc/nginx/conf.d`. It would just take extra work to get that `web.conf.template` you linked ready to work alongside the configuration we use in production. I would prefer doing that in a separate issue, however, or removing the `proxy` service entirely. I've suggested that we could switch to using [caddy to test local HTTPS](https://caddyserver.com/docs/automatic-https#local-https) because it would be a bit easier overall. Switching to that would be a different issue, however.

In short: we could try to combine the two, but it could take a bit of effort to make the two configurations work together without making the production configuration unnecessarily more complicated. Alternatively, we could take a different approach to testing HTTPS locally altogether.

stacimc (approved):
🚀 Thanks for the great testing instructions -- worked seamlessly for me 👍

PR #2853, opened 2023-08-21
Add API routes and controllers for additional search views

### LINKED ISSUE #2849: Add additional search views endpoints to the API
## Problem

API does not allow for getting media from specific source/creator or that has a specific tag.

## Description

Description in the Additional search views Implementation plan: https://docs.openverse.org/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.html#new-api-endpoints

3 new endpoints should be added, both for audio and images:
- /tag/<str:tag>/
- /source/<str:source>/
- /source/<str:source>/creator/<str:creator>/

## Alternatives

See the [Implementation plan](https://docs.openverse.org/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.html) for the discussion of alternatives

### LINKED ISSUE #2850: Create a controller for Additional search views
## Problem

Search controller in the API cannot return the results necessary for additional search views project.

## Description

The `collection` controller should re-use some of the `search` controller functionality, but the query should be constructed differently:
- only the specific `tag`, `source` or `source`+`creator` filters should be applied. No search query with `q` or search within `title`/`description`/`tag` is necessary.
- items should be sorted by `created_on`.
- `tag` view should probably also use the popularity and authority boosting for ranking the results.

## Additional context

The additional search views project implementation plan: https://docs.openverse.org/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.html

obulat:
## Fixes

Fixes #2849 by @obulat
Fixes #2850 by @obulat

## Note
This PR was changed a lot after reviews, mainly removing the query parameters, so I wrote a new description in [this comment](https://github.com/WordPress/openverse/pull/2853#issuecomment-1747352637). The old description is left here for the record.

## Description

This PR adds the endpoints for additional search views for both audio and image. The endpoints start with `/v1/images` or `/v1/audio`:
- **Tag** view: `/tag/tag_name`
- **Source** view: `/source/source_name`
- **Creator** view: `/source/source_name/creator/creator_name`.
These endpoints return paginated results because they accept `page` query parameter. They also support query parameters such as `license`, `license_type`, `category`,  [image-only] `aspect_ratio`, `size`, [audio-only] `length`, `peaks`. Currently, the frontend plans and designs do not allow for querying the collections using these filter parameters, but we can add this feature later.

To enable this feature, I had to refactor the media serializers and the search controller.

### Search controller
 I extracted functions that are common for both the collection and search requests.
Both the `collection`s and the `search` endpoints now call the `search_controller.query_media` function. `query_media` builds the relevant query (collection or search), sets up other search parameters, executes the search, tallies the results and builds the search context.
The new `build_collection_query` creates a query from a dictionary of the possible query predicates (`filter`, `must_not`, `should` and `must` dictionaries). All of these dictionaries are then combined with a `Bool` query. This makes the query much less nested than the query that we currently create in the `search` method in the `search_controller`.

### Serializers
This PR adds a `MediaCollectionRequestSerializer` to allow for filtering the collections using the filters such as `license` or `category`. This would allow us to make large collections easier to browse by various criteria such as the license or category.

This new serializer reuses some common functionality from the `MediaSearchRequestSerializer`. To make it possible, I had to extract the common functionality into `MediaListRequestSerializer`.

### Sorting of the results
The results are sorted by `created_on`. The tag results also add the boost for `unstable__authority` (for the authorized requests).

## Testing Instructions

Run the app using `just up`.
Try various collection routes:
1. http://localhost:50280/v1/images/source/stocksnap (source collection, only the sources that are present in the sample data can be used to test locally)
1. http://localhost:50280/v1/images/source/flickr/creator/Manzabar/ (creator collection)
1. http://localhost:50280/v1/images/tag/cat (tag collection, images)
1. http://localhost:50280/v1/audio/tag/birds (tag collection, audio)
1. http://localhost:50280/v1/audio/tag/birds/?extension=flac (additional query parameter)
1. http://localhost:50280/v1/images/source/flickr/?page=2 (second page of results)

Try the routes that should return errors:
1. http://localhost:50280/v1/images/source/met (We don't have anything from `met` in the sample data, so the source validator throws an error
1. http://localhost:50280/v1/images/source/flickr,stocksnap (The path parameters are not split by a comma, so only a single parameter is allowed)

If there are no items corresponding to the requested values, the empty collection can be returned:
http://localhost:50280/v1/images/source/flickr/creator/Manzaba/ (the creator name is missing the last `r` letter)

Since the image and audio search had to be modified, too, check that they work as expected (the CI should check that, too).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import logging as log
 from math import ceil
-from typing import Literal
+from typing import TYPE_CHECKING
 
 from django.conf import settings
 from django.core.cache import cache
@@ -15,7 +15,8 @@ from elasticsearch_dsl.query import EMPTY_QUERY
 from elasticsearch_dsl.response import Hit, Response
 
 import api.models as models
-from api.constants.media_types import OriginIndex
+from api.constants.media_types import OriginIndex, SearchIndex
+from api.constants.search import SearchStrategy
 from api.constants.sorting import INDEXED_ON
 from api.controllers.elasticsearch.helpers import (
     ELASTICSEARCH_MAX_RESULT_WINDOW,
@@ -23,12 +24,17 @@ from api.controllers.elasticsearch.helpers import (
     get_query_slice,
     get_raw_es_response,
 )
-from api.serializers import media_serializers
 from api.utils import tallies
 from api.utils.check_dead_links import check_dead_links
 from api.utils.dead_link_mask import get_query_hash
 from api.utils.search_context import SearchContext
 
+if TYPE_CHECKING:
+    from api.serializers.audio_serializers import AudioCollectionRequestSerializer
+    from api.serializers.media_serializers import (
+        MediaSearchRequestSerializer,
+        PaginatedRequestSerializer,
+    )
 
 module_logger = logging.getLogger(__name__)
 
@@ -172,24 +178,24 @@ def get_excluded_providers_query() -> Q | None:
     return None
 
 
-def _resolve_index(
-    index: Literal["image", "audio"],
-    search_params: media_serializers.MediaSearchRequestSerializer,
-) -> Literal["image", "image-filtered", "audio", "audio-filtered"]:
-    use_filtered_index = all(
-        (
-            settings.ENABLE_FILTERED_INDEX_QUERIES,
-            not search_params.validated_data["include_sensitive_results"],
-        )
-    )
-    if use_filtered_index:
-        return f"{index}-filtered"
+def get_index(
+    exact_index: bool,
+    origin_index: OriginIndex,
+    search_params: MediaListRequestSerializer,
+) -> SearchIndex:
+    if exact_index:
+        return origin_index
 
-    return index
+    include_sensitive_results = search_params.validated_data.get(
+        "include_sensitive_results", False
+    )
+    if settings.ENABLE_FILTERED_INDEX_QUERIES and not include_sensitive_results:
+        return f"{origin_index}-filtered"
+    return origin_index
 
 
 def create_search_filter_queries(
-    search_params: media_serializers.MediaSearchRequestSerializer,
+    search_params: MediaListRequestSerializer,
 ) -> dict[str, list[Q]]:
     """
     Create a list of Elasticsearch queries for filtering search results.
@@ -230,7 +236,7 @@ def create_search_filter_queries(
 
 
 def create_ranking_queries(
-    search_params: media_serializers.MediaSearchRequestSerializer,
+    search_params: MediaListRequestSerializer,
 ) -> list[Q]:
     queries = [Q("rank_feature", field="standardized_popularity", boost=DEFAULT_BOOST)]
     if search_params.data["unstable__authority"]:
@@ -240,8 +246,8 @@ def create_ranking_queries(
     return queries
 
 
-def create_search_query(
-    search_params: media_serializers.MediaSearchRequestSerializer,
+def build_search_query(
+    search_params: MediaListRequestSerializer,
 ) -> Q:
     # Apply filters from the url query search parameters.
     url_queries = create_search_filter_queries(search_params)
@@ -315,8 +321,49 @@ def create_search_query(
     )
 
 
-def search(
-    search_params: media_serializers.MediaSearchRequestSerializer,
+def build_collection_query(
+    search_params: MediaListRequestSerializer,
+    collection_params: dict[str, str],
+):
+    """
+    Build the query to retrieve items in a collection.
+    :param collection_params: `tag`, `source` and/or `creator` values from the path.
+    :param search_params: the validated search parameters.
+    :return: the search client with the query applied.
+    """
+    search_query = {"filter": [], "must": [], "should": [], "must_not": []}
+    # Apply the term filters. Each tuple pairs a filter's parameter name in the API
+    # with its corresponding field in Elasticsearch. "None" means that the
+    # names are identical.
+    filters = [
+        # Collection filters allow a single value.
+        ("tag", "tags.name"),
+        ("source", None),
+        ("creator", None),
+    ]
+    for serializer_field, es_field in filters:
+        if serializer_field in collection_params:
+            if not (argument := collection_params.get(serializer_field)):
+                continue
+            parameter = es_field or serializer_field
+            search_query["filter"].append({"term": {parameter: argument}})
+
+    # Exclude mature content and disabled sources
+    include_sensitive_by_params = search_params.validated_data.get(
+        "include_sensitive_results", False
+    )
+    if not include_sensitive_by_params:
+        search_query["must_not"].append({"term": {"mature": True}})
+    if excluded_providers_query := get_excluded_providers_query():
+        search_query["must_not"].append(excluded_providers_query)
+
+    return Q("bool", **search_query)
+
+
+def query_media(
+    strategy: SearchStrategy,
+    search_params: MediaListRequestSerializer,
+    collection_params: dict[str, str] | None,
     origin_index: OriginIndex,
     exact_index: bool,
     page_size: int,
@@ -325,10 +372,17 @@ def search(
     page: int = 1,
 ) -> tuple[list[Hit], int, int, dict]:
     """
-    Perform a ranked paginated search from the set of keywords and, optionally, filters.
+    If ``strategy`` is ``search``, perform a ranked paginated search
+    from the set of keywords and, optionally, filters.
+    If `strategy` is `collection`, perform a paginated search
+    for the `tag`, `source` or `source` and `creator` combination.
 
-    :param search_params: Search parameters. See
-     :class: `ImageSearchQueryStringSerializer`.
+    :param collection_params: The path parameters for collection search, if
+    strategy is `collection`.
+    :param strategy: Whether to perform a default search or retrieve a collection.
+    :param search_params: If `strategy` is `collection`, `PaginatedRequestSerializer`
+    or `AudioCollectionRequestSerializer`. If `strategy` is `search`, search
+    query params, see :class: `MediaRequestSerializer`.
     :param origin_index: The Elasticsearch index to search (e.g. 'image')
     :param exact_index: whether to skip all modifications to the index name
     :param page_size: The number of results to return per page.
@@ -337,46 +391,54 @@ def search(
     Elasticsearch shards.
     :param filter_dead: Whether dead links should be removed.
     :param page: The results page number.
-    :return: Tuple with a List of Hits from elasticsearch, the total count of
+    :return: Tuple with a list of Hits from elasticsearch, the total count of
     pages, the number of results, and the ``SearchContext`` as a dict.
     """
-    if not exact_index:
-        index = _resolve_index(origin_index, search_params)
+    index = get_index(exact_index, origin_index, search_params)
+
+    if strategy == "collection":
+        query = build_collection_query(search_params, collection_params)
     else:
-        index = origin_index
+        query = build_search_query(search_params)
 
-    s = Search(index=index)
+    s = Search(index=index).query(query)
 
-    search_query = create_search_query(search_params)
-    s = s.query(search_query)
+    if strategy == "search":
+        # Use highlighting to determine which fields contribute to the selection of
+        # top results.
+        s = s.highlight(*DEFAULT_SEARCH_FIELDS)
+        s = s.highlight_options(order="score")
+        s.extra(track_scores=True)
 
-    # Use highlighting to determine which fields contribute to the selection of
-    # top results.
-    s = s.highlight(*DEFAULT_SEARCH_FIELDS)
-    s = s.highlight_options(order="score")
-    s.extra(track_scores=True)
     # Route users to the same Elasticsearch worker node to reduce
     # pagination inconsistencies and increase cache hits.
     # TODO: Re-add 7s request_timeout when ES stability is restored
     s = s.params(preference=str(ip))
 
-    # Sort by new
-    if search_params.validated_data["sort_by"] == INDEXED_ON:
-        s = s.sort({"created_on": {"order": search_params.validated_data["sort_dir"]}})
+    # Sort by `created_on` if the parameter is set or if `strategy` is `collection`.
+    sort_by = search_params.validated_data.get("sort_by")
+    sort_dir = search_params.validated_data.get("sort_dir", "desc")
+    if strategy == "collection" or sort_by == INDEXED_ON:
+        s = s.sort({"created_on": {"order": sort_dir}})
 
-    # Paginate
-    start, end = get_query_slice(s, page_size, page, filter_dead)
-    s = s[start:end]
-    search_response = get_es_response(s, es_query="search")
-
-    results = _post_process_results(
-        s, start, end, page_size, search_response, filter_dead
+    # Execute paginated search and tally results
+    page_count, result_count, results = execute_search(
+        s, page, page_size, filter_dead, index, es_query=strategy
     )
 
-    result_count, page_count = _get_result_and_page_count(
-        search_response, results, page_size, page
-    )
+    result_ids = [result.identifier for result in results]
+    search_context = SearchContext.build(result_ids, origin_index)
 
+    return results, page_count, result_count, search_context.asdict()
+
+
+def tally_results(
+    index: SearchIndex, results: list[Hit] | None, page: int, page_size: int
+) -> None:
+    """
+    Tally the number of the results from each provider in the results
+    for the search query.
+    """
     results_to_tally = results or []
     max_result_depth = page * page_size
     if max_result_depth <= 80:
@@ -405,13 +467,33 @@ def search(
         # check things like provider density for a set of queries.
         tallies.count_provider_occurrences(results_to_tally, index)
 
-    if not results:
-        results = []
 
-    result_ids = [result.identifier for result in results]
-    search_context = SearchContext.build(result_ids, origin_index)
+def execute_search(
+    s: Search,
+    page: int,
+    page_size: int,
+    filter_dead: bool,
+    index: SearchIndex,
+    es_query: str,
+) -> tuple[int, int, list[Hit]]:
+    """
+    Execute search for the given query slice, post-processes the results,
+    and returns the results and result and page counts.
+    """
+    start, end = get_query_slice(s, page_size, page, filter_dead)
+    s = s[start:end]
 
-    return results, page_count, result_count, search_context.asdict()
+    search_response = get_es_response(s, es_query=es_query)
+
+    results: list[Hit] = (
+        _post_process_results(s, start, end, page_size, search_response, filter_dead)
+        or []
+    )
+    result_count, page_count = _get_result_and_page_count(
+        search_response, results, page_size, page
+    )
+    tally_results(index, results, page, page_size)
+    return page_count, result_count, results
 
 
 def get_sources(index):
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -347,7 +347,7 @@ def build_collection_query(
         # Collection filters allow a single value.
         ("tag", "tags.name.keyword"),
         ("source", None),
-        ("creator", None),
+        ("creator", "creator.keyword"),
     ]
     for serializer_field, es_field in filters:
         if serializer_field in collection_params:
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -369,6 +369,16 @@ def build_collection_query(
     return Q("bool", **search_query)
 
 
+def build_query(
+    strategy: SearchStrategy,
+    search_params: MediaListRequestSerializer,
+    collection_params: dict[str, str] | None,
+) -> Q:
+    if strategy == "collection":
+        return build_collection_query(search_params, collection_params)
+    return build_search_query(search_params)
+
+
 def query_media(
     strategy: SearchStrategy,
     search_params: MediaListRequestSerializer,
@@ -405,10 +415,7 @@ def query_media(
     """
     index = get_index(exact_index, origin_index, search_params)
 
-    if strategy == "collection":
-        query = build_collection_query(search_params, collection_params)
-    else:
-        query = build_search_query(search_params)
+    query = build_query(strategy, search_params, collection_params)
 
     s = Search(index=index).query(query)
 
@@ -426,8 +433,8 @@ def query_media(
 
     # Sort by `created_on` if the parameter is set or if `strategy` is `collection`.
     sort_by = search_params.validated_data.get("sort_by")
-    sort_dir = search_params.validated_data.get("sort_dir", "desc")
     if strategy == "collection" or sort_by == INDEXED_ON:
+        sort_dir = search_params.validated_data.get("sort_dir", "desc")
         s = s.sort({"created_on": {"order": sort_dir}})
 
     # Execute paginated search and tally results
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -354,6 +354,7 @@ def build_collection_query(
     )
     if not include_sensitive_by_params:
         search_query["must_not"].append({"term": {"mature": True}})
+
     if excluded_providers_query := get_excluded_providers_query():
         search_query["must_not"].append(excluded_providers_query)
 
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -29,11 +29,19 @@ from api.utils.check_dead_links import check_dead_links
 from api.utils.dead_link_mask import get_query_hash
 from api.utils.search_context import SearchContext
 
+
+# Using TYPE_CHECKING to avoid circular imports when importing types
 if TYPE_CHECKING:
     from api.serializers.audio_serializers import AudioCollectionRequestSerializer
     from api.serializers.media_serializers import (
         MediaSearchRequestSerializer,
         PaginatedRequestSerializer,
+    )
+
+    MediaListRequestSerializer = (
+        AudioCollectionRequestSerializer
+        | MediaSearchRequestSerializer
+        | PaginatedRequestSerializer
     )
 
 module_logger = logging.getLogger(__name__)
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -337,7 +337,7 @@ def build_collection_query(
     # names are identical.
     filters = [
         # Collection filters allow a single value.
-        ("tag", "tags.name"),
+        ("tag", "tags.name.keyword"),
         ("source", None),
         ("creator", None),
     ]
--- a/api/api/docs/base_docs.py
+++ b/api/api/docs/base_docs.py
@@ -140,11 +140,27 @@ def get_collection_description(media_type, collection):
         return f"""
 Get a collection of {media_type} with a specific tag.
 
-This endpoint returns only the exact matches, case-insensitive matches for the
-specified tag. For example, 'birds' and 'birding' are not matches for 'bird'.
-To search within the tag values, or to match several tags, use the `search` endpoint
-with `tags` query parameter instead of `q` parameter. In this case, the matches will
- not be exact, so 'cat' would match both 'cat' and 'cats'.
+This endpoint matches a single tag, exactly and entirely.
+
+Differences that will cause tags to not match are:
+- upper and lower case letters
+- diacritical marks
+- hyphenation
+- spacing
+- multi-word tags where the query is only one of the words in the tag
+- multi-word tags where the words are in a different order
+
+Examples of tags that **do not** match:
+- "Low-Quality" and "low-quality"
+- "jalapeño" and "jalapeno"
+- "Saint Pierre des Champs" and "Saint-Pierre-des-Champs"
+- "dog walking" and "dog  walking" (where the latter has two spaces between the
+last two words, as in a typographical error)
+- "runner" and "marathon runner"
+- "exclaiming loudly" and "loudly exclaiming"
+
+For non-exact or multi-tag matching, using the `search` endpoint's `tags` query
+parameter.
 
 The returned results are ordered based on the time when they were added to Openverse.
     """
--- a/api/api/docs/base_docs.py
+++ b/api/api/docs/base_docs.py
@@ -140,13 +140,13 @@ def get_collection_description(media_type, collection):
         return f"""
 Get a collection of {media_type} with a specific tag.
 
-This endpoint returns only the exact matches. To search within the
-tag values, or to match several tags, use the `search` endpoint
-with `tags` query parameter instead of `q` parameter.
+This endpoint returns only the exact matches, case-insensitive matches for the
+specified tag. For example, 'birds' and 'birding' are not matches for 'bird'.
+To search within the tag values, or to match several tags, use the `search` endpoint
+with `tags` query parameter instead of `q` parameter. In this case, the matches will
+ not be exact, so 'cat' would match both 'cat' and 'cats'.
 
-The returned results are ordered primarily based on their popularity
-and authority. However, note that the exact order may vary over time
-or across requests.
+The returned results are ordered based on the time when they were added to Openverse.
     """
     elif collection == "source":
         return f"""
@@ -168,8 +168,7 @@ this endpoint only returns the items from the specified source. To search within
 the creator value, use the `search` endpoint with `source` query parameter
 instead of `q`.
 
-The order in the results is not guaranteed to stay the same. Most likely, the images
-in the collection will be sorted by the order in which they were added to Openverse.
+The items will be sorted by the date when they were added to Openverse.
     """
 
 
--- a/api/api/docs/base_docs.py
+++ b/api/api/docs/base_docs.py
@@ -1,8 +1,23 @@
 from http.client import responses as http_responses
 from textwrap import dedent
+from typing import Literal
 
-from drf_spectacular.openapi import AutoSchema, OpenApiResponse
-from drf_spectacular.utils import OpenApiExample, extend_schema
+from drf_spectacular.openapi import AutoSchema
+from drf_spectacular.utils import (
+    OpenApiExample,
+    OpenApiParameter,
+    OpenApiResponse,
+    extend_schema,
+)
+
+from api.constants.media_types import MediaType
+from api.serializers.audio_serializers import (
+    AudioCollectionRequestSerializer,
+    AudioSerializer,
+)
+from api.serializers.error_serializers import NotFoundErrorSerializer
+from api.serializers.image_serializers import ImageSerializer
+from api.serializers.media_serializers import PaginatedRequestSerializer
 
 
 def fields_to_md(field_names):
@@ -77,3 +92,123 @@ class MediaSchema(AutoSchema):
             else:
                 operation_tokens.append("detail")
         return "_".join(operation_tokens)
+
+
+source_404_message = "Invalid source 'name'. Valid sources are ..."
+source_404_response = OpenApiResponse(
+    NotFoundErrorSerializer,
+    examples=[
+        OpenApiExample(
+            name="404",
+            value={"detail": source_404_message},
+        )
+    ],
+)
+
+
+def build_source_path_parameter(media_type: MediaType):
+    valid_description = (
+        f"Valid values are source_names from the stats endpoint: "
+        f"https://api.openverse.engineering/v1/{media_type}/stats/."
+    )
+
+    return OpenApiParameter(
+        name="source",
+        type=str,
+        location=OpenApiParameter.PATH,
+        description=f"The source of {media_type}. {valid_description}",
+    )
+
+
+creator_path_parameter = OpenApiParameter(
+    name="creator",
+    type=str,
+    location=OpenApiParameter.PATH,
+    description="The name of the media creator. This parameter "
+    "is case-sensitive, and matches exactly.",
+)
+tag_path_parameter = OpenApiParameter(
+    name="tag",
+    type=str,
+    location=OpenApiParameter.PATH,
+    description="The tag of the media. Not case-sensitive, matches exactly.",
+)
+
+
+def get_collection_description(media_type, collection):
+    if collection == "tag":
+        return f"""
+Get a collection of {media_type} with a specific tag.
+
+This endpoint returns only the exact matches. To search within the
+tag values, or to match several tags, use the `search` endpoint
+with `tags` query parameter instead of `q` parameter.
+
+The returned results are ordered primarily based on their popularity
+and authority. However, note that the exact order may vary over time
+or across requests.
+    """
+    elif collection == "source":
+        return f"""
+Get a collection of {media_type} from a specific source.
+
+This endpoint returns only the exact matches. To search within the source value,
+use the `search` endpoint with `source` query parameter.
+
+The results in the collection will be sorted by the order in which they
+were added to Openverse.
+    """
+    elif collection == "creator":
+        return f"""
+Get a collection of {media_type} by a specific creator from the specified source.
+
+This endpoint returns only the exact matches both on the creator and the source.
+Notice that a single creator's media items can be found on several sources, but
+this endpoint only returns the items from the specified source. To search within
+the creator value, use the `search` endpoint with `source` query parameter
+instead of `q`.
+
+The order in the results is not guaranteed to stay the same. Most likely, the images
+in the collection will be sorted by the order in which they were added to Openverse.
+    """
+
+
+COLLECTION_TO_OPERATION_ID = {
+    ("images", "source"): "images_by_source",
+    ("images", "creator"): "images_by_source_and_creator",
+    ("images", "tag"): "images_by_tag",
+    ("audio", "source"): "audio_by_source",
+    ("audio", "creator"): "audio_by_source_and_creator",
+    ("audio", "tag"): "audio_by_tag",
+}
+
+
+def collection_schema(
+    media_type: Literal["images", "audio"],
+    collection: Literal["source", "creator", "tag"],
+):
+    if media_type == "images":
+        request_serializer = PaginatedRequestSerializer
+        serializer = ImageSerializer
+    else:
+        request_serializer = AudioCollectionRequestSerializer
+        serializer = AudioSerializer
+
+    if collection == "tag":
+        responses = {200: serializer(many=True)}
+        path_parameters = [tag_path_parameter]
+    else:
+        responses = {200: serializer(many=True), 404: source_404_response}
+        path_parameters = [build_source_path_parameter(media_type)]
+        if collection == "creator":
+            path_parameters.append(creator_path_parameter)
+    operation_id = COLLECTION_TO_OPERATION_ID[(media_type, collection)]
+    description = get_collection_description(media_type, collection)
+    return extend_schema(
+        operation_id=operation_id,
+        summary=operation_id,
+        auth=[],
+        description=description,
+        responses=responses,
+        parameters=[request_serializer, *path_parameters],
+    )
--- a/api/api/serializers/audio_serializers.py
+++ b/api/api/serializers/audio_serializers.py
@@ -48,8 +48,8 @@ class AudioSearchRequestSerializer(
 ):
     """Parse and validate search query string parameters."""
 
-    fields_names = [
-        *MediaSearchRequestSerializer.fields_names,
+    field_names = [
+        *MediaSearchRequestSerializer.field_names,
         *AudioSearchRequestSourceSerializer.field_names,
         "category",
         "length",
--- a/api/api/serializers/audio_serializers.py
+++ b/api/api/serializers/audio_serializers.py
@@ -11,6 +11,7 @@ from api.serializers.media_serializers import (
     MediaReportRequestSerializer,
     MediaSearchRequestSerializer,
     MediaSerializer,
+    PaginatedRequestSerializer,
     get_hyperlinks_serializer,
     get_search_request_source_serializer,
 )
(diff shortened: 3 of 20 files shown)


api/api/serializers/media_serializers.py
```diff
@@ -334,6 +295,149 @@ def needs_db(self) -> bool:
...
+                f"Invalid source `{value}`. Allowed sources are: {allowed_sources}."
+            )
+        return source
+
+    def validate(self, data):
+        data = super().validate(data)
+        if "creator" in self.initial_data and "source" not in self.initial_data:
+            raise serializers.ValidationError(
+                "Cannot set `source` without `creator`. "
+                "Use both of these or neither of them."
```
stacimc, line 431:
```suggestion
                "Cannot set `creator` without `source`. "
```

I think that was backwards. `or neither of them` also implies to me that you also can't set source without creator, which isn't accurate.

api/api/serializers/media_serializers.py
```diff
@@ -334,6 +295,149 @@ def needs_db(self) -> bool:
...
+            )
+        if not data.get("source") and not data.get("creator") and not data.get("tag"):
+            raise serializers.ValidationError(
+                "At least one of `source`, `creator` or `tag` must be "
+                "set for a collection view."
```
stacimc, line 436: Is it possible to have `tag` at the same time as either of the others? Does that need to be validated?

stacimc (commented):
I think this generally looks good — all the tests in the PR description passed (although I have some questions below about additional query parameters). A few times while testing this I had an API request take a very long time (> 10 seconds), for example on a creator search and when getting /source/flickr, but I couldn’t reproduce it consistently and it seemed likely that it was a local env problem. Mentioning in case other folks also notice it.

For the creator collections: I noticed that some records have a _url _for `creator` (example sample data with id 040db469-9f73-4fed-b4d3-66592d554eb0, creator is `https://www.rosanetur.com`). Should those be handled any differently?

__

I don’t want to take up time rehashing things that were explained in the IP stage, but I was pretty confused about the intended behavior and how it differs from the existing search parameters. For example, how does searching http://localhost:50280/v1/images/source/flickr/creator/Manzabar differ from the existing http://localhost:50280/v1/images/?source=flickr&creator=Manzabar? Is it that the authority/popularity boost isn’t applied in the new route? Likewise, how does the tag collection differ from querying something like http://localhost:50280/v1/audio/?source=jamendo&tags=acoustic ? Is it mostly that the tags aren’t ‘fuzzy’ matched? (I played around and got results for ‘acoustical’, for example, when searching `?tags=acoustic`). Or is it a performance difference?

I also wasn’t clear what additional filters can be applied on the collection views. So for example:
* http://localhost:50280/v1/audio/source/freesound/?tags=spring appears to silently ignore the `tags` filter and just returns all Freesound results. (In contrast, http://localhost:50280/v1/audio/?source=freesound&tags=spring appears to work and filters by both)
* q is also ignored, so http://localhost:50280/v1/audio/source/freesound/?q=spring returns all Freesound results
* http://localhost:50280/v1/audio/source/freesound/?creator=KTManahan **does** work and returns the same results as http://localhost:50280/v1/audio/source/freesound/creator/KTManahan 
* Other filters like `extension` and `mature` work

I haven’t been involved in this project so I think I’m just missing a lot of context (especially, about how our own frontend actually consumes our API) — as I said, I don’t want to take up all your time time having things re-explained, sorry! But I think if  it’s unclear to me from the API documentation, it might be confusing to others as well? Could we update the `<media>_source / _source_creator / _tag` documentation to include (a) what additional query parameters are available for each of those routes and (b) how they differ from the equivalent `search`?

obulat:
Thank you for your review, @stacimc. Your comments about confusions due to lack of context are especially valuable because they help make the code and docs clearer - I can't see what isn't clear about it as I'm so immersed in the context now :)

> A few times while testing this I had an API request take a very long time (> 10 seconds), for example on a creator search and when getting /source/flickr, but I couldn’t reproduce it consistently and it seemed likely that it was a local env problem. Mentioning in case other folks also notice it.

I **also* noticed the requests taking a long time locally, so I guess this is something I should investigate.

> For the creator collections: I noticed that some records have a _url _for `creator` (example sample data with id 040db469-9f73-4fed-b4d3-66592d554eb0, creator is `https://www.rosanetur.com`). Should those be handled any differently?

For the creator collections, we want to show all of the items by the selected creator **in Openverse**, and also link to their external `url`, if it's available, with the "Open creator page" button:

Screenshot 2023-09-12 at 8 08 18 AM

> For example, how does searching http://localhost:50280/v1/images/source/flickr/creator/Manzabar differ from the existing http://localhost:50280/v1/images/?source=flickr&creator=Manzabar? Is it that the authority/popularity boost isn’t applied in the new route?

The existing filters match fuzzily by a stemmed value of the query. It's not so easy to understand with "Manzabar". A better example would be "photo": if you try searching for "photo" as creator, you will see that you'll get all items for which the creator has the words "photo" or "photos" in the "creator" field: https://api.openverse.engineering/v1/images/?source=flickr&creator=photo
The http://localhost:50280/v1/images/source/flickr/creator/photo endpoint will return only the exact matches for the creator whose name is "photo", and would not return images by "Obtuse Photo" or "JSmith Photo", as the current search using the query parameter does. The IP describes that if the user clicks on the creator button on the single result page, they would expect to see the media by _this specific_ creator, not by the creators that contain the word in their names:
Screenshot 2023-09-12 at 8 17 35 AM

I suppose that there will also be the performance difference, since the boolean filter check (whether `creator` is equal to "photo") should be faster than a search within the field (whether "photo" is contained in the `creator` field). That is a side effect for this change, though, with the main reason being the exact match.

> I also wasn’t clear what additional filters can be applied on the collection views.

The IP did not mention anything about the filters since the current designs do not show filtering, only displaying all of the items by the creator/source/tag. However, when I started making the changes in the API, I realized that we might want to filter the results (especially the Flickr images, for instance :) ). This is why I left filtering in the collections. 

At first, I thought that I should remove the creator/source/tag as filters since the views are already filtered by these parameters. I mainly wanted to have the licenses, categories, extensions, lengths, sensitivity as filters. This is why in your examples filtering by `tags` does not work. The `creator` filter (http://localhost:50280/v1/audio/source/freesound/?creator=KTManahan) wasn't supposed to work, oops...

After your comment, though, I realized that we might want to, for instance, filter the creator view by a tag, or filter a tag view by a source. I'll look into how I can add this to the code. If it turns out to require too many changes, this should probably be implemented in another PR.
Note that the filtering will not be available on the frontend as the designs  (and the IP) don't include implementation of filtering.

> I haven’t been involved in this project so I think I’m just missing a lot of context (especially, about how our own frontend actually consumes our API) — as I said, I don’t want to take up all your time time having things re-explained, sorry! But I think if it’s unclear to me from the API documentation, it might be confusing to others as well? Could we update the `<media>_source / _source_creator / _tag` documentation to include (a) what additional query parameters are available for each of those routes and (b) how they differ from the equivalent `search`?

As I said, these requests for clarification are very valuable, thank you! I'll add more documentation to make it clearer.

stacimc:
Thanks so much for your explanation, @obulat, now I understand why the new endpoints are needed :) 

> For the creator collections, we want to show all of the items by the selected creator in Openverse, and also link to their external url, if it's available, with the "Open creator page" button:

To clarify what I meant -- so for most records, the `creator` is a name and then `creator_url` is the external url (eg "John Smith", "www.provider-site/users/john-smith" or whatever). I had noticed that at least in our sample data we have some records where **both** `creator` and `creator_url` are urls. In the example I gave, `https://www.rosanetur.com` is the 'creator' for that record (and `creator_url` is some other url). So my question is, what does that look like for fetching that creator's collection? Since naively, it would be `http://localhost:50280/v1/images/source/flickr/creator/https://www.rosanetur.com/`. How should that be urlencoded?

api/api/serializers/audio_serializers.py
```diff
@@ -61,13 +64,29 @@ class AudioSearchRequestSerializer(
...
-        index = super().validate_internal__index(value)
-        if index is None:
-            return None
-        if not index.startswith(AUDIO_TYPE):
-            raise serializers.ValidationError(f"Invalid index name `{value}`.")
-        return index
+
+class AudioSearchRequestSerializer(
+    AudioListRequestSerializer,
+    AudioSearchRequestSourceSerializer,
```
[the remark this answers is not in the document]
obulat: Oops, I forgot to update the audio serializers. Updated in [f052720](https://github.com/WordPress/openverse/pull/2853/commits/f0527206444bbb455f0dd6389bc6ba38334b8287)

api/api/views/media_views.py
```diff
@@ -58,15 +65,30 @@ def get_queryset(self):
...
         context.update({"validated_data": req_serializer.validated_data})
         return context
 
-    def _get_request_serializer(self, request):
-        req_serializer = self.query_serializer_class(
-            data=request.query_params, context={"request": request}
+    def _get_request_serializer(
+        self,
+        request,
+        strategy: Literal["search", "collection"],
```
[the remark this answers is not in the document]
obulat: Extracted in [1f2a351](https://github.com/WordPress/openverse/pull/2853/commits/1f2a351b184af51ea630cf7291c822a8e53800c4). I'm not sure about the name, though. It is what the `strategy` for search view is set to, but since it describes two kinds of lists that we can get for media, I called it `ListView`. If you have a better suggestion, I would be happy to update.

obulat:
I rebased this PR on to #3039 to make the API documentation debugging easier.

api/api/serializers/media_serializers.py
obulat: Since the path parameters are checked for existence in the `view`s, this check is not necessary.

obulat:
> If there are creators with the same name within a source, then the endpoint will merge their works in the results, right? I'm okay with this being a first approach but then it will take more work to strictly comply with what the plan says.

I wrote the code with the assumption that a single source does not have any creators with identical names. I think this is a fair assumption: for instance, you probably cannot register with a name that is already taken on Flickr.
There could be creators with the same name on different providers. They may or may not be the same creators. For instance, different people could be using the name "Olga" on Stocksnap and Flickr. At the same time, Rembrandt's works may be published by different sources. But we cannot make assumptions that they are the same (or different) creators, so the only possible way is to show the media by a specific _creator_ on a specific _source_ site.

@krysal, I added _some_ tests for the collection parameters, but I'm not sure what kind of tests I can add for collections. Could you please elaborate on what kind of tests you think are necessary here?

> To clarify what I meant -- so for most records, the creator is a name and then creator_url is the external url (eg "John Smith", "www.provider-site/users/john-smith" or whatever). I had noticed that at least in our sample data we have some records where both creator and creator_url are urls. In the example I gave, https://www.rosanetur.com is the 'creator' for that record (and creator_url is some other url). So my question is, what does that look like for fetching that creator's collection? Since naively, it would be http://localhost:50280/v1/images/source/flickr/creator/https://www.rosanetur.com/. How should that be urlencoded?

Thank you so much for explaining this in detail, @stacimc! I wasn't aware that the creator names can be URLs or can have other special symbols. I updated the `creator` endpoint regex to allow for `/`. You can now query by such a creator: `localhost:50280/v1/images/source/flickr/creator/https://www.rosanetur.com/`, or even filter them by some filter parameter: `http://localhost:50280/v1/images/source/flickr/creator/https://www.rosanetur.com/?aspect_ratio=tall`

I updated the DRF spectacular documentation (e.g. http://localhost:50280/v1/#tag/images/operation/images_source_creator)
Each endpoint now documents the `path` parameters and the `query` parameters, and shows the sample responses. Please let me know if this is insufficient and how to update it.

api/api/controllers/search_controller.py
```diff
@@ -428,36 +478,78 @@ def search(
...
+        ("creator", "creator.keyword", False),
+        # The following filters allow multiple comma-separated values.
+        ("license", "license.keyword", True),
+        ("license_type", "license.keyword", True),
+        ("extension", None, True),
+        ("category", None, True),
+        ("categories", "category", True),
+        ("length", None, True),
+        ("aspect_ratio", None, True),
+        ("size", None, True),
```
stacimc, line 513: Should this include `peaks` as well? (If we want to keep filters?)

api/api/views/audio_views.py
```diff
@@ -48,6 +53,50 @@ def get_queryset(self):
...
 
+    @creator_collection
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="source/(?P<source>[^/.]+)/creator/(?P<creator>.+)",
+    )
+    def creator_collection(self, request, source, creator):
+        """
+        Get a collection of audio items by a specific creator from the specified source.
```
stacimc, line 64: Why are these overridden in the audio/image views? Is this supposed to override the less specific docstring ("collection of media items...") from `creator_collection`? In the [docs](http://localhost:50280/v1/#tag/audio/operation/audio_source_creator) I still see the `media` verbiage, and I don't see anything different in the definitions here.

api/api/serializers/media_serializers.py
```diff
@@ -334,6 +295,118 @@ def needs_db(self) -> bool:
...
+@extend_schema_serializer(
+    exclude_fields=[
+        "unstable__sort_by",
+        "unstable__sort_dir",
+        "unstable__authority",
+        "unstable__authority_boost",
+        "unstable__include_sensitive_results",
+        "internal__index",
+        # These are path parameters, not query parameters.
+        "source",
```
stacimc, line 359: What is the effect of this? I can still use these as query parameters. E.g.: `http://localhost:50280/v1/audio/source/jamendo/?tag=peaceful` works and filters as expected.

api/api/controllers/search_controller.py
```diff
@@ -428,36 +478,78 @@ def search(
...
+    include_sensitive_results = search_params.validated_data.get(
+        "include_sensitive_results", False
     )
+    if not include_sensitive_results:
+        search_query["must_not"].append({"term": {"mature": True}})
+    if excluded_providers := get_excluded_providers():
+        search_query["must_not"].append({"terms": {"provider": excluded_providers}})
+
+    # Rank the `tag` results by popularity and authority
+    if "tag" in search_params.data:
```
stacimc, line 534: Does this result in the popularity/authority boosting being applied if you have a tag query on a source or creator collection? E.g.: http://localhost:50280/v1/audio/source/jamendo/?tag=peaceful

If so, I think that's potentially confusing behavior since the creator/source endpoints are described as being sorted by ingestion order.

stacimc (commented):
Thanks @obulat! I totally understand the intent of the collection endpoints now, and the main endpoints are testing really well. I do still have a number of questions about how the API is supposed to work, but it all comes down to questions about the _additional filters_. Here’s a couple of examples:

* It looks like filtering by `tag` might cause the results to be ranked by popularity on a non-tag collection (/source/jamendo/?tag=vocal)?
* It’s a little confusing that you can filter a collection by ‘tag’, but a search by ‘tags’
* It looks like filters applied to collection endpoints are *all* exact matches. So `/source/jamendo/creator/John` and `source/jamendo/?creator=John` would both be exact matches on a creator named “John”, but /?source=jamendo&creator=John` would fuzzy match the creator. Is that what we want?

This could definitely be helped with documentation (which it looks like you added, but isn’t showing up for me locally — more on that below), but I think we still have to make a lot of decisions about how these parameters should work together, which is tricky because it will be really difficult to go back on any of those decisions once they’re live in the API. From my last read of the IP, my understanding is that the extra filters on these endpoints are really nice to have (I can definitely see the use cases!), but aren’t _necessary_ for the project. Is that correct? If so, the more I think about it my preference would be to just **remove all additional filters from these endpoints for now**. That still allows us to add them in the future, but keeps the API _much_ simpler and easier to test for now, and gets rid of all these edge cases. What do you think?

_______________________________

About the documentation: I see your updates but the documentation looks the same to me locally 😓 — I don’t know if I’m doing something wrong, totally possible.

Screenshot 2023-09-20 at 1 38 23 PM

Specifically the changes that I thought would be useful would be:

* Update the docs for the query parameters for the /search endpoint to explicitly mention that `creator` and `tags` are fuzzy searches
* Update the text of the creator/source/tag endpoints to also mention that they are exact matches (in that top-level docstring, since that would help make it clear how it differs from a regular search with a filter)
* Add a list of accepted query parameters to the new endpoints, and specify where there behavior differs (like being exact matches). If we get rid of extra filters, we don’t need to do this part 😄 

And then just a couple small things from testing:

* The creator collection for creators that are URLs still doesn’t work for me 😞 (http://localhost:50280/v1/images/source/flickr/creator/https://www.rosanetur.com/ 404s). 
* Filtering is case sensitive, it looks like — do we want that? That seems reasonable for `creator`/`source`, but what about `tag`?

stacimc:
> I wrote the code with the assumption that a single source does not have any creators with identical names. I think this is a fair assumption: for instance, you probably cannot register with a name that is already taken on Flickr.

+1 for this reasoning. Maybe we could add a comment or something in the catalog -- maybe in the new provider scripts, or the documentation for the media schemas? Just to let people know that this is an expectation when adding a new provider. But I think that's a reasonable assumption regardless!

api/api/serializers/image_serializers.py
```diff
@@ -59,13 +60,31 @@ class ImageSearchRequestSerializer(
...
-        return index
+
+class ImageSearchRequestSerializer(
+    ImageListRequestSerializer,
+    ImageSearchRequestSourceSerializer,
+    MediaSearchRequestSerializer,
+):
+    """Parse and validate search query string parameters."""
+
+    fields_names = [
```
sarayourfriend, line 71: I never noticed it before but the double plural here is a bit strange, `field_names`, like `ImageSearchRequestSourceSerializer` has would be clearer and consistency between all these serializers even better. DRF itself also uses `field_names`, and consistency with the library would be good as well. If we're adding many new instances of the awkward double plural, we might as well just fix the awkwardness, be consistent, and match the library, rather than perpetuate the awkward naming.

api/api/serializers/image_serializers.py
```diff
@@ -59,13 +60,31 @@ class ImageSearchRequestSerializer(
...
-        if index is None:
-            return None
-        if not index.startswith(IMAGE_TYPE):
-            raise serializers.ValidationError(f"Invalid index name `{value}`.")
-        return index
+
+class ImageSearchRequestSerializer(
+    ImageListRequestSerializer,
+    ImageSearchRequestSourceSerializer,
+    MediaSearchRequestSerializer,
```
sarayourfriend, line 67: I know this was discussed, and I did go and follow the inheritance chain here to confirm that `MediaSearchRequestSerializer` actually isn't in the inheritance chain, but I can't help but feel that there's something a bit wonky going on with the composition of these types. I did a bit of thinking and came up with a solution I think is a good potential alternative, which is shared below the horizontal rule at the end of this comment. I also wrote two other long suggestions that I think are _bad_ and don't help matters and are just confusing in different ways. I didn't want to delete them becuase I think they're _tempting_ alternatives, and I wanted to share my thought process to anticipate different rabbit holes we could work ourselves into and might be able to avoid :slightly_smiling_face: .

<details>
<summary>Two bad alternatives</summary>
One thing that could help because it would (a) untangle the path params from the query params (which Staci pointed out as an issue) and (b) reduce the number of layers of inheritance, is to not use the same serializer for the path and the query parameters. For example, rather than having the collections serializer inherit from the list serializer (with all the query params), rename the "list" serializer to e.g., "MediaSearchFiltersSerializer", and keep the path parameters in a separate, non-heirarchical `MediaCollectionPathParamsSerializer`. The bonus is that this new approach would require no additional inheritance for the collection, and the `SearchQuery` serializer (renamed from "MediaSearchRequestSerializer") would just need to add those collection params.

Although, now that I'm writing it out, I think there's an even better approach: `FilterParamsSerializer`, `CollectionParamsSerializer`, which are individually used as the query and path param serializers respectively for the new collections endpoints and then combined into a `SearchParamsSerializer` used for the query params of the existing search endpoint.

This eliminates one class from these inheritance lists at the media type level. We'd have the following:

```py
# media_serializers.py
class FilterParamsSerializer:
    """Parameters accepted by all media search endpoints."""
    ...

class CollectionParamsSerializer:
    """
    Parameters used to narrow media by collection.

    Collections are defined as sources, creators, and tags.
    """
    ...

class SearchParamsSerializer(CollectionParamsSerializer, FilterParamsSerializer):
    """Base class for media type search parameters."""
    ...

# image_serializers.py
class ImageFilterParamsSerializer(FilterParamsSerializer):
    """Image-specific parameters for filtering images."""
    ...

class ImageSearchRequestSerializer(ImageFilterParamsSerializer, SearchParamsSerializer):
    """All-image search parameters"""
    ...

# audio_serializers.py
"""
Same as image
"""
```

Endpoint serializer configuration would be as follows:

* `/{media_type}/`: `*SearchRequestSerializer` only
* `/{media_type}/{collection_type}/{collection}/`: `*FilterParamsSerializer` for query params and `*CollectionParamsSerializer` for the path (actually doesn't make much sense :confused:, `collection_type` and `collection` are the keys here, not `creator`, `source`, and `tag`).

</details>

---

**Here's what I think is a _good_ alterative.**

After writing all that, I actually disagree with myself and have yet another suggestion: don't use the serializer for the collection path parameters at all. We'd still need to split the collection params out from the filter params, but we wouldn't need a `SearchRequest` base class as it does nothing to help reduce the code at the media type side. Rather than passing the collection filters as part of the serializer payload (artificially putting them into the `data`, pulling them from the path params), they would be passed as a dataclass with two keys: `collection_type` and `collection`.

In that case, we'd end up with:

* `FilterParamsSerializer`, as above
* `CollectionParamsSerializer`, as above
* No `MediaSearchRequestSerializer` base class anymore
* `{MediaType}FilterParamsSerializer` inherits from `FilterParamsSerializer` and adds media-type specific filters.
* `{MediaType}SearchParamsSerializer` inherits from `CollectionParamsSerializer` and `{MediaType}FilterParamsSerializer`.

For the routes:

* `/{media_type}/` uses `{MediaType}SearchParamsSerializer` for the query parameters
* `/{media_type}/{collection_type}/{collection}/` uses `{MediaType}FilterParamsSerializer` for the query params and passes a dataclass to search entrypoint to configure the collection aspect of the query.

I think this also requires slightly changing the way the search strategy is organised. Rather than a single entrypoint that switches on a strategy parameter (and therefore needs to accept the same parameters otherwise for both strategies, or else have very confusing parameters), have two separate search functions that share certain implementation details, for example:

```py
def search("""The current parameters"""):
    ...

def search_collection(Params<Search> + [collection_type: CollectionType, collection: str]):
    ...
```

Those two functions would then compose whatever shared behaviour was extracted from the current `search` function, rather than `search` needing to compose for both strategies.

I think the issue right now is that we're trying to compose from the inside out, rather than bottom up. This causes us to end up with really confusing inheritance trees like this one on the current `ImageSearchRequestSerializer`.

api/api/serializers/media_serializers.py
sarayourfriend: This part specifically only removes them from the documentation. The fields need to not exist on the serializer for them to not appear in the serializer data. I've shared a proposal in another comment for a way to untangle this a bit and prevent this confusing cross-pollination of parameters. It prevents us from needing to modify the documentation as well.

api/api/serializers/image_serializers.py
```diff
@@ -133,14 +152,13 @@ class OembedSerializer(BaseModelSerializer):
     spec: https://oembed.com.
     """
 
-    version = serializers.ReadOnlyField(
+    version = serializers.SerializerMethodField(
```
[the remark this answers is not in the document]
obulat: No, thank you for flagging!

api/api/views/audio_views.py
sarayourfriend: I think the reason we might be missing the media-specific documentation in the docs is because [these schema extensions do not include a `desc` parameter](https://github.com//WordPress/openverse/blob/ae04e7ebf33b3107bbcb5b693f0ebde110590862/api/api/docs/image_docs.py#L149-L176), which the other endpoint docs do.

sarayourfriend (commented):
Looking good. I agree with the existing comments, and left one big suggestion for a way to untangle the serializer inheritance that is making this PR a little more complex than it necessarily needs to be, I think.

I'm not a requested reviewer and I'm going on AFK next week, so I'm not going to "request changes" officially on this PR. However, if circumstances were different, I would do so. I don't know that my proposed solution is necessarily the best correct approach, but the current implementation is overly complex (unless I'm missing a reason why the serializers to be combined and the search entrypoint a single one rather than one-per-strategy, like `search` and `related` already are) and would be difficult to undo, especially if we try to maintain backwards compatibility in any sense. I think the best thing to do is to ensure we absolutely, under no circumstances, introduce query parameters that are not documented or intended to work, but actually do have an effect on how search is performed. Not only is this undocumented behaviour confusing (and makes the code very confusing), it is also a big risk: making search more complex than it already is is something we should not do lightly. It's our main entrypoint, the most important API endpoint, and it needs to be understandable. As much as we can reduce conditional checks and branches for each mode of searching, whether the default search, related, or the new collection searches, we should do so. Right now we're adding a few that I don't think are necessary. I would request changes on this PR if circumstances were different because of the difficulty of undoing this and because of the significant complexity we're adding to something that needs to be easily understandable (and is already complex as is). I would urge us not to commit and deploy parameters that do things we don't intend them to do, which Staci has documented is happening in this PR, even aside from the complexity issue I'm also concerned with.

krysal:
I'll put this on draft while we all re-review it.

obulat:
> I think the best thing to do is to ensure we absolutely, under no circumstances, introduce query parameters that are not documented or intended to work, but actually do have an effect on how search is performed. Not only is this undocumented behaviour confusing (and makes the code very confusing), it is also a big risk: making search more complex than it already is is something we should not do lightly. It's our main entrypoint, the most important API endpoint, and it needs to be understandable.

Thank you for bringing this to my attention! I was really excited by the fact that we _could_ filter the collections that I haven't taken the maintainability into consideration. I decided to remove all the collection filters from this PR. Instead, the views will only accept the path parameters and `page` query parameter. If we need to add more query parameters in the future, we can do that in a different PR.

obulat:
## Updated PR Description

This PR adds the endpoints for additional search views for audio and image. The endpoints start with `/v1/images` or `/v1/audio`:
- **Tag** view: `/tag/tag_name`
- **Source** view: `/source/source_name`
- **Creator** view: `/source/source_name/creator/creator_name`.

These endpoints return paginated results, and only accept `page` and `page_size` query parameters. To enable this feature, I had to refactor the media serializers and the search controller.

### Search controller
The `search` method is replaced with `query_media` method that builds the relevant query (collection or search), sets up other search parameters, executes the search, tallies the results and builds the search context.

Several functions were extracted from this large method:
- `get_index`
- `build_collection_query` or `build_search_query`
- `execute_search`

The new `build_collection_query` method creates a query from a dictionary of the possible query predicates (`filter`, `must_not`, `should` and `must` dictionaries). All of these dictionaries are then combined with a `Bool` query. This makes the query much less nested than the query that we currently create in the `search` (created using the `elasticsearch_dsl` library).

### Serializers
This PR extracts a `PaginatedRequestSerializer` as a parent to the `MediaRequestSerializer` to allow the collection endpoints to accept `page`, `page_size` query parameters. For audio, there's also an `AudioCollectionRequestSerializer` that allows to set `peaks` parameter (whether to add the audio peaks array to the response). 

### Sorting of the results
The results are sorted by `created_on`. The tag endpoint also adds a boost for `unstable__authority` and  (for the authorized requests).

## Update to search docs
As requested in , this PR adds documentation to the similar-named search query parameters (`source`, `creator` and `tags`) to specify that those parameters match fuzzily (unlike the collection views), and accept 1 or more values (comma-separated).

## Testing Instructions

Run the app using `just up`.
Try various collection routes:
1. http://localhost:50280/v1/images/source/stocksnap (source collection, only the sources that are present in the sample data can be used to test locally)
1. http://localhost:50280/v1/images/source/flickr/creator/Manzabar/ (creator collection)
1. http://localhost:50280/v1/images/tag/cat (tag collection, images)
1. http://localhost:50280/v1/audio/tag/birds (tag collection, audio)
1. http://localhost:50280/v1/images/source/flickr/?page=2 (second page of results)

Try the routes that should return errors:
1. http://localhost:50280/v1/images/source/met (We don't have anything from `met` in the sample data, so the source validator throws an error
1. http://localhost:50280/v1/images/source/flickr,stocksnap (The path parameters are not split by a comma, so only a single parameter is allowed)

If there are no items corresponding to the requested values, the empty collection can be returned:
http://localhost:50280/v1/images/source/flickr/creator/Manzaba/ (the creator name is missing the last `r` letter)

Since the image and audio search had to be modified, too, check that they work as expected (the CI should check that, too).

api/api/controllers/search_controller.py
```diff
@@ -296,21 +295,23 @@ def _exclude_sensitive_by_param(s: Search, search_params):
...
 
     return index
 
 
-def search(
-    search_params: media_serializers.MediaSearchRequestSerializer,
+def query_media(
+    strategy: ListView,
+    search_params: media_serializers.MediaSearchRequestSerializer
+    | media_serializers.MediaListRequestSerializer,
```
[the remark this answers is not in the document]
obulat: These types were causing circular import errors. I fixed it using conditional imports, as is advised in the typing documentation.

api/api/views/media_views.py
```diff
@@ -99,8 +119,63 @@ def retrieve(self, request, *_, **__):
...
 
     def list(self, request, *_, **__):
-        params = self._get_request_serializer(request)
+        params = self._get_request_serializer(request, strategy="search")
+        return self.get_media_results(request, "search", params)
+
+    def collection(self, request, tag, source, creator, *_, **__):
+        if tag:
+            collection_params = {"tag": [tag]}
+        elif creator:
```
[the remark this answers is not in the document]
sarayourfriend: I think the `+` in the regex prevents an empty source in the URL:

```
>>> path = re.compile(r"source/(?P<source>[^/.]+)")
>>> path.match("source/") is None
True
>>> path.match("source/hello") is None
False
>>> path.match("source//") is None
True
```

Same goes for the other collection parameters. In the logic of this function, I think it's safe to just `else` without checking for `source`. These are less of _existence_ checks and more _inferences_ of the strategy based on the parameters.

That _might_ be clearer if this method also accepted a `collection_type` parameter of `creator | source | tag`, but I'm ambivalent. Alternatively, each action handler could construct the `collection_params` itself and pass it already initialised to `::collection`. That would leave collection completely (blissfully) ignorant of these details. If anywhere _should_ check that `tag`, etc aren't empty, it would be the action handler, I would think, not a sub-function of it.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -759,3 +772,59 @@ def test_post_process_results_recurses_as_needed(
...
+            [
+                {"term": {"source.keyword": "flickr"}},
+                {"term": {"creator.keyword": "nasa"}},
+            ],
+            [],
+            id="filter_by_creator",
+        ),
+    ],
+)
+@mock.patch("api.controllers.search_controller.Search", wraps=Search)
```
sarayourfriend, line 815: I would recommend _against_ mocking `Search`. I did it before in [`test_media_report` but replaced it with pook matchers recently](https://github.com/WordPress/openverse/commit/503ac82ad432df120497674a77fb8d72066c6c7f#diff-59d0ff474f161029e004a8e550ee72d2cb61875dd1650cd079d0079f3e64c852) when it turned out it was actually necessary to check the real calls to Elasticsearch. In particular, I think it was the client API using different parameters, but the ES request remaining the same, which revealed to me that we were more or less testing the implementation details of the method rather than the actual functionality to Elasticsearch.

Not a blocker or strong suggestion, just sharing an observation from some recent work I did with testing against ES.

api/test/unit/views/test_media_views.py
```diff
@@ -42,6 +44,36 @@ def test_retrieve_query_count(api_client, media_type_config):
...
+        pytest.param("tag/cat/", {"tag": "cat"}, id="tag"),
+        pytest.param("source/flickr/", {"source": "flickr"}, id="source"),
+        pytest.param(
+            "source/flickr/creator/cat/",
+            {"source": "flickr", "creator": "cat"},
+            id="source_creator",
+        ),
+    ],
+)
+def test_collection_parameters(path, expected_params, api_client):
```
sarayourfriend, line 60: This is a good test, I wonder if we could convert it into an integration test instead of using mocks, the same we do for the other search endpoints. Here's an example from a recent PR by @dhruvkb that added such a test: https://github.com/WordPress/openverse/pull/3013/files#diff-b8ef7387c8d809dde36c541047533b8c3f1e23ccaa1455ab98de1428f0518143R158-R180

sarayourfriend (commented):
I've just reviewed the code and will test this locally later today or early tomorrow, but the difference in implementation of the serializers in this new version is amazing. Fantastic work finding a simpler approach for this. It looks really good!

stacimc (commented):
It took me awhile to review the code and I ran out of time to test locally, but this is looking fantastic :) The documentation updates are great, and I feel much more comfortable with the simplification for now. I'll prioritize finishing up testing asap!

api/api/controllers/search_controller.py
```diff
@@ -428,36 +489,61 @@ def search(
...
+    :param s: the search client to build the query for.
+    :return: the search client with the query applied.
+    """
+    search_query = {"filter": [], "must": [], "should": [], "must_not": []}
+    # Apply the term filters. Each tuple pairs a filter's parameter name in the API
+    # with its corresponding field in Elasticsearch. "None" means that the
+    # names are identical.
+    filters = [
+        ("tag", "tags.name.keyword"),
+        ("source", "source.keyword"),
```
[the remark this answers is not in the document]
sarayourfriend: Yes.

Hopefully we will wait until after the ES changes are through to merge this PR, at which point we can make those adjustments?

sarayourfriend (commented):
This LGTM, but I suppose we'll need to re-review after the ES field changes are in. Provided all that needs to change are the field mappings, I'll approve it then (maybe we should draft until then?)

Regarding the tags being an exact match: I think that's a fine way to approach this for the MVP. I do also think that it would be an improvement to use a match search against `tags.name` to get stemming in the future. For now though, it's nice to be able to avoid that complexity.

obulat:
> It took me awhile to review the code and I ran out of time to test locally, but this is looking fantastic :) The documentation updates are great, and I feel much more comfortable with the simplification for now. I'll prioritize finishing up testing asap!

I'm sorry for not communicating this earlier, but I think it would be best if you test this PR after I un-draft it again (after the ES index PRs are merged and deployed), @stacimc .

stacimc:
>  I think it would be best if you test this PR after I un-draft it again

That works out perfectly :) +1

api/api/docs/base_docs.py
```diff
@@ -83,3 +98,123 @@ def get_operation_id(self) -> str:
...
+
+The results in the collection will be sorted by the order in which they
+were added to Openverse.
+    """
+    elif collection == "creator":
+        return f"""
+Get a collection of {media_type} by a specific creator from the specified source.
+
+This endpoint returns only the exact matches both on the creator and the source.
+Notice that a single creator's media items can be found on several sources, but
```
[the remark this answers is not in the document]
obulat: Not only because source =/= provider.
Several sources (and providers) can have works by the same provider. For example, if you search for Rembradt, you can see media from several sources, such as Wikimedia, Met, inaturalist, Finnish Heritage Agency. Not all of them refer to the same creator. Flickr and Inaturalist, for instance, can have users that use "Rembrandt" as their username. This URL allows you to see media by creators with "Rembrandt" in their name, with the sources I listed above excluded: https://api.openverse.engineering/v1/images/?creator=rembrandt&excluded_source=wikimedia,met,inaturalist,finnish_heritage_agency 
You can also see that the same creator can have a different name in different providers (it's "Rembrandt van Rijn" in SMK)

api/api/docs/base_docs.py
```diff
@@ -83,3 +98,123 @@ def get_operation_id(self) -> str:
...
+        return f"""
+Get a collection of {media_type} with a specific tag.
+
+This endpoint returns only the exact matches. To search within the
+tag values, or to match several tags, use the `search` endpoint
+with `tags` query parameter instead of `q` parameter.
+
+The returned results are ordered primarily based on their popularity
+and authority. However, note that the exact order may vary over time
+or across requests.
```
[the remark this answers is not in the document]
obulat: I was trying to add a statement to show that the order can change and is not constant. Since the popularity and authority boosts were removed from tags, I updated the docs to only include the order added to Openverse.

api/api/controllers/search_controller.py
```diff
@@ -300,8 +315,49 @@ def create_search_query(
...
+    :param search_params: the validated search parameters.
+    :return: the search client with the query applied.
+    """
+    search_query = {"filter": [], "must": [], "should": [], "must_not": []}
+    # Apply the term filters. Each tuple pairs a filter's parameter name in the API
+    # with its corresponding field in Elasticsearch. "None" means that the
+    # names are identical.
+    filters = [
+        # Collection filters allow a single value.
+        ("tag", "tags.name"),
```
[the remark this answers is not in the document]
sarayourfriend: Weird, while digging through our data to find examples where white-space sensitivity could matter for tags, I found some tags that are _not_ normalised to lower case:

https://api.openverse.engineering/v1/audio/03ea0149-d9c5-47f3-97db-ed7f715b46af/

That result has "Low-Quality" rather than "low-quality" which other audio results have...

api/api/docs/base_docs.py
```diff
@@ -77,3 +92,122 @@ def get_operation_id(self) -> str:
...
+def get_collection_description(media_type, collection):
+    if collection == "tag":
+        return f"""
+Get a collection of {media_type} with a specific tag.
+
+This endpoint returns only the exact matches, case-insensitive matches for the
+specified tag. For example, 'birds' and 'birding' are not matches for 'bird'.
+To search within the tag values, or to match several tags, use the `search` endpoint
+with `tags` query parameter instead of `q` parameter. In this case, the matches will
+ not be exact, so 'cat' would match both 'cat' and 'cats'.
```
sarayourfriend, line 147: Just a note that "case-insensitive" here is incorrect unless we normalise tag inputs (as mentioned in another one of my comments).

Regarding the clarification of exact/fuzzy matching, the last two sentences are technically documentation for the search endpoint, not this one. It's good information to have on the search endpoint, but is displaced here. I'm having a hard time coming up with a good alternative though, mostly because a full description of how matching
works for this endpoint is mutli-faceted:

- It matches entire tags, so multi-word tags must be multi-word inputs. An input of `dog` will not match works tagged `dog walking`. Nor would `runner` match `marathon runner`.
- It is whitespace sensitive, so `dog walking` is distinct from `dog walking `, though not sure how much that matters. I can't really think of a reasonable example for this that wouldn't just be obtuse. Is it worth documenting though? It _could_ be important, maybe for hyphenated terms, for example? Place names sometimes have hyphens in them and are sometimes left out for orthographic or other reasons. In production data we have both "Saint Pierre des Champs" and "Saint-Pierre-des-Champs" in tags.
- Diacritical marks are also important. "jalapeño" and "jalapeno" match different tags (should we "deburr" tags?)
- It doesn't do text analysis, so plural and conjugated forms of the same word are distinct ("dog"/"dogs")

The last one _feels_ most important to me, but the diacritical distinction could also be pretty important. Here's the only text I could come up with as an alternative, but is it too long? When I try to write a concise version, I end up basically re-writing what you've already got here. I'm struggling to decide what level of detail is _necessary_ for this documentation.

```suggestion
This endpoint matches a single tag, exactly and entirely.

Differences that will cause tags to not match are:
- upper and lower case letters
- diacritical marks
- hyphenation
- spacing
- multi-word tags where the query is only one of the words in the tag
- multi-word tags where the words are in a different order

Examples of tags that **do not** match:
- "Low-Quality" and "low-quality"
- "jalapeño" and "jalapeno"
- "Saint Pierre des Champs" and "Saint-Pierre-des-Champs"
- "dog walking" and "dog  walking" (where the latter has two spaces between the last two words, as in a typographical error)
- "runner" and "marathon runner"
- "exclaiming loudly" and "loudly exclaiming"

For non-exact or multi-tag matching, using the `search` endpoint's `tags` query parameter.
```

This makes me wonder whether there's actually a larger improvement to make here, where we have a separate piece of documentation describing what we mean by "exact" and "non-exact" matching and where, to be linked from here and other relevant endpoints. It would be a less technical version of https://docs.openverse.org/api/reference/search_algorithm.html#text-analysis-tokenization

api/api/controllers/search_controller.py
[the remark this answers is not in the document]
sarayourfriend: One more thing, which definitely needs a change: `tags.name` is a text field, which the ES documentation says _not_ to use a terms query on: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html#avoid-term-query-text-fields

We should change this to `tags.name.keyword` if we want to use a terms query.

Alternatively, if we want to avoid all the exact-matching stuff, we could use `tags.name`, but then just need to change the `term` query to a `match` query below.

```suggestion
        ("tag", "tags.name.keyword"),
```

api/api/controllers/search_controller.py
```diff
@@ -300,8 +315,49 @@ def create_search_query(
...
+        ("source", None),
+        ("creator", None),
+    ]
+    for serializer_field, es_field in filters:
+        if serializer_field in collection_params:
+            if not (argument := collection_params.get(serializer_field)):
+                continue
+            parameter = es_field or serializer_field
+            search_query["filter"].append({"term": {parameter: argument}})
```
sarayourfriend, line 343: Alternative that uses `match` for the tags:

```suggestion
    filters = [
        # Collection filters allow a single value.
        ("tag", "tags.name", "match"),
        ("source", None, "term"),
        ("creator", None, "term"),
    ]
    for serializer_field, es_field, query_type in filters:
        if serializer_field not in collection_params:
            continue
        if not (argument := collection_params.get(serializer_field)):
            continue

        parameter = es_field or serializer_field

        match query_type:
            case "match":
                query = {"match": {parameter: {"query": argument}}}
            case "term":
                query = {"term": {parameter: argument}}

        search_query["filter"].append(query)
```

If we take this approach, then we can cut down on the documentation complexity too, I guess, by just saying it uses regular fuzzy matching. But I know this is something we've talked about a lot, and I'm not totally remembering why we decided to use a terms query for tags, so forgive me if I'm just re-treading ground we've already covered.

sarayourfriend (changes requested):
Looks good to me. I haven't had a chance to test this locally yet, and will do so right after leaving this review. I wanted to make sure I left these comments now though, just in case I get interrupted later today while testing locally.

The only requested changes are related to if we use a terms query:

- Use `tags.name.keyword` to avoid using terms on a text field
- ~Cast tag inputs to lowercase~ (https://github.com/WordPress/openverse/pull/2853#discussion_r1390562124)

However, with the last one, I'm noticing we actually have _lots_ of works where the tags are not lowercased, which I believe might be because tag cleanup isn't actually running? I opened an issue to look into here: #3342. It would be good to understand what's going on there if we're going to use a terms query so that we know what to expect and what to document.

api/api/controllers/search_controller.py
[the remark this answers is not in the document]
sarayourfriend: BTW, noticing when testing locally that case insensitivity is working. I didn't notice the call to lower the tag in the media view's implementation of the tag endpoint. Please disregard my comments about case-sensitivity not working. However, if we switch to the `match` query, then we can remove the code calling tolower on the tags.

But! Because we _do_ have tags that are not lower cased, because of #3342, I think that bug is much more significant, because if we roll this feature out as-is, then there are works tagged such that they would never appear in the tagged collection even though they have that tag. That's only relevant if we stick with a terms query for tags, though.

sarayourfriend:
This is testing well for me locally :rocket:. Just a couple of small details with the tags route to iron out, but the rest of this is solid :rock:

api/test/unit/views/test_media_views.py
obulat: The integration tests were added.

api/api/controllers/search_controller.py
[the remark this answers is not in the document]
obulat: I think we should stick to the terms query for tags here. I've added some comments on #3342 on proposed solution. I think we can finish it by the time the additional search views are shipped.

api/api/docs/base_docs.py
obulat: > The last one _feels_ most important to me, but the diacritical distinction could also be pretty important. Here's the only text I could come up with as an alternative, but is it too long? When I try to write a concise version, I end up basically re-writing what you've already got here. I'm struggling to decide what level of detail is _necessary_ for this documentation.

I think that for tags, it is more important to have the exact match. Even in case of plural vs singular: the users might want to filter out only the images where several items are present. For instance, `cats` implies that the image has more than one cat, which might be important for the user.
Since we already have a way to search (not match exactly) in tags using `tags=` parameter, I think it's important to add the exact match for the tags view.

As for the docs, I think extracting the description of what "exact" and "non-exact" means and linking to it from the shorter endpoint descriptions would be a good idea. @stacimc, what do you think?

api/api/controllers/search_controller.py
```diff
@@ -37,6 +44,14 @@
 DEFAULT_BOOST = 10000
 DEFAULT_SEARCH_FIELDS = ["title", "description", "tags.name"]
 
+FILTER_TYPE = Literal["filter", "exclude"]
```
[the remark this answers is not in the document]
obulat: I was going to accept the suggested change, but then realized that this is not used anyways, so I removed it.

api/api/controllers/search_controller.py
```diff
@@ -37,6 +44,14 @@
...
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
obulat: This is a fix for circular imports: if we import the serializers here directly, the app will fail with the `circular import` error. Here's more info on how `TYPE_CHECKING` is used for resolving circular import errors: https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles
Should I add a comment with this link?

api/api/controllers/search_controller.py
```diff
@@ -322,46 +385,54 @@ def search(
...
     else:
-        index = origin_index
+        query = build_search_query(search_params)
 
-    s = Search(index=index)
+    s = Search(index=index).query(query)
 
-    search_query = create_search_query(search_params)
-    s = s.query(search_query)
+    if strategy == "search":
```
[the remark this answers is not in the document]
obulat: I've extracted `build_query` to make this part a bit clearer.

api/api/controllers/search_controller.py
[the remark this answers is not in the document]
sarayourfriend: Oh cool! I only found the reference in the Python docs. Just a comment saying "avoid circular import" is fine from me.

sarayourfriend (approved):
LGTM!

stacimc (approved):
Tests well for me! I would prefer that one docstring update, but nothing blocking. Kudos @obulat, this is great 🚀

Issue #2839, opened 2023-08-17
New popover and popover item


fcoveram:
## Description

As part of the design changes agreed upon in #460, we need to update the styles of **Popover** and **Popover item**.

### Popover

(description from #460)

The popover follows the [Modal](https://www.figma.com/file/e4tLacmlPuZV47l7901FEs/WordPress-Design-Library?type=design&node-id=28524-24040&mode=design&t=ZV4igMACfeHQZBv2-4) implemented in Gutenberg.

* Popover is divided into three sections to make it reusable: Header, description, and content.
* Header area has the popover title and close action.
* Close button follows [Icon](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=4658-12470&mode=design&t=JzKg0h3PdFx8XQ1v-4) component in `size: small`, `style: transparent`, `color: gray`.
* The container has a border radius of `4px` in all breakpoints.
* In `XS` and `SM`, the container has a padding of `16px` and a margin of `32px`.
* In MD and beyond, the container has a padding of `24px` and a `max-width` (TBD).

You can find the popover structure in the [Design Library](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=5327-30390&mode=design&t=UejtvitW7rxf1Smu-4), and the Metasearch case in its respective [design file](https://www.figma.com/file/AsD7zOnUQxuLaUBiptTYAn/Metasearch-improvement?type=design&node-id=22-13656&mode=design&t=jYPJaXJ3UNeMQvis-0).

![Popover structure in small and big size](https://github.com/WordPress/openverse/assets/895819/6a9fa169-2269-4804-b9c6-b14550a8dfde)

### Popover item

In line with the improvements made in #415, this component inherits the fixed height and has three variants for different interaction cases. In #460 designs, the variant used is `Label`.

![Popover item in three variants and three interaction states](https://github.com/WordPress/openverse/assets/895819/fae83829-24d1-46cc-aad0-57306a9c094c)

You can find the Popover item component in the [Design Library](https://www.figma.com/file/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?type=design&node-id=5327-30331&mode=design&t=UejtvitW7rxf1Smu-4). You will also see a Popover header component in the same section that is not part of #460 changes and will be requested in a different ticket.

## Additional context

[...]

PR #2860, opened 2023-08-23
Update the external search modal and items

### LINKED ISSUE #460: Update external sources modal styles

obulat:
## Fixes

Fixes #460 by @fcoveram

## Description

This PR updates the external search items modal according to the updated designs.
- external sources use modal on all breakpoints (instead of selecting modal for small breakpoints and popover for larger ones). This allows us to use `VModal` to simplify the component, instead of handling the modal behavior manually.
- "centered" modal is full width, except for 24px of horizontal margins, on `xs`, and has a max-width set to 400px for breakpoints `sm` and above
- the sizes, paddings and text styles of the external sources buttons were updated

The `VModalContent` had a bug where the `hide` prop in the top-bar slot was not passed to the `VModal`, so it was impossible to close the modal using the close button. This PR fixes that, too.

## Designs
**Figma file:** [Metasearch improvement](https://www.figma.com/file/AsD7zOnUQxuLaUBiptTYAn/Metasearch-improvement?type=design&node-id=22-13656&mode=dev).

## Testing Instructions

Search for images or audio, scroll to the bottom and click on the "External sources" button. The modal should match the designs, should be closable, and should update sizes if you change the screen width. You should be able to open the external sources links.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/e2e/external-sources.spec.ts
```diff
@@ -1,38 +1,55 @@
...
 
-  expectEventPayloadToMatch(viewEvent, {
-    searchType: "image",
-    query: "cat",
-    resultPage: 1,
-  })
   expectEventPayloadToMatch(selectEvent, {
     name: "Centre For Ageing Better",
-    mediaType: "image",
+    mediaType: "audio",
```
obulat, line 52:
```suggestion
    mediaType: "image",
```

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
```diff
@@ -36,72 +36,71 @@
...
-        :variant="isMd ? undefined : 'centered'"
-      >
-        <VExternalSourceList
-          class="flex flex-col"
-          :search-term="searchTerm"
-          @close="closeDialog"
-        />
-      </Component>
-    </template>
+    <VModal
```
fcoveram, line 39: Modal radius should be `4px`.

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
```diff
@@ -36,72 +36,71 @@
...
+          <VCloseButton
+            size="small"
+            icon-size="medium"
+            variant="filled-white"
+            :label="$t('modal.close')"
+            @close="close"
+          />
+        </header>
+      </template>
+      <VExternalSourceList class="flex flex-col" :search-term="searchTerm" />
```
fcoveram, line 75: Icon size of external link item needs to be like [VIconButton](https://docs.openverse.org/storybook/?path=/story/components-viconbutton--default-story&args=size:small) in `size: small`, but not interactable. This is for consistenty with the new popover item requested in #2839.

fcoveram (changes requested):
It looks nice ⭐ I noticed a few details and tried to be as accurate as possible. The mismatch is caused by how Figma sets spacing per frame instead of per section, as in the code.

frontend/tailwind.config.ts
```diff
@@ -226,7 +226,6 @@ export default {
       },
       borderRadius: {
         inherit: "inherit",
-        md: "0.5rem",
```
obulat, line 229: I didn't realize that `rounded-lg` is actually `border-radius: 0.5rem` when I added this new value, `md`. Since `md` is redundant now, I removed it and replaced it in the components with `rounded-lg`

frontend/src/components/VExternalSearch/VExternalSearchForm.vue
fcoveram: Sorry for not explaining it correctly.

The popover item in #2839 has an optional icon on the right side that in some cases you can interact with it. Here is an example.

recent searches item interaction

Although the above has not been implemented yet, the new style was thought to allow this internal action.

When the action is not needed, like in this case where `↗` only indicates a link type, the icon area is the same (`32px` for the container, `24px` as icon size) to keep consistency across all popover items. Here is a screenshot of popover item in wireframe.

Outline version of Popover item component

The smallest box area is the default icon size of `24px`, and the container is like `VIconButton` in `32px (size: small)`.

obulat: At first when I read this note, I thought maybe we should create a separate component in code. However, because they have different functions (the external icon is a "decoration" that has no functionality and the remove "cross" icon is actually an icon button that should have some click handlers) - I think it's best to leave the "external icon" here as VIcon. I set the icon size to 5, which makes the icon itself around 16px high - just like in the mockups. I think the size of the icon now matches the designs. Could you please confirm, @fcoveram ?

fcoveram: > because they have different functions (the external icon is a "decoration" that has no functionality and the remove "cross" icon is actually an icon button that should have some click handlers) - I think it's best to leave the "external icon" here as VIcon

That makes sense to me. But VIcon needs to be `size: 6 (24px)` and placed inside a `32px` box to look balanced. Without the 32px container, it looks broken.

Here is a design test where the color-filled area is the VIcon in size 6, and the border-only area is the 32px container.

Diagram of popover item. Left, icon with a container. Right, icon without container

I reviewed the PR's last changes and still see the icon in 20px. Here is a gif of what I see.

popover item - chrome inspector

fcoveram: VIcon in size 8 is 32px element that scales up the icon, whereas the above refers to a container area to balance spaces between popover items while keeping the icon in 24px.

fcoveram (approved):
It looks excellent ✨ 🚀  Thanks for the wonderful work.

AetherUnbound (approved):
Confirmed that everything looks correct visually when run locally!

PR #2121, opened 2023-05-17
Allow feature flags to use `sessionStorage` and ignore `ff_` query params

dhruvkb:
## Description

This PR is a PoC for #2118 that demonstrates feature flags that can be backed by session-storage. It also allows feature flags to ignore `ff_` query params when reading from them can open the possibility for malicious activity.

### Session storage

0. Check out the PR and run the dev server.
1. Visit `/preferences`.
2. Flip the `fetch_sensitive` flag to on.
3. Navigate around the site and come back to `/preferences`.
4. See that the flag is still on.
5. Close the tab and open a new one.
6. Visit `/preferences`.
7. See that the flag is off now.

### No `ff_` query

0. Check out the PR and run the dev server.
1. Visit `/preferences`. Ensure `sensitive_content` and `fetch_sensitive` are both off.
2. Visit `/preferences?ff_sensitive_content=on`. See that the switch is flipped on.
3. Visit `/preferences?ff_fetch_sensitive=on`. See that the switch is still off.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
[...]

PR #2880, opened 2023-08-24
Enabling fetching sensitive results when the UI switch is toggled

### LINKED ISSUE #2879: Update the store `fetch_sensitive` flag when it's toggled in the UI
## Problem

Currently, when you toggle "Sensitive results" in the filters sidebar, only the feature flag store is updated, so you cannot fetch sensitive results using the UI toggle.

## Description

 The search store's filter state should be updated when the UI "Sensitive results" toggle changes state.

obulat:
## Fixes

Fixes #2879 by @obulat

## Description

This PR enables the users to fetch the sensitive results by toggling the UI switch in the filters sidebar/modal.

To do so, a new `setFilter` method is added to the search store. When the switch is toggled, `setFilter` is called to update the value in the store to match the UI value. This enables users to fetch `sensitive` results.

Note that as per the Implementation plan requirements, the URL on the frontend does not have the parameter for sensitivity set, but the API requests still set the `unstable__include_sensitive_results` parameter set to true if the toggle is On.

Not directly related to this issue, but this PR moves the watcher that triggers media fetching on changes to the filters from `VSearchGridFilter` to `search.vue` (which already handles fetching on changes to the query parameters)

## Testing Instructions

I've updated the PR in the last commit to use the `unstable__sensitivity` values as the Media.sensitivity values, so now sensitive toggle works as expected, and you can actually see the results blurring based on the API filtered index values.

Enable the `sensitive_results` feature flag on the `/preferences` page.
Go back to the homepage and search for "cat". Notice that there are no query parameters in the URL besides `q=cat`.
Go to the filters sidebar (or the filters tab in the content switcher modal on a mobile screen), and turn on the "Sensitive results" switch. Notice that the URL ~~is changed to http://localhost:8443/search/?q=cat&unstable__include_sensitive_results=true~~ **doesn't change**, and a new set of results is loaded.

If you reload the same page (using CMD+R or the browser's Reload button), it should not show sensitive results (blurred, if you didn't unblur them using the switch) because we don't have a cookie for `fetch_sensitive` (only for `sensitive content`). Should we actually add this cookie?

If you try searching for something that is text-sensitive, you should see mostly blurred results.
When you turn off the "Sensitive results" switch, you should see no blurred media items in the results. 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (approved):
Thanks for integrating the two work streams!

sarayourfriend:
@obulat Can you add testing instructions?

sarayourfriend:
>  Go to the filters sidebar (or the filters tab in the content switcher modal on a mobile screen), and turn on the "Sensitive results" switch. Notice that the URL is changed to http://localhost:8443/search/?q=cat&unstable__include_sensitive_results=true, and a new set of results is loaded. If you try searching for something that is text-sensitive, you should see mostly blurred results.

@obulat Can you confirm whether the inclusion of the query parameter represents the final feature behaviour? The implementation plan explicitly calls for there _not_ to be a query parameter present on the frontend for the sensitive results' inclusion (to prevent the potential for abusively sending a link to sensitive results without the recipient realising what they're clicking).

obulat:
> > 
> 
> @obulat Can you confirm whether the inclusion of the query parameter represents the final feature behaviour? The implementation plan explicitly calls for there _not_ to be a query parameter present on the frontend for the sensitive results' inclusion (to prevent the potential for abusively sending a link to sensitive results without the recipient realising what they're clicking).

I didn't realize that this is the required behavior. I'll update the PR to remove the query parameter from the URL

sarayourfriend:
One more clarification, sorry! I don't see a separate issue for this, so I wanted to make sure that the setting persistence requirements from the project proposal (https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#settings-persistence) and implementation plan (https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230506-implementation_plan_frontend_blurring_sensitive.html#sensitive-content) are implemented here. If they're meant to be covered in a separate issue, then that issue needs to be created.

My understanding is that the change that would need to happen for this based on the current state of the PR is to set the toggled setting into the feature flags store rather than in the search store filters and then derive the query param as sent to the API based on the feature flag rather than from the search store. @dhruvkb can you confirm that is correct and/or what the intention is as far as implementing that part of the feature?

obulat:
> My understanding is that the change that would need to happen for this based on the current state of the PR is to set the toggled setting into the feature flags store rather than in the search store filters and then derive the query param as sent to the API based on the feature flag rather than from the search store. @dhruvkb can you confirm that is correct and/or what the intention is as far as implementing that part of the feature?

Thank you for this question, @sarayourfriend! I feel the `fetch_sensitive` flag should be moved to the search store because it's not a feature flag per se. It's not a feature that can be turned on and off for development purposes, but rather a setting a user has for search. We could, of course, check the feature flag store before sending a search request, but it would be unusual for the app: all settings currently are either in the `search` or `media` stores.

sarayourfriend:
IIRC the reason @dhruvkb wrote to use a feature flag in the implementation plan was to simplify the cookie-based persistence of the setting. That general requirement is described more deeply [in the project proposal](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#settings-persistence).

If it is easier to store the setting elsewhere, that's fine and good, of course, but it doesn't change the need to persist it as a cookie to match the behaviour laid out in the project proposal and implementation plan; admittedly, the implementation plan only makes oblique rather than direct references to this, I think because it's just deferring to the project proposal. In any case, it's a requirement that either needs to be handled here or in a separate issue, which would need to be created.

obulat:
> IIRC the reason @dhruvkb wrote to use a feature flag in the implementation plan was to simplify the cookie-based persistence of the setting. That general requirement is described more deeply [in the project proposal](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230309-project_proposal_detecting_sensitive_textual_content.html#settings-persistence).

Thank you for the context. I was thrown off by the fact that we do **not** have a cookie for `fetch_sensitive`, only the `sensitive_content` one (which is the feature flag that enables the feature as a whole). So, currently, the saving of this setting in a cookie is not implemented.

We do have cookie-based settings in the UI store, so we could add cookie-based settings to the `search` store as well. @dhruvkb, do you think it would make sense to move `fetch_sensitive` to the search store and add the `fetch_sensitive` cookie that would be used to initialize this setting in the search store, if the `sensitive_content` feature flag is enabled?

dhruvkb:
This project has 3 feature flags associated with it.

- `sensitive_content`
  
  This feature flag is to enable/disable the entire blurring project, similar to regular feature flags that we use to develop features on `main`. It controls the visibility of the sidebar toggles and the content safety wall etc. and is stored to the cookie for persistence.

- `fetch_sensitive`

  This is a feature flag that, when set to switchable, acts as a user preference. It is linked to the sidebar toggle "Sensitive results" (the first one). The reason it is set as a feature flag and not in the store is to persist it across a session in the session store (not cookie, because we want it cleared when the session ends!)

- `fake_sensitive`

obulat:
I would like to first confirm the **location** where we should store the `fetch_sensitive` setting.

I found the link to our exchange with @sarayourfriend on where the `fetch_sensitive` value should be stored in the Project Proposal: https://github.com/WordPress/openverse/pull/873#discussion_r1136313457 @sarayourfriend specifically stated that the setting should be saved in a cookie, not in `window.SessionStorage`.

The Implementation plan, on the other hand, describes what needs to be done to save the `fetch_sensitive` setting in the `window.SessionStorage`[^1]. These changes have been implemented.

Even if we store the setting in the `window.SessionStorage`, I believe it would be clearer if the syncing with the storage is done in the search store instead of the feature flags store. This would convey that this feature is not a temporary, beta version feature that is about to be turned on, but is actually a working feature in the frontend. There is a mechanism for initializing the values from cookies in the UI store. A similar thing can be used for the `fetch_sensitive` setting in the search store.

[^1]: [Feature flag system enhancements](https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230506-implementation_plan_frontend_blurring_sensitive.html#step-2-feature-flag-system-enhancements)

dhruvkb:
The reason I used the feature flag store to hold `fetch_sensitive` was that I had the plan to extend the feature flags store to all kinds of user preferences, with `fetch_senstive` being one of them. Currently, with only one such preference that too only scoped to a single session, it does not make much sense for why this particular toggle is being stored away from everything else. If you think it makes more sense to move it, I do not mind.

Secondly, the reason for using `sessionStorage` over cookie was that we can make use of the [browser's guarantees](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to handle the copying, management and clearing of the session storage. It differs slightly from cookies in one respect (see below).

Screenshot 2023-08-29 at 3 27 51 PM

The scoping of the feature flag to `sessionStorage` and its use as user preferences was documented in the implementation plan as well as PR #2121.

obulat:
> The reason I used the feature flag store to hold fetch_sensitive was that I had the plan to extend the feature flags store to all kinds of user preferences, with fetch_senstive being one of them. Currently, with only one such preference that too only scoped to a single session, it does not make much sense for why this particular toggle is being stored away from everything else. If you think it makes more sense to move it, I do not mind.

We could create a new `user-preferences` store to hold these preferences and re-use the functionality from the feature flag store for storing the settings in cookies/session/flag. I think we can defer this store creation until we create at least one more such user preference. For now, I changed this PR to only use the `featureFlagStore` to keep the `fetch_sensitive` preference. 

> Secondly, the reason for using sessionStorage over cookie was that we can make use of the [browser's guarantees](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) to handle the copying, management and clearing of the session storage. It differs slightly from cookies in one respect (see below).

Thank you for explaining the reasons, it makes sense. I think we should update the Project proposal to reflect this and to avoid such ambiguities in the future.

@sarayourfriend, to summarize, I removed the `fetch_sensitive` settings from the search store, and used the featureFlagStore's value to determine whether the request should have the `unstable__include_sensitive_results` parameter or not.
After writing this comment I realized that we probably should completely remove the `sensitive`/`mature` filter from the search store now that it's not used anyways. I'll update the PR.

sarayourfriend:
> I think we should update the Project proposal to reflect this and to avoid such ambiguities in the future.

Definitely. The project proposal shouldn't have made a specific technical recommendation and instead just said something like "stored for the 'session'" or something. Apologies for the confusion between the two being different. I think we should also consider implementation plans to be more authoritative with respect to implementation details than the project proposal, sort of like they supersede the project proposal in those respects.

In any case, thanks you and Dhruv for sorting out what was necessary to do in this PR and good work. I'll review this later today!

sarayourfriend (approved):
Looks really good to me. I found one bug in testing this locally. I'll leave it up to you whether you think it should be fixed in a fast-follow issue or if it would be simpler to fix it in this PR. Here's a gif demonstrating the issue, I describe the steps I followed after the gif. **Note: This requires enabling sensitive content fetching and is tested with a query that (apparently) does return sensitive textual content results in the first page. However, both of those results strike me as potential false-positives derived from descriptions as there is no sensitive text in the text available on the result on the webpage. There is one image and one audio result. The image is of a Galah, a type of bird, and the audio result is a recording of bird sounds. In any case, proceed with caution if you are worried about interacting with sensitive content. To clarify, I haven't unblurred any of the results marked sensitive in the gif below.**

Peek 2023-08-31 14-40

The issue demonstrated in this gif is significant layout shift between the initial SSR-derived render and the page after it is responsive in the browser. After the refresh button is clicked, notice that the initial results show _without anything blurred_, flashes to the skeleton all results mode, then finally lands on the results including the two blurred results. I believe what is happening here is that the "fetch sensitive results" flag isn't making its way to the server, resulting in the client re-requesting the search after hydration.

Steps to reproduce:

1. Turn the feature on via the /preferences page according to the test instructions. Do not turn on "fake sensitive results", you need real sensitive results for this to work.
2. Go back to home and search "galah". **This search returns results designated with sensitive textual content. Reproduction does not require unblurring them, but I would also consider them false positives. Use care here if you are concerned.** You should not see the same type of flash, just the regular (unfortunate) flash that comes from audio and image results coming back at different times and needing to be interlaced client-side.
3. Now refresh the page. You should see the flash described above and demonstrated in the gif.

Otherwise, this looks to work really well for me.

However, I did notice another bug, but it's present on `main` (tested on staging) so I'll open an issue for it and link in this comment afterwards.

I've approved because I don't want to block if you decide to address the flashing layout issue in a separate issue.

dhruvkb (approved):
LGTM!

Regarding the flash seen by @sarayourfriend, this would be because of `sessionStorage`. When the page is loaded from the server, it is with sensitive content turned off. After loading, the page populates the preferences from `sessionStorage` and flips the sensitive content toggle (if you look carefully, the sensitive content toggle switches from off to on after the page is refreshed). This causes the page to go back into loading mode and get the blurred results.

So we see
- flash of safe results
- skeleton loader
- mix of safe and blurred unsafe results

Switching to a session cookie would be the only way to access this information on the server side. Approving to unblock this as the migration to session cookie is a big change that should be its own PR.

PR #2909, opened 2023-08-28
Fix Elasticsearch memory hogging the right way

sarayourfriend:
## Fixes

Related to https://github.com/WordPress/openverse/commit/6cdfb59acf8bfe3aa3dc5528e8846e6ae2d62a4f

## Description

I noticed that Elasticsearch was happily using 60% of my system RAM. At one point today this actually caused an OOM issue in my kernel during a `just recreate` run, despite having 32 GB RAM! The overrides file does _not_ work for me (ES just crashes instead of staying within the limit). However, I remembered from the work I recently did to fix our production ES 8 cluster's heap configuration, that ES will respect heap limits passed as JVM args. Therefore, we can get rid of the overrides file (which we didn't really like any way) and use the heap limits. I've set them to half a gig, which should be _just fine_ for local testing.

## Testing Instructions

Checkout this branch and run `just down && just up` and confirm that Elasticsearch does not use 60% of your system memory :grin: 

Run `just down -v && just up && just init` and confirm that ingestion still works locally.

Note: This might slow down ingestion locally by a _very small amount of time_. When I run it locally I don't see an issue but for other computer configurations it may behave differently. If that's the case, we can increase it to 1 GB or all the way to the 2 GB that the previous overrides would have forced ES to use (remember it doesn't use the full amount of system memory available to it by default!).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker-compose.overrides.yml
+++ b/docker-compose.overrides.yml
@@ -1,4 +1,0 @@
-services:
-  es:
-    # Memory limit for ES, as it tends to be a memory hoarder
-    mem_limit: 4294967296
--- a/docker/es/env.docker
+++ b/docker/es/env.docker
@@ -6,3 +6,9 @@ discovery.type="single-node"
 
 http.cors.enabled="true"
 http.cors.allow-origin="/.*/"
+
+# Elasticsearch will reserve 50-60% of available system memory
+# if left to its own devices. Setting this to half a GB is sufficient
+# for local testing and prevent ES from hogging such a significant
+# amount of system memory.
+ES_JAVA_OPTS="-Xms512m -Xmx512m"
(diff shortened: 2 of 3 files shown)


dhruvkb (approved):

krysal (approved):
Tested with `api/recreate` as it does the same in one command for the API only 😄 I confirm it uses fewer memory resources.

Nice to remove a file here. If it's slowing down ingestion even by a bit then it's better to give it a full GB, a reasonable middle term I think.

PR #2973, opened 2023-09-04
Proof of Concept: Silo ESLint configuration into plugin package

sarayourfriend:
## Description

I've mentioned the possibility of this before and because I'm blocked on local environment issues with Python stuff, I thought I would give this a shot. Especially after working through the issues in #2972 (this PR depends on that one), this particular idea is a bit easier.

The concept here is that all our ESLint related dependencies and configuration get moved into the `@openverse/eslint-plugin` package. This means we only have one place to update ESLint and its related dependencies (plugins, parsers, etc). It makes the configuration at the root of the project very concise as it now just references the plugin and enables one of the built-in configs. I've also gone ahead and split the configuration out in a couple different ways, and there are further improvements available here, especially if we move `overrides` around. This allows us to split the ESLint configuration into easier to understand and navigate files, without creating multiple `eslintrc.js` files in our repository root.

I realised something weird was going on in our ESLint configuration. We had the prettier plugin installed but completely disabled (`prettier/prettier` was `off`). I've removed the plugin itself and just added the `eslint-prettier-config` and add the `prettier` ruleset to `extends` so that the appropriate built-in ESLint rules are disabled. This accomplishes the same end result without the weird situation of enabling the formatting plugin and then disabling it to allow prettier to do its formatting and prevent ESLint from fighting it.

## Testing Instructions

Checkout the branch and run `pnpm install`. Then try out `pnpm run eslint`, confirm it works. Make changes that should cause eslint to fail in various areas (like adding semi-colons, moving imports around, etc) and check that `pnpm run eslint` raises the appropriate errors. Try things in `json`, `vue`, and `ts` files to confirm that all the various plugins are working. A good tip is to remove `eslint-disable` comments and see if those are still working (i.e., if removing them and running `pnpm run eslint` causes errors to appear).

Also confirm that editor integrations work.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


packages/eslint-plugin/src/configs/import.ts
```diff
@@ -0,0 +1,69 @@
+export = {
+  extends: ["plugin:import/recommended"],
+  plugins: ["import"],
+  rules: {
+    "import/no-unresolved": [
```
obulat, line 5: I think this svg rule can be removed since we are not using `svg-module` anymore.

packages/eslint-plugin/src/configs/project.ts
```diff
@@ -0,0 +1,184 @@
+/**
+ * ESLint configuration used by the Openverse project itself.
```
obulat, line 2: I don't quite understand what "used by the Openverse project itself" means here. At first I thought it's for the top-level files, or maybe for the automations etc, but then I saw rules that are for `vue`.

packages/eslint-plugin/src/configs/import.ts
```diff
@@ -0,0 +1,69 @@
+export = {
```
obulat, line 1: A docstring would be really nice here.

packages/eslint-plugin/src/configs/import.ts
sarayourfriend: Removed in [29ee0ab](https://github.com/WordPress/openverse/pull/2973/commits/29ee0abbd725a1053ef39a7018dda1c26064b972)

packages/eslint-plugin/src/configs/project.ts
sarayourfriend: Updated in the latest commits.

packages/eslint-plugin/src/configs/import.ts
sarayourfriend: Added docstrings to each of the config parts in the latest commit.

dhruvkb (approved):
I preferred the earlier setup where the root ESLint config was the final config and `eslint-plugin` had only our _custom rules_. But I understand the reasons behind this and it is definitely an improvement in many ways. LGTM!

obulat (approved):
Everything works well in local testing. Thank you for figuring out the eslint & prettier setup.

A bit unrelated, but what is the best way to run `prettier` now?

sarayourfriend:
> A bit unrelated, but what is the best way to run prettier now?

Should still be `just lint prettier`.

PR #2982, opened 2023-09-05
Wrap usage of `useSensitiveMedia` with `computed`

### LINKED ISSUE #2927: `useSensitiveMedia` composable doesn't update when media changes
## Description

The media sensitivity and blurring is not updated when the global media item changes.

## Reproduction

1. Go to staging.openverse.org/preferences and turn on the 3 sensitivity flags (including the `fake_sensitive`)
2. Narrow the screen to mobile width and search for some audio.
3. Play one of the "non-sensitive" audio items so that the global audio player appears. Then, click on one of the audio items that is blurred because it is sensitive.
4. See error: the audio title is not blurred in the global audio player because the composable is still using the first audio item to detect sensitivity. 
Note: it works the other way around, too: if you first play a sensitive item, then all audio titles will be blurred in the global audio player, even if the items are not sensitive.

## Screenshots

Kapture 2023-08-31 at 09 38 51

## Additional context

To fix this, we should pass the media as a `ref` to the `use-sensitive-media` composable.
I set the priority to high because this must be fixed before rolling out the Frontend blurring of sensitive content project.

dhruvkb:
## Fixes

Fixes #2927 by @obulat

## Description

This PR wraps usages of the `useSensitiveMedia` composable with `computed` so that the value of `shouldBlur` is updated with the prop `audio` changes.

## Testing Instructions

Try to reproduce the problem in the linked issue. If you are unable to, consider the issue fixed.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VAudioThumbnail/VAudioThumbnail.vue
+++ b/frontend/src/components/VAudioThumbnail/VAudioThumbnail.vue
@@ -36,7 +36,7 @@
 </template>
 
 <script lang="ts">
-import { ref, onMounted, defineComponent, PropType } from "vue"
+import { toRefs, ref, onMounted, defineComponent, PropType } from "vue"
 
 import { rand, hash } from "~/utils/prng"
 import { lerp, dist, bezier, Point } from "~/utils/math"
@@ -60,7 +60,8 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const { isHidden: shouldBlur } = useSensitiveMedia(props.audio)
+    const { audio } = toRefs(props)
+    const { isHidden: shouldBlur } = useSensitiveMedia(audio)
 
     const i18n = useI18n()
     const helpText = (
--- a/frontend/src/components/VAudioTrack/layouts/VGlobalLayout.vue
+++ b/frontend/src/components/VAudioTrack/layouts/VGlobalLayout.vue
@@ -20,7 +20,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType } from "vue"
+import { toRefs, defineComponent, PropType } from "vue"
 
 import type { AudioDetail } from "~/types/media"
 
@@ -42,7 +42,8 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const { isHidden: shouldBlur } = useSensitiveMedia(props.audio)
+    const { audio } = toRefs(props)
+    const { isHidden: shouldBlur } = useSensitiveMedia(audio)
 
     return {
       shouldBlur,
--- a/frontend/src/composables/use-sensitive-media.ts
+++ b/frontend/src/composables/use-sensitive-media.ts
@@ -1,4 +1,4 @@
-import { computed } from "vue"
+import { computed, unref, Ref } from "vue"
 
 import type { Media } from "~/types/media"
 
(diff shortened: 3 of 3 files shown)


frontend/src/components/VAudioThumbnail/VAudioThumbnail.vue
```diff
@@ -60,7 +60,9 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const { isHidden: shouldBlur } = useSensitiveMedia(props.audio)
+    const shouldBlur = computed(
+      () => useSensitiveMedia(props.audio).isHidden.value
+    )
```
sarayourfriend, line 65: Calling `useSensitiveMedia` inside the computed callback could be risky if `useSensitiveMedia` provisions `ref` objects, which it does (`isHidden` itself is a `computed`).

Olga suggested the following fix in the issue:

> To fix this, we should pass the media as a ref to the use-sensitive-media composable.

I think this is the better fix and would require less code overall. To implement the fix, we'd just need to wrap `props` in `toRefs`:

```suggestion
    const { audio } = toRefs(props)
    const { isHidden: shouldBlur } = useSensitiveMedia(audio)
```

As Olga pointed out the issue, the problem with the current implementation is that `props.audio` is getting detached from updates to `props`, so `audio` is no longer reactive. The current implement is identical to destructuring from the raw `props` object, as opposed to using `props.audio` inside the ref. `toRefs` ensures that the new `audio` ref pulled from `toRefs(props)` is updated whenever the reactive `props` object is updated.

sarayourfriend (changes requested):
Left a request for changes to the implementation to avoid issues with creating computeds inside of computeds.

frontend/src/components/VAudioThumbnail/VAudioThumbnail.vue
dhruvkb: I was under the impression that both solutions were functionally equivalent and had no idea about it being "risky". Could you elaborate on what the risk is and how? Is that related to the memory leak issue we had a while before?

sarayourfriend: Yes! The memory leak issue we saw before is exactly it. Each time the computed re-runs, the call to `useSensitiveMedia` provisions the computed refs it uses each time instead of reusing the existing ones, and those don't get garbage collected. It might not be visible on the server side because of how the page renders, but this one would be evident client side.

Here's the PR where we fixed it before: https://github.com/WordPress/openverse-frontend/pull/1747

I can't find the explanation we had for why it happened.

If the other implementation is preferable for some reason, you could always profile it locally and see whether it has the same issue.

obulat (commented):
We can use the `unref` to allow for passing `audio` **or** `ref<audio>`.

I think that we only need the ref for global audio, because it's the only element in which the value of `audio` _can change_. For the other audio track components, `audio` stays the same throughout the lifetime of the component.

dhruvkb:
I made the changes and force-pushed 2219a4e (to overwrite d95c8a3), but that doesn't work at all. It needs more investigation.

sarayourfriend (approved):

dhruvkb:
Yeah, the thumbnail also needs to account for the audio change.

PR #2995, opened 2023-09-06
Check flag 'sensitive_content' with 'fetch_sensitive'

### LINKED ISSUE #2994: Fetch sensitive feature flag stays active even if sensitive content feature flag is disabled
## Description

If the sensitive content feature flag is toggled off, the fetch sensitive feature flag stays on, but without the UI being present to toggle it back off. This means the browser gets stuck in a mode where it always fetches sensitive content without any way of turning it off.

Found while testing https://github.com/WordPress/openverse/pull/2955

@dhruvkb I'm not sure whether this is a blocker. I think it would be because it would prevent us from fully rolling back the sensitive content feature, even if we turn the sensitive content feature flag off, the session's flag to make requests for sensitive content would be stuck on. This would be bad if we discovered we _needed_ to roll the feature back because of some critical issue (however unlikely that seems at the moment).

## Reproduction

1.  Visit https://staging.openverse.org/preferences and turn on the sensitive content feature flag
2.  Visit search and turn on the sensitive content fetching
3.  Go back to https://staging.openverse.org/preferences and turn off the sensitive content feature flag
4. Visit search and confirm you can still get sensitive results (or, more easily, that the fetch sensitive feature flag is still on in the preferences). I've confirmed locally that this does still fetch the sensitive results.

To emulate the rollback scenario, test it locally by running the dev server and then:

1. Turn on the sensitive content feature flag and fetch sensitive feature flags
2. Leaving your browser tab open, update `feature-flags.json` to set `sensitive_content` to `disabled` at the staging level. This will also disable it locally.
3. Refresh your browser tab and note that the fetch sensitive feature flag is still on and that your requests will still include sensitive content. This is true on `main` and https://github.com/WordPress/openverse/pull/2955, so it isn't an issue with that PR, but the PR also won't fix this.

@dhruvkb Is the best way to fix this by making checks against `fetch_sensitive` also check the general feature flag? Basically making `fetch_sensitive` dependent on `sensitive_content` also being true? Also, do you agree with my assessment of this being an issue due to rollbacks? Even if we don't block on it, I think we could fix this before #2550 is unblocked anyway (:confused:).

dhruvkb:
## Fixes

Fixes #2994 by @sarayourfriend 

## Description

This PR checks for the `sensitive_content` flag alongside the `fetch_sensitive` flag in places where checking only the latter could cause a situation where sensitive content fetching would be frozen because the UI to toggle it would not be visible.

## Testing Instructions

1. In `/preferences`, enable `fetch_sensitive` but not `sensitive_content`.
2. Search for sensitive content, you should not see any.

With this PR sensitive content will only be fetched when there the UI to disable it is also visible.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/stores/search.ts
+++ b/frontend/src/stores/search.ts
@@ -95,7 +95,12 @@ function computeQueryParams(
   }
 
   // `INCLUDE_SENSITIVE_QUERY_PARAM` is used in the API params, but not shown on the frontend.
-  if (mode === "API" && useFeatureFlagStore().isOn("fetch_sensitive")) {
+  const ffStore = useFeatureFlagStore()
+  if (
+    mode === "API" &&
+    ffStore.isOn("sensitive_content") &&
+    ffStore.isOn("fetch_sensitive")
+  ) {
     search_query[INCLUDE_SENSITIVE_QUERY_PARAM] = "true"
   }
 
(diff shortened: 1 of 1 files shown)


obulat (approved):
LGTM!

sarayourfriend (approved):
LGTM!

PR #3006, opened 2023-09-11
`mature` -> `sensitive` frontend copy and code

### LINKED ISSUE #2623: `mature` -> `sensitive` copy changes for the frontend
## Description

Change copy references of the term "mature" to "sensitive". This work is described in the implementation plan here: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.html#id2

## Additional context

Part of #383

### LINKED ISSUE #2624: `mature` -> `sensitive` code changes for the frontend
## Description

Change **code** references of the term "mature" to "sensitive". This work is described in the implementation plan here: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230517-implementation_plan_mature_copy_change.html#id1

## Additional context

Part of #383

sarayourfriend:
## Fixes

Fixes #2623 by @AetherUnbound 
Fixes #2624 by @AetherUnbound 

Also related to the overall #377 project as the new feature uses "sensitive" across the board and the content report modal is really the last place that was still using the legacy term "mature".

## Description

After doing the copy changes first, the code changes were so minimal (I think because a lot of things have been changed during the sensitive content feature implementation since the implementation plan for these two issues was written) that it did not make sense to separate them. If they were separated, there would be weird decisions like leaving the `mature` key in `en.json5` even though the copy itself was already updated. If we really want to separate them, I can open two separate PRs, but I don't see the value in it anymore since the scope of these issues has so significantly reduced since the implementation plan was written.

The only "dangling" bit of code changes is caused by the Openverse API only accepting "mature" for the report reason.

To handle that, I've created one commit that maps the reason on the frontend to `mature`. That is a standalone commit. However, because it seemed just as straightforward to update the API to accept "sensitive" as the reason, and map it API side, I've added a subsequent commit that removes the mapping on the frontend, and updates the API to map "sensitive" report reason to "mature" (which the database expects). If that's too much to include in this PR and it would be preferable to just map it on the frontend, reviewers, please let me know, and we can revert the commit with API changes.

## Testing Instructions

Checkout the branch and run `just p frontend dev`. Visit the frontend and confirm the content report modal looks good and the options work. That's the only real copy change here. The rest are code changes. Unit tests and e2e tests should all pass. Double check the updated visual regression snapshots to confirm everything there makes sense as well.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
I'll let playwright re-run when I push new changes in a minute or two here and open an issue if those tests are flaky. They passed locally so I believe they probably are flaky.

frontend/src/data/report-service.ts
```diff
@@ -1,5 +1,5 @@
 import type { MediaType } from "~/constants/media"
-import type { ReportReason } from "~/constants/content-report"
+import { ReportReason } from "~/constants/content-report"
```
dhruvkb, line 2: It's only being used as a type, so why this change?

frontend/test/unit/fixtures/audio.js
```diff
@@ -48,7 +47,7 @@ export const getAudioObj = (overrides = {}) =>
       related_url:
         "http://localhost:8000/v1/audio/e19345b8-6937-49f7-a0fd-03bf057efc28/recommendations",
       peaks: [],
-      isSensitive: false,
+      unstable_sensitivity: [],
```
dhruvkb, line 50: 👍

dhruvkb (approved):
Looks good to me! Also thanks for updating the API serializers.

api/api/serializers/media_serializers.py
```diff
@@ -373,8 +384,37 @@ def validate(self, attrs):
             raise serializers.ValidationError(
                 "Description must be at least be 20 characters long"
             )
+
         return attrs
 
+    def _map_reason(self, value):
+        """
+        Map `sensitive` to `mature` of forwards compatibility.
```
obulat, line 392:
```suggestion
        Map `sensitive` to `mature` for forwards compatibility.
```

frontend/src/data/report-service.ts
sarayourfriend: Whoops, just accidentally left in when I removed the frontend adapter code in favour of having the API accept the new term at the serialiser.

sarayourfriend: Fixed in https://github.com/WordPress/openverse/pull/3006/commits/c47176e7e2dd9dc16bef3e969690305116919514

obulat (approved):
When testing locally, with the local API, everything works fine and there are no warnings in the logs or in the console.
It's nice that you can add changes to both the API and the frontend in a single PR thanks to the monorepo :)

sarayourfriend:
>  It's nice that you can add changes to both the API and the frontend in a single PR thanks to the monorepo :)

It is! This is making me realise that the next time we deploy the API and frontend, we need to deploy the API first so that it is ready to handle the new frontend changes. For staging they'll both deploy automatically and the timing doesn't matter for a non-live environment, but it will matter for prod.

I'll add this to the MSR notes for this week and request @stacimc to share this in the MSR hand-off/summary for redundancy.

PR #3011, opened 2023-09-11
Run the app as ASGI

### LINKED ISSUE #2790: Convert the API to ASGI

sarayourfriend:
## Fixes

Fixes #2790 by @sarayourfriend 

## Description

In #3000, I've tried to implement the async client sharing necessary for #2788. The idea of that issue is to do async client instantiation "the right way". However, in that PR, I've discovered the deep complexities of trying to do things this way. There is, in fact, no way to do things the right way in an "async under WSGI" approach that we wouldn't immediately remove after we switched the app to run under ASGI.

On the other hand, switching the app to run under ASGI, without making other changes, does not create any such issues. Django automatically wraps sync views in `sync_to_async` and our application runs perfectly (the integration tests show that).

It does, however, require making some relatively significant changes to the way we deploy the application. Namely, that is to switch to uvicorn and remove gunicorn in favour of scaling our ECS tasks with smaller individual task resources. In other words, rather than having tasks with more resources split their resources with four gunicorn workers, we'll have tasks with fewer individual resources, and scale the number of "workers" by increasing the number of tasks ECS provisions. This removes the gunicorn worker-manager middleman and relies on the orchestrator (ECS in our case) to create the correct number of "workers". Indeed, FastAPI's deployment advice recommends this approach for containerised applications outside of special cases (which I don't think we are): https://fastapi.tiangolo.com/deployment/docker/#one-process-per-container

So, with all of that in mind, this PR does the following:

1. Run the application using uvicorn and convert gunicorn configuration to uvicorn
2. Implement an ASGI handler class that adds lifespan support to Django (compensates for https://code.djangoproject.com/ticket/31508)

This PR will require some changes to the API infrastructure deployment. Specifically, we'll want to do the following before deploying this PR:

1. Increase the number of tasks for the API
2. Then deploy this PR
3. Then scale the task resources down and redeploy

For scaling the resources down, we need to consider the existing settings:

5 tasks with 1 vCPU and 4 GB RAM each.

Each task runs 4 workers. To get the exact same number of workers, we'd need 20 tasks. If we scale each part by a quarter, that is 0.25 vCPU and 1 GB RAM each. Surprising, even with the high number of individual tasks, that puts us at 1k USD less per 30-day period than we spend now.

That being said, I don't think this mathematical approach is going to be reliable. Instead, we should consider that we do not fully exercise our API instances to the maximum capacity. That means there could be room to deploy fewer tasks with more resources or just fewer tasks overall. So long as we match roughly the same worker-to-resource allocation ratio we have now, the price differences are either negligible or offer significant savings.

## Testing Instructions

Checkout this branch and run `just build web` to get the new dependencies. Then run `just api/up` and make requests to the app. Check the logs with `just logs web` and confirm they look good and match the expected output patterns for our logging configuration (it should just match `main`).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Adding @obulat and @dhruvkb to this PR because we discussed the issue in the team meeting. But please, @krysal and @stacimc give y'all's input on whether you think this approach is appropriate and if it makes sense or needs further testing or clarification.

api/conf/settings/base.py
```diff
@@ -60,8 +60,6 @@
     },
 ]
 
-WSGI_APPLICATION = "conf.wsgi.application"
-
```
dhruvkb, line 64: I had a faint recollection that `ASGI_APPLICATION` was needed here. Turns out it was just a requirement for [Daphne](https://docs.djangoproject.com/en/4.2/howto/deployment/asgi/daphne/).

api/run.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+import uvicorn
+
+
+if __name__ == "__main__":
+    uvicorn.run(
```
dhruvkb, line 7: Should we explicitly set [`workers=1`](https://www.uvicorn.org/settings/#production) here (even though that's the default)?

dhruvkb, line 7: The [dev setup](https://www.uvicorn.org/settings/#development) of uvicorn needs the `--reload` flag (or `reload=True`?).

sarayourfriend: Reload is passed on line 11 :slightly_smiling_face:

sarayourfriend: Sure!

api/run.py
```diff
@@ -0,0 +1,21 @@
...
+        "conf.asgi:application",
+        host="0.0.0.0",
+        port=8000,
```
sarayourfriend, line 10:
```suggestion
        port=8000,
        workers=1,
```

api/run.py
dhruvkb: 🤦‍♂️ Ah yes. Sorry, I don't know how I did not see it.

api/conf/urls/__init__.py
```diff
@@ -37,5 +37,5 @@
     path("v1/", include(versioned_paths)),
 ]
 
-if settings.ENVIRONMENT == "local":
-    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
+# if settings.ENVIRONMENT == "local":
+#     urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```
stacimc, line 41: Is this left in intentionally?

stacimc (approved):
LGTM! I really wanted to make sure I wrapped my head around this 😅 Having gone back through the previous PRs and discussions, this makes sense and looks like a good (and exciting!) path forward.

Everything worked well in my testing, including the comparison of logs to `main`. I also tested the static file handling by changing the `ENVIRONMENT` variable locally. Thanks for the additional explanation in the commit messages 👍

krysal (approved):
LGTM 🚀 I left just a few non-blocking questions. Thanks for explaining the move in so much detail. I surfed a bit the endpoints and everything seems to be working normally.

api/conf/urls/__init__.py
sarayourfriend: Nope! It's unnecessary now with the ASGI handler approach for the static files and I just forgot to remove the commented old code :sweat_smile: 

```suggestion
```

sarayourfriend, line 41:
```suggestion
```

api/Pipfile
```diff
@@ -34,14 +34,14 @@ djangorestframework = "~=3.14"
...
 future = "~=0.18"
-gunicorn = "~=21.2"
 limit = "~=0.2"
 Pillow = "~=10.0"
 psycopg2 = "~=2.9"
 python-decouple = "~=3.8"
 python-xmp-toolkit = "~=2.0"
 sentry-sdk = "~=1.30"
 django-split-settings = "*"
+uvicorn = {extras = ["standard"], version = "*"}
```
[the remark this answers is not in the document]
sarayourfriend: Sure! After working with pipenv locally a bit more and seeing how it updates dependencies, I don't think these versions do very much for us. It will update any dependency that has any newer version that matches the constraint any time it runs `install` :cry:

sarayourfriend:
I'm going to hold off merging this until after the Pillow update is released. It'll be better to orchestrate the necessary infrastructures changes anyway. Adding DO NOT MERGE to the title for posterity.

api/conf/urls/__init__.py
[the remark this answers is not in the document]
sarayourfriend: Yes, done in https://github.com/WordPress/openverse/pull/3011/commits/10a4c47d25ea45a7c7baedc93fb022e7a1a79968

sarayourfriend: Done in https://github.com/WordPress/openverse/pull/3011/commits/10a4c47d25ea45a7c7baedc93fb022e7a1a79968.

sarayourfriend:
This PR depends on two related infrastructure changes:

https://github.com/WordPress/openverse-infrastructure/pull/618 to enable a canary service, to prevent increasing the task count from making our deployments run for hours.

https://github.com/WordPress/openverse-infrastructure/pull/616 to resize the tasks and desired counts to accommodate a one-worker-per-task configuration we will end up with after deploying this PR.

sarayourfriend:
I've rebased this branch and re-written the history a bit to squash a few commits that didn't need to be separated from other ones and were complicating rebases (especially changes to `Pipfile` et co).

This PR should be ready to go, as long as it passes CI, and tried out in staging, after staging is increased to 4 tasks by https://github.com/WordPress/openverse-infrastructure/pull/616.

api/Pipfile
```diff
@@ -34,15 +34,14 @@ drf-spectacular = "*"
...
 future = "~=0.18"
-gunicorn = "~=21.2"
 limit = "~=0.2"
 Pillow = "~=10.0.1"
 psycopg2 = "~=2.9"
 python-decouple = "~=3.8"
 python-xmp-toolkit = "~=2.0"
 sentry-sdk = "~=1.30"
 django-split-settings = "*"
-uvloop = "~=0.17"
```
krysal, line 45: Is this removing `uvloop`?

sarayourfriend: [`uvicorn[standard]` already includes it along with other significant improvements to the stock asyncio interface and sets the asyncio loop policy for it](https://www.uvicorn.org/#quickstart), so we don't need to add an explicit dependency (which could lead to weird version drift between uvicorn and uvloop if we don't update them together). We do still need to keep the uvloop policy setting in `manage.py`, though, because we don't run that with uvicorn, just regular old Python.

AetherUnbound (approved):
ASGI build runs great for me locally, code looks good and makes sense!

sarayourfriend:
We'll remove the do not merge label once the changes to staging task count is ready.

PR #3013, opened 2023-09-11
Populate sensitivities for single result view

### LINKED ISSUE #2926: Single result endpoint does not return correctly populated `sensitivity`
## Description

The single results endpoint does not populate `unstable__sensitivity`. This prevents the frontend from blurring sensitive results loaded from direct links.

Reproduction is easy on production. I don't share direct links because I am using results that are designated with sensitive textual content in production (though I believe this example to be a false positive). It is possible to reproduce locally using the testing sensitive terms present locally (running, water, and birds). Just choose a result that is designated to have sensitive text by the testing terms and load it's single result. It will not have the sensitive textual content designation in `unstable__sensitivity`.

## Reproduction

1.  Search `api.openverse.engineering/v1/audio/` with `?q=Sofia's Dawn Chorus&unstable__include_sensitive_results=true`. This returns just one result titled "Sofia's Dawn Chorus" that is designated as having sensitive textual content. Verify this in the `unstable__sensitivity` array on the result in the results list.
2.  Visit the detail url for that result, UUID `06ea8f9b-68b0-4ffd-939b-cb020ad70b0e`.
3.  `unstable__sensitivity` will be an empty array.

## Additional context

Marked high because it prevents the frontend's sensitivity feature roll out. cc @dhruvkb as a heads up to the project lead.

dhruvkb:
## Fixes

Fixes #2926 by @sarayourfriend 

## Description

This PR populates the sensitivies for single result view by manually defining the `retrieve` view in the `MediaViewSet` and extending the `SearchContext` to work for single results too.

## Testing Instructions

1. Check out the branch and run the dev server `just api/up` (and `just api/init` if you haven't already)
0. Find a media item that has some sensitive content. To do so, add the the query param `unstable__include_sensitive_results=true` to the URL.
1. Open the same media's detail page, you should see the same sensitivities as the results page.
2. Report some media item as mature. Do the same for this media item. You should see identical sensitivity in the search resuls and single result pages.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/search_context.py
+++ b/api/api/utils/search_context.py
@@ -1,15 +1,11 @@
 from dataclasses import asdict, dataclass
-from typing import Protocol, Self
+from typing import Self
 
 from django.conf import settings
 
 from elasticsearch_dsl import Q, Search
 
 from api.constants.media_types import OriginIndex
-
-
-class Identifiable(Protocol):
-    identifier: str
 
 
 @dataclass
@@ -25,11 +21,11 @@ class SearchContext:
     """Subset of result identifiers for results with sensitive textual content."""
 
     @classmethod
-    def build(cls, results: list[Identifiable], origin_index: OriginIndex) -> Self:
-        if not results:
+    def build(
+        cls, all_result_identifiers: list[str], origin_index: OriginIndex
+    ) -> Self:
+        if not all_result_identifiers:
             return cls(list(), set())
-
-        all_result_identifiers = [result.identifier for result in results]
 
         if not settings.ENABLE_FILTERED_INDEX_QUERIES:
             return cls(all_result_identifiers, set())
@@ -46,17 +42,14 @@ class SearchContext:
                 "terms",
                 **{"identifier.keyword": all_result_identifiers},
             )
-            if len(results) > 1
-            else Q(
-                "term",
-                **{"identifier.keyword": all_result_identifiers[0]},
-            )
         )
 
         # The default query size is 10, so we need to slice the query
         # to change the size to be big enough to encompass all the
         # results.
-        results_in_filtered_index = filtered_index_search[: len(results)].execute()
+        results_in_filtered_index = filtered_index_search[
+            : len(all_result_identifiers)
+        ].execute()
         filtered_index_identifiers = {
             result.identifier for result in results_in_filtered_index
         }
--- a/api/api/utils/search_context.py
+++ b/api/api/utils/search_context.py
@@ -44,7 +44,12 @@ class SearchContext:
             # cf: https://github.com/WordPress/openverse/issues/2154
             Q(
                 "terms",
-                **{"identifier.keyword": [result.identifier for result in results]},
+                **{"identifier.keyword": all_result_identifiers},
+            )
+            if len(results) > 1
+            else Q(
+                "term",
+                **{"identifier.keyword": all_result_identifiers[0]},
             )
         )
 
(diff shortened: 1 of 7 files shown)


api/api/utils/search_context.py
```diff
@@ -41,7 +44,12 @@ def build(cls, results: list[Hit], origin_index: OriginIndex) -> Self:
...
             Q(
                 "terms",
-                **{"identifier.keyword": [result.identifier for result in results]},
+                **{"identifier.keyword": all_result_identifiers},
+            )
+            if len(results) > 1
+            else Q(
+                "term",
+                **{"identifier.keyword": all_result_identifiers[0]},
             )
```
sarayourfriend, line 53: This is fine, but FWIW I don't think it's entirely necessary. `terms` should still work even with a list of 1 and I'd be surprised if there is a significant performance difference between the two. Have you already measured that? The only reason I'd suggest not doing this is (a) it feels a little like pre-optimising and (b) it makes the code slightly more obscure.

api/api/utils/search_context.py
```diff
@@ -1,32 +1,35 @@
...
 from django.conf import settings
 
 from elasticsearch_dsl import Q, Search
-from elasticsearch_dsl.response import Hit
 
 from api.constants.media_types import OriginIndex
 
 
+class Identifiable(Protocol):
+    identifier: str
```
sarayourfriend, line 12: Cool!

api/api/views/media_views.py
```diff
@@ -87,6 +89,20 @@ def get_db_results(self, results):
...
 
+    def retrieve(self, request, *_, **__):
+        @dataclass
+        class PseudoHit:  # implements Identifiable protocol
+            identifier: str
+
+        instance = self.get_object()
+        hit = PseudoHit(identifier=str(instance.identifier))
+
+        search_context = SearchContext.build([hit], self.default_index).asdict()
```
sarayourfriend, line 100: FWIW, `SearchContext::build` could be simplified if it just received the list of identifiers. Considering it gets mapped in the `::build` function out of the `results` list anyway, and those results aren't used in any _other_ way, I think it would be a nice way to simplify the `::build` function and the type juggling we're doing. It prevents us from needing to create a thin container class to satisfy `SearchContext` and reduces `SearchContext`'s concerns by restricting it to just the identifiers, the part of the results it actually cares about.

```suggestion
        instance = self.get_object()
        search_context = SearchContext.build([instance], self.default_index).asdict()
```

sarayourfriend (changes requested):
LGTM! My only request for a change is to add a unit or integration test against the single result endpoint to check that the `unstable__sensitivity` is populated when we expect it to be. We can use the local testing terms found in `mock_sensitive_terms.txt` to help find a result marked sensitive in the local data to test with.

The other suggestions to reduce `SearchContext` to just the identifiers is just a suggestion :slightly_smiling_face:

dhruvkb:
Added integration tests in 15c12b7 and simplified the `SearchContext` implementation to avoid dataclasses and protocols and deal with simple `list[str]` instead.

sarayourfriend (approved):
LGTM!

stacimc (approved):
Took me awhile to figure out how to set everything up for the manual testing, sorry 😅 Got everything working and this looks good to me!

dhruvkb:
Thank you for the effort @stacimc!

PR #3021, opened 2023-09-13
Resolve flakiness in global audio player VR tests

### LINKED ISSUE #3009: Flaky visual regression test: `visual-regression/components/global-audio-player.spec.ts:15:9 › screen at breakpoint xs with width 340 › Global audio player on the search page - rtl`
## Description

Failed run: https://github.com/WordPress/openverse/actions/runs/6140660047/job/16659950996
Successful run (no relevant changes between the commits): https://github.com/WordPress/openverse/actions/runs/6140997596/job/16660728211

Debugging artefact:
[playwright_vr_test_results.zip](https://github.com/WordPress/openverse/files/12571843/playwright_vr_test_results.zip)
 
## Additional context

Marked as critical according to our flaky test procedure.

dhruvkb:
## Fixes

Fixes #3009 by @sarayourfriend 

## Description

This PR
- <s>updates the track used in the global audio player interactions to a longer one, allowing more time to test the pause + seek interactions.</s>
- adds a check for network idle to allow the track to buffer to the point of the click, ensuring the yellow fill gets populated.

## Testing Instructions

As this is a flaky test, the only way to ensure it passes consistently is to repeat the test many times and expect no failures.

```bash
$ count=1; while pnpm test:playwright visual-regression/components/global-audio-player.spec.ts; do (( count++ )) && echo "$count"; done
```

I ran this script and it completed over 30 consecutive iterations without failures.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/visual-regression/components/global-audio-player.spec.ts
```diff
@@ -13,22 +13,25 @@ import audio from "~~/test/playwright/utils/audio"
 for (const dir of languageDirections) {
   breakpoints.describeXs(async ({ expectSnapshot }) => {
     // https://github.com/WordPress/openverse/issues/3009
```
sarayourfriend, line 15:
```suggestion
```

sarayourfriend (approved):
LGTM, but wondering @dhruvkb if we should be intercepting the network call for the audio like we do for images and returning a dummy audio file? Maybe we already are :thinking:

dhruvkb:
We should definitely consider intercepting outgoing requests. A locally present file should reduce any network-lag considerably and prevent requests going out to the provider (possibly affecting their popularity metrics). But I would prefer to merge this PR as-is, because it fixes the flake, and consider intercepts as a separate feature.

I couldn't find the code in the VR tests where we intercept image requests and return a sample, can you point me to that? Possibly unrelated: I though this was it but it's in E2E and I didn't find any references to these resources either.

Screenshot 2023-09-14 at 1 39 17 AM

sarayourfriend:
@dhruvkb You're right, I can't find it either. @obulat might know best.

FWIW it looks like we have a utility to mock audio provider requests: https://github.com//WordPress/openverse/blob/9471728b4686d054493cd4528554e957dfee4350/frontend/test/playwright/utils/route.ts#L2-L3

Maybe the fix for this flaky test would best be applied by calling that utility (potentially in addition to waiting for networkidle). Those intercepted API calls don't return audio files though, so the test wouldn't really be able to test seeking, I assume. :thinking: Might just need a separate issue for all this.

obulat (commented):
When I tried to fix this issue previously, I especially chose the shortest audio: using a long audio causes more problems due to loading/playing/pausing (when you expect the yellow focus part to be in specific spot, but it's not, because the audio isn't fully loaded yet).

For the storybook image, I added a base64-encoded image to the fixtures, and used its URL as the storybook URL. Maybe we can do the same here: replace the audio URL in the search results tape with the base64-encoded audio?

sarayourfriend:
> replace the audio URL in the search results tape with the base64-encoded audio?

If we do this, please lets use a script to automatically do so (and check for it in CI) so that the next time we run `update-tapes` we don't wonder why all our tests are flaky again :sweat_smile:

dhruvkb:
@obulat using a short audio can be fine too as shown by 35 consequtive passing iterations on 51846b0 in my testing.

obulat (approved):

PR #3025, opened 2023-09-14
Prevent iNaturalist from running alongside any other DAGs

### LINKED ISSUE #1276: Prevent iNaturalist from running alongside any other DAGs
## Description

Similar to WordPress/openverse#1277 in reasoning, iNaturalist can be intensive and disruptive for the other DAGs running on the instance. If possible, we would like to prevent iNaturalist from running while other DAGs are in progress. One way to do this would be to leverage [Airflow pools](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/pools.html). Currently most of our tasks sit in the `default_pool`, which has 128 slots. (We could increase this to a higher number if desired).

Tasks which use pools can [request multiple slots](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/pools.html#using-multiple-pool-slots). One way to prevent iNaturalist concurrency would be to to set its `pool_slots` to the maximum size for `default_pool`, meaning it would require all pool slots to be available. This, paired with a reduced [priority weight](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/priority-weight.html) would mean that all other tasks would run prior to iNaturalist, and the latter could only run if all slots are available. This would apply per-task, so even if iNaturalist ran a task, as soon as it completed a task it would free up the pool slots for other DAGs.

## Additional context

See WordPress/openverse#1277 for an additional/alternate option.

rwidom:
## Fixes

Fixes #1276  by @AetherUnbound 

## Description

The longest running part of the iNaturalist DAG is joining the source tables for insertion into `image`, which happens in batches of approximately 2 million records at a time. Added required slots and lower priority to those tasks so that they can happen in between other dag run tasks, and lowered the batch size to shorten the interruptions to the rest of the schedule.

## Testing Instructions

Running `just catalog/test` is going fine, but see #3631 .

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
Hi @rwidom 👋 Do you plan to continue this work? Need help with anything? Lowering the batch size sounds like a good idea to me!

rwidom:
> Hi @rwidom 👋 Do you plan to continue this work? Need help with anything? Lowering the batch size sounds like a good idea to me!

Hey, sorry to be gone for a bit. I'm back and did update the batch size here. What are your thoughts on the catalog of life test data @krysal ? I'm getting back up to speed, but I'll try and take another look today.

krysal (approved):
The logic looks good to me :) Thanks for coming back to this!

stacimc (approved):
Very cool! Looks great, especially since iNaturalist is already offset to not run on the same day as the other monthly DAGs. 🚀 


I was initially concerned that this would prevent us from running a high-priority manually-triggered DAG while iNaturalist is running (like an urgent index update or something), especially since the load steps take days to finish. But since these are mapped tasks and you've given them a lower priority than all other DAGs, I think we would only ever have to wait for a single batch to finish processing. From production data, I think that's max 40 minutes or so. In some sort of emergency case, we could even fail the running iNaturalist task and clear it once we've finished whatever else we needed to do :) So I'm comfortable with that, and pleased to see another advantage to those mapped tasks! 

I'm also going to add a note to the DAG Status page not to leave DAGs disabled while they have a dagrun in the `running` state, because that would hold a slot and prevent iNaturalist from completing indefinitely. I don't think that's something we can really prevent here, though -- LGTM 🚢

Issue #3028, opened 2023-09-15
API memory leak


sarayourfriend:
## Description

The API's memory consumption pattern changed significantly in the last month.

![image](https://github.com/WordPress/openverse/assets/24264157/e2489b43-2fa0-47c0-b58d-0ec30126126a)

This corresponds with the switch over to ES 8 in production and with this changelog: https://docs.openverse.org/changelogs/api/2023.08.28.20.43.01.html

The changelog doesn't have any changes that seem like they could possibly be related, so I suspect this has something to do with the ES 8 switch.

First thing that comes to mind is that we forgot to update the ES-py client version to match the cluster.

I'll get a PR for that up ASAP and then run a memory analysis locally to see if I can identify a leak.

Marked as critical because we need to resolve this before the ASGI switch.

sarayourfriend:
I've assigned this to myself as I'm working on the ES client upgrade, but will need to hand this off if that doesn't solve the issue.

sarayourfriend:
We deployed the ES client upgrade 12 hours ago, but it has not solved the memory leak. @obulat identified that it is still happening in production.

I unassigned myself the issue and am pinging @WordPress/openverse-api to hand off to someone else.

PR #3029, opened 2023-09-15
Upgrade ES dependencies to match cluster version

sarayourfriend:
## Fixes

Related to #3028 by @sarayourfriend 
Fixes #2744 by @obulat 

## Description

Adds an explicit dependency on the Elasticsearch client library, `elasticsearch-py` that is fixed to our cluster version (8.8.2). This also updates the DSL library to the latest version 8.9, which is the first version to support ES 8. We hardly use the DSL in any significant way here anyway, so I don't think the version mismatch will be an issue.

I've done this in the API and the ingestion server. I've also updated the code to work with the new version. I read through the breaking changes and didn't see anything that would affect us, but I need to look again more closely for the ingestion server, with whose ES usage I'm less familiar.

https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/migration.html

## Testing Instructions

CI should pass. Build should pass and unit and integration tests. To test locally, you must build `web` and `ingestion_server` (`just build web && just build ingestion_server`) to get the new dependencies. Then run `just api/up` and `just api/init` and make searches and test the Django app generally. `just api/init` exercises the ingestion server in addition to what the explicit integration tests test.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/models/media.py
```diff
@@ -275,14 +275,15 @@ def _perform_index_update(self, method: str, raise_errors: bool, **es_method_arg
                     refresh=True,
                     **es_method_args,
                 )
-            except TransportError as e:
-                if e.status_code == 404:
+            except (NotFoundError, BadRequestError) as e:
+                if isinstance(e, NotFoundError):
```
dhruvkb, line 279: Why not two separate except clauses for `NotFoundError` and `BadRequestError`? Also if we are raising the latter again, why do we need to catch it at all?

api/test/unit/conftest.py
```diff
@@ -129,8 +159,16 @@ def cleanup_elasticsearch_test_documents(request, settings):
 
     es: Elasticsearch = settings.ES
 
+    # If pook was activated by a test and not deactivated
+    # (usually because the test failed and something prevent
+    # pook from cleaning up after itself), disable here so that
```
dhruvkb, line 164: I didn't know there was a possibility that pook could not clean up. That's quite bad for the isolation of the tests.

ingestion_server/ingestion_server/es_helpers.py
```diff
@@ -45,9 +46,16 @@ def _elasticsearch_connect() -> Elasticsearch:
...
 
+    elasticsearch_scheme = config("ELASTICSEARCH_SCHEME", default="http://")
     elasticsearch_url = config("ELASTICSEARCH_URL", default="localhost")
     elasticsearch_port = config("ELASTICSEARCH_PORT", default=9200, cast=int)
 
+    es_endpoint = "{scheme}{url}:{port}".format(
+        scheme=elasticsearch_scheme,
+        url=elasticsearch_url,
+        port=elasticsearch_port,
+    )
```
dhruvkb, line 57: Why not f-strings? It's been a long time since I've seen `.format` being used.

sarayourfriend:
This is marked critical as part of the API memory leak issue. I don't know that this will fix the problem but I wouldn't be surprised, given the changes in ES, if a client version mismatch caused something like this.

I'm out of time today, but if someone else is able to reproduce the memory increase on main and then check if it's fixed here, that would be great. We need to do this update _regardless_, and deploying it does not need to happen _now_ because any deployment causes the memory to drop back down. My thinking is that we can just deploy this at the start of the week like we usually do, see if the memory leak is fixed, and if not, start digging into other areas (if we aren't able to reproduce it locally and identify the root cause).

In any case, it needs to be reviewed "ASAP" for us to be able to deploy it early next week, so it is still "critical" in that respect. This also needs to be deployed before #3011 to avoid too many significant changes going in at once. That PR still needs infrastructure PR work anyway.

dhruvkb (approved):
Code-change look good so to me. I'll wait for CI to complete before approving. I also have two small nits.

api/api/models/media.py
```diff
@@ -6,7 +6,7 @@
 from django.db import models
 from django.utils.html import format_html
 
-from elasticsearch import Elasticsearch, TransportError
+from elasticsearch import BadRequestError, Elasticsearch, NotFoundError
```
sarayourfriend, line 9: TransportError changed and it's much easier/clearer to catch the specific errors we care about.

Realising now that we can change this to only import BadRequestError and ignore NotFoundError completely, just letting it raise the way we want it to, now that we're not catching a unified error type.

api/conf/settings/elasticsearch.py
```diff
@@ -2,24 +2,28 @@
...
 
     :return: An Elasticsearch connection object.
     """
 
+    es_scheme = config("ELASTICSEARCH_SCHEME", default="http://")
     es_url = config("ELASTICSEARCH_URL", default="localhost")
     es_port = config("ELASTICSEARCH_PORT", default=9200, cast=int)
     es_aws_region = config("ELASTICSEARCH_AWS_REGION", default="us-east-1")
 
+    es_endpoint = f"{es_scheme}{es_url}:{es_port}"
```
sarayourfriend, line 25: It is no longer possible to easily construct the endpoint from the Elasticsearch client internals. We do this in tests. That was hacky anyway. It's much clearer and cleaner to just save and export it from the place it's centrally built anyway.

api/test/factory/models/media.py
```diff
@@ -95,6 +96,12 @@ def create(cls, *args, **kwargs) -> AbstractMedia | tuple[AbstractMedia, Hit]:
         skip_es = kwargs.pop("skip_es", False)
         with_hit = kwargs.pop("with_hit", False)
 
+        pook_active = pook.isactive()
+        if pook_active:
+            # Temporarily disable pook so that the calls to ES to create
+            # the factory document don't fail
+            pook.disable()
```
sarayourfriend, line 103: This change (and the change later on in this function) makes it so we don't need to juggle pook inside a function to avoid it catching factory requests. Now `@pook.on` can be applied to a whole test function without worry.

I included this in this PR because as part of the ES client's breaking changes, I had to make several updates to tests, mostly removing really flaky and ugly implementation-specific mocks and replacing them with pook matchers. Needing to juggle pook being on or off in all those functions was too tedious.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -714,6 +716,7 @@ def test_post_process_results_recurses_as_needed(
         .body(re.compile('from":0'))
         .times(1)
         .reply(200)
+        .header("x-elastic-product", "Elasticsearch")
```
sarayourfriend, line 719: This really is necessary. The client checks for this header and throws an error about "not supporting an unknown product" :sweat_smile:. It's an easy one trick, luckily!

api/test/unit/conftest.py
sarayourfriend: I've only seen it in some cases, and I wonder if it has to do with async stuff, but yes, I have seen this particular factory run and have errors logged at the end of the test run because pook was still intercepting requests. IIRC it was only on tests that relied on `pook.use` as a context manager, where the code inside the context raised an exception, so there may be some kind of bug in pook there.

api/api/models/media.py
sarayourfriend: Totally! I only realised this while annotating the changes just now. Will remove the BadRequestError here.

api/api/models/media.py
sarayourfriend: Cleaned up in https://github.com/WordPress/openverse/pull/3029/commits/fe96cb145ddd7ea8548805411b208dfd8ab47c7b

api/api/models/media.py
sarayourfriend: Cleaned up in https://github.com/WordPress/openverse/pull/3029/commits/fe96cb145ddd7ea8548805411b208dfd8ab47c7b

ingestion_server/ingestion_server/es_helpers.py
sarayourfriend: Cleaned up in https://github.com/WordPress/openverse/pull/3029/commits/a4402065a62eb1bcc9a80d38f2b9915b317e6a96

api/api/models/media.py
```diff
@@ -275,17 +275,14 @@ def _perform_index_update(self, method: str, raise_errors: bool, **es_method_arg
...
-                else:
-                    raise e
+            except NotFoundError:
+                # This is expected for the filtered index, but we should still
+                # log, just in case.
+                logger.warning(
+                    f"Document with _id {document_id} not found "
+                    f"in {index} index. No update performed."
+                )
+                continue
```
dhruvkb, line 285: Super-nit: I also realised, this `continue` is redundant here.

sarayourfriend: I left it in to make explicit that the loop is meant to continue in spite of the 404. Do you think that's necessary? Or is there a better way to communicate that (I suppose a comment would do).

dhruvkb: No problem either way, we can leave it in.

api/conf/settings/elasticsearch.py
obulat: Not for this PR, but will we need this endpoint for the ingestion server tests in the future? The functions seem so similar, and the return is one of the biggest differences.

ingestion_server/test/integration_test.py
```diff
@@ -464,16 +465,17 @@ def test_task_count_after_two(self):
...
-        return Elasticsearch(
-            host="localhost",
-            port=service_ports["es"],
-            connection_class=RequestsHttpConnection,
-            timeout=10,
+        endpoint = f"http://localhost:{service_ports['es']}"
+        es = Elasticsearch(
+            endpoint,
+            node_class=RequestsHttpNode,
+            request_timeout=10,
```
obulat, line 472: Why was `timeout` changed to `request_timeout` here? These seem to be different values.

Also, the `http_auth` from line 475 is [deprecated](https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/migration.html#migration-options), and should be replaced with `basic_auth` (I don't know if it has any effect when the value is None, though)

ingestion_server/ingestion_server/es_helpers.py
```diff
@@ -55,19 +59,18 @@ def _elasticsearch_connect() -> Elasticsearch:
...
         aws_service="es",
     )
     auth.encode = lambda x: bytes(x.encode("utf-8"))
     es = Elasticsearch(
-        host=elasticsearch_url,
-        port=elasticsearch_port,
-        connection_class=RequestsHttpConnection,
+        es_endpoint,
+        node_class=RequestsHttpNode,
         http_auth=auth,
```
obulat, line 74: `http_auth` from line 74 is [deprecated](https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/migration.html#migration-options), and should be replaced with `basic_auth`.

obulat (changes requested):
So nice that the DSL library is finally updated!
Thank you for opening this PR and explaining the changes to the pook setup. My main blocker here is replacement of `http_auth` with `basic_auth`. It's deprecated, so probably still usable, but it's better to replace it now than wait until its completely removed.

ingestion_server/ingestion_server/es_helpers.py
sarayourfriend: Thanks for catching that!

ingestion_server/test/integration_test.py
sarayourfriend: `timeout` does not exist any more. `request_timeout` configures the length of time to wait for a request to ES before timing out: https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/config.html#timeouts

Here's the documentation for the old `timeout` parameter, which appears to do the same thing, but by a different name: https://elasticsearch-py.readthedocs.io/en/v7.17.11/api.html?highlight=timeout#timeout

Both configure the global request timeout for all requests, as far as I can tell? Are you finding anything different that I'm missing?

api/conf/settings/elasticsearch.py
sarayourfriend: Maybe? The ingestion server tests already constructs its own ES connection, it doesn't re-use the code's, so I don't know how relevant each implementation is to the other at the moment (not to mention we don't have the facility to share Python code between projects yet, I don't think).

ingestion_server/ingestion_server/es_helpers.py
sarayourfriend: BTW, do you know if we still need this authentication stuff? Our clusters are not accessible on the public internet and we don't need any auth to make connections via our jumpbox, for example :thinking: This is old code, and I'm not sure how to determine whether it's necessary. Something to keep an eye on though in case these parameters keep changing or get more complex.

ingestion_server/test/integration_test.py
sarayourfriend: Elasticsearch does not support AWS auth directly, and trying to pass it to `basic_auth` breaks the integration. I've found a nasty workaround, but it would probably be much better to find out if we can remove this bit altogether.

sarayourfriend: Removed in https://github.com/WordPress/openverse/pull/3029/commits/f138fec3e865d5e1f4959c436c8f435a56259286

ingestion_server/ingestion_server/es_helpers.py
sarayourfriend: I've removed it in https://github.com/WordPress/openverse/pull/3029/commits/f138fec3e865d5e1f4959c436c8f435a56259286 as per the other conversation about the parameter.

ingestion_server/test/integration_test.py
obulat: Sorry, you're right.  I was reading the `__init__` code of the Elasticsearch class, and saw both `timeout` and `request_timeout` there, but didn't see the check that shows that `timeout` is deprecated below that: `"The 'timeout' parameter is deprecated in favor of 'request_timeout'"`.
Both seem to configure the global request timeouts for all requests at `init` time. It is also possible to set a different `request_timeout` for the specific request using `options`: `es.options(request_timeout=5).search()`: https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/config.html#timeouts

> I'll go ahead and remove it and we can see whether staging is still able to communicate with Elasticsearch (I really can't see why it wouldn't, but I could very easily be missing something) and roll back to the workaround if staging fails.
So, the `auth` failure would be something to watch for when deploying this change, right?

ingestion_server/ingestion_server/es_helpers.py
obulat: > BTW, do you know if we still need this authentication stuff? 

I really don't know, and your explanations sound reasonable. Let's try :)

obulat (approved):

PR #3035, opened 2023-09-19
Correctly set up constant fields in API responses

### LINKED ISSUE #3034: Fix the API documentation set up for the constant fields
## Problem

Oembed endpoint returns several constant values that are not derived from the Image model.
They are set up as `ReadOnlyField`s.
```py
    version = serializers.ReadOnlyField(
        help_text="The oEmbed version number, always set to 1.0.",
        default="1.0",
    )
```
DRF-spectacular is trying to find `version` field in the `Image` model, and throws a warning when it cannot:

```
/api/api/serializers/image_serializers.py: Warning [ImageViewSet > OembedSerializer]: could not resolve field on model <class 'api.models.image.Image'> with path "version". This is likely a custom field that does some unknown magic. Maybe consider annotating the field/property? Defaulting to "string". (Exception: Image has no field named 'version')
```
## Description
Update the serializer to use a `serializers.SerializerMethodField`:

```py
version = serializers.SerializerMethodField(help_text="The oEmbed version number, always set to 1.0.")

@staticmethod
def get_version() -> Literal["1.0"]:
    return "1.0"
```

The list of fields that need to be updated:

`OembedSerializer`: `type`, `version`
`ProviderSerializer`: `logo_url`

## Alternatives

Leave the fields as they are, the warnings don't break the documentation and the default can be fine.

obulat:
## Fixes

Fixes #3034 by @obulat

## Description

This PR replaces `ReadOnlyField`s with `SerializerMethodField`s that return constant values for values that never change and are not derived from models:
`OEmbedSerializer`: `version`, `type`
`ProviderSerializer`: `logo_url`

This makes the documentation a little bit clearer and fixes the warnings in logs.

Currently, `logo_url` type is given as `string`, but it always returns `null`. The updated version shows `string or null`. I tried setting it to "only null", but spectacular still throws a warning that it couldn't get the type of the value. `string or null` is more accurate than `string`.
Provider serializer logo url type

### OEmbed serializer
Currently, the documentation says that these values are "default", making it seem like a different value is also possible.
OEmbed Serializer current

In this PR, the documentation says "Value: 1.0", making it clear that it's only a single value.
OEmbed Serializer updated

## Testing Instructions

If you run the app on `main` (`just up`), and open the documentation (e.g. http://0.0.0.0:50280/v1/#tag/audio/operation/audio_stats), you should see the warnings mentioned in the related issue in the Docker logs.
These warnings should not be visible in this PR. The documentation for these fields should be updated to match snapshots in the PR description.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/image_serializers.py
+++ b/api/api/serializers/image_serializers.py
@@ -1,3 +1,5 @@
+from typing import Literal
+
 from rest_framework import serializers
 
 from api.constants.field_order import field_position_map
@@ -133,14 +135,13 @@ class OembedSerializer(BaseModelSerializer):
     spec: https://oembed.com.
     """
 
-    version = serializers.ReadOnlyField(
+    version = serializers.SerializerMethodField(
         help_text="The oEmbed version number, always set to 1.0.",
-        default="1.0",
     )
-    type = serializers.ReadOnlyField(
+    type = serializers.SerializerMethodField(
         help_text="The resource type, always set to 'photo' for images.",
-        default="photo",
     )
+
     width = serializers.SerializerMethodField(
         help_text="The width of the image in pixels."
     )
@@ -155,6 +156,14 @@ class OembedSerializer(BaseModelSerializer):
         help_text="A direct link to the media creator.",  # copied from ``Image``
         source="creator_url",
     )
+
+    @staticmethod
+    def get_type() -> Literal["photo"]:
+        return "photo"
+
+    @staticmethod
+    def get_version() -> Literal["1.0"]:
+        return "1.0"
 
     class Meta:
         model = Image
--- a/api/api/serializers/image_serializers.py
+++ b/api/api/serializers/image_serializers.py
@@ -158,11 +158,11 @@ class OembedSerializer(BaseModelSerializer):
     )
 
     @staticmethod
-    def get_type() -> Literal["photo"]:
+    def get_type(*_) -> Literal["photo"]:
         return "photo"
 
     @staticmethod
-    def get_version() -> Literal["1.0"]:
+    def get_version(*_) -> Literal["1.0"]:
         return "1.0"
 
     class Meta:
--- a/api/api/serializers/provider_serializers.py
+++ b/api/api/serializers/provider_serializers.py
@@ -29,7 +29,7 @@ class ProviderSerializer(serializers.ModelSerializer):
     )
 
     @staticmethod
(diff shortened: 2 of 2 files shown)


dhruvkb (approved):
While the change to the documentation is good, something feels off about using a "method" field to generate a constant value. One alternative might be to create a custom field but that seems like over-engineering. 

Approving because I don't know a better way.

obulat:
Added args that are discarded (`*args`) to the `get_type` and `get_version` functions to fix `TypeError: OembedSerializer.get_version() takes 0 positional arguments but 1 was given` error that's causing error in the oembed endpoint: [490095d](https://github.com/WordPress/openverse/pull/3035/commits/490095d62d3cc20d1e4dcc870c698cd144c5ea22)

sarayourfriend:
[The swagger docs say, regarding "null"](https://swagger.io/docs/specification/data-models/data-types/) (TIL):

> Note that there is no null type; instead, the [nullable](https://swagger.io/docs/specification/data-models/data-types/#null) attribute is used as a modifier of the base type.

So a field cannot _only_ be null, it seems, and within spec.

For the static fields, I wrote this custom field that seems to work:

```
class StaticField(serializers.Field):
    Value = TypeVar("Value")

    def __new__(cls, value, **kwargs):
        decorated = extend_schema_field({"type": "string", "nullable": True} if value is None else type(value))(cls)
        return super().__new__(decorated, value, **kwargs)

    def __init__(self, value: Value, **kwargs):
        self.value = value
        kwargs["source"] = "*"
        kwargs["read_only"] = True

        super().__init__(**kwargs)

    def to_representation(self):
        return self.value
```

The `__new__` method is necessary to pull the type from the passed value (or set it to a nullable string, if the value is None). `extend_schema_field` comes from `drf_spectacular.utils`. Not sure whether it's necessary, the method fields work just fine, but if you want something a little more "magical", there you go :slightly_smiling_face:

sarayourfriend (approved):
LGTM!

obulat:
> but if you want something a little more "magical", there you go 🙂

Thank you for the implementation and sharing the info on `nullability` 😄  I think we can go with what we have here, and change if needed later, if this starts to cause problems

PR #3046, opened 2023-09-20
Add tracemalloc utility view to API and improve k6 scripts for memory leak debugging

sarayourfriend:
## Fixes

Related to #3028 

## Description

Two things in this PR help to reproduce and debug the memory leak:
1. Add a new `/_trace/` view to the API to make it easy to see tracemalloc changes over time.
2. Update the k6 script to use random tags from actual results. This ensures we don't make requests that do not have any results, any tag returned will definitely have at least one result.

With the k6 script I'm able to consistently reproduce the leak locally. The API's memory usage climbs seemingly indefinitely now. To do so:

1. Run `just down -v && just api/init`. Deleting the volumes ensures you do not have any dead links cached in redis so that dead link filtering is exercised to the fullest extent.
1. `cd` into `utilities/load_testing` and run `just k6-local`
1. In another terminal, run `docker stats` and observe the memory usage of `openverse_web_1`

You can run multiple `just k6-local` scripts in different terminals to increase the load but I haven't noticed a big difference when doing this, probably because it's a single worker locally. Speaking of which, you can increase the number of workers to 4 by removing `-w 1` in `docker-compose.yml` for the `web` service's `command`. I've been able to observe jumps as big as 20MiB within 2 seconds when I do this locally.

If you enable the trace view (set `ENABLE_TRACE_VIEW=1` in you `api/.env`), you can watch `docker stats` for when you see a jump in memory usage and then make a request to `/_trace/` on your local API. With multiple workers this isn't reliable, so reduce the workers back to 1 if you increased them before (or tweak the implementation if you have ideas for how to make it work well).

## Testing Instructions

See above.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
It does seem to kind of stabilise around 982MiB, but there may be further improvements to the script to make it more similar to production behaviour. It'd be good to add things like searching for heaps of sources the way the Gutenberg integration does, for example.

dhruvkb (approved):
The trace view seems valuable enough to me to make permanent (take out of the `if`), move to its own file (instead of the URL config) and keep even after this project (enabled conditionally with the env var).

AetherUnbound (approved):
LGTM and I've confirmed the trace route doesn't work when the setting is enabled, but does work as expected when enabled! I'm not quite sure how to read the tracemalloc output, maybe we could implement something like the [pretty top](https://docs.python.org/3/library/tracemalloc.html#pretty-top) example from the docs?

utilities/load_testing/k6/utils.js
```diff
@@ -5,12 +5,17 @@ import { randomItem } from "https://jslib.k6.io/k6-utils/1.2.0/index.js"
...
+  if (discoveredUnconsumedTags.size) {
+    const tag = randomItem(Array.from(discoveredUnconsumedTags))
+    discoveredUnconsumedTags.delete(tag)
+    return tag
+  }
+  return randomItem(["dog", "run", "cat", "honey", "happy"])
+}
 
 export const getProvider = (media_type) => {
   let url = `${API_URL}${media_type}/stats`
```
obulat, line 21: How does this work for you? `media_type` is the singular `image`, and `image/stats` fails, it needs to be `images/stats` (plural).

utilities/load_testing/k6/main.js
```diff
@@ -1,6 +1,6 @@
 import { group } from "k6"
 import { searchBy } from "./search.js"
-import { getProvider, getRandomWord } from "./utils.js"
+import { getProvider } from "./utils.js"
```
obulat, line 3: Lines 40-41:
```
    provider_image_page_20: createScenario("images", "20", "searchByProvider"),
    provider_image_page_500: createScenario("images", "500", "searchByProvider"),
```

sarayourfriend:
Drafting and set priority to low. It'd be nice to get the load testing script working reliably, but this quick job I've done does only a very small amount of what would actually make these scripts more useful today. Because we think we have a good handle on what caused the memory leak (#3047), this isn't as high priority anymore.

Issue #2502, opened 2023-06-29
API thumbnails response time alarms


sarayourfriend:
## Description

Project thread: https://github.com/WordPress/openverse/issues/2344
Implementation plan: https://docs.openverse.org/projects/proposals/monitoring/20230606_implementation_plan_ecs_alarms.html

Create alarms in `next/modules/monitoring/production-api-thumbnails` for the following:

- Average response time anomaly
- Average response time over threshold
- p99 response time anomaly
- p99 response time over threshold

## Additional context

This issue will remain open until the alarms are stabilised.

Blocked by the baseline configuration changes in https://github.com/WordPress/openverse/issues/2499

AetherUnbound:
@stacimc I'm going to reopen this issue until we've stabilized the alarm, per @obulat's similar note here: https://github.com/WordPress/openverse/issues/2505#issuecomment-1725017023

PR #3053, opened 2023-09-21
Add thumbnail repsonse time runbooks

stacimc:
## Fixes

Related to #2502 by @sarayourfriend
Related to https://github.com/WordPress/openverse-infrastructure/pull/619 

## Description

Adds runbooks for thumbnail response time alarms.

Question for reviewers: the content for all of these is identical. I could make a single runbook and update the links in the alarms. The reason I didn't is that it seems _possible_ (although perhaps unlikely) that false positives could occur separately for each of them 🤔 

## Testing Instructions

Make sure they look alright in the preview!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
LGTM. Regarding sharing them: I think we could share the _anomaly_ ones, maybe, but I do think it's good to have separate instructions for debugging average vs p99. For example, for average response time, you probably don't need to do anything very special to find "the slow requests", because overall things are slow. For p99, however, as long as average is still okay, it is revealing some kind of edge case where we'd need to query the nginx logs for target response time. I don't know whether that needs to be in the runbook, because it is a different in principle that applies to _all_ average and p99 response time debugging, but it is a notable difference between the two. But yeah, not sure whether that's a difference in the _runbook_ or a difference in some other resource.

The anomaly and threshold alarms also have a difference, I think. Like the threshold alarm runbooks could probably have an instruction to say "check whether it is a one-off spike or persistent" and guide severity along those lines. If we spike to 10 seconds average response time, that's not good, but it also doesn't mean we need to drop everything to debug it if things have gone back down.

That's also just a general principle that could apply to almost every alarm, so I don't know whether that needs to be in the runbooks or, like above, in some other document that they link to. I think this relates somewhat to the level of confidence that something _persistently bad_ is happening based on the alarm, which does also relate to the false alarms you mentioned. Ideally anomaly alarms would be higher confidence that something bad is _persistently_ happening, once we've tuned them, and threshold alarms are kind of more up in the air, could be really bad or could just be a one-off spike that we should look into, but not necessarily with urgency.

Anyway, all of that is just my thoughts on this. I don't have strong recommendations either way, but whatever we come up with, ideally we can apply it consistently between the alarms. It looks like there are some important differences but also important similarities in the principles of these alarms and what they can indicate, I just don't know what the correct or most flexible way of sharing _and_ differentiating between those is.

obulat (approved):

PR #3068, opened 2023-09-27
Transfer UUID validation inside serializer

dhruvkb:
## Description

This PR prevents the oEmbed endpoint from crashing when an invalid UUID is encountered. The UUID is now validated in the serializer, leading to DRF's `ValidationError` being raised if invalid. This is handled by DRF and converted to a 400 Bad Request reponse.

Earlier, Django would validate the UUID and raise a Django core `ValidationError` which does not map to a 400 response, and bubbles uncaught till it exits the worker.

## Testing Instructions

1. Visit the oEmbed endpoint `http://localhost:50280/v1/images/oembed/?url=` with the following `url` values:
   - valid and existing UUID like `2982c0bf-9c26-4543-a712-4d50a117ded8`
   - URL containing valid and existing UUID like `https://abc.xyz/def/2982c0bf-9c26-4543-a712-4d50a117ded8`
   - valid but non-existent UUID like `f6e28bad-860d-4cab-980f-de5aa79892d2`
   - invalid UUID like `f6e28bad-860d-4cab-980f-de5aa79`
2. In all cases you should see a DRF page with a proper error message and not the Django error page.

The PR also adds a test case checking for 400 response (before the PR the test would fail because the code would be 500).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/image_serializers.py
+++ b/api/api/serializers/image_serializers.py
@@ -1,4 +1,5 @@
 from typing import Literal
+from uuid import UUID
 
 from rest_framework import serializers
 
@@ -124,7 +125,15 @@ class OembedRequestSerializer(serializers.Serializer):
 
     @staticmethod
     def validate_url(value):
-        return add_protocol(value)
+        url = add_protocol(value)
+        if url.endswith("/"):
+            url = url[:-1]
+        identifier = url.rsplit("/", 1)[1]
+        try:
+            uuid = UUID(identifier)
+        except ValueError:
+            raise serializers.ValidationError("Could not parse identifier from URL.")
+        return uuid
 
 
 class OembedSerializer(BaseModelSerializer):
--- a/api/api/views/image_views.py
+++ b/api/api/views/image_views.py
@@ -83,10 +83,7 @@ class ImageViewSet(MediaViewSet):
 
         context = self.get_serializer_context()
 
-        url = params.validated_data["url"]
-        if url.endswith("/"):
-            url = url[:-1]
-        identifier = url.rsplit("/", 1)[1]
+        identifier = params.validated_data["url"]
         image = get_object_or_404(Image, identifier=identifier)
         if not (image.height and image.width):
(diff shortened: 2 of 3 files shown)


AetherUnbound (approved):
LGTM and changes work locally, thanks for the quick turnaround on this!

stacimc (approved):
LGTM!

api/api/serializers/image_serializers.py
```diff
@@ -124,7 +125,15 @@ class OembedRequestSerializer(serializers.Serializer):
 
     @staticmethod
     def validate_url(value):
-        return add_protocol(value)
+        url = add_protocol(value)
+        if url.endswith("/"):
+            url = url[:-1]
+        identifier = url.rsplit("/", 1)[1]
+        try:
+            uuid = UUID(identifier)
```
krysal, line 133: I find it odd that the URL validator is returning the image identifier.

krysal (changes requested):

api/api/serializers/image_serializers.py
AetherUnbound: This seems to just be the way this parameter is set up - ultimately we want an identifier out of it, even if the parameter itself is the `url` field.

krysal: It's confusing since it doesn't even check if `value` is a valid URL or not. I'd like us to go with this [alternative solution](https://github.com/WordPress/openverse/pull/3069/).

stacimc: It is for sure confusing that it doesn't check if `value` is a valid URL, but http://localhost:50280/v1/images/oembed/?url=2982c0bf-9c26-4543-a712-4d50a117ded8 for example works on `main` and on this branch but not on the proposed alternative. Given this is critical, do we know we want to make that change in behavior as part of the fix? FWIW I agree it makes more sense in general, but worth asking.

PR #3107, opened 2023-10-01
Remove monetization meta tag

sarayourfriend:
## Fixes

I'm not sure if this is a valid PR. Described below. There is no issue, just something I noticed that was weird.

## Description

I noticed this on a single results page when looking through the meta tags and realised I had no idea what it does. Reading more about the monetization meta tag, I think there are two problems with the current one (info here https://webmonetization.org/):

1. The uphold account is probably not controlled by WP Foundation. I don't know where it leads, but I've never heard us mention this _literally ever_, and the earliest code reference I can find dates back to the CC Search days: https://github.com/cc-archive/cccatalog-frontend/issues/1014
2. The same monetization tag shows on single result pages, which seems to indicate that the same uphold account gets monetization benefits from single results pages. This was determined in https://github.com/cc-archive/cccatalog-frontend/issues/1189, but I'd like to understand more about why this is okay, and whether the intention was to eventually allow creators to add their own monetization tokens somehow?

For now though, I think the correct thing to do is remove it and, if we want to (if we understand this feature sufficiently and its costs/benefits) work towards re-enabling it with clear and open documentation about the intentions and rationale behind it.

## Testing Instructions

Run the frontend locally (`just p frontend dev`) and confirm the meta tag is gone. Really this is just testing that Nuxt works the way it says it does. As long as the references are gone from the code then the meta tag should be gone.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/constants/meta.ts
+++ b/frontend/src/constants/meta.ts
@@ -16,11 +16,6 @@ export const meta = [
   // search, content and home pages.
   { hid: "robots", name: "robots", content: "noindex" },
   {
-    vmid: "monetization",
-    name: "monetization",
-    content: "$ilp.uphold.com/edR8erBDbRyq",
-  },
-  {
     hid: "theme-color",
     name: "theme-color",
     content: "#ffffff",


sarayourfriend:
I'll add that if users have enabled passive monetization payments, then they've been sending money to a wallet that isn't actually contributing to Openverse at all, which is at best tragic (we're missing the benefits of those payments, if there are any), at worst some kind of liability issue due to the money going somewhere that isn't actually benefitting the project, which is ostensibly the idea behind web monetization "standards".

dhruvkb (approved):
Removing it makes sense. I didn't even know we had this and uphold.com seems to be some crypto platform. Additionally, they've also deprecated ILP payments so I would guess this meta tag adds nothing anyway.

Screenshot 2023-10-02 at 9 38 16 AM

obulat (approved):

krysal:
The wallet belongs to CC, so I'd say it is not super tragic, but I agree the time is due to remove it. I had totally forgotten about it. Thanks!

Issue #2858, opened 2023-08-22
Update the Nuxt store to allow for searches by source/creator/tag


obulat:
## Problem

The Nuxt app store and URL construction functions need to be updated to allow for getting the tag/source/creator collections.

## Description

The `searchBy` filter in the search store should be updated with the new values..

The API request URL is constructed from the search store state in the [`prepare-search-query-params` method](https://github.com/WordPress/openverse/blob/b4b46b903731870015c475d2c08eebef7ec6b25b/frontend/src/utils/prepare-search-query-params.ts#L22-L25).
This method should be updated to use the `searchBy` filter value to construct the API request path as described in the "API Changes" section.

## Additional context

Additional search views Implementation plan: https://docs.openverse.org/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.html

PR #3136, opened 2023-10-04
Simplify Nuxt query parameter computation

obulat:
## Fixes

Related to #2858 by @obulat

## Description

This PR simplifies one of the most complicated parts of the Nuxt application: the building of API query string from the query parameters, and parsing of the query string into the filter parameters. This will make adding the store changes for Additional search views project much easier to do.

`searchBy` filter kind was removed because it does not allow for the flexibility required by the Additional search views project. It will be replaced by a separate parameter on the search store state in the future PR.

`prepare-search-query-params` module was completely removed because it is no longer needed. Its functions are implemented more flexibly in `computeQueryParams` function.
Some unused functions were removed. For example, we no longer parse the query string into filters, and use the `route.params` instead.

This PR also adds more concise types for the search query.

## Testing Instructions

The querying should work as expected, and CI should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (approved):
The changes look good to me.

frontend/src/stores/search.ts
```diff
@@ -62,36 +63,28 @@ export interface SearchState {
...
   mode: "frontend" | "API"
 ) {
-  // The filters object is converted to a Record<string, string> object.
-  // e.g., { licenseTypes: [{ code: "commercial", checked: true }] }
-  // => { license_type: "commercial" }
-  const query = { ...filtersToQueryData(filters, searchType) }
-
+  if (!isSearchTypeSupported(searchType)) {
+    return { q: searchTerm.trim() }
+  }
```
sarayourfriend, line 80: What are the circumstances in which this would be the case? I'm not fully remembering how the supported/non-support search type thing works or why the code needs to handle both cases when only one is actually used (supported).

frontend/src/stores/search.ts
```diff
@@ -62,36 +63,28 @@ export interface SearchState {
...
-  const param_names = Object.keys(query).filter(
-    (key): key is Exclude<keyof ApiQueryParams, "q"> =>
-      !["q", INCLUDE_SENSITIVE_QUERY_PARAM].includes(key)
-  )
-
-  for (const api_param_name of param_names) {
-    if (query[api_param_name]?.length) {
-      search_query[api_param_name] = query[api_param_name]
-    }
+  const search_query: SearchQuery = {
```
sarayourfriend, line 82: Can we change this to use camel case, while we're modifying this function anyway?

```suggestion
  const searchQuery: SearchQuery = {
```

frontend/src/stores/media/index.ts
```diff
@@ -442,9 +443,8 @@ export const useMediaStore = defineStore("media", {
       mediaType: SupportedMediaType
       shouldPersistMedia: boolean
     }) {
-      const queryParams = prepareSearchQueryParams({
-        ...useSearchStore().searchQueryParams,
-      })
+      const queryParams = useSearchStore()
+        .searchQueryParams as PaginatedSearchQuery
```
sarayourfriend, line 447: Why is the case necessary here? Could we move it into the search store's `searchQueryParams` getter? Is there some other way to refine this type that wouldn't require a cast?

frontend/src/stores/search.ts
```diff
@@ -223,35 +203,32 @@ export const useSearchStore = defineStore("search", {
...
+      if (queryParams === undefined) {
+        queryParams =
+          searchType === undefined
+            ? this.frontendSearchUrlParams
+            : computeQueryParams(
+                searchType,
+                this.filters,
+                this.searchTerm,
+                "frontend"
+              )
```
sarayourfriend, line 226: What's the difference between using `frontendSearchUrlParams` and calling `computeQueryParams` with the frontend strategy? `frontendSearchUrlParams`'s implementation looks the same as the else branch of this ternary. Is the difference in passing `searchType` directly vs using `this.searchType` as `frontendSearchUrlParams` does? In what circumstances would `type` be undefined? Can those circumstances safely fallback to `ALL_MEDIA`, maybe?

If anything, a comment here would be helpful to explain why the branches are different and why an undefined search type dictates which we choose. Whatever the distinction is between these two branches seems to be very subtle and is completely lost on me.

frontend/src/stores/search.ts
```diff
@@ -508,7 +485,7 @@ export const useSearchStore = defineStore("search", {
       filterCategory: FilterCategory
     ): boolean | undefined {
       if (!["licenseTypes", "licenses"].includes(filterCategory)) {
-        return
+        return false
```
sarayourfriend, line 488: Does this change allow us to remove `undefined` from the return type of this function on line 486?

sarayourfriend (approved):
Tests fine for me locally. A handful of non-blocking questions. One is an important request for an explanatory comment, but I don't want to further delay this PR as I was late to review it.

frontend/src/stores/search.ts
obulat: If you turn on `additional_search_types` feature flag and look into the content switcher in the search results page header, the links to the types that are "Coming soon" will have the `q=cat` (or whatever) parameter set instead of just linking to `/search/video` (which will probably just redirect to the homepage since we don't have a generic media type page without a search term).
Screenshot 2023-10-23 at 1 29 44 PM

frontend/src/stores/media/index.ts
obulat: `search` store can compute all of the parameters except for the `page`, which is saved in the `media` store.
I was trying not adding the `page` parameter to the `queryParams` when sending the first search request, that's why I wrote this code like this.
I refactored it now to remove the type cast.

frontend/src/stores/search.ts
obulat: I simplified this code. The `type` is passed here when creating a search page link (in the content switcher or in the  content link on the All content search page) for a type that's different from the current search type.

PR #3151, opened 2023-10-05
Simplify the related search

### LINKED ISSUE #3149: Improve related endpoint performance
## Problem

The related endpoint is anywhere from 3 to 5 times slower than a regular search. Even in best case scenarios (outside the search response time peaks), when regular search is <200ms, related will be anywhere from 600ms to 1 second. The related endpoint uses the black-box and potentially non-performant MoreLikeThis query. 99% of the time it just returns works from the same creator. They are also always filtered to exclude mature results (but does not query the filtered index, so doesn't actually exclude sensitive terms). 
Something about the low-performance of the related endpoint might be causing memory pressure on Elasticsearch, causing it to need to GC more heap than otherwise necessary. 

### Must do change
Switch to use the filtered index. The filtered indexes are slightly smaller. The performance benefit here is probably negligible but (a) it's a bug that it doesn't query the filtered index and (b) it's a trivial change. Sensitivity annotations (SearchContext.build) can also be skipped if we do this because none of the documents will be mature or have sensitive text due to the search excluding them. That removes one additional Elasticsearch query from the related endpoint that looks for sensitive term documents.

### Additional changes

Removing MoreLikeThis and using a simpler heuristic for finding "related" documents would be preferable and would probably have higher impact that just removing "creator" from more like this, but it is a more complex change. 

#### Use a simpler heuristic for "related" documents than `MoreLikeThis`
Stop using MoreLikeThis. Instead, pull the title and tags from Postgres for the parent result, use multi_match query on title, description, and tags, mirror our regular approach to search. Limit to 10 tags, maybe sort them by length so to prioritise simpler terms.

The key to this change is that it makes related more like a regular search request, and therefore should have far more predictable performance compared to MoreLikeThis, which is a black box and only used in this specific instance.

#### Remove creator from MoreLikeThis.

 The creator matching here isn't really the most useful and often leads to baffling results from a user perspective. Prioritising tags might be the best option here. title is also probably fine. Removing a field from this search could help it go faster.

obulat:
## Fixes

Fixes #3149 by @obulat

## Description

This PR:
- uses the filtered index for `related` search
- removes the search context building. Search context allows us to return sensitivities for the media items, but since we only use the filtered index, all sensitivities are empty anyways.
~~removes the `mature`: False filter since we use the filtered index~~ (removing this filter would show the user-reported mature items, so I put it back)
- removes `creator` from the fields that are used in the `MoreLikeThis` query

## Testing Instructions

Check that the related endpoints (the `related` in image and audio results) work as expected.
To see this work visually, you can run the frontend Nuxt app using the local API:
1. Set the API_URL variable in the `frontend/.env`: `API_URL=http://0.0.0.0:50280/`
2. Run the API using `just up`.
3. Run Nuxt using `just frontend/run dev`.
Now, you can search for something, open the single result page and see that it has the related items. 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -283,7 +283,8 @@ def _exclude_filtered(s: Search):
             key=filter_cache_key, timeout=FILTER_CACHE_TIMEOUT, value=filtered_providers
         )
     to_exclude = [f["provider_identifier"] for f in filtered_providers]
-    s = s.exclude("terms", provider=to_exclude)
+    if to_exclude:
+        s = s.exclude("terms", provider=to_exclude)
     return s
 
 
@@ -495,7 +496,7 @@ def search(
     return results, page_count, result_count, search_context.asdict()
 
 
-def related_media(uuid, index, filter_dead):
+def related_media(uuid: str, index: str, filter_dead: bool) -> tuple[list[Hit], int]:
     """Given a UUID, find related search results."""
 
     search_client = Search(index=index)
@@ -508,14 +509,12 @@ def related_media(uuid, index, filter_dead):
     s = search_client
     s = s.query(
         MoreLikeThis(
-            fields=["tags.name", "title", "creator"],
+            fields=["tags.name", "title"],
             like={"_index": index, "_id": _id},
             min_term_freq=1,
             max_query_terms=50,
         )
     )
-    # Never show mature content in recommendations.
-    s = s.exclude("term", mature=True)
     s = _exclude_filtered(s)
     page_size = 10
     page = 1
@@ -526,12 +525,7 @@ def related_media(uuid, index, filter_dead):
 
     result_count, _ = _get_result_and_page_count(response, results, page_size, page)
 
-    if not results:
-        results = []
-
-    result_ids = [result.identifier for result in results]
-    search_context = SearchContext.build(result_ids, index)
-    return results, result_count, search_context.asdict()
+    return results or [], result_count
 
 
 def get_sources(index):
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -515,6 +515,9 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> tuple[list[Hit], 
             max_query_terms=50,
         )
     )
+    # Prevent the items that users set as `mature` from showing up in
+    # recommendations.
+    s = s.exclude("term", mature=True)
     s = _exclude_filtered(s)
     page_size = 10
     page = 1
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -497,23 +497,43 @@ def search(
 
 
 def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
-    """Given a UUID, find related search results based on title and tags."""
+    """
+    Given a UUID, finds 10 related search results based on title and tags.
+
+    Uses Match query for title or SimpleQueryString for tags.
+    If the item has no title and no tags, returns items by the same creator.
+    If the item has no title, no tags or no creator, returns empty list.
+
+    :param uuid: The UUID of the item to find related results for.
+    :param index: The Elasticsearch index to search (e.g. 'image')
+    :param filter_dead: Whether dead links should be removed.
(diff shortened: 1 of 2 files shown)


sarayourfriend:
> removes the mature: False filter since we use the filtered index

Quick note that the filtered index only filters sensitive text, so user reported sensitive media will still be in there (just with `mature=true`). For full mature filtering (sensitive text _and_ user reported) we still need to filter `mature=false`. (These names are confusing!)

dhruvkb (approved):
LGTM! This is a solid improvement to the `related/` endpoint.

AetherUnbound (approved):
LGTM, I see that #3156 builds off of this but in a way that makes some of these changes irrelevant, is that correct?

obulat:
> LGTM, I see that #3156 builds off of this but in a way that makes some of these changes irrelevant, is that correct?

Yes. I wanted to implement the simple "quick win" approach on Friday, thinking that the "more correct" approach would take a long time and be more difficult. Then, next day I realized that #3156 would actually be doable, I opened it, and based it on this PR (because here we remove the `creator` from the query and use the filtered index).

api/api/controllers/search_controller.py
```diff
@@ -495,43 +496,42 @@ def search(
...
 
-def related_media(uuid, index, filter_dead):
-    """Given a UUID, find related search results."""
+def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
+    """Given a UUID, find related search results based on title and tags."""
 
-    search_client = Search(index=index)
+    # Search the default index for the item itself as it might be sensitive.
+    item_search = Search(index=index)
+    item_hit = item_search.query("match", identifier=uuid).execute().hits[0]
```
sarayourfriend, line 504: This should use `identifier__keyword` and `term` to avoid stemming on the `text` field that `identifier` is currently. This issue came up during the development of the `SearchContext::build` method where I noticed I wasn't getting the exact list of identifiers back :scream:

api/api/controllers/search_controller.py
```diff
@@ -495,43 +496,42 @@ def search(
...
-        )
-    )
-    # Never show mature content in recommendations.
-    s = s.exclude("term", mature=True)
+    # Match related using tags, if the item has any.
+    if tags := getattr(item_hit, "tags", None):
+        # Only use the first 10 tags
+        tags = ",".join([tag.name for tag in tags[:10]])
+        tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
+        related_query |= tags_query
```
sarayourfriend, line 516: I wonder if a multi_match wouldn't be more straightforward than constructing a simple query string query? If we do use simple query string, though, `,` isn't an operator available, as far as I can tell: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#simple-query-string-syntax

A space should be sufficient to use the default `OR` operator. In this case even `query_string` could work fine.

api/api/controllers/search_controller.py
```diff
@@ -495,43 +496,42 @@ def search(
...
+    item_search = Search(index=index)
+    item_hit = item_search.query("match", identifier=uuid).execute().hits[0]
 
-    # Convert UUID to sequential ID.
-    item = search_client
-    item = item.query("match", identifier=uuid)
-    _id = item.execute().hits[0].id
+    # Match related using title.
+    title = item_hit.title
+    title_query = SimpleQueryString(query=title, fields=["title"])
```
sarayourfriend, line 508: I wonder if `match_phrase` is more appropriate for title? https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query-phrase.html

sarayourfriend (commented):
Just wanted to share some initial comments but this is very cool, I'm optimistic it could lead to a genuine improvement in the related results. Do we have an event on related result clicking? I'd be keen to compare before/after if this change goes through to see whether it improves things, especially if we deploy the creator change _first_ and then this afterwards. If there are progressively positive changes, we'd at least have a gut check about what users are looking for in the related query (that being a point besides potentially improving the performance of this route).

PR #3153, opened 2023-10-05
Wait for Plausible active state

### LINKED ISSUE #2489: `just frontend/init` fails with cryptic error if plausible service isn't ready for requests
## Description

```
$ just frontend/up
...
$ just frontend/init
env COMPOSE_PROFILES="frontend" just ../up 
env COMPOSE_PROFILES="frontend" docker-compose -f docker-compose.yml up -d
[+] Running 3/0
 ✔ Container openverse-plausible_db-1  Running                                                      0.0s 
 ✔ Container openverse-plausible_ch-1  Running                                                      0.0s 
 ✔ Container openverse-plausible-1     Running                                                      0.0s 

================================================================================
                                 Service Ports                                  
================================================================================
plausible (plausible/analytics):
-  http://0.0.0.0:50288 (→ 8000)
================================================================================
echo "🚧 TODO"
🚧 TODO
cd .. && ./setup_plausible.sh
--rpc-eval : RPC failed with reason :nodedown
error: Recipe `init` failed on line 66 with exit code 1
```

## Reproduction

1.  `just down`
2.  `just frontend/up` and right after `just frontend/init`
3. See the error above.

## Fix

Add a "wait for" recipe that waits for Plausible to be available before running `setup_plausible.sh` similar to the `docker/es/wait-for-index` recipe.

ngken0995:
## Fixes

Fixes #2489  by @sarayourfriend 

## Description

Implement `wait` in justfile. `Wait` will have host as localhost:50288 because it points to plausible localhost. There is a loop going through to check the activity of plausible. We can check the health of plausible API(https://plausible.io/docs/stats-api#breakdown-custom-event-by-custom-props).

## Testing Instructions

- `just down -v`
- `just frontend/up`
- `just frontend/init`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/justfile
```diff
@@ -59,7 +59,7 @@ up *flags:
 
 # Wait for all profile services to be up
 wait-up: up
-    echo "🚧 TODO"
+    just wait
```
AetherUnbound, line 62: I think it'd be fine to move the contents of the `wait` recipe here and remove the `@wait` recipe below entirely!

AetherUnbound (approved):
This works great! Added one note, otherwise this is good to merge IMO 🙂

frontend/justfile
```diff
@@ -71,3 +71,19 @@ run *args:
...
 types:
     cd .. && pnpm exec vue-tsc -p frontend --noEmit
+
+##########
+# Health #
+##########
+
+# Check the health of the service
+@health host="localhost:50288":
+    -curl -s -o /dev/null -w '%{http_code}' 'http://{{ host }}/api/health'
```
sarayourfriend, line 81: Can you change the name of this to `health-plausible` so that it is clear _which_ service it is checking (specifically, that it isn't checking for the "frontend" service, which doesn't exist).

frontend/justfile
```diff
@@ -71,3 +71,19 @@ run *args:
...
+
+# Wait for the service to be healthy
+@wait host="localhost:50288":
+    # The just command on the second line is executed in the context of the
+    # parent directory and so must be prefixed with `api/`.
+    just ../_loop \
+    '"$(just frontend/health {{ host }})" != "200"' \
+    "Waiting for the frontend to be healthy..."
```
dhruvkb, line 89: Along the same lines as @sarayourfriend's feedback, this should be updated as follows.

```suggestion
    "Waiting for Plausible to be healthy..."
```

frontend/justfile
```diff
@@ -71,3 +71,19 @@ run *args:
...
+##########
+
+# Check the health of the service
+@health host="localhost:50288":
+    -curl -s -o /dev/null -w '%{http_code}' 'http://{{ host }}/api/health'
+
+# Wait for the service to be healthy
+@wait host="localhost:50288":
+    # The just command on the second line is executed in the context of the
+    # parent directory and so must be prefixed with `api/`.
```
dhruvkb, line 86:
```suggestion
    # parent directory and so must be prefixed with `frontend/`.
```

frontend/justfile
dhruvkb: I think this is implemented this way for consistency with the API justfile which has both `@wait` and `wait-up`. I can get behind this, as this can make working in a Dockerised context locally easier in the future.

frontend/justfile
```diff
@@ -71,3 +71,19 @@ run *args:
...
+    just ../_loop \
+    '"$(just frontend/health {{ host }})" != "200"' \
```
dhruvkb, line 88: I think we should also check that the response contains 3 `"ok"`s. 
```json
{
  "clickhouse": "ok",
  "postgres": "ok",
  "sites_cache": "ok"
}
```

I haven't seen what the status code and response is when Plausible is not healthy so there is a possibility that it returns some "not ok" but with a HTTP 200 response.

dhruvkb (commented):
Overall the changes look good to me, and you're going in the right direction. Thanks for the contribution @ngken0995! But I some suggestions.

frontend/justfile
ngken0995: @dhruvkb I found a two ways the parse the JSON object. First option, we can use a `jq` which is a JSON parsing and transformation tool for linux and determine if the response contains 3 `"ok"`s. Second options, we can create python to parse through the JSON object and determine  if the response contains 3 `"ok"`s. Let me know if there is better solution. What is the best option to take?

dhruvkb: I don't want to add a dependency on `jq`. Using Python might be fine 👍.

ngken0995:
@dhruvkb and @AetherUnbound I created a python file to check the JSON object response for the three `ok`s. I believe it can be improved because I had to add a try-except. Please let me know what I can do to making a sustainable code to check the health of plausible,

frontend/justfile
sarayourfriend: Agreed that not adding jq as a dependency is good. We've avoided it in other instances as well, opting instead for Python then too.

sarayourfriend: I think we don't need something as complex as Python or jq or even JSON parsing. Looking at the [plausible code for the healthcheck endpoint](https://github.com/plausible/analytics/blob/master/lib/plausible_web/controllers/api/external_controller.ex#L50-L80), we can just as easily do this and it would catch the unhealthy cases:

```
curl -s 'http://localhost:50288/api/health' --fail | grep -vq 'error'
```

`-vq` inverts the match and suppresses grep output (stops it from printing matched lines, in this case all the `ok` that don't match `error`, inverted) so we just get the exit code. `--fail` in curl causes it to fail if the response code is 500, bypassing grep entirely.

dhruvkb (commented):
The script but we can use it more effectively by returning a non-zero exit code when the status is not "ok".

frontend/plausible_health_check.py
```diff
@@ -0,0 +1,22 @@
+import sys
+
+import requests
```
sarayourfriend, line 3: `requests` isn't standard library and not everyone will have it installed in their global python installation (I don't, for example). `urlopen` is a better choice here, but less convenient, if we stick with a python script instead of a shell-based approach.

+1 to Dhruv's comment that try/except is just fine here, totally appropriate usage.

frontend/justfile
ngken0995: @sarayourfriend your solves only require curl and grep which I would prefer.

ngken0995: @dhruvkb Please take a look at the suggested resolution to use grep and invert matching.

sarayourfriend: Sorry for not responding here. You'd need to use the exit code for my original suggestion, which you can get with `$?`. Your approach looks more conventional though, I always have to look up what `$?` is :sweat_smile:

ngken0995: I tried `'http://localhost:50288/api/health' --fail | grep -vq 'error' | echo $?` and getting this error:
 ```
--rpc-eval : RPC failed with reason :nodedown
error: Recipe `init` failed on line 66 with exit code 1
```
Also, I'm not sure how I can pass a exit code into a command substitution.

It seems like the `curl -s 'http://localhost:50288/api/health' --fail | grep -v -c 'error'` is working as expected. Let me know if there could be any edge case. 

What are the other API status result from `clickhouse`, `postgres` and `sites_cache` beside "ok" and "error"?

File: frontend/justfile [hunk 7]
```diff
@@ -71,3 +71,19 @@ run *args:
...
 types:
     cd .. && pnpm exec vue-tsc -p frontend --noEmit
+
+##########
+# Health #
+##########
+
+# Check the health of the service
+@plausible-health host="localhost:50288":
+    -curl -s 'http://localhost:50288/api/health' --fail | grep -v -c 'error'
```
dhruvkb, line 81: Verified that this returns 0 for unhealthy and 1 for healthy.

frontend/justfile
```diff
@@ -71,3 +71,19 @@ run *args:
...
+
+# Wait for the service to be healthy
+@wait host="localhost:50288":
+    # The just command on the second line is executed in the context of the
+    # parent directory and so must be prefixed with `frontend/`.
+    just ../_loop \
+    '"$(just frontend/plausible-health {{ host }})" != "1" ' \
+    "Waiting for the plausible to be healthy..."
```
dhruvkb, line 89: Nit: grammar

```suggestion
    "Waiting for Plausible to be healthy..."
```

dhruvkb (approved):
LGTM, verified that this works. Thanks for the contribution @ngken0995! Will merge it in once CI has passed.

PR #3156, opened 2023-10-06
Use simple query for related

### LINKED ISSUE #3149: Improve related endpoint performance

obulat:
## Fixes

Fixes #3149 by @obulat

## Description

This PR changes the related endpoint: now, instead of using `MoreLikeThis` query, it gets the title and the tags of the media, and searches for title/tag matches.
We use `or` between the title and tag queries (ES's should, not must) to make sure that there's a match (even if there's no match by title or no match by tags alone). I'm not sure what will happen if both are blank. What happens now?

The related no longer checks for `sensitivities` because it filters out the sensitive and mature results.
It also now returns 10 as the result_count because it does not make sense to return the full result count: there's no way to get other pages of the results, and this piece of information is not meaningful.

On main, the `/related` endpoint makes 3 ES requests. One to get the item itself and its `id`, one to get the MoreLikeThis query, and one to get all the sensitivities.
```
[f26e9f8778024f46b6652f5e48865267] POST http://es:9200/image/_search [status:200 duration:0.198s]
[2023-10-06 05:47:55,926 - elastic_transport.transport - 335][INFO] [f26e9f8778024f46b6652f5e48865267] POST http://es:9200/image/_search [status:200 duration:0.250s]
[2023-10-06 05:47:57,051 - elastic_transport.transport - 335][INFO] [f26e9f8778024f46b6652f5e48865267] POST http://es:9200/image-filtered/_search [status:200 duration:0.015s]
```

With this change, we get 2 ES requests. One gets the item itself to extract its title and tags. The other gets the title/tag matches.

```
[2023-10-06 05:45:37,855 - elastic_transport.transport - 335][INFO] [62341dcfaff941e197f6a083250b559c] POST http://es:9200/image-filtered/_search [status:200 duration:0.040s]
[2023-10-06 05:45:38,043 - elastic_transport.transport - 335][INFO] [62341dcfaff941e197f6a083250b559c] POST http://es:9200/image-filtered/_search [status:200 duration:0.074s]
```

## Testing Instructions

Run the app and check the `related` results. 
To see this work visually, you can run the frontend Nuxt app using the local API:
1. Set the API_URL variable in the `frontend/.env`: `API_URL=http://0.0.0.0:50280/`
2. Run the API using `just up`.
3. Run Nuxt using `just frontend/run dev`.

Compare the related items for the same image locally and in prod:
http://localhost:8443/image/c8e21cd7-ef1a-4a18-a2aa-5ec6730a33b4?q=japan
https://openverse.org/image/9cebb303-3787-4ac0-a9bc-47c182bbe237?q=%E9%81%A0%E3%81%8F%E8%BF%91%E3%81%8F%E3%81%AB%E3%81%BE%E3%81%9A%E3%81%B2%E3%81%A8%E3%81%A4
I think that since we removed the `creator` from the `related` query, the items are much closely "related" now. 
_I don't know why the uuids for the same image are different in the sample data and the prod data._

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -11,7 +11,7 @@ from django.core.cache import cache
 
 from elasticsearch.exceptions import NotFoundError, RequestError
 from elasticsearch_dsl import Q, Search
-from elasticsearch_dsl.query import EMPTY_QUERY, MoreLikeThis, Query
+from elasticsearch_dsl.query import EMPTY_QUERY, Match, Query, SimpleQueryString, Term
 from elasticsearch_dsl.response import Hit, Response
 
 import api.models as models
@@ -496,39 +496,37 @@ def search(
     return results, page_count, result_count, search_context.asdict()
 
 
-def related_media(uuid: str, index: str, filter_dead: bool) -> tuple[list[Hit], int]:
-    """Given a UUID, find related search results."""
+def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
+    """Given a UUID, find related search results based on title and tags."""
 
     search_client = Search(index=index)
 
     # Convert UUID to sequential ID.
     item = search_client
-    item = item.query("match", identifier=uuid)
-    _id = item.execute().hits[0].id
+    item_query = item.query("match", identifier=uuid)
+    item_hit = item_query.execute().hits[0]
+    title = item_hit.title
+    tags = ",".join([tag.name for tag in item_hit.tags])
 
     s = search_client
-    s = s.query(
-        MoreLikeThis(
-            fields=["tags.name", "title"],
-            like={"_index": index, "_id": _id},
-            min_term_freq=1,
-            max_query_terms=50,
-        )
-    )
-    # Prevent the items that users set as `mature` from showing up in
-    # recommendations.
-    s = s.exclude("term", mature=True)
+
+    # Match the title or tags
+    title_query = SimpleQueryString(query=title, fields=["title"])
+    tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
+    related_query = title_query | tags_query
+
+    # Exclude the current item and mature content.
+    s = s.query(related_query & ~Match(identifier__keyword=uuid) & ~Term(mature=True))
+    # Exclude the dynamically disabled sources.
     s = _exclude_filtered(s)
-    page_size = 10
-    page = 1
+
+    page, page_size = 1, 10
     start, end = _get_query_slice(s, page_size, page, filter_dead)
     s = s[start:end]
+
     response = s.execute()
     results = _post_process_results(s, start, end, page_size, response, filter_dead)
-
-    result_count, _ = _get_result_and_page_count(response, results, page_size, page)
-
-    return results or [], result_count
+    return results or []
 
 
 def get_sources(index):
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -510,7 +510,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
 
     # Match related using tags, if the item has any.
     if tags := getattr(item_hit, "tags", None):
-        tags = ",".join([tag.name for tag in tags])
+        # Only use the first 10 tags
+        tags = ",".join([tag.name for tag in tags[:10]])
         tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
         related_query |= tags_query
 
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -499,16 +499,16 @@ def search(
 def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
     """Given a UUID, find related search results based on title and tags."""
 
-    search_client = Search(index=index)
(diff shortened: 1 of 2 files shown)


api/api/controllers/search_controller.py
```diff
@@ -496,39 +496,39 @@ def search(
...
+
+    # Match the title or tags
+    title_query = SimpleQueryString(query=title, fields=["title"])
+    tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
+    related_query = title_query | tags_query
+
+    # Exclude the current item and mature content.
+    # TODO: remove `__keyword` after
+    #  https://github.com/WordPress/openverse/pull/3143 is merged.
+    s = s.query(related_query & ~Match(identifier__keyword=uuid) & ~Term(mature=True))
```
dhruvkb, line 521: Is the `& ~` in `s.query()` equivalent to the previous `s.exclude()`?

obulat: I _think_ ~ is converted to `must_not` query. Here's the elasticsearch_dsl code for `invert`ing the query:

https://github.com/elastic/elasticsearch-dsl-py/blob/82418c01b4b0350636877e81e105164b8fcf289c/elasticsearch_dsl/query.py#L178-L196

So, the query is excluding identifier match (so, the item itself) and `mature=True`.
The `and` means that all of the clauses must match.

dhruvkb:
a54d4f8 handles cases where items have no tags. In these cases the `Hit` instances don't have the `tags` field at all (instead of it being set to `None` or `[]`). Now, for those instances, the related query will not have a tags component.

dhruvkb (approved):
I tried this locally and it works well. 

Unfortunately I see mostly the same results in related as I see in the original query. I'm not sure if it's because the queries are kind of similar now or because of the limited sample size in the test data. Nonetheless, it is a solid change code-wise.

AetherUnbound:
> Unfortunately I see mostly the same results in related as I see in the original query. I'm not sure if it's because the queries are kind of similar now or because of the limited sample size in the test data. Nonetheless, it is a solid change code-wise.

That's a really good point...this does make the related very similar to the original query, albeit without any of the filters that might have been specified on the original query 🤔 I guess I hadn't realized before, the related items _were_ frequently different from the other search results. It seems like this significantly improves performance, which is a good thing in the short term. But improving our related items results could be a good focus area for us in the future!

AetherUnbound (approved):
These changes look good to me and the logic makes sense! I'm surprised that there weren't any tests which needed changes 😮

obulat:
## Comparison of the `related` results

I think the main reason for not seeing a lot of difference between the related results locally is the fact that the sample image set is really small, @dhruvkb and @AetherUnbound. Here is an image that shows the improvements in the relevance of related results:

https://openverse.org/image/9cebb303-3787-4ac0-a9bc-47c182bbe237?q=%E9%81%A0%E3%81%8F%E8%BF%91%E3%81%8F%E3%81%AB%E3%81%BE%E3%81%9A%E3%81%B2%E3%81%A8%E3%81%A4

Screenshot 2023-10-07 at 2 36 15 PM

In prod, this image has the following related results:

Screenshot 2023-10-07 at 2 37 07 PM

And locally, here are the related results for the same image (which for some reason has a different UUID/Openverse identifier in the sample data):

http://localhost:8443/image/c8e21cd7-ef1a-4a18-a2aa-5ec6730a33b4?q=japan
Screenshot 2023-10-07 at 2 36 37 PM

PR #3171, opened 2023-10-09
Allow data refresh poke interval for Image.

### LINKED ISSUE #1472: Allow data refresh poke interval to be configured per media type
## Description

Presently the data refresh poke interval can only be configured by an environment variable across all data refresh DAGs:

https://github.com/WordPress/openverse-catalog/blob/3ee97b6cbe45337bec6289bd77b9a57512a82522/openverse_catalog/dags/data_refresh/data_refresh_task_factory.py#L143

It would be ideal to allow each media type to have its own poke interval. For instance, audio data refreshes could probably be every minute, whereas for the image data refresh 15 minutes makes more sense.

## Implementation

ngken0995:
## Fixes

Fixes #1472 by @AetherUnbound 

## Description

Add an `poke interval` variable to `POPULARITY_REFRESH_CONFIGS` for media_type as "image".

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_refresh/data_refresh_task_factory.py
+++ b/catalog/dags/data_refresh/data_refresh_task_factory.py
@@ -101,8 +101,6 @@ def create_data_refresh_task_group(
 
     target_alias = data_refresh.media_type  # TODO: Change when using versioned aliases
 
-    # Retrieve the correct poke amount for media_type
-
     with TaskGroup(group_id="data_refresh") as data_refresh_group:
         tasks = []
         # Wait to ensure that no other Data Refresh DAGs are running.
--- a/catalog/dags/data_refresh/data_refresh_task_factory.py
+++ b/catalog/dags/data_refresh/data_refresh_task_factory.py
@@ -60,7 +60,7 @@ from common import ingestion_server
 from common.constants import REFRESH_POKE_INTERVAL, XCOM_PULL_TEMPLATE
 from common.sensors.single_run_external_dags_sensor import SingleRunExternalDAGsSensor
 from common.sensors.utils import get_most_recent_dag_run
-from data_refresh.data_refresh_types import DataRefresh
+from data_refresh.data_refresh_types import DATA_REFRESH_CONFIGS, DataRefresh
 
 
 logger = logging.getLogger(__name__)
@@ -101,6 +101,12 @@ def create_data_refresh_task_group(
 
     target_alias = data_refresh.media_type  # TODO: Change when using versioned aliases
 
+    # Retrieve the correct poke amount for media_type
+    poke_interval = REFRESH_POKE_INTERVAL
+    for data in DATA_REFRESH_CONFIGS:
+        if data.media_type == target_alias:
+            poke_interval = data.poke_interval
+
     with TaskGroup(group_id="data_refresh") as data_refresh_group:
         tasks = []
         # Wait to ensure that no other Data Refresh DAGs are running.
@@ -108,7 +114,7 @@ def create_data_refresh_task_group(
             task_id="wait_for_data_refresh",
             external_dag_ids=external_dag_ids,
             check_existence=True,
-            poke_interval=REFRESH_POKE_INTERVAL,
+            poke_interval=poke_interval,
             mode="reschedule",
             pool=DATA_REFRESH_POOL,
         )
@@ -131,7 +137,7 @@ def create_data_refresh_task_group(
             # Wait for the whole DAG, not just a part of it
             external_task_id=None,
             check_existence=False,
-            poke_interval=REFRESH_POKE_INTERVAL,
+            poke_interval=poke_interval,
             execution_date_fn=lambda _: get_most_recent_dag_run(
                 create_filtered_index_dag_id
             ),
--- a/catalog/dags/data_refresh/data_refresh_task_factory.py
+++ b/catalog/dags/data_refresh/data_refresh_task_factory.py
@@ -57,10 +57,10 @@ from airflow.utils.task_group import TaskGroup
 from airflow.utils.trigger_rule import TriggerRule
 
 from common import ingestion_server
-from common.constants import REFRESH_POKE_INTERVAL, XCOM_PULL_TEMPLATE
+from common.constants import XCOM_PULL_TEMPLATE
 from common.sensors.single_run_external_dags_sensor import SingleRunExternalDAGsSensor
 from common.sensors.utils import get_most_recent_dag_run
-from data_refresh.data_refresh_types import DATA_REFRESH_CONFIGS, DataRefresh
+from data_refresh.data_refresh_types import DataRefresh
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


stacimc (changes requested):
I think there's some confusion caused by the `DATA_REFRESH_POKE_INTERVAL` being used in both the data refresh and popularity refresh dags. The popularity refresh DAGs are working as expected and already allow configuring a different poke interval per media type (you can see that audio pokes every 1 minute while image pokes every 30).

Rather than changing the popularity refresh DAGs, I think we need to look at the _data refresh_ DAGs [(here)](https://github.com/WordPress/openverse/blob/main/catalog/dags/data_refresh/data_refresh_task_factory.py). In their case, we use the same poke interval regardless of media type.

One way of doing this would be to update the [data_refresh_types](https://github.com/WordPress/openverse/blob/main/catalog/dags/data_refresh/data_refresh_types.py) to have a configuration for `poke_interval`. It should mirror the way it is implemented in `popularity_refresh_types`. Let me know if that makes sense!

Totally optionally, since it _is_ confusing, we could also rename the environment variable to `REFRESH_POKE_INTERVAL` :)

ngken0995:
@stacimc Thank you for clarifying my confusion with `popularity_refresh_types` and `data_refresh_type`. I have used a similar approach with `popularity_refresh_types` to retrieve the `poke_interval`. I used `REFRESH_POKE_INTERVAL` as a default value.

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -101,14 +101,20 @@ def create_data_refresh_task_group(
 
     target_alias = data_refresh.media_type  # TODO: Change when using versioned aliases
 
+    # Retrieve the correct poke amount for media_type
+    poke_interval = REFRESH_POKE_INTERVAL
```
stacimc, line 105: We receive the `DataRefresh` config object in this factory method as `data_refresh`, so we can actually just access the poke interval directly as `data_refresh.poke_interval`, instead of iterating through the configs.

stacimc (changes requested):
This is looking great to me. Just one comment about how we can simplify things.

_________________
This is doing exactly what the issue asked, so apologies for the back-and-forth here but now I think about it the `poke_interval` in a data refresh is used in two different ways:
* Waiting for the _other_ data refresh to be complete. In this case, since `image` takes a long time and `audio` completes quickly, the poke interval should be **longer** for `audio` (to stop it from polling super frequently for a long-running operation) and **shorter** for `image`
* Waiting for its own `filtered_index_creation` step to complete. For this, it's just the opposite -- the poke interval should be longer for `image`, since filtered image index creation is a long running operation, and shorter for `audio`.

So ideally if we want to tweak these poke intervals, I think we ought to also allow separate poke intervals for these two tasks within the data refresh. That might look like adding separate `data_refresh_poke_interval` and `filtered_index_poke_interval` options to the `DataRefresh` type. Then some reasonable values for those poke intervals might be:
* `audio_data_refresh.data_refresh_poke_interval`: This is waiting on an operation that takes ~3 days, so I would keep the **30 minutes**
* `audio_data_refresh.filtered_index_poke_interval: This is waiting on an operation that takes ~1 hr, so I would poke every **1 minute**
* `image_data_refresh.data_refresh_poke_interval` = This is waiting on an operation that takes ~1 hr, so I would poke every **1 minute**
* `image_data_refresh.filtered_index_poke_interval` = This is waiting on an operation that takes ~1 day, so I would poke every **30 minutes**, although 15 is probably fine.

Pinging @AetherUnbound as the author of the issue to make sure that fits in with your intentions. We could also do this part in a separate PR if you want, particularly as the image data refresh is disabled anyway :)

catalog/dags/data_refresh/data_refresh_task_factory.py
ngken0995: @stacimc The explanation is incredible. I have added the `data_refresh_poke_interval` and `filtered_index_poke_interval` options to the `DataRefresh` type.

catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -41,6 +44,9 @@ class DataRefresh:
     index_readiness_timeout:           timedelta expressing amount of time it may take
                                        to await a healthy ES index after reindexing
     doc_md:                            str used for the DAG's documentation markdown
+    poke_interval:                     int number of seconds to wait between
```
stacimc, line 47: We should update the docstring here with the new options.

catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -68,6 +76,16 @@ def __post_init__(self):
         refresh_matview_timeout=timedelta(hours=72),
         create_pop_constants_view_timeout=timedelta(hours=24),
         create_materialized_view_timeout=timedelta(hours=72),
+        data_refresh_poke_interval=int(os.getenv("DATA_REFRESH_POKE_INTERVAL", 60)),
+        filtered_index_poke_interval=int(
+            os.getenv("FILTERED_INDEX_POKE_INTERVAL", 60 * 30)
```
stacimc, line 81: Since this is referencing a new environment variable, would you mind adding it to the catalog's `env.template` [here](https://github.com/WordPress/openverse/blob/main/catalog/env.template#L103), similar to the `DATA_REFRESH_POKE_INTERVAL`? Alternatively I actually think it's fine to reuse the `DATA_REFRESH_POKE_INTERVAL` here as well, since we basically just use that to make all the poke intervals very quick in local testing.

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -101,14 +101,16 @@ def create_data_refresh_task_group(
 
     target_alias = data_refresh.media_type  # TODO: Change when using versioned aliases
 
+    # Retrieve the correct poke amount for media_type
```
stacimc, line 104:
```suggestion
```

stacimc (commented):
Thanks @ngken0995, this is great! I see this PR was put back in draft -- I left some comments about docstrings but this otherwise looks good to me :)

catalog/dags/data_refresh/data_refresh_task_factory.py
stacimc: Great :) I got the timings of the operations we're waiting on by looking at production runs in Airflow, which does require prod access.

catalog/dags/data_refresh/data_refresh_types.py
ngken0995: I will reuse `DATA_REFRESH_POKE_INTERVAL`.

stacimc (approved):
Fantastic 🚀

krysal (approved):
Thanks for your help, @ngken0995!

PR #3175, opened 2023-10-10
Fix MRO to return 405 for unsupported methods

### LINKED ISSUE #2374: Prevent failing at /v1/auth_tokens/token/ due to `MethodNotAllowed`
## Description

The token endpoint fails with `500 Internal Server Error` when the method is different to POST. It should return a 405 error instead.

## Reproduction

1. Spin up the API `just api/up`
2. Try a request with any method different to POST, for example:
```sh
curl --request GET --url http://localhost:50280/v1/auth_tokens/token/
```
3. See error.

## Additional context

[Sentry context](https://openverse.sentry.io/share/issue/1f39e7a8ddb24891b449c25e009da88f/).

dhruvkb:
## Fixes

Fixes #2374 by @krysal

## Description

This PR fixes the method resolution order of the token `APIView` so that the exception handler will be invoked and a 405 Method Not Allowed response will be sent. When the MRO is incorrect the exception is not handled and results in a 500 Internal Server Error.

## Testing Instructions

1. Make a GET request to the `/v1/auth_tokens/token/` endpoint.
2. Receive a 405 response.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/oauth2_views.py
+++ b/api/api/views/oauth2_views.py
@@ -1,3 +1,4 @@
+import json
 import logging as log
 import secrets
 import smtplib
@@ -138,7 +139,7 @@ class VerifyEmail(APIView):
 @extend_schema(tags=["auth"])
 class TokenView(APIView, BaseTokenView):
     @token
-    def post(self, *args, **kwargs):
+    def post(self, request, *args, **kwargs):
         """
         Get an access token using client credentials.
 
@@ -154,7 +155,9 @@ class TokenView(APIView, BaseTokenView):
         endpoint.
         """
 
-        return super().post(*args, **kwargs)
+        res = super().post(request._request)
+        data = json.loads(res.content)
+        return Response(data, status=res.status_code)
 
 
 @extend_schema(tags=["auth"])
--- a/api/api/views/oauth2_views.py
+++ b/api/api/views/oauth2_views.py
@@ -136,7 +136,7 @@ class VerifyEmail(APIView):
 
 
 @extend_schema(tags=["auth"])
-class TokenView(BaseTokenView, APIView):
+class TokenView(APIView, BaseTokenView):
     @token
     def post(self, *args, **kwargs):
         """
--- a/api/api/views/oauth2_views.py
+++ b/api/api/views/oauth2_views.py
@@ -139,7 +139,7 @@ class VerifyEmail(APIView):
 @extend_schema(tags=["auth"])
 class TokenView(APIView, BaseTokenView):
     @token
-    def post(self, request, *args, **kwargs):
+    def post(self, request):
         """
         Get an access token using client credentials.
 
--- a/api/test/test_auth.py
+++ b/api/test/test_auth.py
@@ -53,10 +53,9 @@ def test_auth_token_exchange(client, test_auth_tokens_registration):
 
 
 @pytest.mark.django_db
-@pytest.fixture
 def test_auth_token_exchange_unsupported_method(client):
     res = client.get(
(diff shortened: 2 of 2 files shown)


api/api/views/oauth2_views.py
```diff
@@ -136,9 +137,9 @@ def get(self, request, code, format=None):
 
 
 @extend_schema(tags=["auth"])
-class TokenView(BaseTokenView, APIView):
+class TokenView(APIView, BaseTokenView):
     @token
-    def post(self, *args, **kwargs):
+    def post(self, request, *args, **kwargs):
```
krysal, line 142: Looks like after `request` the arguments aren't needed anyomre.
```suggestion
    def post(self, request):
```

dhruvkb: Sure, I thought it was best to keep the `*args` and `**kwargs` params when overridding methods but they are truly not needed in this case.

api/api/views/oauth2_views.py
```diff
@@ -154,7 +155,9 @@ def post(self, *args, **kwargs):
         endpoint.
         """
 
-        return super().post(*args, **kwargs)
+        res = super().post(request._request)
+        data = json.loads(res.content)
+        return Response(data, status=res.status_code)
```
sarayourfriend, line 160: Is it necessary to `json.loads` the content? Would something like this work without the need for that?

```suggestion
        data = json.loads(res.content)
        return HttpResponse(content=res.content, content_type="application/json", status=res.status_code)
```

`HttpResponse` is the more low-level version from `django.http` that allows you to bypass the DRF `Response` data parsing. Doing `json.loads` and then passing it directly to `Response` means we're unnecessary marshalling the JSON string to Python and then immediately back to a string.

sarayourfriend: Although, more general question: why is it necessary to do anything manual here at all, other than just return `res` like we used to? Does `APIView` complain about not receiving the right type of response object or something?

sarayourfriend (changes requested):
Just one small request to remove some unnecessary (at least I _think_, could be wrong) JSON parsing. Krystle's change to the unit test is also critical too :slightly_smiling_face: 

Otherwise LGTM!

api/api/views/oauth2_views.py
dhruvkb: > Using the `Response` class simply provides a nicer interface for returning content-negotiated Web API responses, that can be rendered to multiple formats.
> &mdash; DRF docs

We can use `HttpResponse` for sure, but sending DRF's Response is more consistent as it renders the DRF API UI in the browser.

sarayourfriend: > We can use HttpResponse for sure, but sending DRF's Response is more consistent and it renders the DRF API UI in the browser like our other endpoints.

Ahh, interesting, I didn't realise that was a side effect. For this view, it doesn't seem like a tragic loss, though? It's not exactly a "browseable" response :thinking: At least not any less so if your browser just renders a nice JSON explorer (Firefox does this, not sure of others).

I also don't think we _want_ to support anyone using this view over anything than JSON anyway, but I guess `Response` handles it regardless so :shrug: I would personally use HttpResponse but if you prefer to keep it, all good with me as well :+1:

sarayourfriend (approved):

krysal (approved):
Excellent! Thank you, @dhruvkb.

PR #3198, opened 2023-10-16
Raise error when space is in URL

### LINKED ISSUE #1279: Raise an exception if a space is found in URL fields
## Description

Based on [this discussion](https://github.com/WordPress/openverse-catalog/pull/1003#issuecomment-1447467476) - URLs which get written to TSV should not have spaces in them. It should be up to the provider scripts to handle how those spaces are handled, but we should know before they make it into the database that there's an issue with the URLs.

We can add this in the URL validation step of `MediaStore`:

https://github.com/WordPress/openverse-catalog/blob/9d758e9fea1087cbb2761f96674a7686c788a5ce/openverse_catalog/dags/common/storage/media.py#L129-L132

ngken0995:
## Fixes

Fixes #1279 by @AetherUnbound 

## Description

URLs which get written to TSV should not have spaces in them. I had created a class to state the error. There is a if condition to determine if a space is in the `url_string` then the raise error will populate a string message.

## Testing Instructions

run `just catalog/test`

## Checklist

- [X] I tried running the project locally and verified that there are no visible errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/urls.py
+++ b/catalog/dags/common/urls.py
@@ -37,7 +37,7 @@ def validate_url_string(url_string, strip_slash: bool = True):
     strip_slash: Flag (bool) decides whether to strip slashes in URL or not,
     Default value is `True`
     """
-    if " " in url_string:
+    if url_string and " " in url_string:
         raise SpaceInURLError("space is present in URL")
 
     logger.debug(f"Validating_url {url_string}")
--- a/catalog/dags/common/urls.py
+++ b/catalog/dags/common/urls.py
@@ -37,14 +37,15 @@ def validate_url_string(url_string, strip_slash: bool = True):
     strip_slash: Flag (bool) decides whether to strip slashes in URL or not,
     Default value is `True`
     """
-    if url_string and " " in url_string:
-        raise SpaceInUrlError("space is present in URL")
 
     logger.debug(f"Validating_url {url_string}")
     if not type(url_string) == str or not url_string:
         return
-    else:
-        upgraded_url = _add_best_scheme(url_string, strip_slash)
+
+    if url_string and " " in url_string:
+        raise SpaceInUrlError(f"space is present in URL: {url_string}")
+
+    upgraded_url = _add_best_scheme(url_string, strip_slash)
 
     parse_result = urlparse(upgraded_url)
     tld = tldextract.extract(upgraded_url)
--- a/catalog/dags/common/urls.py
+++ b/catalog/dags/common/urls.py
@@ -17,6 +17,10 @@ from requests.exceptions import RequestException
 logger = logging.getLogger(__name__)
 
 
+class SpaceInURLError(Exception):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


dhruvkb (approved):
I'm not 100% sure of the base premise that URLs with spaces should raise errors because in most cases URL-encoding the space as `%20` makes them valid.

E.g. https://www.google.com/search?q=some%20random%20cat

But if the exception must be raised, this LGTM!

catalog/dags/common/urls.py
```diff
@@ -33,6 +37,9 @@ def validate_url_string(url_string, strip_slash: bool = True):
     strip_slash: Flag (bool) decides whether to strip slashes in URL or not,
     Default value is `True`
     """
+    if url_string and " " in url_string:
```
stacimc, line 40: Maybe a nit, but I would move this check below L45 so that it is after the log statement about beginning URL validation, and after we have already checked that `url_string` is a non-null string.

You can also remove the unnecessary `else` on [L46](https://github.com/WordPress/openverse/pull/3198/files#diff-18a4035f56c5e8dbe16c08fb824fc99a5479d5974726f159017feb25779df742R46).

catalog/dags/common/urls.py
```diff
@@ -33,6 +37,9 @@ def validate_url_string(url_string, strip_slash: bool = True):
...
+        raise SpaceInUrlError("space is present in URL")
```
stacimc, line 41: Can we also log the `url_string` in the error for debugging purposes?

stacimc (changes requested):
> I'm not 100% sure of the base premise that URLs with spaces should raise errors because in most cases URL-encoding the space as %20 makes them valid.

I'm also not sure about this, but we do have a precedent of requiring the provider classes to handle this sort of validation, and erroring in the MediaStore if a provider fails to do so. Interested to see what kind of errors we see as a result of this -- it'll lead to improved provider code :)

@ngken0995 This works great! Could you also add a small unit test?

catalog/dags/common/urls.py
ngken0995: Yes, I will add `url_string` in the error message.

ngken0995:
> > I'm not 100% sure of the base premise that URLs with spaces should raise errors because in most cases URL-encoding the space as %20 makes them valid.
> 
> I'm also not sure about this, but we do have a precedent of requiring the provider classes to handle this sort of validation, and erroring in the MediaStore if a provider fails to do so. Interested to see what kind of errors we see as a result of this -- it'll lead to improved provider code :)
> 
> @ngken0995 This works great! Could you also add a small unit test?

@stacimc I wrote an unit test for space in url. Please take a look at it.

stacimc (approved):
Looks good -- sorry for the delay in review @ngken0995, thanks for re-requesting!

PR #3203, opened 2023-10-16
Limit indexer workers using env var

dhruvkb:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/653 by @AetherUnbound 

## Description

This PR allows the ingestion server to limit the number of indexer workers used in the indexing operation to a subset of all available workers.

This limit is set using an env var and prevents the need to actually change the number of workers provisioned in the infrastructure.

## Testing Instructions

This is non-local change so it must be testing in live environment.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/ingestion_server/distributed_reindex_scheduler.py
```diff
@@ -87,6 +88,13 @@ def _prepare_workers():
         servers.append(server)
         ids.append(_id)
     log.info(f"Selected worker instances {servers}")
+
+    if worker_limit > 0:
+        log.info(f"Truncating worker instances under limit {worker_limit}")
+        servers = servers[:worker_limit]
+        ids = ids[:worker_limit]
+        log.info(f"Truncated worker instances {servers}")
```
sarayourfriend, line 96: If we pass a worker limit _larger_ than the available worker count, what would be the side-effect here? Is there work that's already done that would get unnecessary re-done to recover from the error? Should we put an upper bound on `len(servers)`, or is it better to let it completely fail and make it obvious that there's a configuration error?

sarayourfriend (commented):
> This limit is set using an env var and prevents the need to actually change the number of workers provisioned in the infrastructure.

Such a great idea! To avoid needing to re-provision to modify the environment variable, we can SSH into the ingestion server box and modify the environment variables in the docker-compose.yml, either manually or with Ansible, depending on whether we need to make this change before or after the Ansible setup is available.

Just one request for clarification about error recovery, but nothing major, and will approve as long as there's a clear course of action in case of a hypothetical misconfiguration issue.

ingestion_server/ingestion_server/distributed_reindex_scheduler.py
dhruvkb: Currently, slicing an array with a number larger than the length would have no effect at all and we would use all provisioned workers.

```python
servers = [1, 2, 3]
assert servers[:10] == servers
```

I wanted to avoid raising errors and failing the job so I felt it safer to let the job proceed with all available resources.

We could log something to the effect of f"Limit is higher than number of provisioned workers, using all {worker count} workers." if you feel that would be beneficial?

sarayourfriend: > slicing an array with a number larger than the length would have no effect at all and we would use all provisioned workers

TIL! I assumed Python would through an index out of bounds error, but if there is no error from using a number larger than the array length, then my question is irrelevant. The approach you took makes perfect sense :slightly_smiling_face:

sarayourfriend (approved):
LGTM!

obulat (approved):
LGTM!

PR #3219, opened 2023-10-18
Allow runtime configuration of nginx listening port

sarayourfriend:
## Fixes

Related to https://github.com/WordPress/openverse-infrastructure/pull/648

## Description

ECS FARGATE requires that container and host ports match on a container definition. In the related PR we tried to map Nuxt to a different host port, in order to allow Nginx to listen on 8443 so that the target group port doesn't need to change. We did that by juggling the _host_ ports, but we can't have different host and container ports. Nuxt lets you override the port by passing `-p`, but we need to enable runtime configuration of the Nginx port by using an environment variable.

The net change in this PR is that you should be able to pass `OPENVERSE_NGINX_LISTEN_PORT` environment variable to the container and have Nginx listen on a port other than 8080. Proof that it works is in the fact that it still listens to 8080 now, as that is set as the default for that variable in the container build.

## Testing Instructions

Run `just build frontend_nginx && just frontend/up` and confirm you can still access the `frontend_nginx` service at `localhost:50290`. You also need to run the frontend locally on port `8443` using `just frontend/dev`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
I tried changing the port from 8080 to something else locally, and saw that `docker-compose.yaml` port needs to change, too:

https://github.com/WordPress/openverse/blob/90396b2a0484658acc052009f7bd89d6fc66e380/docker-compose.yml#L333

Once this value is updated, the nginx listening port setting using the env variable works well.

sarayourfriend:
> Can we set the env variable on the parent directory? Or do we need to add documentation saying that the port needs to be updated if the env variable inside frontend/ directory is set to something different than 8080?

We can, but it's just support for testing a feature that shouldn't need to be tested after this. I guess if we want it to be an end-to-end test, we can set the variable in `docker-compose.yml`.

obulat:
> > Can we set the env variable on the parent directory? Or do we need to add documentation saying that the port needs to be updated if the env variable inside frontend/ directory is set to something different than 8080?
> 
> We can, but it's just support for testing a feature that shouldn't need to be tested after this. I guess if we want it to be an end-to-end test, we can set the variable in `docker-compose.yml`. What is your suggested change?

I guess we will not change the 8080 locally anyways, so I'm going to approve this PR. I'm not sure if we should add a comment in `docker-compose` saying that 8080 can be changed using an environment variable in the `frontend/.env`.

obulat (approved):

dhruvkb (approved):
LGTM!

Issue #2653, opened 2023-07-16
National Library of Australia "Trove"


sarayourfriend:
### Provider API Endpoint / Documentation

https://trove.nla.gov.au/about/create-something/using-api/v3/api-technical-guide

### Provider description

The National Library of Australia aggregates works from libraries, museums, and galleries across Australia. The API supports search by rights, pagingation, and date sorting.

The API also includes cultural sensitivity information that should be captured in Openverse and presented on the Openverse frontend (would need new designs for this subset of the work). Related to #2593.

### Licenses Provided

Various CC and PDM

### Provider API Technical info

Requires a key and has rate limits. I'm not sure how many records exist but it would probably be a significant provider.

### Checklist to complete before beginning development

- [X] Verify there is a way to retrieve the entire relevant portion of the provider's collection in a systematic way via their API.
- [X] Verify the API provides license info (license type and version; license URL provides both, and is preferred)
- [X] Verify the API provides stable direct links to individual works.
- [X] Verify the API provides a stable landing page URL to individual works.
- [X] Note other info the API provides, such as thumbnails, dimensions, attribution info (required if non-CC0 licenses will be kept), title, description, other meta data, tags, etc.
- [ ] Attach example responses to API queries that have the relevant info.

### Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
NLA also has a Flickr commons page that could be added as a separate, distinct source: https://www.flickr.com/photos/national_library_of_australia_commons/

This is different from "trove" which is the NLA's tool that aggregates multiple GLAM institutions across Australia.

sarayourfriend:
I've just discovered that Trove also includes a good deal of Flickr stuff and we already overlap with them for some stuff. We'd need to filter out Flickr results, I guess?

E.g.: https://openverse.org/image/ffed1feb-0f8b-42fb-a016-2d824983156a?q=galah
https://trove.nla.gov.au/work/231741142?keyword=galah&l-advrights=Free%20with%20conditions&l-advrights=Free

zackkrida:
This is a really remarkable resource! Very compelling approaches to text correction, cultural sensitivity info, and other community features.

PR #3258, opened 2023-10-25
Retrieve Auckland Museum Image Data

### LINKED ISSUE #1771: Auckland Museum
## Source Site
https://www.aucklandmuseum.com/discover/collections/our-data

## Provider API Endpoint / Documentation
Tutorial: https://github.com/AucklandMuseum/API/wiki/Tutorial#fields-available-for-query-string-searches
Endpoint: https://api.aucklandmuseum.com/

## Provider description
The Auckland Museum displays wide varieties of historical artifacts and collections in addition to providing education and research resources.

## Provider API Technical info
Rate Limit: 10 requests per second/ 1000 requests per day.

--------

This issue has been migrated from the [CC Search Catalog repository](https://github.com/creativecommons/cccatalog/issues/280)

```
Author: akshgpt7
Date: Mon Feb 24 2020
Labels: Hacktoberfest,help wanted,providers,✨ goal: improvement,🙅 status: discontinued
```
### Original Comments: 

*akshgpt7* commented on [Tue Feb 25 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-590608109):

>@annatuma, @mathemancer I found out that the Auckland Museum's online collection has CC licensed images, which can be identified by the `copyright` field in the response JSON.

*mathemancer* commented on [Tue Feb 25 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-590792082):

>@akshgpt7 That's great, thank you!  Further info to gather would be:

> - [ ] What is the overall volume of objects to be queried?
> - [ ] Is it possible to catalog *all* CC-licensed items via API calls in some reasonable way?  (I.e., there must be some way to systematically loop through all objects, gathering their metadata)

*akshgpt7* commented on [Fri Feb 28 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-592194998):

>@mathemancer 
- The current volume of objects with `CC` in the `copyright` field is a bit over 100k.
- To catalog the CC-licensed items, the following endpoint can be used `https://api.aucklandmuseum.com/search/collectionsonline/_search?q=copyright:CC&has_image=true`. For each object, there's a `primaryRepresentation` field that contains the image url. I believe it's very much possible to extract the items through the same. 
Most of the images are licensed under this license: http://creativecommons.org/licenses/by/4.0/ (Needs confirmation).

*mathemancer* commented on [Mon Mar 02 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-593598110):

>This is great info @akshgpt7 , thank you!

*annatuma* commented on [Thu Mar 05 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-595443990):

>@akshgpt7 you are welcome to tackle this integration if you're interested in doing so. Let us know, for now this ticket is assigned to you to work on.

*akshgpt7* commented on [Fri May 29 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-635614162):

>@mathemancer I have been working on this script. 
I have one question moving forward. There is a rate limit of 1000 requests per day. How do I go about handling checks to not exceed that in a day?

I was thinking something like getting the time of the day at the start of the script, and maintaining a `request_count`. Then getting the time of the day on each request and stop the script if `request_count` hits 1000 (on the same day), or refresh it if the day passes before completing 1000 requests.

However, I'm not sure if this is the right way to go about handling the 1000 requests/day limit. Moreover, how do we make sure to start off the next day from the same page where we left on the previous day? 

*mathemancer* commented on [Mon Jun 15 2020](https://github.com/creativecommons/cccatalog/issues/280#issuecomment-644166001):

>@akshgpt7 I suggest using the `DelayedRequester` class with a delay of 87 seconds.  This will keep the overall number of requests under the limit.

ngken0995:
## Fixes

Fixes #1771 by @obulat 

## Description

Adds the script to get all the media from aucklandmuseum.com. Currently, there are 10000 images because the search result max amount is 10000 images.

To collect the filesize, this script makes head requests for individual media items, which make the script slower than expected. But it should be okay considering the number of available images.
Image dimensions are not available, so they will need to be collected separately in the future.

## Testing Instructions

run `just catalog/test -k auckland_museum`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
This is awesome work, @ngken0995 !

To add the DAG that would run your script at specific intervals, you'll also need to add it to `catalog/dags/providers/provider_workflows.py`. Then, if you run `just up`,  go to 0.0.0.0:9090 (log in with `airflow`/`airflow`), you can see all of the current DAGs, and you can start the Auckland museum DAG by clicking on the green triangle. Then, you can view the logs from the dag, and can "mark" the "pull_data" step "successful" after several minutes. The DAG will then go on to save the data to a Postgres database on your local machine!

ngken0995:
@obulat I was able to find the `foreign_landing_url` and `foreign_identifier` by cleaning up the `_id` url. I have add AucklandMuseumDataIngester to provider_workflow and a successful run from DAG. Can you double check if it ran correctly?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,143 @@
...
+        url_parameter = data.get("_id").split("id/")[-1].replace("/", "-")
+        foreign_landing_url = f"{LANDING_URL}{url_parameter}"
+
+        foreign_identifier = data.get("_id").split("/")[-1]
+
+        information = data.get("_source")
+
+        url = information.get("primaryRepresentation")
+
+        thumbnail_url = f"{url}?rendering=thumbnail.jpg"
```
stacimc, line 90: The thumbnail their API provides is tiny, fixed width of 70px. @obulat would know best -- should we use this or just default to `None` here and use our own thumbnail service? They also have a slightly bigger `preview` rendering with a fixed width of 100px.

stacimc (changes requested):
Fantastic @ngken0995, I'm excited to see you jumping in to adding a new provider! I suspect you might be having trouble running the DAG locally because of the `start_date` being in the future (see my comment below). Once you update the `start_date`, you'll be able to run it locally and actually get data.

Once you've ingested some data locally, you can see what it looks like by querying your local catalog database. Run `just catalog/pgcli` to open pgcli in your terminal, and then you can run sql queries (ie `select * from image where provider='aucklandmuseum' limit 10;`).

I was not initially able to ingest any data, because the DAG fails on every image when trying to fetch the file size with a 301. I commented this part out for the sake of testing the rest of the code.

On a higher-level note: the `100,000` number comes from the rate limit and max response size, right? Meaning there's actually more data than we can fetch in a day? My concern is that when the DAG is run a second time, it'll start processing from the beginning all over again. As it currently stands, I don't think we'll ever be able to ingest more than those first 100k rows.

If the API supports date range queries, we could consider making this a dated DAG instead. Otherwise, we might need to be a bit creative. The absolute simplest, somewhat silly solution I can think of is to give it a huge timeout and greatly extend the `delay` between requests such that the DAG runs over a week or so, only fetching 100k a day in order to respect the rate limit. I'm not sure what their total dataset size is, so not sure if that's feasible!

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,143 @@
...
+        # On the first request, `prev_query_params` will be `None`. We can detect this
+        # and return our default params.
+        if not prev_query_params:
+            # Return default query params on the first request
+            # primaryRepresentation contain a image url for each data
+            # "+" is a query string syntax for must be present
+            # copyright:CC state Creative Commons Attribution 4.0
+            return {
+                "q": "_exists_:primaryRepresentation+copyright:CC",
+                "size": "100",
```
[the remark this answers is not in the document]
ngken0995: The max amount of data retrieved from the api is `10,000`. Look at `hits -> total -> value`([api](https://api.aucklandmuseum.com/search/collectionsonline/_search?q=_exists_:primaryRepresentation+copyright:CC&size=1&from=200)) `Size` is the amount of data to present in a get request. `From` is the index of the total value. We can keep incrementing `From` till it reach `10,000` and `get_should_continue` function should know when it reach the limit.

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,143 @@
...
+        information = data.get("_source")
+
+        url = information.get("primaryRepresentation")
+
+        thumbnail_url = f"{url}?rendering=thumbnail.jpg"
+        license_info = self.DEFAULT_LICENSE_INFO
+        filesize = self._get_file_info(url)
+
+        creator = (
+            information.get("dc_contributor")[0]
```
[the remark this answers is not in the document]
ngken0995: Thank you for the review. It does make sense to set `get` with a default value. please take a look at the if-else condition.

ngken0995:
>I was not initially able to ingest any data, because the DAG fails on every image when trying to fetch the file size with a 301. I commented this part out for the sake of testing the rest of the code.

@stacimc I tried to add the url with `?rendering=original.jpg` and some of them gave me `message | "Internal server error"` response page. What is the best way to handle every image when trying to fetch the file size with a 301?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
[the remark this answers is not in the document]
stacimc: I meant that by default `self.batch_limit` is 100, and that you can just say:

```
    "size": self.batch_limit,
```

Rather than hard-coding it separately here.

stacimc:
> What is the best way to handle every image when trying to fetch the file size with a 301?

From a quick look, it does look like the API returns image urls with `http` but redirects to `https` in browser. I tried updating the urls to https in `get_filesize` and this results in 403s on the HEAD request. I didn't have much time to look into this, but if it's a problem on their side we may have to just skip getting file size information for now.

> On a higher-level note: the 100,000 number comes from the rate limit and max response size, right? Meaning there's actually more data than we can fetch in a day? My concern is that when the DAG is run a second time, it'll start processing from the beginning all over again. As it currently stands, I don't think we'll ever be able to ingest more than those first 100k rows.

Reiterating this point from earlier: it occurred to me as I was looking at this again, is there a reason the `batch_limit` here _needs_ to be the default 100? The API limits to 1,000 requests per day, but now that I look I don't see anywhere in the documentation saying that we have to limit it to 100 records per request. Can you increase that limit in order to process more data?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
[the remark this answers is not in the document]
ngken0995: I will update "size" with batch_limit.

ngken0995:
> > On a higher-level note: the 100,000 number comes from the rate limit and max response size, right? Meaning there's actually more data than we can fetch in a day? My concern is that when the DAG is run a second time, it'll start processing from the beginning all over again. As it currently stands, I don't think we'll ever be able to ingest more than those first 100k rows.
> 
> Reiterating this point from earlier: it occurred to me as I was looking at this again, is there a reason the `batch_limit` here _needs_ to be the default 100? The API limits to 1,000 requests per day, but now that I look I don't see anywhere in the documentation saying that we have to limit it to 100 records per request. Can you increase that limit in order to process more data?

My apologizes, I should have stated why I set the size as a default for 100. I didn't know what the correct `batch_limit` should be and used 100 as a placeholder. I believe I can set it as 2,000. How should I determine what is a good `batch_limit`?

stacimc:
> I didn't know what the correct batch_limit should be and used 100 as a placeholder. I believe it can set to 2,000. How should I determine what is a good batch_limit?

Nice! Ideally, we want to be able to ingest all the records that the provider has. I'm not sure how much that is -- I didn't see it in their API response or documentation anywhere. I would start by increasing the batch size to the max and testing it locally to see if you can get it to complete ingestion (i.e., run out of records to process from the API). From there, we can dial the batch limit back if needed.

To do that, though, I think you'll need to get rid of the logic in `get_should_continue` that stops ingestion prematurely. I would try just commenting out the method temporarily so it defaults to the base class implementation (which just returns `True`). Hopefully it will halt gracefully on its own, but if not you can adjust the logic once you're done testing.

ngken0995:
@stacimc I have increased `batch_limit` to 2000 and `size` is set to 2000. Once `from` plus `size` is above 10000, there will be an error [message](https://api.aucklandmuseum.com/search/collectionsonline/_search?q=_exists_:primaryRepresentation+copyright:CC&size=2000&from=8001). I kept `get_should_continue` because without it Auckland museum DAG will be on queue and doesn't success or fail. Will it ever timeout on it's own?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,159 @@
...
+            if information.get("dc_contributor", [])
+            else ""
+        )
+
+        title = (
+            information.get("appellation").get("Primary Title")[0]
+            if information.get("appellation", [])
+            and information.get("appellation").get("Primary Title", [])
+            else ""
+        )
```
stacimc, line 112: Maybe:
```suggestion
        appellation = information.get("appellation", {})
        title = (
            appellation.get("Primary Title")[0]
            if appellation.get("Primary Title")
            else None
        )
```

catalog/dags/providers/provider_api_scripts/auckland_museum.py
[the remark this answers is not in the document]
stacimc: Thanks! I think we'll just want to default to `None` instead of an empty string here in the `else`.

stacimc (commented):
@ngken0995 I really appreciate your patience on this issue. I answered a few of your questions below, but basically: long term, I think we'll need to make the modifications I describe in this comment below in order for the DAG to be fully functional. But since this issue has gotten more complicated than we initially thought, I'd _also_ be happy to merge this as-is and create a separate issue for making those updates later 🙂 Let me know if you have a preference!

To get this PR across the line, I just had a couple more comments, but feel free to investigate the dated DAG idea further if you're interested.

_______________________________________________
_Everything that follows could be addressed in a follow-up issue:_

> Once from plus size is above 10000, there will be an error [message](https://api.aucklandmuseum.com/search/collectionsonline/_search?q=_exists_:primaryRepresentation+copyright:CC&size=2000&from=8001).

Oh dear -- that error message comes from Elasticsearch. It indicates their API _is_ limited to returning a max of 10k records for any particular query, so simply increasing the batch size will not work for us after all. As is, every time the DAG is run it will just ingest the first 10k records and then stop.

To get around this on our end, we'll need to come up with a way to break their dataset up into smaller subqueries that each have less than 10k results. The most obvious way I can think of is to make this a `dated` DAG that runs on a `@daily` schedule and ingests only records that were `lastModifiedOn` the logical date. (You can look at Metropolitan Museum for an example of a dated provider).

As for modifying the query to filter on `lastModifiedOn`, I think this is possible based on the API documentation, but it would require overriding some of the base ProviderDataIngester behavior.  The tutorial describes support for [complex searches](https://github.com/AucklandMuseum/Online_Cenotaph_API/wiki/Tutorial#complex-searches) that involve POSTing more complex Elasticsearch queries, including ranged queries. It would be worth a shot to play around with this and see if we can query on `lastModifiedOn` using `self.date` (Wikimedia does something similar [here](https://github.com/WordPress/openverse/blob/main/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L182)). I haven't gotten a chance to try this out but it looks promising 🤞 

> I kept get_should_continue because without it Auckland museum DAG will be on queue and doesn't success or fail. Will DAGs ever timeout on it's own if get_should_continue is removed?

Yes, by default provider DAGs will end gracefully without overriding `get_should_continue`, _as long as_ the provider API being queried doesn't raise an error when you exceed the last batch. Brooklyn and Cleveland Museums are examples. In this provider's case, it looks like exceeding 10k records will cause a failure so we _do_ want to handle it. But I actually think it's a good idea for us to raise an error here (rather than silently halting ingestion), because if this condition is met it means that the DAG failed to ingest all the available records. Maybe we should re-raise with a better error message here instead.

ngken0995:
@stacimc Thank you for clarifying all my questions and giving suggestion on a `dated` DAG approach. I'm going to set this to draft and look into `lastModifiedOn` using `self.date`. I prefer a function which collect majority of the data than limited to a specific amount.

ngken0995:
@stacimc There is a postman setup for complex searches([link](https://github.com/amcollections/APIDocs/wiki/Postman-Set-Up#setting-up-postman-with-auckland-museum-api)). Here is an example of a POST query using the boolean search including a wildcard query to search for `Auckland` in the copyright field, an exist query to populate `primaryRepresentation` that is the image url, a date range query stating `lastModifiedOn` from a date to another date. What should I look up to override ProviderDataIngester and send a POST query with a body?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,159 @@
+"""
+Content Provider:       AucklandMuseum
```
sarayourfriend, line 2:
```suggestion
Content Provider:       Auckland War Memorial Museum Tāmaki Paenga Hira
```

catalog/dags/common/loader/provider_details.py
```diff
@@ -12,6 +12,7 @@
 
 
 # Default provider names
+AUCKLAND_MUSEUM_IMAGE_PROVIDER = "aucklandmuseum"
```
sarayourfriend, line 15: Unfortunately it's not realistic to have diacritical marks (macron) on these slugs, but we can still have both. I left out "war memorial" from the colonial name because it's less important and the institution itself doesn't always include it.

```suggestion
AUCKLAND_MUSEUM_IMAGE_PROVIDER = "aucklandmuseum_tamakipaengahira"
```

sarayourfriend (commented):
Super excited about this provider addition! Fun fact, my spouse used to work there!

Based on that, I have some cultural sensitivity information to incorporate into this PR: we should use the full name of the museum including the te reo Maori name. That's the correct name of the museum in full and leaving out the te reo Maori part in particular should not happen and would be deeply offensive to folks in Aotearoa/New Zealand and people related to the museum in particular.

"Auckland War Memorial Museum Tāmaki Paenga Hira" (Wikipedia has a comma between the two, but I have it on good authority that the comma should not be used)

Anywhere we've referred to the name of the museum should ideally include the te reo name, especially if it's visible by users of the data (e.g., in the provider name).

Ideally it's in the provider name so that this is evident through the API. I know it's long to slugify that, but we've got very long source names already anyway (Smithsonian) and this is a cultural sensitivity/colonial issue that is very important in Aotearoa/New Zealand. I've left a suggestion for that.

I considered suggesting using `TamakiPaengaHiraDataIngester` but the codebase is in English, and we use English names for other non-English (or non-exclusively-English) institutions like SMK and Finnish Museums. In other words, the English name is consistent there, and as long as both are used at all, the English name will be consistent with the code base.

For presentation and for the sake of respect, it's critical that we don't accidentally leave out the te reo Maori name in places users will come across it.

stacimc:
> What should I look up to override ProviderDataIngester and send a POST query with a body?

We'll probably want to override [get_response_json](https://github.com/WordPress/openverse/blob/main/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L404) on the base class, which currently makes a GET request using the [DelayedRequester](https://github.com/WordPress/openverse/blob/main/catalog/dags/common/requester.py#L31).

catalog/dags/common/loader/provider_details.py
ngken0995: This is good to hear. The changes are made.

sarayourfriend:
Nina (my spouse) mentioned to me something about Auckland Museum's image of their reproduction of the Magna Carta once being their most viewed collection item, because it was the image on the Wikipedia page for the Magna Carta (it isn't any longer). That made me wonder if Auckland Museum's collection is already in Openverse by way of Wikimedia Commons, and indeed, they have _heaps_ of works in Wikimedia Commons (and apparently work closely with Wikimedia, even having a [Wikimedian in residence at the museum](https://www.aucklandmuseum.com/your-museum/about-us/blog/2017/wikimedian-in-residence)).

https://commons.wikimedia.org/wiki/Category:Images_from_Auckland_Museum

That page reports 175,608 items from the Museum in Wikimedia Commons.

Indeed, I can find [this one from the museum](https://commons.wikimedia.org/wiki/File:A_caught_shark_laid_over_ropes_(AM_75645-1).jpg) in [Openverse here, with sadly terrible metadata and presentation](https://openverse.org/image/4a0e6e65-205e-4ae8-893c-0a6bae0a4054?q=A%20caught%20shark%20laid%20over%20ropes).

If we add the museum as a separate _provider_, then we'd effectively duplicate every single entry in Wikimedia Commons. Either we'd need to suppress the entries from Wikimedia Commons, or, (probably my preference) improve our ingestion of Wikimedia Commons to be able to identify sources like this in Wikimedia Commons. Glancing at the Wikimedia Commons provider script, I don't think we currently save the "collections" metadata present in the file summary on the Wikimedia Commons page.

I think this is a big opportunity to expand the list of high quality sources without introducing duplicates, and while cleaning up the Wikimedia Commons data ingestion, cleanup, and overall handling. For this institution in particular, there is a great page describing how the metadata is structured: https://commons.wikimedia.org/wiki/Commons:Batch_uploading/AucklandMuseumCCBY

The same information would also be relevant for the National Gallery of Art (see [this Wikimedia result](https://commons.wikimedia.org/w/index.php?curid=78060790), which [is in Openverse with similarly poorly handled metadata](https://openverse.org/image/f1a2f23d-1e3d-4759-8933-962cb0bbd0bb) and is in a NGA collection in Wikimedia Commons's data). I imagine there are a handful of other such institutions that we could add, just by improving the Wikimedia Commons script and our handling of their data.

And actually, when digging through Wikimedia Commons and Wikidata pages researching this comment, I found this _amazing_ spreadsheet that would help us identify these exact kinds of institutions, for Wikimedia Commons, Europeana, Flickr, and even #2653: https://docs.google.com/spreadsheets/d/1WPS-KJptUJ-o8SXtg00llcxq0IKJu8eO6Ege_GrLaNc/edit#gid=1216556120

ngken0995:
> If we add the museum as a separate _provider_, then we'd effectively duplicate every single entry in Wikimedia Commons. Either we'd need to suppress the entries from Wikimedia Commons, or, (probably my preference) improve our ingestion of Wikimedia Commons to be able to identify sources like this in Wikimedia Commons. Glancing at the Wikimedia Commons provider script, I don't think we currently save the "collections" metadata present in the file summary on the Wikimedia Commons page.
> 
> I think this is a big opportunity to expand the list of high quality sources without introducing duplicates, and while cleaning up the Wikimedia Commons data ingestion, cleanup, and overall handling. For this institution in particular, there is a great page describing how the metadata is structured: https://commons.wikimedia.org/wiki/Commons:Batch_uploading/AucklandMuseumCCBY

Is this PR blocked until there is a way to clean up Wikimedia Commons and avoid duplicates?

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,206 @@
...
+    DEFAULT_LICENSE_INFO = get_license_info(license_url=license_url)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.delay = 4
+        self.batch_start = 0
+        self.batch_limit = 2000
+        self.headers = {"Content-Type": "application/json"}
+        self.date_from = convert_date_format(datetime.now() - timedelta(days=1))
+        self.date_to = convert_date_format(datetime.now())
```
stacimc, line 58: Instead of basing these on `datetime.now()`, we should use the logical date of the DagRun, which you can get using `self.date` (see Metropolitan for example). We should use `self.date` as the `date_from`, and `self.date + timedelta(days=1)` for `date_to`.

This would allow us to do backfills, for example -- if you run the Dag with `date` set to `10-01-23`, it will ingest records for that date, instead of re-ingesting today's records.

catalog/dags/providers/provider_workflows.py
```diff
@@ -192,6 +193,10 @@ def __post_init__(self):
 
 
 PROVIDER_WORKFLOWS = [
+    ProviderWorkflow(
+        start_date=datetime(2023, 10, 1),
+        ingester_class=AucklandMuseumDataIngester,
```
stacimc, line 198:
```suggestion
        ingester_class=AucklandMuseumDataIngester,
        schedule_string="@daily",
        dated=True,
```

We just need to add these in to make the DAG dated 🎉

stacimc (commented):
This looks fantastic, I'm so pleased you were able to turn this into a dated DAG! I think this is extremely close, just a few quick configuration changes needed to make sure it handles ingestion dates properly.

> Is this PR blocked until there is a way to clean up Wikimedia Commons and avoid duplicates?

I don't think so! That's a great discovery and we'll definitely want to make issues for the Wikimedia cleanup and deduplication effort. But we should absolutely be able to get this merged and start ingesting data into our catalog at the very least. Worst case scenario, we don't need to enable the source in the API until we're satisfied that the records are deduplicated -- but in any case I think it's a great idea to start this backfilling ASAP.

catalog/dags/providers/provider_api_scripts/auckland_museum.py
ngken0995: That is a great idea!

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,214 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.delay = 4
+        self.batch_start = 0
+        self.batch_limit = 2000
+        self.headers = {"Content-Type": "application/json"}
+        if self.date:
+            date_from = datetime.strptime(self.date, "%Y-%m-%d").date()
+            self.date_from = str(date_from)
+            self.date_to = str(date_from + timedelta(days=1))
```
stacimc, line 60:
```suggestion
            date_from = datetime.strptime(self.date, "%Y-%m-%d")
            self.date_from = date_from.isoformat()
            self.date_to = (date_from + timedelta(days=1)).isoformat()
```

Just one last thing! We're sending just the date strings to the API, so for example `2023-10-01` to `2023-10-02` as the range. I tested this locally and it looks like the API interprets that as any records last modified on 2023-10-01 _or_ 2023-10-02, so a 48 hour range.

If we instead send full timestamps `2023-10-01T00:00:00` to `2023-10-02T00:00:00`, it correctly restricts it to the 24-hr window we want :)

Also, now that the DAG is dated we can count on `self.date` to always be passed (the base class handles that for us). So you can remove the conditional and the `convert_date_format` method too 🎉

stacimc (approved):
It's working great locally! 🎉 It looks like there's just a merge conflict that will need to be fixed before this can be merged.

I also left a comment about fixing the date range and cleaning up a branch we don't need. I'm not blocking on it since this does work, we just end up getting duplicates during ingestion (that get handled during the load steps), but I do think we should make that change for clarity.

Thanks so much for pushing this over the line, especially as it became quite a bit more complex than initially planned! 💯

catalog/dags/providers/provider_api_scripts/auckland_museum.py
ngken0995: It took awhile to find the correct method and function to convert `self.date`. Thank you!

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+        return True
+
+    def get_media_type(self, record: dict):
+        return IMAGE
+
+    def get_record_data(self, data: dict) -> dict | list[dict] | None:
+        # check if _id is empty then foreign_landing_url and
+        # foreign_identifier doesn't exist
+
+        if not data.get("_id"):
```
obulat, line 108:
```suggestion
        if not (identifier := data.get("_id")):
```
We could simplify the code by saving the value of `_id` and reusing it below.

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+            return None
+
+        url_parameter = data.get("_id").split("id/")[-1].replace("/", "-")
```
obulat, line 111:
```suggestion
        url_parameter = identifier.split("id/")[-1].replace("/", "-")
```

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+        foreign_landing_url = f"{LANDING_URL}{url_parameter}"
+
+        foreign_identifier = data.get("_id").split("/")[-1]
```
obulat, line 114:
```suggestion
        foreign_identifier = identifier.split("/")[-1]
```

catalog/dags/providers/provider_api_scripts/auckland_museum.py
obulat: We have previously discussed the thumbnail sizes, and decided against using thumbnails smaller than 600px: https://github.com/WordPress/openverse/issues/675#issuecomment-1472399310

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+
+        information = data.get("_source", {})
+
+        if not (url := information.get("primaryRepresentation")):
+            return None
+
+        thumbnail_url = f"{url}?rendering=thumbnail.jpg"
```
obulat, line 121:
```suggestion
```

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+            else None
+        )
+        meta_data = self._get_meta_data(information)
+
+        return {
+            "foreign_landing_url": foreign_landing_url,
+            "foreign_identifier": foreign_identifier,
+            "url": url,
+            "license_info": license_info,
+            "thumbnail_url": thumbnail_url,
```
obulat, line 143:
```suggestion
```

catalog/dags/providers/provider_api_scripts/auckland_museum.py
```diff
@@ -0,0 +1,201 @@
...
+            "creator": creator,
+            "title": title,
+            "meta_data": meta_data,
+        }
+
+    def _get_meta_data(self, object_json: dict) -> dict | None:
```
obulat, line 149:
```suggestion
    @staticmethod
    def _get_meta_data(self, object_json: dict) -> dict | None:
```
This method does not use any data from the class instance (`self`), so it's better to make it static (as Pycharm IDE suggests :) )

obulat (commented):
Thank you for such a great contribution, @ngken0995 ! I ran the DAG locally, and it works well.

I am concerned with the quality of data we collect, though. I got around 60 items locally, and a very large proportion of them either show "Server error" or "Online image not available" image for the main image file.

I think we should check the main `url` before saving the item to the catalog for this provider, otherwise we risk downloading a lot of dead links here. What do you think, @openverse-catalog?

Here's a sample of urls I got locally:
```
+---------------------------------------------------+---------------------------------------------------------------------------------------------------------+------------------------------------+
| url                                               | foreign_landing_url                                                                                     | title                              |
|---------------------------------------------------+---------------------------------------------------------------------------------------------------------+------------------------------------|
| https://api.aucklandmuseum.com/id/media/v/2882    | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-18430     | jar                                |
| https://api.aucklandmuseum.com/id/media/v/117250  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-15507     | glass, wine                        |
| https://api.aucklandmuseum.com/id/media/v/3191    | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-9923      | jar, lidded                        |
| https://api.aucklandmuseum.com/id/media/v/861840  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-22679     | briefs, pair                       |
| https://api.aucklandmuseum.com/id/media/v/370276  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-608181    | cartridges                         |
| https://api.aucklandmuseum.com/id/media/v/325015  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-4375      | teabowl                            |
| https://api.aucklandmuseum.com/id/media/v/528116  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-1151      | bowl, lidded                       |
| https://api.aucklandmuseum.com/id/media/v/34541   | https://www.aucklandmuseum.com/collections-research/collections/record/am_naturalsciences-object-368805 | Carex resectans Cheeseman          |
| https://api.aucklandmuseum.com/id/media/v/828322  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-90135     | skirt, wool                        |
| https://api.aucklandmuseum.com/id/media/v/229298  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-61319     | tablecloth, signature              |
| https://api.aucklandmuseum.com/id/media/v/75802   | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-14260     | cup                                |
| https://api.aucklandmuseum.com/id/media/v/117280  | https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-12592     | goblet                             |
+---------------------------------------------------+---------------------------------------------------------------------------------------------------------+------------------------------------+

```

sarayourfriend:
@obulat which ones did you see that errored? I chose some random ones from the list you shared and they all worked for me. I'm curious to see whether they're present in the wikimedia commons dataset (perhaps they've already sorted through those ones in the Wikimedia dump).

obulat:
> @obulat which ones did you see that errored? I chose some random ones from the list you shared and they all worked for me. I'm curious to see whether they're present in the wikimedia commons dataset (perhaps they've already sorted through those ones in the Wikimedia dump).

I think you _can_ get to the main images for the next 2 sections from their landing URLs, but it's not easy to derive their URLs.

### "Online image not available" placeholder
https://api.aucklandmuseum.com/id/media/v/3191, https://api.aucklandmuseum.com/id/media/v/2882,

### Internal server error
https://api.aucklandmuseum.com/id/media/v/861840
https://api.aucklandmuseum.com/id/media/v/528116
https://api.aucklandmuseum.com/id/media/v/828322
https://api.aucklandmuseum.com/id/media/v/229298

### No image on the landing page
https://www.aucklandmuseum.com/collections-research/collections/record/am_humanhistory-object-9923

### Placeholder image on the landing page
https://www.aucklandmuseum.com/collections-research/collections/record/am_naturalsciences-object-368805

Maybe there's a geographical access problem?

It's a good idea to check for these items in Wikimedia.

ngken0995:
I have the same errors and I'm based in the US.

sarayourfriend:
From what my spouse has told me, the museum is quick to remove public access things for cultural sensitivity reasons, in favour of having controlled access with a culturally relevant approach. I wouldn't be surprised if there are public _records_ for those items where the images aren't available.

If we want to check, I'm pretty sure that the museum would respond to an email from us, and if not, I can ask my spouse to get us in touch with someone, they still have connections with folks there. At the very least we could clarify which of these intentionally do not have access so that we 100% do not index them (knowing that they won't ever be available) and which are technical errors, if such a distinction even exists.

obulat:
> In the meantime, I think this PR could be merged as-is, although the data quality should be addressed before we actually turn the DAG on in production and begin ingesting. This is in line with how we've managed the addition of some other providers that ended up being very complex (eg iNaturalist). We should add a row to the DAG Status page explaining why the DAG is not yet enabled, and prioritize that work separately.

I like this solution, @stacimc, nice to have such a workaround.

obulat (approved):
Thank you for adding a new DAG, @ngken0995, and all of your patience during the review! It's interesting that this DAG adds a new POST request to the ingester.

PR #3261, opened 2023-10-26
Simplify search query

### LINKED ISSUE #3243: Search controller uses `should` for filtering instead of `filter`
## Description

If you filter the search by `license`, the search controller creates the following query:
```
 "filter": [ { "bool": { "should": [
              { "terms": {
                  "license": ["by", "by-sa", "by-nd", "by-nc", "by-nc-nd"],
                  "boost": 1.0
                }  }
            ] } },
       ...
      ],
```

There are two problems with it:
- it uses `should`, which means that the results with other licenses _can_ be returned (it's not a mandatory filter)
- it uses a `boost` parameter which is unnecessary: instead of increasing the score of the item that has the required licenses, we should just filter out all of the items that don't have the required license.

## Solution
We should replace `should` with `filter` here:

https://github.com/WordPress/openverse/blob/09879490d0f8d0177dd2877e4aead4cc9a24f201/api/api/controllers/search_controller.py#L260-L270

should be:

```
method = getattr(s, behaviour) 
return method(query) 
```
This will use `elasticsearch_dsl`'s `filter` or `exclude` methods to create `filter` or `must_not` clauses.

## Reproduction

1.  Add a line `DEBUG_SCORES=True` to the `api/.env` file
2. Search for items with a license filter: api.localhost:50280/v1/images/?license=by,by-sa
3. Check the logs for the `web` docker container. It should show the query similar to the one in this issue.

obulat:
## Fixes

Fixes #3243 by @obulat

## Description

Instead of using `elasticsearch_dsl` for creating the search query, this PR uses a simple dictionary for the 4 query clauses: `filter`, `must_not`, `must` and `should`. This dictionary is then used to create a much cleaner query which is almost identical in function to the existing query. 

The biggest difference is that the url parameter queries (license, aspect ratio, width, extension, etc.) are actually filters. They remove all of the non-matching items from the results, instead of simply assigning them a lower score. Not calculating the score should also in theory make the filters faster.

### Query examples
To make these queries similar to the production queries, I checked "Hide content" in the Flickr ContentProvider object form in Django admin. This should dynamically add this provider to `exclude` query.
#### No parameters
http://localhost:50280/v1/images/?format=json&page_size=1
**Updated**
```
{'bool': {
    'must': [{'match_all': {}}],
    'must_not': [{'term': {'mature': True}}, {'terms': {'provider': ['flickr']}}],
    'should': [{'rank_feature': {'boost': 10000, 'field': 'standardized_popularity'}}]
}}
```
**Old**
```
{'bool': {'must': [{'bool': {'filter': [{'bool': {'must_not': [{'term': {'mature': True}}]}},
                                                  {'bool': {'must_not': [{'terms': {'provider': ['flickr']}}]}}]}}],
                    'should': [{'rank_feature': {'boost': 10000,
                                                 'field': 'standardized_popularity'}}]}}
```
#### `q` search with filters
http://localhost:50280/v1/images/?q=cat&format=json&license=by,cc0&aspect_ratio=wide&page_size=1
**Updated**
```
{'bool': {
    'filter': [{'terms': {'license': ['by', 'cc0']}}, {'terms': {'aspect_ratio': ['wide']}}],
    'must': [{'simple_query_string': {'default_operator': 'AND', 'fields': ['title', 'description', 'tags.name'], 'query': 'cat'}}],
    'must_not': [{'term': {'mature': True}}, {'terms': {'provider': ['flickr']}}],
    'should': [{'simple_query_string': {'boost': 10000, 'fields': ['title'], 'query': 'cat'}}, {'rank_feature': {'boost': 10000, 'field': 'standardized_popularity'}}]
}}
```
**Old**
```
{'bool': {
    'must': [{'bool': {'must': [{'bool': {
                 'filter': [
                   {'bool': {'should': [{'terms': {'license': ['by', 'cc0']}}]}},
                   {'bool': {'should': [{'terms': {'aspect_ratio': ['wide']}}]}},
                   {'bool': {'must_not': [{'term': {'mature': True}}]}},
                   {'bool': {'must_not': [{'terms': {'provider': ['flickr']}}]}}
                 ],
                 'must': [{'simple_query_string': {
                                 'default_operator': 'AND', 'fields': ['tags.name', 'title', 'description'],  'query': 'cat'}}]}}],
                 'should': [{'simple_query_string': {'boost': 10000,  'fields': ['title'], 'query': 'cat'}}]}}],
                  'should': [{'rank_feature': {'boost': 10000, 'field': 'standardized_popularity'}}]}},}
```

### `title` search (without `q`) with `excluded_source` param
http://localhost:50280/v1/images/?title=cat&format=json&excluded_source=stocksnap&page_size=1

**New**
```
{'query': {'bool': {
    'must': [{'simple_query_string': {'fields': ['title'], 'query': 'cat'}}],
    'must_not': [{'terms': {'source': ['stocksnap']}},{'terms': {'provider': ['flickr']}}, {'term': {'mature': True}}],
     'should': [{'rank_feature': {'boost': 10000, 'field': 'standardized_popularity'}}]}}
}
```
**Old**
```
{ 'query': {'bool': {
    'must': [{'bool': {'filter': [{'bool': {'must_not': [{'terms': {'source': ['stocksnap']}}]}},
                                                  {'bool': {'must_not': [{'term': {'mature': True}}]}},
                                                  {'bool': {'must_not': [{'terms': {'provider': ['flickr']}}]}}],
    'must': [{'simple_query_string': {'fields': ['title'], 'query': 'cat'}}]}}],
     'should': [{'rank_feature': {'boost': 10000, 'field': 'standardized_popularity'}}]
}}}
```

### Unit tests
I added unit tests for  the newly extracted `search_controller.create_search_query` method. I wish it was easy to parametrize these tests, but because we need to check the whole query for each test, it's not easy to do.

While writing the tests, I realized that the `categories` filter doesn't work anymore. In the `search_controller`, we still try to handle it converting the deprecated `categories` param to the current `category` parameter. However, the serializer does not pass the deprecated `categories` to the controller! So, I just removed the deprecated `categories` param from the controller.

Another thing I learned when writing the tests is that `license` and `license_type` filters don't interact at all, so they create 2 completely separate terms filters. We should probably combine them, taking an intersection of the two license lists as the ES query filter (e.g., if `license=by-nc&license_type=commercial` would not return any results with `by-nc` because the commercial filter would exclude them).

## Testing Instructions

Set `DEBUG_SCORES=True` to the `api/.env` file so that the Elasticsearch queries a logged in the `web` Docker container logs. 
Run the API using `just up`.
Go to localhost:50280/admin (use deploy/deploy to log in), and check "Hide content" for one of the providers so that one of the providers is excluded dynamically in the queries (to match the current prod settings).
Try the queries from the PR description and look at the queries logged in Docker logs.
Also compare the queries to the ones on the `main` branch and see that they are different.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -11,7 +11,7 @@ from django.core.cache import cache
 
 from elasticsearch.exceptions import BadRequestError, NotFoundError
 from elasticsearch_dsl import Q, Search
-from elasticsearch_dsl.query import EMPTY_QUERY, Match, Query, SimpleQueryString, Term
+from elasticsearch_dsl.query import EMPTY_QUERY, Match, SimpleQueryString, Term, Terms
 from elasticsearch_dsl.response import Hit, Response
 
 import api.models as models
@@ -34,10 +34,7 @@ PROVIDER = "provider"
 DEEP_PAGINATION_ERROR = "Deep pagination is not allowed."
 QUERY_SPECIAL_CHARACTER_ERROR = "Unescaped special characters are not allowed."
 DEFAULT_BOOST = 10000
-
-
-class RankFeature(Query):
-    name = "rank_feature"
+DEFAULT_SEARCH_FIELDS = ["title", "description", "tags.name"]
 
 
 def _unmasked_query_end(page_size, page):
@@ -235,43 +232,12 @@ def _post_process_results(
     return results[:page_size]
 
 
-def _apply_filter(
-    s: Search,
-    search_params: media_serializers.MediaSearchRequestSerializer,
-    serializer_field: str,
-    es_field: str | None = None,
-    behaviour: Literal["filter", "exclude"] = "filter",
-):
+def get_dynamically_excluded_providers():
     """
-    Parse and apply a filter from the search parameters serializer.
-
-    The parameter key is assumed to have the same name as the corresponding
-    Elasticsearch property. Each parameter value is assumed to be a comma
-    separated list encoded as a string.
-
-    :param s: The ``Search`` instance to apply the filter to
-    :param search_params: the serializer instance containing user input
-    :param serializer_field: the name of the parameter field in ``search_params``
-    :param es_field: the corresponding parameter name in Elasticsearch
-    :param behaviour: whether to accept (``filter``) or reject (``exclude``) the hit
-    :return: the input ``Search`` object with the filters applied
+    Hide data sources from the catalog dynamically.
+    To exclude a provider, set ``filter_content`` to ``True`` in the
+    ``ContentProvider`` model in Django admin.
     """
-
-    if serializer_field in search_params.data:
-        arguments = search_params.data.get(serializer_field)
-        if arguments is None:
-            return s
-        arguments = arguments.split(",")
-        parameter = es_field or serializer_field
-        query = Q("terms", **{parameter: arguments})
-        method = getattr(s, behaviour)
-        return method("bool", should=query)
-
-    return s
-
-
-def _exclude_filtered(s: Search):
-    """Hide data sources from the catalog dynamically."""
 
     filter_cache_key = "filtered_providers"
     filtered_providers = cache.get(key=filter_cache_key)
@@ -282,16 +248,7 @@ def _exclude_filtered(s: Search):
         cache.set(
             key=filter_cache_key, timeout=FILTER_CACHE_TIMEOUT, value=filtered_providers
         )
-    to_exclude = [f["provider_identifier"] for f in filtered_providers]
-    if to_exclude:
-        s = s.exclude("terms", provider=to_exclude)
-    return s
-
-
-def _exclude_sensitive_by_param(s: Search, search_params):
-    if not search_params.validated_data["include_sensitive_results"]:
-        s = s.exclude("term", mature=True)
-    return s
+    return [f["provider_identifier"] for f in filtered_providers]
 
 
 def _resolve_index(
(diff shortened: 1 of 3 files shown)


api/api/controllers/search_controller.py
```diff
@@ -310,88 +267,91 @@ def _resolve_index(
...
         query = _quote_escape(search_params.data["q"])
         base_query_kwargs = {
             "query": query,
-            "fields": search_fields,
+            "fields": DEFAULT_SEARCH_FIELDS,
             "default_operator": "AND",
         }
 
         if '"' in query:
             base_query_kwargs["quote_field_suffix"] = ".exact"
```
sarayourfriend, line 352: Just noticing this, but we don't have any `.exact` subfields. If we switch this to `.raw`, though, I believe it would start to work, because title, description and tags.name do have `.raw` subfields that are not analysed (and therefore not stemmed, so should theoretically be able to service exact match queries).

Separate issue though!

api/api/controllers/search_controller.py
```diff
@@ -310,88 +267,91 @@ def _resolve_index(
...
 
-        s = s.query(
-            "simple_query_string",
-            **base_query_kwargs,
-        )
+        search_queries["must"].append(Q("simple_query_string", **base_query_kwargs))
         # Boost exact matches on the title
         quotes_stripped = query.replace('"', "")
         exact_match_boost = Q(
```
sarayourfriend, line 357: This would also theoretically change with the `.raw` subfields.

api/api/controllers/search_controller.py
```diff
@@ -400,35 +360,78 @@ def search(
...
-        feature_boost = {"standardized_popularity": DEFAULT_BOOST}
-        if search_params.data["unstable__authority"]:
-            feature_boost["authority_boost"] = (
-                search_params.data["unstable__authority_boost"] * DEFAULT_BOOST
-            )
+        search_queries["should"].extend(create_ranking_queries(search_params))
+
+    # If there are no must query clauses, only the results that match
+    # the`should` clause are returned. To avoid this, we add an empty
+    # query clause to the `must` list.
```
sarayourfriend, line 384: Thanks for adding this comment, I hadn't realised this distinction before or what `EMPTY_QUERY` was doing. What's the problem with only matching the should clauses? That should also go into the search algorithm documentation, I think.

```suggestion
    # If there are no `must` query clauses, only the results that match
    # the `should` clause are returned. To avoid this, we add an empty
    # query clause to the `must` list.
```

api/api/controllers/search_controller.py
```diff
@@ -541,7 +544,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
     # Exclude the current item and mature content.
     s = s.query(related_query & ~Term(identifier=uuid) & ~Term(mature=True))
     # Exclude the dynamically disabled sources.
-    s = _exclude_filtered(s)
+    excluded_providers = get_dynamically_excluded_providers()
+    s = s.exclude(Terms(provider=excluded_providers))
```
sarayourfriend, line 548: Shouldn't the excluded providers _always_ be excluded? Would it make sense to add this to the base `create_search_query` function?

sarayourfriend (changes requested):
My only request for change is to also update the search algorithm documentation along with these changes: https://docs.openverse.org/api/reference/search_algorithm.html. In particular, using `filter` instead of `should` is an important note for the metadata filters. Recording an explanation of why that is the correct choice for each of the fields (or at least a general explanation that is clearly application to each one) would be helpful.

One note about the tests: they're quite specific right now, in that they test the specific functions, and I wonder if they would be less fiddly in the future if we make adjustments to how the search query is built by doing a more "integration" style test by checking the actual query sent to Elasticsearch. This is done elsewhere using `pook` to intercept the Elasticsearch client call. Not a request for change here because I'm not confident it's correct, but wanted to suggest it in case you had thoughts about the particular testing approach. I could also see the integration style test being more fiddly because it relies on the ES client's implementation of how it sends the query... on the other hand, that's an important thing we should know if it changes. There are trade-offs, in any case.

api/api/controllers/search_controller.py
obulat: That's interesting! I didn't look into what `.exact` does, just copied the existing code :) I'll open a new issue.

api/api/controllers/search_controller.py
obulat: Yes, they should always be excluded. We shouldn't add any provider to `excluded` providers if its media should not be excluded.  Thank you for suggestion!

api/api/controllers/search_controller.py
obulat: Found the PR that added the `.exact` filter: [Make quoted queries behave as described in the API documentation (return exact matches only)](https://github.com/WordPress/openverse-api/pull/1012)

I'll open an issue for this.

obulat: Issue: [Correctly set up mixing of exact search with stemming(https://github.com/WordPress/openverse/issues/3269)

api/api/controllers/search_controller.py
obulat: I'm not fully certain why, but without `EMPTY_QUERY` in `must` clause, the search without parameters does not match any documents. In theory, all of the documents should match all of the `should` clauses.

api/test/unit/controllers/test_search_controller_search_query.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def test_create_search_query_q_search_with_filters(image_media_type_config):
+    serializer = image_media_type_config.search_request_serializer(
+        data={
+            "q": "cat",
+            "license": "by-nc",
+            "aspect_ratio": "wide",
+            # this is a deprecated param, and it doesn't work because it doesn't exist in the serializer
+            "categories": "digitized_artwork",
```
AetherUnbound, line 109: Are we including it here just to showcase that?

AetherUnbound (approved):
This is excellent! No blocking comments, just some notes and thoughts 🙂 it's very cool to see this so improved!

api/test/unit/controllers/test_search_controller_search_query.py
```diff
@@ -0,0 +1,240 @@
+from django.core.cache import cache
+
+import pytest
+
+from api.controllers import search_controller
+
+
+pytestmark = pytest.mark.django_db
```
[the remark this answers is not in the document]
sarayourfriend: `pytestmark` is a pytest feature: https://docs.pytest.org/en/7.4.x/reference/reference.html#globalvar-pytestmark

It has to have this name or it won't have any effect.

sarayourfriend (approved):
LGTM!

api/test/unit/controllers/test_search_controller_search_query.py
obulat: I included it in the PR to showcase this to the reviewers, but I should probably remove it from the test that will be merged.

obulat:
> One note about the tests: they're quite specific right now, in that they test the specific functions, and I wonder if they would be less fiddly in the future if we make adjustments to how the search query is built by doing a more "integration" style test by checking the actual query sent to Elasticsearch. This is done elsewhere using `pook` to intercept the Elasticsearch client call. Not a request for change here because I'm not confident it's correct, but wanted to suggest it in case you had thoughts about the particular testing approach. I could also see the integration style test being more fiddly because it relies on the ES client's implementation of how it sends the query... on the other hand, that's an important thing we should know if it changes. There are trade-offs, in any case.

Thank you for the suggestion, @sarayourfriend. I will leave the tests as they are here, but checking the ES queries seems valuable. Maybe we should use both?

api/test/unit/controllers/test_search_controller_search_query.py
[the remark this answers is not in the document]
AetherUnbound: OH, TIL! That's super cool 😮

Issue #2910, opened 2023-08-28
Audio components improvement


fcoveram:
## Problem

From the audio control update requested in #2602, the following components need to be updated to incorporate the button in a clean way.

- Box (large, and medium)
- Row (large, medium, and small)
- Global

## Proposal

All components and their variants mentioned above have new spacing, size, and style values. Before diving into each, here is a set of all pages involved in the component update.

### Search results. All content

![All content search results in all breakpoints](https://github.com/WordPress/openverse/assets/895819/3861b8dc-268c-422f-8637-d6091c300153)

### Search results. Audio

![Audio search results in all breakpoints](https://github.com/WordPress/openverse/assets/895819/d6ed4ce3-fcd8-4bf4-85ad-3f6b40abb00a)

### Single result page

![Single result page mockups in all breakpoints](https://github.com/WordPress/openverse/assets/895819/c818f5ae-6984-4786-ae83-5d8abec9f6df)

### Audio box

The new box audio component has the following main changes:
- The component was reduced to two variants: Large and Small. The medium size was not needed for Figma mockups.
- The audio control is `size: small` .
- Both large and small variants allow users to control the audio with the current play, pause, and repeat actions.
- Only large variant shows the wave area. 
- All texts styles were updated to be consistent with other library changes.

| Small | Large |
|---|---|
| <img src="https://github.com/WordPress/openverse/assets/895819/9a8bb8d6-b75e-4281-928a-359e8a3646ae" alt="Small variant of audio box" width="230"> | <img src="https://github.com/WordPress/openverse/assets/895819/1b8b1ab0-d494-4aa5-9186-9aee35160da4" alt="Large variant of audio box" width="230"> |

### Audio row

For the main audio component, the following changes were applied:

**Size: Large**

- The current variant named “Medium” is now “Large”. This change comes from the variant used per breakpoint.
- The audio control is `size: large`.
- Cover image is now `64px` to match the audio control button
- All texts styles were updated to be consistent with other library changes.
- Spacings in the audio text info were updated.

![Large variant of audio row component](https://github.com/WordPress/openverse/assets/895819/ecb70726-01f9-4489-8e17-f78c2a60f956)

**Size: Medium**

- The current variant named “Large” is now “Medium”. This change comes from the variant used per breakpoint.
- The audio control is `size: medium`.
- Cover image is the height formed by the sum of text info and wave area. In Figma it was set as `95px`, but this should be revisited in code to meet a visual balance.
[...]

PR #3265, opened 2023-10-27
Update the global audio player

obulat:
## Fixes

Related to #2910 by @fcoveram

## Description

This PR updates the Global audio player to match the design requirements in #2910.

Screenshot showing the previous version of the global audio component compared to the updated version

It also fixes the problem we had with global audio player: it sometimes appeared on the home page, but didn't appear on the single result page. This was caused by the updates in the layouts, when I accidentally added the component into the wrong layout.
The old no longer used `VCloseButton` component is also removed.

## Testing Instructions

Open the app on a mobile-width screen and search for something. When you click on Play button, you should be able to see the global player. You can seek, pause or play using it. If you click Play on a different audio, the audio should be updated. If you click on the audio title, the single result should open, and the global player should stay on the screen with the same audio playing.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


fcoveram (changes requested):
The behavior described works correctly.

The only wrong style I see is the border color, it should be 20% opacity of `dark-charcoal` rather than `dark-charcoal-20`. Secondly, the design has a border-radius of 2px, but I think it looks better in 4px.

frontend/src/components/VGlobalAudioSection/VGlobalAudioSection.vue
```diff
@@ -1,14 +1,19 @@
...
-      v-if="audio"
-      class="!absolute end-0 top-0 z-30"
-      variant="transparent-dark"
-      size="large"
-      :label="$t('audioTrack.close')"
-      @close="handleClose"
-    />
+  <div
+    class="global-audio sticky z-global-audio rounded-sm bg-white sm:hidden"
+    :class="{ 'mx-2 mb-2 border border-dark-charcoal-20': !!audio }"
```
obulat, line 4:
```suggestion
    :class="{ 'mx-2 mb-2 border border-dark-charcoal border-opacity-20': !!audio }"
```

frontend/src/components/VGlobalAudioSection/VGlobalAudioSection.vue
```diff
@@ -1,14 +1,19 @@
...
-    <VCloseButton
-      v-if="audio"
-      class="!absolute end-0 top-0 z-30"
-      variant="transparent-dark"
-      size="large"
-      :label="$t('audioTrack.close')"
-      @close="handleClose"
-    />
+  <div
+    class="global-audio sticky z-global-audio rounded-sm bg-white sm:hidden"
```
obulat, line 3:
```suggestion
    class="global-audio sticky z-global-audio rounded bg-white sm:hidden"
```

fcoveram (commented):
I see the new border color, but I think it's set outside rather than inside because cover image and the white area on the right side have the same shade. Besides, the component corners look cut off.

Here is what I see

Global audio player with two areas highlighted

Global audio player and chrome inspector

And here is the design

Mockup of global audio player with two areas highlighted

I tried deleting the branch and running the PR again, but I see the same.

obulat:
@fcoveram, there were some child elements that were covering the border and making it look not `rounded`. I've updated the code:
Screenshot 2023-11-01 at 2 11 28 PM

The corners of the `VAudioControl` are the same as the global audio layout (4px) but it seems that they look a little bit more rounded so that there's a tiny gap between the button and the border. I tried making the button rounded corner 2px, but it didn't look right either. What do you think, @fcoveram ?

AetherUnbound:
@obulat I'm seeing an issue where when I play an audio from the results list, then click on a single result, the audio stops playing. Is that intended? Otherwise this works as expected!

obulat:
> @obulat I'm seeing an issue where when I play an audio from the results list, then click on a single result, the audio stops playing. Is that intended? Otherwise this works as expected!

The expected behavior was to stop the playback if you are opening a single result that is different from the audio that is currently playing. And if you start playing an audio and then go on to **its** single result page, the playback continues.

fcoveram (commented):
> The corners of the VAudioControl are the same as the global audio layout (4px) but it seems that they look a little bit more rounded so that there's a tiny gap between the button and the border.

The change is sublte. Let's keep the `The VAudioControl` corner as 4px is fine. Although, I did notice that effect in the cover image, especially in white ones. I think it inherits the brand yellow from the dot pattern. If that change requires too much time, feel free to dismiss it. It's also very subtle.

Cover image details in Global audio player

Regarding opacity in borders, I think the change does not work as expected and keeps the line outside the component area. I revisited Tailwind documentation and saw that "ring width" and "ring color" are used for these inner borders.

You can see how it works in Tailwind's homepage, in the Projects component example

Tailwind's homepage screenshot

AetherUnbound (approved):
Looks good to me and works as expected!

frontend/src/types/button.ts
```diff
@@ -36,5 +36,11 @@ export type ButtonSize = (typeof buttonSizes)[number]
 export const buttonTypes = ["button", "submit", "reset"] as const
 export type ButtonType = (typeof buttonTypes)[number]
 
-export const buttonConnections = ["start", "end", "none", "all"] as const
+export const buttonConnections = [
```
dhruvkb, line 39: Technically the `buttonConnections` array should only contain 4 entries (one for each edge) and the `connections` prop on the `VButton` component should accept an array of connections. That would also remove the need for elements like `"none"` and `"all"`.

It can be a complex change so I won't block on it, especially because this works.

dhruvkb (approved):
The changes look good to me. The new design is great. I have a suggestion but if its implementation is too big of a change for the PR, feel free to skip.

frontend/src/types/button.ts
obulat: Having an array for `buttonConnections` would be much cleaner! I'll see if I can implement this.

fcoveram (approved):
It looks nice. Thanks for the great work ✨ 🚀

PR #3303, opened 2023-11-01
Create filtered index before promoting primary index during data refresh

### LINKED ISSUE #2981: Wait to change all ES aliases until filtered index creation is complete
## Problem

Presently, the primary media index alias is changed from the old index to the new one during a data refresh _before_ the filtered index creation has run. This means that both the reindex initiated by the filtered index DAG _and_ the production API are working against the same index. We believe that this may be impacting production performance (see #2975).

## Description

It may be best to wait until the filtered index creation is complete to change the primary media alias. This would prevent the API from using an index which is also being used as the source of a reindex vis-a-vis the creation of the filtered index. The goal with this change would be to allow a higher throttle & throughput of the filtered index creation without affecting the API.

I believe this could be done by bringing the filtered index creation steps into the media refresh DAGs (currently they exist as a separate, triggered DAG). This could actually make orchestration _easier_ since it would mean that we could remove the "check filtered index creation" step. That would remove our ability to trigger the filtered index creation on its own however, so perhaps it might be better to add a step to the primary DAG wait for the triggered index DAG to complete. This may require some changes to the logic that checks for running DAGs, since the data refresh itself will be running until the filtered index creation step completes.

## Additional context

This came up alongside, but is not directly related to, #2980

stacimc:
## Fixes

Fixes #2981 by @AetherUnbound 

## Description

This PR updates the data refresh DAG to run the steps to create the filtered index _before_ promoting the primary index . The goal is to prevent API instability during data refreshes, potentially caused by the API using an ES index that is also being used as the source of a reindex (to create the filtered index).

To accomplish this, I refactored the `create_filtered_index` DAG factory to add a new factory that generates a `create_filtered_index` TaskGroup and a separate `promote_filtered_index` TaskGroup (that does the index readiness check, promotion, and deletion). The data refresh DAGs now, rather than triggering the separate `create_filtered_index` DAGs, use this factory method to create the taskgroups and add them directly into its own flow:

Screenshot 2023-11-01 at 2 23 31 PM

_Collapsed view, showing that we complete `ingest_upstream`, wait for the index readiness check on the primary index, then **create the filtered index before promoting anything**, and only then do we promote the primary index followed by the filtered index._

Screenshot 2023-11-01 at 2 23 52 PM

_Expanded view of the filtered index creation_

Screenshot 2023-11-01 at 2 24 06 PM

_Expanded view of the filtered index promotion_

We keep the separate filtered index creation DAGs as these are still useful for running manually when testing filtered indices or updating the sensitive terms list. These DAGs now just reuse the same task groups and add the concurrency check. Because the DAG is _only_ run manually, I removed the `force` parameter to break through the concurrency check.

Screenshot 2023-11-01 at 2 24 38 PM

_New view of the create filtered index DAG_

## Testing Instructions

Test the data refreshes locally. Also test the filtered index DAGs: I tried running the DAG locally with `origin_index_suffix` empty (so it would use the current index) and `destination_index_suffix` set to `foo`. After the DAG completed I checked Elasticvue to see that I now had an `audio-foo-filtered` index replacing the old filtered index :)

Test that the concurrency checks work to prevent either a data refresh or filtered index dag from running at the same time as its counterpart. You can do this by triggering both at once in the shell:

```
$ just catalog/shell
> airflow dags trigger image_data_refresh && airflow dags trigger create_filtered_image_index
```

Try swapping the order of the two commands as well to test the opposite direction. (If the filtered index Dag starts second, it should fail immediately. If the data refresh starts second, it should wait for the filtered index dag. This behavior was not changed.)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -5,21 +5,27 @@
...
-creation if an urgent change to the sensitive terms calls for an immediate
-recreation of the filtered indexes.
+The DAGs generated by the ``create_filtered_index_creation_dag`` can be used
+to manually run the filtered index creation and promotion steps described above in
+isolation from the rest of the data refresh. These DAGs also include checks to ensure
+that race conditions with the data refresh DAGs are not encountered (see ``Race
+conditions`` section below).
+
+The DAGs generated in this module are on a `None` schedule and are only triggered
+manually. This is primarily useful in two cases: for testing changes to the filtered
```
obulat, line 26: Nice to have this functionality!

obulat (approved):
I tested following the instructions, and the steps work well. Thank you for excellent documentation, as always, @stacimc!

catalog/dags/data_refresh/create_filtered_index.py
```diff
@@ -0,0 +1,133 @@
...
+
+        do_create, await_create = create_and_populate_filtered_index(
+            origin_index_suffix=origin_index_suffix,
+            destination_index_suffix=final_destination_index_suffix,
+        )
+
+        # Determine the destination index suffix and get the current index. The current
+        # index retrieval has to happen prior to any of the index creation steps to
+        # ensure the appropriate index information is retrieved.
+        [get_current_index_if_exists, destination_index_suffix]
```
AetherUnbound, line 92: This step looks like a no-op, do you know if it's necessary?

catalog/dags/data_refresh/create_filtered_index_dag.py
```diff
@@ -193,64 +141,25 @@ def point_alias(destination_index_suffix: str) -> TaskGroup:
...
-            action="DELETE_INDEX",
-            model=data_refresh.media_type,
-            data={
-                "index_suffix": XCOM_PULL_TEMPLATE.format(
-                    get_current_index_if_exists.task_id, "return_value"
-                ),
-            },
+        prevent_concurrency = prevent_concurrency_with_data_refresh.override(
+            retries=0
+        )()
```
AetherUnbound, line 146: I _might_ be wrong, but I think if this isn't called with any arguments, then it doesn't need to be called at all? The bitflow definition below should ensure it's actually run in the course of the DAG (_I think_).
```suggestion
        prevent_concurrency = prevent_concurrency_with_data_refresh.override(
            retries=0
        )
```

catalog/dags/data_refresh/create_filtered_index.py
```diff
@@ -0,0 +1,133 @@
...
+    destination_index_suffix: str | None,
+) -> tuple[TaskGroup, TaskGroup]:
+    """
+    Create the TaskGroups that performs filtered index creation and promotion for
+    the given DataRefresh.
+    """
+    media_type = data_refresh.media_type
+    target_alias = f"{media_type}-filtered"
+
+    def create_and_populate_filtered_index(
```
AetherUnbound, line 41: I know these functions were within the DAG function in the previous implementation, but it does feel a little odd to have them be a pure function within a pure function  in the new setup. Do you think it makes sense to try and extract these out so they can be functions at the top level that then get called within `create_filtered_index_creation_task_groups` rather than defined within it? Or is it easier to keep as-is?

AetherUnbound (approved):
This is excellent 🤩 I tested the DAGs locally and they run _great_, I'm glad to have the functionality all encapsulated within one DAG now rather than have it spread across too, while maintaining the filtered index DAG in case we need it. I have a few comments but no blocking objections - awesome work!

catalog/dags/data_refresh/create_filtered_index_dag.py
stacimc: This caused errors for me at DAG parsing in testing.

catalog/dags/data_refresh/create_filtered_index.py
stacimc: Totally, and in fact I'll move the task definitions around a little to be more straightforward while I'm at it.

catalog/dags/data_refresh/create_filtered_index.py
stacimc: We certainly can, it just means we have to pass more params through to those methods.

PR #3307, opened 2023-11-02
Simplify related query to remove nesting and make more performant

### LINKED ISSUE #3306: Excluded providers clause in the related query is inefficient
## Description

The related query's excluded provider clause nests the query in `filter -> bool -> must_not`, which is unnecessary.
An example of the current related query with excluded providers:

Current related query
```
{
  "from": 0, "size": 20,
  "query": {
    "bool": {
      "filter": [{"bool": {"must_not": [{"terms": {"provider": ["thorvaldsensmuseum", "eol", "500px", "behance", "iha", "deviantart", "mccordmuseum"], "boost": 1.0}}], "boost": 1.0}}],
      "must_not": [{"term": {"identifier": {"value": "464a3673-6438-48e1-8f55-67ebefe6a9cc"}}}, {"term": {"mature": {"value": true}}}],
      "should": [
        {"match": {"title": {"query": "Dandelion in Green Grass"}}},
        {"simple_query_string": {"query": "dandelion | grass | bright | closeup | dandelion | delicate | downy | flora | flower | garden", "fields": ["tags.name^1.0"]}}
      ],
      "minimum_should_match": "1", "boost": 1.0
    }
  }
}
```

## Solution

This query should be simplified, similar to the way the search query was simplified:

```
{'query': {'bool': {'must_not': [{'terms': {'provider': ['stocksnap']}},
                                 {'term': {'mature': True}},
                                 {'term': {'identifier': '027e86f8-ab89-4634-84ed-fb17bfdb0342'}}],
                    'should': [{'match': {'title': 'Big CAT on the Beach'}},
                               {'simple_query_string': {'fields': ['tags.name'],
                                                        'query': '3290 | adelaide | beach | brand | cat | caterpillar | digger | dunes | earthmover | excavator'}}]}}}
```

## Additional context

#3261

obulat:
## Fixes

Fixes #3306 by @obulat

## Description

This PR refactors the way related query is created, using more low-level queries to make them less nested.

The main changes for this PR are within the `elasticsearch/related.py` file. All other changes are just moving the functions to a different file. The `related` function is moved from `search_controller` module to a separate file within `elsticsearch` folder, and some other common functions for pagination to `elasticsearch/helpers.py` file to make the files easier to read.

### Converting the simple search query to the terms query

After opening this PR, I also checked the queries logged by slowlog: all of the queries logged in the recent hours were related queries. 
Using the `simple_query_string` for the list of tags is not performant. This PR converts the query to use the terms query for tags. It will mean that the tags will not match if the form is different (so, `cat` will not match `cats`), but since we are trying to match all of the tags, I think we should still have enough matches. 

## Testing Instructions

Go to the `/admin` endpoint and set `hide content` for either Stocksnap or Flickr to true (in ContentModel).
Add logging to the related endpoint (`s.to_dict()` to view the ES query), and check that the query sent to ES when the related endpoint is requested does not have the nesting described in the issue.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from elasticsearch_dsl import Search
-from elasticsearch_dsl.query import Match, Q, SimpleQueryString, Term
+from elasticsearch_dsl.query import Match, Q, Term
 from elasticsearch_dsl.response import Hit
 
 from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
@@ -47,11 +47,10 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
             related_query["should"].append(Match(title=title))
 
         # Match related using tags, if the item has any.
+        # Only use the first 10 tags
         if tags:
-            # Only use the first 10 tags
-            tags = " | ".join([tag.name for tag in tags[:10]])
-            tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
-            related_query["should"].append(tags_query)
+            tags = [tag["name"] for tag in tags[:10]]
+            related_query["should"].append(Q("terms", tags__name=tags))
 
     # Exclude the dynamically disabled sources.
     if excluded_providers_query := get_excluded_providers_query():
@@ -69,6 +68,6 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
     start, end = get_query_slice(s, page_size, page, filter_dead)
     s = s[start:end]
 
-    response = get_es_response(s, "related_media")
+    response = get_es_response(s, es_query="related_media")
     results = _post_process_results(s, start, end, page_size, response, filter_dead)
     return results or []
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -1,0 +1,74 @@
+from __future__ import annotations
+
+from elasticsearch_dsl import Search
+from elasticsearch_dsl.query import Match, Q, SimpleQueryString, Term
+from elasticsearch_dsl.response import Hit
+
+from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
+from api.controllers.search_controller import (
+    _post_process_results,
+    get_excluded_providers_query,
+)
+
+
+def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
+    """
+    Given a UUID, finds 10 related search results based on title and tags.
+
+    Uses Match query for title or SimpleQueryString for tags.
+    If the item has no title and no tags, returns items by the same creator.
+    If the item has no title, no tags or no creator, returns empty list.
+
+    :param uuid: The UUID of the item to find related results for.
+    :param index: The Elasticsearch index to search (e.g. 'image')
+    :param filter_dead: Whether dead links should be removed.
+    :return: List of related results.
+    """
+
+    # Search the default index for the item itself as it might be sensitive.
+    item_search = Search(index=index)
+    item_hit = item_search.query(Term(identifier=uuid)).execute().hits[0]
+
+    # Match related using title.
+    title = getattr(item_hit, "title", None)
+    tags = getattr(item_hit, "tags", None)
+    creator = getattr(item_hit, "creator", None)
+
+    related_query = {"must_not": [], "must": [], "should": []}
+
+    if not title and not tags:
+        if not creator:
+            return []
+        else:
+            # Only use `creator` query if there are no `title` and `tags`
+            related_query["should"].append(Term(creator=creator))
+    else:
+        if title:
+            related_query["should"].append(Match(title=title))
+
+        # Match related using tags, if the item has any.
+        if tags:
+            # Only use the first 10 tags
+            tags = " | ".join([tag.name for tag in tags[:10]])
+            tags_query = SimpleQueryString(fields=["tags.name"], query=tags)
+            related_query["should"].append(tags_query)
+
+    # Exclude the dynamically disabled sources.
+    if excluded_providers_query := get_excluded_providers_query():
+        related_query["must_not"].append(excluded_providers_query)
+    # Exclude the current item and mature content.
+    related_query["must_not"].extend(
+        [Q("term", mature=True), Q("term", identifier=uuid)]
+    )
+
+    # Search the filtered index for related items.
+    s = Search(index=f"{index}-filtered")
+    s = s.query("bool", **related_query)
+
+    page, page_size = 1, 10
+    start, end = get_query_slice(s, page_size, page, filter_dead)
+    s = s[start:end]
+
+    response = get_es_response(s, "related_media")
+    results = _post_process_results(s, start, end, page_size, response, filter_dead)
+    return results or []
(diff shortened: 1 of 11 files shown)


api/test/unit/controllers/test_search_controller.py
```diff
@@ -15,6 +15,7 @@
 from django_redis import get_redis_connection
 from elasticsearch_dsl import Search
 
+import api.controllers.elasticsearch.helpers
```
sarayourfriend, line 18: Nit: using an alias for this would help simplify the code in the tests a bit:

```suggestion
import api.controllers.elasticsearch.helpers as es_helpers
```

Though the alternate form reads more "Python-y" (familiar?) to me:

```suggestion
from api.controllers.elasticsearch import helpers as es_helpers
```


```

**Review by sarayourfriend — CHANGES_REQUESTED** [review]
@obulat could you add some new unit tests, preferrably ones that run against the previous implementation and then still pass with the new one? The only test for related I can find is that it returns a 200 (https://github.com//WordPress/openverse/blob/bc7b2a7f6ee32fc3ca367baae7681abc41ef7f79/api/test/media_integration.py#L152-L155) but it doesn't confirm anything about the query actually doing what we want it to do (for example, by testing that the query result is the expected one).

As it stands, this might work fine (I haven't tested it locally yet, the code looks fine) but I'm uncomfortable approving it without any tests that would verify the behaviour of the query is the same (or at least, changing in an expected way).

Also: I'd like to have a deployment plan for this that includes setting up a dashboard to monitor the changes in response times for the related endpoint. I've saved a Logs Insights query in CloudWatch called "parsed requests" (something along those lines) that exposes an `isRelated` flag, which we could use to create a query like this:

```
# ... saved query with parsing and creation if `isRelated` field
| filter isRelated
| stats avg(upstream_response_time) by bin(5m)
```

I've started this work for https://github.com/WordPress/openverse-infrastructure/issues/651 but likely won't be able to get to it until my Friday, but would seem to be a prerequisite to me for any significant changes we make to the related endpoint.

File: api/api/controllers/elasticsearch/related.py [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+            related_query["should"].append(Term(creator=creator))
+    else:
+        if title:
+            related_query["should"].append(Match(title=title))
+
+        # Match related using tags, if the item has any.
+        if tags:
+            tags = [tag["name"] for tag in tags[:10]]
+            # Only use the first 10 tags
+            related_query["should"].append(Q("terms", tags__name=tags))
```
AetherUnbound, line 53: These comments seem a little out of order:
```suggestion
        # Match related using only the first 10 tags, if the item has any.
        if tags:
            tags = [tag["name"] for tag in tags[:10]]
            related_query["should"].append(Q("terms", tags__name=tags))
```

api/api/controllers/elasticsearch/related.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Exclude the dynamically disabled sources.
+    if excluded_providers_query := get_excluded_providers_query():
```
AetherUnbound, line 56: Love that we have this now!

AetherUnbound (commented):
Agreed with @sarayourfriend - I like the code changes and they make sense, having tests for this would be useful and having a dashboard for monitoring production when deployed feels necessary. I appreciate how much more readable the search building code is with all the changes you've been making @obulat!

sarayourfriend:
Regarding the dashboard: I actually have a few minutes now I think I can have the basic thing up and we can iterate it after deployment if needed. The data will be there before/after because it's just derived from the logs anyway, so the dashboard doesn't need to be perfect at deployment time, just comprehensible enough that we can monitor the deployment and make sure this change doesn't somehow negatively effect the response times (which, to be clear, I don't think will be the case, just want to make sure we can move forward with confidence!).

sarayourfriend:
Update on the dashboard: I implemented it earlier today, and it's available at a link in the issue: https://github.com/WordPress/openverse-infrastructure/issues/651

With unit tests this will be good-to-go from my perspective :grin:

obulat:
Thank you so much for setting up the dashboard, @sarayourfriend!

I've updated this PR so that the first commit is adding the test for `related_media`: https://github.com/WordPress/openverse/pull/3307/commits/b312c5eecc6781d9f7fbc47a634d82e7945bfbea
The following commits add the changes, and the last commit updates the related test.

I've added assertions for the ES query to check that the correct query is set. It is not very consistent with other tests, because instead of using `assert`, it uses the `.json` in `pook` request setup.

sarayourfriend:
> I've added assertions for the ES query to check that the correct query is set. It is not very consistent with other tests, because instead of using assert, it uses the .json in pook request setup.

For what it's worth, I think that's fine, there are other ES tests that do this:

https://github.com//WordPress/openverse/blob/95b59110eff953f7d30d9b9444791ae2e0dc6d68/api/test/unit/controllers/test_search_controller.py#L584-L591

> I wanted to add a check that the results' title and tags have common words, but realized that it's not possible with this implementation since we use mocked results instead of the "related" results. I will see if I can use the integrated tests for this.

You're right, I think it's only at the integration test level that we can actually test this in a meaningful way.

obulat:
> For what it's worth, I think that's fine, there are other ES tests that do this:

I think this example is a little bit different from what I did in the test here. In the linked example, there's only 1 `.json`, which is used to set the response. In this PR, however, `.json` is used twice: first for making sure that the _request body_ matches the shape we expect, and second to mock the response:
```
    mock_related = (
        pook.post(es_filtered_index_endpoint)
 >      .json(es_related_query)  # Testing that ES query is correct 
        .times(1)
        .reply(200)
        .header("x-elastic-product", "Elasticsearch")
 >      .json(mock_es_response)
        .mock
    )
```

sarayourfriend:
Sure, but the example I showed just uses a regex to test the body of the request. There's no real difference, the end result is pook either compares the body as strings using `==` (`.json` just tells pook to serialise the argument to a JSON string, it then compares the bodies as strings), or executes a regex test. In other words, the tests are really the same, the example I shared just tests a _part_ of the body, whereas passing `.json` on the pook request tests the whole body.

Anyway, it's all good, both are perfectly fine approaches as far as I'm concerned and well suited for different testing needs.

obulat:
I added the check that there are at least 1 word in common in the related results with the main item (in the title and tags) in https://github.com/WordPress/openverse/pull/3307/commits/8a5e3e4e0b54c991155497816c184716bdb62b9d

sarayourfriend (approved):
LGTM! Excited to see how/if the endpoint timings respond :crossed_fingers:

AetherUnbound (approved):
Thanks for adding the new tests, I'm excited to try this out!

api/api/controllers/elasticsearch/related.py
```diff
@@ -0,0 +1,73 @@
...
+            related_query["should"].append(Term(creator=creator))
+    else:
+        if title:
+            related_query["should"].append(Match(title=title))
+
+        # Match related using tags, if the item has any.
+        # Only use the first 10 tags
+        if tags:
+            tags = [tag["name"] for tag in tags[:10]]
+            related_query["should"].append(Q("terms", tags__name=tags))
```
sarayourfriend, line 53: Note for follow up @obulat:

```suggestion
related_query["should"].append(Q("terms", tags__name__keyword=tags))
```

obulat: See #3346

sarayourfriend:
Just wanted to share an update. We deployed this an hour ago and the related image response times are _dramaticly_ improved: 
image

Nice work @obulat!

PR #3312, opened 2023-11-03
Update renovate config to handle all dependency types and group upgrades

### LINKED ISSUE #3304: Both dependabot and renovate open dependency update PRs in the monorepo
## Description

Renovate and dependabot both individually open dependency update PRs. I'm not sure which of these we want to stick with, but we should pick one and turn the other off.

Related to [this issue in the infrastructure repository (private repo) about self hosting renovate](https://github.com/WordPress/openverse-infrastructure/issues/484). With the new Ansible setup, this should be a lot easier to manage. If we want to use Renovate for whatever reason we can spin up the self-hosted version relatively easily now, and then switch to the free version later once we don't need the `.pnpmfile.cjs`, which the free version does not support.

## Additional context

Dependabot PR: https://github.com/WordPress/openverse/pull/3283
Renovate PR: https://github.com/WordPress/openverse/pull/3278

Both of those are for the same dependency.

sarayourfriend:
## Fixes

Fixes #3304 by @sarayourfriend
Closes https://github.com/WordPress/openverse-infrastructure/issues/484 by @sarayourfriend

## Description

Updates the existing renovate configuration in the following ways:

- Match dependabot stack and tech labeling
- Group dependencies that can be grouped (note: we can continue iterating on this, there are probably lots of good opportunities for this in the frontend especially, I just picked some obvious ones)
- Update configuration options that were in an older format
- Add a pre-commit hook to validate the renovate configuration
- Remove the dependabot configuration (also requires disabling it in the repo settings, I think)

**This is _one_ approach we could take**. The other is, of course, to continue using dependabot. It has rudimentary support for pnpm and grouped dependencies, both added very recently, but it's a bit more of a chore to configure (lots of repeated labeling rules, for example) and tends to take a very long time to support new features (e.g., pnpm, grouped dependencies, sensible github action matching), if it ever gets them at all.

## Testing Instructions

Not entirely sure. I suppose we can run the renovate workflow from this branch and see what happens.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (approved):
Since renovate has had a longer support for groups and pnpm, I agree with going forward with renovate. Thank you for fixing the labelling (renovate PRs didn't have all the necessary labels before) and setting everything up!

.github/workflows/renovate.yml
```diff
@@ -1,18 +1,95 @@
...
+          # are different than the ones given after the cache is restored. We have to
+          # change ownership to solve this. We also need to have correct permissions in
+          # the entire /tmp/renovate tree, not just the section with the repo cache.
+          sudo chown -R runneradmin:root /tmp/renovate/
+          ls -R "$cache_dir"
+
       - uses: renovatebot/github-action@v39.1.1
         with:
-          configurationFile: .github/renovate.json
+          configurationFile: .github/renovate-config.js
```
dhruvkb, line 72: Sorry if I'm missing something here but the config file is still named `renovate.json` so I don't get this change.

.github/workflows/renovate.yml
```diff
@@ -1,18 +1,95 @@
...
 jobs:
   renovate:
     runs-on: ubuntu-latest
     if: github.repository == 'WordPress/openverse'
     steps:
       - uses: actions/checkout@v4
+
+      # This third party action allows you to download the cache artifact from different workflow runs
+      # Note that actions/cache doesn't work well because the cache key would need to be computed from
+      # a file within the cache, meaning there would never be any data to restore. With other keys, the
```
dhruvkb, line 37: > actions/cache doesn't work well because the cache key would need to be computed from a file within the cache

That's not necessary, the cache key can be set to a constant value as well. I think this would work equally well:

```yaml
- uses: actions/cache@v3
  if: github.event.inputs.repoCache != 'disabled'
  continue-on-error: true
  with:
    path: cache-download
    key: ${{ env.cache_key }}
```

dhruvkb (changes requested):
Although I prefer personally prefer Dependabot over Renovate, Renovate has clear advantages in our case and settling on one tool instead of two makes more sense. Thanks for this. The addition of the pre-commit hook is also a great touch, thanks!

I do have some questions and suggestions though.

.github/workflows/renovate.yml
sarayourfriend: Oops! It should be the json file. I experimented with a js file at first but it wasn't worth the confusion in the end, I think.
```suggestion
          configurationFile: .github/renovate.json
```

.github/workflows/renovate.yml
sarayourfriend: I'll take a deeper look! I copied this from the Renovate github-action README: https://github.com/renovatebot/github-action

They must be working off old information! I'll update this, and maybe open an issue in their repository as well for them to update the example. Glad to have your in-depth knowledge of GitHub actions, Dhruv!

sarayourfriend: Updated in the latest commit to use your approach, Dhruv.

dhruvkb (approved):

PR #3315, opened 2023-11-05
Update TypeScript to v5.2.2

obulat:
## Fixes

No issue for this update

## Description

This PR updates the TypeScript version to the latest stable, v5.2.2.

I removed `typeRoots` setting from `tsconfig` because with it, type checks were failing with 
```
 error TS2688: Cannot find type definition file for '@nuxt/types
│   The file is in the program because
│     Entry point of type library '@nuxt/types' specified in compilerOptions
│ error TS2688: Cannot find type definition file for '@nuxtjs/i18n
│   The file is in the program because
│     Entry point of type library '@nuxtjs/i18n' specified in compilerOptions
│ error TS2688: Cannot find type definition file for '@pinia/nuxt
│   The file is in the program because
│     Entry point of type library '@pinia/nuxt' specified in compilerOptions
│ error TS2688: Cannot find type definition file for '@sentry/core
│   The file is in the program because
│     Entry point of type library '@sentry/core' specified in compilerOptions
│ error TS2688: Cannot find type definition file for '@types/jest
│   The file is in the program because
│     Entry point of type library '@types/jest' specified in compilerOptions
```
This happens if types are specified in `types` parameter in tsconfig, but not included in the `typeRoots` [^1].

@sarayourfriend, I'm not sure what effect excluding `typeRoots` will have.
I found a very old note about this setting being added for backward compatibility with typings: https://github.com/Microsoft/TypeScript/issues/22217#issuecomment-370019383
And here are the docs about the change that caused the new behavior during resolution, in TS 5.1: https://devblogs.microsoft.com/typescript/announcing-typescript-5-1-rc/#typeroots-are-consulted-in-module-resolution

## Testing Instructions

The CI should pass, both VS Code and WebStorm should not have additional warnings about TS or unresolved files in JS.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

[^1]: [Related test file in TS repository](https://github.com/microsoft/TypeScript/blob/d2fdf851cc630d812b7b336a43cc1f363ba25083/tests/baselines/reference/tscWatch/programUpdates/types-should-not-load-from-config-file-path-if-config-exists-but-does-not-specifies-typeRoots.js#L38)

### DIFF
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -15,7 +15,6 @@
       "@pinia/nuxt",
       "@sentry/core"
     ],
-    "typeRoots": ["./typings", "./node_modules/@types"],
     "paths": {
       "~/*": ["./src/*"],
       "~~/*": ["./*"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -91,7 +91,7 @@ repos:
         additional_dependencies:
           - prettier@2.8.3
           - prettier-plugin-tailwindcss@0.3.0
-          - typescript@4.9.5
+          - typescript@5.2.2
 
   - repo: https://github.com/koalaman/shellcheck-precommit
     rev: v0.9.0
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -137,7 +137,7 @@
     "tailwindcss": "^3.3.5",
     "talkback": "^3.0.1",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.5",
+    "typescript": "^5.2.2",
     "vue-demi": "^0.14.6",
     "vue-i18n-extract": "^2.0.7",
     "vue-jest": "^3.0.7",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -144,7 +144,7 @@
     "vue-loader": "^15.10.0",
     "vue-server-renderer": "^2.7.14",
     "vue-template-compiler": "^2.7.14",
-    "vue-tsc": "1.2.0",
+    "vue-tsc": "1.8.22",
     "webpack": "^4.46.0"
   },
   "browserslist": [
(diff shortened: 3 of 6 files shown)


frontend/tsconfig.json
```diff
@@ -15,7 +15,6 @@
       "@pinia/nuxt",
       "@sentry/core"
     ],
-    "typeRoots": ["./typings", "./node_modules/@types"],
```
sarayourfriend, line 18: Removing the reference to typings would (theoretically) exclude our custom type definitions in that directory from being used.

I actually noticed a similar thing locally, but couldn't reproduce it when running the type check, and was able to fix it by adding `../node_modules/@types` (the monorepo root) as well:

```suggestion
    "typeRoots": ["./typings", "./node_modules/@types", "../node_modules/@types"],
```

However, if our custom extensions of library types are working, then there might not be any need for typings at all :shrug:

sarayourfriend: So I checked locally and the typings are definitely still working!

@obulat It might also be necessary/wise to update `vue-tsc`, I think :thinking:

obulat: > However, if our custom extensions of library types are working, then there might not be any need for typings at all

I think adding the `typeRoots` helps locating the typings instead of making TypeScript look through all of the possible sources in the whole monorepo. However, I couldn't find the correct path to make it work, and when I take it out, it works 🤷

sarayourfriend: I see! In that case, this appears to work!

```
    "typeRoots": ["./typings", "../node_modules"],
```

sarayourfriend (approved):
LGTM!

dhruvkb (approved):
LGTM!

PR #3344, opened 2023-11-13
Configure the number of proxies to fix anon rate limiting

sarayourfriend:
## Description

Anonymous throttling is currently broken and only works by chance. Given our live API configuration, where we always have X-Forwarded-For, and do not have `NUM_PROXIES` set, DRF's default `get_ident` implementation is using the entire X-Forwarded-For header to build the client identifier. This is a problem because X-Forwarded-For will always include three additional IPs for the reverse proxies in our live environment: Cloudflare's proxy, the AWS load balancer and finally Nginx (which handles static files). The resulting X-Forwarded-For header will look something like this once it reaches Django: `<client IP>, <Cloudflare IP>, <load balancer IP>, <Nginx IP>`.

The most problematic of these are really the load balancer and the nginx IP, because there are 6 load balancers (one for each availability zone) and it's purely chance which one you get, and additionally there are 5 API tasks, and which of those you get is also up to chance. The Cloudflare IP might be stable for requesters within a geographic area, but I'm not 100% sure if that's true. Assuming it is, the chances of the other two are not necessarily evenly weighted, because it's dependent on request volume, so it isn't a clean 1/(5 tasks * 6 lbs) (or 1/30) chance of having the same identifier. Still the chances are high that you'll get a different identifier anyway!

Indeed, I tried three different anonymous requests that all missed the cache and I got the same rate limiting response each time, indicating that I'd only use 1 of my request quota for sustained anonymous.

## Testing Instructions

N/A. This is just adding a rest framework setting. The implementation details are all in the rest framework code, so there's nothing for us to test.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/conf/settings/rest_framework.py
+++ b/api/conf/settings/rest_framework.py
@@ -57,3 +57,5 @@ if config("DISABLE_GLOBAL_THROTTLING", default=True, cast=bool):
         **{k: None for k, _ in REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"].items()}
     )
     del REST_FRAMEWORK["DEFAULT_THROTTLE_CLASSES"]
+
+NUM_PROXIES = config("NUM_PROXIES", default=3, cast=int)
--- a/api/conf/settings/rest_framework.py
+++ b/api/conf/settings/rest_framework.py
@@ -58,4 +58,6 @@ if config("DISABLE_GLOBAL_THROTTLING", default=True, cast=bool):
     )
     del REST_FRAMEWORK["DEFAULT_THROTTLE_CLASSES"]
 
-NUM_PROXIES = config("NUM_PROXIES", default=3, cast=int)
+# https://www.django-rest-framework.org/api-guide/throttling/#how-clients-are-identified
+# We override this in live environments to an appropriate number based on our deployment
+NUM_PROXIES = config("NUM_PROXIES", default=0, cast=int)


api/conf/settings/rest_framework.py
```diff
@@ -57,3 +57,5 @@
         **{k: None for k, _ in REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"].items()}
     )
     del REST_FRAMEWORK["DEFAULT_THROTTLE_CLASSES"]
+
```
obulat, line 60: Would be nice to have a comment on why we set this and why set it to 3.

obulat (approved):
Great find!

api/conf/settings/rest_framework.py
sarayourfriend: We're really just setting a default, and 3 is a production implementation detail. I think this should actually default to 0 instead and then we can set the actual value in our live environments (and document the number there).

api/conf/settings/rest_framework.py
```diff
@@ -57,3 +57,5 @@
...
+NUM_PROXIES = config("NUM_PROXIES", default=3, cast=int)
```
sarayourfriend, line 61:
```suggestion
# https://www.django-rest-framework.org/api-guide/throttling/#how-clients-are-identified
# We override this in live environments to an appropriate number based on our deployment
NUM_PROXIES = config("NUM_PROXIES", default=0, cast=int)
```

api/conf/settings/rest_framework.py
sarayourfriend: I've updated it to default to 0 and added a comment linking to the docs and mentioning that we set a specific value in live environments.

AetherUnbound (approved):
Nice catch!

PR #3358, opened 2023-11-15
Use the correct suffix for exact matches field

### LINKED ISSUE #3269: Correctly set up mixing of exact search with stemming
## Problem

Openverse [search algorithm](https://docs.openverse.org/api/reference/search_algorithm.html#general-query-searching) boosts exact query matches on title, description and tags. However, the field that's used for that was set incorrectly in  [Make quoted queries behave as described in the API documentation (return exact matches only)](https://github.com/WordPress/openverse-api/pull/1012). It used `.exact` suffix, and Openverse index has `.raw` fields, not `.exact`.

## Description

@sarayourfriend suggested in [Simplify search query #3261](https://github.com/WordPress/openverse/pull/3261) [comment](https://github.com/WordPress/openverse/pull/3261#discussion_r1375557637) to replace `.exact` subfield with `.raw` subfields that are not analyzed and, therefore, not stemmed.

## Additional context

[Mixing exact search with stemming](https://www.elastic.co/guide/en/elasticsearch/reference/current/mixing-exact-search-with-stemming.html)

obulat:
## Fixes

Fixes #3269 by @obulat

## Description

This PR uses the correct suffix for the exact matches: in the ES mappings we don't have any fields with `.exact` suffix, but we have `.raw` instead. These are non-analyzed fields that can be used for exact matching in search.

## Testing Instructions

To test this PR, I changed the titles of the first three items in the `sample_image.csv` to "penguin is dancing", "penguins are dancing" and "penguin be dance".

On `main`, I tried `/v1/images/?q=penguins+are+%22dancing%22&unstable__include_sensitive_results=true` request, and got 3 results (where the title has both `dance` and `dancing`) 

Then I checked out this PR, deleted all indexes using Elasticvue, ran `just down` and `just up`/`just init`, and tried the same request.

I got only 2 results, the item with "penguin be dance" in the title did not match.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
Cool that it works! Do you think it's possible to add an integration test for it (sorry to sound like a broken record :cry:)? This looks like something that would be _so_ easy to break again without realising it.

Approving to unblock, but if you decide not to add the test now, please create a follow up issue. It's probably an okay "help wanted" issue for someone :slightly_smiling_face:

obulat:
> Cool that it works! Do you think it's possible to add an integration test for it (sorry to sound like a broken record 😢)? This looks like something that would be _so_ easy to break again without realising it.

An integration test here could be flaky because it would depend on the exact titles of the media items in sample data. I will need to find some items that have the same words in different forms in the sample data. I'll try adding the tests.

sarayourfriend:
> An integration test here could be flaky because it would depend on the exact titles of the media items in sample data. I will need to find some items that have the same words in different forms in the sample data. I'll try adding the tests.

Totally. I wish it was easier to do this. For the record, our media factories create real Elasticsearch records (that get cleaned up after each test), so if you want to create specific media items that would match rather than digging through the sample data (and having that switch from under us causing a flaky test, as you mentioned), that's one option.

https://github.com/WordPress/openverse/blob/95b59110eff953f7d30d9b9444791ae2e0dc6d68/api/test/factory/models/media.py#L120-L124

This fixture would need to be moved up so that the integration tests also use them, though:

https://github.com/WordPress/openverse/blob/95b59110eff953f7d30d9b9444791ae2e0dc6d68/api/test/unit/conftest.py#L149-L174

obulat:
> This fixture would need to be moved up so that the integration tests also use them, though:

I moved this file one level up, but it doesn't seem to work. 
I get `RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.`, even if I add the pytest django db mark.

I added creation of the media with specific titles, and some more checks to the quoted test checking that the quoted text is in the results' titles.

@WordPress/openverse-api , could you help fixing this test?

dhruvkb:
This is not totally ideal but I think renaming the media in the sample data is also one valid approach to do this. It's not the best solution where factories and fixtures are more suitable but it would

- work immediately (as you've already tested it)
- keep working in CI (as long as the samples don't change)
- push this PR over the finish line (🎉)

obulat:
> This is not totally ideal but I think renaming the media in the sample data is also one valid approach to do this. It's not the best solution where factories and fixtures are more suitable but it would
> 
>     * work immediately (as you've already tested it)
> 
>     * keep working in CI (as long as the samples don't change)
> 
>     * push this PR over the finish line (🎉)
> 
> 
> We cannot use factories with integration tests because afaik, the integration tests are from the outside perspective where we're making requests to the Django server and expecting/validating the response without being allowed access to the server-side code. The server-side behaviour is happening outside the test Python context if that makes sense?

Thank you, @dhruvkb!

I've updated the sample data and the tests as you suggested.

After these changes, I tried removing the `.raw`->`.exact` change, and the tests failed (so, it appears the change is working):
```
>       assert all(
            [q in title for title in quoted_result_titles]
        ), f"Not all titles contain exact match for `{q}`: {quoted_result_titles}"
E       AssertionError: Not all titles contain exact match for `dancing penguins`: ['Dancing Penguin', 'dancing penguins']
```

dhruvkb (approved):
Change looks good, tests look good, CI passes too!

PR #3360, opened 2023-11-15
Remove unused simple query string features

### LINKED ISSUE #3327: Disable unused simple query string features
## Problem

Simple query string makes it easy for us to accept user input to directly pass as a query term to Elasticsearch without worrying about messy or bad inputs causing unexpected behaviour for users. However, simple query string is expensive, so if it's possible to turn off any of the features that (a) we don't see people using (when we check logs) or (b) don't _want_ people using (fuzzy, wildcard, etc?) then we can gain some performance benefits.

## Description

This is a two part issue. For each of the [simple query string features that can be turned off by the limit operators](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html#supported-flags) investigate whether they've been used in the last month by querying the API logs in CloudWatch. Any that have been used less than 100 times in the last month are worth turning off (that would represent a statistically meaningless number of search requests, 3 per day in the context of several hundred thousand requests per day).

My guess is we can probably turn off the following:

- Fuzzy
- Near
- Prefix
- Slop
- Escape

**Note**: This also requires modifying the frontend "search syntax" guide to remove mention of features we no longer support after this change, hence the "frontend" label alongside the API label.

## Additional context

Suggested to us by Greg who works on Jetpack.

obulat:
## Fixes

Fixes #3327 by @sarayourfriend

## Description

This PR removes the following feature from the simple query string:
- Fuzzy
- Near (Slop)
- Prefix
- Escape
The following features are left:
- And
- Not
- Phrase
- Whitespace

For more reasoning on why each feature is removed or left in, see the linked issue.

The search guide page removed mentions of the features that were removed. The text was also simplified to make it clearer now that there are only 2 features left. The `+` feature wasn't removed from simple search query in the API, however, since the behavior of "cat dog" is the same as "cat+dog", I removed it from the search syntax description page to prevent confusion.

## Testing Instructions

Run the app. The features that were turned off should not affect the search results (`()` precedence, `~` fuzziness,), but the `"` quotes for the exact match, `-` for exclusion, `+` for `and` should work as they did before.
The frontend `/search-help` page text should be clear and describe the search query syntax.

Also, look through the query features that were left in/removed, and see if you agree with it.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
I'm hesitant to ask this because I know it's a significant amount of work, but this change (even though so few people use it!) does seem non-trivial for how our search functions. Do you feel it might be appropriate to have this change be under a different API version (e.g. `/v2/`). I guess we're not changing the _shape_ of the API per-se, but rather how our search functions behind the scenes? Interested to hear @WordPress/openverse-maintainers's thoughts on that.

sarayourfriend (approved):
Awesome. I'm so excited to see if this improves anything on our search timings :eyes:

AetherUnbound:
Thanks for your thoughts Sara 🙂 I agree, my impression was to move forward regardless but I wanted to check with folks to make sure that sense was shared. And we can move forward with confidence thanks to the great analysis Olga did, too!

obulat:
> If we hear that someone is actually using and relying on this feature, then we can selectively turn them on when a query includes key characters (`*`, `|`, `~`, etc), and keep them _off_ for the the rest. We can even do that now if it would assuage concerns about rolling this out non-versioned. However, I keep coming back to Olga's excellent analysis which for me completely satisfactorily shows that these features are not used in a meaningful way.

Do you think it would be helpful to add this and add logging for when the query has a key character so that we can analyze how much they are used. Now that the frontend page does not have these links, we will probably get more accurate usage numbers.

dhruvkb (approved):
LGTM! Approving to unblock, although breaking compatibility is not very appealing. It we can conditionally enable those features when the strings have the special characters, that would be ideal.

sarayourfriend:
> It we can conditionally enable those features when the strings have the special characters, that would be ideal.

I asked one of the Elasticsearch expert folks we've been talking to about the response time issues whether this would make a difference. My thinking being, I'd be surprised if ES doesn't itself already do something like a `"*" in query` check before doing anything with the prefix feature. It would be weird if even without `*` in the query it still had to spend time working on prefix related queries.

In other words, if these features aren't used very heavily, then maybe they aren't causing any performance difference.

On the other hand, if they are enabled, even if conditionally (whether by us or by relying on ES to correctly ignore features that aren't in the query syntax), then it leaves open the possibility for someone to cause our service overwhelming load by sending heaps of these kinds of requests with complex features. From a service resilience standpoint, it's worth turning these features off entirely _because_ they break compatibility. From that perspective, ideally we turn them off soon, before there is any large scale reliance on them.

sarayourfriend:
> Do you think it would be helpful to add this and add logging for when the query has a key character so that we can analyze how much they are used? Now that the frontend page does not have these links, we will probably get more accurate usage numbers.

This sounds like a good option too, @obulat. If you're comfortable with this PR drafting and us getting that analysis information for a few weeks, it would be good information to help us make an accurate decision.

obulat:
I've added the logging for query strings that contain one of the 4 symbols, but kept the flags off. This is why I requested a re-review, @dhruvkb, @dhruvkb.

sarayourfriend (approved):

api/api/controllers/search_controller.py
```diff
@@ -339,6 +347,18 @@ def build_search_query(
...
+def log_query_features(query: str, query_name) -> None:
+    flags = [
+        ("PRECEDENCE", r"\(.*\)"),
+        ("ESCAPE", r"\\"),
+        ("FUZZY|SLOP", r"~\d"),
+        ("PREFIX", r"\*"),
+    ]
+    for flag, pattern in flags:
+        if bool(re.search(pattern, query)):
+            log.info(f"Special feature in `{query_name}` query string. {flag}: {query}")
```
sarayourfriend, line 359: Just a suggestion, but it would produce less logs overall while making it easier to tally them if the logging did something like, counting the instances of each flag and logging all flags at once in JSON like `{ "log_message": "Special features present in query", "query_name": query_name, "query": query} | {flag: count_flag(flag, query) for flag in flags }` (one-liner just for commenting, not to suggest that's the way to do it).

No big deal, though. This way works too, it just produces slightly more logs (not many, because we know these are used so rarely and mostly following the frontend examples) and requires parsing the message string rather than using Logs Insights discovered fields feature.

obulat: Thank you for this suggestion, I was actually not certain that the logging pattern would be good, and I didn't even realize that there could be several flags used in a single search. I'll update the logging

obulat: Updated the logging, now if you go to `http://0.0.0.0:50280/v1/images/?q=cat+net*+(dog)+%5C(+someth~2`, you get the following message logged:
`[2023-11-30 08:32:53,795 - root - 362][INFO] [3419191a256b4058b2e8ffd5066b1ec2] {'log_message': 'Special features present in query', 'query_name': 'q', 'query': 'cat net* (dog) \\( someth~2', 'flags': ['PRECEDENCE', 'ESCAPE', 'FUZZY|SLOP', 'PREFIX']}`

I don't think we need to log the flag cound separately, do you, @sarayourfriend ?

sarayourfriend: I guess the raw count isn't that important, is it! Good call :+1: The change LGTM.

PR #3361, opened 2023-11-15
Unify DAG creation/database cleaning fixtures for testing

### LINKED ISSUE #2622: Unify DAG creation/database cleaning fixtures for testing
## Description

We have a few places in our catalog tests with very similar logic for creating the DAG necessary for running some tests. Examples:

https://github.com/WordPress/openverse/blob/9056029e6c354fe2dbf5b0ad49d60648bd380d27/catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py#L23-L34

https://github.com/WordPress/openverse/blob/9056029e6c354fe2dbf5b0ad49d60648bd380d27/catalog/tests/dags/data_refresh/test_dag_factory.py#L22-L25

https://github.com/WordPress/openverse/blob/9056029e6c354fe2dbf5b0ad49d60648bd380d27/catalog/tests/dags/common/test_ingestion_server.py#L21-L31

https://github.com/WordPress/openverse/blob/9056029e6c354fe2dbf5b0ad49d60648bd380d27/catalog/tests/dags/common/sensors/test_utils.py#L18-L21

https://github.com/WordPress/openverse/blob/9056029e6c354fe2dbf5b0ad49d60648bd380d27/catalog/tests/dags/providers/test_provider_dag_factory.py#L27-L37

Although there are some slight differences, it would be ideal if we could find a way to combine these into a single fixture that's used within all the affected tests, rather than redefining the fixture for every test.

## Additional context

Came up in discussing https://github.com/WordPress/openverse/pull/2209#discussion_r1232622950

ngken0995:
## Fixes

Fixes #2622  by @AetherUnbound 

## Description

There will be one `clean_db()` in conftest.py file because tests from multiple test modules in the directory can access the fixture function. `Pool` and `TaskInstance` from `airflow.models` will have a if condition because not all tests runs a filter and deletion for them. There is a override method for fixture on a test module level([link](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#override-a-fixture-on-a-test-module-level)) There are default values `get_test_dag_id`, `get_test_id`, `isPool`, `isTaskInstance` in `conftest.py`. On each test involving `clean_db()`, there should be a `get_test_dag_id()` fixture and options to run `isPool` fixture and `isTaskInstance` fixture by returning True

## Testing Instructions

- run `just down -v`
- run `just catalog/up`
- run `just catalog/init`
- run `just catalog/test`

## Checklist

- [X] I added or updated tests for the changes I made (if applicable).
- [X] I tried running the project locally and verified that there are no visible errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/tests/conftest.py
+++ b/catalog/tests/conftest.py
@@ -1,4 +1,6 @@
 import pytest
+from airflow.utils.session import create_session
+from airflow.models import DagRun, Pool, TaskInstance
 
 
 def pytest_addoption(parser):
@@ -24,3 +26,31 @@ def pytest_addoption(parser):
 # Use this decorator on tests which are expected to take a long time and would best be
 # run on CI only
 mark_extended = pytest.mark.skipif("not config.getoption('extended')")
+
+@pytest.fixture()
+def get_test_dag_id():
+    return ""
+
+@pytest.fixture()
+def get_test_pool():
+    return ""
+@pytest.fixture()
+def isTaskInstance():
+    return False
+
+@pytest.fixture()
+def isPool():
+    return False
+
+@pytest.fixture(autouse=True)
+def clean_db(get_test_dag_id, get_test_pool, isTaskInstance, isPool):
+    with create_session() as session:
+        # synchronize_session='fetch' required here to refresh models
+        # https://stackoverflow.com/a/51222378 CC BY-SA 4.0
+        session.query(DagRun).filter(DagRun.dag_id == get_test_dag_id).delete()
+        if isTaskInstance:
+            session.query(TaskInstance).filter(
+                TaskInstance.dag_id == get_test_dag_id
+            ).delete(synchronize_session="fetch")
+        if isPool:
+            session.query(Pool).filter(id == get_test_pool).delete()
--- a/catalog/tests/conftest.py
+++ b/catalog/tests/conftest.py
@@ -36,26 +36,26 @@ def _normalize_test_module_name(request) -> str:
 
 
 @pytest.fixture
-def get_test_dag_id(request):
+def sample_dag_id_fixture(request):
     return f"{_normalize_test_module_name(request)}_dag"
 
 
 @pytest.fixture
-def get_test_pool(request):
+def sample_pool_fixture(request):
     return f"{_normalize_test_module_name(request)}_pool"
 
 
 @pytest.fixture
-def clean_db(get_test_dag_id, get_test_pool):
+def clean_db(sample_dag_id_fixture, sample_pool_fixture):
     with create_session() as session:
         # synchronize_session='fetch' required here to refresh models
         # https://stackoverflow.com/a/51222378 CC BY-SA 4.0
-        session.query(DagRun).filter(DagRun.dag_id.startswith(get_test_dag_id)).delete(
-            synchronize_session="fetch"
-        )
-        session.query(TaskInstance).filter(
-            TaskInstance.dag_id.startswith(get_test_dag_id)
+        session.query(DagRun).filter(
+            DagRun.dag_id.startswith(sample_dag_id_fixture)
         ).delete(synchronize_session="fetch")
-        session.query(Pool).filter(Pool.pool.startswith(get_test_pool)).delete(
+        session.query(TaskInstance).filter(
+            TaskInstance.dag_id.startswith(sample_dag_id_fixture)
+        ).delete(synchronize_session="fetch")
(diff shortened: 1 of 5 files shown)


AetherUnbound (changes requested):
Thanks for taking this on Kenneth! I think there are a few changes I'd like to see with this. A pytest fixture that exists in the highest level `conftest.py` file and also has `autouse=True` means that the fixture will be used _in ever test that's run_. Most of our tests don't even interact with the Airflow database, so this becomes quite excessive and unnecessary. Additionally, we don't need the extra fixtures for conditionally cleaning task instances and pools if we opt to 1) automatically define a task instance/pool name for each test and 2) always clean them up by selecting records that match those names (whether they exist or not doesn't matter).

I think one way to do this is to create some helper fixtures that generate a unique DAG ID and pool based on the _test module_. (I say module here instead of test name because we have the test multi-processing binned by module, see [here](https://github.com/WordPress/openverse/blob/910cc68248a623c43becfbc0b5f630cd128ece79/catalog/pytest.ini#L17) and [docs](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#running-tests-across-multiple-cpus). Here's an example of what that might look like:

```python
# in conftest.py
def _normalize_test_module_name(request) -> str:
    # Extract the test name
    name = request.module.__name__
    # Replace periods with two underscores
    return name.replace(".", "__")


@pytest.fixture
def get_test_dag_id(request):
    return f"{_normalize_test_module_name(request)}_dag"


@pytest.fixture
def get_test_pool(request):
    return f"{_normalize_test_module_name(request)}_pool"


@pytest.fixture
def clean_db(get_test_dag_id, get_test_pool):
    with create_session() as session:
        # synchronize_session='fetch' required here to refresh models
        # https://stackoverflow.com/a/51222378 CC BY-SA 4.0
        session.query(DagRun).filter(DagRun.dag_id == get_test_dag_id).delete()
        session.query(TaskInstance).filter(
            TaskInstance.dag_id.startswith(get_test_dag_id)
        ).delete(synchronize_session="fetch")
        session.query(Pool).filter(id == get_test_pool).delete()
```

Tests that required a clean database, test DAG, test pool, etc, could then request them as fixtures in their function signature. For example, the test in `test_ingestion_server.py` could be changed to:

```python


@pytest.fixture()
def index_readiness_dag(get_test_dag_id, clean_db):
    # Create a DAG that just has an index_readiness_check task
    with DAG(dag_id=get_test_dag_id, schedule=None, start_date=TEST_START_DATE) as dag:
        ingestion_server.index_readiness_check(
            media_type="image", index_suffix="my_test_suffix", timeout=timedelta(days=1)
        )

    return dag
```

(note the usage of `get_test_dag_id` and `clean_db` when creating the fixture)

This might require a few more changes to the affected tests, but I think it's much better than adding the `clean_db` step to every test!

ngken0995:
@AetherUnbound unittest.TestCase only allow auto-use fixture([link](https://pytest.org/en/latest/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses)) In `test_single_run_external_dags_sensor.py`, `TEST_POOL` and `DAG_PREFIX` are hardcode with the same outcome to fixture `get_test_dag_id` and `get_test_pool`. In `clean_db`, `DagRun`, `TaskInstance` and `Pool` will filter with startswith to search for prefix of names and delete them.

AetherUnbound (commented):
Thanks so much for reworking this, and sorry for the delay on review! 

That's frustrating about the `unittest.TestCase` scenario 😭 I'll make an issue for converting all our catalog tests from `unittest`, I don't think there's a specific reason they need to be in that format (but that's not the purview of this particular PR!). Do you mind adding a "TODO" comment right above the test pool/test prefix constants in `test_single_run_external_dags_sensor.py` noting that they can be removed and the fixtures can be used instead once this test is converted to pytest?

Also, I apologize for suggestion one thing then requesting another, but I'm realizing the `get_*` prefix doesn't really fit our pytest fixtures standard. Would you mind renaming `get_test_dag_id` to `sample_dag_id_fixture` and `get_test_pool` to `sample_pool_fixture`?

AetherUnbound (approved):
Fantastic, everything looks great! Thanks Kenneth!

stacimc (approved):
This looks great, thanks @ngken0995!

Issue #3257, opened 2023-10-25
Filter dead link-only providers in the Catalog


stacimc:
## Problem

The search controller includes a filter to [dynamically exclude providers](https://github.com/WordPress/openverse/blob/main/api/api/controllers/search_controller.py#L279) with the `filter_content` attribute set. This allows us to exclude providers whose content is exclusively dead links, but it results in an additional ES filter on every query.

## Description

We could instead filter out these providers on the Catalog side by removing them from the Catalog's media tables. On subsequent data refreshes, the records will no longer be copied into the API tables -- no filtering necessary. This makes this a one-time operation.

Rather than simply deleting the records, we should move them into a new table. For now, these tables should be as simple as possible. This will be `deleted_image` and `deleted_audio`. They should be exact copies of the `image` and `audio` tables respectively, with two new columns:
* `deleted_on`: timestamp for when the record was added to the table/deleted from the corresponding media table
* `reason`: reason the record was deleted, which for the purposes of this work will just be `dead_link` for all records

In the future, there may be a project to consider other cases for handling deleted media in the Catalog. The deleted media tables may be changed as part of that work. The goal in this issue is to make something simple that can easily be modified when the project is planned more thoroughly (for that reason, for example, we'll just preserve every column from the media tables exactly so that the records could be easily restored).b 

We'll need to:
* Create the tables
* Write queries for moving the archived provider records into these tables, and then deleting them from the main media tables. It might make the most sense to do this in a small DAG.
* Remove the filtering from the API

## Alternatives

Alternatively this could be implemented more quickly by updating the ingestion server to [filter out these records](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/queries.py#L178) during the copy_data query step so that they are not copied into the API tables. This would be a very small change, and is similar to how we already filter out records that are in the API's `DeletedMedia` table.

The primary disadvantage is that this filtering needs to happen on every data refresh. I am not certain how much longer the

stacimc:
In an ad-hoc discussion, @obulat and I agreed that the approach described in this issue (moving the records into a deleted media table in the Catalog) would be the better solution. However as I've written up this issue it's occurred to me how much quicker the ingestion-server approach would be to implement, so I think it's worth touching base again to make sure this is still how we want to proceed. Another option is to do the filtering in the ingestion server _temporarily_, as it would also be easy to undo. For the moment I'll take a time-boxed approach into looking at both options.

obulat:
[...]

PR #3362, opened 2023-11-15
Do not Merge: Remove dynamic provider filtering

stacimc:
## Fixes

Part of #3257 by @stacimc
Follow up to #3259

## Description

This PR removes dynamic provider filtering in the API, following the work in #3257. This dynamic filtering was *only* being used to filter out archived providers, who have all now been removed from the dataset at the catalog level. After the image data refresh is run, those records will not even be available in the API db & elasticsearch indices, so filtering them is no longer necessary at the API level. This PR entirely gets rid of that filter in the API

As I started to post this, however, I realized that the dynamic provider filtering can still be useful because it allows us to immediately stop serving results for a provider without needing to wait for a `delete_records` run + data refresh. It's also more flexible because we can turn the filtering on and back off again instantly. This is an important tool we shouldn't remove.

So I think the correct solution here is actually to:
* Wait for the image data refresh to finish.
* Remove the affected image providers from the `content_provider` table in the API. This makes sense because no records from those providers will be in the API anymore.
* Leave the filtering as is

As long as no providers have `filter_content` enabled, we won't have any additional ES filters applied. We should note that removing records at the catalog level is preferred for this reason, but the ES filter can be used for hiding records quickly (or when data refreshes aren't running) or temporarily.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -5,10 +5,7 @@ from elasticsearch_dsl.query import Match, Q, Term
 from elasticsearch_dsl.response import Hit
 
 from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
-from api.controllers.search_controller import (
-    _post_process_results,
-    get_excluded_providers_query,
-)
+from api.controllers.search_controller import _post_process_results
 
 
 def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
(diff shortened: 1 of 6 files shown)


stacimc:
Pulled the formatting change out into #3363, closing this for the reasons described in the description.

PR #3140, opened 2023-10-05
Add additional search view pages to the Nuxt app

obulat:
~~This PR is based on #3136, and will be undrafted after that PR is merged.~~

## Fixes

Fixes #2858 by @obulat
Fixes #2774 by @obulat 

## Description

With this PR, you can view the tag/source/creator collections in the Nuxt app.

To show the collections, we add `_tag.vue` and `_source.vue` pages to both `image` and `audio`, with dynamic routes. These pages check the path to make sure that the path has the necessary parameters: tag name for the tag page; an existing source for source/creator pages; `/creator/` path part and the creator name for the creator pages. 

In the first iteration, I used a single page with a fully-dynamic path (`/_mediaType/_collection/_.vue`) that would then validate the `mediaType`, `collection` and the collection parameters. However, because `_mediaType` can accept any string path, it would also match the locale routes such as `/ar/non_existent_page`), so I split this single page into 4.

The collection pages have the collection headers (using `VCollectionHeader` from #3365) and either a `VAudioCollection` or `VImageGrid`, as well as the "Load more" button. When you navigate from the collection page to a single result page, you also get the "Back to results" link at the top that goes back to the collection page.

This PR adds a `strategy` and `collectionParams` props to the `search` store. Whenever we set the `searchTerm` (which is only available for search, not collection), we reset the `strategy` to `default` and nullify the `collectionParams`.

The API service had to be updated to accept a `slug` for paths like `/tag/tag_name`, `/source/source_name`, `/source/source_name/creator/creator_name`, which are added to the `/<mediaType>` path.
The media store was updated to use the necessary slug from the search store `collectionParams`.

## Testing Instructions

Run the app using `just frontend/run dev`. Go to `/preferences` page and turn on the `additional_search_views` feature flag.

The easiest way to land on a collection page is from the single result page. Click on the source, creator or tag buttons - and you will be taken to the corresponding page.
Check that in non-English locales, the links have the correct locale path, and when navigating, the locale is kept. 
Check that if you go from a collection page to a single result page, you can go back to the collection page using "Back to results" page.
Check that if you go back to homepage from the collection page, you can search for something.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
[...]

PR #3365, opened 2023-11-16
Update VCollectionHeader.vue

obulat:
## Fixes

Updated implementation of #2729 by @obulat

## Description

This PR is extracted from #3140 to make them easier to review.

When adding Nuxt page for the tag/source/creator collections in #3140, I first tried to add a single dynamic page for all of them to prevent code duplication.

I used a dynamic catch-all path `_mediaType/_collection/_.vue`, and thought that `_mediaType` would catch the `image` or `audio` path parameter, `_collection` would catch the first part of the path (`tag` or `source`),  and `_.vue` would be the catch-all that collects the rest of the path parameters: `tag/tagName`, `source/sourceName`, `source/sourceName/creator/creatorName`.

However, this strategy catches more routes than expected. For instance, it would catch `/<localePrefix>/non-existent-page`, and would return the English error page because the collection page was seeing the locale prefix as `_mediaType`, and path validation would fail because the locale prefix does not match the `_mediaType` (image/audio).

This is why I added 4 pages that used less dynamic path parameters (`/image/tag/_tag.vue`, `/image/source/_.vue`).
Previously, I planned to use the page to compute all the necessary parameters, and have `VCollectionHeader` as a simple presentational component. However, now that there are 4 page component, moving the computation to the store and the component makes more sense. 

Now `VCollectionHeader` accepts the `collectionParams` prop from the page component, and then computes the title, URL and the results label. There is a complication for the creator pages: we can only get the creator URL after we fetch the media, from the media item's `creator_url` param. This is why the `creatorUrl` is still computed in the page (which handles media fetching) and is passed as a prop to `collectionParams`.

This PR also converts the `CollectionParams` type to a [discriminated union type](https://basarat.gitbook.io/typescript/type-system/discriminated-unions) to make it easier to narrow the type when handling it.

## Testing Instructions

Check the Storybook to see that `VCollectionHeader` is rendered correctly: https://docs.openverse.org/_preview/3365/storybook/?path=/story/components-vcollectionheader--all-collections
To run the Storybook locally, you may need to run `just frontend/run test:storybook` to run Storybook inside docker (it currently does not run locally on Mac due to ssl errors).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/locales/scripts/en.json5
```diff
@@ -826,5 +826,43 @@
       source: "Open source site",
       creator: "Open creator page",
     },
+    resultCountLabel: {
+      creator: {
+        audio: {
+          zero: "No audio files by this creator in {source}",
+          count: "{count} audio file by this creator in {source}|{count} audio files by this creator in {source}",
+          countMore: "Over {count} audio files by this creator in {source}",
```
AetherUnbound, line 834: These seem very similar, would it also be possible to interpolate the media type here too? Or is it just easier to define them separately?

AetherUnbound (approved):
The code looks good and makes sense to me! One question about the translations, nothing to block a merge 🚀

frontend/src/locales/scripts/en.json5
obulat: I think it's better to leave as much as possible fully written out because interpolation does not always work well for different languages, which might have very different pluralization and other grammar rules.

frontend/src/locales/scripts/en.json5
```diff
@@ -826,5 +826,43 @@
       source: "Open source site",
       creator: "Open creator page",
     },
+    resultCountLabel: {
+      creator: {
+        audio: {
+          zero: "No audio files by this creator in {source}",
```
dhruvkb, line 832: Is "track" a better word than "file" here? I would like us to identify a word that's more descriptive than "file" and use it consistently in the UI.

dhruvkb (approved):
The changes make sense, LGTM. Why did the snapshots change from "10000 images" to "No image", is it possible to get 10000 and avoid the diffs?

obulat:
> The changes make sense, LGTM. Why did the snapshots change from "10000 images" to "No image", is it possible to get 10000 and avoid the diffs?

Thank you for this comment, I did update the test to keep the number. The snapshots still had to be changed, though, due to the fact that I didn't add "over" in the original implementation, and instead of "Over 10000 results", it was just "10000 results" before 😢
This is not accurate since 10000 is a cut-off that we return every time there are more than 10000 results.

Issue #2246, opened 2023-05-31
Add Django request ID to nginx log output


sarayourfriend:
## Problem

Django attaches a request ID header to responses (`X-Request-Id`) but this doesn't show up in Nginx logs. That means we don't have any way to easily tie Django events to specific Nginx requests.

## Description

Add `X-Request-Id` to the Nginx JSON logging configuration.

## Additional context

Testing this might be a pain as we don't have anything set up for running the Nginx configuration locally. [The testing instructions from this PR](https://github.com/WordPress/openverse-api/pull/990) will be a good starting point for getting it wired up. Essentially the Nginx image needs to be run manually via docker run with environment variables set such that the requests are forwarded to you local Django container.

PR #3369, opened 2023-11-19
Add middleware to log application name and verification status

### LINKED ISSUE #3368: Add response headers and nginx logging to identify requests and registered applications used for a request
## Problem

Right not it's not possible for us to identify applications related to a request. If any particular registered application is causing issues, it is essentially impossible for us to identify that through our existing request logging.

## Description

Add the following parameters to Nginx logging (some will require new response headers added to relevant requests):

- Request ID (from the request ID logging middleware, present in the `x-request-id` response header)
- Registered client application name and verified status

The last two will require new headers added so that we can log them in Nginx.

This is not a privacy concern. Our API terms of service already allow for this kind of introspection into registered application usage, and we theoretically have _more_ granular data if we really wanted to fingerprint or identify individual users. To clarify, that is not our intention with this. If we wanted to do that we'd fingerprint off of user agent and source IP. The intention behind this is to help identify potentially nefarious API requesters and to see if any of the identifiable bot traffic we see in traffic patterns is tied to specific registered applications. I'm adding this caveat just to clarify that adding this logging does not make anything _less_ private than it already is.

sarayourfriend:
Also add authorization, app name and verified status to nginx logs

## Fixes

Closes https://github.com/WordPress/openverse/issues/2246 by @sarayourfriend 
Fixes #3368 by @sarayourfriend 

## Description

Add logging as described in the issue.

To add the application name and verified status, I added a new middleware. The tests for this are in the `test_auth` suite.

## Testing Instructions

You need to rebuild the nginx image to get the new logging: `just build nginx`. Afterwards, make requests to your local API via the nginx container: `localhost:50270`. Follow that container's logs (`just logs nginx`) and confirm that you see the following JSON log properties:

- `request_id`
- `client_application_name`
- `client_application_verified`

These come from the following headers, respectively, which you should also see on responses:

- `x-request-id`
- `x-ov-client-application-name`
- `x-ov-client-application-verified`

Note that the last two are only present on authorized requests. Nginx will log an empty string when those headers are not present, but if you make an authorized request locally (by following the auth flow), you'll be able to see these populated both in your response headers and in Nginx's logs.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/middleware/client_application_middleware.py
+++ b/api/api/middleware/client_application_middleware.py
@@ -1,0 +1,16 @@
+from api.utils.oauth2_helper import get_token_info
+
+
+def client_application_middleware(get_response):
+    def middleware(request):
+        response = get_response(request)
+
+        if hasattr(request, "auth") and request.auth:
+            token_info = get_token_info(str(request.auth))
+            if token_info:
+                response["x-ov-client-application-name"] = token_info.application_name
+                response["x-ov-client-application-verified"] = token_info.verified
+
+        return response
+
+    return middleware
--- a/api/api/middleware/response_headers_middleware.py
+++ b/api/api/middleware/response_headers_middleware.py
@@ -1,7 +1,16 @@
 from api.utils.oauth2_helper import get_token_info
 
 
-def client_application_middleware(get_response):
+def response_headers_middleware(get_response):
+    """
+    Add standard response headers used by Nginx logging.
+
+    These headers help Openverse more easily and directly connect
+    individual requests to each other. This is particularly useful
+    when evaluating traffic patterns from individual source IPs
+    to identify malicious requesters or request patterns.
+    """
+
     def middleware(request):
         response = get_response(request)
 
--- a/api/api/utils/oauth2_helper.py
+++ b/api/api/utils/oauth2_helper.py
@@ -1,5 +1,6 @@
 import datetime as dt
 import logging
+from dataclasses import dataclass
 
 from oauth2_provider.models import AccessToken
 
@@ -8,10 +9,22 @@ from api import models
 
 parent_logger = logging.getLogger(__name__)
 
-_no_result = (None, None, None)
+
+@dataclass
+class TokenInfo:
+    """Extracted ``models.ThrottledApplication`` metadata."""
+
+    client_id: str
+    rate_limit_model: str
+    verified: bool
+    application_name: str
+
+    @property
+    def valid(self):
+        return self.client_id and self.verified
 
 
-def get_token_info(token: str):
+def get_token_info(token: str) -> None | TokenInfo:
     """
     Recover an OAuth2 application client ID and rate limit model from an access token.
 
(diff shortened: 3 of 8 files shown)


api/api/middleware/client_application_middleware.py
```diff
@@ -0,0 +1,16 @@
+from api.utils.oauth2_helper import get_token_info
+
+
+def client_application_middleware(get_response):
```
dhruvkb, line 4: A few docstrings about this middleware would be nice! Just what's in the PR description is good enough.

dhruvkb (approved):
LGTM, the code works as expected! I added a small note about more docs but it's not blocking.

api/api/middleware/client_application_middleware.py
sarayourfriend: Good idea. I'll rename this to add "headers" in there as well. The name I've chosen to commit is bafflingly obscure :sweat_smile:

obulat (approved):
Works well locally, and has great tests.
A question not related to this PR: what is `request_id` for? It's different every time I request something. (`'"request_id":"$sent_http_x_request_id",'`). 

Seeing these `x_` headers here, I thought that we could also add some `x_` header to all the client frontend requests to identify them instead of trying to set the header in Axios (which is not allowed as it's not safe).

sarayourfriend:
>  A question not related to this PR: what is request_id for? It's different every time I request something.

The ID assigned to the request is meant to help us trace an individual requests logs. In a context where logs relevant to different requests can interweave (either in an async context or a multi-worker context), it helps us determine which request caused what logs.

Take the following scenario, for example, considering the same working in an async context.

Request A comes in at timestamp 0
Request B comes in at timestamp 1

Request A will run until it reaches an `await`, for example, to send the query to ES or to make a database request.

At that point, Request B will run. Imagine that request B finishes completely before request A's `await` resolves. For example, if request B is not an async route or if request B's async operations finish faster than A's. In that case, all the logging for request B will happen between the logging for request A before and after request A's `await` point. It could look like this:

```
req_a: start
req_a: Sending query {query}
req_b: start
req_b: some important error log!
req_b: end
req_a: Received query response {response}
req_a: end
```

Without a request ID, it wouldn't be either impossible or very difficult to determine whether the error log caused by request b was caused by request B or some other request. If the two requests are totally different code paths, then that makes it possible to deduce which request it's for. But if they're for the same endpoint? It's impossible without reading further logs, and even then, only if further logs are clearly tied to a specific request via the request's context (e.g., by logging the query).

Request ID eliminates the need to do this kind of mental juggling and reliance on deep knowledge of the code when trying to trace the lifecycle of a request in logs and also eliminates the possibility of two requests' logs getting mixed up altogether.

sarayourfriend:
> Seeing these x_ headers here, I thought that we could also add some x_ header to all the client frontend requests to identify them instead of trying to set the header in Axios (which is not allowed as it's not safe).

We can identify frontend client requests in two ways:

1. After this PR, via the client application name, for SSR requests
2. Already existing, we can see the `openverse.org` referrer header (already in our Nginx logs)

Would that cover what you're looking to find, @obulat?

PR #3020, opened 2023-09-13
Add ADRF and make the thumbnail view async

sarayourfriend:
## Fixes

First step of #2789 by @sarayourfriend

## Description

This is the first step to convert the image proxy route to async. It converts the overall view to async. I've split this into a separate PR because (a) it is complex enough on its own and (b) converting the actual `image_proxy.get` function is blocked on sharing the aiohttp client session (for which I'll have another PR up later today, it's already finished I just forgot to push it last night :facepalm:).

So this PR only converts the _view_ side of things to async. It does this by abstracting the thumbnails route in a slightly different way than it was before, which allows the audio and image views to ignore most of the safe-sync-in-async compatibility issues. To make that interplay a little easier, I decided to refactor `image_proxy.get` to use two MRO objects for the input, rather than the one-by-one parameter definitions. Now that I'm writing this, I'm realising that I could have avoided the changes to `image_proxy.get` itself by creating a new response definition for `get_image_proxy_media_info` and then using `asdict` to spread that into `image_proxy.get`. If y'all want me to make that change, I think it's a fine one to simplify the overall PR. On the other hand, I think the new API is also a bit cleaner, if potentially too abstract. Let me know what y'all think because I'm two minds about it.

## Testing Instructions

Checkout the PR and run `just build web` to get the new dependencies. Also keep in mind that this PR depends on #3011.

Run `just api/up` and make various search requests, thumbnail, and related requests. Generally: exercise the image and audio media views and their various routes. Everything should essentially work as it did before. The existing API integration tests should also continue to work, along will all unit tests. This change should be generally totally transparent and not result in any observable differences on the consumer side, which is why I didn't add any new tests for this: the existing tests already cover the media view and should continue to work without changes. The only tests I changed were to use pook a little more fluidly and to accommodate the changes to the `image_proxy.get` function's interface.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

[...]

Issue #2542, opened 2023-07-03
Replace instances of "Photon" with "Site Accelerator"


zackkrida:
## Problem

Photon, the image CDN we currently use, has been renamed to "Site Accelerator":

https://developer.wordpress.com/docs/site-accelerator/

## Description

Update all references in the codebase to use this new name.

sarayourfriend:
I'm converting this to staff only @AetherUnbound because it will also require updating these environment variables, which will require coordinating infrastructure changes and staged deployment:

https://github.com/WordPress/openverse/blob/f54c71c4b7d4be49ea94f0501b5089171695235c/api/conf/settings/thumbnails.py#L4-L10

PR #3388, opened 2023-11-22
Convert entire image proxy route async

### LINKED ISSUE #2789: Convert the image proxy route to async

sarayourfriend:
## Fixes

Fixes #2789 by @sarayourfriend 

## Description

This PR converts `image_proxy::get` to an asynchronous function. This makes the entire `thumbnails` route fully asynchronous, meaning we finally start to gain benefits from running the application under ASGI. We should see increased thumbnail throughput after this (though, it might not be noticeable because we probably won't receive more traffic), but as a result, we should be able to lower the number of thumbnail tasks in the mean time.

We should assume this will increase CPU and memory usage, however, as now each thumbnails worker will be able to handle multiple requests, which will require more context switching (read: increased memory). This is really the first "real" async benefit we get though, because the thumbnails route is _heavily_ asynchronous. The vast majority of time our servers are handling thumbnail requests they're merely waiting for an upstream response! That means they're wasting cycles that could be spent serving (preparing) additional requests to the proxy.

This PR will finalise the Django ASGI project as far as the currently created issues. However, I want to clarify something about the project, and have done so in the project thread.

## Testing Instructions

Everything should continue to work as before. You should be able to follow the testing instructions in #3020, but on top of that you should also test to confirm that Redis tallying, in particular the "fire and forget" parts, are still working. You can do this via the redis cli: `just exec cache redis-cli -n 3` and reading the relevant keys before/after a request to ensure they've incremented as expected.

This PR is a draft for now because I still need to refactor the unit tests to handle the new async function. I ran out of time today but wanted to get the working code up in the meantime. I'll work on the automated tests later this week and hopefully have this ready to review by end of week.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/asyncio.py
+++ b/api/api/utils/asyncio.py
@@ -1,0 +1,29 @@
+import asyncio
+import logging
+from collections.abc import Awaitable
+
+
+parent_logger = logging.getLogger(__name__)
+
+
+_fire_and_forget_logger = parent_logger.getChild("fire_and_forget")
+
+
+def fire_and_forget(awaitable: Awaitable) -> None:
+    """
+    Consume an awaitable without waiting for it to finish.
+
+    This allows us to "fire and forget" an async function that
+    we don't care about the result of. This is useful, for example,
+    if some operation creates a side effect that isn't necessary for
+    the response we're processing (e.g., Redis tallying).
+    """
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError as exc:
+        _fire_and_forget_logger.error(
+            "`fire_and_forget` must be called inside a running" " event loop."
+        )
+        raise exc
+
+    loop.create_task(awaitable)
--- a/api/api/utils/image_proxy/__init__.py
+++ b/api/api/utils/image_proxy/__init__.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Literal
 from urllib.parse import urlparse
@@ -7,11 +8,15 @@ from django.conf import settings
 from django.http import HttpResponse
 from rest_framework.exceptions import UnsupportedMediaType
 
+import aiohttp
 import django_redis
 import requests
 import sentry_sdk
+from asgiref.sync import sync_to_async
 from sentry_sdk import push_scope, set_context
 
+from api.utils.aiohttp import get_aiohttp_session
+from api.utils.asyncio import fire_and_forget
 from api.utils.image_proxy.exception import UpstreamThumbnailException
 from api.utils.image_proxy.extension import get_image_extension
 from api.utils.image_proxy.photon import get_photon_request_params
@@ -75,11 +80,40 @@ def get_request_params_for_extension(
     )
 
 
-def get(
+@sync_to_async
+def _tally_response(
+    tallies,
+    media_info: MediaInfo,
+    month: str,
+    domain: str,
+    response: aiohttp.ClientResponse,
+):
+    """
+    Tally image proxy response without waiting for Redis to respond.
+
+    Pulled into a separate function to help reduce overload when skimming
+    the `get` function, which is complex enough as is.
+    """
+    tallies.incr(f"thumbnail_response_code:{month}:{response.status}"),
+    tallies.incr(
+        f"thumbnail_response_code_by_domain:{domain}:" f"{month}:{response.status}"
+    )
+    tallies.incr(
+        f"thumbnail_response_code_by_provider:{media_info.media_provider}:"
+        f"{month}:{response.status}"
+    )
+
+
+_UPSTREAM_TIMEOUT = aiohttp.ClientTimeout(15)
+
+
+async def get(
     media_info: MediaInfo,
     request_config: RequestConfig = RequestConfig(),
 ) -> HttpResponse:
     """
+    Retrieve proxied image from site accelerator.
+
     Proxy an image through Photon if its file type is supported, else return the
     original image if the file type is SVG. Otherwise, raise an exception.
     """
@@ -88,9 +122,10 @@ def get(
 
     logger = parent_logger.getChild("get")
     tallies = django_redis.get_redis_connection("tallies")
+    tallies_incr = sync_to_async(tallies.incr)
     month = get_monthly_timestamp()
 
-    image_extension = get_image_extension(image_url, media_identifier)
+    image_extension = await get_image_extension(image_url, media_identifier)
 
     headers = {"Accept": request_config.accept_header} | HEADERS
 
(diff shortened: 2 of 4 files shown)


api/test/conftest.py
```diff
@@ -1,11 +1,34 @@
...
+
+    def _get_new_loop() -> asyncio.AbstractEventLoop:
+        loop = asyncio.new_event_loop()
+        loops.append(loop)
+        return loop
+
+    yield _get_new_loop
+
+    for loop in loops:
+        loop.close()
```
sarayourfriend, line 20: I've moved this here to unify the async support fixtures into a single place.

api/test/conftest.py
```diff
@@ -1,11 +1,34 @@
...
+
+
+@pytest.fixture(scope="session")
+def session_loop() -> asyncio.AbstractEventLoop:
+    loop = asyncio.new_event_loop()
+    yield loop
+    loop.close()
```
sarayourfriend, line 27: This can't use `get_new_loop` because `get_new_loop` is intentionally function scoped, whereas the session loop, as its name implies, is scoped to the session, and therefore cannot depend on the function scoped fixture.

api/api/utils/asyncio.py
```diff
@@ -0,0 +1,29 @@
...
+from collections.abc import Awaitable
+
+
+parent_logger = logging.getLogger(__name__)
+
+
+_fire_and_forget_logger = parent_logger.getChild("fire_and_forget")
+
+
+def fire_and_forget(awaitable: Awaitable) -> None:
```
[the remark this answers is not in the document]
sarayourfriend: For folks reviewing this, would `do_not_wait`, `do_not_wait_for`, `bypass_await`, or something along those lines be descriptive enough? I'm also toying with `schedule_task`, which is very technically correct and potentially has a similar "read the doc string" effect by way of being a little broad?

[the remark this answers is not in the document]
obulat: How about `do_not_await`? Other than that, I think `do_not_wait_for` seems the clearest option here.

api/api/utils/image_proxy/__init__.py
```diff
@@ -106,26 +141,36 @@ def get(
     )
 
     try:
-        upstream_response = requests.get(
+        # todo: refactor to use aiohttp shared session
```
obulat, line 144: Do we have an issue for this?

api/api/utils/image_proxy/__init__.py
```diff
@@ -88,9 +122,10 @@ def get(
 
     logger = parent_logger.getChild("get")
     tallies = django_redis.get_redis_connection("tallies")
+    tallies_incr = sync_to_async(tallies.incr)
```
obulat, line 125: Sorry I'm missing a lot of sync-to-async context since this is the first such PR that I'm reviewing. Why do we not use `async def`, but  need an `sync_to_async` for tallies?

api/api/utils/image_proxy/__init__.py
```diff
@@ -75,11 +80,40 @@ def get_request_params_for_extension(
...
+_UPSTREAM_TIMEOUT = aiohttp.ClientTimeout(15)
+
+
+async def get(
     media_info: MediaInfo,
     request_config: RequestConfig = RequestConfig(),
 ) -> HttpResponse:
     """
+    Retrieve proxied image from site accelerator.
+
```
obulat, line 116: It seems that we should prioritize the issue of renaming all instances of Photon to `site accelerator` to keep consistency: a single function's docstrings use both names now.

obulat (approved):
Tested well locally, all Redis keys are updated correctly, and the search keeps working as expected. Awesome work!

api/api/utils/asyncio.py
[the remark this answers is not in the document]
sarayourfriend: `do_not_wait_for` it is! Thanks, Olga.

api/api/utils/image_proxy/__init__.py
sarayourfriend: We are using `async def`, unless I unintentionally missed it somewhere. `async def` is only necessary when awaiting, though.

`sync_to_async` is necessary to prevent context issues in the Redis client, which assumes that only one thread is every touching the connection pool. `sync_to_async` creates a safe context for such synchronous functions. Relevant to Django ORM as well. [From Django's documentation](https://django.readthedocs.io/en/stable/topics/async.html#asgiref.sync.sync_to_async):

> The reason this is needed in Django is that many libraries, specifically database adapters, require that they are accessed in the same thread that they were created in. Also a lot of existing Django code assumes it all runs in the same thread, e.g. middleware adding things to a request for later use in views.

As an additional point of clarification, Django's own async cache support just wraps the synchronous versions of functions with `sync_to_async`, rather than using fully async clients. In the future we could switch to an asynchronous Redis client (like https://github.com/Andrew-Chen-Wang/django-async-redis, but I'm not sure that it's 100% stable), but for now it's not necessary and Django's own caching behaviour would still use the `sync_to_async` adapter approach for this anyway.

So the reason here is to create a new async version of `tallies.incr` that we can call later with `await`.

sarayourfriend: Let me know if I answered your question or if I misunderstood. The Django documentation on async adaption is a good read and important for understanding how and when it's necessary to adapt synchronous functions to an async context.

api/api/utils/image_proxy/__init__.py
sarayourfriend: Good point. This is such a significant change already, I'll just switch this back to use Photon or maybe just avoid mention of the service used. We already have an issue to cover this (#2542), and it's better not to mix the names up until we're actively fixing it everywhere.

sarayourfriend: Fixed in https://github.com/WordPress/openverse/pull/3388/commits/1d6ebd52f5759a93a7c7cae346faef8cc3726f02

api/api/utils/image_proxy/__init__.py
sarayourfriend: Cleaned up in https://github.com/WordPress/openverse/pull/3388/commits/1d6ebd52f5759a93a7c7cae346faef8cc3726f02

krysal (approved):
Everything works the same for me in appearance, which is excellent! 😄 Exciting movement!

api/test/unit/utils/test_image_proxy.py
```diff
@@ -56,8 +58,26 @@ def auth_key():
...
+    def do(*args, **kwargs):
+        try:
+            res = session_loop.run_until_complete(_photon_get(*args, **kwargs))
+            return res
+        finally:
+            tasks = asyncio.all_tasks(session_loop)
+            for task in tasks:
+                session_loop.run_until_complete(task)
+
+    yield do
```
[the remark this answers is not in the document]
sarayourfriend: Yes, I meant `do_not_wait_for`. I wrote this comment before Olga and I had decided on the better name for it. Everything is still accurate in the comment aside from the name.

sarayourfriend:
That sentry issue is fine. It only looks like a new issue because requests threw an `OSError` and aiohttp throws a `ClientOSError`, so our error handling, which always reports the first and then only ever the nth of an error in the proxy handler, thinks they're all new errors.

PR #3381, opened 2023-11-21
Fix dynamically excluded providers caching

obulat:
## Fixes

Fixes #3380 by @obulat

## Description

This PR saves a list of `provider_identifier`s for the dynamically-excluded providers to the redis cache, instead of saving the QuerySet.
It also adds two unit tests for the specific functions, but there are also several other tests that test dynamically excluded providers using both cache and the `ContentProvider` model saved in the database.

## Testing Instructions

Run the app, go to `http://localhost:50280/admin`, select one of the `ContentProvider` models and check its "Hide Content" property.
Screenshot 2023-11-21 at 5 22 01 PM

 Execute a search (http://localhost:50280/v1/images). Check Redis value for `:1:filtered_providers` (I used the RedisInsights GUI): 

Screenshot 2023-11-21 at 5 23 35 PM

Also, check that on `main`, if you set "Hide content" to true for one of the providers, the value saved in Redis should look like the one in the related issue instead of a list of filtered providers.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
[...]

PR #3399, opened 2023-11-27
Revert "Fix dynamically excluded providers caching"

sarayourfriend:
Reverts WordPress/openverse#3381

https://openverse.sentry.io/share/issue/65f52cf79486429c86ad957532094188/

and 

https://openverse.sentry.io/share/issue/d5341bad88cf421081b2551b36ef8c33/


dhruvkb (approved):
Sad to see a revert, is the resolution for the errors too complex to rollout as a hotfix?

sarayourfriend:
> Sad to see a revert, is the resolution for the errors too complex to rollout as a hotfix?

I haven't looked into it, honestly, and there are multiple parts of the code with problems, so in the interest of just being able to deploy the production API I think a revert is better. The original code change didn't address a critical problem, so there's no reason to rush to a solution.

obulat (approved):

PR #3408, opened 2023-11-28
Use DB to augment ES hits for related media with required info

### LINKED ISSUE #3403: Related endpoint does not return all of the necessary propertes
## Description

The related results are missing required properties such as `license_version`.

## Reproduction

1. Check that the image single result and search result have the necessary properties including the `license_version`: (https://api.openverse.engineering/v1/images/cf016656-2584-4f6e-827e-386d124a2d0a/ and https://api.openverse.engineering/v1/images/?q=Eastern%20Phoebe%20(Sayornis%20phoebe)%202022-04-28%20(2))
2. See that the same image has `null` for `license_version`, `height`, `width`, `attribution`, `creator_url` when it's returned as a related result.

## Additional context

I guess the related endpoint does not get the search context.

dhruvkb:
## Fixes

Fixes #3403 by @obulat

## Description

This PR uses the DB to augment ES hits for related media with info that's not in ES. It uses the exact same pattern (and for that matter, code) as `get_media_results`.

Also updated the related media tests to check for the presence of these fields.

## Testing Instructions

Try, and fail, to repro the issue described in #3403 on this branch.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -267,6 +267,10 @@ class MediaViewSet(AsyncViewSetMixin, AsyncAPIView, ReadOnlyModelViewSet):
 
         serializer_context = self.get_serializer_context()
 
+        serializer_class = self.get_serializer()
+        if serializer_class.needs_db:
+            results = self.get_db_results(results)
+
         serializer = self.get_serializer(results, many=True, context=serializer_context)
         return self.get_paginated_response(serializer.data)
 
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -105,6 +105,18 @@ class MediaViewSet(AsyncViewSetMixin, AsyncAPIView, ReadOnlyModelViewSet):
         return req_serializer
 
     def get_db_results(self, results):
+        """
+        Map ES hits to ORM model instances.
+
+        ORM instances have all necessary info needed for serializers whereas ES
+        hits only contain the subset of fields needed for indexing and search.
+        This function issues one query to the DB, using the ``identifier`` field
+        which is both unique and indexed, so it's quite performant.
+
+        :param results: the list of ES hits
+        :return: the corresponding list of ORM model instances
+        """
+
         identifiers = []
         hits = []
         for hit in results:
--- a/api/test/media_integration.py
+++ b/api/test/media_integration.py
@@ -227,6 +227,10 @@ def related(fixture):
             or result["creator"] == item["creator"]
         ), f"{terms_set} {get_terms_set(result)}/{result['creator']}-{item['creator']}"
 
+        assert result["license_version"] is not None
+        assert result["attribution"] is not None
+        assert result["creator_url"] is not None
+
 
 def sensitive_search_and_detail(media_type):
     search_res = requests.get(


api/test/media_integration.py
```diff
@@ -227,6 +227,10 @@ def get_terms_set(res):
             or result["creator"] == item["creator"]
         ), f"{terms_set} {get_terms_set(result)}/{result['creator']}-{item['creator']}"
 
+        assert result["license_version"] is not None
+        assert result["attribution"] is not None
+        assert result["creator_url"] is not None
+
```
obulat, line 233: This is a high priority PR that fixes an important issue, so I'm not proposing the change here, but I think we should use something like https://github.com/maticardenas/drf-openapi-tester/ to test that _all_ responses conform to the OpenAPI schema that we have. I'll open an issue for that.

api/api/views/media_views.py
```diff
@@ -267,6 +267,10 @@ def related(self, request, identifier=None, *_, **__):
 
         serializer_context = self.get_serializer_context()
 
+        serializer_class = self.get_serializer()
+        if serializer_class.needs_db:
+            results = self.get_db_results(results)
```
obulat, line 272: Do we have any idea of how much getting the db results affects the performance of the searches?
I wonder how we can measure if it's better to get this data from the database, or to keep it in ES indexes and get it from there?

obulat (approved):
Thank you for such a quick fix, @dhruvkb ! It works well and matches what we do on another endpoint.
I added non-blocking questions inline.

api/test/media_integration.py
dhruvkb: That is a great idea!

api/api/views/media_views.py
dhruvkb: It is a single query to the DB for all results using the field `identifier` that's  both `unique=True` and `index=True`. I don't think it is inefficient. Personally I prefer one the way it is rn because we have to query the DB anyway for relational info that can't be stored in ES. Also the amount of relational info will only increase with new projects slated for 2024.

obulat: Thank you for the explanation, @dhruvkb , makes sense!

sarayourfriend: @dhruvkb 100% in agreement here as well, that would be a good thing to document somewhere as the "rationale" of the serialiser approach, prefereably in code or, if not there, then in the API docs, in something of a "design decisions" document. It's similar to the the ES document `_id` not being the same as `identifier`, for example, in that it's easy to forget why this is the "right" way to do it and then re-ask this same question (or for new folks to ask this question). A "design decisions" doc I think would be a great thing to have in general (to summarise my point :grinning:).

sarayourfriend: Actually, rather than a new documentation page, pulling this behaviour into a function that the two view methods can call and adding the rationale in the doc string there would be more resilient to future changes.

dhruvkb: Both views ultimately use the `get_db_results` function, I've documented that function in 324fe92. I think the `needs_db` mechanism is largely moot because almost all operations ultimately need the DB and once that is simplified, `get_db_results` will be the common functionality entirely.

krysal (approved):
LGTM!  Tested with the first general image result and in `main` the mentioned fields are `null` and here are correctly filled ✅ 

http://localhost:50280/v1/images/3d0af15e-5f3a-42f9-8833-1a3d6baf90c3/related/

Issue #3080, opened 2023-09-29
Handle slightly invalid license URL


dhruvkb:
## Description

Airflow raises error when it encounters a URL like `https://creativecommons.org/licenses/by-nd-nc/2.0/jp/`. This is fair because the URL does not exist. But this is only a typo because the `ND` and `NC` components are swapped in order.

The fixed URL `https://creativecommons.org/licenses/by-nd-nc/2.0/jp/` points to a valid license deed. This fix can be made by the workflow itself to avoid these exceptions.

## Screenshots

<img width="656" alt="Screenshot 2023-09-29 at 11 29 10 AM" src="https://github.com/WordPress/openverse/assets/16580576/51018809-a285-483a-95f2-e076e8f990d3">

## Additional context

Although not observed by me, this exception can occur with other multi-component license names like CC BY-NC-SA etc. in which case rearranging the components to the "right" order might fix the broken URLs.

sarayourfriend:
Could this be fixed by breaking the licence identifier into the CC "licence elements" and re-building the URL when needed?

stacimc:
#3440 explains why this was not the case specifically with Flickr (from Flickr we receive just the components, and were actually incorrectly applying the invalid url on our side).

I think this is still a possible issue, but is not currently being observed by any of our providers so I am re-prioritizing this to `low`. If it does occur, we will get alerts from Airflow and can bump this back up.

PR #3440, opened 2023-12-01
Fix incorrectly parsed license url by-nc-nd for Flickr

stacimc:
## Fixes

I noticed this problem while testing locally in Flickr.

## Description

In the Flickr DAG, we often get the following error (which is currently silenced in production):

```
Failed to rewrite URL: https://creativecommons.org/licenses/by-nd-nc/2.0/jp/
```

That URL is indeed invalid. These urls, however, are correct (note the swapped components, `by-nc-nd` vs `by-nd-nc`):

- https://creativecommons.org/licenses/by-nc-nd/2.0/jp
- https://creativecommons.org/licenses/by-nc-nd/2.0/

In the Flickr DAG, for each record we get a `license` and `license_version` from the Flickr API. We use our [`get_license_info`](https://github.com/WordPress/openverse/blob/adb56f128ca9f0835e0e882fbe960539a0076a37/catalog/dags/common/licenses/licenses.py#L41) utility to parse out the `license_url` (which ends up being stored in the `meta_data` column).

`get_license_info` attempts to look up the license url in a reverse mapping of (license, license_version) pair to url, which is built [here](https://github.com/WordPress/openverse/blob/adb56f128ca9f0835e0e882fbe960539a0076a37/catalog/dags/common/licenses/constants.py#L119) by merging two hard-coded dicts of simple cases and special cases. We had an entry in the _SPECIAL_CASE_LICENSE_PATHS mapping `by-nd-nc/2.0/jp` (the invalid url) to `('by-nc-nd', '2.0')`, but we _also_ had an entry in _SIMPLE_LICENSE_PATHS mapping the correct, generic license url `by-nc-nd/2.0` to the same `('by-nc-nd', '2.0')` pair. Because special cases are applied after simple cases when building the mappings, the special case (with the invalid url) overrides the simple (with the valid). So when Flickr calls `get_license_info` with that license/version pair, it always gets the invalid URL from the mapping and then fails when trying to validate it.

**This PR handles this by just getting rid of that invalid special case**. When a DAG such as flickr ingests a record with license `by-nc-nd` and version `2.0` and uses `get_license_info` to look up the URL, it will now always return the valid, generic URL https://creativecommons.org/licenses/by-nc-nd/2.0/

### Mapping from license url to license/version

Note that DAGs may also ingest just the URL from their provider API, and use `get_license_info` instead to parse out the type and version (for which we maintain the mapping in the other direction, defined [here](https://github.com/WordPress/openverse/blob/adb56f128ca9f0835e0e882fbe960539a0076a37/catalog/dags/common/licenses/constants.py#L109)).
* If passed the incorrect url, on `main` it will return `('by-nc-nd', '2.0')` for the license/version pair __but preserve the invalid URL in meta_data without validating it__
* On this branch, it will instead **raise an error** (because the invalid url is no longer present in the mapping)

This change should be okay (and in fact, what we *want* to happen), because I checked the production catalog and found 0 records outside of Flickr that have this URL:
```sql
openledger> select count(*) from image where meta_data->>'license_url' = 'https:/
 /creativecommons.org/licenses/by-nd-nc/2.0/jp' and provider != 'flickr';
l+-------+
| count |
|-------|
| 0     |
+-------+
openledger> select count(*) from audio where meta_data->>'license_url' = 'https:/
 /creativecommons.org/licenses/by-nd-nc/2.0/jp';
+-------+
| count |
|-------|
| 0     |
+-------+
```

If we do encounter this in the future (a DAG that gets the slightly invalid URL from its provider), then #3080 could be implemented to handle that situation flexibly. But I do not believe it is currently a problem.

TL;DR, this change should fix the error in Flickr and should *not* cause problems with any other DAGs.  

### Backfilling old Flickr records

A final note: We *do* have many Flickr records containing the invalid `license_url` in `meta_data`, with the license components swapped. I haven’t looked too deeply into the history of how this happened, but it seems in a previous implementation of either the Flickr DAG or the license utilities, we allowed this license_url to get written to records.

Here's an example you can see in our API: https://api.openverse.engineering/v1/images/85c1bedb-a32c-4ff7-8e97-0d04d55cfd31/ The license url is invalid, and the attribution uses this invalid url. For thoroughness, I checked out this image [on Flickr](https://www.flickr.com/photos/47651805@N00/3834147370) and verified that the license given there is the generic https://creativecommons.org/licenses/by-nc-nd/2.0/.

In conclusion: after merging this change, we’ll need to backfill these flickr records. We can do that via the `batched_update` DAG with this config:

```
{
    "query_id": “update-invalid-flickr-licenses”,
    "table_name": “image”,
    "select_query": "WHERE provider='flickr' and meta_data->>'license_url' = 'https:/
 /creativecommons.org/licenses/by-nd-nc/2.0/jp'",
    "update_query": "SET meta_data->>'license_url' = 'https:/
 /creativecommons.org/licenses/by-nc-nd/2.0'",
    "batch_size": 10000,
    "update_timeout": 3600,
    "resume_update": false,
    "dry_run": false
}
```

We should also remove the `SILENCED_SLACK_NOTIFICATIONS` configuration in prod that's currently silencing this issue for Flickr. Unfortunately, I think this has been incorrectly silenced and we have probably been failing to ingest many valid Flickr records during that time.

## Testing Instructions

Try running Flickr locally using `date`: 2023-11-30. On main, it should fail very quickly with the `InvalidLicense` error (because it is not silenced in local environments). Check out this branch and re-run the dagrun for a few minutes to see that it passes the step successfully.

This isn't really unit-testable as far as I can see, but please verify my explanation in the description makes sense and that I am not misunderstanding something fundamental!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/licenses/constants.py
+++ b/catalog/dags/common/licenses/constants.py
@@ -79,7 +79,6 @@ _SPECIAL_CASE_LICENSE_PATHS = {
     # set the license and/or version manually, and for which we can
     # recover the path from a valid license_, license_version pair.
     "licenses/by-nd-nc/1.0": ("by-nc-nd", "1.0"),
-    "licenses/by-nd-nc/2.0/jp": ("by-nc-nd", "2.0"),
     "licenses/publicdomain": ("publicdomain", NO_VERSION),
     "publicdomain/mark/1.0": ("pdm", "1.0"),
     "publicdomain/zero/1.0": ("cc0", "1.0"),


krysal (approved):
I confirmed that it fails on `main` but passes in this branch.

> We should also remove the SILENCED_SLACK_NOTIFICATIONS configuration in prod that's currently silencing this issue for Flickr.

I was about to ask/suggest that as well. So everything looks covered here! Good find, and thank you for the thoughtful explanation :)

AetherUnbound (approved):
Thank you for the very thorough PR description!! What a ratio to the change of code hehe - tested on both main and this branch, and it failed and then worked as expected!

stacimc:
The Slack notification is unsilenced. I tested the batched_update locally and it is ready to go, but I will until next week so the currently running popularity refresh can finish, as the backfill is not urgent.

PR #3444, opened 2023-12-03
Add API schema tests using schemathesis

### LINKED ISSUE #3446: Use property-based testing for the API
## Problem

There are several issues about 500 errors being returned by some inputs for some endpoints (#3342, #1232). The OpenAPI specs also don't match the responses fully.

## Description

Add property-based testing to the API responses using [Schemathesis](https://schemathesis.readthedocs.io/en/stable/index.html). This will allow us to use the [following checks](https://schemathesis.readthedocs.io/en/stable/cli.html#how-are-responses-checked): 
- `not_a_server_error`. The response has 5xx HTTP status;
- `status_code_conformance`. The response status is not defined in the API schema;
- `content_type_conformance`. The response content type is not defined in the API schema;
- `response_schema_conformance`. The response content does not conform to the schema defined for this specific response;
- `response_headers_conformance`. The response headers do not contain all defined headers.

## Additional context

In a recent PR, we had to add a test to check if the `related` endpoint returns all of the necessary properties. This [made me think](https://github.com/WordPress/openverse/pull/3408/files#r1408661163) of some OpenAPI specs-based tests. There is a DRF-specific library, https://github.com/maticardenas/drf-openapi-tester/. But I think a better alternative is the `schemathesis` which allows for testing more things.

obulat:
## Fixes

Fixes #3446 by @obulat

## Description

~~This PR adds the `schemathesis` step to the API testing in the CI.~~
~~The checks run for 2 flags: `response_headers_conformance` and  `not_a_server_error`. I changed the related endpoint to fix the `not_a_server_error` problem.~~

~~The Schemathesis docs say that running the tests as part of the Pytest test suite is possible, but I couldn't get it to run, so I added the CLI. I would be happy if someone could find a way to run the tests as part of the test suite.~~
This PR adds Schemathesis tests running as part of the pytest suite (thanks to @sarayourfriend ). It checkes that all of the response codes are documented in the openapi schema correctly.

The code is also updated to make the tests pass.

## Testing Instructions

Check out this PR and run `just api/test`. The tests should run the schema tests and should pass.

~~If you revert the changes to the related function and run `just api/test-schema`, you should see the following output:~~
This is for the first implementation of this as part of the just script in CI.

<details>
<summary> Schemathesis output </summary>
```
============================================================ Schemathesis test session starts ===========================================================
Schema location: file:///Users/obulat/Documents/openverse/openverse/api/openapi.json
Base URL: http://localhost:50280
Specification version: Open API 3.0.3
Workers: 1
Collected API operations: 24

GET /v1/audio/ .                                                                                                                                   [  4%]
GET /v1/audio/{identifier}/ .                                                                                                                      [  8%]
GET /v1/audio/{identifier}/related/ .                                                                                                              [ 12%]
POST /v1/audio/{identifier}/report/ .                                                                                                              [ 16%]
GET /v1/audio/{identifier}/thumb/ .                                                                                                                [ 20%]
GET /v1/audio/{identifier}/waveform/ .                                                                                                             [ 25%]
GET /v1/audio/source/{source}/ .                                                                                                                   [ 29%]
GET /v1/audio/source/{source}/creator/{creator}/ .                                                                                                 [ 33%]
GET /v1/audio/stats/ .                                                                                                                             [ 37%]
GET /v1/audio/tag/{tag}/ .                                                                                                                         [ 41%]
POST /v1/auth_tokens/register/ .                                                                                                                   [ 45%]
POST /v1/auth_tokens/token/ .                                                                                                                      [ 50%]
GET /v1/images/ .                                                                                                                                  [ 54%]
GET /v1/images/{identifier}/ .                                                                                                                     [ 58%]
GET /v1/images/{identifier}/related/ .                                                                                                             [ 62%]
POST /v1/images/{identifier}/report/ .                                                                                                             [ 66%]
GET /v1/images/{identifier}/thumb/ .                                                                                                               [ 70%]
GET /v1/images/{identifier}/watermark/ .                                                                                                           [ 75%]
GET /v1/images/oembed/ .                                                                                                                           [ 79%]
GET /v1/images/source/{source}/ .                                                                                                                  [ 83%]
GET /v1/images/source/{source}/creator/{creator}/ .                                                                                                [ 87%]
GET /v1/images/stats/ .                                                                                                                            [ 91%]
GET /v1/images/tag/{tag}/ .                                                                                                                        [ 95%]
GET /v1/rate_limit/ .                                                                                                                              [100%]

======================================================================== SUMMARY ========================================================================

Performed checks:
    response_headers_conformance                    2103 / 2103 passed          PASSED 

WARNINGS:
  - Most of the responses from `POST /v1/auth_tokens/token/` have a 401 status code. Did you specify proper API credentials?
  - Most of the responses from `GET /v1/rate_limit/` have a 403 status code. Did you specify proper API credentials?

Tip: Use the `--report` CLI option to visualize test results via Schemathesis.io.
We run additional conformance checks on reports from public repos.

================================================================== 24 passed in 29.73s ==================================================================
============================================================ Schemathesis test session starts ===========================================================
Schema location: file:///Users/obulat/Documents/openverse/openverse/api/openapi.json
Base URL: http://localhost:50280
Specification version: Open API 3.0.3
Workers: 1
Collected API operations: 24

GET /v1/audio/ .                                                                                                                                   [  4%]
GET /v1/audio/{identifier}/ .                                                                                                                      [  8%]
GET /v1/audio/{identifier}/related/ F                                                                                                              [ 12%]
POST /v1/audio/{identifier}/report/ .                                                                                                              [ 16%]
GET /v1/audio/{identifier}/thumb/ .                                                                                                                [ 20%]
GET /v1/audio/{identifier}/waveform/ .                                                                                                             [ 25%]
GET /v1/audio/source/{source}/ .                                                                                                                   [ 29%]
GET /v1/audio/source/{source}/creator/{creator}/ .                                                                                                 [ 33%]
GET /v1/audio/stats/ .                                                                                                                             [ 37%]
GET /v1/audio/tag/{tag}/ .                                                                                                                         [ 41%]
POST /v1/auth_tokens/register/ .                                                                                                                   [ 45%]
POST /v1/auth_tokens/token/ .                                                                                                                      [ 50%]
GET /v1/images/ .                                                                                                                                  [ 54%]
GET /v1/images/{identifier}/ .                                                                                                                     [ 58%]
GET /v1/images/{identifier}/related/ F                                                                                                             [ 62%]
POST /v1/images/{identifier}/report/ .                                                                                                             [ 66%]
GET /v1/images/{identifier}/thumb/ .                                                                                                               [ 70%]
GET /v1/images/{identifier}/watermark/ .                                                                                                           [ 75%]
GET /v1/images/oembed/ .                                                                                                                           [ 79%]
GET /v1/images/source/{source}/ .                                                                                                                  [ 83%]
GET /v1/images/source/{source}/creator/{creator}/ .                                                                                                [ 87%]
GET /v1/images/stats/ .                                                                                                                            [ 91%]
GET /v1/images/tag/{tag}/ .                                                                                                                        [ 95%]
GET /v1/rate_limit/ .                                                                                                                              [100%]

======================================================================== FAILURES =======================================================================
__________________________________________________________ GET /v1/audio/{identifier}/related/ __________________________________________________________
1. Test Case ID: 1IqpOZ

- Server error

[500] Internal Server Error:

    `{"detail":"Could not find items."}`

Reproduce with: 

    curl -X GET http://localhost:50280/v1/audio/00000000-0000-4000-8000-000000000000/related/

Or add this option to your command line parameters: --hypothesis-seed=110085890912020084465361027240064100643
__________________________________________________________ GET /v1/images/{identifier}/related/ _________________________________________________________
1. Test Case ID: sLiRsm

- Server error

[500] Internal Server Error:

    `{"detail":"Could not find items."}`

Reproduce with: 

    curl -X GET http://localhost:50280/v1/images/00000000-0000-4000-8000-000000000000/related/

Or add this option to your command line parameters: --hypothesis-seed=121590553949486563256538290782721953302
======================================================================== SUMMARY ========================================================================

Performed checks:
    not_a_server_error                    1903 / 1909 passed          FAILED 

WARNINGS:
  - Most of the responses from `POST /v1/auth_tokens/token/` have a 401 status code. Did you specify proper API credentials?
  - Most of the responses from `GET /v1/rate_limit/` have a 403 status code. Did you specify proper API credentials?
```
</details>

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/justfile
```diff
@@ -121,6 +121,11 @@ doc-test: wait-up
       --fail \
       'http://localhost:50280/v1/schema/' > openapi.json
 
+# Run the API schema tests using schemathesis
+test-schema: wait-up
+    just doc-test
```
dhruvkb, line 126: `doc-test` can be changed to a pre-requisite recipe here because we depend on the output of that recipe `openapi.json` to run this step.
```suggestion
test-schema: wait-up doc-test
```

api/justfile
```diff
@@ -121,6 +121,11 @@ doc-test: wait-up
...
+    python -m pip install schemathesis && st run --checks response_headers_conformance openapi.json --base-url http://localhost:50280 && st run --checks not_a_server_error openapi.json --base-url http://localhost:50280
```
dhruvkb, line 127: This step seems installs the `schemathesis` package on the host Python environment, which is not ideal. It's also not guaranteed to work on macOS for example, where there is no unversioned `python` (but `python3` is present across all platforms afaik).

Screenshot 2023-12-11 at 7 43 24 AM

The [Getting Started](https://schemathesis.readthedocs.io/en/stable/getting-started.html#getting-started) guide lists a bunch of ways to run this, including a Docker image. That seems like a more cross-platform way to run the check that does not affect the host.

.github/workflows/ci_cd.yml
```diff
@@ -404,6 +404,9 @@ jobs:
           name: api_logs
           path: api_logs
 
+      - name: Run API schema tests
+        run: just api/test-schema
```
dhruvkb, line 408: This would fit better as a matrix entry in the `django-checks` job below, as those are all checks based on `just` recipes. Additionally the recipe this depends on i.e. `doc-test` is also a job in that matrix.

dhruvkb (changes requested):
The changes to the API code LGTM! But I have some notes about improvements to the execution of the test and provisioning of `schemathesis`.

AetherUnbound:
Drafting this until feedback can be addressed

sarayourfriend:
Okay! All tests except the auth token ones are passing.

You can run the schemathesis tests with `just api/test -k test_schema`, because they're just part of the normal API test suite.

I was able to do some cleanup here of the error serializers, which aren't necessary, because drf-spectuacular understands DRF's exception classes. I found this because the validation error serializer did not match the schema of 400 responses, which schemathesis didn't like.

Right now I've had repeated failures with the creator path, because schemathesis keeps trying to pass the newline character in percent encoding to the creator path parameter. I'm not sure why it's doing that, or how to prevent. It also seems like it might be flaky.

obulat:
@sarayourfriend, I really like having the tests run by pytest. However, having seen the amount of warnings of deprecations coming from them, I wonder if the pytest part of Schemathesis is not maintained as well as the main repository. Hopefully, it won't fall behind more, and not make us regret moving this test to pytest runner.

sarayourfriend:
I didn't notice the deprecation warnings :thinking:

api/api/serializers/media_serializers.py
```diff
@@ -71,11 +70,13 @@ def validate_page_size(self, value):
...
             try:
                 validator(value)
-            except ValidationError as e:
-                raise NotAuthenticated(
-                    detail=e.message,
-                    code=e.code,
-                )
+            except Exception as e:
+                # TODO: figure out why except ValidationError doesn't catch this
+                if type(e).__name__ == "ValidationError":
```
sarayourfriend, line 75: Is this todo still valid?

sarayourfriend (commented):
This LGTM! There's a TODO left in the code, so not approving just yet, just to clarify whether that is still a dangling todo item, if we need a new issue for it, or if it can be solved here.

api/api/serializers/media_serializers.py
obulat: Yes, it's still valid. I could not figure out why this didn't work:
```py
import rest_framework.exception.ValidationError
catch ValidationError
```
Even though the name of this error is `ValidationError`.

I need help with this

sarayourfriend: It's probably the Django rather than DRF class. The DRF class does not inherit from the Django classes:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L143

https://github.com/django/django/blob/main/django/core/exceptions.py#L134

sarayourfriend: DRF has the same difficulty itself, it turns out. Here they use `except (ValidationError, DjangoValidationError)` as a "catch all" for the two potential approaches: https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L453

obulat: Thank you, @sarayourfriend! I updated the code and removed the `TODO` comment.

One thing I'm not sure about in this PR is whether the tests do all of the checks.
This is a list of checks that Schemathesis can perform: [`not_a_server_error`, `status_code_conformance`, `content_type_conformance`, `response_headers_conformance`, `response_schema_conformance`]. The Pytest command we use is supposed to do all of the checks by default [^1]. 

In the first iteration of this PR, I opened 2 PRs. This one only added the test for response status codes, and #3447 added the changes for checking the `response_schema_conformance` and other checks. There, for instance, I had to add a new serializer for audio alt_files to make the responses conform the schema. Here, we don't seem to have such a problem. Is it because something changed in the responses so that they are now conformant, or because the check is not run correctly?

[^1]: https://schemathesis.readthedocs.io/en/stable/python.html#how-are-responses-checked

sarayourfriend: The spec in this branch has `any or null` for alt files. That matches production: https://api.openverse.engineering/v1/#tag/audio/operation/audio_detail

I didn't run into that issue when I was working on the changes I contributed to this PR, and I don't know how I would reproduce it. Based on the schema in main on this PR, I can't think of any reason schemathesis would ever error on any value at all in `alt_files`, unless there was a version of this PR that changed the schema type for that field away from `any or null`.

The schema type checks are working fine, as far as I can tell. The changes to the stats endpoint documentation are proof of that (schemathesis will fail if you remove the `many=True` because without that the schema does not wrap the response in a list, which the endpoint always returns).

sarayourfriend:
As I've contributed heavily to the changes in this PR, I am _not_ going to review it. I can't figure out how to remove myself from the requested reviewers, though.

api/api/serializers/media_serializers.py
obulat: Thank you! I didn't realize that `audio_detail` is set to `any or null`. I added the `alt_files` serializer from the other PR to make the types more accurate.
I didn't realize that you've also added the `many=True` to this PR. Those were also causing the response shape mismatch error.

api/api/views/oauth2_views.py
```diff
@@ -182,20 +193,15 @@ def get(self, request, format=None):
 
         # TODO: Replace 403 responses with DRF `authentication_classes`.
```
AetherUnbound, line 194: Is this still accurate now that we're using `PermissionDenied`?

AetherUnbound (approved):
This looks great and the new tests work for me locally! I'm getting an unrelated failure:

```
FAILED test/integration/test_media_integration.py::test_collection_by_tag[audio] - assert 0 > 0
```

I'll look into it to try and figure out why but I don't think it's related to this work at all.

Regardless, I was baffled that the testing code was _so small_ 😮 🚀

AetherUnbound:
Actually, it seems like when I run `j api/test` the test fails, but I can run `j api/test -k test_media_integration` and all tests pass 🤔 is anyone else seeing this?

obulat:
@AetherUnbound, some tests rely on the sample data, and if your local sample data was changed in another PR (for instance, to add some special characters to the creator/tag values), the tests might fail until you update the indices.

> Regardless, I was baffled that the testing code was _so small_ 😮 🚀

Now I see why there's so much talk about property-based testing :)

api/api/views/oauth2_views.py
obulat: @dhruvkb, do you think it's still accurate, and could you explain what does "replacing with `authentication_classes`" mean? Asking you because it looks like you've added it in https://github.com/WordPress/openverse/pull/1876

dhruvkb: This is being addressed separately in #3663.

api/api/serializers/audio_serializers.py
```diff
@@ -114,6 +114,41 @@ class Meta:
 AudioHyperlinksSerializer = get_hyperlinks_serializer("audio")
 
 
+class AudioAltFileSerializer(serializers.Serializer):
```
dhruvkb, line 117: I think this can be defined as a subclass of `AudioSerializer` with different `Meta.fields`. I think it's worth exploring that possibility to reduce the code as the fields will not have to be defined manually.

api/api/utils/exceptions.py
```diff
@@ -1,4 +1,4 @@
-from rest_framework.serializers import ValidationError
+from rest_framework.exceptions import ValidationError
```
dhruvkb, line 1: Please correct me if I'm mistaken but I thought the practice was to raise `ValidationError` from `rest_framework.serializers` as per [these docs](https://www.django-rest-framework.org/api-guide/serializers/).

dhruvkb (approved):
LGTM 🎉 for the most part but I have some questions and some concerns. Since the additional checks from schemathesis are good to have, I'm not blocking the PR.

api/api/utils/exceptions.py
sarayourfriend: `serializers.ValidationError` is an alias for the `exceptions.ValidationError`. The only good reason to use `serializers.ValidationError`, as in the docs, is because the convention in _serializer definition modules_ is to `from rest_framework import serializers`, and so it's convenient and consistent to address it from the alias in that module. There's no reason not to use the exceptions module, though, and I think it makes particular sense in a file that otherwise has nothing to do with serialization (it's named `exceptions.py`).

At least, that's my interpretation of why the docs are written that way, and why I would choose to write code one way or the other, but I also don't think there's a reason for/against always using one or the other, and there's nothing in the docs you linked that explicitly implies you should use one or the other, aside from the examples using the alias (I suspect for the reason above).

api/api/views/oauth2_views.py
sarayourfriend: It's not part of that PR you've linked, Dhruv, to be clear. That PR handles a different thing entirely than permissions, it handles validation of credentials, not permissions. A small distinction, but to be clear, that does not have anything to do with _permissions_. 

This comment is still valid, because the manual `PermissionDenied` raising can be implemented in DRF Permissions class, and should be, to follow DRF conventions and separate the concerns.

api/api/serializers/audio_serializers.py
obulat: I tried fixing this in this PR, but got confused because making this a subclass of `AudioSerializer` would make it a recursive serializer. I'll open a new issue for this.

sarayourfriend (commented):
@obulat I had some comments staged on this PR, including a tip to fix the issue with the circular serializer. Can you create an issue to follow up on that, to create a model serializer instead of manually defining the videos?

obulat:
@sarayourfriend, sorry I only now read your comments here. 

I think we cannot use a model serializer because `AltAudioFile` is a child of `AltFile` class which is not a Django Model:

https://github.com/WordPress/openverse/blob/775c0df6845c58ea50b87b9856ca6cfb51fe50c5/api/api/models/media.py#L408-L418

@dhruvkb, is there a reason why the `AltFile` is not based on the `FileMixin` and thus a subclass of Model?

PR #3459, opened 2023-12-05
Add ESLint rules to mandate braces around if-else and switch-case

### LINKED ISSUE #3386: Add ESLint rules to mandate braces around `if`-`else` and `switch`-`case`
## Problem

Currently it's possible to write one-line `if`-`else` statements without braces. It's also possible to write `case` blocks inside `switch` without braces.

## Description

The lack of braces is easier to read initially but makes it possible to make errors when adding more statements to these constructs or updating code in the future.

Mandating braces with ESLint (with the possibility to autofix can prevent these problems.

## Additional context

This scope for improvement to our ESLint config was raised in the review for #3375.

bhavik001:
Fixes #3386 by @dhruvkb 

## Description
This PR incorporates changes to the ESLint configuration to enforce the use of braces in one-line if-else statements and case blocks inside switch statements. This update aims to enhance code readability and prevent potential issues in the future. The specific ESLint rules added are:

- curly: Enforce braces for all control statements.
- default-case: Enforce a default case in switch statements.
- eqeqeq: Enforce the use of strict equality (===) and inequality (!==).

For more details on ESLint rules, please refer to the [official ESLint documentation](https://eslint.org/docs/rules).

## Testing Instructions

1. Ensure ESLint is installed by running npm install eslint --save-dev.
2. Run ESLint on the project to check for adherence to the new rules: npx eslint .
3. If necessary, automatically fix fixable issues using: npx eslint --fix .

## Checklist

- [x] My pull request has a descriptive title.
- [x] My pull request targets the default branch of the repository (main) or a parent feature branch.
- [x] My commit messages follow best practices.
- [x] I added or updated documentation regarding the ESLint rule changes.
- [x] I ran the project locally and verified that there are no visible errors.


packages/eslint-plugin/src/rules/analytics-configuration.ts
```diff
@@ -213,6 +215,35 @@ export const analyticsConfiguration = OpenverseRule<Options, MessageIds>({
...
+          for (const switchCase of node.cases) {
+            if (!switchCase.consequent || switchCase.consequent.length === 0) {
+              context.report({
+                node: switchCase,
+                messageId: 'missingBraces',
+              });
+            }
+          }
+        }
+      },
```
dhruvkb, line 246: Although I can't say for sure without extensively testing it locally, this rule, from a first glance, seems to me to be implemented correctly. However, this specific file `packages/eslint-plugin/src/rules/analytics-configuration.ts` is supposed to be for our rules specifically related to the analytics area. You should put this in a separate, new file under `packages/eslint-plugin/src/rules/`.

dhruvkb (changes requested):
This is really great work @bhavik001, commendable really. Thank you for the contribution!

I have some notes, and some changes that may be helpful to pass the lint checks.

packages/eslint-plugin/src/rules/analytics-configuration.ts
sarayourfriend: Really we don't need this custom implementation at all. The ESLint unicorn plugin already has this rule, and we already use it. The unicorn implementation also has automatic fixes, which would solve some of the other issues in this PR as well caused by manually adding the braces.

Here's the unicorn rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/switch-case-braces.md

To enable it, add it to project's ESLint configuration here: https://github.com/WordPress/openverse/blob/8b9dabe011b133f7edfc84e59ddd58db57872d38/packages/eslint-plugin/src/configs/index.ts#L32-L36

```
"unicorn/switch-case-braces": ["error"],
```

It also won't incorrectly cause an error on fall-through cases like this implemention does here: https://github.com/WordPress/openverse/pull/3459/files#diff-b3a01ee15275f0b7ecd3aa87dec5fefd76e9f6d324a11042ce4c9bb3a4fb500eR428

sarayourfriend (changes requested):
Thanks for the contribution, however, we shouldn't use custom implementation of rules that already exist. I'm also confused, though, because the PR description says that the curly, eqeqeq, and default-case are also enabled. But they aren't. In particular `curly` wasn't used, but a custom implementation of it is proposed.

Please update the PR to only address the issue's specific requirements. Enable ESLint's built in `curly` rule and the unicorn plugin's `switch-case-braces` rule, then run ESLint with `--fix` (`pnpm run eslint --fix`) and commit the changes.

If you believe we should enable eqeqeq and default-case, please open separate a separate issue for them to explain the rationale. I strongly disagree that `default-case` is necessary in a project that has Typescript (TS already enforces that all switch conditions are handled). Same with eqeqeq, which TypeScript also enforces. It is not necessary to include them in the scope of this PR and it would only delay further review and merging the changeset.

Thank you again for the contribution. Once you've addressed the requested changes, I'll be happy to review the PR again and approve it.

bhavik001:
Hey @dhruvkb 
I don't know what happen with the repository I have forked to my github. I am not able to find it. So, I am forking it again and creating new pull request.

dhruvkb:
@bhavik001 it seems that instead of closing the PR and deleting your branch, you have accidentally deleted the entire repo.

Screenshot 2023-12-06 at 10 39 21 AM

bhavik001:
I don't know how it get deleted because I am still working on it. At the time of pushing the code I was getting the error and I saw the repository is not there.

PR #3482, opened 2023-12-07
Add external dag sensor utilities

stacimc:
## Fixes

Fast-follow to #3232

## Description

Several of our DAGs have concurrency checks that prevent them from running while some other DAG is running. Depending on the behavior we want, we need:
* A task that fails immediately if the external DAG is running
* A task that *waits* for the external DAG to complete

We will soon be adding even more DAGs that have similar logic, so this PR refactors those tasks out into reusable tasks that make it much simpler to set up.

## Testing Instructions

Test the existing use cases.

### staging_database_restore and recreate_full_staging_index

First make sure you have both of those DAGs enabled,

Run `just catalog/shell`, then the following:

```
# Trigger both at the same time, staging_db_restore first
> airflow dags trigger staging_database_restore && airflow dags trigger recreate_full_staging_index

# Locally, verify that both DAGs ran. The staging_database_restore will fail locally, but it should've passed the
# `wait_for_recreate_full_staging_index` step. The recreate_full_staging_index should have failed
# immediately.

# Trigger recreate_full_staging_index first
> airflow dags trigger recreate_full_staging_index && airflow dags trigger staging_database_restore

# Locally, verify both ran. The recreate_full_staging_index should run to completion. The staging_db_restore
# will again fail, but check the logs for the `wait_for_recreate_full_staging_index` task to verify that it
# went UP_FOR_RESCHEDULE initially and then passed.
```

### data_refresh and create_filtered_index

Enable the `audio_data_refresh` and `create_filtered_audio_index`. Repeat the same steps as above, using the airflow CLI to trigger the dags close-to simultaneously.

If the data refresh is started first, it should run successfully and the filtered index DAG should fail immediately. If the filtered index starts first, it should run successfully and the data refresh should wait until it is completed before continuing past the sensor step.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/common/sensors/external_dag_sensor.py
```diff
@@ -0,0 +1,39 @@
+from airflow.sensors.external_task import ExternalTaskSensor
+from airflow.utils.state import State
+
+from common.constants import REFRESH_POKE_INTERVAL
+from common.sensors.utils import get_most_recent_dag_run
+
+
+class ExternalDAGSensor(ExternalTaskSensor):
```
sarayourfriend, line 8: What's the reason to use a class that just extends init (and doesn't override any other behaviour) rather than a function that constructs `ExternalTaskSensor` and returns it directly, as in a factory function?

The only downside I can think of is that this creates an intermediary type with a slightly different name than the built-in Airflow one. That might be slightly more confusing to read stack traces for or keep track of `isinstance` usage (though that's probably unlikely to be used with this class, right?).

catalog/dags/common/sensors/utils.py
```diff
@@ -27,3 +31,30 @@ def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
...
+
+
+@task(retries=0)
+def prevent_concurrency_with_dag(external_dag_id: str, **context):
+    """
+    Prevent concurrency with the given external DAG, by failing
+    immediately if that DAG is running.
+    """
+
+    wait_for_dag = ExternalTaskSensor(
```
sarayourfriend, line 43: Can this be implemented using `ExternalDAGSensor` (or a factory function if you replace `ExternalDAGSensor` with one)? The only difference between the two looks like the timeout.

sarayourfriend (commented):
This LGTM, it's nice to get rid of the duplication. Just two questions to clarify the implementation.

catalog/dags/common/sensors/external_dag_sensor.py
stacimc: Just a stylistic preference, but weakly held -- and frankly, doing it this way makes it harder to organize the code for reuse because of circular imports, so I'm definitely convinced to change it. Updated, thanks!

AetherUnbound (approved):
These changes look great! Love the unification 🚀

dhruvkb (approved):
LGTM! See deduplication feels so good, thank you.

PR #3486, opened 2023-12-07
Add referrer based throttle scope

### LINKED ISSUE #3484: Add naively higher rate limit for requests with `openverse.org` referrer
## Problem

We've had to turn off rate limiting due to accidentally rate limiting users in school settings that share a single IP.

## Description

These are all from the Openverse.org website. As a temporary holdover until we can create a better solution, we'll just use a separate rate limit class when the referrer header is Openverse.org.

## Additional context

We know this is imperfect but there's no good way to solve this right now that doesn't include just leaving rate limiting off entirely for an indefinite period of time. This is a stop gap solution we sould remove once something like #3473 is implemented.

### LINKED ISSUE #3485: Thumbnail rate limit incorrectly applied to non-thumbnail routes
## Description

The thumbnail throttle classes are in the default list of throttles, which are applied to all routes with no explicit throttle classes (which is most of our routes, including search). This means that search requests incorrectly count towards thumbnail requests count, from a throttling perspective.

## Reproduction

1.  Visit https://api.openverse.engineering/v1/ and check out the response headers and see that `x-ratelimit-limit-anon_thumbnail` is incorrectly present, meaning that throttle class was applied to the request.

sarayourfriend:
## Fixes

Fixes #3484 by @sarayourfriend 
Fixes #3485 by @sarayourfriend

## Description

Adds a new set of throttle classes to match requests with the `openverse.org` referrer header. Refer to the issue for rationale behind this approach. Keep in mind it is a temporary stop gap with known issues we seek to resolve with more advanced approaches later on.

The rollout for this will need infrastructure updates to include the new rate limiting scope environment variables.

## Testing Instructions

Check over the new unit tests. I made significant modifications here to remove artificiallity from the tests. Instead, the tests run against endpoints known to have specific rate limiting configuration (whether the default or otherwise) and confirms that those rate limit scopes are used, based on the rate limiting headers).

To manually test the new throttle, run the API with `DISABLE_GLOBAL_THROTTLING=False` in your `api/.env` file (`just api/up` to get the new setting loaded). Then make a request to localhost:50280/v1/ in your browser and inspect the response. Confirm that the anon rate limit scope is applied based on the X-RateLimit headers. Now use your browser's dev tools to edit and resend the request, but set the referrer header to `openverse.org`. Inspect the response, and confirm that the `ov_referrer` scope is used instead, with separate rate limit counts. If you made a single anon request, you should see 4 remaining burst requests for `ov_referrer`. For extra confidence, run out of requests on the anon by refreshing the page 5 times (then once more to show the rate limit response) and then make the request with the referrer header. You should now have a separate limit.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/throttle.py
+++ b/api/api/utils/throttle.py
@@ -3,8 +3,6 @@ import logging
 
 from rest_framework.throttling import SimpleRateThrottle
 
-from django_redis import get_redis_connection
-
 from api.utils.oauth2_helper import get_token_info
 
 
@@ -49,7 +47,7 @@ class AbstractAnonRateThrottle(SimpleRateThrottleHeader, metaclass=abc.ABCMeta):
     logger = parent_logger.getChild("AnonRateThrottle")
 
     def get_cache_key(self, request, view):
-        logger = self.logger.getChild("get_cache_key")
+        self.logger.getChild("get_cache_key")
         # Do not apply anonymous throttle to request with valid tokens.
         if request.auth:
             token_info = get_token_info(str(request.auth))
@@ -57,16 +55,6 @@ class AbstractAnonRateThrottle(SimpleRateThrottleHeader, metaclass=abc.ABCMeta):
                 return None
 
         ident = self.get_ident(request)
-        redis = get_redis_connection("default", write=False)
-        if redis.sismember("ip-whitelist", ident):
-            logger.info(f"bypassing rate limiting for ident={ident}")
-            """
-            Exempt internal IP addresses. Exists as a legacy holdover and usages of this
-            should be replaced with the exempt API key as it is easier to manage via
-            Django admin and doesn't require leaky permissions in our production infra.
-            """
-            return None
-
         return self.cache_format % {
             "scope": self.scope,
             "ident": ident,
--- a/api/api/utils/throttle.py
+++ b/api/api/utils/throttle.py
@@ -129,9 +129,14 @@ class AbstractOAuth2IdRateThrottle(SimpleRateThrottle, metaclass=abc.ABCMeta):
     """
 
     scope: str
-    # The name of the scope. Used to retrieve the rate limit from settings.
-    applies_to_rate_limit_model: str
-    # The ``ThrottledApplication.rate_limit_model`` to which the scope applies.
+    """The name of the scope. Used to retrieve the rate limit from settings."""
+    applies_to_rate_limit_model: set[str]
+    """
+    The set of ``ThrottledApplication.rate_limit_model`` to which the scope applies.
+
+    Use a ``set`` specifically to make checks O(1). All default throttles run on
+    almost every single request and must be performant.
+    """
 
     def get_cache_key(self, request, view):
         # Find the client ID associated with the access token.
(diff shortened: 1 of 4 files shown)


api/conf/settings/rest_framework.py
```diff
@@ -13,6 +13,45 @@
...
+    "api.utils.throttle.BurstRateThrottle",
+    "api.utils.throttle.SustainedRateThrottle",
+    "api.utils.throttle.OpenverseReferrerBurstRateThrottle",
+    "api.utils.throttle.OpenverseReferrerSustainedRateThrottle",
+    "api.utils.throttle.OAuth2IdBurstRateThrottle",
+    "api.utils.throttle.OAuth2IdSustainedRateThrottle",
+    "api.utils.throttle.EnhancedOAuth2IdBurstRateThrottle",
+    "api.utils.throttle.EnhancedOAuth2IdSustainedRateThrottle",
+    "api.utils.throttle.ExemptOAuth2IdRateThrottle",
+)
```
sarayourfriend, line 53: I moved these so that they are easier to reference in the tests when we need to enable the throttling for a scoped set of tests.

api/api/utils/throttle.py
```diff
@@ -116,14 +140,14 @@ def get_cache_key(self, request, view):
...
-        if token_info.rate_limit_model != self.applies_to_rate_limit_model:
+        if token_info.rate_limit_model not in self.applies_to_rate_limit_model:
             return None
 
         return self.cache_format % {"scope": self.scope, "ident": token_info.client_id}
 
 
 class OAuth2IdThumbnailRateThrottle(AbstractOAuth2IdRateThrottle):
-    applies_to_rate_limit_model = "standard"
+    applies_to_rate_limit_model = ["standard", "enhanced"]
```
sarayourfriend, line 150: "enhanced" previously had no thumbnail rate applied to it. I don't think that's the correct behaviour, but I could be wrong. This change makes enhanced tokens have the same thumbnail rate limit as standard oauth. I think that's right because we have no separate thumbnail rate limit configured for enhanced tokens.

Exempt tokens, I assume, should be exempted from all rate limiting, so there's no change for that. That already worked.

api/api/utils/throttle.py
```diff
@@ -116,14 +140,14 @@ def get_cache_key(self, request, view):
         if not (token_info and token_info.valid):
             return None
 
-        if token_info.rate_limit_model != self.applies_to_rate_limit_model:
+        if token_info.rate_limit_model not in self.applies_to_rate_limit_model:
```
sarayourfriend, line 143: `not in` works for both a single string and for a list of strings, but I could see it being too flaky (`"standard" in "nonstandard"` is `True`). If folks prefer, I can either make all of them lists or change this to check the type of `self.applies_to_rate_limit_model` and using either `!=` or `not in`. Making all of them sets/tuples/lists would probably keep the code simpler overall.

api/api/utils/throttle.py
```diff
@@ -1,17 +1,15 @@
...
+from rest_framework.throttling import SimpleRateThrottle as BaseSimpleRateThrottle
 
 from api.utils.oauth2_helper import get_token_info
 
 
 parent_logger = logging.getLogger(__name__)
 
 
-class SimpleRateThrottleHeader(SimpleRateThrottle, metaclass=abc.ABCMeta):
+class SimpleRateThrottle(BaseSimpleRateThrottle, metaclass=abc.ABCMeta):
```
sarayourfriend, line 12: I renamed this because it no longer only handled header-related enhancements.

stacimc (approved):
Tests well for me, and the stop gap approach makes sense. No blocking feedback. The test improvements in particular are great and so easy to read, cheers!

api/api/utils/throttle.py
sarayourfriend: Done in https://github.com/WordPress/openverse/pull/3486/commits/297d74b67ada1763f4ffddd5fad8a669056432a8

api/test/unit/utils/test_throttle.py
```diff
@@ -1,32 +1,47 @@
...
+    # to emulate the application startup. Without this, MediaViewSet has cached the
+    # initial setting and won't re-retrieve it after we've called `api_settings.reload`
+    MediaViewSet.throttle_classes = api_settings.DEFAULT_THROTTLE_CLASSES
+    throttle.SimpleRateThrottle.THROTTLE_RATES = api_settings.DEFAULT_THROTTLE_RATES
 
-    yield fake_redis
-    fake_redis.client().close()
+    yield
+
+    # Set everything back as it was before and reload settings again
```
dhruvkb, line 39: This is not needed. The [docs for the `settings` fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) specify that all changes are automatically reverted.

dhruvkb (approved):
LGTM! There is one potential scope for improvement but I don't think like it should block a PR that works.

I also have mixed feelings about having to redo #3505 and #3544.

api/test/unit/utils/test_throttle.py
sarayourfriend: That is true about `settings`, but DRF's `api_settings` object will _not_ automatically reload the settings. So unless there's a way to add a fixture that wraps the `settings` fixture and runs after it has reset, where we could call `api_settings.reload()` then this is, unfortunately, still all necessary :cry:

dhruvkb: Ah, that's unfortunate. Cool then, feel free to merge it.

sarayourfriend:
> I also have mixed feelings about having to redo https://github.com/WordPress/openverse/pull/3505 and https://github.com/WordPress/openverse/pull/3544.

Well, yes, I felt the same when I saw those PRs opened after this one when this one had no reviews :wink:.

Issue #1966, opened 2023-05-02
Implementation Plan: Django admin access control and tool improvements


sarayourfriend:
## Description

Project proposal: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#volunteer-moderator-django-admin-tools-and-access-control-requirement-2

Refer to the project proposal section linked above for the requirements for this plan.

PR #3494, opened 2023-12-08
Implementation Plan: Django admin moderator access control and base improvements

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/1966 by @sarayourfriend

## Description

@krysal @stacimc I've chosen y'all to review this plan with me. I'll ping y'all privately about this, but I'd like to schedule time to do a synchronous review, if possible, to avoid the long review process getting even further prolonged by my part-time work schedule.

This discussion is following the Openverse decision-making process. Information about this process can be found [on the Openverse documentation site](https://docs.openverse.org/meta/decision_making/index.html).

## Current round

This discussion is currently in the [**Proposal approval** round](https://docs.openverse.org/meta/decision_making/process_description.html#proposal-approval).

The deadline for review of this round is **9 January 2024**.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
@@ -234,6 +234,10 @@ update the selected reports to set the `decision_id` to point to the new
 moderation decision. All actions rely on the selected set of reports and should
 not modify any other reports.
 
+Remove all logic from the report models `save` methods that bulk actions
+reports. That approach will no longer be valid and we will operate on specific
+reports individually following the `ModerationDecision`.
+
 ##### Multiple-decision prevention (soft-locking)
 
 To prevent multiple moderators unintentionally looking at the same report, we'll
@@ -445,6 +449,7 @@ For each step description, ensure the heading includes an obvious reference to t
 - Create the new `ModerationDecision` models
   - Create an `AbstractModerationDecision` for each media type
 - Add new `decision_id` column to report models
+- Remove bulk report updates from the report models' `save` methods
 
 ### 5. Media moderation view
 
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
@@ -1,4 +1,4 @@
-# {2023-12-D08} Implementation Plan: Django admin tools and access control for moderators
+# 2023-12-08 Implementation Plan: Django admin tools and access control for moderators
 
 **Author**: @sarayourfriend
 
@@ -30,8 +30,8 @@ There are two broad aspects of this implementation plan:
 
 1. Access control for moderator users (in other words, users who should only
    have access to moderation tools and nothing else)
-1. Improvements and additions to Django admin tools for the content report
-   models
+1. Improvements and additions to Django admin tools for working through content
+   reports
 
 There is also a significant bug that needs addressing: If a media object has
 multiple sensitivity reports, only one can be saved as confirmed because there
@@ -40,8 +40,8 @@ is a
 and the report model admin does not handle this. The likelihood of this should
 be all but eliminated by the process and tooling changes in this implementation
 plan (because moderation reports will be handled in a single unified view for a
-given work) but we'll solve this bug by just not creating the mature media
-object on save.
+given work). However, we'll solve this bug directly anyway by not creating the
+mature media object on save if one already exists for the work.
 
 ### Additional tools and improvements
 
@@ -61,45 +61,55 @@ relationship with the _media_ and tracks the history of moderation decisions
 made for the media, including explanatory notes from moderators. It is not
 possible to use `LogEntry`, like the project plan suggests to do, because
 `LogEntry`'s `object_id` column is not indexed, and querying it to find all the
-decisions for a given record would not scale.
+changes to all content reports for a given work would not scale.
 
 This table anticipates additional abilities for moderators in the future
 including reversing previous decisions (currently impossible, but needed if
 users and especially creators can contest decisions) or explicitly marking works
-as "not sensitive" (for example, if the work was sensitive because of erroneous
-sensitive text detection). Right now the data model and approach to sensitivity
-and deindexing don't make these possible. It is outside the scope of this
-project to add those features, but this new table, especially its `action`
+as "not sensitive" (for example, if the work was only marked sensitive because
+of erroneous sensitive text detection). Right now the data model and approach to
+sensitivity and deindexing don't make these possible. It is outside the scope of
+this project to add those features, but this new table, especially its `action`
 column, is intentionally flexible to allow for these kinds of things in the
-future.
+future. This is also why it isn't just a new `explanation` column on the content
+report models: because we want to preserve a _record_ of moderation decisions in
+chronological order, and describes the progressive changes of a work's
+sensitivity designation (or de-indexing). It also separates the moderation
+decision (and metadata) from the report, which has its own important metadata
+that should not be confused with the moderation decision metadata.
 
 It will have the following columns:
 
 - `id` - basic auto-incrementing id
+- `created_on` - the date the moderation decision was created, in other words,
+  the date the decision was made
+  - Provided by `OpenLedgerModel`
 - `moderator_id` - a foreign key relationship with moderator who made the
   decision
 - `media_identifier` - a one-to-one relationship to the media the decision is
   related to
   - Because deindexing media removes them from the API database, this needs to
-    be a non-constrained column, so not a foreign key
-- `explanation` - moderator notes explaining the decision
+    be a non-constrained column, so not a foreign key, just an indexed
+    _non-unique_ UUID column
+- `explanation` - moderator notes explaining the decision; optional
 - `action` - slug column of the action taken
   - `confirmed_sensitive` (marked a work as sensitive)
   - `deindexed` (deindexed the work from Openverse search)
-  - `rejected_report` (marked as `no_action` when the work is not already marked
-    sensitive)
-  - `duplicate` (marked a report `no_action` when the work is already marked
-    sensitive; i.e., a re-report that did not result in deindexing)
+  - `rejected_report` (marked pending report(s) `no_action` when the work is not
+    already marked sensitive)
+  - `duplicate` (marked pending report(s) `no_action` when the work is already
+    marked sensitive; i.e., a re-report that did not result in deindexing)
 
-Additionally, the report models will be expanded with an additional column,
-`decision_id` that ties the report to a specific moderation decision. A single
-moderation decision can happen as a result of multiple pending reports, so this
-facilitates the many-to-one relationship between a report and the moderation
-decision. See
-[the moderation view section below for additional information regarding this](#media-moderation-view)
+Additionally, the report models will be expanded with an additional nullable
+column, `decision_id` that ties the report to a specific moderation decision. A
+single moderation decision can happen as a result of multiple pending reports,
+so this facilitates the many-to-one relationship between a report and the
+moderation decision. See
+[the moderation view section below for additional information regarding this](#media-moderation-view).
 
 There will not be a new admin view for `ModerationDecision`. Instead, moderation
-decisions should show up on an individual media item's admin description page.
+decisions will
+[show up on an individual media item's admin description page](#media-moderation-view).
 The only part of a `ModerationDecision` moderators can manually interact with is
 the explanation on the decision. Everything else is handled by the report and
 media model admins.
@@ -133,7 +143,7 @@ works with "pending" reports, with the option to include other statuses if
 moderators choose. This creates a queue to work through progressively, from top
 to bottom. Remove any manual sorting options for now. We can add those back if
 we get feedback that moderators want them, but for now they just clutter and
-confuse the interface.
+confuse the interface, and make the foundational feature more complex.
 
 #### Media moderation view
 
@@ -150,17 +160,18 @@ Clicking on single report should show the following:
 - A chronological list of all moderation decisions for the work
 - A chronological list of all reports for the work regardless of status
 
-It is possible that to make displaying the work easier, we could iframe the
-Openverse.org view of the work into Django admin. However, that would probably
-take a lot longer to load, and I'm not sure if the website still handles being
-in an iframe very well. It would also make it harder to blur the work if it
-isn't already marked sensitive.
+The idea here is that the report view is not focused on taking an action for a
+single _report_, but rather showing all relevant information for a single work
+to make a moderation decision based on all pending reports for a work and
+historical moderation decisions for the work (if any).
 
-The idea here is that the report view is less focused on taking an action for a
-single _report_, but rather all relevant information for a single work.
+In the future, we could try to include links that open tables to show all
+moderation decisions for works by the same creator, etc, in case there is useful
+information, particularly for identifying potential bulk moderation actions.
+These kinds of additional views are out of scope for this current project,
+however.
 
-The view should have the following context-sensitive action buttons, all of
-which require a populated "explanation" free-text field:
+The expanded view should have the following context-sensitive actions:
 
 - "Mark sensitive: confirm a sensitive content report and mark the work as
   sensitive"
@@ -369,50 +380,58 @@ For each step description, ensure the heading includes an obvious reference to t
 
 [See "Media moderation view" for rationale](#media-moderation-view)
 
-- Create a new media view to replace the existing view
-  - Use the basic model admin but with only "view" settings for all model fields
-  - We'll need to expand the rendered fields to include things like the rendered
-    (interactive) thumbnail and audio player
-  - It should display (and not allow editing) the following:
-    - title
-    - description
-    - tags
-    - creator with link to creator landing page on source (if applicable)
-    - source/provider
-    - the work's thumbnail itself, blurred if the moderator has not turned off
-      blurring
-      - clicking on the thumbnail should un-blur the work
-    - if an audio work, show the basic HTML5 audio player
-    - links to the foreign landing URL and Openverse.org page of the work
-    - whether sensitive text detection matched on the work
-      - check if the work is in the filtered index, if not, then it has
-        sensitive text
+This is in two separate issues:
+
+1. Create a new media view to replace the existing view
+
+- Use the basic model admin but with only "view" settings for all model fields
+- We'll need to expand the rendered fields to include things like the rendered
+  (interactive) thumbnail and audio player
+- It should display (and not allow editing) the following:
+  - title
+  - description
+  - tags
+  - creator with link to creator landing page on source (if applicable)
+  - source/provider
+  - the work's thumbnail itself, blurred if the moderator has not turned off
+    blurring
+    - clicking on the thumbnail should un-blur the work
+  - if an audio work, show the basic HTML5 audio player
+  - links to the foreign landing URL and Openverse.org page of the work
+  - whether sensitive text detection matched on the work
+    - check if the work is in the filtered index, if not, then it has sensitive
+      text
 - Show chronolical lists of all moderation decisions and reports for the work
   - Do this using an
     [`TabularInline`](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#inlinemodeladmin-objects)
     for each
-- If the user has content moderation permissions, add the relevant actions for
-  the work based on it
-  - See the
-    [relevant overview section for described actions](#media-moderation-view)
-    and the conditions for when to show them
-  - Add an free text form field "explanation" to collect for the moderation
-    decision
-  - For any action taken, we only need to update/save a single report with the
-    new status. The `AbstractMediaReport` class already handles updating any
-    other pending reports in it's `save` method without creating
-    duplicate/conflicting mature or deleted media objects.
-  - For each of these actions, set the following new status:
-    - `confirmed_sensitive` -> `mature_filtered`
-    - `deindexed` -> `deindexed`
-    - `rejected_report` or `duplicate` -> `no_action`
 
-```{note}
-This looks like a lot of work, but most of it is handled entirely by Django admin's
-model admin and inline model admin classes, with careful configuration of their
-respective options. Most of the actual effort for this will be in the actions and
-permissions.
-```
+2. If the user viewing the page has content moderation permissions, add the
+   relevant actions for the work based on it
+
+- See the
+  [relevant overview section for described actions](#media-moderation-view) and
+  the conditions for when to show them
+- Add an free text form field "explanation" to collect for the moderation
+  decision
+- For any action taken, we only need to update/save a single report with the new
+  status. The `AbstractMediaReport` class already handles updating any other
+  pending reports in it's `save` method without creating duplicate/conflicting
+  mature or deleted media objects.
+- For each of these actions, set the following new status:
+  - `confirmed_sensitive` -> `mature_filtered`
+  - `deindexed` -> `deindexed`
+  - `rejected_report` or `duplicate` -> `no_action`
+- Prevent the race condition described in the rational section for this step by
+  setting a hidden form field with the list of pending report IDs at page load.
+  Instead of erroring when that list does not match the list of pending reports
+  at save, which would slow down the moderation workflow, just make sure that
+  only those report IDs _at load_ are the ones that are modified. To do this,
+  update the report `save` method to only bulk update reports _older_ than the
+  one being saved. Then, in the admin action, only update the oldest report of
+  the ones the moderator is known to have seen. The next time moderators come
+  past the work they'll have the most recent decision to work off of and can
+  action on the new report then.
 
 ### 6. Finalise access control
 
@@ -423,6 +442,10 @@ permissions.
   in the Terraform access module
   - Preferably, only stage this change for once we're ready to actually give
     people access and test this
+- Add the necessary permissions to the "Content Moderator" group in Django admin
+  so that content moderators can only view the new report table view, the
+  expanded individual work view with the moderation and report chronology, and
+  can issue moderation decisions on that page.
 
 ## Dependencies
 
@@ -431,7 +454,7 @@ permissions.
 <!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
 
 Small infrastructure change to add new GitHub team and allow access through
-Cloudflare Access
+Cloudflare Access.
 
 ### Other projects or work
 
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
@@ -143,7 +143,7 @@ in the columns list above.
 
 #### Table view improvements
 
-We will scrap the existing report table. Instead of showing a list of reports,
+We will scrap the existing reports view. Instead of showing a list of reports,
 we'll show a list of reported media, sorted by the total number of pending
 reports, and then by the age of the oldest pending report. This will have the
 effect of sorting re-reported works to the top of the list, under the assumption
@@ -170,8 +170,8 @@ Clicking on a single reported media object should show the following:
   - provider/source
   - links to foreign landing url and Openverse.org page for the work
 - A chronological list of all moderation decisions for the work
-- A chronological list of all reports for the work regardless of status
-  - Include at least the following columns: status, description, reason, date
+- A chronological list of all reports for the work
+  - Include at least the following columns: description, reason, date
 
 The idea here is that the "report view" is not focused on taking an action for a
 single _report_, but rather showing all relevant information for a single work
@@ -399,10 +399,10 @@ others or are non-blocking.
 3. Report table view improvements (non-blocking)
 4. `ModerationDecision` (blocking media moderation view)
 5. Media moderation view
-6. The view itself
-7. The actions
-8. The moderation soft-lock
-9. Finalise access control
+   1. The view itself
+   2. The actions
+   3. The moderation soft-lock
+6. Finalise access control
 
 ## Step details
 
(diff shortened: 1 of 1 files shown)


sarayourfriend:
This is ready for early review, but I still need to do basic proofreading of it. However, I've covered all the basic elements of the implementation, so if @krysal or @stacimc, either of y'all take a look at this before it's marked ready for review again, that's totally fine, leave any comments you like :+1:.

stacimc:
@sarayourfriend I'll plan on taking a look Monday. Scheduling a synchronous review sounds good as well 👍

krysal (commented):
I started reading, but my time today is over. I'll finish on Monday. Thanks for drafting this IP!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,468 @@
...
+It will have the following columns:
+
+- `id` - basic auto-incrementing id
+- `moderator_id` - a foreign key relationship with moderator who made the
+  decision
+- `media_identifier` - a one-to-one relationship to the media the decision is
+  related to
+  - Because deindexing media removes them from the API database, this needs to
+    be a non-constrained column, so not a foreign key
+- `explanation` - moderator notes explaining the decision
```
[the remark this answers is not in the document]
sarayourfriend: I had imagined it as required but it's a good point that it can be optional. Making it required would complicate the bluk actions stuff that the other IP (that Staci is assigned to) needs to implement.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,468 @@
+# {2023-12-D08} Implementation Plan: Django admin tools and access control for moderators
```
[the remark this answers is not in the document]
sarayourfriend: Thanks, fixed in https://github.com/WordPress/openverse/pull/3494/commits/5acaa5b00c3830845200c5033e722133b659e2e4

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: I've updated the text to say it isn't required in https://github.com/WordPress/openverse/pull/3494/commits/5acaa5b00c3830845200c5033e722133b659e2e4

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+information, particularly for identifying potential bulk moderation actions.
+These kinds of additional views are out of scope for this current project,
+however.
+
+The expanded view should have the following context-sensitive actions:
+
+- "Mark sensitive: confirm a sensitive content report and mark the work as
+  sensitive"
+  - Only available if a work is not already marked sensitive
+  - `confirmed_sensitive` action
```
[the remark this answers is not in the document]
sarayourfriend: Yep, there's nothing to prevent reports on media that's already marked sensitive. Keep in mind that reports aren't _only_ for sensitivity. The same report object handles DMCA. Additionally, something could be manually marked sensitive and get re-reported with a reason why it should be deindexed. In this case, the original moderation decision might have missed something that should have resulted in the work getting deindexed before. That's why it's "no_action" explicitly as "duplicate" if nothing happens on an already sensitive work, because there's nothing new to do.

However, now that I think about it, "duplicate" doesn't work if the report is DMCA. So maybe we shouldn't say "duplicate" but rather just "no_action" if the work is already marked sensitive? Maybe having two different "no action" statuses doesn't make sense right now and they could just both be covered under "rejected" meaning, the reports added no new information that led to a different moderation decision than whatever already stands?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+- "Deindex: entirely exclude the work from Openverse"
+  - There are no real conditions for this because deindexing can happen for
+    works that are or are not marked sensitive
+  - `deindexed` action
+- "Reject all pending reports: take no action and mark all pending reports for
+  the work as reviewed"
+  - Only available if a work is not already marked sensitive
```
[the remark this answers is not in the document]
sarayourfriend: Yep! For that reason exactly.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: After reading [this other comment of yours](https://github.com/WordPress/openverse/pull/3494#discussion_r1423181916) I realised that a moderator could make any decision based on reports regardless of the stated report reason. There isn't a hard 1-1 relationship between marking a work sensitive and the report for the work having a "sensitive" reason. It could be "other" reason, for one, or even if something is reported as DMCA, it could not be a DMCA issue but also need to have the work marked sensitive.

I think the list of actions should be this then:

- `confirm_sensitive_keep` -> set all pending reports to `sensitive_filtered`
  - Only available if the work is not already marked sensitive
- `confirm_sensitive_deindex` -> set all pending reports to `deindexed` and deindex the work
- `confirm_dmca` -> set all pending reports to `deindexed` and deindex the work
- `dismiss` -> Anything that would result in "no action" regardless of context. Reducing this to a single action simplifies things a lot here with respect to how to handle the bulk update when rejecting sensitivity reports vs dmca reports. Moderators probably just want a way to say "none of the pending reports result in a change to the work's information" and be done with it. It would be annoying to have to reject sensitivity reports and dmca reports separately on a single work to fully clear it from the queue.

Having separate actions for sensitive and DMCA deindex actions lets us mark everything with the same single "deindexed" status while also recording the _reason_ why it was deindexed, and acknowledging that the reason a user chose to report a work is not always the reason the moderator made a particular choice. It also removes the complication of the report type also potentially being "other" which I wasn't handling at all before.

We'd have to get rid of this check, though, which filters the reason of the bulk-updated reports for decisions that didn't result in deindexing: https://github.com//WordPress/openverse/blob/aa16d4f1be7607b12c428886b9890bdd947cc71c/api/api/models/media.py#L238-L239

We could actually do well to remove the bulk-updating logic from the media report `save` and handle it in either `ModerationDecision`'s `save` method or, probably more reliably, in the admin view's method to save the form. It also helps separate the logic of report creation from report actioning instead of stuffing it into a single `save` method.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+  - `rejected_report` action
+- "Mark all pending reports duplicates: take no action and mark all pending
+  reports for the work as duplicates"
+  - Only available if the work is already marked sensitive
+  - `duplicate` action
```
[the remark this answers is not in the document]
sarayourfriend: Answering this in the other related thread: https://github.com/WordPress/openverse/pull/3494#discussion_r1423179555

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+
+Each of these actions should create a single `ModerationDecision` with
+explanation, action, moderator ID, and so forth populated. The admin view should
+bulk update the reports pending at the time of the decision to set the
```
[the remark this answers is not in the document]
sarayourfriend: Answering this here as well: https://github.com/WordPress/openverse/pull/3494#discussion_r1423179555

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+to create a new "Content Moderator" user group. This user group will have only
+the following permissions:
+
+- read and update content reports
+- read related models like works, moderation decisions, and the mature and
+  deleted models for each media type
+
+Moderator users will not be allowed to delete anything and will only be allowed
+to create moderation decisions, though only indirectly through taking actions on
+reports. This includes not being able to create reports via the admin. If they
```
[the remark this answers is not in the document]
sarayourfriend: Yep, exactly. We can enable arbitrary actions later on down the line if it turns out we need them (like if somehow we're able to moderate works without needing a user to report them first).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+  decision
+- For any action taken, we only need to update/save a single report with the new
+  status. The `AbstractMediaReport` class already handles updating any other
+  pending reports in it's `save` method without creating duplicate/conflicting
+  mature or deleted media objects.
+- For each of these actions, set the following new status:
+  - `confirmed_sensitive` -> `mature_filtered`
+  - `deindexed` -> `deindexed`
+  - `rejected_report` or `duplicate` -> `no_action`
+- Prevent the race condition described in the rational section for this step by
```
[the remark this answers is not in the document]
sarayourfriend: Oh true :thinking: Should moderators obtain a "lock" on a work while they're reviewing it? We could keep a list of "in-review" works by moderator ID in Redis and exclude in-review works from the table view query and also show a message on the screen when the form loads for the work to say that someone else is currently reviewing that work?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
stacimc: +1 for the new list of actions, and separating out `confirm_sensitive_deindex` from `confim_dmca`!

I think it's still a bit unintuitive for me that taking an action updates the status of _all_ pending reports, when some of them might theoretically be unrelated/dismissable. Using the example where a record has several sensitive content reports that the moderator agrees with, and a dmca report that the moderator determines is invalid. They would take the action `confirm_sensitive_keep` and then all reports, including the invalid dmca report that wasn't really part of the decision, get set to `sensitive_filtered`. I guess that's not really an _error_, but it does feel like a slight misrepresentation of the decision.

> We could actually do well to remove the bulk-updating logic from the media report save and handle it in either ModerationDecision's save method or, probably more reliably, in the admin view's method to save the form. It also helps separate the logic of report creation from report actioning instead of stuffing it into a single save method.

Now that you say this, I think that would be excellent -- it makes much more sense to me that way.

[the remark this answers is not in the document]
stacimc: > It also removes the complication of the report type also potentially being "other" which I wasn't handling at all before.

Do we have any examples of what 'other' would be? Wondering if perhaps we need an additional `confirm_deindex` separate from `confirm_sensitive_deindex` and `confim_dmca` 🤔

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
stacimc: > We could keep a list of "in-review" works by moderator ID in Redis and exclude in-review works from the table view query and also show a message on the screen when the form loads for the work to say that someone else is currently reviewing that work?

That sounds great!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: Here are a random 10 "other" reports from production:

```
+-----+--------------------------------------+--------+------------------------------------------------------------------------+----------------+-------------------------------+
| id  | identifier                           | reason | description                                                            | status         | created_at                    |
|-----+--------------------------------------+--------+------------------------------------------------------------------------+----------------+-------------------------------|
| 71  | e48fca1b-94da-46fa-a69c-90965ad5a619 | other  | Test description is 20 characters.                                     | pending_review | 2021-12-24 01:52:01.434245+00 |
| 388 | f9384235-b72e-4f1e-9b05-e1b116262a29 | other  | Testing the functionality of this field                                | no_action      | 2022-01-23 19:33:04.130104+00 |
| 407 | d7f3ed3f-0872-4012-9b63-dea62518c1fe | other  | the site is blocked and it wont let me use it                          | pending_review | 2022-01-27 16:12:56.817752+00 |
| 408 | 8d5a18ec-71d2-4161-9de2-3b1d9978076f | other  | what hapen two this im copi this en mi clas                            | pending_review | 2022-01-27 20:34:22.130701+00 |
| 411 | 2408d871-1c5f-40fc-8b2f-3421a1d45f88 | other  | License has been altered, is no longer Public Domain.                  | pending_review | 2022-02-02 19:41:57.068421+00 |
| 412 | f6ae6dde-701d-454e-9548-03df31cf8046 | other  | aint working                                                           | pending_review | 2022-02-02 19:50:33.699505+00 |
|     |                                      |        | stoping it                                                             |                |                               |
| 442 | 40df2568-a921-4252-bae9-110cf8b4d2ac | other  | There is no Kharkov Republic, this image supports separatist ideology. | pending_review | 2022-02-24 14:20:04.713105+00 |
| 463 | 832c00d2-cadb-49a0-b026-1875b71c3a09 | other  | inapropriot content kids use this                                      | pending_review | 2022-03-28 15:48:52.421191+00 |
| 479 | a2077e8f-ac72-462a-9309-4902f826bbe1 | other  | Wrong license in Open Verse                                            | pending_review | 2022-04-08 13:17:05.290978+00 |
| 480 | b0a52325-f249-4606-ad85-56cc7b85fec7 | other  | Wrong license in Open Verse                                            | pending_review | 2022-04-08 13:17:10.682073+00 |
+-----+--------------------------------------+--------+------------------------------------------------------------------------+----------------+-------------------------------+
```

Are there reasons we would deindex something that _aren't_ related to sensitivity or DMCA (or I suppose copyright issues more broadly)?

Some of these look like feedback. We have the wrong licence for some things. That's not DMCA or sensitivity, but we also have no way to action it :open_mouth:.

Maybe combining all report reasons into a single decision view is a terrible idea, actually, and we should just have separate queues for each report reason? But then have the same set of actions for all, considering the fact that any report could lead to any type of decision, even if just because the user chose the wrong type of report. The problem with other is that we can see that sometimes people use it when they mean DMCA or sensitivity (411 and 463 respectively).

This is really tricky! I'm starting to wonder if handling reports _separately_ is actually the simplest approach to this, and we should just handle them chronologically, with previous decisions visible in the moderation view, but not having _any_ bulk handling of reports?

Maybe we can start with that, handling single reports at a time, and then if in the future moderators give us feedback that they need bulk actioning of reports for a single media item, then we can implement it with a clearer idea of the precise use case? It would simplify things for now without accidentally creating difficult-to-undo patterns?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+- The details of the reported media:
+  - title
+  - description
+  - tags
+  - the work itself
+  - whether sensitive text detection thought it had sensitive text
+  - provider/source
+  - links to foreign landing url and Openverse.org page for the work
+- A chronological list of all moderation decisions for the work
+- A chronological list of all reports for the work regardless of status
```
stacimc, line 161: It would be nice to make sure the reports are listed with their status visible, though, and sorted by date so the pending ones are at the top and easily distinguished from reports that have already been acted upon.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
stacimc: I think single reports at a time would be fine for the number of reports we have now, and could be expanded upon in the future 👍 But the moderation view as described, where you can still see the context of _all_ the reports for a single image regardless of their report reason, still seems valuable to me even if we're only updating one report at a time. I think you'd still want the full context of reports and historic moderation decisions.

Ideally it would be nice if from the moderation view, you could select one **or more** pending reports and then take an action on just the selected reports. Is that doable? I guess you'd want to only allow taking action on pending reports, which adds a complication 🤔 We could have `pending_reports` be an actionable tabular inline list, and then separate (non-actionable) lists of previous moderation decisions and non-pending reports?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Yes, absolutely. I'll make the list of columns to display explicit.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: That's true. So we'll show a list, have a checkbox next to each, and then the actions only apply to the selected ones? If there's only a single report, then it'll be automatically selected?

The separated lists sounds good too.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+in the columns list above.
+```
+
+#### Table view improvements
+
+We will scrap the existing report table. Instead of showing a list of reports,
+we'll show a list of reported media, sorted by the total number of pending
+reports, and then by the age of the oldest pending report. This will have the
+effect of sorting re-reported works to the top of the list, under the assumption
+that a work with multiple pending reports is probably higher priority, because a
```
krysal, line 140: Great idea!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,491 @@
...
+
+#### Media moderation view
+
+Clicking on single report should show the following:
+
+- The details of the reported media:
+  - title
+  - description
+  - tags
+  - the work itself
```
krysal, line 156: Is this the direct URL?

It may also be a given, but it would be good to include in this list the `description` sent with the report (different from the media's description), if any.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
krysal: Interesting conversation. The same doubts were coming to me as I read 😄  I think separate views for each report reason is a good idea, that way, the bulk actions should apply only to reports with the same reason because I _assume_ the people reviewing and making decisions on sensitive work would not be the same as the people confirming a DCMA infraction? It's an option, but I'm okay with starting with single reports moderation too.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: It would be either the image or audio available to view/play.

I'll make explicit that the report description should be present as well :+1:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: Yeah, we'll start there :+1: The "other" reason really throws this whole thing for a loop, and the fact that a user could choose the wrong reason as well (which appears to happen for "other" especially, in fact). I'll re-write this to simplify it along the lines of this discussion and we can deal with complexities once we have people actually using the tools who can give feedback about what needs to change :slightly_smiling_face:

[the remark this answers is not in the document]
sarayourfriend: I've updated the plan to require the moderator to select a set of reports to action, and updated the list of actions to be more flexible. Nothing will operate on unselected reports and moderators can mark any type of report as having led to a sensitivity decision or copyright decision (those are the only two we have for now).

sarayourfriend:
@krysal @stacimc this is ready for review again. I set the deadline to Wednesday to give me time on my Thursday and Friday before I go on Holiday to review any additional comments and (hopefully) create the issues.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+  - Include at least the following columns: status, description, reason, date
+
+The idea here is that the "report view" is not focused on taking an action for a
+single _report_, but rather showing all relevant information for a single work
+to make a moderation decision on the work, based on all pending and historical
+reports.
+
+We could take two approaches for this view: either create a brand new view or
+build this into the media view itself. My preference and recommendation is the
+latter, because while it creates complexity in the media admin view, that view
```
krysal, line 183: Would this be possible with the Django's Access Control features? Ideally moderators shouldn't be able to edit media works, only reports (and indirectly, moderation decisions) as you say further below.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+confuse the interface, and make the foundational feature more complex.
+
+#### Media moderation view
+
+##### Information presentation
+
+Clicking on a single reported media object should show the following:
+
+- The details of the reported media:
+  - title
```
krysal, line 165:
```suggestion
##### Information presentation

Clicking on a single reported media object should show the following:

- Description provided by the reporting user
- The details of the reported media:
  - title
```

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+
+[See "Media moderation view" for rationale](#media-moderation-view)
+
+This is in three separate issues:
+
+1. Create a new media view to replace the existing view
+
+- Use the basic model admin but with only "view" settings for all model fields
+- We'll need to expand the rendered fields to include things like the rendered
+  (interactive) thumbnail and audio player
```
krysal, line 464: "interactive" thumbnail? Does that mean that thumbnail can be blurred as well?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+- Add new `decision_id` column to report models
+- Remove bulk report updates from the report models' `save` methods
+
+### 5. Media moderation view
+
+[See "Media moderation view" for rationale](#media-moderation-view)
+
+This is in three separate issues:
+
+1. Create a new media view to replace the existing view
```
krysal, line 460: I think it still valuable to keep the current media detail view. In case someone needs to check something from a specific work having the uuid, it could be easier from the admin UI than accessing the production DB.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+
+Each of these actions should create a single `ModerationDecision` with
+explanation, action, moderator ID, and so forth populated. The admin view should
+update the selected reports to set the `decision_id` to point to the new
+moderation decision. All actions rely on the selected set of reports and should
+not modify any other reports.
+
+Remove all logic from the report models `save` methods that bulk actions
+reports. That approach will no longer be valid and we will operate on specific
+reports individually following the `ModerationDecision`.
```
krysal, line 239: Removing that logic from the `AbstractMediaReport` model, would that also imply removing its `status` column? It won't make sense with the current option values since the model will represent exclusively a user input, not the decision, and from what I understand if that a report will now have only two possible states: `pending` or `no_pending`.

To determine if a model if pending from decision can be a simply check of if the new `decision_id` column is null or not. Am I right this is what is expected here?

krysal (commented):
The simplified proposal makes perfectly sense, and it is quite complete with the security details from CloudWatch and Terraform. Thank you!

From the comments I left, my main question is related to how it would change the report models, specifically on the status.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+- `created_on` - the date the moderation decision was created, in other words,
+  the date the decision was made
+  - Provided by `OpenLedgerModel`
+- `moderator_id` - a foreign key relationship with moderator who made the
+  decision
+- `media_identifier` - a one-to-one relationship to the media the decision is
+  related to
+  - Because deindexing media removes them from the API database, this needs to
+    be a non-constrained column, so not a foreign key, just an indexed
+    _non-unique_ UUID column
```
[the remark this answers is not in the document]
stacimc: We _do_ actually [delete the media](https://github.com/WordPress/openverse/blob/main/api/api/models/media.py#L329) from the database, in addition to removing it from Elasticsearch. They are not deleted in the _catalog_ database, but we have logic in the ingestion server that prevents them from being added back to the API db during a data refresh as well.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
stacimc: What does `description provided by the reporting user` mean? The description field from the Report object? That would presumably already be included per-report in the chronological list of reports below, right? Since there may be multiple reports.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+  - Only available if a work is not already marked sensitive
+  - `confirmed_sensitive` action
+- "Deindex sensitive: entirely exclude the work from Openverse due to
+  sensitivity based on selected reports"
+  - There are no real conditions for this because deindexing can happen for
+    works that are or are not marked sensitive
+  - `deindexed_sensitive` action
+- "Deindex copyright: entirely exclude the work from Openverse due to copyright
+  based on selected reports"
+  - `deindexed_copyright`
```
stacimc, line 224: Should there be guidance for what to do if a record ought to be indexed for _both_ copyright and sensitivity concerns?

I think there's value in the separation especially if a work is deindexed without an official report (for example if it was the result of a bulk action), because the reason is preserved. And I suppose that if there are reports of multiple types that lead to an indexing decision, they're all linked to the ModerationDecision, so that context is not missed.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: It deletes the media as well as removing the ES result: https://github.com/WordPress/openverse/blob/aa16d4f1be7607b12c428886b9890bdd947cc71c/api/api/models/media.py#L329

It's a hard delete from the API database and future reindex jobs exclude them altogether based on the deleted media table. There is also the issue of the recreate causing foreign key issues, but we solve that with Django level contraints elsewhere (`db_constraint=False` [on the report models, for example](https://github.com/WordPress/openverse/blob/eb0e906e7300e32fae945eb9222ed47a50fb72a2/api/api/models/image.py#L111-L119)).

So it's both :slightly_smiling_face: 

Also occurring to me that we might want to make this a many-to-many for the bulk moderation actions in the future :thinking:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Yes, definitely, the plan is to use Django's access control for this.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: The report description would be in the report table.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Yes. That's covered in this implementation plan (see the moderator preferences section, for example). The project plan calls for this.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
```diff
@@ -0,0 +1,587 @@
...
+
+- See the
+  [relevant overview section for described actions](#media-moderation-view) and
+  the conditions for when to show them
+- Add an free text form field "explanation" to collect for the moderation
+  decision
+- For any action taken, we only need to update/save a single report with the new
+  status. The `AbstractMediaReport` class already handles updating any other
+  pending reports in it's `save` method without creating duplicate/conflicting
+  mature or deleted media objects.
```
stacimc, line 495: This sentence should be deleted, right? Now that we are removing that logic from the AbstractMediaReport, and instead updating each selected report individually.

stacimc (commented):
The updates look fantastic :) Really pleased with the flexibility of this approach!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: This view would retain all the functionality of the existing view. I can reword this to say extend, rather than replace. The view would still be registered as the model admin for the media items and would have all the same features, just with the additional moderation features added at the bottom of the current view and with the media data shown in an uneditable state for moderators.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: That's true, we could get rid of the status column altogether and treat `decision_id`'s presence to indicate the reviewed state. That would simplify the data model a bit and even simplify https://github.com/WordPress/openverse/issues/1965 because we wouldn't have to migrate those fields anymore.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: Oops, sorry, Staci's comment hadn't loaded for me yet when I wrote mine! Staci's got it :+1:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Oops, yes, definitely.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Hmm, that's interesting. I really don't know, other than either adding a list field that includes the reasons (so like, the action would be `deindexed` and then the moderator would select the reasons as both sensitivity and copyright)? Or maybe a `deindexed_sensitive_and_copyright`? To me, I feel like copyright nullifies the sensitivity question.

> And I suppose that if there are reports of multiple types that lead to an indexing decision, they're all linked to the ModerationDecision, so that context is not missed.

By this, do you mean to say that you feel the context is already sufficiently covered?

This case, where we need to know that it was both for sensitivity and copyright, would only be an issue if we're overturning deindex decisions based on appeals at some future date, right? Those appeals would be manually reviewed, and the moderator would presumably look at the moderation decision history then. But for bulk actions or if moderators are working quickly and not necessarily adding explanations, it could cause an issue.

What do you think? Is it worth increasing the complexity to capture the reason as a separate thing?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
krysal: Ok, it seems we understand kind of the same here! A deletion registry is still _kept in the API DB_, not in the `<media>` table _but_ in the `api_<media>deleted` or `nsfw_reports*` (oh lord, forgive the names!) tables, depending on the action selected. Otherwise, it wouldn't be possible to determine which works have been marked as sensitive or deleted when reingesting.

My fear was that saying that it's a "_hard delete from the API database_" is not totally correct and might be missing the context of the data refresh process, but if we all understand the same then it's fine.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
krysal: Oh right! I was reading it in terms of the old media report detail view. It makes perfect sense then 👍 Forget this.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
krysal: Totally! If reports don't have another state, I can't see a more straightforward approach. I think we should go with it.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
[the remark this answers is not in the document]
sarayourfriend: Yup, you've got it. The "hard delete" is from the perspective of just the media object, which is what this identifier is for. I left out further details for no real reason other than they felt unrelated, but I can see the potential confusion if someone doesn't have all the domain context... but then I don't know when to stop writing :joy:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
sarayourfriend: Sounds good. The question is then how we back-fill Moderation Decision objects. We can backfill them, for each one:

- Create a moderation decision with moderation_id set to the `opener` django admin user
- Set the appropriate action on the decision based on the status of the report object
- Add an explanation on the decision indicating it is a backfill record

There are currently a small number of these, so it would be a fast backfill to complete in a simple management command.

stacimc: +1 to all this from me as well :) Especially since the backfill should be simple

documentation/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.md
stacimc: I meant that I came to the same conclusion as you, I think -- the copyright nullifies the sensitivity question. So if a record has copyright and sensitivity reports, I think it would make sense for a moderator to use the `deindexed_copyright` action as that takes "priority", but link that ModerationDecision to _all_ the reports, including the sensitivity ones.

Then, when you're looking at the ModerationDecision in the future, the sensitivity reports will still be linked to it, so you'll still be able to see the full context.

Basically I think I talked myself into the same thing -- I think the context is covered sufficiently, and it's not worth adding more complexity 😄 Just reasoning through the edge cases!

stacimc (approved):
I did not see this formally went into the Decision round -- officially LGTM!

sarayourfriend:
@stacimc @krysal I've updated the implementation plan with the new backfill for moderation decision and dropping report status. The plan is in the ["proposal approval" stage](https://docs.openverse.org/meta/decision_making/process_description.html#proposal-approval), so please re-review and share any outstanding blockers that prevent the plan from moving forward, or confirm approval so we can move forward and I can create issues for the plan.

krysal (approved):
Looks perfect!

sarayourfriend:
I've created the issues. Milestone is here: https://github.com/WordPress/openverse/milestone/22

PR #3515, opened 2023-12-12
Fix incorrect default production data refresh limit of 0

stacimc:
## Description

#3232
The ingestion server allows setting a `DATA_REFRESH_LIMIT` used to cap the number of records processed in a data refresh. By default on production we don't have a limit. Historically we have represented this by setting the limit to `0`. The copy data step has a simple[ if conditional ](https://github.com/WordPress/openverse/blob/c6ce2c6d819c0c316f64bdd616cf1378976fa766/ingestion_server/ingestion_server/queries.py#L184)before applying limits which meant that 0 was correctly considered "no limit".

When #3232 added logic to the reindexing step to also apply the record limit here, it did not have a similar check. The result is that in production, the default of 0 is applied... and 0 records are reindexed, effectively totally breaking the data refresh.

This PR:
* Adds a check to the index worker scheduler to only apply the record limit if in fact one has been set
* Changes the default limit in production from 0 to None, as this is more intuitive.

## Testing Instructions

Locally, explicitly set `DATA_REFRESH_LIMIT` to None in your `ingestion_server/.env` to imitate production. Then run a data refresh and ensure that records make it all the way to the new es index.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/distributed_reindex_scheduler.py
+++ b/ingestion_server/ingestion_server/distributed_reindex_scheduler.py
@@ -42,9 +42,11 @@ def _assign_work(db_conn, workers, model_name, table_name, target_index):
         cur.execute(est_records_query)
         estimated_records = cur.fetchone()[0]
 
-    records_per_worker = math.floor(
-        min(estimated_records, get_record_limit()) / len(workers)
-    )
+    # If a record_limit has been set, cap the number of records to be indexed.
+    if record_limit := get_record_limit():
+        estimated_records = min(estimated_records, record_limit)
+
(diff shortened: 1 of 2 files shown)


sarayourfriend (approved):
LGTM!

AetherUnbound (approved):
LGTM, I looked though this logic and thought we could get tripped up here but assumed that we'd be okay since the value would end up being `None` - alas!

PR #3541, opened 2023-12-17
Handle the possibility of an ES hit not being present in the database

dhruvkb:
## Description

This PR makes the `get_db_results` function in `MediaViewSet` more robust by handling the case where an ES hit is not present in the database and in those situations, it just returns the `Hit` instead of the DB entry.

> [!NOTE]
> This situation is unlikely to occur as the ingestion server pushes the same data into the downstream API database as it indexes in Elasticsearch.

This change preserves the contract for the function to return as many entries as were passed to it, but in the case of inconsistencies, more fields will be `None` in the response as they are not present on the `Hit` during serialization. I don't know if that's an improvement over silently returning fewer results.

## Testing Instructions

The included CI check verifies this behaviour. Additionally,

- Search for some term.
- Delete an item visible in the results from the DB.
- Search for the same term.
- You should still see the item, but with more fields set to `None` this time.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -101,7 +101,7 @@ class MediaViewSet(AsyncViewSetMixin, AsyncAPIView, ReadOnlyModelViewSet):
         req_serializer.is_valid(raise_exception=True)
         return req_serializer
 
-    def get_db_results(self, results):
+    def get_db_results(self, hits):
         """
         Map ES hits to ORM model instances.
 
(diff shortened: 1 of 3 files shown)


sarayourfriend:
This change would force the serializer to continue to need to handle a hit and a Django DB result, right? Because now, if the db result doesn't exist, then we are serializing the hit?

I think it's better to just leave these results out entirely. Like you said, it shouldn't happen, it's an obscure edge case (I can't even think when it _could_ happen, like what would cause this other than a very serious issue), that makes the serializer code more complex.

Do you think this is worth it? Is the side effect of returning fewer hits that the result count for the page will be incorrect? That seems like a minor issue that isn't worth trying to handle at the expense of keeping a consistent return type from `get_db_results`. Now the name is also incorrect. It doesn't get db results. It _tries_ to get DB results but then silently mixes in the hits if it couldn't get them. Unless there's a really bad side effect to not finding the DB result for a hit and leaving out that result from the search response (keeping in mind that it should apparently never happen), I'm worried this complicates things for no clear benefit.

dhruvkb:
> Do you think this is worth it?

No, it was meant to be a draft PR where I wanted to experiment with this bit of the code. You're right that it complicates a bunch of things without any meaningful benefit. Sorry for wasting your time, I did not intend for it to be reviewed, at least not in this state. I'll just close it because the tests for the change are broken too.

sarayourfriend:
No worries at all, Dhruv.

PR #3552, opened 2023-12-19
Return status 404 instead of 500 when media not found

### LINKED ISSUE #3520: Related endpoint on a missing/non-existent image raises a 500
## Description

When the `/related/` endpoint is requested for a record with a valid identifier which does not exist in the database, the request returns a 500 instead of a 404. We will want to change this logic to return a 404 instead.

## Reproduction

1. `j api/init`
2. Visit http://localhost:50270/v1/images/ and find a valid image identifier
3. Change one character in the identifier and ensure the new path 404s (example I used: http://localhost:50270/v1/images/94104cd4-9c38-4d6e-a27e-1a4a02e8c253/)
4. Add `/related/` to the end and verify it now 500s instead (example: http://localhost:50270/v1/images/94104cd4-9c38-4d6e-a27e-1a4a02e8c253/related/)

dhruvkb:
## Fixes

Fixes #3520 by @AetherUnbound 

## Description

This PR changes the 500 exception raised by the `/related` endpoint to 404 instead because that's more appropriate for when the identifier does not point to a media item.

## Testing Instructions

Visit the non-existent related endpoint and observe the response code. See the original issue for examples.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -3,7 +3,7 @@ from typing import Union
 
 from rest_framework import status
 from rest_framework.decorators import action
-from rest_framework.exceptions import APIException
+from rest_framework.exceptions import APIException, NotFound
 from rest_framework.response import Response
 from rest_framework.viewsets import ReadOnlyModelViewSet
 
@@ -272,7 +272,7 @@ class MediaViewSet(AsyncViewSetMixin, AsyncAPIView, ReadOnlyModelViewSet):
             raise APIException(getattr(e, "message", str(e)))
         # If there are no hits in the search controller
         except IndexError:
-            raise APIException("Could not find items.", 404)
+            raise NotFound
 
         serializer_context = self.get_serializer_context()
 
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -27,6 +27,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
 
     # Search the default index for the item itself as it might be sensitive.
     item_search = Search(index=index)
+    # This will raise ``IndexError`` if no hits are found. This error is caught
+    # in the viewset handler function.
     item_hit = item_search.query(Term(identifier=uuid)).execute().hits[0]
 
     # Match related using title.


api/api/views/media_views.py
```diff
@@ -272,7 +272,7 @@ def related(self, request, identifier=None, *_, **__):
             raise APIException(getattr(e, "message", str(e)))
         # If there are no hits in the search controller
         except IndexError:
-            raise APIException("Could not find items.", 404)
```
AetherUnbound, line 275: If this was supposed to be raising a 404, why wasn't it working before?

AetherUnbound (approved):
Tested locally, looks good! Thanks for the quick fix!

api/api/views/media_views.py
sarayourfriend: Because the second parameter isn't the status code, it's the _error_ code: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L99-L114. Typically this is something like a slug that identifies the error type within the status code: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L183-L192. This lets you have multiple distinct error types within a single semantic HTTP status code that are identifiable on the client without needing to parse the potentially arbitrary error message string.

api/api/views/media_views.py
```diff
@@ -272,7 +272,7 @@ def related(self, request, identifier=None, *_, **__):
...
+            raise NotFound
```
sarayourfriend, line 275: TIL raise accepts a class object rather than an instance for lazy evaluation of the parameterless constructor :open_mouth: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

> It must be either a subclass or an instance of [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException). If it is a class, the exception instance will be obtained when needed by instantiating the class with no arguments.

sarayourfriend (approved):
LGTM!

api/api/views/media_views.py
AetherUnbound: Woah, TIL too!

PR #3560, opened 2023-12-20
Use api_client, move fixtures around to share basic defaults

sarayourfriend:
Some potential changes for Dhruv's PR.

### DIFF
--- a/api/test/integration/test_auth.py
+++ b/api/test/integration/test_auth.py
@@ -3,7 +3,6 @@ import uuid
 from test.constants import API_URL
 
 from django.urls import reverse
-from django.utils.http import urlencode
 
 import pytest
 from oauth2_provider.models import AccessToken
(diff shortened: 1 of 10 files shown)


api/test/integration/test_auth.py
```diff
@@ -54,20 +53,19 @@ def test_auth_tokens_registration(client):
...
+        "client_id": api_client_id,
+        "client_secret": api_client_secret,
+        "grant_type": "client_credentials",
+    }
+
+    res = api_client.post(
         "/v1/auth_tokens/token/",
         data,
-        "application/x-www-form-urlencoded",
+        "multipart",
```
dhruvkb, line 68: What is the reason for this change?

dhruvkb (approved):
All changes LGTM and while they may not fix the async errors, they're an improvements to the test suite overall 👍!

api/test/integration/test_auth.py
sarayourfriend: It's part of DRF's `api_client`'s testing request renderer formats. It takes a slug corresponding to the renderer rather than the content type string: https://www.django-rest-framework.org/api-guide/testing/#setting-the-available-formats

Here's an SO answer explaining it as well: https://stackoverflow.com/questions/68528408/difference-between-content-type-application-json-and-format-json-in-drf-apiclien

DRF's testing APIClient actually gave a pretty easy to understand error message about this, I was not aware of this until fixing that error when I changed this test to api_client.

PR #3567, opened 2023-12-21
Add initial vale configuration and apply editorial changes to documentation

### LINKED ISSUE #3566: Integrate Vale for linting docs according to editorial style guide
## Description

We can integrate Vale (https://github.com/errata-ai/vale) into the pre-commit hooks to ensure that our documentation has a consistent style guide that can be enforced. Vale has a significant number of stars on GitHub, seems actively maintained and [supports pre-commit](https://github.com/errata-ai/vale/blob/v2/.pre-commit-hooks.yaml).

This is related to WordPress/openverse#368.

sarayourfriend:
## Fixes

Fixes #3566 by @dhruvkb 

## Description

I've picked this work up from @ngken0995 so that Ken can focus on other PRs.

When we originally started looking into Vale, we were misinformed about how it worked. We thought that you had to commit the styles files to the repository. However, that's not the intended way of using Vale. Instead, the Vale documentation instructs you to git-ignore all "synced" configuration files, and only commit custom configuration files. This allows us to avoid introducing big long lists of offensive terms into the monorepo. Those files would still appear if you ran `vale sync` from within the new `.vale` directory. But that is _not_ necessary to do, not even when developing the Vale configuration, because everything can run from within the built docker image. Running Vale using Docker also allows us to avoid contributors needing to install Vale locally, and avoids potential conflicts in Vale versions by pinning the Vale version we use in the base docker image's tag.

We will not use a published image in any circumstance because: (a) the vale docker image is small and after the first time it is downloaded it is reused every time, unlike needing to redownload the entire published image each time the configuration changes, and (b) it avoids a whole host of complexity in our CI and local linting when changes to the Vale configuration conflict with the previous Vale configuration. For example, if a new configuration conflicts with the currently published configuration, the Vale pre-commit step would fail when running the published image! This was too hard to resolve, and would require looking at the git history compared to `HEAD` in `pre-commit`, or juggling new conditionals in our already complex ci_cd.yml workflow. Using a local image every time avoids all of this complexity without any downsides. In fact, it is positive because the base vale image is always reused instead of needing to redownload it any time the Vale configuration changes (excluding updates to the version of the base image).

Run vale using `just .vale/run`. This also builds the docker image. Refer to `.vale/README.md` and other new in-code documentation for rationale behind various configuration choices.

The pre-commit hook also excludes project proposals and changelogs. Changelogs are automatically generated, and as such, I don't think they should be linted for editorial content. For example, if a PR title doesn't pass Vale, I don't think it's accurate to change the title in the changelog. That's my personal judgement, but if other folks want to run Vale there, I would also think that's fine and wouldn't push against it.

The project proposals, on the other hand, are complex. They raise a lot of errors in Vale and are the bulk of our documentation, line-by-line. They are also not "living" documents, that we make changes to over time. Once they're in, they're more or less meant to stay as-is, with changes to the project reflected in the project threads (that's my understanding). Going back and making retroactive editorial changes to those files doesn't feel right to me. On the other hand, if folks _do_ want to do that, then I also wouldn't push back here, I'd just ask that it gets done in a separate PR.

This configuration is just a starting point, and we can incorporate additional rules in the future, if we like, even from other style guides. Red Hat's guide in particular has a lot of compelling options, including scentence complexity checks and a much better passive voice check than the other ones I've seen.

## Testing Instructions

Run `just .vale/run`. Make changes that will cause issues with Vale and confirm they raise errors as you expect. Confirm no unnecessary errors are present.

Read over the Vale configuration and through the Vale documentation to confirm that it all makes sense and is appropriate for our project.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -60,6 +60,7 @@ lint:
   - .pre-commit-config.yaml
   - tsconfig.base.json
   - .codespell
+  - .vale
 mgmt:
   - .github/**
   - .devcontainer/**
@@ -68,3 +69,5 @@ mgmt:
 ci_cd:
   - .github/actions/**
   - .github/workflows/ci_cd.yml
+vale:
+  - .vale
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -69,5 +69,3 @@ mgmt:
 ci_cd:
   - .github/actions/**
   - .github/workflows/ci_cd.yml
-vale:
-  - .vale
--- a/catalog/dags/providers/provider_api_scripts/inaturalist.py
+++ b/catalog/dags/providers/provider_api_scripts/inaturalist.py
@@ -7,7 +7,7 @@ Notes:      The iNaturalist API is not intended for data scraping.
             https://api.inaturalist.org/v1/docs/
             But there is a full dump intended for sharing on S3.
             https://github.com/inaturalist/inaturalist-open-data/tree/documentation/Metadata
-            Because these are very large normalized tables, as opposed to more document
+            Because these are exceptionally large normalized tables, as opposed to more document
             oriented API responses, we found that bringing the data into postgres first
             was the most effective approach. More detail in slack here:
             https://wordpress.slack.com/archives/C02012JB00N/p1653145643080479?thread_ts=1653082292.714469&cid=C02012JB00N
--- a/justfile
+++ b/justfile
@@ -89,7 +89,7 @@ precommit:
 
 # Run pre-commit to lint and reformat files
 lint hook="" *files="": precommit
-    python3 pre-commit.pyz run {{ hook }} {{ if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
+    python3 pre-commit.pyz run {{ hook }} {{ if files =~ "all-files" { "" } else if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
 
 ########
 # Init #
--- a/justfile
+++ b/justfile
@@ -23,6 +23,7 @@ DC_USER := env_var_or_default("DC_USER", "opener")
     cd automations/python && just
     cd automations/js && just
     cd documentation && just
+    cd .vale && just
     printf "\nTo run a nested recipe, add the folder path before it, like \`just frontend/install\`.\n"
 
 ###########
--- a/justfile
+++ b/justfile
@@ -90,7 +90,7 @@ precommit:
 
 # Run pre-commit to lint and reformat files
 lint hook="" *files="": precommit
-    python3 pre-commit.pyz run {{ hook }} {{ if files =~ "all-files" { "" } else if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
+    python3 pre-commit.pyz run {{ hook }} {{ if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
 
 ########
 # Init #
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -25,6 +25,7 @@ tsconfig.base.json  @WordPress/openverse-frontend
 
 # Specific assignments for the 'openverse-maintainers' group
 .codespell/                 @WordPress/openverse-maintainers
+.vale/                 @WordPress/openverse-maintainers
 .devcontainer/              @WordPress/openverse-maintainers
 .github/                    @WordPress/openverse-maintainers
 automations/                @WordPress/openverse-maintainers
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -219,7 +219,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: ${{ matrix.context }}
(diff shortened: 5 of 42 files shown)


sarayourfriend:
An alternative to using a published image is to run the vale linting step using `just .vale/ci` that builds the image and then runs it with the configuration we need. That would make it easier to test the CI configuration locally too, without needing to duplicate the CLI args between pre-commit and the justfile.

.github/filters.yml
```diff
@@ -68,3 +69,5 @@ mgmt:
 ci_cd:
   - .github/actions/**
   - .github/workflows/ci_cd.yml
+vale:
+  - .vale
```
krysal, line 73: Does it need its own section? Why not just keep it under "lint"?

justfile
```diff
@@ -89,7 +90,7 @@ precommit:
 
 # Run pre-commit to lint and reformat files
 lint hook="" *files="": precommit
-    python3 pre-commit.pyz run {{ hook }} {{ if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
+    python3 pre-commit.pyz run {{ hook }} {{ if files =~ "all-files" { "" } else if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
```
krysal, line 93: Can this be broken down at least into two lines to make it more readable?
```suggestion
    python3 pre-commit.pyz run {{ hook }} \
    {{ if files =~ "all-files" { "" } else if files == "" { "--all-files" } else { "--files" } }}  {{ files }}
```

krysal (commented):
This looks pretty cool! I haven't read everything, so for now, I left a few questions.

.github/filters.yml
sarayourfriend: Totally, I'll move it to `lint` :+1:

justfile
sarayourfriend: I can remove this change entirely now.

.github/filters.yml
sarayourfriend: Ah, actually, I remember now why I had this separated in addition to under `lint`: the `get-changes` CI step uses this filters.yml to determine the groups of changes, and we need to detect changes in `.vale` to build and publish the vale image when the vale configuration changes. This is what makes the changes in `set_matrix_images.py` actually cause the docker image to get built (and published), because that script checks for `vale` in the set of changes:

https://github.com//WordPress/openverse/blob/b6ccc486ebd561f9777ee5128948e6b2e262c77b/automations/python/workflows/set_matrix_images.py#L61

sarayourfriend: I've removed this altogether in favour of a much simpler local-docker image approach. I updated the PR description with details and rationale.

stacimc (approved):
~Very~ Exceptionally cool :) Documentation improvements all look good to me as well; this is a welcome change!

I agree with you re: project proposals and especially the changelogs.

One thing I noticed when testing is that Vale appears to ignore anything in code blocks using the backticks. In our documentation some of these are actual code blocks but others are notes/warnings that probably ought to be checked. For example if I add this in a doc file, it won't raise any warnings or errors:

> \`\`\`{warning}
This note is very likely to to cause Vale errors.
\`\`\`

Not a blocker but perhaps something to investigate in a follow up? I looked briefly and was unable to determine if this is a bug or intentional behavior by Vale.

krysal (approved):
Nice addition! The last thing I think is adding the `.vale/` folder to the CODEOWNERS. 
Thank you for this comprehensive setup.

catalog/dags/providers/provider_api_scripts/inaturalist.py
```diff
@@ -7,7 +7,7 @@
             https://api.inaturalist.org/v1/docs/
             But there is a full dump intended for sharing on S3.
             https://github.com/inaturalist/inaturalist-open-data/tree/documentation/Metadata
-            Because these are very large normalized tables, as opposed to more document
+            Because these are exceptionally large normalized tables, as opposed to more document
```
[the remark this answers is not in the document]
sarayourfriend: Yes, that is precisely why. Sorry, I should have left a note regarding this particular change.

sarayourfriend:
Okay, I've updated the PR with the following:

1. Check Markdown files outside of `documentation`. As part of this I cleaned up the glob configuration in `.vale/justfile` to make it slightly easier to read and maintain (though not by much, due to limitations in the glob syntax and just/scripting while avoiding too much complexity)
2. Check MDX files as Markdown
3. Check admonitions
4. Fix a lot of false positives by updating `Openverse.TermCasing` to ignore path-like phrases

sarayourfriend:
@obulat @dhruvkb I think this PR needs a review from one of y'all to be able to get merged? I can't understand why else the existing 2 approvals wouldn't be sufficient.

dhruvkb (approved):
I think that's because "frontend" owns at least one of the modified files and Staci isn't a member of that group.

PR #3590, opened 2023-12-26
Additional service checks for ingestion server health endpoint

### LINKED ISSUE #2019: Add additional checks to ingestion server healthcheck endpoint
## Problem

The ingestion server healthcheck endpoint returns 200 without actually doing any checks for things like database or Elasticsearch connectivity. The endpoint could return 200 even if the ingestion server lost database or ES connectivity, both of which are necessary for the critical paths of the application.

## Description

Similar to https://github.com/WordPress/openverse-api/issues/14 and https://github.com/WordPress/openverse-api/issues/1079 for the API, we should add some new checks to the healthcheck endpoint. At minimum, we should confirm that the databases and Elasticsearch are all accessible.

## Additional context

Optionally, we may want to add a check for the indexer workers and for running task healthchecks, but that is probably more involved (requires a bit more thought to how best to do it and what conditions to look for) so we can defer that. The most high value checks are the connectivity ones above.

AetherUnbound:
## Fixes

Fixes #2019 by @sarayourfriend

## Description

This PR adds checks for Elasticsearch and both databases to the ingestion server health check endpoint behind the `check_deps` query parameter. This also refactors the database connection into a `db_helpers` module similar to the `es_helpers` one that already exists. This helped simplify some of the code and centralize _where_ the database connection logic was being defined.

The failed response has a `status` field with a list of possible failures due to the number of upstream services to be checked. If there's a better way to structure that, let me know!

## Testing Instructions

1. `just i` to start the stack
1. Run `curl -L http://localhost:50281/` and verify the output is `{"status": "200 OK"}`
1. Run `curl -L http://localhost:50281/?check_deps=true` and verify the output is `{"status": "200 OK"}`
1. Run `docker stop openverse-db-1 openverse-es-1 openverse-upstream_db-1` to simulate the resources becoming unavailable
1. Run `curl -L http://localhost:50281/` and verify the output is `{"status": "200 OK"}` (unaffected)
1. Run `curl -L http://localhost:50281/?check_deps=true` and verify the output is `{"status": "503 Service Unavailable", "dependencies": {"es": ["Elasticsearch could not be reached"], "db": ["Database connection for 'upstream' could not be established", "Database connection for 'api' could not be established"]}}` (error messages shown)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -15,6 +15,11 @@ from sentry_sdk.integrations.falcon import FalconIntegration
 
 from ingestion_server import slack
 from ingestion_server.constants.media_types import MEDIA_TYPES, MediaType
+from ingestion_server.db_helpers import (
+    DB_UPSTREAM_CONFIG,
+    DB_API_CONFIG,
+    database_connect,
+)
 from ingestion_server.es_helpers import elasticsearch_connect, get_stat
 from ingestion_server.indexer import TableIndexer
 from ingestion_server.state import clear_state, worker_finished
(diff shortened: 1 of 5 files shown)


sarayourfriend (approved):
This LGTM, it would be nice to add unit tests, though.

I know I opened the issue... but I was initially a little sceptical about whether this is actually needed. We never ended up using the API healthcheck features. But I've got the answer wrong, really. After adding these, we should go back and actually start using the API and these ones _somewhere_, maybe in the load balancer healthcheck for the API, and we can update the ingestion server healthcheck playbook to use these new parameters (not sure where else we actually use the ingestion server healthcheck).

dhruvkb (approved):
Functionally LGTM. In terms of the schema of the response, `status` as a collection of message strings feels a bit unstructured. This is my suggestion for a bit more structured (albeit verbose) way of conveying the info.

```ts
interface Response {
  status: "200 OK" | "503 Service Unavailable"
  dependencies: { // only if `check_deps=true`
    es: {
      isOk: boolean
      messages: string[]
    },
    // ...
  }
}
```

AetherUnbound:
Thanks for the suggestion @dhruvkb! I updated the response to the following, since the presence of the dependency in the response should indicate whether or not it's OK:

```typescript
interface Response {
  status: "200 OK" | "503 Service Unavailable"
  dependencies: { // only if `check_deps=true`
    es: string[], // only if es problems
    db: string[], // only if db problems
    // ...
  }
}
```

PR #3596, opened 2023-12-28
Don't trigger PR limit reminder on PRs that are ignored

AetherUnbound:
## Description

This PR attempts to fix an issue wherein our PR limit reminder bot will ping you even if your most recent PR is one of the PRs that can be safely ignored for the check (e.g. documentation or critical). While the count only includes the non-ignored PRs, the ping is still issued if the user submits another, ignorable PR. I found this issue because I submitted 3 documentation PRs over the limit, but each bot reminder said "you have 4 PRs open". 

## Testing Instructions

I have no idea if this works or is even valid! Opening this to try and find out :) (CC @dhruvkb as you might have the most familiarity with GHA)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/pr_limit_reminders.yml
+++ b/.github/workflows/pr_limit_reminders.yml
@@ -11,6 +11,7 @@ jobs:
     outputs:
       pr_count: ${{ steps.count-prs.outputs.pr_count }}
       slack_id: ${{ steps.count-prs.outputs.slack_id }}
+      should_alert: ${{ steps.count-prs.outputs.should_alert }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/github-script@v6
@@ -26,7 +27,7 @@ jobs:
   send_message:
     needs: analyze-user-prs
     name: Send Slack message
-    if: ${{ needs.analyze-user-prs.outputs.pr_count >= 3 }}
+    if: ${{ needs.analyze-user-prs.outputs.pr_count >= 3 && needs.analyze-user-prs.outputs.should_alert }}
     runs-on: ubuntu-latest
     env:
       slack_url: ${{ secrets.SLACK_DM_WEBHOOK_URL }}
--- a/automations/js/src/count_user_reviewable_prs.js
+++ b/automations/js/src/count_user_reviewable_prs.js
@@ -78,13 +78,26 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
       }
     }
 
+    let should_alert = false
+    if (reviewablePRs.length >= 1) {
+      // TODO sort by open date
+      should_alert = !reviewablePRs[0].labels.nodes.some((label) =>
+        ignoredLabels.includes(label.name)
+      )
+    }
+
     const result = {
       pr_count: reviewablePRs.length,
       slack_id: slackID,
+      should_alert: should_alert,
     }
     core.info(`Current user has ${result.pr_count} PR(s).`)
+    core.info(
+      `Most recent PR should trigger alert based on labels: ${result.should_alert}`
+    )
     core.setOutput('pr_count', result.pr_count)
     core.setOutput('slack_id', result.slack_id)
+    core.setOutput('should_alert', result.should_alert)
   } catch (error) {
     core.setFailed(`Error fetching pull requests: ${error.message}`)
   }


AetherUnbound:
I _just_ got pinged again for this PR 😆 which means that drafts _also_ make this happen, so I'll need to rework this a bit to make it more resilient to the other checks we make.

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -78,13 +77,26 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
       }
     }
 
+    let should_alert = false
```
sarayourfriend, line 80: We use camel case for JavaScript variable names.

I wonder why prettier/eslint doesn't catch this? :thinking: 

```suggestion
    let shouldAlert = false
```

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -38,6 +38,7 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
           }
         }
         isDraft
+        createdAt
```
sarayourfriend, line 41: The event that triggers this script already has the PR in context (https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) from which we can retrieve the latest PR. @dhruvkb told me GitHub's rate limits for GraphQL are a bit different (https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api#primary-rate-limit) so anything we can avoid requesting would be good to avoid.

Instead of getting the most recent PR this way, we can pass the `pull_request` context object to the script and check if it's a relevant PR at the top, and avoid making any requests at all if it isn't. Something like:

```
def (..., pullRequest):
  if not isRelevantPR(pullRequest):
    return { pr_count: 0 }

  # the rest of the function
```

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -49,12 +50,15 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
     '🟥 priority: critical',
   ]
   const [owner, repo] = GITHUB_REPOSITORY.split('/')
+  const isValidPR = (pr) =>
```
sarayourfriend, line 53: Just a nit to suggest "relevant" rather than "valid". Valid makes it sound like we're asking whether the PR is a real PR, but relevant captures what we actually want to know: whether it's a PR we care to evaluate in this context.

sarayourfriend (changes requested):
I think this is a correct change conceptually. I am requesting changes just to shortcut making any requests to save us on the GitHub rate limits and (I think) also really significantly simplify this by-pass edge case. Even in reviewing this PR, the way this fixed the issue was inscruitble to me for almost 10 minutes while I read and re-read the code. Reading it after the fact, I don't think I'd ever understand that this edge case was handled in the way that it is. Adding an early return and avoiding the entire operation would be clearer as well as avoiding the rate limit hit.

automations/js/src/count_user_reviewable_prs.js
AetherUnbound: Oops 😅 force of habit!

automations/js/src/count_user_reviewable_prs.js
AetherUnbound: It looks like we have a few helper methods for getting the pull request object in other scripts:

https://github.com/WordPress/openverse/blob/8b0ec498e12d73a0fc3499a5a8f5753911b2fe9c/automations/js/src/label_pr.mjs#L57-L72

https://github.com/WordPress/openverse/blob/8cd1dbcf8861238bac15dcaae28b587f244a377c/automations/js/src/project_automation/prs.mjs#L67-L75

These use GraphQL under the hood themselves, so I'm not sure we'd change cut down much on usage by using this method instead. Is it really as simple as putting `pull_request` in the exports here? https://github.com/WordPress/openverse/blob/b8230bea5c9ca00a14b69c0622f7862fd046fadb/automations/js/src/count_user_reviewable_prs.js#L11

Do you know where I might be able to find the shape of the object that's available when I do? I'm struggling with the documentation for Octokit 😅

sarayourfriend: No, you'd need to get it in from the GitHub action context. You _might_ be able to get it from `context.payload.pull_request`, and that _might_ have the same shape as the `github.event.pull_request` present in the YAML contexts, [which I found a dump of via this SO answer](https://stackoverflow.com/questions/70104600/complete-list-of-github-actions-contexts):

https://gist.github.com/colbyfayock/1710edb9f47ceda0569844f791403e7e#file-github-context-json-L19

So something like `context.payload.pull_request.created_at` could get you the creation date.

Those other scripts you found are pulling the PR information out of a temporary file built from the event context, and is only necessary because of what Dhruv explains in this comment: https://github.com/WordPress/openverse/blob/889c3b5959a26c18cf09b8be652b7992640f527d/.github/workflows/pr_automations_init.yml#L1-L18

It isn't relevant for this particular script because this script only runs on `pull_request_target` and no other event type, and so will always have the `pull_request` in the event context (unlike `pull_request_review`).

If for some reason `context.payload.pull_request` is _not_ a full `pull_request` object, then you can add it as a property on the ` module.exports = async ({ github, context, core }) => { ` you shared as:

```js
 module.exports = async ({ github, context, core, triggeringPullRequest }) => { 
```

and then pass that in the workflow file like so:

https://github.com/WordPress/openverse/blob/889c3b5959a26c18cf09b8be652b7992640f527d/.github/workflows/pr_limit_reminders.yml#L24

to

```yml
await script({ github, context, core, triggeringPullRequest: ${{ toJSON(github.event.pull_request) })
```

I'm hoping you don't need to do that though and that we can get it from `context.payload.pull_request`. I'm sharing this alternative _just in case_ the simpler approach doesn't work. Unfortunately GitHub Action contexts are not super well documented (as you can see from that SO question I linked).

automations/js/src/count_user_reviewable_prs.js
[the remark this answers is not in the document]
AetherUnbound: A few of the other existing variables use `PR` in their name, so I'll leave this for consistency if that's okay!

.github/workflows/pr_limit_reminders.yml
```diff
@@ -26,7 +27,7 @@ jobs:
   send_message:
     needs: analyze-user-prs
     name: Send Slack message
-    if: ${{ needs.analyze-user-prs.outputs.pr_count >= 3 }}
+    if: ${{ needs.analyze-user-prs.outputs.pr_count >= 3 && needs.analyze-user-prs.outputs.should_alert }}
```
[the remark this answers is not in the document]
AetherUnbound: Oh, good to know! I've removed the `should_alert` output entirely, but I'll remove the brackets on this 🙂

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -78,13 +77,26 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
...
       slack_id: slackID,
+      should_alert: should_alert,
     }
     core.info(`Current user has ${result.pr_count} PR(s).`)
+    core.info(
+      `Most recent PR should trigger alert based on labels: ${result.should_alert}`
+    )
     core.setOutput('pr_count', result.pr_count)
     core.setOutput('slack_id', result.slack_id)
+    core.setOutput('should_alert', result.should_alert)
```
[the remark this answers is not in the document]
AetherUnbound: Thanks for the note! I decided not to go with adding a new output so we should hopefully be able to avoid this for now 😄

automations/js/src/count_user_reviewable_prs.js
AetherUnbound: Thanks for all that info! And amazing find with that gist. I think I've got a workable approach in dfa97590327e9da9c06a20cb00a0e941a01629bd - I had to add a separate function for checking relevancy since the shape of the context's PR is different from the one returned by the GraphQL API. Let me know what you think!

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -50,53 +49,49 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
...
+    const result = {
+      pr_count: 0,
+      slack_id: slackID,
+    }
 
-      const { nodes, pageInfo } = result.repository.pullRequests
-      const relevantPRs = nodes.filter(isRelevantPR)
-      reviewablePRs.push(...relevantPRs)
+    // Check that this pull request is relevant, otherwise skip the action entirely
+    if (isRelevantPRFromContext(pullRequest)) {
```
sarayourfriend, line 70: Nice! This is much easier to understand, at least to me :slightly_smiling_face:

sarayourfriend (approved):
LGTM!

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -49,39 +49,46 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
     '🟥 priority: critical',
   ]
   const [owner, repo] = GITHUB_REPOSITORY.split('/')
+  const isRelevantPRFromGraphQL = (pr) =>
```
dhruvkb, line 52: Nit:

```suggestion
  const isRelevantPrFromGraphql = (pr) =>
```

I don't want to block a PR for a variable name, but I also want to avoid creating something similar to `XMLHttpRequest`.

As for GraphQL, the most common identifier I've seen for it is 'graphql' (all lowercase), so just uppercasing the 'G' for the camelCase version makes sense to me.

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -49,39 +49,46 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
...
+    pr.author.login === context.actor &&
+    !pr.isDraft &&
+    !pr.labels.nodes.some((label) => ignoredLabels.includes(label.name))
+  const isRelevantPRFromContext = (pr) =>
```
dhruvkb, line 56: Nit:

```suggestion
  const isRelevantPrFromContext = (pr) =>
```

automations/js/src/count_user_reviewable_prs.js
```diff
@@ -49,39 +49,46 @@ query ($repoOwner: String!, $repo: String!, $cursor: String) {
...
+    !pr.draft && !pr.labels.some((label) => ignoredLabels.includes(label.name))
 
   try {
     let hasNextPage = true
     let cursor = null
     let reviewablePRs = []
+    const pullRequest = context.payload.pull_request
```
dhruvkb, line 63: This should be fine based on the shape of `Context`: https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts

dhruvkb (approved):
LGTM! This is a welcome improvement to these notifications!

While I appreciate the handling of pagination in the GraphQL call, my YAGNI-sense is tingling because we never reach that many open PRs and there are many other pieces of code where we do not do this. Losing it can significantly reduce the complexity.

automations/js/src/count_user_reviewable_prs.js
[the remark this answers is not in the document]
dhruvkb: Awesome! Best to avoid these complexities/oddities altogether.

automations/js/src/count_user_reviewable_prs.js
AetherUnbound: Updated!

Issue #416, opened 2023-02-17
Data migration protocols


obulat:
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| YYYY-MM-DD | YYYY-MM-DD | TBD          | TBD              |

## Description

Develop a runbook for adding columns and TSV versioning.
Add an automated migration system.
General documentation on procedures and practices for when and how to run migrations. Avoid hindering deployment by long data migrations.

This project is a prerequisite for adding more media types.
This project might require different implementation plans for the API and the Catalog. 
Related to the API ECS migration.

This project also may be a prerequisite for backpropagation

## Documents

- [ ] Project Proposal
- [ ] Implementation Plan

## Prior Art

- [[Infrastructure] Investigate the use of alembic for openledger migrations](https://github.com/WordPress/openverse-catalog/issues/235)

PR #3620, opened 2024-01-03
Add media properties generation

### LINKED ISSUE #2186: Create a script to automatically generate media properties
## Problem

This is the first part of the [Documenting Media Properties in the Catalog](https://docs.openverse.org/projects/proposals/document_all_media_properties/20230426-implementation_plan_documenting_media_properties.html) project.

## Description

Create the script that parses SQL DDLs and Python column definition files to generate the markdown file for the documentation site.

Doing steps 1 and 2 from the [implementation plan](https://docs.openverse.org/projects/proposals/document_all_media_properties/20230426-implementation_plan_documenting_media_properties.html#outlined-steps-parallelizable-workstreams) in the same PR will make testing this much easier: you would just invoke the just script to run the documentation generation.

obulat:
## Fixes

Fixes #2186 by @obulat

## Description

This PR adds the page with documentation on the media properties: a table with all of the audio and image properties with their database and Python characteristics.

The main part of this PR, though, is the script that parses the SQL definition files and the Python column definitions to create the media property tables, and the `catalog/utilities/media_props_gen/media_props.md` file that holds the more detailed descriptions of the media properties. This latter markdown file in this PR only contains the list of properties and the headings for descriptions. It will be filled in in the follow-up PR.

A new `just catalog/generate-media-props` script runs the parsers and markdown generators. If the generated markdown file is different from the file that was saved previously, a warning is issued to run the script again and commit the updated markdown file.

To parse the Python column definitions, this PR uses the `ast` module.

## Testing Instructions

Change some property in the SQL file, the `catalog/dags/common/storage/columns.py` file or the `catalog/utilities/media_props_gen/media_props.md` file, and run `just catalog/generate-media-props`. The `documentation/meta/media_properties.md` file should reflect the changes, and a warning should appear in the console.

Also, review the media properties table and the full document from [the comment below](https://github.com/WordPress/openverse/pull/3620#issuecomment-1980754953), and see if anything needs to be changed.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
@obulat, Would you be working on this PR in the following days? Or should it be closed?

catalog/utilities/media_props_gen/column_parser.py
```diff
@@ -0,0 +1,125 @@
...
+COLUMN_PROPS = COLUMN.keys()
+
+
+def file_to_ast(file_name) -> ast.Module:
+    with open(file_name) as f:
+        file_contents = f.read()
+    return ast.parse(file_contents)
+
+
+CODE = file_to_ast(COLUMNS_PATH)
```
AetherUnbound, line 29: Since `COLUMNS_PATH` is a `pathlib.Path` object, this can just be:
```suggestion
CODE = ast.parse(COLUMNS_PATH.read_text())
```

catalog/utilities/media_props_gen/column_parser.py
```diff
@@ -0,0 +1,125 @@
...
+    python_column: dict[str, any],
+    python_column_lines: dict[str, tuple[int, int]],
+) -> str:
+    """
+    Format the Python column properties dictionary to a string that can be
+    used in the markdown file.
+    """
+    col_type = python_column.pop("python_type")
+    start, end = python_column_lines[col_type]
+    python_column_string = f"[{col_type}]({COLUMNS_URL}#L{start}-L{end}) (`"
```
AetherUnbound, line 42: Love this!

catalog/utilities/media_props_gen/column_parser.py
```diff
@@ -0,0 +1,125 @@
...
+        for item in ast.iter_child_nodes(CODE)
+        if isinstance(item, ast.Assign)
+        and isinstance(item.value, ast.Call)
+        and isinstance(item.value.func, ast.Name)
+        and item.value.func.id.endswith("Column")
+    ]
+
+    columns = {}
+    for col in cols:
+        parsed_column = copy.copy(COLUMN) | {
```
AetherUnbound, line 107: I think defining `COLUMN` as a [`dataclass`](https://docs.python.org/3/library/dataclasses.html) or even a typed `NamedTuple` here might be a little easier 🤔 we could define defaults in the same way and then just splat the keywords into the constructor at this step. Some of this logic could even be in the `__post_init__` function rather than having to access dictionary keys repeatedly! It could help make the code a bit more maintainable down the line as well if the types/attributes are clearly defined.

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+    "integer",
+    "boolean",
+    "uuid",
+    "double precision",
+    "jsonb",
+    "timestamp with time zone",
+    "character varying",
+]
+sql_type_regex = re.compile(f"({'|'.join(sql_types)})")
+MediaType = Literal["audio", "image"]
```
AetherUnbound, line 22: This is already present in `common/constants.py`:

https://github.com/WordPress/openverse/blob/002066fdea540ebba81c7574df526c7db285401d/catalog/dags/common/constants.py#L13

I know that this is able to run outside the docker container presently, but if we followed a similar pattern as the DAG documentation generation script, that would mean we could reuse _even more_ of the `just` recipe when combining the logic for those two! What do you think?

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+
+def create_db_props_dict(
+    media_type: MediaType,
+) -> dict[Any, Any] | dict[Any, dict[str, FieldSqlInfo]]:
+    """
+    Parse the DDL for a media type and returns a list of field
+    sql definitions.
+    """
+
+    create_table_regex = re.compile(r"CREATE\s+TABLE\s+\w+\.(\w+)\s+\(([\s\S]*?)\);")
```
AetherUnbound, line 40: Can this be defined at the top with the other regex constants?

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+    sql_path = SQL_PATH[media_type]
+
+    with open(sql_path) as f:
+        contents = f.read()
+        table_description_matches = create_table_regex.search(contents)
```
AetherUnbound, line 45: Same note here about `Path::read_text()`:
```suggestion
    table_description_matches = create_table_regex.search(sql_path.read_text())
```

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+sql_types = [
+    "integer",
+    "boolean",
+    "uuid",
+    "double precision",
+    "jsonb",
+    "timestamp with time zone",
+    "character varying",
+]
+sql_type_regex = re.compile(f"({'|'.join(sql_types)})")
```
AetherUnbound, line 21: Nit: can these constants be in caps case?

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+        return {}
+    field_descriptions = [
+        field.strip()
+        for field in table_description_matches.group(2).split("\n")
+        if field.strip()
+    ]
+    fields = {}
+    for field in field_descriptions:
+        field_name = field.split(" ")[0]
+        # False if "not null" in field.lower() else True
```
AetherUnbound, line 61: Should this be here?

catalog/utilities/media_props_gen/db.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    create_table_regex = re.compile(r"CREATE\s+TABLE\s+\w+\.(\w+)\s+\(([\s\S]*?)\);")
+    sql_path = SQL_PATH[media_type]
+
+    with open(sql_path) as f:
+        contents = f.read()
+        table_description_matches = create_table_regex.search(contents)
+    if not table_description_matches:
+        print(f"Could not find table description for {media_type} in {sql_path}")
```
AetherUnbound, line 47: This module is doing printing but the generate media properties script is using logging. Do you mind converting the `print`s into `log.info` calls for consistency?

catalog/utilities/media_props_gen/generate_media_properties.py
```diff
@@ -0,0 +1,137 @@
...
+log = logging.getLogger(__name__)
+# Silence noisy modules
+logging.getLogger("common.storage.media").setLevel(logging.WARNING)
+
+# Constants
+PARENT = Path(__file__).parent
+DOC_MD_PATH = PARENT / "media_properties.md"
+SOURCE_MD_PATH = PARENT / "media_props.md"
+
+PREAMBLE = open(Path(__file__).parent / "preamble.md").read()
```
AetherUnbound, line 21: We have a `postamble` file defined as well, but no variable for it. Should it be referenced here or removed?

catalog/utilities/media_props_gen/generate_media_properties.py
```diff
@@ -0,0 +1,137 @@
...
+
+MEDIA_TYPES: list[MediaType] = ["audio", "image"]
```
AetherUnbound, line 23: Same note about constants here: 
https://github.com/WordPress/openverse/blob/002066fdea540ebba81c7574df526c7db285401d/catalog/dags/common/constants.py#L11

catalog/utilities/media_props_gen/md.py
```diff
@@ -0,0 +1,41 @@
...
+        return f"{text}\n"
+
+    @staticmethod
+    def parse(file_name: Path) -> dict[str, dict[str, str]]:
+        """
+        Parse the markdown documentation file and return a dictionary with the
+        field name as key and the description as value.
+        """
+        with open(file_name) as f:
+            contents = [line for line in f.readlines() if line.strip()]
```
AetherUnbound, line 22:
```suggestion
        contents = [line for line in file_name.read_text().split("\n") if line.strip()]
```

catalog/utilities/media_props_gen/md.py
```diff
@@ -0,0 +1,41 @@
...
+        properties = {}
+        prop = ""
+        value = {}
+        for i, line in enumerate(contents):
+            if line.startswith("# "):
+                if current_field and value:
+                    properties[current_field] = value
+                current_field = line.replace("# ", "").strip()
+                value = {}
+                continue
```
AetherUnbound, line 33: I think due to the nature of the if/else block here, the `continue` statements are unnecessary here and below.

catalog/utilities/media_props_gen/generate_media_properties.py
```diff
@@ -0,0 +1,137 @@
...
+        media_types = [
+            f"`{media_type}`"
+            for media_type, value in media_properties.items()
+            if prop in value.keys()
+        ]
+        prop_heading += f"_Media Types_: {', '.join(media_types)}\n\n"
+
+        prop_doc = "".join(
+            [f"{Md.heading(4, k)}{Md.line(v)}" for k, v in description.items()]
+        )
```
AetherUnbound, line 100: The generated document has a lot of headers that have empty content (at this stage). I wonder, do you think it makes sense to only add these subheadings if there is actually content? E.g.:
```suggestion
        prop_doc = "".join(
            [f"{Md.heading(4, k)}{Md.line(v)}" for k, v in description.items() if v]
        )
```

This might make the document less verbose when we don't have everything filled out, even if the template has all those fields. Here's a screenshot (with the line breaks I added in the review comment):
image

AetherUnbound (commented):
This is an excellent start!! I'm really excited to see us make progress on this and I'm loving the final output of the generation. I have a number of structure/code suggestions, as well as potential formatting improvements. I think we will also need tests for most of the utilities that are added here.

Additionally, the media properties themselves feel pretty squished together and difficult to delineate. I think adding a horizontal bar between each property can help visually differentiate the various properties, without affecting the table of contents in any way. Below is some code to accomplish that, and some screenshots of the difference:

```diff
diff --git a/catalog/utilities/media_props_gen/generate_media_properties.py b/catalog/utilities/media_props_gen/generate_media_properties.py
index 0505af318..106cb037d 100644
--- a/catalog/utilities/media_props_gen/generate_media_properties.py
+++ b/catalog/utilities/media_props_gen/generate_media_properties.py
@@ -98,7 +98,7 @@ def generate_long_form_doc(markdown_descriptions: dict, media_properties: dict)
         prop_doc = "".join(
             [f"{Md.heading(4, k)}{Md.line(v)}" for k, v in description.items()]
         )
-        media_docs += prop_heading + prop_doc
+        media_docs += prop_heading + prop_doc + Md.horizontal_line
 
     return media_docs
 
diff --git a/catalog/utilities/media_props_gen/md.py b/catalog/utilities/media_props_gen/md.py
index 8010fb011..2fb3a7a09 100644
--- a/catalog/utilities/media_props_gen/md.py
+++ b/catalog/utilities/media_props_gen/md.py
@@ -2,6 +2,9 @@ from pathlib import Path
 
 
 class Md:
+
+    horizontal_line = "\n---\n\n"
+
     @staticmethod
     def heading(level: int, text: str) -> str:
         """Add a heading to a markdown string."""
```

**Before**
image
**After**
image

catalog/utilities/media_props_gen/preamble.md
```diff
@@ -0,0 +1,10 @@
+# Media Properties
+
+_This document is auto-generated from the source code in
+[/catalog/utilities/media_props_gen/generate_media_properties.py](https://github.com/WordPress/openverse/blob/main/catalog/utilities/media_props_gen/generate_media_properties.py)._
+
+This is a list of the media properties, with the descriptions of corresponding
+database columns and Python objects that are used to store and retrieve media
+data. The order of the properties corresponds to their order in the `image_view`
+materialized view. Property names typically match those of the database columns,
```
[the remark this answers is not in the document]
obulat: This shows the age of this PR :)

catalog/utilities/media_props_gen/generate_media_properties.py
obulat: I added the POSTAMBLE here and in the document.

catalog/justfile
```diff
@@ -142,6 +142,27 @@ generate-dag-docs fail_on_diff="false":
       fi
     fi
 
+
+# Generate the DAG documentation
+generate-media-props fail_on_diff="true":
```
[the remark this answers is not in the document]
obulat: I tried moving this workflow into docker, but then realized that we don't mount the local `.sql` files that we parse to generate the db docs into the container. Here's the error I got:
```
FileNotFoundError: [Errno 2] No such file or directory: '/opt/airflow/docker/upstream_db/0006_openledger_audio_schema.sql'
```
Is it possible to open them from the host, @AetherUnbound ? Seems like this would make the setup unnecessarily complex.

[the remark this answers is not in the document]
AetherUnbound: ohhhh dang, I didn't even think about having to mount those files 😮 We can actually mount those files relatively easily in the recipe at least, it would just require modifying the `just ../run` command here: 

https://github.com/WordPress/openverse/blob/78d799c4d42f2708fbe9149321ff58aec1126d7f/catalog/justfile#L127-L129

Something similar to the way the tests are mounted (note the `--volume` here):

https://github.com/WordPress/openverse/blob/78d799c4d42f2708fbe9149321ff58aec1126d7f/catalog/justfile#L101-L108

Even though it's not needed for the DAG markdown generation, it can be done for all the generation recipes anyway!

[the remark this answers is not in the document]
obulat: Thank you, I'll try that.
Not related to this PR, but is the process of keeping those files up-to-date automated?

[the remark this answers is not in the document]
AetherUnbound: It is not! I believe that's the purview of #416 😄

obulat:
The code in this PR was substantially refactored and expanded:
- the code was organized into `helpers` folder for SQL file, Python source code parsing functions and `docs` folder for the markdown source files
- added tests 
- added `<hr>` horizontal lines in the long-form documentation. This turned out to be more difficult than I expected because prettier converts any sequence of hyphens longer than 3 symbols (e.g., "----") to "---", but Sphinx expects 4 hyphens for an `<hr>`.
- just script was updated to be used for both DAG docs and the media props docs. Both run inside Docker now, and both are ran in the "Run catalog checks" step in the CI.

After the incident caused by #3930, I think we should also parse the code to generate the documentation on the contents of JSONFields and ArrayFields and the possible values for fields inside them. We could probably do this by using dataclasses for properties like `tags`, `meta_data` or `alt_files`, and then parsing the Python code AST files. However, I think it should be done in a separate PR. What do you think, @AetherUnbound?

catalog/tests/utilities/media_props_gen/test_generate_media_properties.py
```diff
@@ -0,0 +1,117 @@
...
+            ),
+            ("standardized_popularity", "double precision, nullable"),
+            id="double_precision_nullable",
+        ),
+    ],
+)
+def test_generate_db_props_string(field, expected_output):
+    field_name, props_string = generate_db_props_string(field)
+    assert field_name == f"[`{expected_output[0]}`](#{expected_output[0]})"
+    assert props_string == expected_output[1]
```
AetherUnbound, line 117: Instead of indexing into the field, I might recommend unpacking the tuple here:
```suggestion
def test_generate_db_props_string(field, expected_output):
    expected_column, expected_props = expected_output
    field_name, props_string = generate_db_props_string(field)
    assert field_name == f"[`{expected_column}`](#{expected_column})"
    assert props_string == expected_props
```

catalog/utilities/media_props_gen/generate_media_properties.py
```diff
@@ -1,27 +1,36 @@
...
 from dataclasses import dataclass
 from pathlib import Path
 
-from column_parser import parse_python_columns
-from db import create_db_props_dict
-from md import Md
 
-from common.constants import MEDIA_TYPES
+# Avoid import error in tests
+sys.path.insert(0, str(Path(__file__).parents[3]))
```
AetherUnbound, line 10: This feels like it might not be advisable...is it necessary? I know we did **not** have to do something similar for the DAG documentation generation 🤔

.prettierignore
```diff
@@ -2,6 +2,10 @@
 .github/**/*.md
 .github/CODEOWNERS
 
+# Documentation
+# This file is linted when created, but the hr elements are replaced for correct rendering.
+documentation/meta/media_properties.md
```
AetherUnbound, line 7: It looks like the linting step in the `justfile` happens _after_ the file is moved here, not before. So I think this will prevent any linting from happening at all 😞

AetherUnbound (commented):
I cannot stress how exciting this is! 🤩 It's so neat that we're going to have this _in_ our documentation site. I love the reference linking, and thanks for making the changes that you did. I have a concern about the prettier ignoration though.

catalog/utilities/media_props_gen/generate_media_properties.py
obulat: DAG documentation is all in one file.
Here, I broke the code into several files. Without this, there's a "Module not found" error when testing. 
Could we somehow update the path when testing so we don't need this line?

.prettierignore
obulat: Fixed in [b5cd89b](https://github.com/WordPress/openverse/pull/3620/commits/b5cd89b8aa65f3612b6dc9c2a01489d5a86721d4)

catalog/tests/utilities/media_props_gen/test_column_parser.py
```diff
@@ -0,0 +1,65 @@
...
+            f"{COLUMN_DEFINITIONS['StringColumn']} (`upsert_strategy=newest_non_null, nullable=True, required=False`)",
+            id="default upsert strategy",
+        ),
+        pytest.param(
+            {"name": "genres", "required": False, "base_column": "StringColumn"},
+            True,
+            f"{COLUMN_DEFINITIONS['ArrayColumn']} "
+            f'(`name="genres", upsert_strategy=newest_non_null, nullable=True, required=False`)',
+            id="base_column",
+            marks=pytest.mark.xfail(reason="base_column is not implemented"),
```
AetherUnbound, line 43: I think we want to use `pytest.mark.raises` here instead (since we can assert that the test raises a specific exception). Using `xfail` here will mean that _any_ failure, even an unrelated one, will "pass" this test.

catalog/utilities/media_props_gen/generate_media_properties.py
AetherUnbound: I tried experimenting with this, and realized we could make _huge_ improvements here by deleting one (1) file 😆: https://github.com/WordPress/openverse/pull/4029

If you're able to rebase your PR once this is merged, this should prevent having to alter the system path like this at all!

catalog/justfile
```diff
@@ -120,24 +120,46 @@ test *args:
 # Utilities #
 #############
 
-# Generate the DAG documentation
-generate-dag-docs fail_on_diff="false":
+# Generate the documentation
```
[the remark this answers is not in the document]
AetherUnbound: @obulat this doesn't look like it was changed

AetherUnbound:
@obulat I rebased this onto #4029 and experimented a bit. My assumption was that python would always add the _working directory_ to the path, but as you noted it adds the _directory the script is in_ to the path. We can correct this by altering the `PYTHONPATH` environment variable when running these generation scripts (pytest does this automatically, which is why the tests were working fine).

I prefer this approach because altering the `sys.path` variable _in flight_ feels very worrisome to me 😅 I'd much rather us explicitly define the path at the start of runtime using an environment variable than get into a pattern of changing the import path set while code is running, if possible.

Here's the diff:

```diff
diff --git a/catalog/justfile b/catalog/justfile
index 9f93f2090..be496e8df 100644
--- a/catalog/justfile
+++ b/catalog/justfile
@@ -141,6 +141,7 @@ generate-docs doc="dag" fail_on_diff="false":
     GENERATED_ABS_PATH="/opt/airflow/catalog/${GENERATED_REL_PATH}"
     just ../run \
       --volume {{ justfile_directory() }}/../docker:/opt/airflow/docker/ \
+      -e PYTHONPATH=/opt/airflow/catalog:/opt/airflow/catalog/dags \
       {{ SERVICE }} \
       "bash -c 'python $SCRIPT_PATH && chmod 666 $GENERATED_ABS_PATH'"
     TEMP="documentation/meta/temp"
diff --git a/catalog/tests/utilities/media_props_gen/test_column_parser.py b/catalog/tests/utilities/media_props_gen/test_column_parser.py
index 6b4d5e1f2..754b10979 100644
--- a/catalog/tests/utilities/media_props_gen/test_column_parser.py
+++ b/catalog/tests/utilities/media_props_gen/test_column_parser.py
@@ -1,6 +1,6 @@
 import pytest
 
-from catalog.utilities.media_props_gen.helpers.column_parser import (
+from utilities.media_props_gen.helpers.column_parser import (
     COLUMN_DEFINITIONS,
     Column,
 )
diff --git a/catalog/tests/utilities/media_props_gen/test_db.py b/catalog/tests/utilities/media_props_gen/test_db.py
index 7bda0814c..fbf29a329 100644
--- a/catalog/tests/utilities/media_props_gen/test_db.py
+++ b/catalog/tests/utilities/media_props_gen/test_db.py
@@ -1,6 +1,6 @@
 import pytest
 
-from catalog.utilities.media_props_gen.helpers.db import (
+from utilities.media_props_gen.helpers.db import (
     FieldSqlInfo,
     get_table_description_matches,
     parse_field,
diff --git a/catalog/tests/utilities/media_props_gen/test_generate_media_properties.py b/catalog/tests/utilities/media_props_gen/test_generate_media_properties.py
index 24b0fb8ae..86eedbd5f 100644
--- a/catalog/tests/utilities/media_props_gen/test_generate_media_properties.py
+++ b/catalog/tests/utilities/media_props_gen/test_generate_media_properties.py
@@ -2,14 +2,14 @@ from unittest.mock import patch
 
 import pytest
 
-from catalog.utilities.media_props_gen.generate_media_properties import (
+from utilities.media_props_gen.generate_media_properties import (
     generate_db_props_string,
     generate_markdown_doc,
 )
-from catalog.utilities.media_props_gen.helpers.db import FieldSqlInfo
+from utilities.media_props_gen.helpers.db import FieldSqlInfo
 
 
-DIRECTORY = "catalog.utilities.media_props_gen"
+DIRECTORY = "utilities.media_props_gen"
 MODULE = f"{DIRECTORY}.generate_media_properties"
 
 # Sample mock data
diff --git a/catalog/tests/utilities/media_props_gen/test_md.py b/catalog/tests/utilities/media_props_gen/test_md.py
index 9c424e361..2831a4426 100644
--- a/catalog/tests/utilities/media_props_gen/test_md.py
+++ b/catalog/tests/utilities/media_props_gen/test_md.py
@@ -1,6 +1,6 @@
 import pytest
 
-from catalog.utilities.media_props_gen.helpers.md import Md
+from utilities.media_props_gen.helpers.md import Md
 
 
 @pytest.mark.parametrize(
diff --git a/catalog/utilities/media_props_gen/generate_media_properties.py b/catalog/utilities/media_props_gen/generate_media_properties.py
index ddc1e9336..ff5f7dadd 100644
--- a/catalog/utilities/media_props_gen/generate_media_properties.py
+++ b/catalog/utilities/media_props_gen/generate_media_properties.py
@@ -1,23 +1,13 @@
 """Automatic media properties generation."""
 
 import logging
-import sys
 from dataclasses import dataclass
 from pathlib import Path
 
-
-# Avoid import error in tests
-sys.path.insert(0, str(Path(__file__).parents[3]))
-
-from catalog.utilities.media_props_gen.helpers.column_parser import (  # noqa: E402
-    parse_python_columns,
-)
-from catalog.utilities.media_props_gen.helpers.db import (  # noqa: E402
-    FieldSqlInfo,
-    create_db_props_dict,
-)
-from catalog.utilities.media_props_gen.helpers.md import Md  # noqa: E402
-from common.constants import MEDIA_TYPES  # noqa: E402
+from utilities.media_props_gen.helpers.column_parser import parse_python_columns
+from utilities.media_props_gen.helpers.db import FieldSqlInfo, create_db_props_dict
+from utilities.media_props_gen.helpers.md import Md
+from common.constants import MEDIA_TYPES
 
 
 log = logging.getLogger(__name__)
```

catalog/utilities/media_props_gen/docs/preamble.md
```diff
@@ -0,0 +1,10 @@
+# Media Properties
```
AetherUnbound, line 1: To match the pattern of #3966:
```suggestion
# Catalog Media Properties
```

AetherUnbound (approved):
Fantastic!! So excited for this, I have one recommendation but nothing to block a merge 🚀

catalog/utilities/media_props_gen/helpers/db.py
```diff
@@ -0,0 +1,100 @@
...
+    "audio": LOCAL_POSTGRES_FOLDER / "0006_openledger_audio_schema.sql",
+}
+SQL_TYPES = [
+    "integer",
+    "boolean",
+    "uuid",
+    "double precision",
+    "jsonb",
+    "timestamp with time zone",
+    "character varying",
```
dhruvkb, line 22: Interestingly, the output from the API table shows this as `varchar(<length>)` and not `character varying`.

dhruvkb (approved):
LGTM! 

I feel like keeping the notes for fields that are specific to a single media type (like `height` and `width` for images, and `duration` and `genre` for audio) below the respective media type would be better because having them all together at the bottom is slightly confusing. But I will make a separate issue for that and not hold up this perfectly good PR.

catalog/utilities/media_props_gen/helpers/db.py
obulat: TIL: In Postgres, you can use both interchangeably (https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER).
I guess we could replace `character varying` with `varchar(<length>)` in the catalog docs and in the local catalog SQL files for consistency.

catalog/utilities/media_props_gen/generate_media_properties.py
```diff
@@ -0,0 +1,147 @@
...
+PREAMBLE = open(DOCS / "preamble.md").read()
+POSTAMBLE = open(DOCS / "postamble.md").read()
+
+
+@dataclass
+class FieldInfo:
+    name: str
+    nullable: bool
+    datatype: str
+    constraint: str
```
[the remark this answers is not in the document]
obulat: You're right. I was trying to explicitly state everywhere whether a value is nullable or not to prevent the errors when we have an unexpected `NULL` in a field upstream. Do you think it's better to convert it to a part of a constraint? I will open an issue for it if you think so, @dhruvkb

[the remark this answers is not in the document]
dhruvkb: It's okay as it is, makes the nullability easier to read at a glance. I'm just noting a difference between the this and the [API docs](https://docs.openverse.org/meta/media_properties/api.html) which put `not null` in the constraints column.

catalog/utilities/media_props_gen/helpers/db.py
dhruvkb: TIL as well. `varchar` is technically shorter so fits better in a narrow table 🤷 but that's not justification to use one over the other.

AetherUnbound: Consistency there would be nice though! Would one of you mind making a follow-up issue for it?

PR #3622, opened 2024-01-03
Convert unittest to pytest

### LINKED ISSUE #3425: Convert unittest.TestCase tests to pytest
## Description

Most of our tests are set up in the functional format that [pytest expects](https://docs.pytest.org/en/7.1.x/example/simple.html). We have two exceptions:

- `catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py`
- `ingestion_server/test/integration_test.py`

Both of these tests use `unittest.TestCase` instead of individual test functions, and both should be able to be converted to pytest-formatted tests. `TestCase`-based tests don't allow the use of fixtures, which hinders the simplicity of some tests (see https://github.com/WordPress/openverse/pull/3361#issuecomment-1819321403).

Examples of other pytest-formatted tests can be found throughout the `catalog/tests` directory.

## Additional context

This came up as part of #3361, and the workarounds created in that PR should be removed when the tests are converted.

ngken0995:
## Fixes

Fixes #3425  by @AetherUnbound 

## Description

On `catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py`, `setUp` function is renamed to `setup_pool`, `set_pool` will have a parameter `sample_pool_fixture` retrieved from conftest.`TestExternalDAGsSensor` class will be removed and all test function will be in the same module. 

On `ingestion_server/test/integration_test.py`, `setUpClass` is rename to `setup_fixture`, `setup_fixture` use scope as module, autouse is set as True  and it will return a dictionary for other function to retrieve them. Install `pytest-subtests` to replace unittest `subTest`.

## Testing Instructions

For catalog:

run `just catalog/test`

For ingestion server:

```
source ingestion_server/venv/bin/activate
pipenv install --dev
just ingestion_server/test_local
```

if ingestion server fail:

run `just ingestion_server/test-clean-dc`
run `just ingestion_server/test_local`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/test/integration_test.py
```diff
@@ -66,575 +65,650 @@ def handle_task_callback():
...
-    def test_stat_endpoint_for_non_existent(self):
-        res = requests.get(f"{ingestion_server}/stat/non-existent")
-        data = res.json()
-        assert not data["exists"]
+            }
+            res = es.search(index=index, query=query)
+            count = res["hits"]["total"]["value"]
+            assert (
+                count == 0
+            ), f"There should be no results that exactly match {sensitive_term} in {index}. Found {count}."
```
sarayourfriend, line 621: I think we only used subtests for this because unittest doesn't support `pytest.mark.parametrize`. If it is possible to convert this to use parametrize and avoid including a new pytest plugin, that would be preferrable. I can't leave a normal PR review suggestion because the diff of this function is split on a weird diff match of `}` :confused:, but here's what I'm thinking might work to avoid the new dependency:

```py
@pytest.mark.order(
    after=["test_point_filtered_audio_alias", "test_point_filtered_image_alias"]
)
@pytest.mark.parametrize(
    "sensitive_term, index",
    zip(mock_sensitive_terms, ["audio-filtered", "image-filtered"]),
)
def test_filtered_indexes(sensitive_term, index):
    """
    Check that the sensitive terms are correctly filtered out.

    Each sensitive term should have zero exact matches for it when querying
    the filtered index. We use `term` against the `.raw` fields to ensure this.
    """
    es = _get_es()
    queryable_fields = ["title", "description", "tags.name"]
    query = {
        "bool": {
            "should": [
                {"term": {f"{field}.raw": sensitive_term}}
                for field in queryable_fields
            ]
        }
    }
    res = es.search(index=index, query=query)
    count = res["hits"]["total"]["value"]
    assert (
        count == 0
    ), f"There should be no results that exactly match {sensitive_term} in {index}. Found {count}."
```

ingestion_server/test/integration_test.py
```diff
@@ -66,575 +65,650 @@ def handle_task_callback():
...
+@pytest.mark.order(after="test_upstream_indexed_images")
+def test_upstream_indexed_audio():
+    """
+    Check that the audio data has been successfully indexed in Elasticsearch.
+
+    The number of hits for a blank search should match the size of the loaded mock
+    data.
+    """
+
+    es = _get_es()
```
sarayourfriend, line 559: Maybe `_get_es` could also be converted to a fixture? It's used as such, but I don't know what the specific benefit would be either way, other than that it _looks_ like a fixture to me.

sarayourfriend (commented):
Looks really good to me, I just have a couple questions for clarification, but nothing is necessarily a blocker :+1:

catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py
```diff
@@ -65,6 +59,16 @@ def create_dagrun(dag, dag_state):
...
 
 
+@pytest.fixture
+def setup_pool(sample_pool_fixture):
+    Pool.create_or_update_pool(
+        sample_pool_fixture,
+        slots=1,
+        description="test pool",
+        include_deferred=False,
+    )
```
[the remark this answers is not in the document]
ngken0995: The `setup_pool` is identical. I will delete it from `conftest`.

ingestion_server/test/integration_test.py
ngken0995: Yes, I have converted `_get_es` to `sample_es` as a fixture.

ingestion_server/test/integration_test.py
ngken0995: Thank you for the suggestion. This is a great to use parametrize instead of a third party plugin.

catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py
```diff
@@ -216,26 +142,127 @@ def test_retries_if_running_dags_with_completed_sensor_task(self):
...
+    "ignore:This class is deprecated. Please use "
+    "`airflow.utils.task_group.TaskGroup`.:airflow.exceptions.RemovedInAirflow3Warning"
+)
+# This also appears to be coming from Airflow internals during testing as a result of
+# loading the example bash operator DAG:
+# /home/airflow/.local/lib/python3.10/site-packages/airflow/models/dag.py:3492: RemovedInAirflow3Warning  # noqa: E501
+@pytest.mark.filterwarnings(
+    "ignore:Param `schedule_interval` is deprecated and will be removed in a future release. "
+    "Please use `schedule` instead.:airflow.exceptions.RemovedInAirflow3Warning"
+)
```
sarayourfriend, line 217: Just want to ping @WordPress/openverse-catalog on this to make sure this is a known deprecation. I couldn't find an issue for resolving it.

sarayourfriend (approved):
LGTM! In looking at the ingestion server tests, I noticed this warning: https://github.com/WordPress/openverse/actions/runs/7405567531/job/20148753309?pr=3622#step:5:31

I was going to open an issue, but saw this comment on the code from @dhruvkb talking about using Bottle instead of Falcon:

https://github.com//WordPress/openverse/blob/95b59110eff953f7d30d9b9444791ae2e0dc6d68/ingestion_server/test/integration_test.py#L20-L21

Dhruv: Do you remember what the issue with Falcon was? Can you open an issue describing it? It would be nice to find a workaround to use Falcon instead or to find an alternative to Bottle, less so because of the deprecation warning and more to simplify the dependencies (which in turn means we also have fewer moving parts to worry about deprecations).

Anyway, that's a separate issue from this PR, I just wanted to bring it up for Dhruv to look at to make sure we have it recorded just in case it's an issue in the future.

catalog/tests/dags/common/sensors/test_single_run_external_dags_sensor.py
stacimc: Both warnings are being thrown by loading Airflow's example DAGs for use in the tests. It looks like the second was resolved on their end -- I just removed the filter and confirmed it's not being raised locally. The TaskGroup warning is still raised.

```suggestion
@pytest.mark.filterwarnings(
    "ignore:This class is deprecated. Please use "
    "`airflow.utils.task_group.TaskGroup`.:airflow.exceptions.RemovedInAirflow3Warning"
)
```

ngken0995: Thank you @sarayourfriend and @stacimc for looking into both warnings

sarayourfriend:
@krysal @obulat are either of y'all able to review this PR this week? It'd be great to merge this.

ingestion_server/test/integration_test.py
```diff
@@ -66,575 +65,634 @@ def handle_task_callback():
...
+            ORDER BY conrelid::regclass::text, contype DESC;
+    """
+    with conn.cursor() as cursor:
+        cursor.execute(constraint_sql)
+        return {constraint: name for name, constraint in cursor}
+
+
+def _get_index_parts(index: str, table: str) -> list[str]:
+    """
+    Strip out common keywords from the index to get a the name & columns.
```
krysal, line 206:
```suggestion
    Strip out common keywords from the index to get the name & columns.
```

krysal (approved):
Great work, @ngken0995! Thank you!

PR #3623, opened 2024-01-03
Add secret-key check

### LINKED ISSUE #3456: Add a check on startup to ensure application is not using default key in live environments
## Description

In #3422, we altered the Django secret key to use a more obvious shim value. @dhruvkb suggested:

> We can also update settings.py to raise an error if it's unchanged.

To be specific, we should check to ensure that the value is not `example_key` in any `ENVIRONMENT` that's not `local` or `development`, and raise an appropriate error if it is.

firatbezir:
## Fixes

Fixes #3456 by @AetherUnbound 

## Description
This pull request updates the Django secret key to use a more evident shim value. Additionally, it enhances the settings.py file to raise an error if the secret key remains unchanged. Specifically, the code checks that the key is not set to "example_key" in any environment other than local or development, triggering an appropriate error if the condition is met.

-- Changes I've made:
image
image

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


firatbezir:
I've made changes to the security.py file with the aim of addressing issue #3456. I believe these modifications should effectively resolve the mentioned problem. As for testing, I didn't explicitly check related items as I'm uncertain about the necessity. Please feel free to reach out and let me know if testing is required. I'm more than willing to handle it promptly.

api/conf/settings/security.py
```diff
@@ -52,3 +54,12 @@
...
     SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
+# Adding DJANGO_SECRET_KEY check
+if (
+    config("DJANGO_SECRET_KEY") == "example_key"
+    and config("ENVIRONMENT", default="local") not in ["local", "development"]
+):
+    raise ImproperlyConfigured(
+        "DJANGO_SECRET_KEY should not be 'example_key' in local or production environment."
+    )
```
obulat, line 65:
```suggestion
    )

```
This file should have a blank line at the end

obulat (commented):
The changes look correct, @firatbezir.

Could you lint your changes? You should be able to run `just lint` in the `openverse` directory, and then commit the changes.

firatbezir:
thanks @obulat, apparently linting is successfull now. please let me know if it requires anything else.

api/conf/settings/security.py
```diff
@@ -52,3 +54,11 @@
...
     # https://docs.djangoproject.com/en/4.0/ref/settings/#secure-proxy-ssl-header
     SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
+# Adding DJANGO_SECRET_KEY check
+if config("DJANGO_SECRET_KEY") == "example_key" and config(
+    "ENVIRONMENT", default="local"
+) not in ["local", "development"]:
+    raise ImproperlyConfigured(
+        "DJANGO_SECRET_KEY should not be 'example_key' in local or production environment."
+    )
```
sarayourfriend, line 64: We should only check if `ENVIRONMENT != "local"`. We've historically used the word "development" to mean both local and staging environments, and it is critical that we continue to disambiguate these as distinct terms (by, in fact, leaving `development` behind entirely because it's too easy to confuse it's meaning). We have no other instances of checking "development" as an environment setting in any of our Python code, so we should not introduce it again here.

`ENVIRONMENT` is already defined in `BASE` and `SECRET_KEY` is already defined in this module (line 12), so we don't need to call `config` again to retrieve them. Doing so introduces multiple places that would need changing if those environment variable names ever changed.

The error message also needs a slight re-wording:

```suggestion
if SECRET_KEY == "example_key" and ENVIRONMENT != "local":
    raise ImproperlyConfigured(
        "DJANGO_SECRET_KEY must not be 'example_key' non-local environments."
    )
```

You'll need to add `ENVIRONMENT` to the list of imported variables from `conf.settings.base` on line 5.

sarayourfriend (changes requested):
This is a great start, thanks @firatbezir. The changes I have requested are in the comment I left with a suggestion for a change to satisfy my request. I've added additional explanation for the sake of other reviewers, in particular for @AetherUnbound as I see that the issue refers to "development" as an environment name, when we should not use this term any longer for the reasons I've shared.

Your implementation is fine for what the issue described, leaving the simplifications of avoiding duplicate calls to `config` aside. Thanks very much for the contribution. Once you implement the simplifications I've requested I will approve this :+1:

sarayourfriend (approved):
LGTM, thanks for the contribution and fast revision @firatbezir :rocket:

api/conf/settings/security.py
```diff
@@ -52,3 +54,9 @@
     USE_X_FORWARDED_HOST = True
     # https://docs.djangoproject.com/en/4.0/ref/settings/#secure-proxy-ssl-header
     SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
+# Adding DJANGO_SECRET_KEY check
+if SECRET_KEY == "example_key" and ENVIRONMENT != "local":
+    raise ImproperlyConfigured(
+        "DJANGO_SECRET_KEY must not be 'example_key' non-local environments."
```
krysal, line 61:
```suggestion
        "DJANGO_SECRET_KEY must not be 'example_key' in non-local environments."
```

PR #2975, opened 2023-09-04
Add 10k throttle to filtered index reindex query

sarayourfriend:
## Fixes

Fixes #2963 by @AetherUnbound 

## Description

I haven't been able to test this locally due to local environment issues preventing upstream db from building and initialising correctly.

I figured I'd open the PR and see how CI handles it and get it ready ASAP in order to prevent any further delays before the next time the reindex comes around.

## Testing Instructions

Existing tests should pass. Audit Elasticsearch request logs if you want to confirm the query parameter is sent.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [N/A] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #3644, opened 2024-01-08
Increase ES throttling rate in Ingestion Server

krysal:
## Fixes

Since the last time the `request_per_second` setting was modified for the filtered index creation in #2975, the ES's average CPU usage has been constant below the 43%. Observe the graph for last six weeks, the high peaks correspond to the base index creation and the following lower raises to the filtered index task.

Screenshot 2024-01-08 at 15-35-01 CloudWatch us-east-1

Having approximately 60% of computing power available we can try increasing this value so that the process finishes faster, or alternatively, removing the throttling entirely since indexes are not promoted until both are updated. Live users shouldn't be affected.

## Description

This PR makes this setting take its value from a new environment value with an increased default value.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/indexer.py
+++ b/ingestion_server/ingestion_server/indexer.py
@@ -46,6 +46,11 @@ DATABASE_USER = config("DATABASE_USER", default="deploy")
 DATABASE_PASSWORD = config("DATABASE_PASSWORD", default="deploy")
 DATABASE_NAME = config("DATABASE_NAME", default="openledger")
 
+# See https://www.elastic.co/guide/en/elasticsearch/reference/8.8/docs-reindex.html#docs-reindex-throttle
+ES_FILTERED_INDEX_THROTTLING_RATE = config(
+    "ES_FILTERED_INDEX_THROTTLING_RATE", default=20_000, cast=int
+)
+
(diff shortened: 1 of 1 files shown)


ingestion_server/ingestion_server/indexer.py
```diff
@@ -523,9 +528,7 @@ def create_and_populate_filtered_index(
             },
             slices="auto",
             wait_for_completion=True,
-            # 10k derived from in-production testing
```
sarayourfriend, line 526: This comment was already out of date :open_mouth:

sarayourfriend (approved):
Nice find! Do API response times still go up a bit during indexing? That's the only reason I could think of to maintain the lower throttling rate and accept the longer amount of time. Not a blocker though, you have a much better handle on these things than I do :slightly_smiling_face: 

Also: I wonder if this scales linearly. If we go from 15 to 20k per second, will we see a 1/3 decrease in the time it takes?

stacimc (approved):
👍 Nice to configure this by environment variable as well. LGTM.

PR #3656, opened 2024-01-11
Simplify load sample data

sarayourfriend:
## Fixes

I set up Openverse from scratch today and ran into a very strange issue with `load_sample_data.sh` that I cannot track down the precise cause of. The version on `main` hangs completely, without any feedback, and apparently passes malformed SQL commands to `psql`.

## Description

The changes in this PR get the script running as expected for me again. They also, I believe, somewhat simplify things, through the introduction of two patterns:

1. The new `exec_sql` function which not only DRYs up the code a bit, but also uses what is apparently a more reliable approach for passing a set of SQL commands to `psql`. The same approach is used for interacting with Redis as well.
2. Taking advantage of bash's built-in multiline string support, rather than using docstring style. This is easier to read and keep track of (IMO) and has less potential variations. It's also easier to assign to variables or pass as arguments to bash functions.

## Testing Instructions

CI should work, all API testing steps rely on this script. You should also be able to run it locally.

I really don't know _why_ the script as-is didn't work on this machine. The only thing I can think of is maybe the bash version is different?

```
$ bash --version
GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -27,8 +27,12 @@ function exec_sql {
 
 # Load sample data
 function load_sample_data {
-  exec_sql "$UPSTREAM_DB_SERVICE_NAME" "DELETE FROM $1"
-  exec_sql "$UPSTREAM_DB_SERVICE_NAME" "\copy $1 from './sample_data/sample_$1.csv' with (FORMAT csv, HEADER true);"
+  exec_sql "$UPSTREAM_DB_SERVICE_NAME" "
+  	DELETE FROM $1;
+	\copy $1 \
+		from './sample_data/sample_$1.csv' \
+		with (FORMAT csv, HEADER true);
+"
 }
 
 function verify_loaded_data {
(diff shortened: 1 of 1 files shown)


dhruvkb:
The changes look fine to me, and the CI tests continue to pass, which means there isn't anything logically wrong here. But the more interesting thing to me here is why the previous version of the script did not work on your machine if it worked everywhere else (like CI, for example).

The Mac users on the team are on very old versions of Bash so, if that script worked for us, it should've worked on all newer versions.

```
❯ bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)
Copyright (C) 2007 Free Software Foundation, Inc
```

load_sample_data.sh
```diff
@@ -56,35 +58,35 @@ fi
...
+usernames = ['continuous_integration', 'deploy']
+for username in usernames:
+	if User.objects.filter(username=username).exists():
+		print(f'User {username} already exists')
+		continue
+	if username == 'deploy':
+		user = User.objects.create_superuser(username, f'{username}@example.com', 'deploy')
+	else:
+		user = User.objects.create_user(username, f'{username}@example.com', 'deploy')
+	user.save()
```
dhruvkb, line 75: The Python part here should be indented with 4 spaces, instead of tabs, as per PEP-8.

dhruvkb (approved):
LGTM, except for a stylistic change and a bigger question around why the script did not work before these refactors.

krysal (approved):
Works locally and the simplication of multiline strings is quite nice!

Issue #3626, opened 2024-01-04
Return 401 for API requests with invalid `Authorization` headers


sarayourfriend:
## Problem

The API automatically downgrades requests with invalid authentication details to an anonymous request. This has two problems:

1. It does not give clear feedback to API users that their credentials are incorrect
2. It can open up ways of bypassing anonymous rate limiting structures that exist before the API handles the request (these do not yet exist but will eventually)

## Description

Requests where `"Authorization" in requests.HEADERS and not request.auth` should be rejected with a 401 response, with an explanation that authentication failed due to invalid credentials.

## Additional context

Related to this issue to combine anonymous rate limiting from the frontend and API: https://github.com/WordPress/openverse-infrastructure/issues/746

PR #3663, opened 2024-01-15
Fix: Return 401 for API request with Invalid Authorization

carlosm22700:
## Fixes
Related to #3626 by @sarayourfriend

## Description
This PR implements middleware 'StrictAuthMiddleware' as a solution to issue [#3626](https://github.com/WordPress/openverse/issues/3626). The goal is to enforce a consistent response across the API for requests with invalid Authorization headers by returning a 401 Unauthorized status. If the request does not contain a header, it proceeds as anonymous. 

During the development and testing phases, I tested different endpoints to see how they would handle various scenarios, including valid and invalid Authorization headers (tokens), as well as requests lacking these headers. By implementing this middleware, my goal was to avoid the need for repetitive modifications across individual views. It also avoids using individual permissions for specific resources and uses a wrapper instead.

However, I ran into some challenges with authenticated requests also returning 401 and the auth tests not passing when running just/api test. I would very much appreciate any feedback on my approach to this!

## Testing
- **Invalid Token:**  
  `curl -H "Authorization: Bearer invalid_token" -I "http://localhost:50280/v1/images/"` 
  `curl -H "Authorization: Bearer invalid_token" -I "http://localhost:50280/v1/checkrates/"`
  Expect a 401 Unauthorized response.

- **No Authorization Header (Anonymous request):**  
  `curl -I "http://localhost:50280/v1/images/"`  
  Expect a 200 OK response, treating the request as anonymous.

## Checklist
- [X] My pull request has a descriptive title.
- [X] Targets the _default_ branch of the repository (`main`).
- [X] My code adheres to the established code style of the repository.
- [ ] Tests for the changes have been added/updated (if applicable).
- [ ] Documentation has been added/updated (if applicable).
- [X] Project runs locally without errors.
- [ ] (If applicable) Ran the DAG documentation generator.wow too mu

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


carlosm22700:
Thanks for the help! @dhruvkb

dhruvkb (changes requested):
A more elegant approach to this would be to use `permission_classes` (I believe L176 was a typo).

```diff
@extend_schema(tags=["auth"])
  class CheckRates(APIView):
      throttle_classes = (OnePerSecond,)
+     permission_classes = (TokenHasScope,)
+     required_scopes = ('read',)
```

This automatically raises 401 when the token is not provided, invalid or expired (and also provides the reason in the `WWW-Authenticated` header as recommended by DRF.

E.g.
```
WWW-Authenticate: Bearer realm="api",error="invalid_token",error_description="The access token has expired."
```
or 
```
WWW-Authenticate: Bearer realm="api",error="invalid_token",error_description="The access token is invalid."
```

krysal:
@carlosm22700 Do you think you could resume this work in the following days? We would appreciate it if you let us know in any case.

carlosm22700:
Sorry for the delay, just wrapped up school for the semester!

Thanks for the review @dhruvkb @krysal. Would either of you mind taking another look to make sure everything looks good? I made sure to implement the changes recommended.

dhruvkb (approved):
Thanks for the contribution @carlosm22700 and for implementing the changes from the review. LGTM!

stacimc (changes requested):
Thanks for your contribution, @carlosm22700! Would you mind updating the PR description to include testing instructions? It looks like the description might also be slightly out of date.

If I'm reading the linked issue correctly, I think we want to return a 401 when an invalid authorization header is sent to any API endpoint. The code changes here only address the CheckRates view. For example, when testing I would expect that if I try:

```
> curl -H "Authorization: Bearer faketoken123" \
  -I "http://localhost:50280/v1/images/"
```

I should get a 401. Instead it continues to silently downgrade it to an unauthorized request (returns successfully until the unauthorized rate limit is exceeded, then 429s).

I'm also still getting a 403 when I call the CheckRates endpoint with a bad token, too 🤔 Though I may be misunderstanding the intended behavior there. @dhruvkb can you weigh in if I'm off base?

sarayourfriend:
You're right about the intended scope, Staci. Any API endpoint sent with invalid authorisation header should raise a 401 to indicate the credentials are invalid, without downgrading the request to anonymous. What this means is:

1. Any request _with_ authorization header should have this check happen
2. Any request _without_ it should bypass the check and be treated as anonymous

To accomplish that, we need a wrapper around the authentication middleware, not a permissions class. The permissions class would be the right choice if we were trying to guard access to particular resources based on the scope of the authenticated request. That's not the goal, though.

If it helps clarify the intention and goal, we could use 400 instead of 401 for these, with a message indicating the authorization header credentials were invalid.

carlosm22700:
Thank you all for your patience! If anyone could take a moment to review this PR or provide feedback, I'd greatly appreciate it! Your help is invaluable as I continue to learn and contribute.

api/api/middleware/strict_auth_middleware.py
```diff
@@ -0,0 +1,21 @@
...
+def strict_auth_middleware(get_response):
+    # Inner function to process each request
+    def middleware(request):
+        # Extract the Authorization header from the request
+        auth_header = request.headers.get('Authorization', None)
+        
+        # If the Authorization header is present
+        if auth_header:
+            # Check if the user is anonymous or authentication failed
+            if request.user.is_anonymous or request.auth is None:
```
sarayourfriend, line 13: Because of how the OAuth2 DRF authenticator works, we can replace all of these checks with a check against `oauth2_error` on the request object:

https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/contrib/rest_framework/authentication.py#L30

The request attribute comes from this method, where you can see various error messages are turned as part of it:

https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L219-L249

We haven't defined set scopes anywhere, so we don't need to worry about `insufficient_scope`, and can just assume the reason is for an invalid token. It'd be nice to return the error description as well, so users understand _why_ it's invalid.

Something along these lines is where I would start:

```py
oauth2_error = getattr(request, "oauth2_error", None)

if oauth2_error is None:
    return get_response(request)

raise AuthenticationFailed(
    detail=oauth2_error.get("error_description", None),
    code=oauth2_error.get("error", None),
)
```

sarayourfriend (changes requested):
You are on the right track, @carlosm22700, but there are two requested changes I've left. This is a bit of a tricky issue, and to understand the approach I've suggested, I had to dig through the [code for `django-oauth-toolkit`](https://github.com/jazzband/django-oauth-toolkit) to understand how it handles errors.

With these requested changes, I'd also like to request new unit tests to cover this, added in `test_auth.py`. The test should make a request with a bad token, and confirm that the response is a 401.

sarayourfriend:
@carlosm22700 have you moved this PR elsewhere/do you plan to continue working on it? Just wanted to clarify since you closed the PR, just let us know :slightly_smiling_face:

PR #3682, opened 2024-01-19
Add check to skip toggling CloudWatch alarms

krysal:
## Fixes

Follow up to #3652.

## Description

@AetherUnbound mentioned in the PR above a problem running the data refresh locally when AWS credentials aren't set (as is usual locally, since they are not necessary): https://github.com/WordPress/openverse/pull/3652#issuecomment-1899405418

This PR adds an Airflow variable, `AIRFLOW_VAR_TOGGLE_CLOUDWATCH_ALARMS`,  to decide when to actually modify Cloudwatch alarms, and change the connection to CloudWatch to use a more idiomatic Airflow AWS connection with a  hook.

### Connection

1. To test the AWS connection, set the `AIRFLOW_CONN_AWS_CLOUDWATCH` variable with the required credentials as shown in the `env.example` file. You could also generate it using [this snippet to create Connection and convert to URI](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/connections/aws.html#snippet-to-create-connection-and-convert-to-uri).
2. Run the catalog and ingestion server stack.
```
just up
```
3. Run a data_refresh DAG (audio or image), wait for it to finish and then observe in CloudWatch the updates of the alarm in the History tab: [ES Production CPU utilization above 50%](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#alarmsV2:alarm/ES+Production+CPU+utilization+above+50$25) 

CleanShot 2024-02-02 at 14 32 53@2x

### Config for whether to toggle the alarm

1. Add the new variable through the Airflow UI, `Header menu > Admin > Variables`, with the value `false`.
```
TOGGLE_CLOUDWATCH_ALARMS
```
2. Run a data refresh again and observe the tasks for enabling/disabling the alarm are skipped.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/cloudwatch.py
+++ b/catalog/dags/common/cloudwatch.py
@@ -4,9 +4,11 @@ https://github.com/awsdocs/aws-doc-sdk-examples/blob/54c3b82d8f9a12a862f9fcec449
 """
 
 import logging
+
 import boto3
-from botocore.exceptions import ClientError
 from airflow.models import Variable
+from botocore.exceptions import ClientError
+
 
 logger = logging.getLogger(__name__)
 
@@ -49,8 +51,8 @@ class CloudWatchWrapper:
 
 
 def enable_or_disable_alarms(enable):
-    skip_toggling = Variable.get("SKIP_TOGGLING_CLOUDWATCH_ALARMS", False)
-    if skip_toggling:
+    toggle = Variable.get("TOGGLE_CLOUDWATCH_ALARMS", True)
+    if not toggle:
         logger.info("Skipping toggling CloudWatch alarms.")
         return
 
--- a/catalog/dags/common/cloudwatch.py
+++ b/catalog/dags/common/cloudwatch.py
@@ -62,5 +62,4 @@ def enable_or_disable_alarms(enable):
         resource_type="cloudwatch",
     )
     cw_wrapper = CloudWatchWrapper(cloudwatch.get_conn())
-
-cw_wrapper.enable_alarm_actions("ES Production CPU utilization above 50%", enable)
+    cw_wrapper.enable_alarm_actions("ES Production CPU utilization above 50%", enable)
--- a/catalog/dags/common/cloudwatch.py
+++ b/catalog/dags/common/cloudwatch.py
@@ -4,6 +4,7 @@ https://github.com/awsdocs/aws-doc-sdk-examples/blob/54c3b82d8f9a12a862f9fcec449
 """
 import logging
 
+from airflow.exceptions import AirflowSkipException
 from airflow.models import Variable
 from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
 from botocore.exceptions import ClientError
@@ -52,8 +53,7 @@ class CloudWatchWrapper:
 def enable_or_disable_alarms(enable):
     toggle = Variable.get("TOGGLE_CLOUDWATCH_ALARMS", True)
     if not toggle:
-        logger.info("Skipping toggling CloudWatch alarms.")
-        return
+        raise AirflowSkipException("TOGGLE_CLOUDWATCH_ALARMS is set to False.")
 
     cloudwatch = AwsBaseHook(
         aws_conn_id="aws_default",
--- a/catalog/dags/common/cloudwatch.py
+++ b/catalog/dags/common/cloudwatch.py
@@ -1,0 +1,64 @@
+"""
+CloudwatchWrapper extracted partially from
+https://github.com/awsdocs/aws-doc-sdk-examples/blob/54c3b82d8f9a12a862f9fcec44909829bda849af/python/example_code/cloudwatch/cloudwatch_basics.py
+"""
+
+import logging
+import boto3
+from botocore.exceptions import ClientError
+from airflow.models import Variable
+
+logger = logging.getLogger(__name__)
+
+
+class CloudWatchWrapper:
+    """Encapsulates Amazon CloudWatch functions"""
+
+    def __init__(self, cloudwatch_resource):
+        """:param cloudwatch_resource: A Boto3 CloudWatch resource."""
+        self.cloudwatch_resource = cloudwatch_resource
+
+    def enable_alarm_actions(self, alarm_name, enable):
+        """
+        Enable or disable actions on the specified alarm. Alarm actions can be
+        used to send notifications or automate responses when an alarm enters a
+        particular state.
+
+        :param alarm_name: The name of the alarm.
+        :param enable: When True, actions are enabled for the alarm. Otherwise, they
+                       disabled.
+        """
+        try:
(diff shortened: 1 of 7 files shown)


catalog/dags/common/cloudwatch.py
```diff
@@ -48,6 +49,11 @@ def enable_alarm_actions(self, alarm_name, enable):
 
 
 def enable_or_disable_alarms(enable):
+    skip_toggling = Variable.get("SKIP_TOGGLING_CLOUDWATCH_ALARMS", False)
+    if skip_toggling:
+        logger.info("Skipping toggling CloudWatch alarms.")
+        return
```
stacimc, line 55: Can we raise an `AirflowSkipException` so that the task is actually skipped, instead of just logging?

Note that we'd also need to update the `trigger_rule` for the tasks immediately downstream (`trigger_ingest_upstream` and `trigger_delete_index`, I think) to be `TriggerRule.NONE_FAILED` so they don't also get skipped.

documentation/catalog/reference/DAGs.md
```diff
@@ -85,6 +85,7 @@ The following are DAGs grouped by their primary tag:
 
 | DAG ID                                                          | Schedule Interval | Dated   | Media Type(s) |
 | --------------------------------------------------------------- | ----------------- | ------- | ------------- |
+| [`auckland_museum_workflow`](#auckland_museum_workflow)         | `@daily`          | `True`  | image         |
 | `brooklyn_museum_workflow`                                      | `@monthly`        | `False` | image         |
```
stacimc, line 89: Woah, weird. How did this end up in this PR? 😮 No worries since the change is good, just odd!

stacimc (approved):
Approving to unblock local testing of the data refreshes -- good catch @AetherUnbound, I did not think of that! I do think it would be preferable to actually skip the tasks if possible.

dhruvkb (commented):
Sorry for the drive-by review but I find the name `SKIP_TOGGLING...` a bit confusing like a multiple negative. I feel like `ENABLE_TOGGLING...` or `TOGGLE...` would be clearer. This would default to `true` so that the outcome is the same as `SKIP_TOGGLING...` that defaults to `false`.

catalog/dags/common/cloudwatch.py
krysal: Done ✅

krysal:
This is good to go now! Thanks for the suggestions; they are excellent! I am asking for a review from @AetherUnbound, as you may want to check this too.

catalog/dags/common/cloudwatch.py
```diff
@@ -0,0 +1,71 @@
...
+        resource_type="cloudwatch",
+    )
+    cw_wrapper = CloudWatchWrapper(cloudwatch.get_conn())
+
+    sensitive_alarms_list = [
+        "ES Production CPU utilization above 50%",
+    ]
+
+    for alarm in sensitive_alarms_list:
+        cw_wrapper.enable_alarm_actions(alarm, enable)
```
AetherUnbound, line 71: I know this was a holdover from when we had more than one alarm that was disabled, would you be alright with simplifying it?
```suggestion
cw_wrapper.enable_alarm_actions("ES Production CPU utilization above 50%", enable)
```

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -137,13 +138,21 @@ def create_data_refresh_task_group(
         generate_index_suffix = ingestion_server.generate_index_suffix.override(
             trigger_rule=TriggerRule.NONE_FAILED,
         )()
-        tasks.append(generate_index_suffix)
+        disable_alarms = PythonOperator(
+            task_id="disable_sensitive_cloudwatch_alarms",
+            python_callable=cloudwatch.enable_or_disable_alarms,
+            op_args=[False],
```
AetherUnbound, line 144: Even though it's not strictly necessary, we usually try to use `op_kwargs` instead to make it clearer what each argument is for:
```suggestion
            op_kwargs={
                "enable": False,
            },
```

catalog/dags/data_refresh/data_refresh_task_factory.py
```diff
@@ -181,6 +190,13 @@ def create_data_refresh_task_group(
         # running against an index that is already promoted in production.
         tasks.append(create_filtered_index)
 
+        enable_alarms = PythonOperator(
+            task_id="enable_sensitive_cloudwatch_alarms",
+            python_callable=cloudwatch.enable_or_disable_alarms,
+            op_args=[True],
```
AetherUnbound, line 196: Same note as the above:
```suggestion
            op_kwargs={
                "enable": True,
            },
```

catalog/env.template
```diff
@@ -117,3 +121,7 @@ SQLALCHEMY_SILENCE_UBER_WARNING=1
 
 AIRFLOW_VAR_AIRFLOW_RDS_ARN=unset
 AIRFLOW_VAR_AIRFLOW_RDS_SNAPSHOTS_TO_RETAIN=7
+
+# Whether to toggle production CloudWatch alarms when running a data refresh DAG.
+# Used to prevent requiring AWS credentials when running locally.
+AIRFLOW_VAR_TOGGLE_CLOUDWATCH_ALARMS=true
```
AetherUnbound, line 127: For defaults for local development (e.g. contributors), this should definitely default to false:
```suggestion
AIRFLOW_VAR_TOGGLE_CLOUDWATCH_ALARMS=false
```

In fact, this could probably be removed from the `env.template` entirely if it is put into the DAG documentation instead (like [`SLACK_MESSAGE_OVERRIDE`](https://github.com/WordPress/openverse/blob/ee77ef5fe752fa5f4f89ca6d9c8f5af7fd4816ba/catalog/dags/common/slack.py#L15)).

AetherUnbound (commented):
I've got a few small, suggested changes. Thanks for modifying this to use the idiomatic approach! 😄

catalog/env.template
krysal: From other projects, I was used to the idea of making the defaults the values used in production, but it's true that the approach of the catalog is different. Change included!

Also, given it defaults to true in the function where it's used, I think it would be need it for local development, as you said.

AetherUnbound (changes requested):
image
I tried running this locally, but it looks like if the disable alarms tasks is skipped, it skips the trigger ingest upstream step too. I think we'll need to change the trigger rule on that one as well!

krysal:
@AetherUnbound Ah, good catch! They were supposed to inherit the TriggerRule from the group task 🤔 Anyway, I changed it.

stacimc (approved):
Updates look great, I'm really excited to get this back in!

PR #3687, opened 2024-01-21
Remove needs db and add patch

### LINKED ISSUE #3436: Skip `needs_db` and always map ES hits to DB results
## Problem

Currently the serializers and request params can opt-in and out of mapping ES hits to DB results using the `needs_db` param. This was added when the DB mapping was introduced as a way to test the functionality's viability and peformance impact (although it was not used as such).

## Description

Now that our indices are leaner and contain no info beyond what's indexed and searchable the flag is almost always set to `True` and hence does not need to exist at all. Remove all occurrences of `needs_db` and replace it with a code branch that is always active.

This will simplify a lot of logic and clarify that there is only one path taken by the API from ES hits to JSON response.

ngken0995:
## Fixes

Fixes #3436 by @dhruvkb 

## Description

Remove `needs_db` because ES hit will alway map to DB result. Creating expected data by storing them in pickle file and retrieving those data in testing with open file and use patch to mock `get_db_result` return value.

## Testing Instructions

`just api/test`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/views/media_views.py
```diff
@@ -1,6 +1,5 @@
 import logging
 from typing import Union
-
```
dhruvkb, line 3: That's weird! This separator between imports was justified!

dhruvkb (changes requested):
In `test_auth.py` we're not testing for the contents of the results so for those tests, mocking `get_db_results` to return the input argument `results` as-is would've worked.

api/api/views/media_views.py
ngken0995: How can a mock retrieve the input and return input? is in this documentation related to unittest? https://docs.python.org/3/library/unittest.mock.html

dhruvkb: I meant mock a function so as to return the same argument back, something like

```python
mock_func = Mock(side_effect=lambda x: x)
```

This mock function, if called, will return its argument back.

api/test/test_auth.py
```diff
@@ -206,21 +206,25 @@ def test_unauthed_response_headers(client):
...
+        mock_get_db_result.side_effect = lambda value: value
+        time.sleep(1)
+        token = test_auth_token_exchange["access_token"]
+        query_params = {
+            "unstable__sort_by": "indexed_on",
+            "unstable__sort_dir": sort_dir,
+        }
+        res = client.get(
+            "/v1/images/", query_params, HTTP_AUTHORIZATION=f"Bearer {token}"
+        )
```
dhruvkb, line 222: Only these lines, where the view code is being executed, need to be inside the `patch` context, everything else can be at at the outer level as it was before.

dhruvkb (changes requested):
I have small change, but otherwise it looks good to me. Thanks for  the contribution and for redoing the work based on reviews @ngken0995. The PR is almost ready to go 🚀!

api/test/test_auth.py
ngken0995: Thank you for trying to push me to understand testing.

AetherUnbound (approved):
These changes look great, thanks for making them @ngken0995!

dhruvkb (approved):
LGTM! This is great.

PR #3688, opened 2024-01-22
Fix tag casing match in new tag view

### LINKED ISSUE #3495: Tag casing is not matched in new tag view
## Description

In the new tag page view, it appears that casing is ignored when clicking on a tag. For media that has tags with casing in it (e.g. `This Is An Example`), clicking on the tag will bring you to a page which doesn't have the previous result on it.

We'll likely need to modify the search logic on the API end in order to address this.

## Reproduction

1. Visit https://staging.openverse.org/preferences and enable `additional_search_views`
2. Visit https://staging.openverse.org/image/f93eb5f1-a816-4f0b-9154-a949e360b043
3. Click on the tag "Culinary equipment for serving and consuming food"
4. Observe no results

## Additional context

Related to #410

obulat:
## Fixes

Fixes #3495 by @Aetherunbound

## Description

This PR removes the `lower()` call from the `tag` view in the API. Now, the tags are matched exactly, and the case IS important. For instance, `images/tag/cat` and `images/tag/Cat` and `/images/tag/CAT` will return different results.
I updated the sample data to make it easy to add the integration test for this feature.

I completely forgot that I added the `tag.lower()` to the view, and was expecting to research Elasticsearch to fix this issue :) 

## Testing Instructions

Run `just up` and get the tag view for `"CAT"`: `http://0.0.0.0:50270/v1/images/tag/CAT/`. There should be one result with the excavator. If you search for `"cat"` (lower case), `http://0.0.0.0:50270/v1/images/tag/cat/`, the result should not contain the previous excavator result, only the 14 results with the lower-cased `"cat"` tag.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/test/media_integration.py
+++ b/api/test/media_integration.py
@@ -27,14 +27,14 @@ def search_by_category(media_path, category, fixture):
     assert all(audio_item["category"] == category for audio_item in results)
 
 
-def tag_collection(media_path):
-    response = requests.get(f"{API_URL}/v1/{media_path}/tag/cat")
+def tag_collection(media_path, tag="cat"):
+    response = requests.get(f"{API_URL}/v1/{media_path}/tag/{tag}")
     assert response.status_code == 200
 
     results = response.json()["results"]
     for r in results:
         tag_names = [tag["name"] for tag in r["tags"]]
-        assert "cat" in tag_names
+        assert tag in tag_names
 
 
 def source_collection(media_path):
--- a/api/api/views/media_views.py
+++ b/api/api/views/media_views.py
@@ -174,8 +174,7 @@ class MediaViewSet(AsyncViewSetMixin, AsyncAPIView, ReadOnlyModelViewSet):
 
     @action(detail=False, methods=["get"], url_path="tag/(?P<tag>[^/.]+)")
     def tag_collection(self, request, tag, *_, **__):
-        tag_lower = tag.lower()
-        return self.collection(request, tag_lower, None, None)
+        return self.collection(request, tag, None, None)
 
     @action(detail=False, methods=["get"], url_path="source/(?P<source>[^/.]+)")
     def source_collection(self, request, source, *_, **__):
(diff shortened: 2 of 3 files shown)


AetherUnbound (approved):
What a simple fix, glad it was easy! I tested locally and confirmed that the case matching works.

api/test/media_integration.py
```diff
@@ -27,14 +27,14 @@ def search_by_category(media_path, category, fixture):
...
-    response = requests.get(f"{API_URL}/v1/{media_path}/tag/cat")
+def tag_collection(media_path, tag="cat"):
+    response = requests.get(f"{API_URL}/v1/{media_path}/tag/{tag}")
     assert response.status_code == 200
 
     results = response.json()["results"]
     for r in results:
         tag_names = [tag["name"] for tag in r["tags"]]
-        assert "cat" in tag_names
+        assert tag in tag_names
```
sarayourfriend, line 37: I guess it isn't sensible to do `assert tag.upper() in tag_names` because both `Cat` and `cat` could be in the same list of tags :thinking: Just trying to think of how to really make the tests check "this matched only `cat` and not any other capitalisation variant". Maybe combining the tests and doing `[tag, tag.lower(), tag.upper()]` as individual queries and then asserting each one is a different result set (and that the tag as searched casing is in the list of tags)?

sarayourfriend (approved):
LGTM! Left a suggestion for a potential improvement to the test, so that it tests a bit deeper, but up to you whether you think it is is appropriate or worth it.

api/test/media_integration.py
obulat: When you consider that an item can have both a lower and an upper case tag, adding more checks seems to be overcomplicating things. We do run two tests for lower and upper case tags, and confirm that the matches are of the same casing, so I think it's enough?

PR #3719, opened 2024-01-29
Implementation Plan: Bulk moderation actions

### LINKED ISSUE #1967: Implementation Plan: Bulk moderation actions
## Description

Project proposal: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#bulk-moderation-actions-requirement-3

Refer to the project proposal section linked above for the requirements of this plan.

stacimc:
## Fixes

Fixes #1967 by @sarayourfriend 

## Description

I've chosen @obulat and @sarayourfriend as reviewers. In particular I'd like to call out to @sarayourfriend (as I do in the document itself) that this plan build on and makes some slight change to the work in #3494 that should be carefully reviewed.

This discussion is following the Openverse decision-making process. Information about this process can be found [on the Openverse documentation site](https://docs.openverse.org/meta/decision_making/index.html).

## Current round

This discussion is currently in the [Decision round](https://docs.openverse.org/meta/decision_making/process_description.html#decision-round).

The deadline for review of this round is 19 February 2024.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
+++ b/documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
@@ -1,0 +1,526 @@
+# 2024-01-22 Implementation Plan: Bulk Moderation Actions
+
+**Author**: @stacimc
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend
+- [ ] @obulat
+
+## Project links
+
+<!-- Enumerate any references to other documents/pages, including milestones and other plans -->
+
+- [Project Thread](https://github.com/WordPress/openverse/issues/383)
+- [Project Proposal](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20230411-project_proposal_content_report_moderation.html#bulk-moderation-actions-requirement-3)
+
+## Overview
+
+<!-- An overview of the implementation plan, if necessary. Save any specific steps for the section(s) below. -->
+
+Typically, moderation actions are applied to individual records in response to a
+Report (see this
+[IP](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html)
+for more details). However it may occasionally be necessary to perform a
+moderation action to a large set of records at once.
+
+For example, after receiving a large number of user reports for images from a
+particular creator, it may be beneficial to quickly mark all results from that
+creator as sensitive, or else deindex them entirely. Because these bulk actions
+necessarily affect large numbers of records without careful individual review,
+it's also necessary to be able to _undo_ these actions (for example, if further
+investigation reveals that only a subset of the records needs to be marked
+sensitive). We should also be able to view these bulk decisions as part of the
+complete moderation history of an individual record.
+
+This IP extends the work described in the
+[Django admin moderator access control and base improvements](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html)
+IP to add support for these features. I have attempted to summarize all relevant
+parts of that IP as needed here, but for full context that IP should be
+considered prerequisite reading. For brevity it is referred to as "the related
+IP" hereafter.
+
+### Models
+
+No new models need be added; rather, we will make some small changes to the
+`ModerationDecision` model described in
+[this section](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#moderationdecision)
+of the related IP. See the [Alternatives section](#alternatives) of this
+document for a discussion of why we choose this route, rather than adding new
+models.
+
+As a reminder, these are the fields from the `ModerationDecision` model which
+will not be changed:
+
+- `id`
+- `created_on`
+- `moderator_id`
+- `explanation`
+
+The following changes will be made to the `media_identifier` and `action`
+fields:
+
+#### `media_identifiers`
+
+`media_identifier` (previously a one-to-one relationship to the media the
+decision is related to), will be updated to `media_identifiers` _plural_: a
+many-to-many relationship to potentially multiple media records the decision is
+related to. Under the covers, this will be implemented by a join table. As
+pointed out in the related IP, the media identifiers must be non-constrained
+(indexed non-unique UUIDs rather than foreign keys), because deindexing media
+removes them from the API database.
+
+The relationship is many-to-many because:
+
+- A moderation decision _may_ affect many records, in the case of a bulk
+  decision
+- A record may be affected by multiple decisions, with the addition of actions
+  to undo deindexing and marking sensitive.
+
+Using the join table to tie multiple records to a single decision (rather than,
+for example, creating a separate `ModerationDecision` record for each record
+selected for bulk update), has significant cost savings and more accurately
+models the data as a single decision.
+
+#### `action`
+
+This is a slug column describing the action taken. In addition to the action
+types defined in the related IP, we will add two new options:
+
+- `reversed_deindex` - Undo a previous deindexing decision.
+- `reversed_mark_sensitive` - Undo a previous decision marking the applicable
+  records sensitive.
+
+For consistency we will also update the existing action type
+`confirmed_sensitive` to `marked_sensitive` to describe the effect of the action
+without the use of the word "confirm", which is accurate when describing single
+decisions made by moderators but potentially inappropriate for describing bulk
+actions that do not include individual review of records.
+
+## Actions
+
+The related IP describes adding new custom Django
+[admin actions](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#actions)
+for each of the action types described in that IP. These actions are called from
+the Report view and take action on a list of MediaReports to create a
+ModerationDecision. They do not need to be modified, except to use the
+many-to-many `media_identifiers` column (although in this workflow only a single
+record will ever be added.)
+
+We will need to add new actions for making moderation decisions directly from a
+list of Media, as well as two new actions for the `reversed_<x>` action types.
+
+_All_ actions go through an intermediate confirmation page which:
+
+- Indicates the number of records that will be affected by the action
+- Collects the required `explanation` text (all other columns, such as
+  `moderator_id` and `created_on`, are filled in automatically)
+
+### Bulk Mark Sensitive: mark the works as sensitive
+
+This action takes a queryset of `AbstractMedia` records. It creates a
+ModerationDecision with type `marked_sensitive` and adds all records from the
+queryset **which do not already have a related `SensitiveMedia` record**. It
+then creates and saves `SensitiveMedia` records for each of these.
+
+The confirmation page for this action should indicate the number of records that
+will actually be affected. If some records are already marked sensitive, that
+should be indicated on the confirmation page as well.
+
+### Bulk Deindex sensitive: entirely exclude the works from Openverse due to sensitivity
+
+This action takes a queryset of `AbstractMedia` records. It creates a
+ModerationDecision with type `deindexed_sensitive` and adds all records from the
+queryset. It then creates and saves `DeletedMedia` records for each of these,
+deleting the record from the API and deindexing it in elasticsearch. The
+confirmation page for this action should include a highly visible warning that
+once records are deindexed, they cannot be restored immediately.
+
+Note that it is not necessary to check that a `DeletedMedia` record does not
+already exist, because deindexing a record deletes it from the database (so it
+is not possible to call this action on already deleted records).
+
+### Bulk Deindex copyright: entirely exclude the works from Openverse due to copyright
+
+Identical to the other bulk deindex, except the action type is
+`deindexed_copyright`.
+
+### Reverse Mark Sensitive: undo marking the selected works as sensitive
+
+This action takes a queryset of `SensitiveMedia` records, rather than
+`AbstractMedia`. Each `SensitiveMedia` record has a related `media_obj`, which
+are what will be added to the ModerationDecision. It creates a
+ModerationDecision with type `reversed_mark_sensitive`, and adds the related
+media. Then it first calls `_update_es` on the `SensitiveMedia` records to
+update the elasticsearch index with the "sensitive" field for each media record,
+and finally deletes the actual `SensitiveMedia` records.
+
+Although the `SensitiveMedia` records are deleted, the ModerationDecisions
+preserve the history of these actions.
+
+### Reverse Deindex: undo deindexing the selected works
+
+This action takes a queryset of `DeletedMedia` records, rather than
+`AbstractMedia`. Like `SensitiveMedia`, each `DeletedMedia` record has a related
+`media_obj`, which will be added to the ModerationDecision. The action creates a
+ModerationDecision with type `reversed_deindex` and adds the related records.
+Then it deletes the `DeletedMedia` records.
+
+It is extremely important to note that this action does not immediately result
+in deindexed records being added back to the DB and elasticsearch indices. When
+the records were originally deindexed, they were deleted from both, and there is
+no quick way to restore them without running a data refresh. The confirmation
+page should contain a highly visible `info` block explaining this to the user.
+
+Although the `DeletedMedia` records are deleted, the ModerationDecisions
+preserve the history of these actions.
+
+### Other actions
+
+We do _not_ need to add any new action types for `rejected_reports` and
+`duplicate_reports`, which are only applicable to the Report view flow described
+in the related IP.
+
+## Views
+
+We will add one new read-only view for viewing ModerationDecisions, but most
+changes will be to existing views.
+
+### Media views
+
+We will add the custom bulk actions for `deindex_sensitive`,
+`deindex_copyright`, and `mark_sensitive` to the dropdown in the Media views.
+Bulk moderation decisions of these types will be created by taking these
+actions.
+
+However the current Media views only allow filtering records by uuid. To
+facilitate selecting records for bulk action, we'll update the `search_fields`
+for the main searchbar to search text fields (title, description, tags). We'll
+also add some additional text filters to the sidebar:
+
+- `uuid` - it may still be necessary to search by uuid, so we should preserve
+  this functionality. We include it in a separate filter (rather than keeping it
+  in `search_fields`) for search efficiency.
+- `provider`
+- `creator` - this field should have helptext reminding the user to also filter
+  by provider in order to disambiguate a specific creator.
+
+Additional or more sophisticated filters may be added in the future as
+appropriate.
+
+### DeletedMedia List View
+
+We'll add the `reversed_deindex` action to this list view, allowing maintainers
+to select a list of deindexed records and undo deindexing from this view.
+
+To facilitate easily undoing a particular bulk ModerationDecision, we will add a
+text filter for searching by ModerationDecision id. That allows you to select
+all records that were deleted as part of a particular decision.
+
+### SensitiveMedia List View
+
+We will add the `reversed_mark_sensitive` action to this list view, allowing
+maintainers to select a list of sensitive media and unmark them as sensitive
+from this view.
+
+Similar to the DeletedMedia list view, we will add a filter for
+ModerationDecision id to facilitate easily undoing an entire bulk
+ModerationDecision.
+
+### ModerationDecisions List View
+
+To make it easy to view historical ModerationDecisions, we will add one new list
+view. In the view, decisions will be sorted by creation date, and will display:
+
+- `id`
+- `action`
+- `explanation` - truncated `explanation` column
+- `record_count` - a count for the number of records affected by this decision
+
+This view is useful for getting the id of a moderation decision that you want to
+reverse. Clicking on an individual ModerationDecision will take you to a detail
+view that is entirely read-only, as a moderation decision should not be edited
+after it is created. We will add a custom filter to display only bulk moderation
+decisions, defined as a ModerationDecision with more than one media record.
+
+We will disable the option to add a ModerationDecision object directly.
+
+### Report view
+
+No changes are needed to the Report view. Note that by extending the
+ModerationDecision model rather than adding a new model, when a record is viewed
+from the Report view _all_ related moderation decisions (including bulk actions)
+will automatically be visible to moderators.
+
+## Access control
+
+Because bulk moderation decisions can affect a very large number of records,
+access should be limited to only maintainers. We will use Django's existing
+permissions support to ensure that the relevant actions are only visible to
+maintainers.
+
+If we would like to change this policy in the future to allow moderators to make
+bulk decisions, that can be added easily.
+
+## Concurrency
+
+It's possible, even likely, that reports may exist for some records that we'd
+like to take bulk action on. For example, imagine a scenario where a particular
+creator is producing large numbers of sensitive images that are not caught by
+existing filters. We may discover this situation after receiving many user
+reports, and decide to mark all of the creator's records as sensitive in bulk.
+It is therefore necessary to ensure that moderators do not, for example, deindex
+an individual record that is simultaneously being deindexed by a bulk action.
+
+To do this, we will use atomic transactions and Django's
+[select_for_update](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#select-for-update)
+to lock the records being selected for bulk action in each of the custom
+actions.
+
+## Expected Outcomes
+
+<!-- List any succinct expected products from this implementation plan. -->
+
+When this work is completed we will be able to do the following new things:
+
+- Filter media in the media views by search terms, or by provider/creator
+- Mark a selection of multiple records as sensitive, as part of a single
+  decision
+- Deindex a selection of multiple records, as part of a single decision
+- Undo marking sensitive and deindexing, for one or more records
+  - This includes undoing individual ModerationDecisions by moderators
+  - This also includes undoing a subset of a previous ModerationDecision
+
+It is equally important to identify what is out of scope for this work:
+
+Performing a bulk moderation decision will only affect records that are
+currently in the API, with no effect on records in the Catalog. For example, if
+a maintainer uses this workflow to deindex all records made by a particular
+Flickr user, this does not prevent the Flickr DAG from ingesting _new_ records
+by that same user, which will be added to the API in the next data refresh.
+Preventing new records from being added, or otherwise propagating deleted
+records to the Catalog, is out of scope for this project.
+
+## Step-by-step plan
+
+<!--
+List the ordered steps of the plan in the form of imperative-tone issue titles.
+
+The goal of this section is to give a high-level view of the order of implementation any relationships like
+blockages or other dependencies that exist between steps of the plan. Link each step to the step description
+in the following section.
+
+If special deployments are required between steps, explicitly note them here. Additionally, highlight key
+milestones like when a feature flag could be made available in a particular environment.
+-->
+
+The basic steps are:
+
+- Create the actions
+- Update the Media views with new filters and bulk actions
+- Add new bulk actions to the Media views
+- Add action and filter to the DeletedMedia view
+- Add action and filter to the SensitiveMedia view
+- Create the read-only ModerationDecision view
+
+We will also need to modify the
+[existing issue](https://github.com/WordPress/openverse/issues/3636) for
+creating the ModerationDecision model to meet the new requirements.
+
+## Step details
+
+<!--
+Describe all of the implementation steps listed in the "step-by-step plan" in detail.
+
+For each step description, ensure the heading includes an obvious reference to the step as described in the
+"step-by-step plan" section above.
+-->
+
+### Create the actions
+
+[See "Actions" section for a description of all actions](#actions)
+
+Create the custom admin actions for:
+
+- bulk deindex sensitive
+- bulk deindex copyright
+- bulk mark sensitive
+- bulk reverse mark sensitive
+- bulk reverse deindex
+
+### Update the Media views with new filters and bulk actions
+
+[See "Media views" section for details](#media-views)
+
+- Add the actions for bulk deindexing and marking sensitive to the media views,
+  available only for maintainers
+- Add the new filters for search terms, provider, and creator
+
+### Add action and filter to the DeletedMedia view
+
+[See "DeletedMedia List View" section for details](#deletedmedia-list-view)
+
+- Add the bulk reverse deindex action
+- Add a filter for ModerationDecision id
+
+### Add action and filter to the SensitiveMedia view
+
+[See "SensitiveMedia List View" section for details](#sensitivemedia-list-view)
+
+- Add the bulk reverse mark sensitive action
+- Add a filter for ModerationDecision id
+
+### Create the read-only ModerationDecision view
+
+[See "ModerationDecisions List View" section for details](#moderationdecisions-list-view)
+
+- Add the list view
+- Add read-only detail view
+
+## Dependencies
+
+### Infrastructure
+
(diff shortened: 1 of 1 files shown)


stacimc:
@sarayourfriend I'm very happy to extend the deadline for review given your schedule, please just let me know! 🙂

sarayourfriend:
5 Feb should be fine but I will let you know by Thursday if I am going to be late on this.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+
+### SensitiveMedia List View
+
+We will add the `reversed_mark_sensitive` action to this list view, allowing
+maintainers to select a list of sensitive media and unmark them as sensitive
+from this view.
+
+Similar to the DeletedMedia list view, we will add a filter for
+ModerationDecision id to facilitate easily undoing an entire bulk
+ModerationDecision.
```
sarayourfriend, line 233: I guess the importance of these two views being distinct from viewing individual records related to moderation decisions, is that this makes it easier for moderators to reverse a subset or even cross-section of decisions from one or many moderation decisions? Like from these views, a single reversal decision can be used to reverse deindexing caused by multiple different moderation decisions (e.g., a one-off and a bulk decision)?

Just trying to make sure I understand the purpose of these views as opposed to only being able to view the moderation decisions and then within the moderation decision selecting a subset of decisions to reverse.

I suppose the other advantage is that these two views only ever have _current_ decisions, rather than the moderation decisions list which will have current and historical decisions.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+- `id`
+- `action`
+- `explanation` - truncated `explanation` column
+- `record_count` - a count for the number of records affected by this decision
+
+This view is useful for getting the id of a moderation decision that you want to
+reverse. Clicking on an individual ModerationDecision will take you to a detail
+view that is entirely read-only, as a moderation decision should not be edited
+after it is created. We will add a custom filter to display only bulk moderation
+decisions, defined as a ModerationDecision with more than one media record.
```
sarayourfriend, line 249: Would the single object view show a list of the works affected? Or the deleted/sensitive media records? That list would also probably need to be actionable, if it existed. I suppose the alternative is to go here, find the moderation decision you are after, then use it to filter the deleted/sensitive media views?

I'm thinking of a workflow where someone wants to easily reverse an entire decision, rather than a subset.

Maybe the deleted/sensitive list views should also include the moderation decision ID for each line as a clickable filter for that view? It would help in this case to not have to jump between the views.

I might be over complicating things though :sweat_smile:

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+
+## Access control
+
+Because bulk moderation decisions can affect a very large number of records,
+access should be limited to only maintainers. We will use Django's existing
+permissions support to ensure that the relevant actions are only visible to
+maintainers.
+
+If we would like to change this policy in the future to allow moderators to make
+bulk decisions, that can be added easily.
```
sarayourfriend, line 268: Should we impose any limit to the max number of works in a single moderation decision? Thinking about the fact that we're more or less adding the ability to "delete all" or at least "delete a very large subset" of works (even if only temporarily until the next reindex).

Is it worth limiting bulk actions to 50 works maximum?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+
+It is equally important to identify what is out of scope for this work:
+
+Performing a bulk moderation decision will only affect records that are
+currently in the API, with no effect on records in the Catalog. For example, if
+a maintainer uses this workflow to deindex all records made by a particular
+Flickr user, this does not prevent the Flickr DAG from ingesting _new_ records
+by that same user, which will be added to the API in the next data refresh.
+Preventing new records from being added, or otherwise propagating deleted
+records to the Catalog, is out of scope for this project.
```
sarayourfriend, line 307: Ooo this is an interesting edge case that I had not considered. Excellent to call it out here as out of scope.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+
+### Having a special form for creating ModerationDecisions that affect all records by a creator
+
+I considered including the ability to add a ModerationDecision directly, with a
+modified form that would take `creator` and `provider` rather than selecting the
+individual records to add.
+
+However, this duplicates some work, and makes it much more confusing to explain
+the various ways that a decision can be made. It is also already possible to do
+this sort of bulk action by filtering in the Media views on creator/provider.
```
sarayourfriend, line 446: :+1: I like the thought process of "one obvious way to do things"! Very [zen](https://peps.python.org/pep-0020/).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+  some of them needed to be deindexed.
+- Undoing _all_ deindexing/marking sensitive for a particular creator, including
+  ones that were originally acted upon by individual reports.
+
+### A DAG for bulk decisions
+
+It's difficult to anticipate all the ways that we might need to filter records
+for selecting a set to take action on. Moreover, filtering in the Django admin
+views (especially by text search terms) may be extremely slow for some use
+cases.
```
sarayourfriend, line 469: Can we can alleviate the slow search by using Elasticsearch instead? For creator searches I think it might actually be necessary (mentioned in another comment).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+can reindex but we still need to do a data refresh afterward to get the data
+back, which may take several days.
+
+This is mitigated by tightly restricting access and adding a confirmation page
+that reiterates the number of records that are about to be affected by a
+decision, so a maintainer does not accidentally deindex much more than intended.
+We _could_ also put a limit to the number of records that can be deindexed in a
+single moderation decision as an additional precaution; however, this new bulk
+action does not add any _new_ risk compared to the existing bulk delete actions
+so I do not think this is necessary.
```
sarayourfriend, line 519: Ah, you answered my questions! Sounds good to me to not include a new limitation on the max number of records :+1:

sarayourfriend (commented):
This looks really great, and I anticipate approving without issue, provided the usage of Elasticsearch or Postgres for searching records to bulk action against (and the rationale for the decision) is clarified. My hunch is that Elasticsearch is the safer option, and may even be the necessary option, considering `creator` is not indexed in the API Postgres. However, I haven't spent a great deal of time thinking about this, so I am really just asking for clarification on which we will use and why considering the limitations of each. Postgres would be a lot less work overall, I imagine, considering how easily Django admin already interacts with it. But is the performance hit worth that? Probably that balances against how much effort it takes to use Elasticsearch instead of Postgres for that view.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
sarayourfriend: You answered this later, never mind!

https://github.com/WordPress/openverse/pull/3719#discussion_r1470452715

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+- Indicates the number of records that will be affected by the action
+- Collects the required `explanation` text (all other columns, such as
+  `moderator_id` and `created_on`, are filled in automatically)
+
+### Bulk Mark Sensitive: mark the works as sensitive
+
+This action takes a queryset of `AbstractMedia` records. It creates a
+ModerationDecision with type `marked_sensitive` and adds all records from the
+queryset **which do not already have a related `SensitiveMedia` record**. It
+then creates and saves `SensitiveMedia` records for each of these.
```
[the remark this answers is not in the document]
stacimc: Definitely, I'll add that for clarification.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+This action takes a queryset of `SensitiveMedia` records, rather than
+`AbstractMedia`. Each `SensitiveMedia` record has a related `media_obj`, which
+are what will be added to the ModerationDecision. It creates a
+ModerationDecision with type `reversed_mark_sensitive`, and adds the related
+media. Then it first calls `_update_es` on the `SensitiveMedia` records to
+update the elasticsearch index with the "sensitive" field for each media record,
+and finally deletes the actual `SensitiveMedia` records.
+
+Although the `SensitiveMedia` records are deleted, the ModerationDecisions
+preserve the history of these actions.
```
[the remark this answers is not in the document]
stacimc: 💯 That's good context to add to the IP as well, actually! Thanks :)

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+facilitate selecting records for bulk action, we'll update the `search_fields`
+for the main searchbar to search text fields (title, description, tags). We'll
+also add some additional text filters to the sidebar:
+
+- `uuid` - it may still be necessary to search by uuid, so we should preserve
+  this functionality. We include it in a separate filter (rather than keeping it
+  in `search_fields`) for search efficiency.
+- `provider`
+- `creator` - this field should have helptext reminding the user to also filter
+  by provider in order to disambiguate a specific creator.
```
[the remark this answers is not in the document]
stacimc: I think elasticsearch for sure -- there should be no reason we can't, since the text filters will have to be custom anyway. I'll add that.

Also, this makes me realize: we should probably leave the `search_fields` to just the id, and move the search terms for description/tags/title to the sidebar as well. I think that might be easier when customizing the implementation to query elasticsearch, and generally clearer. I'll update!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: Basically what you say here, yep. TL;DR but my initial plan _was_ to support reversing decisions from the ModerationDecision list view, but I ended up spending a ton of time trying to think through complexity and edge cases caused by that approach -- which are entirely eliminated by just implementing the reverse from these views. 

I talked about this a little in the Alternatives section, but some more context: it is important to track when a decision has been reversed, so that it cannot be reversed twice -- which is not so easy without complicating the model. I came up with a lot of edge cases along the lines of:
  * Decision A deindexes a set of records. Decision B undoes the deindex. Decision C deindexes a set of records, some of which were affected by Decision A. Then you try to undo Decision A again, which results in **some** of Decision _C_ being 'undone', and the history doesn't really make sense.
  * Decision A deindexes all of Creator X's records, and Decision B undoes this. Decision C deindexes this creator again at a later date, when a larger number of records by Creator X have been ingested since the last time they were deindexed. Decision D is mistakenly made as a reverse of Decision A instead of Decision C, and not all of Creator X's records get reindexed.
  * and so on 😵 

It's also more difficult to reverse subsets -- for example, if you've reversed a subset of a decision, should you be able to later reverse the remaining records? What does that look like? And it's actually impossible to reverse cross-sections, which I could easily imagine wanting when a bulk decision is prompted by a number of individual decisions by moderators for the same creator -- it would be much cleaner to undo all of those in one decision.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: I think allowing reversal from the detail view introduces a ton of complexity (see [this comment](https://github.com/WordPress/openverse/pull/3719#discussion_r1473372955)), although that was definitely something I initially tried to plan for.

But we could definitely add some convenience links. What about adding a link to the Moderation Decision detail view that will take you to the deleted/sensitive list view (whichever is relevant), filtered by the moderation decision ID?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: Agreed in your earlier comment. I was fixated on changing the `search_fields`, but I think if we add all the new filters as custom filters we can do this in elasticsearch. 🚀

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
sarayourfriend: > What about adding a link to the Moderation Decision detail view that will take you to the deleted/sensitive list view (whichever is relevant), filtered by the moderation decision ID?

Yeah, that sounds perfect!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+`deindexed_copyright`.
+
+### Reverse Mark Sensitive: undo marking the selected works as sensitive
+
+This action takes a queryset of `SensitiveMedia` records, rather than
+`AbstractMedia`. Each `SensitiveMedia` record has a related `media_obj`, which
+are what will be added to the ModerationDecision. It creates a
+ModerationDecision with type `reversed_mark_sensitive`, and adds the related
+media. Then it first calls `_update_es` on the `SensitiveMedia` records to
+update the elasticsearch index with the "sensitive" field for each media record,
```
obulat, line 160:
```suggestion
update the Elasticsearch index with the "sensitive" field for each media record,
```
Sorry for nitpicking :)

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+### Reverse Deindex: undo deindexing the selected works
+
+This action takes a queryset of `DeletedMedia` records, rather than
+`AbstractMedia`. Like `SensitiveMedia`, each `DeletedMedia` record has a related
+`media_obj`, which will be added to the ModerationDecision. The action creates a
+ModerationDecision with type `reversed_deindex` and adds the related records.
+Then it deletes the `DeletedMedia` records.
+
+It is extremely important to note that this action does not immediately result
+in deindexed records being added back to the DB and elasticsearch indices. When
```
obulat, line 175:
```suggestion
in deindexed records being added back to the DB and Elasticsearch indices. When
```

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+should be indicated on the confirmation page as well.
+
+### Bulk Deindex sensitive: entirely exclude the works from Openverse due to sensitivity
+
+This action takes a queryset of `AbstractMedia` records. It creates a
+ModerationDecision with type `deindexed_sensitive` and adds all records from the
+queryset. It then creates and saves `DeletedMedia` records for each of these,
+deleting the record from the API and deindexing it in elasticsearch. The
+confirmation page for this action should include a highly visible warning that
+once records are deindexed, they cannot be restored immediately.
```
obulat, line 142: Should we also wait some time (10 minutes?) before deindexing, since this is such a drastic change that would require a long time to reverse.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
obulat: I wonder if we could in the future add a step to check for media by deindexed creators after the data refresh, and either alert or deindex them all automatically.

obulat (commented):
I will also approve this IP :)

My main 2 questions are:
1. Do we delete the item from the main media table, or only Elasticsearch? (see more details in the inline comment)
2. How will we proceed with updating the related IP? Will we open another issue for it?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,526 @@
...
+The confirmation page for this action should indicate the number of records that
+will actually be affected. If some records are already marked sensitive, that
+should be indicated on the confirmation page as well.
+
+### Bulk Deindex sensitive: entirely exclude the works from Openverse due to sensitivity
+
+This action takes a queryset of `AbstractMedia` records. It creates a
+ModerationDecision with type `deindexed_sensitive` and adds all records from the
+queryset. It then creates and saves `DeletedMedia` records for each of these,
+deleting the record from the API and deindexing it in elasticsearch. The
```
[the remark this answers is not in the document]
stacimc: Currently when an individual record is deindexed, we do both (deleting the record from the API DB's main media table _and_ deindexing in Elasticsearch), so the intention was to apply consistent behavior to bulk actions.

This is an interesting idea, though. When a data refresh runs, records that have an associated DeletedMedia record are not copied over into the new media table or the index. So if we update the admin `deindex` logic to _only_ remove records from Elasticsearch, the record in the API media table will be retained _until the next data refresh_. It still gets removed from the table eventually, but we would have a built in "safety window" where it would be easier for us to reindex quickly, like you say 🤔 

We would:
* Update the `deindex` logic to only remove the record from Elasticsearch
* Make no changes to the data refresh logic, which will result in media with a `DeletedMedia` record being removed from the API media table on the next data refresh
* Update the `reindex` logic to check to see if the record being reindexed exists in the API media table, and if so we can immediately reindex it in elasticsearch

This would apply for bulk and individual actions. The cons I see are:
* The effect of a `reindex` action differs depending on when it is run, which may be confusing.
* A bulk `reindex` might be able to reindex _some_ but not _all_ of the selected records, if it affects records which were initially deindexed at different times :/
* The media tables and indices would become technically out of sync temporarily (and potentially for more than a week at a time if there are issues with the data refresh)

This is tough, bulk actions in particular reveal a lot of complexity.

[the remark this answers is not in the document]
stacimc: I had considered -- and @sarayourfriend also suggested :) -- putting a limit on the number of records that can be deindexed in any single decision, to mitigate risk in a much simpler way. I decided against it mainly because it is already possible to do large bulk `Delete` actions on media records, so I don't really perceive this as a _new_ risk.

Another thought is that once the data refresh is moved entirely into Airflow, I think it would be easy to add DAG conf options for a `SELECT` query that would allow you to run a "mini" data refresh updating only a subset of records more quickly (although I don't think we'll get around to that very soon).

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: Definitely! I almost added some ideas to this IP, but it got complicated really quickly 😅 It required adding provider/creator as nullable fields on the ModerationDecision, and added a ton of complexity to the workflows for creating bulk decisions. That also raises questions like: Do we create DeletedMedia/SensitiveMedia records for the new records, or just avoid copying them over in the data refresh? If the former, do we add those to an existing ModerationDecision or create a new one? 🤔 

Ultimately I think it might be best to have some sort of table for tracking creators that should be deindexed and actually prevent ingestion in the catalog. We could have something similar for creators that should be marked sensitive.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: When I initially responded I did not see Sara's additional comment in this thread, just wanted to acknowledge that I agree with this reasoning.

Also wanted to add another mention of my [earlier comment](https://github.com/WordPress/openverse/pull/3719/files/e14e2465ba6a40bd9b1088cdbfa6f914aa2067f4#r1484704393) -- in the future I think we could implement support for mini/targeted data refreshes that would make this even quicker/easier.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
stacimc: Not at all, thank you! :)

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
sarayourfriend: I have to say this is alot of complexity for what is currently a non-issue (quickly re-adding inappropriately deindexed data). It really only _becomes_ an issue if (a) someone takes a malicious action (which is too broad to try to prevent everything, at a certain level of access) or (b) someone mistakingly deindexes a disasterous number of documents.

Both of these are _exceptional_ circumstances. The easiest solution is to kick off an early data refresh, or accept that we will have _at most_ 7 days of bad data.

I really don't think we need contingencies beyond that, especially if they add so much complexity to an already complex set of issues. Adding a limit on the number of records deleted is much less complex and would be my preference for the _first_ thing we did _if_ someone accidentally deindexed a large chunk of works and for some reason a proactive data refresh proved not to be a suitable remedy.

Everything else feels excessively protectionist about data that is entirely recoverable. There's no risk of permanent loss, which makes me recoil from significant complexity to prevent temporarily deminished data caused by exceptional circumstances.

sarayourfriend:
> whether to https://github.com/WordPress/openverse/pull/3719#discussion_r1484700003 to only remove records from Elasticsearch, to give a 'safety window' for reindexing

To clarify my answer from the comment thread, my vote is a strong "no", it is not worth making a change for this. My reasons are in the thread.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
stacimc: Very strongly in agreement! I wanted to double check on this last option since as described it would be easy to _implement_ unlike some of the other ideas -- but it adds _conceptual_ complexity and edge cases that I personally did not like. It sounds like Olga is also okay with this so I'm going to go ahead and make revisions now.

stacimc:
Re-requesting reviews as this has entered the Decision Round :)

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
sarayourfriend: I'm okay making _no changes_ to prevent malicious or accidental deindexation of massive amounts of data until we have a reason to think it is not something from which we can easily recover within a week.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,531 @@
...
+**Author**: @stacimc
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend
```
sarayourfriend, line 12:
```suggestion
- [x] @sarayourfriend
```

sarayourfriend (approved):
LGTM! Nice work :+1:

dhruvkb (commented):
While I'm not an assigned reviewer, I am still sharing some comments that came up during working on #3760. I'll add more comments if something comes up later, but these are the thoughts I have so far.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,531 @@
...
+selected for bulk update), has significant cost savings and more accurately
+models the data as a single decision.
+
+#### `action`
+
+This is a slug column describing the action taken. In addition to the action
+types defined in the related IP, we will add two new options:
+
+- `reversed_deindex` - Undo a previous deindexing decision.
+- `reversed_mark_sensitive` - Undo a previous decision marking the applicable
```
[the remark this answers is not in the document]
stacimc: In the case of deindexing, the separation of the `deindexed_sensitive` and `deindexed_copyright` actions serve to provide more context for why the records were deindexed. I don't think that context makes sense in the case of reindexing.

Moreover it complicates the action considerably. Currently it is possible to reindex any selection of `DeletedMedia`. If we have two reindex actions, should we only allow a `reversed_deindex_sensitive` action to be taken on DeletedMedia that were previously `deindexed_sensitive`? How would we implement that in the UI? But if we don't make that restriction, the suffix doesn't make sense.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,531 @@
...
+  records sensitive.
+
+For consistency we will also update the existing action type
+`confirmed_sensitive` to `marked_sensitive` to describe the effect of the action
+without the use of the word "confirm", which is accurate when describing single
+decisions made by moderators but potentially inappropriate for describing bulk
+actions that do not include individual review of records.
+
```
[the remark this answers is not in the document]
stacimc: I agree; that also allows us to change the tense to `deduplicated_reports` to match the other action types!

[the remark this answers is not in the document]
stacimc: Done in [12a9be8](https://github.com/WordPress/openverse/pull/3719/commits/12a9be8f33f5d23a3a0ebf448a08ac7c4603b969)!

sarayourfriend:
@obulat are you able to review the implementation plan soon? It would be great to get this approved soon (with any changes needed, of course, if you have blockers to share) so that we can update the issues from the previous implementation plan with the new ModerationDecision model and other modifications to the existing plan.

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
dhruvkb: Fair point, thanks for clarifying!

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,531 @@
...
+The following changes will be made to the `media_identifier` and `action`
+fields:
+
+#### `media_identifiers`
+
+`media_identifier` (previously a one-to-one relationship to the media the
+decision is related to), will be updated to `media_identifiers` _plural_: a
+many-to-many relationship to potentially multiple media records the decision is
+related to. Under the covers, this will be implemented by a join table. As
```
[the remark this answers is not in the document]
dhruvkb: Ah yes, that makes sense. I had not thought about the impact this would have on querying. But wouldn't DB-writes be incredibly slow if we have to write individual records for each media item in a bulk-moderation event?

stacimc:
I've just realized (prompted by @dhruvkb 's comment above) that I may need to clarify the specifics of the implementation for the many to many relationship. I likely do not have time to do so today so I'm going to draft this briefly.

sarayourfriend:
@obulat Can you confirm you've seen the request to review this? Will you confirm if you will approve with the change Staci is making or if you'll require other changes?

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
stacimc: I don't believe so -- we can use [RelatedManager.set](https://docs.djangoproject.com/en/5.0/ref/models/relations/#django.db.models.fields.related.RelatedManager.set) to bulk set the list of related records, if I'm understanding your question!

stacimc:
I was [briefly concerned](https://github.com/WordPress/openverse/pull/3719#issuecomment-1960228526) that there would be an issue with using the `ManyToManyField` as designed, due to the requirement that the record ids be non-constrained. TL;DR I had forgotten that this is possible using the `db_constrained` argument as we already do elsewhere in the code (for example, in the implementation of the MediaReports). I thought it was worth adding a specific mention of that argument to the IP and will also do so in the issues when they are updated/created. This should not require more work and I am undrafting now, apologies -- better to be safe 😅

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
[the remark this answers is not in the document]
obulat: I hadn't considered that some records could have been removed previously and not be in the API table. I agree with the reasoning that this is an exceptionally rare (and possibly malicious) case 👍

documentation/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.md
```diff
@@ -0,0 +1,535 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [x] @sarayourfriend
+- [ ] @obulat
```
obulat, line 13:
```suggestion
- [x] @obulat
```

obulat (approved):
The plan is solid and very detailed as it considers all of the edge cases, and also updates the related plan.
Sorry for the delay

stacimc:
I've created issues under [this milestone](https://github.com/WordPress/openverse/milestone/24), and updated the relevant issue from the related IP.

PR #3721, opened 2024-01-30
Centralise frontend network error reporting

### LINKED ISSUE #3468: Avoid `AxiosError` when requesting bad image links
## Description

There is a recurrent error of Axios trying to request an image with a wrong uuid. It has 112 occurrences since first seen on Nov 20, 2023.

```
GET /image/f51977d1-081a-4e9f-afe1-
```

It raises an error with the message: `Request failed with status code 404` .

It is unknown to me what or who stumbles upon this URL, but it consumes part of our Sentry quota, so it should be prevented somehow from triggering a notification.

## Reproduction

Reproduction steps are unknown. 

## Sentry details
https://openverse.sentry.io/share/issue/20761d85362b468190131dcda0463e9f/

sarayourfriend:
## Fixes

Fixes #3468 by @krysal

## Description

I've kept this PR as a draft because:
- it is just a _proposal_ for an idea for how to solve this problem, but is not necessarily the simplest (but I think it's worth it)
- it still needs additional testing and probably some small refactoring, but I don't want to invest time into that until someone else has looked at this and given the general approach/idea a :+1: 

## Testing Instructions

Force an error to happen in a network request on the frontend. For example, append a bogus string to the end of media IDs so that they result in 404s. Confirm that those don't get sent to Plausible or Sentry anymore (try it on `main` as well to see that it goes to Sentry).

I don't know how to force a network error. You may be able to do it using your browser's network debugging tab and explicitly blocking a request.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/composables/use-search.ts
+++ b/frontend/src/composables/use-search.ts
@@ -6,14 +6,13 @@ import { useMediaStore } from "~/stores/media"
 
 import { useI18nResultsCount } from "~/composables/use-i18n-utilities"
 import { useMatchSearchRoutes } from "~/composables/use-match-routes"
-import type { EventName, Events } from "~/types/analytics"
+import { useAnalytics } from "~/composables/use-analytics"
 
-export const useSearch = (
-  sendCustomEvent: <T extends EventName>(name: T, payload: Events[T]) => void
-) => {
+export const useSearch = () => {
   const mediaStore = useMediaStore()
   const searchStore = useSearchStore()
   const router = useRouter()
+  const { sendCustomEvent } = useAnalytics()
 
   const { matches: isSearchRoute } = useMatchSearchRoutes()
 
--- a/frontend/src/composables/use-search.ts
+++ b/frontend/src/composables/use-search.ts
@@ -8,11 +8,12 @@ import { useI18nResultsCount } from "~/composables/use-i18n-utilities"
 import { useMatchSearchRoutes } from "~/composables/use-match-routes"
 import { useAnalytics } from "~/composables/use-analytics"
 
-export const useSearch = () => {
+export const useSearch = (
+  sendCustomEvent: ReturnType<typeof useAnalytics>["sendCustomEvent"]
+) => {
   const mediaStore = useMediaStore()
   const searchStore = useSearchStore()
   const router = useRouter()
-  const { sendCustomEvent } = useAnalytics()
 
   const { matches: isSearchRoute } = useMatchSearchRoutes()
 
--- a/frontend/src/plugins/errors.ts
+++ b/frontend/src/plugins/errors.ts
@@ -1,0 +1,144 @@
+import axios from "axios"
+import { Plugin, Context } from "@nuxt/types"
+
+import { ERR_UNKNOWN, ErrorCode, errorCodes } from "~/constants/errors"
+import type { FetchingError, RequestKind } from "~/types/fetch-state"
+import type { SupportedSearchType } from "~/constants/media"
+
+import { useAnalytics } from "~/composables/use-analytics"
+
+const isValidErrorCode = (
+  code: string | undefined | null
+): code is ErrorCode => {
+  if (!code) {
+    return false
+  }
+  return (errorCodes as readonly string[]).includes(code)
+}
+
+function isDetailedResponseData(data: unknown): data is { detail: string } {
+  return !!data && typeof data === "object" && "detail" in data
+}
+
+/**
+ * Normalize any error occurring during a network call.
+ *
+ * @param error - Any error arising during a network call
+ * @param searchType - The type of search selected when the error occurred
+ * @param requestKind - The kind of request the error occurred for
+ * @param details - Any additional details to attach to the error
+ * @returns Normalized error object
+ */
+function normalizeFetchingError(
+  error: unknown,
+  searchType: SupportedSearchType,
+  requestKind: RequestKind,
+  details?: Record<string, string>
+): FetchingError {
+  const fetchingError: FetchingError = {
+    requestKind,
+    details,
+    searchType,
+    code: ERR_UNKNOWN,
+  }
+
+  if (!axios.isAxiosError(error)) {
+    fetchingError.message = (error as Error).message
+    return fetchingError
+  }
+
+  // Otherwise, it's an AxiosError
+  if (isValidErrorCode(error.code)) {
+    fetchingError.code = error.code
+  }
+
+  if (error.response?.status) {
+    fetchingError.statusCode = error.response.status
+  }
+
+  const responseData = error?.response?.data
+
+  // Use the message returned by the API.
+  if (isDetailedResponseData(responseData)) {
+    fetchingError.message = responseData.detail as string
+  } else {
+    fetchingError.message = error.message
+  }
+
+  return fetchingError
+}
+
+/**
+ * Record network errors using the appropriate tool, as needed,
+ * based on response code, status, and request kind.
+ * @param error - The error to record
+ */
+function recordError(
+  context: Context,
+  originalError: unknown,
+  fetchingError: FetchingError
+) {
+  if (fetchingError.code === "ERR_NETWORK") {
+    /**
+     * Record network errors in Plausible so that we can evaluate potential
+     * regional or device configuration issues, for which Sentry is not
+     * as good a tool. Additionally, the number of these events are trivial
+     * for Plausible, but do actually affect our Sentry quota enough that it
+     * is worth diverting them.
+     */
+    const { sendCustomEvent } = useAnalytics()
+    sendCustomEvent("NETWORK_ERROR", {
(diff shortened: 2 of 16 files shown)


sarayourfriend:
@krysal interesting in your view on this, in particular, if you have any opinions about this approach.

The simpler alternative I can think of was to either copy/paste the conditionals of error handling to the appropriate areas, or to create a similar unified fetch error processing function that just takes `context` as an argument, rather than creating a plugin. The plugin feels like a convenient approach to me, but it could cause additional pains with Nuxt 3, I'm not sure.

Curious for your thoughts regardless!

AetherUnbound (commented):
I think this approach makes sense! I like that we can differentiate the tool the error is sent to for certain kinds of failures.

frontend/src/composables/use-search.ts
```diff
@@ -6,14 +6,13 @@ import { useMediaStore } from "~/stores/media"
...
+import { useAnalytics } from "~/composables/use-analytics"
 
-export const useSearch = (
-  sendCustomEvent: <T extends EventName>(name: T, payload: Events[T]) => void
-) => {
+export const useSearch = () => {
   const mediaStore = useMediaStore()
   const searchStore = useSearchStore()
   const router = useRouter()
+  const { sendCustomEvent } = useAnalytics()
```
obulat, line 15: Have you checked that this always works? 

The reason that `sendCustomEvent` was passed as a parameter here was that you cannot use a composable that requires Nuxt instance outside of setup. There is an edge case `useSearch` is called, if I remember correctly, in `asyncData`, and Nuxt instance is not available, so a call to `useAnalytics` inside `useSearch` produces a warning.

obulat (commented):
I love centralized recording of errors, and separation of the actionable ones from the non-actionable, @sarayourfriend!
I was thinking whether we should create an Error class for our custom fetchingError instead of an object, in case we need to throw this error somewhere (based on some work in the Nuxt 3 branch). What do you think?

frontend/src/plugins/errors.ts
```diff
@@ -0,0 +1,144 @@
...
+     * regional or device configuration issues, for which Sentry is not
+     * as good a tool. Additionally, the number of these events are trivial
+     * for Plausible, but do actually affect our Sentry quota enough that it
+     * is worth diverting them.
+     */
+    const { sendCustomEvent } = useAnalytics()
+    sendCustomEvent("NETWORK_ERROR", {
+      requestKind: fetchingError.requestKind,
+      searchType: fetchingError.searchType,
+    })
```
krysal, line 93: Although we won't lose information by sending them only to Plausible? Will we know which page, for example, the error occurs on?

krysal (commented):
I love the centralization of error management, too! 💯  I think the plugin approach is right; not so sure about the high coupling with Axios since it may go away with Nuxt 3, but it is what it is right now. If we also move the API calls to a plugin, I think the error management will probably merge with it.

frontend/src/plugins/errors.ts
sarayourfriend: Yes, all of that is automatically recorded for all Plausible events: https://plausible.io/docs/stats-api#properties

All plausible "goal" events (custom events) are essentially just an extension on a regular page view event, with optional additional information on top of the page view props.

sarayourfriend:
> should create an Error class for our custom fetchingError instead of an object, in case we need to throw this error somewhere

In JS you can throw anything. We can wrap it in a class, no problem. But is the idea to be able to check it with `instanceof`? We can do this in a separate issue as well, now it will be easy to switch to a different format for the error if needed.

frontend/src/composables/use-search.ts
sarayourfriend: I see. But then should `useSearch` return a product based on calling `useAnalytics`? I'll revert this change, but that to me sounds like a bit of code smell, if the composable isn't able to call other composables at the top level. Maybe I am misunderstanding the edge case. Either way, I'll remove this from the PR, it definitely doesn't need to be here.

sarayourfriend:
Thanks for the preliminary reviews, Krystle and Olga. I'll fix up this PR and get it ready for a final review.

obulat (approved):
I tried this locally, and sent the following error to Sentry:
Screenshot 2024-02-08 at 11 37 49 AM

Do you know where Sentry gets the "recordError" title from, and whether we can change it?
I also left 2 non-blocking comments inline.

frontend/test/unit/specs/plugins/errors.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { Context } from "@nuxt/types"
```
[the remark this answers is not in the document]
sarayourfriend: I did start with that, but the function the plugin puts into `$processFetchingError` still requires `$sentry` and `useAnalytics`. It also would require further exporting things from the plugin module that aren't really needed anywhere other to make them testable, and would be testing such small units that it would kind of miss the bigger picture (is how I felt).

Happy to change this if you see a way to make the tests much simpler, and not require mocking `$sentry` and `useAnalytics` still, but if they still require that, then the only change is one that also additionally tests that the plugin is configured correctly (calls `inject` with the right parameters).

frontend/src/plugins/errors.ts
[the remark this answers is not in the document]
sarayourfriend: Good question! Those we _can_ see in Cloudflare, so I don't know how much value we get from sending them to Plausible, but I agree that we can skip them in Sentry :+1:

sarayourfriend:
> Do you know where Sentry gets the "recordError" title from, and whether we can change it?

I suppose from the function name, but I don't know how to change it. Maybe in breadcrumbs? I'll have to look at the sentry JS docs, but it's a good one to change, absolutely.

frontend/test/unit/specs/plugins/errors.spec.ts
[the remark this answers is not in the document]
obulat: Not directly related to this comment, but after spending some time this morning investigating the problems with "search time" event, I realized that this change would also require the changes that would make `useAnalytics` usable on the server (see #3775)

obulat (changes requested):
I'm requesting changes because I realized that our plausible tracker breaks on the server (it's basically only loaded on the client side). We would need to make sure that the server-side error reporting does not cause new errors. And related to that, we should probably add e2e tests for errors because the unit tests do not have server/client distinction.

sarayourfriend:
Yes! I actually have staged changes for that I forgot to push, thanks for the reminder!

sarayourfriend:
I remember what I got stuck on, with Playwright. I don't know how to trigger a _network error_ as Axios identifies them. Maybe using `page.route` and `route.abort` will work, I will try it. Just wanted to mention in case you have ideas before I'm able to get back to this.

frontend/src/plugins/errors.ts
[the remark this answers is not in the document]
obulat: Then it's even better, we can completely ignore 429s here :)

frontend/src/plugins/errors.ts
```diff
@@ -0,0 +1,144 @@
...
+    fetchingError.message = error.message
+  }
+
+  return fetchingError
+}
+
+/**
+ * Record network errors using the appropriate tool, as needed,
+ * based on response code, status, and request kind.
+ * @param error - The error to record
```
obulat, line 74: WebStorm is complaining that JSDoc does not describe the parameters correctly (`error` instead of `context`, `originalError` and `fetchingError`.

obulat:
> I remember what I got stuck on, with Playwright. I don't know how to trigger a _network error_ as Axios identifies them. Maybe using `page.route` and `route.abort` will work, I will try it. Just wanted to mention in case you have ideas before I'm able to get back to this, @obulat

You can use Playwright's `route.abort` like this to trigger a network error:
```
await page.route(new RegExp(`v1/(images|audio)/`), async (route) => {
      route.abort("internetdisconnected")
    })
```
I tried adding e2e error tests to this PR, and found that `useAnalytics` does not work when we need to call it from the store and not from the `setup` function. The context is not passed to this composable, even if the store's `fetchMedia` function is called from within the `setup` function. The context is not available, and the `onMounted` also throws an error. We could move the `onMounted` to sync analytics to the layout files to prevent it from running for every mounted component that uses it. However, the solution for the `context` not being available is more complex. Maybe we should move `sendEvent` to a plugin?

sarayourfriend:
> Maybe we should move sendEvent to a plugin?

I was thinking about this as well. There are a few composables like this that might be better suited for plugins instead of composables.

sarayourfriend:
This PR doesn't work due to the analytics composable, so I'm going to close it for now, but leave the branch up. I'll open another PR to move analytics composable into a plugin.

PR #3728, opened 2024-01-31
Add storybook snapshot update command to testing documentation

AetherUnbound:
## Description

This came up in https://github.com/WordPress/openverse/pull/3727#issuecomment-1919615370.

This PR adds the storybook snapshot update command to the frontend testing documentation, so folks updating aspects of the frontend have all the necessary commands readily available in the same document.

I also removed all the preceding `$` characters from these commands. Because they're of the `console` type (or perhaps just a code block in general), Sphinx makes it easy to copy the commands from the browser and paste into a shell. However, those commands come with the leading `$` character too which makes them hard to paste and run immediately. This PR removes them all for this document, I'll have a follow up PR that removes them everywhere else. 

## Testing Instructions

View the generated documentation output and verify it looks correct.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
[...]

PR #3730, opened 2024-01-31
Remove preceding `$` in documentation code blocks, use bash for code blocks, add pre-commit check for console usage

AetherUnbound:
## Description

This PR fixes an issue identified in #3728 wherein `console` blocks wouldn't use the right syntax highlighting in our documentation unless a preceding `$` was added, but the additional character then made it difficult to copy and paste commands directly into a terminal.

I've corrected all of the other instances of the `console` syntax selection, removed the preceding `$` characters, and added a pre-commit check for the ~~`console` block to ensure that `bash` is used instead. The pre-commit hook is a little janky 😅 I don't love it but it does work 🤷‍♀️~~ The pre-commit hook I added now auto-corrects uses of `console` to `bash`! I tried to add a hook for checking for `$` lines as well, but the bash escaping and multi-line grep that would be required were way too complicated so I just gave up.

## Testing Instructions

CI should pass! Check out locally and run `just lint` to observe the new check. You can also change a `bash` block back to `console` and run linting to see that it fails.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,15 +11,15 @@ repos:
         entry: bash -c 'if [ ! -f "$(dirname "$dir")"/frontend/src/locales/ar.json ]; then just frontend/run i18n; fi'
         language: system
         pass_filenames: false
-      - id: check-no-console-code-block
-        name: Check that the "console" block is not used in docs
+      - id: fix-console-code-block
+        name: Ensure "bash" code block is used over "console"
+        files: ^documentation/.*\.md$
         # Check that we are not using the ```console block in our documentation
         # -R recursively checks files
         # -l suppresses normal input and only lists files with errors
         # Print an error message if any files are found
-        entry: bash -c 'grep -Rl --include="*.md" '\''```console'\'' documentation/ && { printf "\n[ERROR] The \"bash\" syntax selection in the above files should be used in place of \"console\" in code blocks"; exit 1; } || exit 0'
+        entry: sed -i -e 's/```console/```bash/g'
         language: system
-        pass_filenames: false
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,15 @@ repos:
         entry: bash -c 'if [ ! -f "$(dirname "$dir")"/frontend/src/locales/ar.json ]; then just frontend/run i18n; fi'
         language: system
         pass_filenames: false
+      - id: check-no-console-code-block
+        name: Check that the "console" block is not used in docs
+        # Check that we are not using the ```console block in our documentation
+        # -R recursively checks files
+        # -l suppresses normal input and only lists files with errors
+        # Print an error message if any files are found
+        entry: bash -c 'grep -Rl --include="*.md" '\''```console'\'' documentation/ && { printf "\n[ERROR] The \"bash\" syntax selection in the above files should be used in place of \"console\" in code blocks"; exit 1; } || exit 0'
+        language: system
+        pass_filenames: false
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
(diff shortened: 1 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -11,6 +11,15 @@ repos:
         entry: bash -c 'if [ ! -f "$(dirname "$dir")"/frontend/src/locales/ar.json ]; then just frontend/run i18n; fi'
         language: system
         pass_filenames: false
+      - id: check-no-console-code-block
+        name: Check that the "console" block is not used in docs
+        # Check that we are not using the ```console block in our documentation
+        # -R recursively checks files
+        # -l suppresses normal input and only lists files with errors
+        # Print an error message if any files are found
+        entry: bash -c 'grep -Rl --include="*.md" '\''```console'\'' documentation/ && { printf "\n[ERROR] The \"bash\" syntax selection in the above files should be used in place of \"console\" in code blocks"; exit 1; } || exit 0'
```
dhruvkb, line 20: Can I haz autofix to change ```` ```bash ```` to ```` ```console ````?

dhruvkb, line 20: `--include` can be replaced with the `files` param on the hook to only accept Markdown. That way the hook won't even run unless a Markdown file has changed.

dhruvkb (approved):
LGTM! It's awesome that you made a custom hook. 

I would love it more if it did autofix and if it accepted a specific set of files that had changed, but it's super fast so I won't hold anything against it.

sarayourfriend:
> but the additional character then made it difficult to copy and paste commands directly into a terminal.

I somewhat think this is a feature, and more or less the purpose for pre-pending lines with $, as I personally like to discourage a mindless copy/paste of terminal commands.

That said, it's my preference, and consistency and working syntax highlighting is nice, so I'm not blocking or requesting a change, just giving one reason why it might be a benefit for copy/paste not to be a one-click operation. The terminal is a powerful thing and I worry about not reading the thing someone told me to put into it :slightly_smiling_face:

AetherUnbound:
I appreciate your thoughts @sarayourfriend! This came up because I was trying to follow a guide for updating tests, and I wanted to use the commands the guide specified but could not without removing that first character on each command. A lot of the files changed are `quickstart.md` files, intended for new contributors. If my own experience with using our documentation was frustrating, I imagine that frustration might be larger for new contributors too who are navigating an entirely new project. I do think that reducing friction in those cases is beneficial. To me, being able to copy and paste commands that just work is more useful than having to avoid the built-in copy button that Sphinx provides or drag my mouse over selections to exclude the `$` character. It's a small thing, but if it gets contributors off the ground faster I feel like that's a win 🙂

AetherUnbound:
Oh I should also mention another thing I noticed: removing the preceding `$` in those code blocks also helped my IDE identify that they were runable blocks! So it now provides an arrow next to those segments of the documentation which allow me to run the lines directly in my IDE, which is neat 🚀

.pre-commit-config.yaml
AetherUnbound: Thanks for the encouragement @dhruvkb! ChatGPT had given me a very wrong impression of things when helping me draft this hook 😑 the new auto-fix version is _way_ simpler 🎉

sarayourfriend:
Yup! I think there are pros and cons, your decision makes sense for this project. If we ever have sensitive stuff we're telling people to run (like `sudo whatever`) then it'd be good to make that _not_ copyable, but we'll deal with that if we get there.

sarayourfriend (approved):
Cool!

I was going to try to come up with the regex that Dhruv suggested for the auto-fix to remove the leading `$/#` but I guess it's kinda complex!

I actually think that a Vale style would be _perfect_ for this, now that I think about it! I'll try and cook something up.

AetherUnbound:
> I actually think that a Vale style would be _perfect_ for this, now that I think about it! I'll try and cook something up.

Ooo, that'd be really neat! And a great example for other possibilities that might come up in the future 😄

sarayourfriend:
I tried writing one up but it's hard to match, because Vale doesn't considering code blocks as single lines, making it difficult to find the "start" of a line :confused: Oh well! It might be possible with a prettier plugin but those are pretty hard to get going!

PR #3755, opened 2024-02-06
Separate release drafting from publishing

### LINKED ISSUE #2264: Change release process to facilitate release notes previews
## Problem

The current release process makes it extremely difficult to know whether a particular app actually has changes to release. This means we're liable to run the API or frontend release workflow (which we try to do weekly, if there are changes) when there are no relevant changes, resulting in an empty changelog.

## Description

To fix this, we need to change the way we use release drafter and the way we trigger releases for particular apps. Rather than using the Release app workflow, we'll configure release drafter to actively draft releases for all our apps whenever changes appear for them. We can use the exact same release drafter configurations we currently have adapted to the old "draft release" workflow. The only caveat is that we need to provide release drafter with the version string because it only supports semver, which we do not use. When these individual releases are published, that will trigger the docker tag creation, the changelog PR, and the production deployment (for frontend and API). The app and datever tag need to be parsed from the release name.

## Solution

Create a new workflow called "Draft releases" with a matrix for each of the four apps that can be released (api, ingestion server, frontend, catalog). This workflow should run on pushes to main. Create the date tag and then for each app, run release drafter with the current configurations for each app.

Create a new workflow called "Release" that runs on release publish. This workflow must parse the app and datever string from the release name, tag the appropriate docker images with the release tag (as Release app does now), create the changelog PR based on the published release's text, and then trigger the API or frontend production deployments when applicable.

Afterwards, delete the release app workflow from the GitHub actions UI to remove clutter.

sarayourfriend:
## Fixes

Fixes #2264 by @sarayourfriend 

## Description

Splits release _drafting_ into its own workflow that runs on every push to main.

Updates the release-app to have a single "release" trigger, to always use the latest staging image to deploy production. It derives which app to release based on the released tag.

As the trigger is the release being published, there's nothing we need to do with the release in the `release-app` workflow, everything is now exclusively about (a) getting the images tagged, (b) deploying them (if API or frontend), and (c) opening the changelog PR.

## Testing Instructions

This is annoyingly hard to test. I've tried to write the release drafting workflow in such a way that it will get tested, but I don't know if it will work. There is no way to really test the release app workflow changes without merging and trying it on main, and then fixing things as we go.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/api/guides/deploy.md
+++ b/documentation/api/guides/deploy.md
@@ -4,17 +4,10 @@
 For more information on how deployments work, please see the [general deployment guide](/general/deployment.md).
 ```
 
-1. Visit
-   [https://api-staging.openverse.engineering/version](https://api-staging.openverse.engineering/version)
-   and
-   [the API Docker image](https://github.com/wordpress/openverse/pkgs/container/openverse-api).
-   Verify that the commit SHA live on the staging site is also tagged with
-   `latest` in the Docker image.
-   ![GitHub package directory screenshot](/_static/package_directory_example.png)
-1. Release the app via
-   [GitHub workflow](https://github.com/WordPress/openverse/actions/workflows/release-app.yml).
-   Click the "Run workflow" button, choose "api" from the dropdown, and supply
-   the SHA identified in step 1.
+1. [Publish the drafted API release in the GitHub release page of the monorepo](https://github.com/WordPress/openverse/releases?q=api-)
+   - Here you can preview the changes included in the API release and decide
+     whether a release is necessary and adjust monitoring during the deployment
+     accordingly.
 1. That's it! The API will be deployed. You can monitor the deployment in the
    maintainers `#openverse-notifications` channel and in the
    [infrastructure repository's workflow listing](https://github.com/WordPress/openverse-infrastructure/actions).
--- a/.github/workflows/draft_releases.yml
+++ b/.github/workflows/draft_releases.yml
@@ -29,13 +29,28 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Calculate tag name
+        id: tag
+        run: |
+          # Format example: 2023.03.22.04.56.29
+          # `-u` forces UTC
+          formatted_date="$(date -u +%Y.%m.%d.%H.%M.%S)"
+
+          {
+            echo "date=$formatted_date";
+            echo "git-tag=${{ env.APP_NAME }}-$formatted_date";
+          } >> "$GITHUB_OUTPUT"
+
+      # Each time this runs, it should replace the previous drafted release matching the tag prefix set in the release drafter config
+      # This means there will always be a running drafted release for all apps
+      # Publishing the drafted release triggers the `release-app` workflow
       - uses: release-drafter/release-drafter@v5
         id: release-drafter
         with:
           config-name: release-drafter-${{ matrix.app }}.yml
-          version: ${{ matrix.app }}-draft
-          tag: ${{ matrix.app }}-draft
-          name: ${{ matrix.app }}-draft
+          version: ${{ steps.tag.outputs.date }}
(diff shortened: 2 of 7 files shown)


sarayourfriend:
The new workflow successfully created the draft releases: 
image

This is possible to test because I've set it to do so on `pull_request`. But for the actual release workflow, there's no way to test that without putting the workflow on `main`.

documentation/api/guides/deploy.md
```diff
@@ -4,17 +4,10 @@
...
-   `latest` in the Docker image.
-   ![GitHub package directory screenshot](/_static/package_directory_example.png)
-1. Release the app via
-   [GitHub workflow](https://github.com/WordPress/openverse/actions/workflows/release-app.yml).
-   Click the "Run workflow" button, choose "api" from the dropdown, and supply
-   the SHA identified in step 1.
+1. [Publish the drafted API release in the GitHub release page of the monorepo](https://github.com/WordPress/openverse/releases?q=api-)
+   - Here you can preview the changes included in the API release and decide
+     whether a release is necessary and adjust monitoring during the deployment
+     accordingly.
```
AetherUnbound, line 10: Note here (and for all other cases):
```suggestion
   - Here you can preview the changes included in the API release, and decide
     whether a release is necessary, and adjust monitoring during the deployment
     accordingly.
```

AetherUnbound (approved):
This makes sense! One editorial note for the release documents. 

Rephrasing this for my own understanding: The release drafts for _every_ service are made on each commit to main. When we want to deploy a release, we are now only able to deploy the `latest` image. That's done by publishing the draft, which creates a release and kicks off the deployment. We can still perform rollbacks by using the deployment workflows in the infrastructure repo with a specific tag. 

Is that all correct?

sarayourfriend:
Yes, everything else remains the same, the only difference is that now you publish a release instead of manually dispatching the release app workflow.

Using the latest tag is the only reasonable option I can think of, and the release notes were always based on that anyway, they didn't, for example, try to only include changes up to the SHA passed in, it always used the latest changes. In any case, the release notes will let us know what is actually in that SHA now, which is more useful than a random guess of "I guess whatever is in staging now is probably fine".

From my perspective there is no good reason to release something that is not the current staging version, unless staging is for some reason broken, in which case it needs to be fixed before we cut a release! If we absolutely need to deploy a specific SHA to production, which includes rollbacks but that is just a side effect of being able to deploy any arbitrary image tag, we can still do so using the processes that already existed.

documentation/api/guides/deploy.md
sarayourfriend: I'm not sure I agree with this editorial change, and it doesn't match the way I would think to read this sentence either. In any case these commas aren't strictly required :thinking:

AetherUnbound:
Awesome, totally aligned with your perspective there! Just wanted to make sure I understood the new approach :)

obulat (approved):
It's so nice to be able to see what you're deploying!
Would you still be able to edit the drafted release notes as you can do it now? Sometimes the PRs are categorized incorrectly, and I update the notes.

sarayourfriend:
>  Would you still be able to edit the drafted release notes as you can do it now? Sometimes the PRs are categorized incorrectly, and I update the notes.

It should be, right before releasing. Whenever a new change is pushed to main I think it always re-writes the body and fully replaces it. But if you edit the draft release body, then publish it (which you can do from the same page), that's the body that will get used when generating the changelog PR. The release app workflow no longer interacts directly with the drafted release, it only reads the body from the github event payload in context, so it's always working from the published release's body. That more flexibility than we have now, where we (technically) have to edit the PR _and_ the GitHub release body. I know that I've edited changelog PRs before and never did it occur to me to also update the GitHub release body (!), so it'll be nice that future ones should match.

Issue #3761, opened 2024-02-07
`create_proportional_by_source_staging_index` DAG does not base proportions off source index


stacimc:
## Problem

As described in the [IP](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.html#id2), the `create_proportional_by_source_staging_index` uses the stats endpoint (e.g. https://api.openverse.engineering/v1/images/stats/) to get the record counts by source in production. It then creates a new index that is a fraction of the size (as determined by the `percentage_of_prod` param), but with the same proportion of records per each source.

The problem is that the stats endpoint only provides stats for the main, unfiltered media index, while the proportional index DAG allows you to create indices based off of any production index (and in fact defaults to the filtered index). So if you try to make a new index that is 50% of the filtered index for example, your new index will match the filtered configuration and draw records from that index -- but it will be 50% of the size of the _unfiltered_ index and match its source proportions.

This is unintuitive. A few potential problems:
* The size of the index may be much bigger than expected if the `source_index` is a small subset of the main media index
* If the `source_index` has very different proportions than the main media index -- for example, if a particular source is heavily filtered -- this will not be apparent in the new index.

## Description

We should be careful about making changes to the public api, for example by adding stats for other indices. However, we could replace this logic in the `create_proportional_by_source_staging_index` DAG by querying elasticsearch directly.

This query can be used:
```
{
    "size": 0,
    "aggs": {
        "unique_sources": {
            "terms": {
                "field": "source",
                "size": 100,
                "order": {"_key": "desc"}
            }
        }
    }
}
```
In fact this is what the search controller is doing for the [stats endpoint](https://github.com/WordPress/openverse/blob/main/api/api/controllers/search_controller.py#L613).

stacimc:
Setting the priority relatively low because for the use cases we can anticipate, the difference won't be severe -- although even then, I think this is very counter-intuitive and needs to be resolved.

I checked the record counts for the filtered and unfiltered image indices in production. Unsurprisingly the proportion for Flickr is the one with the biggest difference. Somewhat surprisingly, it is a _bigger_ proportion of the _filtered_ index.

| | image | image-filtered | Delta |
| --- | --- | --- | --- |
|Total image count | 749,056,559 | 743,672,348 | 5,384,211 |
|Flickr image count | 507,925,257 | 50,4631,702 | 3,293,555 |
[...]

PR #3763, opened 2024-02-07
Add proportional staging index dag

### LINKED ISSUE #3488: Create a DAG for creating es indices proportional by provider
## Description

Create the `create_proportional_by_provider_staging_index` DAG as described in the Staging Elasticsearch Reindex DAGs IP [here](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.html#create-proportional-by-provider-media-type-staging-index-dag).

stacimc:
## Fixes

Fixes #3488 by @stacimc 

## Description

This PR adds a new `create_proportional_by_source_staging_index` DAG as described in the IP [here](https://docs.openverse.org/projects/proposals/search_relevancy_sandbox/20230530-implementation_plan_staging_elasticsearch_reindex_dags.html#create-proportional-by-provider-media-type-staging-index-dag):
Screenshot 2024-02-07 at 12 53 35 PM

It can be used to create a new index in staging which is a subset of a staging index, but maintains the same proportion of records per source.

Note: reindexing is not parallelized with slicing, as it is for `create_new_production_es_index` and other DAGs (see comment thread for explanation). However it is parallelized in the sense that the individual reindexing tasks for each source can run in parallel. Both of these are acceptable because the DAG only touches staging indices, not production.

## Testing Instructions

Tests assume you're starting from fresh local testing data, ie after running `just api/init`.

* Run the `create_proportional_by_source_staging_index` DAG locally with default conf options. Then verify in elasticvue that you have a new es index called `audio-50-percent-proportional-20240214t181238` (the final suffix is a timestamp and so will differ), with the alias `audio-subset-by-source`.
  * By default it will be based on the `audio-filtered` index and should have 50% of the total documents, with the same source proportions. Since the filtered index has 4912 documents, the new index should have 2456.
  * Inspect the logs for `get_staging_source_counts` to see the counts for the source index, and then the `wait_for_reindex` tasks to see how many were reindexed for each source in the destination index. The proportions should remain the same. For example `wikimedia_audio` has 3992 records in the source index, and 1996 in the new one.
  * Note that the `remove_existing_alias` step was skipped in this dagrun
* Run the DAG again, but this time change the `percentage_of_prod` to 0.25 and the `source_index` to `audio`.
  * You should get a new index named something like `audio-25-percent-proportional-20240214t182518`
  * No steps should be skipped this time. The `audio-subset-by-source` alias should have been removed from the previous index and is now applied to this one.
  * The index should have 1250 documents (1/4 of 5000 from the main audio index)
* Run the DAG with media type `image` and ensure this also works. You should get a new index with `image-subset-by-source` alias, a name like `image-50-percent-proportional-20240214t181238`, and 2328 documents (half of the 4656 that are in the filtered image index). Check the proportions are correct.
* Run the `staging_elasticsearch_cluster_healthcheck` and `create_new_staging_es_index` DAGs to ensure that refactoring out shared elasticsearch utilities did not break them.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


stacimc:
I think I am going to hold off on this and implement the fix for #3761 in this branch, because otherwise we have to add a new Airflow connection for the API which will just get removed later.

stacimc:
The fix for #3761 was added in this branch.

I've encountered another issue with auto-slicing to parallelize the reindexing step. This works fine in DAGs that don't use `max_docs` (like `create_new_es_index`), and for dagruns targeting the audio indices (which only have one primary shard). For dagruns targeting the image indices, the reindexing tasks would consistently fail with the error:

```
ApiError: ApiError(500, 'search_phase_execution_exception', '[slice] can only be used with [scroll] or [point-in-time] requests')
```

The [docs state](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/docs-reindex.html#docs-reindex-api-query-params) that max_docs `When set to a value less then or equal to scroll_size then a scroll will not be used to retrieve the results for the operation.` This disabling of scroll then causes slicing to break. Scroll size defaults to 1000. This can be addressed in a few ways:

* Reduce the value of `size` (the scroll size). This may have a negative effect on reindexing efficiency.
* Set slices to `None`. Reindexing will not be parallelized but the problem is avoided.
* Set `conflicts='proceed'` (instead of the default 'abort').
  * After looking into the code, I realized scroll is only disabled if reindex is set to [abort on version conflicts](https://github.com/elastic/elasticsearch/blob/8e53b0f0bffd18e8ac09a2241b900dae98e954a7/modules/reindex/src/main/java/org/elasticsearch/reindex/AbstractAsyncBulkByScrollAction.java#L211). I **think** we can safely set this to 'proceed', since we're reindexing into an empty index and should not expect version conflicts. This prevents the scroll issue.

I implemented the last of these options in [this commit](https://github.com/WordPress/openverse/pull/3763/commits/a87ce548882cd88ae38f3252e87f745300150406), but while this fixed the errors I encountered _another_ issue. As noted in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/docs-reindex.html#docs-reindex-api-query-params):

> using max_docs with slices might not result in exactly max_docs documents being reindexed.

And in fact you can see, if you check out this branch at the commit mentioned above, that if you run it against the `image-filtered` index with `percentage_of_prod=0.25`, we end up with a new index that has fewer records than expected because `max_docs` is not respected when using slices. In my test run, I got 852 records instead of the expected 1164. This is in my opinion not acceptable because it results in the proportions by source being off.

As a result, **I do not think it is possible to use slicing to parallelize reindexing for this DAG**, and still achieve exact record proportions. The most recent commit on this PR handles it by disabling slicing. This will necessarily be less efficient, although I'm not sure how much.

stacimc:
Noting this will have to be rebased when #3805 is merged.

stacimc:
#3805 was merged, so this has been rebased.

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index_dag.py
```diff
@@ -0,0 +1,188 @@
...
+        "media_type": Param(
+            default=AUDIO,
+            enum=MEDIA_TYPES,
+            description="The media type for which to create the index.",
+        ),
+        "percentage_of_prod": Param(
+            default=0.5,
+            type="number",
+            minimum=0,
+            maximum=1,
```
krysal, line 90: I tried 0 for the percentage out of curiosity and it failed in the `trigger_reindex` task with the following error:
```
elasticsearch.BadRequestError: BadRequestError(400, 'x_content_parse_exception', '[1:181] [reindex] failed to parse field [max_docs]')
```

It doesn't make sense to try creating an index of zero items anyway, so an exclusive minimum would be more appropriate here (I learned [this validation](https://json-schema.org/understanding-json-schema/reference/numeric#range) exists recently!).

```suggestion
        "percentage_of_prod": Param(
            default=0.5,
            type="number",
            exclusiveMinimum=0,
            maximum=1,
```

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index.py
```diff
@@ -0,0 +1,98 @@
...
+
+    response = es_conn.search(
+        index=source_index,
+        size=0,
+        aggregations={
+            "unique_sources": {
+                "terms": {"field": "source", "size": 100, "order": {"_key": "desc"}}
+            }
+        },
+    )
```
krysal, line 68: Good idea to count the items of each source by directly querying ES 💯 

To clarify, what does the `"size":100` part?

krysal (approved):
So exciting to see this complete! I tried creating several indexes as suggested, and everything worked wonderfully 👏 Good eye on avoiding related ES/DB DAGs from conflicting with each other. Everything is carefully thought out, as always.

This is excellent. Thank you, @stacimc!

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index_dag.py
stacimc: Cool, TIL!

catalog/dags/common/elasticsearch.py
```diff
@@ -0,0 +1,223 @@
...
+
+from common.constants import REFRESH_POKE_INTERVAL
+
+
+logger = logging.getLogger(__name__)
+
+
+# Index settings that should not be copied over from the base configuration when
+# creating a new index.
+EXCLUDED_INDEX_SETTINGS = ["provided_name", "creation_date", "uuid", "version"]
```
AetherUnbound, line 19: Nit: This should be a set for check speed and clarity

```suggestion
EXCLUDED_INDEX_SETTINGS = {"provided_name", "creation_date", "uuid", "version"}
```

catalog/dags/common/elasticsearch.py
```diff
@@ -0,0 +1,223 @@
...
+            source=source,
+            dest={"index": destination_index},
+            max_docs=max_docs,
+            # Parallelize indexing when not None
+            slices=slices,
+            # Do not hold the slot while awaiting completion
+            wait_for_completion=False,
+            # Immediately refresh the index after completion to make
+            # the data available for search
+            refresh=True,
```
AetherUnbound, line 127: Should this be a configurable? I imagine for standard reindices this would hold true, but for the proportional-by-source, it might mean several dozen refreshes (one after each source is done being indexed). If we went that route, we might need a separate `refresh` step at the very end once indexing is all done.

catalog/dags/common/elasticsearch.py
```diff
@@ -0,0 +1,223 @@
...
+        response = es_conn.tasks.get(task_id=task_id)
+
+        count = response.get("task", {}).get("status", {}).get("total")
+        if expected_docs and count != expected_docs:
+            logger.info(
+                f"Reindexed {count} documents, but {expected_docs} were expected."
+            )
+        else:
+            logger.info(f"Reindexed {count} documents.")
+        return response.get("completed")
```
AetherUnbound, line 145: FWIW it looks like we can use `@task.sensor` here for this instead of `PythonSensor`! https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/python.html#pythonsensor

Might be useful to make this a reusable function too? Your call 😄

catalog/dags/common/elasticsearch.py
```diff
@@ -0,0 +1,223 @@
...
+        index_name: str,
+        alias: str,
+    ):
+        es_conn = ElasticsearchPythonHook(hosts=[es_host]).get_conn
+        response = es_conn.indices.put_alias(index=index_name, name=alias)
+        return response.get("acknowledged")
+
+    exists_alias = check_if_alias_exists(alias, es_host)
+
+    remove_alias = remove_existing_alias.override(task_id="remove_existing_alias")(
```
AetherUnbound, line 213: Why is `.override` necessary if it's directly reflecting the function name?

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # `size` is the max number of buckets to return for the aggregation
+    # query, and should be set to a number that exceeds the known
+    # number of sources. Read more at:
+    # https://github.com/elastic/elasticsearch/issues/18838
+    size = 100
+
+    response = es_conn.search(
+        index=source_index,
+        size=0,
```
AetherUnbound, line 68: Do you mind leaving a note here as to why this `size` is `0` and differs from the other `size` written above? 😅

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index_dag.py
```diff
@@ -0,0 +1,188 @@
...
+* media_type:         The media type for which to create a new index.
+* percentage_of_prod: A float indicating the proportion of items to take from each
+                      source from the total amount existing in the staging
+                      source index
+
+Optional params:
+
+* source_index: An existing staging Elasticsearch index to use as the basis for
+                the new index. If not provided, the index aliased to
+                `<media_type>-filtered` will be used.
```
AetherUnbound, line 21: Honestly, given how clearly this information is conveyed on the Trigger DAG page, I think it's okay to remove it from the DAG description and de-duplicate the places we're referencing it!
image

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index_dag.py
```diff
@@ -0,0 +1,188 @@
...
+        percentage_of_prod="{{ params.percentage_of_prod }}",
+    )
+
+    reindex = es.trigger_and_wait_for_reindex.partial(
+        destination_index=destination_index_name,
+        source_index=source_index_name,
+        timeout=timedelta(hours=12),
+        requests_per_second=Variable.get(
+            "ES_INDEX_THROTTLING_RATE", 20_000, deserialize_json=True
+        ),
```
AetherUnbound, line 162: Do you think we can change this to a [Jinja templated variable get](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html) so we're not making DB calls at DAG parse time for it?

AetherUnbound (approved):
This is **fantastic**! Great work on implementing this Staci, it's very clear and works flawlessly locally. I have a few nits/thoughts, but nothing to block a merge 🚀

catalog/dags/common/elasticsearch.py
stacimc: That's a good optimization, for sure! I updated it to make this configurable, set it to `False` for this DAG in particular, and added a final `refresh` task after the reindexing is complete.

catalog/dags/common/elasticsearch.py
stacimc: Huh, zero reason -- I must have changed the function names at some point 😄 Cleaned up!

catalog/dags/elasticsearch_cluster/create_proportional_by_source_staging_index/create_proportional_by_source_staging_index_dag.py
stacimc: I'm of two minds on this. I agree that the duplication is annoying, but (maybe not in this case specifically, but especially for some of the more complicated DAGs) it can be difficult to convey how a DAG works and what its available options are without just describing the params. Ideally, I think that information _should_ be conveyed from DAGs.md. It would be really nice if we could include this programmatically rather than duplicating the strings though 😓

AetherUnbound: That's fair, we can leave it for now then 🙂

PR #3785, opened 2024-02-12
Refactor the media collection

### LINKED ISSUE #3576: Returning to results from the new content views does not load previously-loaded pages
## Description

In the new tag/creator/source views, returning to the results after loading multiple pages of results does not reload all of the pages that were previously loaded. If an item was on page 3, this would require the user to reload the first two pages every time they go back to the result list.

## Reproduction

**Current behavior**

1. Go to https://staging.openverse.org/search/?q=pixel
2. Load multiple pages
3. Select any result (doesn't have to be from beyond page 1, but it could)
4. Click "Back to results"
5. Observe that the previously loaded remain present (even if the page doesn't scroll back to the original position due to #576)

**New views**

1. Go to https://staging.openverse.org/image/source/rawpixel/
2. Load multiple pages
3. Select any result
4. Click "Back to results"
5. Observe that only the first page is reloaded

### LINKED ISSUE #3830: Navigating between additional search views and single result pages and back does not update the single image
## Description

Single result pages show previously loaded image instead of the one that the path id parameter when navigating back.

## Reproduction

1. Go to `/preferences` and switch the `additional_search_views` on.
2. Search for "cat". Select a single result. Click on the creator button to go to the creator page. Select one of the images. Click on the source button. Select one of the images. Try to select different images to see the error more clearly.
3. Go back using the browser "Back" button.
4. See error.

## Screenshots

Notice that the image of a column is displayed instead of the olive image for the second single result page:

https://github.com/WordPress/openverse/assets/15233243/0e987439-ddfc-4db7-bccb-f6966241ffd3

obulat:
## Fixes

Fixes #3576 by @AetherUnbound
Fixes #3830 by @obulat 

_Note: this PR does not fix the problem when the creator/tag have special symbols (`/` or `?`)_

## Description

Currently, we fetch the search results in two separate places. The `search.vue` component calls `fetchMedia` when the search parameters like the search type, search term or filters change. The `VLoadMore` component calls `fetchMedia` when the user clicks on "Load more" button, fetching the pages after the first one.

This means that we cannot use the list of media returned in `search.vue`'s `fetchMedia` call, because it won't be updated when the "Load more" button is clicked, and only the first page will be shown.
To work around this limitation, we were using the computed that returned the value of media results from the media store.  These values were passed to the image/audio search pages, but the All content page would use a separate computed.

This PR tries to separate the concerns more cleanly.

### Data fetching
The **pages** (`search.vue`, `_tag.vue`, `_.vue` for source and creator views) are responsible for **fetching the data**, including the load more fetching (that was previously done by the `VLoadMore` component). 

### Data display
The `VMediaCollection` is responsible for displaying the media search results or the loading skeleton.

It handles both main and additional search views, and displays the header with the title and link buttons, the media grid, the load more button and the external search form (if necessary) and the scroll to top button.

### Updating the data on navigation

The linked issue is that only the first page of additional search views results was kept if you navigate to the single result page from the second page (or third, fourth, etc) and back. To fix it, we set the media to the media results from the store if the `collectionParams` match the path parameters. 
This fix created a different issue: the old media results were shown when you would navigate between the single result views and the additional search results views. For instance, when going back from the single results views, or when navigating between the views (e.g., tag view -> single result -> creator view -> single result -> source view and back). To fix that, we clear the media in the `media` store if the `collectionParams` parsed from the path do not match the one in the search store.

## Testing Instructions

Make sure the `additional_search_views` feature flag is `on` (use the switch on the `/preferences` page)
Go to a tag view (localhost:8443/image/tag/cat), load a couple more pages of results. Then select one of the results from the last page. Go back by clicking "Back to results" or the browser's "Back" button. You should see that all of the previously loaded results are displayed (instead of the first page as in current implementation).

Also try to navigate between the additional search views and the single result views (e.g. tag/cat -> single result -> creator view -> single result -> source view and back), and check that the title of the additional search view always matches the results. You should never see, for example, results for a creator and title of "tag Cat".

Repeat the steps for #3830 and check that it's fixed.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/pages/search.vue
```diff
@@ -199,27 +190,27 @@ export default defineComponent({
     })
 
     useFetch(async () => {
+      console.log("[useFetch] needsFetching", needsFetching.value)
```
obulat, line 193:
```suggestion
```

frontend/src/pages/search/index.vue
```diff
@@ -1,14 +1,37 @@
...
+    :collection-label="
+      $t('browsePage.aria.results', { query: searchTerm }).toString()
+    "
+    @load-more="handleLoadMore"
+  />
 </template>
 
 <script lang="ts">
-import { defineComponent } from "vue"
+import { defineComponent, PropType } from "vue"
```
obulat, line 13:
```suggestion
import { defineComponent, type PropType } from "vue"
```

frontend/test/playwright/utils/components.ts
```diff
@@ -19,7 +19,11 @@ export const getMenuButton = async (
   page: Page,
   dir: LanguageDirection = "ltr"
 ) => {
-  return page.getByRole("button", { name: t("header.aria.menu", dir) })
+  const menuButton = page.getByRole("button", {
+    name: t("header.aria.menu", dir),
+  })
+  await expect(menuButton).toBeEnabled()
```
obulat, line 25: Ensure that the page is hydrated before we open the pages modal. Otherwise, the tests fail because the event handlers haven't loaded yet.

frontend/test/unit/specs/pages/browse-page.spec.js
obulat, line 1: This was only testing if the scroll button is shown. Since the button was moved to `VMediaCollection`, it's failing now. Should be moved to e2e tests.

AetherUnbound (commented):
I really don't feel like I can give good feedback on the code, but two issues I noticed:
- The "load more results" button on a provider source page doesn't seem to work: http://localhost:8443/audio/source/wikimedia_audio/
- Creator names with slashes in them get interpreted as a full path (and thus 404): http://localhost:8443/audio/source/wikimedia_audio/creator/Holger%20Klein/Helmholtz-Gemeinschaft/ - I'll make a separate issue for this!
- On this PR, clicking Back from the above example, then hitting "back to results" puts you back to the page it couldn't find in the first place, which seems unintended 😮

obulat:
> I really don't feel like I can give good feedback on the code, but two issues I noticed:
> 
> * The "load more results" button on a provider source page doesn't seem to work: http://localhost:8443/audio/source/wikimedia_audio/
> * Creator names with slashes in them get interpreted as a full path (and thus 404): http://localhost:8443/audio/source/wikimedia_audio/creator/Holger%20Klein/Helmholtz-Gemeinschaft/ - I'll make a separate issue for this!
> * On this PR, clicking Back from the above example, then hitting "back to results" puts you back to the page it couldn't find in the first place, which seems unintended 😮

Thank you for testing! I didn't even know that we have creators with slashes in their names.
I rebased this PR onto #3793 that allows slash in the creator name.

I also forgot to add the load more handler to the audio source page, thank you for catching! It now works.

frontend/src/pages/image/_id/index.vue
```diff
@@ -168,7 +168,12 @@ export default defineComponent({
 
     const route = useRoute()
 
-    const image = ref<ImageDetail | null>(singleResultStore.image)
+    const image = ref<ImageDetail | null>(
```
obulat, line 171: This is one part that fixes #3830 (not re-using the store image if the id doesn't match)

obulat:
Closing this PR because it was replaced with the following three:
#3832
#3833
#3835

PR #3789, opened 2024-02-12
Fix newline at end of folded env variable string and prevent confusing skips of the changelog PRs

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/actions/runs/7879001320/job/21498432707

## Description

Releases currently fail to complete due to the `env.APP_NAME` having a newline at the end of the string: https://github.com/WordPress/openverse/actions/runs/7879001320/job/21498432707

Converting it to a single line is necessary because, as Madison pointed out, you cannot use a chomped multiline string (`>-`) in GitHub actions.

This PR also fixes the changelog PR step to only run if the rest of the workflow completes and to use the correct variable in place of `git-tag`. I searched for other usages of `git-tag` to be sure and those were the last remaining ones.

## Testing Instructions

Confirm the change will work as expected based on the information about yaml multiline strings in the linked documentation.

## Checklist

- [N/A] I tried running the project locally and verified that there are no visible errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/release-app.yml
+++ b/.github/workflows/release-app.yml
@@ -29,13 +29,9 @@ jobs:
       # Needed to open the changelog PR
       pull-requests: write
     env:
-      APP_NAME: >-
-        ${{
-          startsWith(github.ref_name, 'api-') && 'api'
-          || startsWith(github.ref_name, 'ingestion_server-') && 'ingestion_server'
-          || startsWith(github.ref_name, 'catalog-') && 'catalog'
-          || startsWith(github.ref_name, 'frontend-') && 'frontend'
-        }}
+      # Do not split this into multiple lines, it will not work:
+      # https://github.com/WordPress/openverse/pull/3789#pullrequestreview-1876525552
+      APP_NAME: ${{ startsWith(github.ref_name, 'api-') && 'api' || startsWith(github.ref_name, 'ingestion_server-') && 'ingestion_server' || startsWith(github.ref_name, 'catalog-') && 'catalog' || startsWith(github.ref_name, 'frontend-') && 'frontend' }}
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/release-app.yml
+++ b/.github/workflows/release-app.yml
@@ -29,7 +29,7 @@ jobs:
       # Needed to open the changelog PR
       pull-requests: write
     env:
-      APP_NAME: >
+      APP_NAME: >-
         ${{
           startsWith(github.ref_name, 'api-') && 'api'
           || startsWith(github.ref_name, 'ingestion_server-') && 'ingestion_server'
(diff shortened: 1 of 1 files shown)


AetherUnbound (changes requested):
Unfortunately, I don't think this will work because newline chomping isn't supported in the GHA YAML spec 😞 see: https://github.com/WordPress/openverse/pull/3702

.github/workflows/release-app.yml
```diff
@@ -122,14 +122,13 @@ jobs:
 
       - name: Open changelog PR
         uses: peter-evans/create-pull-request@v5
-        if: "!cancelled()"
```
stacimc, line 125: I'm not sure I understand why this was removed -- it sounds like it relates to this part of the description:

> This PR also fixes the changelog PR step to only run if the rest of the workflow completes

Was this check broken and is no longer necessary? Is it related to the critical bug?

sarayourfriend:
Whaaat! That's wild! I'll just convert it to a single line then.

AetherUnbound (approved):
LGTM!

.github/workflows/release-app.yml
sarayourfriend: It isn't part of the critical bug, but it doesn't make sense at all, `!cancelled` would also run if the workflow fails, which you can see in the linked failure. That doesn't make sense: if we skip the changelog when we cancel the release workflow, why would we want it if the release workflow failed? We can close the PR if we don't want the changelog, and we'll need to delete the tag in that case too anyway (and generate a new drafted release when we push whatever fix we need).

stacimc: >  if we skip the changelog when we cancel the release workflow, why would we want it if the release workflow failed? 

Agree with the logic here, although I don't understand how removing this check does what is described in the PR description (actually preventing the step from running if the rest of the workflow hasn't completed).

stacimc (approved):

.github/workflows/release-app.yml
sarayourfriend: Because `!cancelled()` is strictly _only if the workflow isn't cancelled_, meaning it still runs if the workflow failed. I don't know how this is the case, but you can see it in the linked failure, where a previous step in the workflow fails but the changelog PR step still runs.

stacimc: I understand that and agree it's not useful to only check that case. I assumed (I think wrongly?) that removing this check would mean that now it will run in all cases, not that it would only run if the previous steps completed as described.

sarayourfriend: Ah, I see! The default behaviour is to run only if the previous step succeeded, in effect having `if: success()` by default (https://docs.github.com/en/actions/learn-github-actions/expressions#success).

To run something in every case, you'd have to explicitly use `always`: https://docs.github.com/en/actions/learn-github-actions/expressions#always

PR #3793, opened 2024-02-13
Allow using special characters in creator name and tags

### LINKED ISSUE #3791: Unable to visit creator page if creator name has `/` in it
## Description

In the new creator view, if a creator has a `/` in it, the part after the `/` is interpreted as a path and 404s.

## Reproduction

1. Run the frontend locally
2. Make sure you have `additional_search_views` enabled in `/preferences`
3. Visit http://localhost:8443/audio/source/wikimedia_audio/
4. Click on any of the results with `/` in them
5. See 404

## Screenshots

image

## Additional context

Related to #410

obulat:
## Fixes

Fixes #3791 by @AetherUnbound

## Description

This PR changes the way the path is parsed to extract the creator name to allow for URL-encoded special symbols like `/` and `?` within it. The creator name is also encoded when creating the app creator page link.
To enable testing this change locally, I added special symbols to the creator names sample data.

## Testing Instructions

Run the local API. 

Then, run the Nuxt app using the local API by running the app using `API_URL=http://localhost:50270 just frontend/run dev:only`

Enable `additional_search_views` on the `/preferences` page.
~~Go to `localhost:8443/audio/source/wikimedia_audio` and navigate to one of the results that have a slash in the creator name. From that single result page, click on the creator link to open the creator view - there should be no errors here.~~

Search for "Wish You Were Here" in audio. You should see an item with a question mark in the creator name. If you click on it, you should see that the name is encoded in the URL: http://localhost:8443/audio/source/jamendo/creator/The.madpix.project%3F/

If you search for "German term Vulkan" in audio, you should see an item with a `/` in the creator name. If you click on it, you should see the slash as is in the URL, however, this slash is actually encoded in the app code. You can get to this page using this URL: http://localhost:8443/audio/source/wikimedia_audio/creator/wiki%2FJeuwre/ 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
Really surprised the API handles this so well, but there is a big edge case... `?`!

http://0.0.0.0:50280/v1/images/source/flickr/creator////me////?you-and-them/\////

Leads to the following query:

```
openverse-web-1  | {'from': 0,
openverse-web-1  |  'query': {'bool': {'filter': [{'term': {'source': 'flickr'}},
openverse-web-1  |                                {'term': {'creator.keyword': '///me///'}}],
openverse-web-1  |                     'must_not': [{'term': {'mature': True}}]}},
openverse-web-1  |  'size': 40,
openverse-web-1  |  'sort': [{'created_on': {'order': 'desc'}}]}
```

I wonder if we should just percent encode the creator name? Or maybe conditionally percent encode, at least `?`, and any other significant things we could find?

Not requesting changes because it could be considered a separate issue. Let me know!

frontend/src/utils/parse-collection-path.ts
```diff
@@ -1,23 +1,32 @@
-import { CreatorCollection, SourceCollection } from "~/types/search"
+import type { CreatorCollection, SourceCollection } from "~/types/search"
+
+const removeSlashes = (path: string): string => {
+  return path.replace(/^\/|\/$/g, "")
+}
```
sarayourfriend, line 5: Does this only remove leading and trailing slashes from the path? What about trailing or leading slashes in creator names? e.g., `image/source/flickr//edgycreator//` where the creator name is `/edgycreator/`?

sample_data/sample_audio.csv
```diff
@@ -1,5 +1,5 @@
...
...ot a single day that passes, without you on my mind. Not even one minute can end up, before you come around. I aim for the days when I see your face here before me. I long for the days when you held me in your arms. How I wish that you were here to give me everything I want. But I know that live's unfair and we can't always have it all.. I wish that you were here."", ""listens"": ""17959942"", ""downloads"": ""0"", ""playlists"": ""0"", ""license_url"": ""https://creativecommons.org/licenses/by-nc-sa/3.0/"", ""release_date"": ""2015-03-29"", ""raw_license_url"": ""http://creativecommons.org/licenses/by-nc-sa/3.0/""}","[{""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""female"", ""provider"": ""jamendo""}, {""name"": ""speed_medium"", ""provider"": ""jamendo""}, {""name"": ""guitar"", ""provider"": ""jamendo""}, {""name"": ""strings"", ""provider"": ""jamendo""}, {""name"": ""energetic"", ""provider"": ""jamendo""}, {""name"": ""acoustic"", ""provider"": ""jamendo""}, {""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""voice"", ""provider"": ""jamendo""}, {""name"": ""funkyhouse"", ""provider"": ""jamendo""}]",false,2021-09-30 19:00:43.800462+00,false,,145774
```
sarayourfriend, line 2:
```suggestion
8624ba61-57f1-4f98-8a85-ece206c319cf,2022-12-06 06:54:25.000000+00,2022-12-06 06:54:25.000000+00,provider_api,jamendo,jamendo,1214935,https://www.jamendo.com/track/1214935,https://mp3d.jamendo.com/download/track/1214935/mp32,https://usercontent.jamendo.com?type=album&id=145774&width=200&trackid=1214935,mp3,270000,128000,44100,music,"{dance,electronic,house}","{""url"": ""https://usercontent.jamendo.com?type=album&id=145774&width=200"", ""title"": ""Wish You Were Here"", ""creator"": ""The.madpix.project?"", ""creator_url"": ""https://www.jamendo.com/artist/441585/the.madpix.project"", ""foreign_identifier"": ""145774"", ""foreign_landing_url"": ""https://www.jamendo.com/album/145774/wish-you-were-here""}",1,,7139840,by-nc-sa,3.0,The.madpix.project?,https://www.jamendo.com/artist/441585/the.madpix.project%3F,Wish You Were Here,"{""lyrics"": ""There's not a single day that passes, without you on my mind. Not even one minute can end up, before you come around. I aim for the days when I see your face here before me. I long for the days when you held me in your arms. How I wish that you were here to give me everything I want. But I know that live's unfair and we can't always have it all.. I wish that you were here."", ""listens"": ""17959942"", ""downloads"": ""0"", ""playlists"": ""0"", ""license_url"": ""https://creativecommons.org/licenses/by-nc-sa/3.0/"", ""release_date"": ""2015-03-29"", ""raw_license_url"": ""http://creativecommons.org/licenses/by-nc-sa/3.0/""}","[{""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""female"", ""provider"": ""jamendo""}, {""name"": ""speed_medium"", ""provider"": ""jamendo""}, {""name"": ""guitar"", ""provider"": ""jamendo""}, {""name"": ""strings"", ""provider"": ""jamendo""}, {""name"": ""energetic"", ""provider"": ""jamendo""}, {""name"": ""acoustic"", ""provider"": ""jamendo""}, {""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""voice"", ""provider"": ""jamendo""}, {""name"": ""funkyhouse"", ""provider"": ""jamendo""}]",false,2021-09-30 19:00:43.800462+00,false,,145774
```

AetherUnbound:
> I wonder if we should just percent encode the creator name? Or maybe conditionally percent encode, at least `?`, and any other significant things we could find?

This was exactly my thought for approaching this too! I'm not sure how much it matters, but leaving the `/` unencoded forces us into the position of accepting that we _won't_ have any future sub-paths to `/creator/`, since those sub-paths might be interpreted as the creator name. FWIW I think it's worth changing this PRs behavior to URL encode things!

obulat:
> This was exactly my thought for approaching this too! I'm not sure how much it matters, but leaving the `/` unencoded forces us into the position of accepting that we _won't_ have any future sub-paths to `/creator/`, since those sub-paths might be interpreted as the creator name. FWIW I think it's worth changing this PRs behavior to URL encode things!

I added encoding to the link that we create for the creator button in [9a535b9](https://github.com/WordPress/openverse/pull/3793/commits/9a535b98e258a9ea8aacfa166b3db15ef57799de). If you want to access a page for a creator that has special symbols like `/` or `?` in it by entering the URL, you would have to manually encode the creator name.

AetherUnbound:
I'm not sure if it's just my machine, but I'm having a really hard time testing this 😞 lots of the pages aren't loading, and I can't navigate to any of the links you posted (even after trying `j down -v && j api/init` and using the API URL variable directly). Some of the pages half load or stay loading forever. It looks like not all the content on the pages is loading because my default font is being used for some of the pages that get half rendered. I tried the _exact_ same steps on `main` (including pointing to the local API) and the pages load and function as expected 😕

obulat:
> I'm not sure if it's just my machine, but I'm having a really hard time testing this 😞 lots of the pages aren't loading, and I can't navigate to any of the links you posted (even after trying `j down -v && j api/init` and using the API URL variable directly). Some of the pages half load or stay loading forever. It looks like not all the content on the pages is loading because my default font is being used for some of the pages that get half rendered. I tried the _exact_ same steps on `main` (including pointing to the local API) and the pages load and function as expected 😕

This is weird. I tried checking out the branch again, and it works well for me locally 🙃 

Just in case, here are the first steps I used for testing:

1. Checkout `fix/allow-slashes-in-creator-name`
2. Run `just install`

4. Run the API locally. Make sure that the updated sample data with special symbols in creator names is used.

6. Run `just up`
7. Delete all indices using Elasticvue
8. Run `just init`

10. `API_URL=http://localhost:50270/ just frontend/run dev:only`

12. Open `http://localhost:8443/preferences` and make sure that the `additional_search_views` is on.

frontend/src/utils/parse-collection-path.ts
sarayourfriend: @obulat have you seen this question? Just want to make sure that `path` gets passed in with the `/`s from the path, and not the raw creator name characters _on both ends of the string and always_.

If that _is_ the case, then can we just do `path.substring(1, path.length - 1)` and avoid the regex altogether? It would communicate things a bit more clearly, especially if the function is renamed to `trimLeadingAndTrailingSlashes` (or some less verbose version :grin: )

sample_data/sample_audio.csv
sarayourfriend: I missed one of the creator URLs in my suggestion.

```suggestion
8624ba61-57f1-4f98-8a85-ece206c319cf,2022-12-06 06:54:25.000000+00,2022-12-06 06:54:25.000000+00,provider_api,jamendo,jamendo,1214935,https://www.jamendo.com/track/1214935,https://mp3d.jamendo.com/download/track/1214935/mp32,https://usercontent.jamendo.com?type=album&id=145774&width=200&trackid=1214935,mp3,270000,128000,44100,music,"{dance,electronic,house}","{""url"": ""https://usercontent.jamendo.com?type=album&id=145774&width=200"", ""title"": ""Wish You Were Here"", ""creator"": ""The.madpix.project?"", ""creator_url"": ""https://www.jamendo.com/artist/441585/the.madpix.project%3F"", ""foreign_identifier"": ""145774"", ""foreign_landing_url"": ""https://www.jamendo.com/album/145774/wish-you-were-here""}",1,,7139840,by-nc-sa,3.0,The.madpix.project?,https://www.jamendo.com/artist/441585/the.madpix.project%3F,Wish You Were Here,"{""lyrics"": ""There's not a single day that passes, without you on my mind. Not even one minute can end up, before you come around. I aim for the days when I see your face here before me. I long for the days when you held me in your arms. How I wish that you were here to give me everything I want. But I know that live's unfair and we can't always have it all.. I wish that you were here."", ""listens"": ""17959942"", ""downloads"": ""0"", ""playlists"": ""0"", ""license_url"": ""https://creativecommons.org/licenses/by-nc-sa/3.0/"", ""release_date"": ""2015-03-29"", ""raw_license_url"": ""http://creativecommons.org/licenses/by-nc-sa/3.0/""}","[{""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""female"", ""provider"": ""jamendo""}, {""name"": ""speed_medium"", ""provider"": ""jamendo""}, {""name"": ""guitar"", ""provider"": ""jamendo""}, {""name"": ""strings"", ""provider"": ""jamendo""}, {""name"": ""energetic"", ""provider"": ""jamendo""}, {""name"": ""acoustic"", ""provider"": ""jamendo""}, {""name"": ""vocal"", ""provider"": ""jamendo""}, {""name"": ""voice"", ""provider"": ""jamendo""}, {""name"": ""funkyhouse"", ""provider"": ""jamendo""}]",false,2021-09-30 19:00:43.800462+00,false,,145774```

File: frontend/src/utils/parse-collection-path.ts (hunk 1 above)
Comment (sarayourfriend) [reply frontend/src/utils/parse-collection-path.ts:5]: I'm realising that potentially my question doesn't make sense anymore, because all slashes that are elements of the creator's name should be URL encoded, right? I am still waiting to approve just to make sure, because http://192.168.20.75:8443/audio/source/jamendo/creator/%2FThe.madpix.project%3F/, with a leading slash in the creator name URL encoded does work, but http://192.168.20.75:8443/audio/source/jamendo/creator//The.madpix.project%3F/ does not.

However, the PR description says that we should be able to visit http://localhost:8443/audio/source/wikimedia_audio/creator/wiki/Jeuwre/, but that doesn't work either, it 404s. The URL encoded version does work.

Except, they only work when navigate to client side. Navigating to them SSR (i.e., reloading the page) does not work. Is it an issue with preferences, perhaps? I also don't see the additional search view links on the single result page, so I think it may be an issue outside this PR, but I want to make absolutely sure it is working the right way, because what I am experiencing differs from the PR description.

**Review by sarayourfriend — CHANGES_REQUESTED** [review]
More a request for clarification than a change, but I do have an additional fix to the sample audio data.

**Comment by sarayourfriend** [comment]
@AetherUnbound I am not having the same issue as you.

We can pair this week to try to discover why your local is behaving differently, if you would like.

**Comment by obulat** [comment]
> I'm realising that potentially my question doesn't make sense anymore, because all slashes that are elements of the creator's name should be URL encoded, right? I am still waiting to approve just to make sure, because http://192.168.20.75:8443/audio/source/jamendo/creator/%2FThe.madpix.project%3F/, with a leading slash in the creator name URL encoded does work, but http://192.168.20.75:8443/audio/source/jamendo/creator//The.madpix.project%3F/ does not.

> However, the PR description says that we should be able to visit http://localhost:8443/audio/source/wikimedia_audio/creator/wiki/Jeuwre/, but that doesn't work either, it 404s. The URL encoded version does work.

> Except, they only work when navigate to client side. Navigating to them SSR (i.e., reloading the page) does not work. Is it an issue with preferences, perhaps? I also don't see the additional search view links on the single result page, so I think it may be an issue outside this PR, but I want to make absolutely sure it is working the right way, because what I am experiencing differs from the PR description.

I was having problems testing the creator names that start with `/` locally: they would be correctly saved as URL-encoded versions, but would redirect to a creator name without a leading slash. Then,  I realized that the `pathMatch` returned by `vue-router` is always URL-decoded, so it never shows the URL-encoded version of the creator.

So, I changed the code a little bit to only allow for URL-encoded creator names. Now, if you use a non-encoded slash `/` in the creator name, it shows a full-page yellow error page. To correctly parse out the encoded creator name, I had to add the `fullPath` as a parameter, as it shows the exact URL instead of always decoding it. Previously, both `%2Fcreator` and `/creator` would be passed as `/creator` in the parser function, and now these are passed as `%2Fcreator` and `/creator`, respectively.

I also moved the path collectionParams validation to the middleware because throwing an error in the setup function shows the Nuxt generic error page instead of Openverse yellow error page.

**Review by sarayourfriend — COMMENTED** [review]
The code LGTM. I'm unable to test this locally at the moment, but will do so later.

Something I didn't realise about the implementation until now, is the use of `_.vue`, rather than typical path parameters. Was that originally to allow for `/` in the creator names? If we are now disallowing it, could we switch to using regular Nuxt path params by changing the file names to `audio/source/[source].vue`, `audio/source/[source]/creator/[creator].vue`, and so forth? Maybe it's even possible to use `[mediaType]/source/[source]/creator/[creator].vue`, if that would simplify the logic altogether (I'm not sure if this is the case).

Would this overall simplify the implementation of these pages or is it more complex for some reason?

Also, do we need to handle URL encoding tags? I don't know how much normalisation we do to them on the presentation layer.

File: frontend/src/components/VCollectionHeader/VCollectionHeader.vue [hunk 3]
```diff
@@ -96,7 +96,7 @@ export default defineComponent({
       if (props.collectionParams.collection === "tag") {
         return props.collectionParams.tag
       } else if (props.collectionParams.collection === "creator") {
-        return props.collectionParams.creator
+        return decodeURIComponent(props.collectionParams.creator)
```
obulat, line 99:
```suggestion
        return props.collectionParams.creator
```

sarayourfriend:
Unfortunately SSR is still broken for creator names, for me :(

I think that this might be actually an issue with trying to use the path at all, potentially with the browser decoding the percent-encoded slash into a plaintext slash upon page load. Then reloading, or copying the URL, results in an usuable URL, because it gives back the URL with the plain-text slash (`http://192.168.20.75:8443/audio/source/wikimedia_audio/creator/wiki/Jeuwre/`). You can't navigate to that.

https://github.com/WordPress/openverse/assets/24264157/86cd4dac-6d57-4fe7-bda0-7c32d1d25c7d

This worries me a bit, because I think it invalidates a core supposition that we had about how to approach this from a technical perspective. Specifically, I think it means we should actually use query parameters instead of trying to treat the creator (and tags) as a RESTful resource. Source is (theoretically) safe from this, because we control the slugs that are used for them, and can ensure they are safe for URL path parameters (do not have `?` or `/`, etc). User generated content, on the other hand, cannot be controlled in this way, and it doesn't look like browsers are going to cooperate with us when we try our URL encoded path workaround.

Unfortunately, I think that means taking a big step back and rethinking at least the creator and tag endpoints, on both the frontend and API side, to ensure they're able to handle the full spectrum of possible values.

AetherUnbound:
I'm having the same difficulties as Sara - even hitting refresh on a page that's client-side rendered will then cause the loading to fail from SSR 😞 And the links in the PR description don't work for accessing media locally. Is moving those potential user-generated values to query parameters the only way? That is disappointing that we seem to be fighting the browser on that front 😕

sarayourfriend:
> That is disappointing that we seem to be fighting the browser on that front

I think it's more that we're fighting the HTTP standard, rather than the _browser_ :shrug: Trying to keep a / in the path parameter was always going to be iffy, that's one of the few significant characters in a URI.

obulat:
> I think it's more that we're fighting the HTTP standard, rather than the _browser_ 🤷 Trying to keep a / in the path parameter was always going to be iffy, that's one of the few significant characters in a URI.

Yes, having a slash in the URL is definitely iffy. But I _think_ I found a solution: if we encode the slash in creator and tag twice, we can make it work without having to fight with the browser's behavior with `/`. I added the changes in [e85b6a8](https://github.com/WordPress/openverse/pull/3793/commits/e85b6a8853b07b4a159e01e5780a7e123f9191c4). To make this work, I also added the necessary decoding step to handle this in the API.

I added the test that handles reloading of the client-rendered page to make sure this case is not missed.

You can test the cases of `?` and `/` in the creator and the tag using the following search terms:
- "Vulkan german term" audio (creator `wiki/Jewre`)
- "house cat" image (tag `house/cat` and `house?cat`)

obulat:
> In fact, it'll be better, because at least in Firefox, you can type whatever cleartext you want in a query parameter, and the browser will URL encode it at request time on your behalf. We were _never_ going to get that with path parameters.

I found a way of allowing for typing creator name as cleartext in the path parameter: in the `collection` middleware, if we're on the server and there's an non-encoded slash, we redirect to the path with creator encoded. Tag cannot use cleartext, though, because it does not use a `catchall` page name as the `source` pages do (it's `_tag.vue`, so it matches only until the following `/`, and source is `_vue`, so it matches everything that comes after). 

However, this would not allow for a clear text `?` question mark in the source name, because the first question mark will be considered the start of the query params. We could also add server-side handling of the question mark: if there are more than 1 question mark, or if there are no `=` signs in the "query params", redirect the page to the URL with the encoded creator path (i.e., for "/creator/me??license=by", we would redirect on the server to "/creator/me%3F?license=by". 

The easiest solution would be not to add clear text handling to the path, i.e, accept only encoded values.
What do you think about this solution, @sarayourfriend ? Is there anything that would make it not feasible and make us refactor the code to use query params instead of the path params?

sarayourfriend:
> The easiest solution would be not to add clear text handling to the path, i.e, accept only encoded values.

I think the issue with this is that browser (or at least Firefox) automatically converts URL encoded values in the path to clear text, so if you copy/paste that value, it ends up as cleartext. It does this with query parameters as well, but browsers know to URL encode query parameters before sending them (because that's standard), whereas they won't know to do that for the path (as it is unusual).

You can see this in my video above. When I navigate to the creator page, despite the link in the HTML actually being url encoded, Firefox even displays it as cleartext in the link preview in the bottom left corner. After navigating to the page, it is cleartext in the URL bar, and when I copy it, it is copied as cleartext.

I think we are doing something fundamentally incompatible with the way browsers expect paths to work. We need to use an approach that will not cause issues when the browser converts the URL to plain text. The only two options I can think of are: using a normalised version of creator names that encodes `/` and `?` is some other format that the browser doesn't know, but we know, and can use to decode to the full name on the API (which seems... bizarre, frankly, especially if we'd need to communicate that to anyone using those API routes) _or_ use query parameters instead of path parameters, which doesn't "look" as nice, but is compatible with every single possible edge case, and works exactly as expected in every single one of those.

sarayourfriend:
I've removed Dhruv as a requested reviewer, I stepped in for him instead.

Also drafted this so that we stop getting review pings while we continue to discuss this and find a solution that will work long term.

obulat:
Closing this PR as we've decided to change the approach and use query parameters for the additional search views.

PR #3808, opened 2024-02-19
Cleanup tag display for long lists of tags

### LINKED ISSUE #2589: Cleanup tag display with long lists of tags
## Problem

Some images, particularly from platforms like Flickr, have many many tags. The display of these tags is distracting, particularly on mobile:

An example: https://openverse.org/image/7e138814-e774-45fb-91ee-7b728c174616?q=Titmouse

Single result page screenshot

## Description

Truncate the tags list to a fixed number of rows. We can do this hackily in CSS with `height` or some other method. Use additional styling to indicate the list is expansible.

obulat:
## Fixes

Fixes #2589 by @zackkrida 

## Description

This PR hides the tags beyond the third row and shows a "Show more tags" button if there are hidden tags.

I first tried simply hiding the tags setting the max height and using `overflow-hidden`. However, this way,  you can navigate to the hidden tags using <kbd>Tab</kbd>. Even though those tags are not visible, they are focusable and are read by the screenreader.

That's why I added a function that detects the index of the tag that is the first on the 4th line of tags, and set the `visibleTags` value to the tags up to the "first on the 4th line". When the user clicks show more, the `visibleTags` is set to all of the tags, and when they click on hide tags, the `visibleTags` array only contains the first 3 lines.

For counting the rows of tags, I used `offsetLeft`, but had to make comparison operator (`<` or `>`) conditional to make it work with RTL (some of the first VR tests here for RTL are failures because I only used one comparison)

The expander button has a `aria-expanded` set to `true` or `false` based on whether the tags are all shown, or hidden.

## Testing Instructions

Enable the `additional_search_views` on the `/preferences` page (note that this is only implemented for when this flag is on)
Go to an image result with many tags, such as `[http://localhost:8443/image/7e138814-e774-45fb-91ee-7b728c174616?q=Titmouse](http://localhost:8443/image/7e138814-e774-45fb-91ee-7b728c174616?q=Titmouse)`. 
Check that only the 3 rows of tags are displayed, with a "Show more tags" button. Try clicking on the button. The first opened tag should be focused, and the "See less" button should be displayed.
Check that this works for pages with other locales, and that the tags for media with fewer tags are displayed as before.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VMediaInfo/VMediaTags.vue
```diff
@@ -1,50 +1,223 @@
...
-  <ul v-else class="flex flex-wrap gap-2">
-    <VMediaTag v-for="(tag, index) in tags" :key="index" tag="li">{{
-      tag.name
+  <div v-if="normalizedTags.length && additionalSearchViews" class="-my-1.5px">
+    <ul
+      ref="tagsContainerRef"
+      :aria-label="$t('mediaDetails.tags.title').toString()"
+      class="flex flex-wrap gap-3 overflow-y-hidden p-1.5px"
+      :class="heightClass"
+    >
```
zackkrida, line 8:
```suggestion
  <div v-if="normalizedTags.length && additionalSearchViews">
    <ul
      ref="tagsContainerRef"
      :aria-label="$t('mediaDetails.tags.title').toString()"
      class="flex flex-wrap gap-3 overflow-y-hidden px-1.5px"
      :class="heightClass"
    >
```

I _think_ this is effectively the same appearance? I might be overlooking something though.

frontend/src/locales/scripts/en.json5
```diff
@@ -498,6 +498,11 @@
     },
     imageInfo: "Image information",
     audioInfo: "Audio information",
+    tags: {
+      title: "Tags",
+      showAll: "Show all tags",
+      seeLess: "See less",
```
zackkrida, line 504:
```suggestion
      showAll: "Show all tags",
      showLess: "See less",
```

I wonder if these keys using the same word, even though the translations do not, would make the relationship of these strings clearer to translators.

frontend/src/utils/focus-management.ts
```diff
@@ -88,8 +88,11 @@ export function isFocusableElement(element: HTMLElement) {
   return element.matches(focusableSelector)
 }
 
-export function focusElement(element: HTMLElement | null) {
-  element?.focus()
+export function focusElement(element: HTMLElement | Element | null) {
+  if (!element || !(element instanceof HTMLElement)) {
+    return
+  }
```
zackkrida, line 94: Nice guard :+1:

frontend/src/components/VMediaInfo/VMediaTags.vue
```diff
@@ -1,50 +1,223 @@
...
+    const localizedTagPath = (tag: string) => {
+      return app.localePath({ path: `tag/${tag}` })
     }
 
-    return { additionalSearchViews, localizedTagPath }
+    const normalizedTags = computed(() => {
+      return Array.from(new Set(props.tags.map((tag) => tag.name)))
+    })
+
+    const fourthRowStartsAt = ref<number>()
```
zackkrida, line 93:
```suggestion
    const rowStartsAtIndex = ref<number>()
```

I would suggest a name like this, to be more explicit, and then we make the "fourth" more of an implementation detail with a constant like `const ROWS_TO_DISPLAY = 4;` at the top of the file. 

Then, instead of `findFourthRowStartsAt` we could have `findRowStartsAtIndex` and replace `4` with our constant.

zackkrida (approved):
This works so well :tada: I made one suggestion for code quality, but functionality-wise this is awesome.

frontend/src/locales/scripts/en.json5
zackkrida: I think you are right, "show" everywhere, in the strings and keys, is probably the best solution.

frontend/src/components/VMediaInfo/VMediaTags.vue
zackkrida: > Probably due to the maxDiffPixels setting?

Must be,  as it's set to `2px`. 

To clarify, my suggestion was that I didn't see a visual difference between having the negative vertical margin on the wrapper div and vertical padding on the ul vs. having no vertical margin at all. It could be something weird with rendering on my machine, though! I see the focus state the same both ways.

frontend/src/components/VMediaInfo/VMediaTags.vue
obulat: Drafting this PR to add these changes

zackkrida (changes requested):
@obulat I've drafted this so we can make the changes discussed in Slack:

- [ ] Add a `TOGGLE_TAG_EXPANSION` (or similar name) analytics event
- [ ] Finalize the language

frontend/src/components/VMediaInfo/VMediaTags.vue
```diff
@@ -38,21 +85,159 @@ export default defineComponent({
...
+      /**
+       * Find the index of the first item after the third row of tags. This is used
+       * to determine which tags to hide.
+       */
+      if (tagsContainerRef.value) {
+        collapsibleRowsStartAt.value = findRowStartsAt(tagsContainerRef.value)
+      }
+    })
+
+    const { sendCustomEvent } = useAnalytics()
```
zackkrida, line 165: Nit: `const { $sendCustomEvent } = useContext()` would be preferable.

frontend/src/locales/scripts/en.json5
```diff
@@ -498,6 +498,11 @@
     },
     imageInfo: "Image information",
     audioInfo: "Audio information",
+    tags: {
+      title: "Tags",
+      showMore: "Show more",
+      showLess: "Show less",
```
dhruvkb, line 504: Grammar nit: While "Show less" feels right too, I'm feeling a little bit of a dissonance that says this should be "Show fewer" because tags are countable.

frontend/src/utils/focus-management.ts
```diff
@@ -88,8 +88,11 @@ export function isFocusableElement(element: HTMLElement) {
...
+  element.focus()
```
dhruvkb, line 95: `!(!c1 || !c2)` can be simplified to `c1 && c2`. Without the double negative, the condition is more readable. It can even be simplified further because `null` and `undefined` aren't instances of `HTMLElement`.

```suggestion
  if (element instanceof HTMLElement) {
    element.focus()
  }
```

frontend/src/components/VMediaInfo/VMediaTags.vue
```diff
@@ -38,21 +83,158 @@ export default defineComponent({
...
 
-    return { additionalSearchViews, localizedTagPath }
+    const heightClass = computed(() => {
+      if (!hasOverflow.value) {
+        return "max-h-none"
+      }
+      /**
+       * Height is 3 rows of tags, gaps, and a padding for the focus rings.
+       */
+      return buttonStatus.value === "show" ? "max-h-[7.6875rem]" : "mah-h-none"
```
dhruvkb, line 193: A bit more explanation about the calculation of this number with the measurements of the components (tags, gaps, paddings etc.) would be nice.

dhruvkb (approved):
LGTM! I have a few suggestions but I will not block this PR.

frontend/src/locales/scripts/en.json5
obulat: That's "show more content" and "show less content" 😆

Issue #1787, opened 2021-04-21
Archive TSVs saved on S3 at lower (cheaper) access level (original #376)


obulat:
This issue has been migrated from the [CC Search Catalog repository](https://github.com/creativecommons/cccatalog/issues/376)

```
Author: mathemancer
Date: Tue Apr 28 2020
Labels: ✨ goal: improvement,🙅 status: discontinued
```
## Current Situation

We now save the TSV resulting from each Provider API Script run in S3, in a Standard Access Object Storage Class bucket.

## Suggested Improvement

After loading the data into the DB, we should archive in a Standard-IA (Infrequent Access) class bucket.

This should be done via a node in the `loader_workflow` Apache Airflow DAG (source file at `src/cc_catalog_airflow/loader_workflow.py`)

## Benefit

We don't currently need to re-access these files very much (or at all under normal circumstances).  Thus, we could save money by putting them in a lower-access tier storage bucket.

sarayourfriend:
Implementing this would cut our catalog data storage costs by at least half, if not more, depending on whether we consider the data "recreatable". There are two infrequent access classes, a multi-zone option and a single-zone option. The multi-zone option is only 0.00625 USD more expensive per GB than the single-zone option. Single-zone is only useful if we consider the data recreatable, because there's no disaster redundancy if the AZ goes out. I think that would be a safe option, but cannot make that call myself, so asking @WordPress/openverse-catalog to give input there.

There are two stages to implement this:
1. The easiest part, updating the storage class of the object when we are finished with it
2. The harder part, backfilling the storage class of previous objects

The harder part might actually not be harder, depending on how the objects are used. For example, if it's safe to do a one-time conversion of every existing object in the bucket to the new storage class then it's very easy, and can be done through the AWS CLI:

```
aws s3 cp s3://<bucket-name>/ s3://<bucket-name>/ --recursive --storage-class <storage_class>
```

The relevant bucket does not have versioning enabled, so this is a safe way to change the storage class. If versioning was enabled, this would just create a new version with the new storage class, but the previous version would still exist with the standard storage class and we'd be paying double.

[...]

PR #3810, opened 2024-02-20
Use infrequent access when uploading provider TSVs

sarayourfriend:
## Fixes

Related to #1787 by @obulat 

This does not close the issue, as we still need to run the s3 command noted in the issue to backfill the storage class onto all existing objects.

## Description

Updates the provider ingestion workflow DAG to use the standard infrequent access storage class for provider TSVs.

I've documented the rationale behind setting the storage class eagerly, despite immediately accessing it, in the code. I really do not think it makes any sense at all to use a new task to set the storage class down to IA after the DAG is finished, and despite the 10th of a US penny spent on the retrieval fee, we will still save >99.99% of the money we. I really think the difference is incalculable and not worth the code complexity. The note about increased energy usage of an additional task (and all the requests it would need to send) helps further justify the fact, even though it doesn't save on our AWS bill.

The configuration is stored on the provider workflow, to allow for theoretical flexibility in setting different storage classes for different providers. We could use single-zone storage for non-dated DAGs, for example, as we always recreate the data 100% of the time on every DAG run, so it wouldn't present an issue if we needed to fully reload that provider's data if the AZ holding the data failed or went temporarily offline.

This does not close the issue, as we still need to run the s3 command noted in the issue to backfill the storage class onto all existing objects.

I also did not bother changing the storage class for staging objects because (a) the loader workflow DAG is retired and (b) those are immediately deleted when the DAG is successfully anyway.

As noted in `env.template` and the provider ingestion configuration data class, minio does not support the AWS storage classes aside from `STANDARD` (the default). Therefore, our local environments need to use a different non-default value. The testing instructions below describe how to verify the behaviour.

## Testing Instructions

There are two things to verify in this PR:
1. That the non-default storage class is used by the DAG
2. That the local environment is able to pass by using a non-standard storage class compatible with Minio.

For both of these, we'll run the wikimedia commons ingestion workflow, as it's the easiest to work with. For all these steps, we will follow this basic pattern:

1. Run the Wikimedia DAG
2. Allow the `load_mixed` task to pull data until it logs `x records retrieved` (usually 160)
3. Mark `load_mixed` as successful
4. Observe the behaviour of `copy_to_s3` (either success or failure, depending on the cases outlined below)
5. Observe the final result of the DAG (success or failure, as above)
6. In cases where we expect a success of the local run, confirm the value in Minio

Log in to your local minio at http://0.0.0.0:5011/ using `test_key` and `test_secret` as the username and password respectively.

First, test the behaviour on main, so that you understand what the existing behaviour here is. In short, because the storage class is not specified in our API call to s3, the default is used. For both minio and AWS, this is the "STANDARD" storage class. Run the Wikimedia DAG on `main` and confirm in minio that the uploaded TSV has the "STANDARD" storage class. Minio does not explicitly list the storage class of an object if it is the default. This is reliable behaviour, as you will see later. If you absolutely want to confirm the storage class of the object, download the "inspect" zip (disable encryption) and open the `xl.meta` file for the TSV inside the zip, and confirm that the string `x-amz-storage-classšSTANDARD` is present in the file.

Screenshot showing standard storage class in the meta.xl file for the uploaded TSV

Now that you understand the behaviour on main, checkout the branch, and copy `env.template` to `.env` OR (if you don't want to override your `.env` entirely) copy the new `AIRFLOW_VAR_DEFAULT_S3_TSV_STORAGE_CLASS` variable into your .env. Run `just dup catalog` to get the new environment variable in.

Follow the steps above to run the Wikimedia DAG to completion, and expect a successful run. Navigate to the TSV in minio, click on it, and then scroll to the bottom of the sidebar to view the "Metadata" section, and confirm you see `REDUCED_REDUNDANCY` as the storage class:

Minio dashboard showing non-default storage class

This confirms that the DAG is sending the storage class with the load_file operation as expected. However, we must confirm that the `STANDARD_IA`, set as default in the code, is used when the Airflow variable isn't present. The variable is currently only meant for testing, though it could be used in the future to avoid a code change if we need to temporarily set the storage class to something else. However, we must ensure that the DAG uses `STANDARD_IA` by default.

To do this, comment out `AIRFLOW_VAR_DEFAULT_S3_TSV_STORAGE_CLASS` from your catalog .env file, and run `just dup catalog` to refresh the environment variable. At this point, if you ran the Wikimedia DAG you would see `copy_to_s3` fail on an error of invalid storage class. This is because Minio does not support `STANDARD_IA` and returns an error during upload. However, it does not specify which storage class is used. You may do so by adding the following code to the top of `copy_file_to_s3`:

```py
import json
logger.error(json.dumps(extra_args))
```

That way, when you run the DAG, the logs for `copy_to_s3` will show the storage class passed to the hook so you can confirm it is `STANDARD_IA` instead of some other value.

In this case, there will be nothing to confirm in Minio, as it will have rejected the upload due to the invalid storage class.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/providers/provider_workflows.py
+++ b/catalog/dags/providers/provider_workflows.py
@@ -166,6 +166,25 @@ class ProviderWorkflow:
     tags: list[str] = field(default_factory=list)
     overrides: list[TaskOverride] = field(default_factory=list)
 
+    # Set when the object is uploaded, even though we access the object later in
+    # the DAG. IA incurs additional retrieval fees per request, unlike plain
+    # standard storage. However, as of writing, that costs 0.001 USD (1/10th of
+    # a US cent) per 1k requests. In other words, a minuscule amount, considering
+    # we will access the object once later in the DAG, to upsert it to the DB,
+    # and then in all likelihood never access it again.
+    # Even if we did, and had to pay the retrieval fee, we would still come out
+    # ahead on storage costs, because IA is so much less expensive than regular
+    # storage. We could set the storage class in a later task in the DAG, to
+    # avoid the one time retrieval fee. However, that adds complexity to the DAG
+    # that we can avoid by eagerly setting the storage class early, and the actual
+    # savings would probably be nil, factoring in the time spent in standard storage
+    # incurring standard storage costs. If it absolutely needs to be rationalised,
+    # consider the amount of energy spent on the extra request to S3 to update the
+    # storage cost to try to get around a retrieval fee (which, again, will not
+    # actually cost more, all things considered). Saving that energy could melt
+    # the glaciers all that much more slowly.
+    s3_tsv_storage_class: str = "STANDARD_IA"
+
     def _get_module_info(self):
         # Get the module the ProviderDataIngester was defined in
         provider_script = inspect.getmodule(self.ingester_class)
@@ -186,11 +205,29 @@ class ProviderWorkflow:
         if not self.doc_md:
             self.doc_md = provider_script.__doc__
 
-        # Check for custom configuration overrides, which will be applied when
-        # the DAG is generated.
+        self._process_configuration_overrides()
+
+    def _process_configuration_overrides(self):
+        """
+        Check for and apply custom configuration overrides.
+
+        These are only applied when the DAG is generated.
+        """
+
+        # Provider-specific configuration overrides
         self.overrides = Variable.get(
             "CONFIGURATION_OVERRIDES", default_var={}, deserialize_json=True
         ).get(self.dag_id, [])
+
+        # Allow forcing the default to something other than `STANDARD_IA`
+        # Primarily meant for use in local development where minio is used
+        # which does not support all AWS storage classes
+        # https://github.com/minio/minio/issues/5469
+        # This intentionally applies to all providers, rather than the provider-specific
+        # overrides above
+        self.s3_tsv_storage_class = Variable.get(
+            "DEFAULT_S3_TSV_STORAGE_CLASS", default_var=self.s3_tsv_storage_class
+        )
 
 
 PROVIDER_WORKFLOWS = [
--- a/catalog/env.template
+++ b/catalog/env.template
@@ -68,7 +68,7 @@ S3_LOCAL_ENDPOINT=http://s3:5000
 # value so that the expected behaviour can be verified (specifically, that the storage
 # class is not the default "STANDARD")
 # https://github.com/minio/minio/issues/5469
-AIRFLOW_VAR_DEFAULT_S3_TSV_STORAGE_CLASS="REDUCED_REDUNDANCY"
+AIRFLOW_VAR_DEFAULT_S3_TSV_STORAGE_CLASS=REDUCED_REDUNDANCY
 
 # Connection to the Ingestion Server, used for managing data refreshes. Default is used to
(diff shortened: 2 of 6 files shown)


catalog/dags/providers/provider_workflows.py
```diff
@@ -166,6 +166,25 @@ class ProviderWorkflow:
...
+    # ahead on storage costs, because IA is so much less expensive than regular
+    # storage. We could set the storage class in a later task in the DAG, to
+    # avoid the one time retrieval fee. However, that adds complexity to the DAG
+    # that we can avoid by eagerly setting the storage class early, and the actual
+    # savings would probably be nil, factoring in the time spent in standard storage
+    # incurring standard storage costs. If it absolutely needs to be rationalised,
+    # consider the amount of energy spent on the extra request to S3 to update the
+    # storage cost to try to get around a retrieval fee (which, again, will not
+    # actually cost more, all things considered). Saving that energy could melt
+    # the glaciers all that much more slowly.
```
dhruvkb, line 185: Chuckled at this! While I feel odd about having such a large, informal comment in the code, thanks for thinking about the glaciers.

justfile
```diff
@@ -200,6 +200,10 @@ init:
 down *flags:
     just dc down {{ flags }}
 
+# Take all services down then call the specified app's up recipe. ex.: `just dup catalog` is useful for restarting the catalog with new environment variables
```
dhruvkb, line 203: A shorter comment would help as this is printed as the help output. The example doesn't add much and can be left out, if that helps.

dhruvkb: Also what does "dup" mean, is is an abbreviation for something? I don't get it.

dhruvkb (approved):
LGTM, I have some non-blocking comments.

catalog/dags/providers/provider_workflows.py
sarayourfriend: I wasn't sure whether to leave this in the code or just in the PR description, for basically the reason of it being informal and more about the decision rather than the implementation.

What do you think? I'm comfortable removing it if it isn't helpful to have as in-code documentation.

sarayourfriend:
Updating the tests for the new parameter now

catalog/env.template
```diff
@@ -63,6 +63,12 @@ AIRFLOW_CONN_SLACK_NOTIFICATIONS=https://slack
 AIRFLOW_CONN_SLACK_ALERTS=https://slack
 
 S3_LOCAL_ENDPOINT=http://s3:5000
+# Set to a non-default value supported by minio in local development to workaround
+# Minio's lack of support for all AWS storage classes, while still using a non-default
+# value so that the expected behaviour can be verified (specifically, that the storage
+# class is not the default "STANDARD")
+# https://github.com/minio/minio/issues/5469
+AIRFLOW_VAR_DEFAULT_S3_TSV_STORAGE_CLASS="REDUCED_REDUNDANCY"
```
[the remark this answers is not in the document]
sarayourfriend: Is there a rationale behind when to quote and not to quote? I used the quotes because the first other variable I checked was quoted, but if there is a consistent way to do it that always works exactly as expected, it would be good to prioritse that one.

PR #3811, opened 2024-02-20
Turn analytics composable into a plugin

sarayourfriend:
## Fixes

A pre-requisite for https://github.com/WordPress/openverse/pull/3721

## Description

Convert the `useAnalytics` composable into a plugin. I've chosen to take an approach where the composable still exists, but just acts as a thin wrapper around `$sendCustomEvent` from the new plugin. This (a) reduces code churn and (b) allows us to confirm that the change behaves exactly as we expected the previous version to behave. All tests asserting against the analytics events continue to work without changes.

## Testing Instructions

All Playwright and unit tests should continue to pass without changes.

Furthermore, to confirm absolutely that the `breakpoint`, `height` and `width` variables still work as expected, run Plausible locally using `just frontend/up` followed by `just frontend/init` to get the custom events loaded. Then visit frontend through the Nginx container at http://0.0.0.0:50290

Perform a search and confirm that the three default props (listed above) match the expected values based on the viewport of your browser. Make multiple searches with different viewport sizes, especially making sure to trigger different breakpoints, and confirm that those events have the correct values.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/composables/use-analytics.ts
+++ b/frontend/src/composables/use-analytics.ts
@@ -1,5 +1,11 @@
 import { useContext } from "@nuxtjs/composition-api"
 
+/**
+ * This wrapper around the plugin, retained to reduce code churn.
+ * @see Refer to frontend/src/plugins/analytics.ts for plugin implementation
+ *
+ * @deprecated For new code, use `$sendCustomEvent` from Nuxt context
+ */
 export const useAnalytics = () => {
   const { $sendCustomEvent } = useContext()
 
--- a/frontend/src/composables/use-analytics.ts
+++ b/frontend/src/composables/use-analytics.ts
@@ -1,68 +1,7 @@
-import { computed, onMounted } from "vue"
 import { useContext } from "@nuxtjs/composition-api"
 
-import type { Events, EventName } from "~/types/analytics"
-import { useUiStore } from "~/stores/ui"
-import { useFeatureFlagStore } from "~/stores/feature-flag"
-
-import { log } from "~/utils/console"
-
-/**
- * The `ctx` parameter must be supplied if using this composable outside the
- * bounds of the composition API.
- */
 export const useAnalytics = () => {
-  const { $plausible } = useContext()
-  const uiStore = useUiStore()
-  const featureFlagStore = useFeatureFlagStore()
+  const { $sendCustomEvent } = useContext()
 
-  onMounted(() => {
-    featureFlagStore.syncAnalyticsWithLocalStorage()
-  })
-
-  /**
-   * the Plausible props that work identically on the server-side and the
-   * client-side; This excludes props that need `window`.
-   */
-  const isomorphicProps = computed(() => ({
-    breakpoint: uiStore.breakpoint,
-  }))
-
-  /**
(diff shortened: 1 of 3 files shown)


frontend/src/composables/use-analytics.ts
```diff
@@ -1,68 +1,7 @@
...
-    log(`Analytics event: ${name}`, payload)
-    $plausible.trackEvent(name, {
-      props: {
-        ...isomorphicProps.value,
-        ...windowProps.value,
-        ...payload,
-      },
-    })
-  }
+  const { $sendCustomEvent } = useContext()
```
obulat, line 4: I don't think this will work when called by the store or the media service, outside of the `setup` function.  Currently, the places where I think this could break is the other composables: `use-sensitive-media` and `use-search`.
It would probably also break the search response time and error handling events in the other open PRs.

sarayourfriend: Why would it break in other composables? They also have access to nuxt context, no? Do you have a reproduction of it breaking somewhere?

obulat: I think I tried using the "Centralise frontend errors" (which calls analytics from store's media fetching, if I remember correctly) and the "Search response times" (which cals analytics from a data service), and they failed there. We currently don't call analytics from the server or from data services, that's why tests should not be affected at all.

obulat: @sarayourfriend, I tried re-adding the "Search response time" event (https://github.com/WordPress/openverse/tree/fix/analytics-plugin-add-search-response-time), and it does work with this code. I did have to mock the `useContext` calls in the unit tests, so maybe we should also add an e2e test to make sure it works. I'm going to approve this PR, glad that it took so few changes!

obulat (approved):

frontend/src/composables/use-analytics.ts
```diff
@@ -1,68 +1,7 @@
...
-import { useUiStore } from "~/stores/ui"
-import { useFeatureFlagStore } from "~/stores/feature-flag"
-
-import { log } from "~/utils/console"
-
-/**
- * The `ctx` parameter must be supplied if using this composable outside the
- * bounds of the composition API.
- */
 export const useAnalytics = () => {
```
zackkrida, line 3:
```suggestion
/**
 * This function is a thin wrapper around a plugin,
 * used to preserve tests and behavior from a prior
 * implementation. See `frontend/src/plugins/analytics.ts`
 * for the full analytics logic.
 */
export const useAnalytics = () => {
```

Nit: Some context might be useful here for future readers. Feel free to write your own copy, just a suggestion.

zackkrida (approved):
Works great! All of my local breakpoint, height, and width values are correct.

frontend/src/composables/use-analytics.ts
sarayourfriend: Oh, yes definitely, thanks for the suggested docs.

Issue #2037, opened 2023-05-05
Move Openverse API and catalog to `openverse.org` subdomains


sarayourfriend:
## Summary

Move the infrastructure which currently exists on `openverse.engineering` to `openverse.org`.

## Description

<details>
<summary>Pseudo-project proposal (collapsed to preference the issue and document list)</summary>

We currently pay for two Cloudflare accounts, one for `openverse.engineering` and `openverse.org`. If we move the API and catalog to live on subdomains of `openverse.org`, we should be able to change our `openverse.engineering` account to a free one, saving 200 USD a month that can go towards e.g., Plausible.

In the initial discussion we had about this we only talked about moving the API. However, because Airflow is behind Cloudflare Access, we may need to move it as well. I've tried to understand whether that is the case based on the Cloudflare pricing and it seems like Cloudflare Access might be free under our current usage but it isn't clear to me whether that's free _for specific paid account types_ or free for _any_ account type.

The end result of this project should be that `openverse.engineering` and its usage should be entirely covered by a free Cloudflare account. The planning for this project must consider the various features we will need that will cause contingencies here:

1. Do we need to route redirects through Cloudflare, or could we do it AWS side with a small redirect service? If we do it via Cloudflare, we might need individual page rules for each domain we are redirecting. For the API there are at least two we need to redirect: `api.openverse.engineering` and `api-production.openverse.engineering`. Do we need to redirect staging (`api-staging`)? What about the legacy staging subdomain, `api-dev`? We also continue to redirect `search.openverse.engineering`, our legacy frontend domain(s). Cloudflare free only supports three page rules. We currently have 3 redirects already (for the frontend). We would need at minimum one more for the API (`api.openverse.engineering`) but to match our existing redirect philosophy with the frontend, we should redirect `api`, `api-production` and `api-staging`. That would lead to 6 minimum page rules if we used page rules for the redirects. The other existing page rules are caching related and would be moved into `openverse.org`, so they do not need to count towards our total page rule utilisation.
2. Can the catalog stay on `openverse.engineering` if it is a free account? Namely, as discussed above, do free Cloudflare accounts support Cloudflare Access? If we manage Cloudflare Access on two different Cloudflare zones, the access terraform module will need to be updated or potentially completely re-written to accommodate multiple Cloudflare zones.

Additional considerations:

- The jumphost is on `openverse.engineering`. Should is stay there?
[...]

PR #3828, opened 2024-02-26
Make User-Agent a default header

### LINKED ISSUE #1362: Include UA string on every request made by the DelayedRequester

krysal:
## Fixes

Fixes #1362 by @krysal
Related to #2037 ([Airflow work](https://github.com/WordPress/openverse-infrastructure/milestone/6))

## Description

Resuming the work done previous to the monorepo. This PR adds the User-Agent as the default header for all requests in the `DelayedRequester` and `ProviderDataIngester` classes, making it unnecessary to repeat code in providers DAGs, so it was removed from museum_victoria, nappy, rawpixel, stocksnap, and wikimedia_commons.

I added a new `CANONICAL_ORIGIN` reading it from the os environment to replace the old domain referenced in the UA string. Is it better to read it from an Airflow variable? CC @AetherUnbound.

## Testing Instructions

Run one of the above-mentioned DAGs, for example rawpixel, to confirm they continue to work. See the code and updated tests and confirm it all makes sense.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/loader/provider_details.py
+++ b/catalog/dags/common/loader/provider_details.py
@@ -124,7 +124,8 @@ SMITHSONIAN_SUB_PROVIDERS = {
 
 # User-Agent header for APIs that require it
 CONTACT_EMAIL = os.getenv("CONTACT_EMAIL")
-UA_STRING = f"Openverse/0.1 (https://wordpress.org/openverse; {CONTACT_EMAIL})"
+CANONICAL_ORIGIN = os.getenv("CANONICAL_ORIGIN", "https://openverse.org")
+UA_STRING = f"Openverse/0.1 ({CANONICAL_ORIGIN}; {CONTACT_EMAIL})"
 
 
 # Available Image Categories for API
--- a/catalog/dags/common/loader/provider_details.py
+++ b/catalog/dags/common/loader/provider_details.py
@@ -124,7 +124,12 @@ SMITHSONIAN_SUB_PROVIDERS = {
 
 # User-Agent header for APIs that require it
 CONTACT_EMAIL = os.getenv("CONTACT_EMAIL")
-CANONICAL_ORIGIN = os.getenv("CANONICAL_ORIGIN", "https://openverse.org")
+
+CANONICAL_DOMAIN: str = os.getenv("CANONICAL_DOMAIN", "openverse.org")
+
+_proto = "http" if "localhost" in CANONICAL_DOMAIN else "https"
+CANONICAL_ORIGIN: str = f"{_proto}://{CANONICAL_DOMAIN}"
+
 UA_STRING = f"Openverse/0.1 ({CANONICAL_ORIGIN}; {CONTACT_EMAIL})"
 
 
--- a/catalog/dags/common/requester.py
+++ b/catalog/dags/common/requester.py
@@ -7,6 +7,7 @@ from airflow.exceptions import AirflowException
 from requests.exceptions import JSONDecodeError
 
 import oauth2
+from common.loader import provider_details as prov
 
 
 # pytest_socket will not be available in production, so we must create a shim for
@@ -45,9 +46,10 @@ class DelayedRequester:
              by kwargs in specific calls to the get method
     """
 
-    def __init__(self, delay=0, headers=None):
+    def __init__(self, delay: int = 0, headers: dict | None = None):
+        headers = {} if headers is None else headers
         self._DELAY = delay
-        self.headers = headers or {}
+        self.headers = {"User-Agent": prov.UA_STRING} | headers
         self._last_request = 0
         self.session = requests.Session()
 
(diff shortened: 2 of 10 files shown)


catalog/dags/common/loader/provider_details.py
```diff
@@ -124,7 +124,8 @@
 
 # User-Agent header for APIs that require it
 CONTACT_EMAIL = os.getenv("CONTACT_EMAIL")
-UA_STRING = f"Openverse/0.1 (https://wordpress.org/openverse; {CONTACT_EMAIL})"
+CANONICAL_ORIGIN = os.getenv("CANONICAL_ORIGIN", "https://openverse.org")
```
sarayourfriend, line 127: I guess if we want to match the API's new variables, it should be an environment variable of `CANONICAL_DOMAIN` with origin derived from it, but I don't know if that was a goal with this PR. It doesn't matter to me either way, as there's so little overlap between the variables for each I don't think we need to be concerned with having them match _except_ when the names are this similar, it would be easier to make a quick mistake.

To clarify, not a suggestion for a change, just noting the slight difference, will at least help me keep it in mind during infra work to migrate Airflow's environment variables to the new approach.

sarayourfriend (approved):
LGTM!

catalog/dags/common/requester.py
```diff
@@ -42,12 +43,13 @@ class DelayedRequester:
...
-             by kwargs in specific calls to the get method
+             by kwargs in specific calls to the `get` method
     """
 
-    def __init__(self, delay=0, headers=None):
+    def __init__(self, delay: int = 0, headers: dict | None = None):
+        headers = {} if headers is None else headers
         self._DELAY = delay
-        self.headers = headers or {}
+        self.headers = {"User-Agent": prov.UA_STRING} | headers
```
stacimc, line 52: Why do we default here in _addition_ to adding this as a default in the ProviderDataIngester?

krysal: That was the suggestion of the issue, and I went that route first, but it turns out it wasn't enough to set the default header. I leave it after updating the `ProviderDataIngester` as it will be better if each request from the Calalog goes with this header.

catalog/dags/common/loader/provider_details.py
krysal: Yes, I intended to make the change aligned with your on moving the domain :) I thought the `CANONICAL_DOMAIN` was a variable we could skip here, but you're right. It would be better to match the API's variables in this case.

Done ✅

catalog/dags/common/loader/provider_details.py
```diff
@@ -124,7 +124,13 @@
 
 # User-Agent header for APIs that require it
 CONTACT_EMAIL = os.getenv("CONTACT_EMAIL")
-UA_STRING = f"Openverse/0.1 (https://wordpress.org/openverse; {CONTACT_EMAIL})"
+
+CANONICAL_DOMAIN: str = os.getenv("CANONICAL_DOMAIN", "openverse.org")
+
+_proto = "http" if "localhost" in CANONICAL_DOMAIN else "https"
+CANONICAL_ORIGIN: str = f"{_proto}://{CANONICAL_DOMAIN}"
```
stacimc, line 131: Is this just for consistency with the api? Would we ever actually use a domain with "localhost" in this context?

catalog/dags/common/requester.py
stacimc: 😮 That's so strange -- do you mean setting it as the default in `DelayedRequester` did not cover all the existing cases? Are we making requests that don't use the requester somewhere?

If it _can_ be added in only one place, I do agree about adding it in the requester instead of the ProviderDataIngester, since that gets complicated with subclasses that override `__init__` etc.

stacimc (approved):
Couple of questions but nothing blocking. LGTM and all DAGs work as expected 🚀

catalog/dags/common/requester.py
krysal: > do you mean setting it as the default in DelayedRequester did not cover all the existing cases? 

Yes, the thing is the `ProviderDataIngester` class passes each time its headers to the `DelayedRequester.get_response_json` method (for the possibility of overriding them) so it wasn't enough to add them to `DelayedRequester` only.

> Are we making requests that don't use the requester somewhere?

It's a possibility, I haven't checked but this isn't the issue here, as I explained above. We should be safe with these changes!

catalog/dags/common/loader/provider_details.py
krysal: Yes, this arose from my conversation [with Sara here](https://github.com/WordPress/openverse/pull/3828#discussion_r1503265924). We opted for keeping consistency with the API in any case.

PR #3836, opened 2024-02-27
Add accesstoken and ThrottledApplication to admin panel

### LINKED ISSUE #3711: Add access token and throttled application models to Django admin
## Problem

When investigating traffic patterns and addressing malicious API usage, it can be useful to use the Django Admin web UI to check the status of an access token and/or client application.

## Description

We should add the `oauth_provider.models.AccessToken` and `api.models.oauth.ThrottledApplication` tables to the Django Admin UI so that these records can be easily accessed and searched against from the web UI.

https://github.com/WordPress/openverse/blob/098925afa90b26881b4b4d3e2baffa69049692c0/api/api/models/oauth.py#L27

Those models will need their own views added to `admin.__init__`:

https://github.com/WordPress/openverse/blob/5e5a2aa376048391a8455ae3063b4f3e7afe0d0d/api/api/admin/__init__.py

While I initially thought this would be good as read-only, we might even want to enable editing of certain fields so that access can be revoked by maintainers using the Admin UI as well.

bjornthiberg:
## Fixes

Fixes #3711 by @AetherUnbound

## Description

This PR adds `oauth_provider.models.AccessToken` and `api.models.oauth.ThrottledApplication` tables to the Django Admin UI.

All fields are set as read-only. We are not sure if this is done in the best way. Also, in the issue (#3711), it was said that: 

> While I initially thought this would be good as read-only, we might even want to enable editing of certain fields so that access can be revoked by maintainers using the Admin UI as well.

We did not know which fields should be set as editable. Hence the draft PR.

## Testing Instructions
Test by opening the Django admin panel and looking at the new entries in the sidebar.

## Checklist

- [ ] ~~I added or updated tests for the changes I made (if applicable).~~
- [ ] ~~I added or updated documentation (if applicable).~~
- [ ] ~~I ran the DAG documentation generator (if applicable).~~

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Dhruv and I are at WordCamp Asia next week, so pinging @WordPress/openverse-api to reassign reviews on this to folks who are available.

Regarding the editable fields, please make `rate_limit_model` editable. It's probably also a good one to add to search fields, so we can easily see client IDs in each rate limit model.

I'm not 100% sure how revoking access would work. Deleting the ThrottledApplication instance doesn't seem like the right way to go (just thinking of that being an option), but if we had a new rate limit model called `"deny"`, or something along those lines, that could work. From my perspective this is something for a separate issue, though, as it needs a bit of thought from @WordPress/openverse-maintainers to decide what the technical and decision making process would be.

zackkrida:
Hi @bjornthiberg, Openverse maintainers are going to take a look at your PR this week! Thank you for the contribution :tada: 

I did want to make sure you saw the [notes from @sarayourfriend](https://github.com/WordPress/openverse/pull/3836#issuecomment-1973900832) concerning editable fields. Thank you!

api/api/admin/__init__.py
```diff
@@ -83,3 +86,54 @@ class ProviderAdmin(admin.ModelAdmin):
...
     ordering = ("media_type", "provider_name")
+
+
+@admin.register(ThrottledApplication)
+class ThrottledApplicationAdmin(admin.ModelAdmin):
+    search_fields = ("client_id", "name")
+    list_display = ("client_id", "name", "user", "rate_limit_model")
+    ordering = ("client_id",)
+
+    def get_readonly_fields(self, request, obj=None):
```
AetherUnbound, line 97: For both this and the class below - since we're not actually doing any modifications of the read only field list itself, we can use the [`readonly_fields` property](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields) instead of having to define a `get_readonly_fields` function. Do you mind changing these to that pattern instead? Here's an example:

https://github.com/WordPress/openverse/blob/57223c279408fe2372503a052f5b179fc9446aa8/api/api/admin/__init__.py#L70

AetherUnbound (changes requested):
One code request change, in addition to Sara's note about editable fields above.

api/api/admin/__init__.py
```diff
@@ -83,3 +86,54 @@ class ProviderAdmin(admin.ModelAdmin):
     list_display = ("provider_name", "provider_identifier", "media_type")
     search_fields = ("provider_name", "provider_identifier")
     ordering = ("media_type", "provider_name")
+
+
+@admin.register(ThrottledApplication)
+class ThrottledApplicationAdmin(admin.ModelAdmin):
+    search_fields = ("client_id", "name")
+    list_display = ("client_id", "name", "user", "rate_limit_model")
+    ordering = ("client_id",)
```
krysal, line 95: The `user` field isn't very useful here, but the date the app was created is! It will also be nice to sort it to show first the most recently created apps. The client_id is quite random. Here is my suggestion of how the code should look:
```suggestion
    list_display = ("client_id", "name", "created", "rate_limit_model")
    ordering = ("-created",)
```

krysal (commented):
Thank you for starting this @bjornthiberg. I'm adding just a quick comment to the previous reviews.

AetherUnbound (approved):
This looks great and works well locally, thanks for addressing all the changes!

PR #3845, opened 2024-02-28
Preemptively fail unparsable/unusable Europeana URLs

### LINKED ISSUE #2784: Europeana data ingester can return `None` in URL
## Description

Europeana ingestion workflow failed because a `None` was saved in the URL field, and then the `MediaStore` class tried to extract the filetype from it.

## Reproduction

See the error in Airflow logs:
```
[2023-08-06, 01:11:48 UTC] {taskinstance.py:1824} ERROR - Task failed with exception
providers.provider_api_scripts.provider_data_ingester.IngestionError: 'NoneType' object has no attribute 'split'
query_params: {"wskey": "[redacted]", "profile": "rich", "reusability": ["open", "restricted"], "sort": ["europeana_id+desc", "timestamp_created+desc"], "rows": "100", "media": "true", "start": 1, "qf": ["TYPE:IMAGE", "provider_aggregation_edm_isShownBy:*"], "query": "timestamp_created:[2016-03-01T00:00:00Z TO 2016-03-02T00:00:00Z]", "cursor": "*"}
The above exception was the direct cause of the following exception:
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 181, in execute
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 198, in execute_callable
  File "/opt/airflow/catalog/dags/providers/factory_utils.py", line 55, in pull_media_wrapper
    data = ingester.ingest_records()
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 269, in ingest_records
    raise error from ingestion_error
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 234, in ingest_records
    self.record_count += self.process_batch(batch)
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 465, in process_batch
    store.add_item(**record)
  File "/opt/airflow/catalog/dags/common/storage/image.py", line 146, in add_item
    image = self._get_image(**image_data)
  File "/opt/airflow/catalog/dags/common/storage/image.py", line 153, in _get_image
    image_metadata = self.clean_media_metadata(**kwargs)
  File "/opt/airflow/catalog/dags/common/storage/media.py", line 145, in clean_media_metadata
    media_data["filetype"] = self._validate_filetype(
  File "/opt/airflow/catalog/dags/common/storage/media.py", line 315, in _validate_filetype
    filetype = extract_filetype(url, self.media_type)
  File "/opt/airflow/catalog/dags/common/extensions.py", line 9, in extract_filetype
    possible_filetype = url.split(".")[-1]
AttributeError: 'NoneType' object has no attribute 'split'
```

## Additional context

Europeana script uses a  [`raise_if_empty` decorator](https://github.com/WordPress/openverse/blob/61c7abdeaed0450558859ff949846f9a1497ad13/catalog/dags/providers/provider_api_scripts/europeana.py#L33) to ensure that only items with valid required data are saved.

AetherUnbound:
## Fixes

Fixes #2784 by @obulat

## Description

This PR modifies the Europeana ingester to preemptively fail URLs that are invalid or unusable. See #2784 for what prompted this. Because we automatically add the scheme if it isn't present when attempting to validate the URL, I didn't want to add any machinery to check if the scheme was missing from the beginning of the URL (due to the diversity of sources Europeana includes, I can't check all possible URLs we might get returned back). Dropbox URLs, however, can't really be used by our API as they can't be accessed directly. It seems Europeana actually proxies them itself, here's an example result: https://www.europeana.eu/en/item/2063627/BEL_280_003

This adds a check to throw out any Dropbox URLs ahead of time, so the CSV writing can occur as intended and the error is raised upstream.

## Testing Instructions

1. Check out this branch
2. Trigger the Europeana workflow with `additional_query_params` as `{"query": "timestamp_created:[2016-03-01T00:00:00Z TO 2016-03-02T00:00:00Z]"}`
3. Observe that the task completes successfully, although no records are ingested

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/providers/provider_api_scripts/europeana.py
+++ b/catalog/dags/providers/provider_api_scripts/europeana.py
@@ -97,7 +97,15 @@ class EuropeanaRecordBuilder:
     @raise_if_empty
     def _get_image_url(self, data: dict) -> str | None:
         group = data.get("edmIsShownBy")
-        return group[0] if group else None
+        if not group:
+            return None
+        url = group[0]
+        # Some Europeana URLs may have prefixes, or reference Dropbox (which we can't
+        # include in our catalog because we cannot access them directly ourselves).
+        # E.g.: L-APC248-https://www.dropbox.com/s/i1pqizm1joof8y1/Belgium_Diptyque%20_MAR-SGP-CO1.jpg?raw=1
+        if "dropbox.com" in url:
+            return None
+        return url
 
     @raise_if_empty
     def _get_foreign_identifier(self, data: dict) -> str | None:
--- a/catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
@@ -9,6 +9,7 @@ from common.licenses import LicenseInfo, get_license_info
 from common.loader import provider_details as prov
 from common.storage.image import ImageStore
 from providers.provider_api_scripts.europeana import (
+    EmptyRequiredFieldException,
     EuropeanaDataIngester,
     EuropeanaRecordBuilder,
 )
@@ -252,6 +253,19 @@ def test_get_foreign_landing_url_without_edmIsShownAt(record_builder):
     assert (
         record_builder.get_record_data(image_data)["foreign_landing_url"] == expect_url
     )
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        {},
+        {"edmIsShownBy": None},
+        {"edmIsShownBy": ["dropbox.com/value"]},
+    ],
+)
+def test_get_image_url_empty(data, record_builder):
+    with pytest.raises(EmptyRequiredFieldException):
+        assert record_builder._get_image_url(data)
 
 
 @pytest.mark.parametrize(


krysal (commented):
Good find on identifying the cause here! The solution looks valid as a last resort, but I feel compelled to ask: isn't there an other field in the API to take the proxied URL for the images?

AetherUnbound:
> isn't there an other field in the API to take the proxied URL for the images?

Ah, I should have mentioned! I did check, and for these results there are not 😅 They are _quite old_ it seems, which is probably why this is the case!

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -97,7 +97,15 @@ def get_record_data(self, data: dict) -> dict | None:
     @raise_if_empty
     def _get_image_url(self, data: dict) -> str | None:
         group = data.get("edmIsShownBy")
-        return group[0] if group else None
+        if not group:
+            return None
+        url = group[0]
+        # Some Europeana URLs may have prefixes, or reference Dropbox (which we can't
+        # include in our catalog because we cannot access them directly ourselves).
```
obulat, line 104: And this specific Dropbox link is actually broken 😆

obulat: Does this link actually contain the "L-APC248-" prefix? If so, maybe we need to validate the scheme here? Something like
```py
from urllib.parse import urlparse
result = urlparse(url)
if result.scheme not in ["http", "https"] or not result.netloc:
    return None
```

Not blocking because this doesn't happen often, IIRC.

obulat (approved):
Great catch! How did you find it?

krysal (approved):
Confirmed that these images are not available in Europeana either. Take this one for example:
- Response's [gist](https://gist.github.com/krysal/9adfa60a79ecc2aa6e406fe1958ddeac)
- Europena's landing page: https://www.europeana.eu/en/item/2063627/BEL_280_008
- Direct URL: https://www.dropbox.com/s/rpv24ftoz01e9s9/Belgium_MAR-BVM-1-384.jpg?raw=1
- Thumbnail: https://api.europeana.eu/thumbnail/v2/url.json?uri=https%3A%2F%2Fwww.dropbox.com%2Fs%2Frpv24ftoz01e9s9%2FBelgium_MAR-BVM-1-384.jpg%3Fraw%3D1&type=IMAGE

The thumbnail works but the other two URL don't show the image and it can't be downloaded. Given that you said that Europeana proxies them I thought they could be saved but that doesn't seem to be the case.

Thanks for fixing it!

stacimc (approved):
LGTM as well 🚀

catalog/dags/providers/provider_api_scripts/europeana.py
AetherUnbound: It does, unfortunately 😅 Quoting from the PR description here:

> Because we automatically add the scheme if it isn't present when attempting to validate the URL, I didn't want to add any machinery to check if the scheme was missing from the beginning of the URL (due to the diversity of sources Europeana includes, I can't check all possible URLs we might get returned back).

It's possible that some other links we receive from Europeana are valid but don't include a scheme. Since this was a rare case with a very specific cause, I'm hesitant to change any of the other Europeana machinery without understanding the impact it might have on ingesting all of the sources!

PR #3860, opened 2024-03-01
Revert "Update Node.js to v18.19.1"

sarayourfriend:
Reverts WordPress/openverse#3854

The deployment failed and I'm not able to fix forward right now. cc @AetherUnbound as MSR, @WordPress/openverse-frontend for looking into this failure. I'm at WordCamp next week and then AFK the week after, so won't be able to look into this myself, as just wrapping up for the week to get ready to travel.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -112,7 +112,7 @@
     "@testing-library/vue": "^5.8.2",
     "@types/express-useragent": "^1.0.2",
     "@types/jest": "^29.5.4",
-    "@types/node": "18.19.21",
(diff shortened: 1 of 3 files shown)


AetherUnbound (approved):

zackkrida:
Closing this PR as although the deploy was reported in GitHub as a failure, the latest version of the image _was_ successfully deployed. 

This seems to be an issue with our GitHub action, either a one-off flake-type issue or some other problem. The `Deploy Amazon ECS task definition` step took 10 minutes which is much longer than typical.

PR #3831, opened 2024-02-27
Combine related media component using `VMediaCollection.vue`

obulat:
## Fixes

Extracted from #3785 to make it easier to review.
Fixes #3832 by @obulat

_Note: 428 lines of the 987 added lines are from "tapes" (saved API responses used for tests)_

## Description

We currently have two separate components for related images and related audio that are basically duplicating each other. This PR combines the 2 related components.

A bigger change is the new `VMediaCollection` component (originally introduced in #3785, but split from it to make reviewing simpler).

This component handles displaying the header, grid skeleton if fetching or the media collection if fetched, and the footer with load more button and the external search form, if necessary.

To simplify this component, I used slots for header and footer, and created `wrapper` components (similar to the audio templates) for the related, search and collection results. This PR contains the `VMediaCollection` and the `VRelatedMedia.vue` wrapper component. It does not use footer because the related media does not have load more button (or the external search form).

This PR renames the "old" components  for audio and image grid to minimize the number of changes (`VImageGrid` and `VAudioList` are "old", and `VImageCollection` and `VAudioCollection` are the new ones).

There's also a tiny unrelated change: the loading skeleton for image single result is now centered (using `mx-auto` class).

## Testing Instructions

Select an audio result and click on one of the related audios. You should see the `SELECT_RESULT` event with the `relatedTo` set correctly. The related images/audio should look correct.

Look at the single result pages. The related media section should look and behave correctly.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


[...]

PR #3866, opened 2024-03-03
Use query parameters for additional search views in Nuxt

### LINKED ISSUE #3864: Replace Additional search views path parameters with query parameters
## Problem

Additional search views on the frontend should use the query parameters instead of the path parameters.

## Description

The following changes are necessary:

- replace `/pages/<mediaType>/tag/_tag.vue` and `/pages/<mediaType>/source/_.vue` pages with `/pages/<mediaType>/collection.vue`. The page should be similar to `_tag.vue`, using the `collection` middleware. It should not be setting the collection state in the `setup` function
- update `collection.ts` middleware to extract the collection parameters from the route name and query parameters. It should also check that the source name is valid, and should update the `search` store state.
- replace path-related functions in the search store to parse or set the collection query parameters.

## Additional context

Temporarily, until the API changes are added, we can use the current `full-text search` query parameters for the query: `tags` for tag view, `creator` and `source` for the creator and source. The `unstable__collection` parameter will be ignored by the API for now. 

The changes were discussed in [Rethinking additional search views URLs](https://github.com/WordPress/openverse/discussions/3825#top)

obulat:
## Fixes

Fixes #3864 by @obulat

## Description

This PR updates the frontend implementation to use the query parameters instead of path parameters:

- `/image/tag/cat` -> `/image/collection?tag=cat`
- `/image/source/wikimedia` -> `/image/collection?source=wikimedia`
- `/image/source/wikimedia/creator/me` -> `/image/collection?source=wikimedia&creator=me`

It updates pages, stores, collection middleware and API service.

### Pages
The two pages (`tag/_tag.vue` and `source/_vue`) with a single `collection.vue` page (which is duplicated for each media type). The parsing and setting of `collectionParams` was removed from the `setup` function and moved to the collection middleware.

### Store
`getCollectionPath` function is updated to create a path using `collection` path part and relevant query parameters.
This function is used for the source and creator link buttons, tags, the sources table.

### API requests
The store collection parameters are converted to an API request URL. For now, before the API changes are implemented, we can use the search without `q` parameter: the full-text search within tags, creator and source.
The following request URLS are used in this PR:
- `/images?unstable__collection=tag&tags=cat`
- `/images?unstable__collection=source&source=flickr`
- `/images?unstable__collection=creator&source=flickr&creator=me`

The API currently ignores the `unstable__collection` parameter.
When the API changes are implemented, `tags` will be replaced with `unstable__tag`, and from then on the real collections with exact matches and sorted by newest will be returned.

**Important limitation**
After these changes, the collection pages are not loaded on SSR because the `CollectionPage` sets the value of results before media is fetched, and server renders an empty result list. The follow-up PR that will use the new `VMediaCollection` component from #3831, and fix this problem.

## Testing Instructions

Make sure that `additional_search_views` is `on` on the `/preferences` page.
Navigate to tags, creator and source views from the single result pages. The client-side navigation should work correctly. _Note that there will be times that the page renders only the loading skeletons. This problem will be solved in a separate PR._

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/e2e/collections.spec.ts
```diff
@@ -16,42 +16,42 @@ test.describe("collections", () => {
...
+    await page.click('[href*="image/collection?tag="]')
 
-    await page.waitForURL(/image\/tag\/cat/)
+    await page.waitForURL(/image\/collection/)
 
     await expect(getH1(page, /cat/i)).toBeVisible()
-    await expect(getLoadMoreButton(page)).toBeEnabled()
-    expect(await page.locator("figure").count()).toEqual(20)
+    // await expect(getLoadMoreButton(page)).toBeEnabled()
+    // expect(await page.locator("figure").count()).toEqual(20)
```
zackkrida, line 32: Could you document why these assertions are commented?

frontend/src/stores/search.ts
```diff
@@ -266,8 +277,9 @@ export const useSearchStore = defineStore("search", {
       type: SupportedMediaType
       collectionParams: CollectionParams
     }) {
-      const path = `/${type}/${collectionToPath(collectionParams)}`
-      return this.$nuxt.localePath(path)
+      const path = `/${type}/collection`
+      const query = buildCollectionQuery(collectionParams, "frontend")
```
zackkrida, line 281:
```suggestion
      const { collection, ...params } = collectionParams
      const query = params
```

Just an example, this could be simpler and then `buildCollectionQuery` wouldn't need to support frontend and API modes, I think.

frontend/src/stores/search.ts
```diff
@@ -95,29 +107,33 @@ export function computeQueryParams(
...
-      return `source/${collectionParams.source}/creator/${collectionParams.creator}/`
-    }
-    case "source": {
-      return `source/${collectionParams.source}/`
-    }
+// TODO: After the API changes are done, replace
+// `tags` with `unstable__tag`
+export function buildCollectionQuery(
+  collectionParams: CollectionParams,
+  mode: "frontend" | "API"
```
zackkrida, line 120: Could you document this `mode` a bit to clarify specifically what it refers to? This is basically a helper that sometimes processes API urls and sometimes frontend urls? If the frontend version simply pulls out the params and returns them, maybe it doesn't need to use this helper at all. See [this example](https://github.com/WordPress/openverse/pull/3866/files#r1513515761).

frontend/src/pages/audio/collection.vue
zackkrida, line 1: Perhaps it's a bit too clever, but I think we could do a single `_media/collection.vue` page file that then passes the path as the media-type to `VCollectionPage` and throws an error if the route path is not one of our valid media types.

frontend/src/middleware/collection.ts
```diff
@@ -1,15 +1,95 @@
...
+  CollectionParams,
+  CreatorCollection,
+  SourceCollection,
+  TagCollection,
+} from "~/types/search"
+
+import type { Context, Middleware } from "@nuxt/types"
+import type { Dictionary } from "vue-router/types/router"
+
+const queryToCollectionParams = (
```
zackkrida, line 20: Separate conditionals might be easier to grok:

```
const queryToCollectionParams = (
  query: Dictionary<string | (string | null)[]>
): CollectionParams | undefined => {
  query = queryDictionaryToQueryParams(query);

  if ("tag" in query) {
    return {
      collection: "tag",
      tag: query.tag,
    } as TagCollection;
  }

  if ("creator" in query && "source" in query) {
    return {
      collection: "creator",
      creator: query.creator,
      source: query.source,
    } as CreatorCollection;
  }

  if ("source" in query) {
    return {
      collection: "source",
      source: query.source,
    } as SourceCollection;
  }

  return undefined;
};
```

frontend/src/data/api-service.ts
```diff
@@ -139,16 +138,14 @@ export const createApiService = ({
...
      * @param params - Url parameter object
      * @returns response  The API response object
      */
     query<T = unknown>(
       resource: string,
-      slug: string = "",
       params: Record<string, string> = {}
     ): Promise<AxiosResponse<T>> {
-      return client.get(`${getResourceSlug(resource)}${slug}`, { params })
+      return client.get(`${getResourceSlug(resource)}`, { params })
```
zackkrida, line 148:
```suggestion
      return client.get(getResourceSlug(resource), { params })
```

No interpolation needed anymore

zackkrida (changes requested):
This was very easy to follow and looks great! I've requested some changes. I tried to make it clear what suggestions feel optional; let me know if you have questions.

frontend/src/stores/search.ts
obulat: Great idea! This will make the code much simpler, clearer and easier to understand :)

frontend/src/stores/search.ts
obulat: Removed in [5cf32fc](https://github.com/WordPress/openverse/pull/3866/commits/5cf32fc5f72742c60081d32a98938a15f1ed5f64)

frontend/test/playwright/e2e/collections.spec.ts
obulat: Added comments in [4645b7b](https://github.com/WordPress/openverse/pull/3866/commits/4645b7b8e335bb6b94378c1fae85222e433a8172)

frontend/src/pages/audio/collection.vue
obulat: I tried that previously but it didn't work when Additional search views were using the "catch-all" path parameter in `/source/_.vue`: it was catching the localized 404 paths. I tried the same in this PR, and for some reason there was a redirect loop in the collection page.
We could try single `_media/collection.vue` in a separate PR.

zackkrida (approved):
This works great locally 🎉 LGTM!

AetherUnbound (commented):
I don't have a lot of notes on the code itself, but one thing I've noticed: the collection pages seem to be using the default text (e.g. "Openly Licensed Images, Audio and More | Openverse"). It feels like that should be specific per collections page, as in "tag: cat | Openverse" or something?

obulat:
> I don't have a lot of notes on the code itself, but one thing I've noticed: the collection pages seem to be using the default text (e.g. "Openly Licensed Images, Audio and More | Openverse"). It feels like that should be specific per collections page, as in "tag: cat | Openverse" or something?

I should have added more details on the scope of this PR. 

The main goal here is to make the creator/tag values with special characters work by making the `store` and the `media-service` use the URI-encoded query parameters, since URI-encoding path parameters is not feasible. So, it's important to see that you can visit the pages for creators with special characters in their names (I did add some special characters to the creator in the `tapes` in e2e tests to test that), and that the links (source/creator and tag links on the single result page and the source links in the sources table) work correctly, set the values in the store correctly, and decode the values for display so that the links don't have `%20` in them instead of spaces.

This PR actually breaks some things like loading of the images (on server-rendered pages), and it does not fix the SEO-related. That should be done in a separate PR.
It also does not fetch the exact matches from the API since the requests using query parameters use the full-text search in the API right now.

AetherUnbound (approved):
Tested this locally and the functionality as described works as expected! 🥳

PR #3868, opened 2024-03-04
Update the Additional search views IP

obulat:
## Fixes

Updates the Implementation Plan in line with the [discussion](https://github.com/WordPress/openverse/discussions/3825) conclusions

## Description

This PR updates the IP to move from path parameters to the query parameters. Also, SEO-related changes were added, too.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
+++ b/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
@@ -32,8 +32,11 @@ API returns all media with the selected tag, from the selected source or by the
 selected creator, sorted by date added to Openverse.
 
 Frontend allows to browse media items by a selected creator, source, or with a
-selected tag. These pages are indexed by search engines, have relevant SEO
-properties and can be shared with an attractive thumbnail and relevant data.
+selected tag using URLs like `/<mediaType>/collection?tag=tagName`,
+`/<mediaType>/collection?source=flickr` and
+`/<mediaType>/collection?source=flickr&creator=Photographer`. These pages are
+indexed by search engines, have relevant SEO properties and can be shared with
+an attractive thumbnail and relevant data.
 
 The single result pages are updated to add the links to source, creator and tag
 collections.
@@ -118,13 +121,13 @@ for fuzzy matches.
 **Finally**, we apply the ranking and sorting parameters, and "highlight" the
 fields that were matched in the results.
 
-The search controller needs to be updated to extract the first 2 steps into
+The search controller needs to be updated to extract the first 2 steps into a
 `build_search_query` method. A new `build_collection_query` should be added and
 used when the `collection` parameter is present. This method should create a
 filter query for the relevant field and value.
 
 The pagination and dead link cleanup should be the same for additional search
-views as for the default search views.
+queries as for the default search queries.
 
 #### Search request serializer updates
 
@@ -142,20 +145,25 @@ launched. In the text below, for brevity, the prefix is omitted.
 ```
 
 The collections will use the `search` endpoint with `collection` query parameter
-set to `tag`, `creator` or `source`, and the relevant query parameters. The
-`collection` parameter validator will check that the request contains the
-necessary additional parameters: `source` for `collection=source`, `creator` and
-`source` for `collection=creator`, and `tag` for `collection=tag`. The value of
-the `source` parameter will also be checked to be an existing source using the
-provider store.
+set to `tag`, `creator` or `source`, and the following additional query
+parameters. The `collection` parameter validator will check that the request
+contains the necessary additional parameters: `source` for `collection=source`,
+`creator` and `source` for `collection=creator`, and `tag` for `collection=tag`.
+
+The middleware validates that the values of parameters is not empty, and that
+the `source` parameter is an existing source in the provider store. If the
+values are invalid, the 404 yellow error page is shown.
 
 For the tag route, the new singular `tag` parameter should be used, rather than
-the existing plural `tags` should be used for legibility since we are presenting
-a single tag.
+the existing plural `tags`, since we are only presenting a single tag.
 
 The existing `source` and `creator` parameters will be reused, but will be
 parsed differently when `collection` parameter is present: they will only allow
-a single value instead of being split by `,` as it is for the default search.
+a single value instead of being split by `,` as it is for the default search. If
+the `source` contains invalid values, such as `source=flickr,europeana`, when
+the `collection` parameter is present, a 400 error with `detail` of "Invalid
+source parameter for source collection: `flickr,europeana`" is returned. If
+possible, a list of valid sources should be returned in the error message.
 
 Their documentation should be updated to reflect that. The additional
 documentation for the search parameters should be added as a draft so that it's
@@ -164,7 +172,11 @@ remove the `unstable__` prefix.
 
 ### 2. Add the `additional_search_views` feature flag
 
-The flag should be switchable, and off by default.
+The flag should be switchable, and off by default. This will mean that the
+changes are visible on [staging](https://staging.openverse.org) when the flag is
+switched on using the preferences page. When the features are stable, we will
+turn the flag on to test the new views in production. After we conclude that the
+project is successful, we will remove the flag and the conditional rendering.
 
 ### 3. Nuxt store and API request changes
 
@@ -184,7 +196,7 @@ the current approach. If it is set to `"collection"`, the query will be
 constructed using the new method (`buildCollectionQuery`) to set `collection`
 and other relevant parameters using the new `collectionParams` object in the
 store. It will _not_ be setting the filter parameters such as license or
-category.
+category, and will ignore such unsupported query parameters.
 
 ### 4. Create collection pages
 
@@ -222,9 +234,14 @@ search engines.
 
 The following titles should be used for the pages:
 
-- "Images by Olga in Flickr" for the creator page
+- "Images by Olga at Flickr" for the creator page
 - "Audio from Wikimedia" for the source page
 - "Images with the cat tag" for the tag page
+
+There are i18n consideration for these titles that we will work on during the
+implementation. It is important to make the titles translatable, which is
+difficult if the non-translatable dynamic names are used inside the sentence due
+to different sentence structures.
 
 The generic Openverse thumbnail will be used. We could also generate a thumbnail
 for the collection pages in the future, but this is not in scope for this
@@ -359,7 +376,7 @@ Analytics events should be added or updated:
 
 - We should also a special event for visiting source `VISIT_SOURCE_LINK`,
   similar to `VISIT_CREATOR_LINK`.
--
+
 - The `REACH_RESULT_END`, `LOAD_MORE` and `SELECT_SEARCH_RESULT` events should
   add strategy (`search`/`tag`/`creator`/`source`) and set `query` to tag name,
   source name or source/creator pair: `cat`, `flickr` or `flickr/Olga`.
@@ -454,10 +471,6 @@ To roll back the changes, we would need to set the feature flag to `OFF`.
 The biggest risk I see is that this project might be seen as an "invitation" to
 scraping. Hopefully, frontend rate limiting and the work on providing the
 dataset would minimize such risks.
-
-## Prior art
-
-<!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
 
 ## Plan revisions
 
--- a/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
+++ b/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
@@ -146,29 +146,49 @@ launched. In the text below, for brevity, the prefix is omitted.
 
 The collections will use the `search` endpoint with `collection` query parameter
 set to `tag`, `creator` or `source`, and the following additional query
-parameters. The `collection` parameter validator will check that the request
-contains the necessary additional parameters: `source` for `collection=source`,
-`creator` and `source` for `collection=creator`, and `tag` for `collection=tag`.
+parameters:
 
-The middleware validates that the values of parameters is not empty, and that
-the `source` parameter is an existing source in the provider store. If the
-values are invalid, the 404 yellow error page is shown.
+- `collection=tag` will require `tag` parameter to be set
+- `collection=source` will require `source` parameter to be set
+- `collection=creator` will require both `source` and `creator` parameter to be
+  set.
 
-For the tag route, the new singular `tag` parameter should be used, rather than
-the existing plural `tags`, since we are only presenting a single tag.
+This means that the existing `source` and `creator` parameters will be reused.
+For the tag collection, the new singular `tag` parameter should be used, rather
+than the existing plural `tags`, since we are only presenting a single tag.
 
-The existing `source` and `creator` parameters will be reused, but will be
-parsed differently when `collection` parameter is present: they will only allow
-a single value instead of being split by `,` as it is for the default search. If
-the `source` contains invalid values, such as `source=flickr,europeana`, when
-the `collection` parameter is present, a 400 error with `detail` of "Invalid
-source parameter for source collection: `flickr,europeana`" is returned. If
-possible, a list of valid sources should be returned in the error message.
+`MediaSearchRequestSerializer` should be updated to validate the `collection`
+parameter. This validator will check that the request contains the necessary
+additional parameters: `source` for `collection=source`, `creator` and `source`
+for `collection=creator`, and `tag` for `collection=tag`. It will only validate
+that the parameters _have_ values, but not the values themselves.
 
-Their documentation should be updated to reflect that. The additional
-documentation for the search parameters should be added as a draft so that it's
-not published on the API documentation site until we launch this project and
-remove the `unstable__` prefix.
+The `MediaSearchRequestSourceSerializer` currently splits the `source` parameter
+value by `,` and validates that each value is a valid source. This validator
+will be updated:
+
+- If the `collection` parameter is not present, it will behave as before.
+- If the `collection` parameter is set to `source` or `creator`, it will take
+  the value of `source` as is, and check if this value is present in the list of
+  valid sources, without splitting it by `,`. If the value is not valid, it will
+  return a 400 error, showing the invalid value, as well as a list of valid
+  sources for the media type.
+- If the `collection` parameter is set to `tag`, it will return the value as-is,
+- because it will be ignored in the search controller.
+
+The documentation for the `source` parameter should be updated to reflect that
+it accepts a single source when there is a `collection` parameter set to
+`source` or `creator`, and a comma-separated list of sources when there is no
+`collection` parameter. Similarly, the documentation for `creator` parameter
+should say that it accepts a comma-separated list of values when there is no `q`
+parameter, and a single URI-encoded value when the `collection=creator`
+parameter is present. The creator serializer does not need any changes as the
+parsing (splitting by comma for regular searches and URI-decoding for `creator`
+collections) is only done in the search controller.
+
+The additional documentation for the search parameters should be added as a
+draft so that it's not published on the API documentation site until we launch
+this project and remove the `unstable__` prefix.
 
 ### 2. Add the `additional_search_views` feature flag
 
@@ -215,6 +235,10 @@ parameters:
 - `source` parameter should be validated to be an existing source using the
   provider store
 - `creator` and `tag` parameters should be decoded using `decodeURIComponent`.
+
+The middleware validates that the values of parameters is not empty, and that
+the `source` parameter is an existing source in the provider store. If the
+values are invalid, the 404 yellow error page is shown.
 
 #### Fetching the media
 
--- a/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
+++ b/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
@@ -29,7 +29,7 @@ The original version of this plan was significantly revised on 2024-03-01 due to
 <!-- List any succinct expected products from this implementation plan. -->
 
 API returns all media with the selected tag, from the selected source or by the
-selected creator, sorted by date added to Openverse.
+selected creator within a given source, sorted by date added to Openverse.
 
 Frontend allows to browse media items by a selected creator, source, or with a
 selected tag using URLs like `/<mediaType>/collection?tag=tagName`,
--- a/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
+++ b/documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
@@ -170,7 +170,7 @@ The flag should be switchable, and off by default.
 
 We can reuse the search store as is for these pages.
 
-_Previously, frontend search store had a `searchBy` filter that allowed to
+_Previously, the frontend search store had a `searchBy` filter that allowed to
 search within the `creator` field. When `searchBy` value was set, the API `q`
 parameter was replaced with the `<searchBy>=<searchTerm>` API query parameter.
 This filter was removed because `searchBy` is not strictly a filter that can be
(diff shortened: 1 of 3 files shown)


documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
 **Finally**, we apply the ranking and sorting parameters, and "highlight" the
 fields that were matched in the results.
 
-The new search controller should allow for using different filters for the first
-step and to not use the full-text search. We should also create a new serializer
-for collection search requests. It should include the common parameters for
-`list` requests, such as `page` and `page_size`, and the parameters for the
-exact matches: `tag`, `creator` and `source`.
+The search controller needs to be updated to extract the first 2 steps into
```
zackkrida, line 121:
```suggestion
The search controller needs to be updated to extract the first 2 steps into a
```

Nit

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+`build_search_query` method. A new `build_collection_query` should be added and
+used when the `collection` parameter is present. This method should create a
+filter query for the relevant field and value.
+
+The pagination and dead link cleanup should be the same for additional search
+views as for the default search views.
```
zackkrida, line 127:
```suggestion
The pagination and dead link cleanup should be the same for additional search
views as for the default search views.
```

Should this read "queries" instead of "views", since there will no longer be dedicated API endpoints for the collections?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+path segments (`/`, `?` and `&`) that cannot be properly encoded, this approach
+turned out to be not feasible._
+
+```{note}
+The new query parameters (`collection` and `tag`) will have `unstable__` prefix until this project is
+launched. In the text below, for brevity, the prefix is omitted.
+```
+
+The collections will use the `search` endpoint with `collection` query parameter
+set to `tag`, `creator` or `source`, and the relevant query parameters. The
```
zackkrida, line 145: > and the relevant query parameters

I am not sure what these are. Could you be more specific here?

zackkrida: Ah, it's in the next sentence. Maybe "the following additional query parameters" instead?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+`collection` parameter validator will check that the request contains the
+necessary additional parameters: `source` for `collection=source`, `creator` and
+`source` for `collection=creator`, and `tag` for `collection=tag`. The value of
+the `source` parameter will also be checked to be an existing source using the
+provider store.
```
zackkrida, line 150: How does it validate? What error code is thrown?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+For the tag route, the new singular `tag` parameter should be used, rather than
+the existing plural `tags` should be used for legibility since we are presenting
+a single tag.
```
zackkrida, line 154:
```suggestion
For the tag route, the new singular `tag` parameter should be used, rather than
the existing plural `tags`, since we are only presenting a single tag.
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+The existing `source` and `creator` parameters will be reused, but will be
+parsed differently when `collection` parameter is present: they will only allow
+a single value instead of being split by `,` as it is for the default search.
```
zackkrida, line 158: Does "allow" just mean that these parameters won't perform the splitting, and instead will read the strings as-is? So `source=flickr,europeana` would search for `flickr,europeana`? As opposed to say, throwing an error or selecting the first item in the list (flickr)?

I just want to make sure we're specific with what "allow" means.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+Their documentation should be updated to reflect that. The additional
+documentation for the search parameters should be added as a draft so that it's
+not published on the API documentation site until we launch this project and
+remove the `unstable__` prefix.
```
zackkrida, line 163: Can we conditionally add documentation that would appear under the `unstable__` versions until we merge them?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+### 2. Add the `additional_search_views` feature flag
+
+The flag should be switchable, and off by default.
+
+### 3. Nuxt store and API request changes
+
+We can reuse the search store as is for these pages.
+
+_Previously, frontend search store had a `searchBy` filter that allowed to
```
zackkrida, line 173:
```suggestion
_Previously, the frontend search store had a `searchBy` filter that allowed to
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+#### Add `strategy` and `collectionParams` to `search` store
+
+The `strategy` parameter will be used to determine the API request query
+parameters. If it is set to `"search"`, then the API query will be created using
+the current approach. If it is set to `"collection"`, the query will be
+constructed using the new method (`buildCollectionQuery`) to set `collection`
+and other relevant parameters using the new `collectionParams` object in the
+store. It will _not_ be setting the filter parameters such as license or
+category.
```
zackkrida, line 187: What happens if unsupported params are present in the URL, are they ignored or do they they throw HTTP errors?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
-for legibility since we are presenting a single tag.
+This page should also have relevant SEO properties, and should be indexed by the
+search engines.
 
-The new views should use the same pagination and dead link cleanup as the search
-views.
+The following titles should be used for the pages:
 
-### 3. New and updated components
+- "Images by Olga in Flickr" for the creator page
```
zackkrida, line 225:
```suggestion
- "Images by Olga in Flickr" for the creator page
```
"
"at", "via", or "from" might be more appropriate than "in".

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+- "Audio from Wikimedia" for the source page
+- "Images with the cat tag" for the tag page
```
zackkrida, line 227:
```suggestion
- "Cat Images" for the tag page
```

This might be more straightforward, what do you think?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +118,156 @@ for fuzzy matches.
...
+
+The generic Openverse thumbnail will be used. We could also generate a thumbnail
+for the collection pages in the future, but this is not in scope for this
+project.
```
zackkrida, line 231: :+1: to keeping the scope constrained.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -213,156 +343,35 @@ page and "provider" in the
...
 
-Two analytics events should be added or updated:
+Analytics events should be added or updated:
 
 - The clicks on external creator link in the `VCollectionHeader` should be
   tracked as `VISIT_CREATOR_LINK` events.
 
 - We should also a special event for visiting source `VISIT_SOURCE_LINK`,
   similar to `VISIT_CREATOR_LINK`.
+-
```
zackkrida, line 362:
```suggestion
```

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -414,6 +423,11 @@ not dependent on anything.
 The work on the single result pages (step 7) can be done in parallel with the
 work on the collection pages (step 6), but should follow the previous steps.
 
+The new frontend views can use the existing query parameters (`source`,
+`creator` and `tags` - instead of `tag`) until the API changes are implemented
+since this _will_ return _some_ relevant results, and changing the query
+parameter names is easy.
```
zackkrida, line 429: Will this apply in production, while this is being developed, or both?

zackkrida (commented):
I've left some feedback for places where some small changes for clarity would be useful. Thank you for doing this! It can be a chore to document changes as they're being made but this has clarified my own understanding of the work needed here.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Since the additional search views are under a feature flag, this will only be seen on staging. It will apply only while the API changes are being developed.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: I was mainly thinking about the i18n difficulties when writing this sentence. For instance, to translate this sentence into Russian, you would have to change the case of "Cat" in this sentence. However, we could also add a note for translators to translate the "Images with the cat tag" sentence if translating "Cat images" would require to change the case of the tag itself (since changing the case of the tag name can't be automated, the "Cat images" would be wrong grammatically in the languages that use cases).
So, do you think changing this sentence to "Cat Images", and adding a comment to translators to translate this sentence as "Images with the Cat tag" if "Cat images" requires changes to the word "Cat", would work, @zackkrida ?

zackkrida: Ah, that is indeed complicated. Perhaps for this plan all we need is a note that there _are_ i18n considerations in developing these strings, and we can figure out the specifics as we work on it.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: > Does "allow" just mean that these parameters won't perform the splitting, and instead will read the strings as-is?

Yes, it will read the strings as-is, without splitting by comma.

I was thinking that the API would throw an error if the source is not in the list of sources for the media type. So, for instance, `source=flickr,europeana` would throw an error if used with `collection=source`. The error would say something like "Invalid source parameter for source collection: `flickr,europeana`".

Do you think it's the most clear way of handling such requests? Or should we rather use the first source?

zackkrida: That sounds great! I think it could be documented just like that in the document.

krysal (commented):
I'm leaving a few initial comments even though it's in draft. I'll be digesting the API Search controller changes. Things have changed a lot since the last time I looked at that part of the code.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
zackkrida: Excellent! Could you add that clarity to the plan?

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: I'm not sure about this, will have to check.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Updated the second step, "2. Add the `additional_search_views` feature flag"

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Added details on validation. The error code is not shown on the frontend, so currently the users see the "yellow error page"

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Added that they are ignored.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: Used "at"

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -444,3 +458,14 @@ dataset would minimize such risks.
 ## Prior art
 
 <!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
+
+## Plan revisions
```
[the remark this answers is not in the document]
obulat: I removed the Prior Art, and left the Revisions as level 2.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +121,170 @@ for fuzzy matches.
...
+For the tag route, the new singular `tag` parameter should be used, rather than
+the existing plural `tags`, since we are only presenting a single tag.
+
+The existing `source` and `creator` parameters will be reused, but will be
+parsed differently when `collection` parameter is present: they will only allow
+a single value instead of being split by `,` as it is for the default search. If
+the `source` contains invalid values, such as `source=flickr,europeana`, when
+the `collection` parameter is present, a 400 error with `detail` of "Invalid
+source parameter for source collection: `flickr,europeana`" is returned. If
+possible, a list of valid sources should be returned in the error message.
```
sarayourfriend, line 166: Can you clarify the actual implementation details of this? Will it use new `validate_` methods on the serializer, and then for each one check if `collection` is present, and if so, what? Will it throw an error any time a comma is present?

I don't think we can realiably validate this for anything other than source. For all collection parameters, they are passed as-is to the ES query: https://github.com//WordPress/openverse/blob/f56f7310e81ac0266807192803e4a451a8983766/api/api/controllers/search_controller.py#L399-L410

They already do not support multiple values.

Which, again, we need to keep in mind that we cannot reliably split on _any_ character for creator or tag names, neither of which are sanitised to remove commas. What would the validator do if a Flickr creator's name was `Photographer, Esq."? What if the tag is "Taipei, Taiwan"? Will these throw errors? How would a user resolve them? By URL encoding the commas? I'm not sure that will work, they may need to double encode them. Normally, you can pass multiple values for a query parameter by using `param=1,param=2`:

```
> p = new URLSearchParams()
URLSearchParams {}
> p.append("param", "1")
undefined
> p.append("param", "2")
undefined
> p.toString()
'param=1&param=2'
```

Some frameworks require adding `[]` to the front of explicit lists.

We, instead, have chosen to use a comma, which creates significant complication for things like creator. Luckily, we already do not treat creator on the search endpoint as a "listable" input. It is passed as-is in the normal search strategy: 

https://github.com//WordPress/openverse/blob/f56f7310e81ac0266807192803e4a451a8983766/api/api/controllers/search_controller.py#L335-L349

I'm highly sceptical we can reliably validate these inputs, and should instead pass them directly to ES regardless of what is in them. We could validate source to check that it _exists_, so that `flickr,wikimedia_commons` would 404 if passed as `source` with `collection` because there is no `flickr,wikimedia_commons` source. But trying to validate and prevent multiple inputs on tag and creator looks to me like it will create similar issues that we had with the path parameters.

sarayourfriend (changes requested):
I have one blocking comment, which is to request clarification and deeper explanation of the implementation details for the validation of the query parameters. The current approach, as described, I am fairly confident will create similar problems as we had with path parameters. I could be misunderstanding the intention, so I am not explicitly requesting a specific change, just either clarification of the proposed approach or a change to address the potential problems I've mentioned.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
zackkrida: I asked a similar question and Olga shared that the intention was to only validate the source, yes. 

https://github.com/WordPress/openverse/pull/3868#discussion_r1513495351

Good call for more specifics here though; I overlooked that this wasn't added in my re-review.

obulat: I'll add more clarification in the IP itself, but will also add a short explanation here.

For collection view, only two validation steps will be used:

1. if `collection` parameter is present, we check that the relevant parameters are also present:
- if `collection=source`, there is the `source` value in the query
- if `collection=creator`, there are `creator` and `source` values in the query
- if `collection=tag`, there is the `tag` value in the query
2. if collection is `source` or `creator`, we check that the value of the `source` parameter, as is, is present in the list of available creators for the media type. So, `flickr` or `stocksnap` will be valid, but `flick`, `Flickr`, `flickr,stocksnap` will throw an error. We will not split this values by comma or lower-case them. And since we control the creator names, it should be okay. 

As for the `tag` and `creator` parameters, they will be passed to the search controller "as is".

By the way, while trying this out, I realized that our current source validation is not very clear. We take a string of the `source` value, split it by comma, leave all of the values that are names of sources and ignore invalid values. So, `https://api.openverse.engineering/v1/images/?source=flick` ignores the source parameter and returns all of the images. This might be very confusing if the user has made a spelling mistake.

sarayourfriend: Sounds, good, Olga. Thanks for clarifying. It would resolve the issue for me if you update this paragraph to clarify that only the source parameter would ever be validated in any capacity. As it reads now, the creator value would also be validated, but I think that's an artefact of trying to clarify the difference between the normal search route's handling of these two, and the creator parameter getting sidelined a bit.

> By the way, while trying this out, I realized that our current source validation is not very clear. We take a string of the source value, split it by comma, leave all of the values that are names of sources and ignore invalid values. So, https://api.openverse.engineering/v1/images/?source=flick ignores the source parameter and returns all of the images. This might be very confusing if the user has made a spelling mistake.

That sounds like a good new issue, maybe even a "help wanted" one!

obulat:
I updated the serializer and validator descriptions, so this PR is ready for re-review, @sarayourfriend.
There's also a draft PR implementing the API changes described here: #3887

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +121,194 @@ for fuzzy matches.
...
+```
+
+The collections will use the `search` endpoint with `collection` query parameter
+set to `tag`, `creator` or `source`, and the following additional query
+parameters:
+
+- `collection=tag` will require `tag` parameter to be set
+- `collection=source` will require `source` parameter to be set
+- `collection=creator` will require both `source` and `creator` parameter to be
+  set.
```
krysal, line 154: This is the part that gets confusing for me because, in the "Expected Outcomes" section, the new URLs are shown as belonging to a new `collection/` endpoint and not a tag with a value like the ones shown here. What would be the final form, as `/<mediaType>/collection?tag=tagName` or `/<mediaType>/?collection=tag&tag=<tag_name>` or maybe like `/<mediaType>/?collection=tag&tagName=<tag_name>`? 

This is fundamental because if they end up in the form of `/<mediaType>/collection?tag=<tag_name>`, then there is not much difference from the original proposal of creating new endpoints, which my experience tells me it should be the easiest and cleanest path, at least for **tags** and **sources**. The `creator` makes perfect sense as a parameter for the `source` collection as it's currently portrayed.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
```diff
@@ -111,35 +121,194 @@ for fuzzy matches.
...
+
+#### Search request serializer updates
+
+_The initial version of this plan proposed to add new API endpoints and use path
+parameters for the values (e.g., `/image/tag/cat`) to make the URLs more
+readable, easier to share, will be easier to cache or perform cache invalidation
+required by [#1969](https://github.com/WordPress/openverse/issues/1969).
+However, since tags and creator names contain characters that are special for
+path segments (`/`, `?` and `&`) that cannot be properly encoded, this approach
+turned out to be not feasible._
```
krysal, line 140: Do you know if we captured the problem in an issue that can be linked here? I want to understand where the problem is precisely because it sounds pretty strange to me that that was an issue for the `tags/` endpoint 🤔 

Do we have an example of a URL involving the tag collection returning erroneous items? I may be missing something else, perhaps related to the casing.

krysal (commented):
I see that there is a certain consensus. Now that I have read the API part more carefully, I have a couple of doubts that I would like to clarify before approving.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: We first noticed this issue in the `/creator` endpoint: #3791
Then, we realized that tags has the same problem. You can see it now in prod for tags with a question mark.

You cannot enter this URL into browser's address bar: `https://api.openverse.engineering/v1/images/tag/18th century?/`, `?` is replaced with `/?/`, and thus the tag you're searching for becomes the one without a question mark.

You can access this tag if you URI-encode this path parameter: `https://api.openverse.engineering/v1/images/tag/18th%20century%3F/`. But as Sara noted elsewhere that encoding path parameters (not the query parameters) is unusual and unexpected. 
It becomes more confusing when you add query parameters to the URL, e.g. `https://api.openverse.engineering/v1/images/tag/18th century?/?page=2`.

In Nuxt, it becomes really confusing if a URL contains special symbols in the path parameter, and also has a query parameter, such as `http://localhost:8443/image/tag/tag? cat & dog?page=2`. The query is split in different places depending on what browser you use: Firefox or Chrome.
Note that currently these views don't support filtering using query parameters, and don't add parameters for `page` or `peaks`, but if we decide to add this in the future - it will be very problematic.

You can see my numerous attempts to make the special characters work in the creator and tag values in #3793

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
obulat: I added sample URLs for API in the expected outcomes and made the section easier to scan.

documentation/projects/proposals/additional_search_views/20230719-implementation_plan_additional_search_views.md
krysal: I see the reasoning now, makes sense the changes 👍

PR #3871, opened 2024-03-04
Fix Vale by passing files to process instead of a glob to ignore

### LINKED ISSUE #3764: Vale sometimes fails to run, complaining of non-existent files
## Description

@stacimc and I have both run into this issue in different ways. Here is Staci's example output of the error:

```
docker run --rm -v $PWD/..:/src:rw,Z --workdir=/src openverse-vale:local . --glob='!*{_build,_static,venv,.venv,.nuxt,.pnpm-store,node_modules,test-results,storebook-static,.ruff-cache,projects/proposals,changelogs/api,changelogs/frontend,changelogs/catalog,changelogs/ingestion_server}*'
stat venv/bin/python: no such file or directory
error: Recipe `run` failed on line 64 with exit code 2
```

I had the same error mode, but the file was `src/static/node_modules/sharp` (I'd created a node_modules folder in `frontend/src/static` when working on #3724). I was able to resolve the issue for me locally by deleting that directory, and didn't think anything much of it because I'd done something rather unusual, and I didn't think it would affect normal usage. However, it appears to indeed affect normal usage, as now Staci is running into the same problem.

Something tricky about it, however, is that we can't easily tell where the file causing the issue is, because Vale's message cuts off the first directory!

## Reproduction

Not sure how to reproduce at the moment. I tried recreating the issue I had locally, but wasn't able to reproduce it.

## Additional context

Marked high because it interrupts local development workflows.

zackkrida:
## Fixes

Fixes #3764 by @sarayourfriend 

## Description

Something with Vale's globbing logic would throw an error if a file wasn't found, typically in the case of broken/invalid symlinks. This was true even if the file wasn't one which would be processed by Vale, i.e, non-markdown files.

To fix this, instead of telling vale which files to ignore, I've written a script in our justfile to:

1. List all MD and MDX files in the repository
2. Filter out the directories we do not want to lint
3. Strip the absolute path from the beginning of the files
4. Replace newlines with spaces

This is then passed as the argument to Vale. 

### Concerns and considerations

Some potential pitfalls or annoyances with this approach:

- It doesn't actually fix whatever the underlying problem was
- The list of filenames to pass to Vale is printed to the terminal and clutters our precommit output
- If the repo or the list of files we want to lint grows considerably, we could run into bash argument length limits

## Testing Instructions

1. Edit a markdown file which should be linted by Vale, for example adding "OpenVerse" to one of our readmes.
2. Also edit a file which _should not_ be linted by vale, like our changelogs, to contain "OpenVerse" as well.
3. Manually run precommit with `pre-commit run --all-files`. Confirm the correct Vale errors for the changes you made appear.
4. **Bonus:** `cd ./vale && j debug`. Manually review the list of files being linted and check if anything you think should be there is missing, or if files that shouldn't be linted are.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
@AetherUnbound and @sarayourfriend I do think this solution is ready for review but I am drafting it due to both of your AFK and in case you have any suggestions for improvements to the justfile, like splitting up the Vale justfile and capturing the various sub-outputs, for example. 

Please feel free to review this whenever you like though! :smile:

.vale/justfile
```diff
@@ -1,64 +1,37 @@
...
+debug:
+    echo {{ FILES }} | tr ' ' '\n'
 
 # Run Vale configured for Openverse in Docker.
 # Using Docker avoids the need for contributors to install the Vale binary.
 #Configuration defaults to what is used for CI.
-run +files=".": build
-    docker run --rm \
+run +files=FILES: build
+    docker run \
```
zackkrida, line 33:
```suggestion
    docker run --rm \
```

AetherUnbound (commented):
This looks right and makes sense! However, I tried making the following diffs and pre-commit didn't run any changes:

```diff
diff --git a/catalog/README.md b/catalog/README.md
index 813b4262d..a224edd95 100644
--- a/catalog/README.md
+++ b/catalog/README.md
@@ -1,4 +1,4 @@
-# Airflow Catalog
+# Airflow Catalog for OpenVerse
 
 This service is documented at
 [docs.openverse.org/catalog](https://docs.openverse.org/catalog/index.html).
diff --git a/ingestion_server/ingestion_server/cleanup.py b/ingestion_server/ingestion_server/cleanup.py
index 6e794f290..1768805d3 100644
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -295,6 +295,7 @@ def save_cleaned_data(result: dict) -> dict[str, int]:
 def clean_image_data(table):
     """
     Clean up data loaded from upstream that is unsuitable for prod before going live.
+    in OpenVerse.
 
     :param table: The staging table for the new data
     :return: None
```

The first file at least seems like it _should_ be included in the files to process, based on `j .vale/debug`'s output:

```
readme_assets/system_diagram/README.md
catalog/dags/oauth2/README.md
catalog/dags/retired/README.md
catalog/docs/provider_data_ingester_faq.md
catalog/docs/data_models.md
catalog/README.md
catalog/utilities/README.md
CONTRIBUTING.md
frontend/README.md
```

zackkrida:
@AetherUnbound super interesting, Vale doesn't seem to suggest corrections for `OpenVerse` in markdown _titles_, just text. You can test this on `main` and confirm. 

I am also fairly certain vale only operates on Markdown files, but not 100%. I tried the `ingestion_server/ingestion_server/cleanup.py` change on main `and` it also didn't appear there.

sarayourfriend:
Vale is not configured to lint code documentation: it is a prose linter (Markdown, HTML, reStructuredText, etc), and cannot lint in-code documentation (unless, like the DAG docs, it is extracted into a prose file that Vale will lint as part of our build step).

Vale not linting spelling in titles is a separate issue, and can be fixed by changing the contexts checked in the Openverse styles rule that enforces the correct capitalisation of Openverse. It isn't an issue with this PR.

.vale/justfile
```diff
@@ -1,64 +1,37 @@
...
+FILES := ```
+    find $PWD/.. -type f \( -name "*.md" -o -name "*.mdx" \) \
+    | grep -vE ".pnpm|changelogs|projects/proposals|node_modules" \
+    | sed "s|$PWD/../||" \
+    | tr '\n' ' '
+```
+
+# Print the list of FILES for Vale to lint
+debug:
+    echo {{ FILES }} | tr ' ' '\n'
```
sarayourfriend, line 27: Just a nit, will `tr ' ' '\n'` incorrect break files with legitimate spaces in their names? Not an issue for us now (and maybe never will be), but if `FILES` is a script now anyway, it can just be a recipe. The only reason it was a variable before was because it was easier to modify the list that way (I thought).

We can even break it into multiple steps to make it easier to read, and make the separator configurable (and conditional). It's a good opportunity to move the ignore pattern into a named space as well:

```
IGNORE_FILES_PATTERN := ".pnpm|changelogs|projects/proposals|node_modules|.?venv"

@files separator="\n":
    #! /usr/bin/env sh
    files=$(
        find "$PWD/.." -type f \( -name "*.md" -o -name "*.mdx" \) \
            | grep -vE "{{ IGNORE_FILES_PATTERN }}" \
            | sed "s|$PWD/../||"
    )
    if [ '{{ separator }}' != '\n' ]; then
        echo "$files" | tr '\n' '{{ separator }}'
        exit 0
    fi
    echo "$files"
```

I also somewhat doubt we should try to support passing a custom list of files to the run recipe, because now if I run `just .vale/run frontend`, it won't ignore the files it should ignore (and I won't find out until _after_ my PC has spent ages marching through the npm module tree linting every single README file).

Anyway, here is a diff that works with `run` able to take the `files` argument still (I just think we could do to remove it as well):

```diff
diff --git a/.vale/justfile b/.vale/justfile
index 4b9ad0043..9737b3c3f 100644
--- a/.vale/justfile
+++ b/.vale/justfile
@@ -11,27 +11,27 @@ NO_COLOR := "\\033[0m"
 build:
     docker build . -t openverse-vale:local
 
-# 1. List all MD and MDX files in the repository
-# 2. Filter out the directories we do not want to lint
-# 3. Strip the absolute path from the beginning of the files
-# 4. Replace newlines with spaces
-FILES := ```
-    find $PWD/.. -type f \( -name "*.md" -o -name "*.mdx" \) \
-    | grep -vE ".pnpm|changelogs|projects/proposals|node_modules" \
-    | sed "s|$PWD/../||" \
-    | tr '\n' ' '
-```
+IGNORE_FILES_PATTERN := ".pnpm|changelogs|projects/proposals|node_modules|.?venv"
 
-# Print the list of FILES for Vale to lint
-debug:
-    echo {{ FILES }} | tr ' ' '\n'
+files separator="\n":
+    #! /usr/bin/env sh
+    files=$(
+        find $PWD/.. -type f \( -name "*.md" -o -name "*.mdx" \) \
+            | grep -vE "{{ IGNORE_FILES_PATTERN }}" \
+            | sed "s|$PWD/../||"
+    )
+    if [ '{{ separator }}' != '\n' ]; then
+        echo "$files" | tr '\n' '{{ separator }}'
+        exit 0
+    fi
+    echo "$files"
 
 # Run Vale configured for Openverse in Docker.
 # Using Docker avoids the need for contributors to install the Vale binary.
 #Configuration defaults to what is used for CI.
-run +files=FILES: build
+run +files="": build
     docker run --rm \
         -v $PWD/..:/src:rw,Z \
         --workdir=/src \
         openverse-vale:local \
-        {{ files }}
+        {{ if files == "" { "`just files`" } else { files } }}
```

sarayourfriend (approved):
This LGTM. I've left a comment with suggestions that I think would improve the code and maintainability moving forward, but in it's present state it is also fine to me. I've confirmed that everything still lints as _configured_. Here is a diff that will fix the term casing on headings:

```diff
diff --git a/.vale/styles/Openverse/TermCasing.yml b/.vale/styles/Openverse/TermCasing.yml
index 90ecadc2e..cf64c6d5f 100644
--- a/.vale/styles/Openverse/TermCasing.yml
+++ b/.vale/styles/Openverse/TermCasing.yml
@@ -5,7 +5,7 @@ message: "Incorrect casing. Use '%s' instead of '%s'."
 level: error
 ignorecase: false
 scope:
-  - paragraph
+  - text
 swap:
   # [^/\.] prevents matching things that look like URLs, file paths, or GitHub team mentions
   # For example: @WordPress/openverse-maintainers
```

And here's a diff that will fix a slightly annoying ordering issue in the Dockerfile that makes iterating on our custom styles a pain:

```diff
diff --git a/.vale/Dockerfile b/.vale/Dockerfile
index 9bd1fd58e..1afd8b6e0 100644
--- a/.vale/Dockerfile
+++ b/.vale/Dockerfile
@@ -2,8 +2,9 @@ FROM jdkato/vale:v2.30.0
 
 WORKDIR /vale
 COPY .vale.ini .
-COPY styles styles
 
 RUN vale sync
 
+COPY styles styles
+
 ENTRYPOINT ["vale", "--config=/vale/.vale.ini"]
```

This could be improved further by using a bind mount for our custom styles, but that is even further out of scope for this PR.

Anyway, all of that is just suggestions, if you want to make them here, or if we want a separate PR for them :+1: This PR looks good to me as-is.

.vale/justfile
zackkrida: @sarayourfriend this is exactly what I was looking for; thank you! The piece I was missing when trying to take this myself was making `files` a recipe, and generally thinking in terms of "composing recipes". 

I did decide to go ahead and remove allowing the user to pass their own files. I originally conceived of that being _specific_ file names and didn't consider the ignore issue when passing directories.

These changes are in cb69954.

AetherUnbound:
Thanks @sarayourfriend and @zackkrida!

> Vale is not configured to lint code documentation: it is a prose linter (Markdown, HTML, reStructuredText, etc), and cannot lint in-code documentation (unless, like the DAG docs, it is extracted into a prose file that Vale will lint as part of our build step).

Totally, I think I might have taken this line a bit too seriously 😅 

> Also edit a file which should not be linted by vale, like our changelogs, to contain "OpenVerse" as well.

I chose a file that I knew wouldn't be: code docstrings. So that was my intended example of "this shouldn't be changed"!

I'll give this another shot now and see if it all works 😄

AetherUnbound (approved):
It works!! 🚀 Here's the output of `j lint` after making some breaking changes:

```
docker run --rm -v $PWD/..:/src:rw,Z --workdir=/src openverse-vale:local `just files`
#! /usr/bin/env sh
files=$(
    find "$PWD/.." -type f \( -name "*.md" -o -name "*.mdx" \) \
        | grep -vE ".pnpm|changelogs|projects/proposals|node_modules|.?venv" \
        | sed "s|$PWD/../||"
)
if [ '
' != '\n' ]; then
    echo "$files" | tr '\n' '
'
    exit 0
fi
echo "$files"

 catalog/README.md
 1:19  error  Incorrect casing. Use           Openverse.TermCasing 
              'Openverse' instead of                               
              'OpenVerse'.                                         
 6:9   error  Incorrect casing. Use           Openverse.TermCasing 
              'Openverse' instead of                               
              'OpenVerse'.                                         

✖ 2 errors, 0 warnings and 0 suggestions in 258 files.
error: Recipe `run` failed on line 41 with exit code 1

```

PR #3878, opened 2024-03-05
Update documentation for how to perform rollbacks

stacimc:
## Fixes

I noticed that some documentation was out of date or slightly unclear while performing a deploy recently.

## Description

This PR updates some documentation around rollbacks following the changes to the release process in https://github.com/WordPress/openverse/pull/3755/files. It also tries to clarify some ambiguity I found confusing (in general, referring to "the workflow" without specifying deployment vs release-app).

## Testing Instructions

Read through the documentation and confirm that the updates are **accurate**. I've left a few questions in comments about things I was not 100% sure of, so please correct me!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/api/guides/deploy.md
+++ b/documentation/api/guides/deploy.md
@@ -21,10 +21,11 @@ For more information on how deployments work, please see the [general deployment
    and verify the SHA is the same as the value you deployed.
 1. Review and Approve the automatically-generated changelog pull request in the
    repository.
-1. In the event of errors or problems, repeat the deployment process using the
-   latest stable version of the application. You can find the release version
-   number in the [changelogs](/changelogs/index), and then the tag to pass to
-   the action is the version number prefixed with "rel-", for example
-   "rel-2023.07.03.17.52.00".
+1. In the event of errors or problems, rollback the application by running the
+   appropriate deployment workflow from the WordPress/openverse-infrastructure
+   repository using the tag of the latest stable version. You can find the
+   release version number in the [changelogs](/changelogs/index), and then the
+   tag to pass to the workflow is the version number prefixed with "rel-", for
+   example "rel-2023.07.03.17.52.00".
 1. If anything else goes wrong or service is disrupted, consider this a
    Production Incident and notify the team.
--- a/documentation/general/deployment.md
+++ b/documentation/general/deployment.md
@@ -152,9 +152,12 @@ publishes the release is pinged in the PR description to help with the
 visibility of the PR.
 
 For the API and frontend, publishing the release also triggers an automated
-production deployment using the [release app workflow](https://github.com/WordPress/openverse/blob/main/.github/workflows/release-app.yml), which itself triggers the appropriate _deploy_ app workflow in the infrastructure repository. For the catalog and
-ingestion server, however, maintainers must manually deploy the changes to
-production using the Terraform deployment process.
+production deployment using the
+[release app workflow](https://github.com/WordPress/openverse/blob/main/.github/workflows/release-app.yml),
+which itself triggers the appropriate _deploy_ app workflow in the
+infrastructure repository. For the catalog and ingestion server, however,
+maintainers must manually deploy the changes to production using the Terraform
+deployment process.
 
 ## Rollbacks
 
--- a/documentation/general/deployment.md
+++ b/documentation/general/deployment.md
@@ -148,13 +148,13 @@ release for an application. Publishing a release automatically tags the latest
 docker image for the application and opens a PR to add the changelog to the
 [documentation site's changelog directory](https://docs.openverse.org/changelogs/index.html).
 This needs to be manually approved and merged by maintainers. The person who
-triggers the release app workflow is pinged in the PR description to help with
-the visibility of the PR.
+publishes the release is pinged in the PR description to help with the
+visibility of the PR.
 
 For the API and frontend, publishing the release also triggers an automated
-production deployment. For the catalog and ingestion server, however,
-maintainers must manually deploy the changes to production using the Terraform
-deployment process.
+production deployment using the release app workflow. For the catalog and
+ingestion server, however, maintainers must manually deploy the changes to
+production using the Terraform deployment process.
 
 ## Rollbacks
 
@@ -163,8 +163,11 @@ realise that the deployed code is not behaving as we expected. In these cases it
 may be necessary to force an environment to be deployed to a specific version,
 usually the previous version of the application.
 
-The same staging and production deployment workflows are used directly to
-rollback any environment for any service. Only members of the
+The same staging and production deployment workflows in the
+WordPress/openverse-infrastructure repository are used directly to rollback any
+environment for any service. **Do not attempt to perform a rollback using the
+release-app workflow in the monorepo, or by creating a new release; this will
+result in the creation of duplicate tags.** Only members of the
 @WordPress/openverse-maintainers GitHub team are able to dispatch the workflows
 to completion. Anyone else who tries to dispatch them will have the workflow
 automatically fail.
@@ -201,12 +204,12 @@ it. Please see the
      this yourself (i.e., if you do not have the ability to make changes to our
      Terraform configuration)
 
-2. Deploy the code that depends on the new or updated variable by dispatching
-   the appropriate deployment workflow for the application and environment that
-   was updated.
+2. Deploy the code that depends on the new or updated variable by publishing the
+   release for the application and environment that was updated.
    - If the application only needs to be re-deployed to the already
-     running-version, dispatch the workflow with the version present at the
-     version endpoint for the application. See the "version endpoint" in the
+     running-version, dispatch the appropriate deployment workflow in the
+     WordPress/openverse-infrastructure repository with the application's
+     current version. See the "version endpoint" in the
      [real-time deployment information](#real-time-deployment-information)
      tables above to determine the currently running version.
 
--- a/documentation/general/deployment.md
+++ b/documentation/general/deployment.md
@@ -152,7 +152,7 @@ publishes the release is pinged in the PR description to help with the
 visibility of the PR.
 
 For the API and frontend, publishing the release also triggers an automated
-production deployment using the release app workflow. For the catalog and
+production deployment using the [release app workflow](https://github.com/WordPress/openverse/blob/main/.github/workflows/release-app.yml), which itself triggers the appropriate _deploy_ app workflow in the infrastructure repository. For the catalog and
 ingestion server, however, maintainers must manually deploy the changes to
 production using the Terraform deployment process.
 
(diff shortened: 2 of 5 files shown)


documentation/api/guides/deploy.md
```diff
@@ -21,10 +21,11 @@ For more information on how deployments work, please see the [general deployment
...
-   latest stable version of the application. You can find the release version
-   number in the [changelogs](/changelogs/index), and then the tag to pass to
-   the action is the version number prefixed with "rel-", for example
-   "rel-2023.07.03.17.52.00".
+1. In the event of errors or problems, rollback the application by running the
+   appropriate deployment workflow from the WordPress/openverse-infrastructure
+   repository using the tag of the latest stable version. You can find the
+   release version number in the [changelogs](/changelogs/index), and then the
+   tag to pass to the workflow is the version number prefixed with "rel-", for
+   example "rel-2023.07.03.17.52.00".
```
stacimc, line 29: Is this accurate (here and in the other applications)? Is the appropriate tag the `rel-` tag, or the tag obtained from the version endpoint?

documentation/general/deployment.md
```diff
@@ -163,8 +163,11 @@ realise that the deployed code is not behaving as we expected. In these cases it
 may be necessary to force an environment to be deployed to a specific version,
 usually the previous version of the application.
 
-The same staging and production deployment workflows are used directly to
-rollback any environment for any service. Only members of the
+The same staging and production deployment workflows in the
+WordPress/openverse-infrastructure repository are used directly to rollback any
+environment for any service. **Do not attempt to perform a rollback using the
```
stacimc, line 168: Again, is this accurate? My understanding is that rollbacks should be done _only_ by running the deployment workflow from the infra repo, with the latest stable version.

Would triggering a new release at https://github.com/WordPress/openverse/releases/new with the stable version result in creating duplicate tags?

documentation/general/deployment.md
```diff
@@ -201,12 +204,12 @@ it. Please see the
...
-   the appropriate deployment workflow for the application and environment that
-   was updated.
+2. Deploy the code that depends on the new or updated variable by publishing the
+   release for the application and environment that was updated.
    - If the application only needs to be re-deployed to the already
-     running-version, dispatch the workflow with the version present at the
-     version endpoint for the application. See the "version endpoint" in the
+     running-version, dispatch the appropriate deployment workflow in the
+     WordPress/openverse-infrastructure repository with the application's
+     current version. See the "version endpoint" in the
```
stacimc, line 212: Just curious why the docs here say to use the version from the version endpoint, rather than the rel-tag. Is there an intended difference?

documentation/api/guides/deploy.md
zackkrida: It is accurate. Here's a screenshot of the latest API releases on https://github.com/wordpress/openverse/pkgs/container/openverse-api:
 
Screenshot from 2024-03-05 19-02-05

I believe with the way our builds work that those two latest images are essentially the same image, or put another way the `rel-` tagged release is a duplicate of the immediate previous release.

sarayourfriend: Yes, the `rel-` tag. We could probably drop that if we wanted to, one day, and just use the date without the `rel-` prefix. That is probably less confusing than needing to swap the application slug for `rel-`? Instead, we'd just drop the application slug?

documentation/general/deployment.md
sarayourfriend: Yes, that is correct. Really we can frame it like this: to deploy a specific, already released version, run the deployment workflow in the infrastructure repository with the version tag to deploy. This applies to _any_ (literally, all possible cases) scenario where we need to trigger a dpeloyment to an existing release version, whether a rollback or re-deployment (e.g., to get new environment variables) or whatever.

Publishing a release _creates the release_ and the release app workflow is just an incidental part of publishing (it would be good if we just renamed it to "Tag stable image and publish changelog" (TSIAPC :cold_sweat:) or something, as it is _not_ in any way independent of publishing). Publishing is only for creating new a new release off of `HEAD` on `main` and has no other purpose.

The _deployment_ workflows are general purpose deployment workflows utilised by the "TSIAPC" workflow (release-app) as an incidental part of publishing a release, but not a _strictly necessary_ one. We could, for example, remove automated deployments for the API and frontend from the "TSIAPC" workflow (release-app), to completely separate the ideas and prevent any possible confusion between the two. _Publishing is publishing_. _Deploying is deploying_. There is no necessary crossover between the two, which is why they are distinct workflows with no direct relationship to each other, aside from that publishing _happens_ to trigger the deployment workflows.

This might sound pedantic, but it's really important to completely separate these two workflows conceptually, and for everyone creating releases and/or deploying the application _in any capacity_ understands the separation between them. Publishing _triggers_ deployment, but is not _part of_ deployment. Deployment is also not part of publishing a release, it just happens to occur after publishing a release because that's basically why we publish releases, to immediately deploy to production.

It really might be best to just separate the two and not have publishing trigger a deployment, to emphasise the separation of the two... but, as it is literally what we want to do every time we publish a release, I think we would quickly find it tedious that the frontend and API no longer automatically deploy to production when we create a new release.

I'm also thinking there is a confusion of the GitHub terms and what would be the ideal, precise words to use: Publishing implies making the thing public, which in our case really means deploying it to production (because we run apps, not libraries). But, ideally, the process would be called "Create a stable version with a changelog, then deploy it", which emphasises the two completely separate parts. It's just that GitHub call's it "publishing a release", wihch overloads a whole lot of ideas that for us are actually separate. If we think of publishing as a combination of creating the release and deployment, it's still kind of confusing.

Anyway, the concepts are currently muddled, and the words GitHub uses and that we've titled our workflow files with are not helping maintain the separation. If removing the automated deployments from the workflow triggered by publishing a release in GitHub would help make that clearer, then it's probably worth the tediousness, as mistakes here are potentially more confusing than the tediousness.

documentation/general/deployment.md
sarayourfriend: There is no underlying difference, they will point to exactly the same image, the rel- tags are aliases for the commit hash tags the version endpoint returns. From a process perspective, however, it is impossible to confuse which image tag to use if you follow the version endpoint, whereas finding the rel- tag is potentially fraught.

sarayourfriend (approved):
This LGTM, I don't see a need for changes now.

**But**. I think that if we remove the automated production deployment from the workflow that runs after publishing a release in GitHub, that all of this documentation would be far simpler. Rather than needing to add any caveats, we would always have a simple instruction:

- Deploy the application to the version you wish to be deployed

That is actually already true in all cases. As I said in my lengthy comment, it just _happens_ that the workflow triggered by publishing a release in GitHub does this for you. If we remove that though, it would completely separate the idea of publishing a release as "a method of deployment". There is only one method of deployment for any app, and any deployment to any version always uses that single method, regardless of whether it is automatically or manually triggered. If we switch it to always be manually triggered, then while it adds a step to pushing a version to production, it separates the question of "how do I deploy a version" only one possible answer: deploy it using the deployment workflow.

This would basically remove all caveats about deployment, aside from zero-downtime concerns, which are already extensively covered in a separate document. The documentation outline would go like this:

```
## Pushing a version to production

1. Publish the drafted release in GitHub for the app you want to run the latest code for in production
2. Copy the changelog title, replace the application slug with `rel-`, and then run the deployment workflow for the application from the infrastructure repository with that new tag.

## Rollback

Run the deployment workflow with the version you wish to roll the application back to.

## Redeploy

Run the deployment workflow with the version tag reported by the `/version` endpoint of the application you wish to redeploy.
```

It would be even simpler if we did away with the annoying task of changing the tag prefix from the app slug to `rel-`.

What do you think, Staci? Would that clarify the ideas? It adds a step to updating production to the latest code, but prevents any possibility of confusion between tagging an image with a "stable release" tag, and the deployment of the app. I think we could delete a fair bit of documentation and remove basically all the caveats that aren't just referncing zero-downtime.

stacimc:
@sarayourfriend Thank you so much for responding so thoroughly!

> What do you think, Staci? Would that clarify the ideas? It adds a step to updating production to the latest code, but prevents any possibility of confusion between tagging an image with a "stable release" tag, and the deployment of the app. I think we could delete a fair bit of documentation and remove basically all the caveats that aren't just referncing zero-downtime.

I'm not opposed to separating them out, but I don't know that I would remove much documentation afterward. For me the confusion in the docs was related entirely to ambiguity in phrasing "the process" or "the workflow" etc.

Colloquially we use the phrase "deploying" a service to mean the process of publishing the release _and then_ actually deploying it, and I think we would still do so even if that work required two steps instead of one. When the documentation for a rollback uses language like "repeat the deployment process", to me that _really_ sounds like it's saying to repeat **all** the steps. The phrases "workflow", "deployment workflow", and "deployment process" all could be interpreted as referring to (1) the actual `deployment - <env> <service>` workflows in the infra repo, (2) the release-app workflows in the monorepo, or (3) the general process of deploying at a higher level (since you're right, 'publishing', 'process', and 'workflow' are all unfortunately loaded terms).

For me all the confusion is immediately cleared up just by specificity. I read "repeat the deployment process" to mean that I should be publishing a new release somehow, and was immediately confused because that doesn't seem right! If we specify that we mean to use `Run Workflow` in the infra repo to run just the deployment workflow that immediately makes sense.

documentation/general/deployment.md
stacimc: Makes sense, thanks!

documentation/api/guides/deploy.md
stacimc: No strong feelings either way -- I think we see/use the `rel-` tags plenty when deploying catalog/ingestion-server so it's not especially confusing as is. I just wanted to double-check :)

sarayourfriend:
> Colloquially we use the phrase "deploying" a service to mean the process of publishing the release and then actually deploying it

> I read "repeat the deployment process" to mean that I should be publishing a new release somehow

I see, I didn't understand that _that_ was the point of ambiguity, but I see now that my suggestion wouldn't address that, basically at all. Ideally we would not use deployment as a catch-all, but you're right that we do, and there's not much we can really do to just change the language. Thanks for clarifying where the confusion was coming from, and sorry I misunderstood it!

documentation/general/deployment.md
```diff
@@ -148,13 +148,13 @@ release for an application. Publishing a release automatically tags the latest
...
-triggers the release app workflow is pinged in the PR description to help with
-the visibility of the PR.
+publishes the release is pinged in the PR description to help with the
+visibility of the PR.
 
 For the API and frontend, publishing the release also triggers an automated
-production deployment. For the catalog and ingestion server, however,
-maintainers must manually deploy the changes to production using the Terraform
-deployment process.
+production deployment using the release app workflow. For the catalog and
```
zackkrida, line 155:
```suggestion
production deployment using the deploy app workflow. For the catalog and
```

This might provide even more clarity, and also better matches the `deploy-{environment}-{service}.yml` naming of the workflows. 

We could consider linking to those files: https://github.com/WordPress/openverse-infrastructure/tree/main/.github/workflows for additional clarity, but not necessary.

zackkrida (approved):
I think this helps quite a bit. I've left one non-blocking suggestion.

PR #3887, opened 2024-03-08
Use search query parameters for additional search views in the API

### LINKED ISSUE #3869: Replace API collection paths with search parameters
## Problem

Additional search views in the API should use the `/search` endpoint with query parameters instead of separate endpoints with path parameters.

## Description

The following changes are necessary:

- remove the new endpoints (tag, creator and source)
- update the serializer to use `unstable__collection` and also add `unstable_tag` (singular, in addition to plural `tags` that's used for search within tags instead of exact match)
- update the `build_collection_query` method

### Additional changes
During the writing and review process, the following changes were also added:
- the `MediaSearchRequestSourceSerializer` was merged into the `MediaSearchRequestSerializer`. We now need to add `media_type` to the serializer context when calling it.
- `SHOW_COLLECTION_DOCS` env variable allows to show the updated documentation for the search endpoint. When it's set to False, no changes are shown in the API doc site.

## Additional context

The changes were discussed in [Rethinking additional search views URLs](https://github.com/WordPress/openverse/discussions/3825#top)

obulat:
## Fixes

Fixes #3869 by @obulat

## Description

This PR removes the collection API endpoints and uses the `search` endpoint with `collection`-related query parameters.

This PR adds an `ENABLE_COLLECTIONS` environment variable. When it's set to `True`, the API documentation includes the additional documentation for `tag`, `collection` query parameters, updated search endpoint descriptions and updated descriptions of `creator` and `source` query parameters.

The source serializer is updated to validate a single source when `collection` parameter is present. 
Note: the `source` serializer takes the value of the last parameter if there is more than one: 
`source=flickr&source=stocksnap` only sends `stocksnap` to the source validator.

### Documentation
Add `ENABLE_COLLECTIONS=True` to your `/api/.env` file
Run the API `just api/up`.
Check the documentation for the images and audio search endpoints. 
Text improvement suggestions/rewrites are very welcome!

Set `ENABLE_COLLECTIONS` env variable to `False` and see that the new parameters (`unstable__collection` and `unstable__tag` are not listed, and the documentation is almost the same as it currently is in prod. _Note I removed `q` from the list of parameters that can be used to filter results since it's not used for that._

### Successful collection requests
_These requests should return the collection results_
- http://localhost:50280/v1/images/?unstable__collection=tag&unstable__tag=CAT
- http://localhost:50280/v1/images/?unstable__collection=source&source=stocksnap
- http://localhost:50280/v1/images/?unstable__collection=creator&source=stocksnap&creator=Living%20Together

### Serializer validation errors
- http://localhost:50280/v1/images/?unstable__collection=creator&source=stocksnap
            ```"The `creator` and `source` parameters are required when `unstable__collection` is set to `creator`."```
- http://localhost:50280/v1/images/?unstable__collection=tag&unstable__tag=CAT&q=cat
             ```"The `collection` parameter cannot be used with the `q` parameter."```
- http://localhost:50280/v1/images/?unstable__collection=source&source=stocksnap,flickr
            ```"Invalid source parameter 'stocksnap,flickr'. Use one of the valid sources: 'stocksnap', 'flickr'"```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
> Note: the `source` serializer takes the value of the last parameter if there is more than one:
> `source=flickr&source=stocksnap` only sends `stocksnap` to the source validator.

I've seen this as a way that some other APIs handle arrays 😮 is this something we should support too?

api/api/docs/base_docs.py
```diff
@@ -149,109 +140,64 @@ def build_source_path_parameter(media_type: MediaType):
...
-        parameters=[
-            request_serializer,
-            *path_parameters,
-        ],
-    )
+SEARCH_DESCRIPTION_DEFAULT = """
+Return audio files that match the query.
+
+This endpoint allows you to search within specific fields, or to retrieve
+a collection of all audio files from a specific source, creator or tag.
```
AetherUnbound, line 147: Nit: should `audio` be here?
```suggestion
a collection of all %(media_type)s from a specific source, creator or tag.
```

api/api/serializers/docs.py
```diff
@@ -0,0 +1,84 @@
...
+is fuzzy, so `creator=john` will match any value that includes the
+word `john`. If the value contains space, items that contain any of
+the words in the value will match. To search for several values,
+join them with a comma."""
+
+CREATOR = """
+_When `q` parameter is present, `creator` parameter is ignored._
+
+**Creator collection**
+When used with `collection=creator`, returns the collection of media
```
AetherUnbound, line 17: Since the source is also required here, can we make that explicit in the example:
```suggestion
When used with `collection=creator&source=sourceName`, returns the collection of media
```

api/api/serializers/docs.py
```diff
@@ -0,0 +1,84 @@
...
+
+Should be used with `{TAG}`, `source` or `creator`+`source`"""
+
+EXCLUDED_SOURCE_HELP_TEXT = """
+A comma separated list of data sources to exclude from the search.
+Valid values are `source_name`s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/.
+"""
+SOURCE_HELP_TEXT_COLLECTIONS_DISABLED = """
+A comma separated list of data sources; valid values are
+``source_name``s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/."""
```
AetherUnbound, line 47: Are the double backticks necessary here?
```suggestion
`source_name`s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/."""
```

api/api/serializers/media_serializers.py
```diff
@@ -628,37 +684,56 @@ class MediaSearchRequestSourceSerializer(serializers.Serializer):
...
-        def validate_source_field(value):
-            """Check whether source is a valid source."""
-
+        def validate_source(self, value):
+            """
+            For the regular searches, split the value.lower() by comma and only return the
+            source names that are in the search controller's sources list for the media type.
+            For the collection=tag, return the value as is. It is ignored in the query builder.
+            For source and creator collections, accept the value as is, without lower-casing
+             or splitting, and check if it's in the valid source name list.
```
AetherUnbound, line 713: Nit:
```suggestion
            or splitting, and check if it's in the valid source name list.
```

api/test/unit/views/test_media_views.py
```diff
@@ -49,36 +47,6 @@ def test_retrieve_query_count(api_client, media_type_config):
...
-        pytest.param("tag/cat/", {"tag": "cat"}, id="tag"),
-        pytest.param("source/flickr/", {"source": "flickr"}, id="source"),
-        pytest.param(
-            "source/flickr/creator/cat/",
-            {"source": "flickr", "creator": "cat"},
-            id="source_creator",
-        ),
-    ],
-)
-def test_collection_parameters(path, expected_params, api_client):
```
AetherUnbound, line 65: Are these removed because they're tested by schemathesis?

AetherUnbound (commented):
This is quite the simplification! Noting that `ENABLE_COLLECTIONS=True just api/up` did not work for me, I explicitly needed to set `ENABLE_COLLECTIONS=true` in the `api/.env` file instead. I also have a few questions.

obulat:
> I've seen this as a way that some other APIs handle arrays 😮 is this something we should support too?

We could, but I don't see a reason for that. From the "Zen of Python": _There should be one-- and preferably only one --obvious way to do it._  😄

api/api/serializers/docs.py
obulat: I copy-pasted if from the old docs, and I also don't know why there are double backticks there.

api/api/docs/base_docs.py
obulat: Do you mean `audio` instead of `audio files` ? I think we've talked about using audio vs audio files a couple of times, but I don't remember the conclusions.

sarayourfriend:
To Madison's point, it really would be ideal if we were able to switch away from using `,` as a delimiter for "list" query parameters, because it makes it hard to send a value where a comma is in the name of e.g., a tag in the `tags` parameter. For example "Paris, France" is impossible to query, because our serialiser splits it to two separate tags. For Paris, France, it isn't such an issue, but for any other town named Paris (or e.g., London, Moscow, Athens, to name a few common ones off to the top of my head), you wouldn't be able to find them because the single tag "paris" will strongly outweigh anything else, and because the city in France is more popular, you'll get results from there instead.

You can verify this with Moscow, Idaho, where queries for `tags=moscow, idaho` return results for the city in Russia by the third result, despite there being heaps of results tagged `moscow` and `idaho`. It isn't even working in _that_ case.

Anyway, all of that to say, the current approach of using a comma doesn't work intuitively. You can URL encode the comma inside the tag, so long as we split on `,` before URL decoding the individual components (and presuming browsers maintain the encoding inside the tags when copying so that copy/paste works well).

It isn't something we can change for existing parameters, and whether it's a pattern we should change for future ones, is subject to Olga's comment. Consistency can be more useful than correctness.

api/conf/settings/misc.py
```diff
@@ -3,6 +3,8 @@
 from decouple import config
 
 
+ENABLE_COLLECTIONS = config("ENABLE_COLLECTIONS", cast=bool, default=False)
```
sarayourfriend, line 6: Just a nit, I think there's a way to clarify this, but not 100% sure what the goal is. The name "enable" made me think at first that it was necessary to set this to true for the parameters to work at all. Is that the case? Is it more like "stabilise_collections" or something along those lines? Enable reads more binary on/off to me, and it would be surprising to me if I hadn't read this PR to see the parameters working when I request them in production but this set to false.

obulat: Maybe `SHOW_COLLECTION_DOCS` better relays what this variable does?

In this PR, the parameters work correctly. This variable only affects the documentation. When set to False, the old documentation is shown (with the `/tag`, `/source` and `/source/sourceName/creator/` endpoints removed). When set to True, we can test the updated documentation for `creator`, `tag` and `collection` parameters.

Stabilizing would mean removing the `unstable__` prefix from the parameters, so it's not what the variable does.

obulat:
> To Madison's point, it really would be ideal if we were able to switch away from using `,` as a delimiter for "list" query parameters, because it makes it hard to send a value where a comma is in the name of e.g., a tag in the `tags` parameter. For example "Paris, France" is impossible to query, because our serialiser splits it to two separate tags. For Paris, France, it isn't such an issue, but for any other town named Paris (or e.g., London, Moscow, Athens, to name a few common ones off to the top of my head), you wouldn't be able to find them because the single tag "paris" will strongly outweigh anything else, and because the city in France is more popular, you'll get results from there instead.

_It was really useful for me to write this response as I clearly understood the distinction between different parameters and how we handle them_

**Additional search views**

The "list" queries are not used in the additional search views. Here, every `,` is considered a part of the [single] value. 

For instance, `https://localhost:50280/v1/images/?unstable__collection=tag&unstable__tag=Paris, France` will match the single `Paris, France` tag exactly. `localhost:50280/v1/audio/?unstable__collection=creator&source=jamendo&creator=Musician, guitar` would match `Musician, guitar` as the creator value. 

**Filter parameters in default or non-q search**

The list is split by a comma for the filter parameters only. There are two ways we handle such parameters
- `EnumCharField`: they split the values by a comma and then validate each value. If it's not in the choices for the field, a validation error is thrown:

https://github.com/WordPress/openverse/blob/697c1789e6268b119350121b79da6ded67081661/api/api/serializers/fields.py#L55

- The `source` filter also splits the value by a comma, checking if the parts are valid source names. However, instead of throwing a `ValidationError` on an invalid part, it just ignores it.

**Analyzing the search query**
When we search with `tags` parameter (or `creator` or `title`), without the `q` or `collection` parameter present, we do not split the value by commas. We use `simple_query_string` search, so the ES algorithms that does the splitting and analyzing.

https://github.com/WordPress/openverse/blob/697c1789e6268b119350121b79da6ded67081661/api/api/controllers/search_controller.py#L335-L349

_Note that `tags` parameter is different from the `tag` collection here (that you can access at http://localhost:50280/v1/images/?unstable__collection=tag&unstable_tag=something, something), and `creator` without `collection` is also a different search_

api/test/unit/views/test_media_views.py
obulat: I thought that we don't need these tests because no longer have extra endpoints, and we _do_ have the serializer tests that check that when a search request has the collection-related query parameters, we create the correct query.

api/conf/settings/misc.py
sarayourfriend: Gotcha! The show docs version you suggest would be great.

AetherUnbound (approved):
This looks and works great! Thanks for taking the time to explain the specifics of how we handle multiple arguments for the various parameters.

api/api/serializers/media_serializers.py
```diff
@@ -628,37 +684,56 @@ class MediaSearchRequestSourceSerializer(serializers.Serializer):
...
+            """
             allowed_sources = list(search_controller.get_sources(media_type).keys())
-            sources = value.lower().split(",")
-            sources = [source for source in sources if source in allowed_sources]
-            value = ",".join(sources)
-            return value
-
-        def validate_source(self, input_sources):
-            return self.validate_source_field(input_sources)
+            sources_list = ", ".join([f"'{s}'" for s in allowed_sources])
```
sarayourfriend, line 718: Nit: this operation is unnecessary for most queries. Can we move it into the error branch itself on line 726 and bypass it for all other search queries?

api/api/serializers/media_serializers.py
```diff
@@ -628,37 +684,56 @@ class MediaSearchRequestSourceSerializer(serializers.Serializer):
...
+            collection = self.initial_data.get(COLLECTION)
+
+            # For collection=tag, return the value as is. It is ignored in the query builder.
+            if collection == "tag":
+                return value
```
sarayourfriend, line 723: Shouldn't this raise a validation error, as they are not compatible parameters, rather than just ignored? I'm having a hard time keeping straight what we're ignoring vs validating... spreading these out between separate serializers (MediaSearchRequestSourceSerializer and the regular search request serializer) is making it hard to keep track of what is validating what and where!

Are we able to reduce down to a single serializer, by chance? Or could the validation be built into a new `SourceField` that validates differently based on the strategy?

api/conf/settings/misc.py
sarayourfriend: @obulat Will you implement this change to the environment variable?

api/test/unit/views/test_media_views.py
sarayourfriend: It's a bit different, unit vs integration level tests. Integration level tests would be nice for these complex endpoints because it's easier to build the real edge cases into the tests. Schemathesis might do that too, but I don't think this style of test is duplicative of the unit/serializer level tests. Just because we have serializer tests to test serializer output, doesn't mean those tests also ensure the serializer output integrates with the actual search requests themselves in all instances. Integration tests do that testing of the glue.

Incidentally, integration tests could obviate the need for some of those unit tests, but only reliably so if we had coverage collection.

With the complexity of the endpoints parameters, I think it's not a bad idea to have tests of the sort removed here that send a variety of query parameter configurations and assert at minimum whether the result is a successful search or not.

sarayourfriend (changes requested):
Requesting changes just for the environment variable name, but also asking for clarification on some things. The serializer changes we made to accommodate the different endpoints sharing code, now seem to be unnecessary, as it's all one endpoint. The abstraction is no longer necessary, and makes the code much harder to understand (_especially_ now that it's all one endpoint). It's not necessarily a blocker, but some interdependencies of the validation are hard to keep track of now when needing to jump between multiple serializers.

api/api/serializers/media_serializers.py
obulat: > MediaSearchRequestSourceSerializer and the regular search request serializer

These two serializers were separate before any additional search views project work. The reason for that, as I understand, is that `media_type` is not available in the regular media serializer, and it's necessary for the source to check the relevant source list (audio sources for audio and image sources for image). The source serializer is dynamic.

obulat:
> The serializer changes we made to accommodate the different endpoints sharing code, now seem to be unnecessary, as it's all one endpoint. The abstraction is no longer necessary, and makes the code much harder to understand (_especially_ now that it's all one endpoint).

I am a bit confused about which serializer changes this comment is.

If the comment is about the separate **source serializer and `MediaSearchRequestSerializer`**, they were separate before any work on the additional search views began.

The main `MediaSearchRequestSerializer` does not have the access to the media type, and has fields and validators that are the same for all media types.
Source serializer is separate because it needs to have access to the media type. The help text uses it to create the link ("valid values are source_names from the stats endpoint: https://api.openverse.engineering/v1/audio/stats/."). 
The source validator uses the media type to get the correct list of valid sources. The validator function itself though is the same for both media types.

I tried moving the source serializer into the `MediaSearchRequestSerializer` by adding the `media_type` to the serializer context. This way, we do have access to the media type inside the search request serializer, and can simplify the source serializer, as well as validating of the `internal__index`. 

@dhruvkb, as a person who worked on this code, do you think there could be any problems with using the `context` for serializer media type?

If the comment is about `MediaListRequestSerializer` that was added specifically for the additional search views, I removed it completely.

sarayourfriend:
The only issue with using context is if there are cases where the context cannot be defined, but the serialiser needs it to initialise. I don't think we have that in our application, we always know whether we're working on an image or audio request at the time we perform deserialization of request parameters. To me, it's a perfectly reasonable and DRF-y use of serializer context :+1: 

I'll re-review this PR in depth later today. Thanks for the explanation of the different serializers and how they worked together. I had not remembered the separation pre-existed the collections work, but now that validation is so much more complex, I think it does make sense to tighten the relationships but a bit, which it sounds like you've done. I look forward to reviewing the changes later today.

sarayourfriend:
Review this now.

sarayourfriend (changes requested):
The serializer changes are excellent, and the cleanup and code deletions you've managed in this PR are awesome! Even with the addition of a large amount of new documentation strings, this PR still nets fewer lines of code :tada: 

I have one thing that might seem small, but I think is actually a very significant issue with the template strings. Here is a patch that should resolve it:

```diff
diff --git a/api/api/docs/audio_docs.py b/api/api/docs/audio_docs.py
index dfaadc78f..06f5d573c 100644
--- a/api/api/docs/audio_docs.py
+++ b/api/api/docs/audio_docs.py
@@ -42,10 +42,10 @@ from api.serializers.provider_serializers import ProviderSerializer
 serializer = AudioSearchRequestSerializer(context={"media_type": "audio"})
 audio_filter_fields = fields_to_md([f for f in serializer.field_names if f != "q"])
 
-audio_search_description = SEARCH_DESCRIPTION % {
-    "filter_fields": audio_filter_fields,
-    "media_type": "audio files",
-}
+audio_search_description = SEARCH_DESCRIPTION.format(
+    filter_fields=audio_filter_fields,
+    media_type="audio files",
+)
 
 search = custom_extend_schema(
     desc=audio_search_description,
diff --git a/api/api/docs/base_docs.py b/api/api/docs/base_docs.py
index c80fd333a..96e3f6cee 100644
--- a/api/api/docs/base_docs.py
+++ b/api/api/docs/base_docs.py
@@ -144,7 +144,7 @@ SEARCH_DESCRIPTION_DEFAULT = """
 Return audio files that match the query.
 
 This endpoint allows you to search within specific fields, or to retrieve
-a collection of all %(media_type)s from a specific source, creator or tag.
+a collection of all {media_type} from a specific source, creator or tag.
 Results are paginated on the basis of the `page` parameter. The `page_size`
 parameter controls the total number of pages.
 
@@ -163,7 +163,7 @@ according to specific needs.
 By default, this endpoint performs a full-text search for the value of `q` parameter.
 You can search within the `creator`, `title` or `tags` fields by omitting
 the `q` parameter and using one of these field parameters.
-These results can be filtered by %(filter_fields)s.
+These results can be filtered by {filter_fields}.
 
 The default search results are sorted by relevance.
 
@@ -180,11 +180,11 @@ additions appearing first. The filters such as `license` are not available for c
 """
 
 SEARCH_DESCRIPTION_COLLECTIONS_DISABLED = """
-Search %(media_type)s using a query string.
+Search {media_type} using a query string.
 
 By using this endpoint, you can obtain search results based on specified
 query and optionally filter results by
-%(filter_fields)s.
+{filter_fields}.
 
 Results are ranked in order of relevance and paginated on the basis of the
 `page` param. The `page_size` param controls the total number of pages.
diff --git a/api/api/docs/image_docs.py b/api/api/docs/image_docs.py
index f193308ec..5770f02c7 100644
--- a/api/api/docs/image_docs.py
+++ b/api/api/docs/image_docs.py
@@ -44,10 +44,10 @@ from api.serializers.provider_serializers import ProviderSerializer
 serializer = ImageSearchRequestSerializer(context={"media_type": "image"})
 image_filter_fields = fields_to_md([f for f in serializer.field_names if f != "q"])
 
-image_search_description = SEARCH_DESCRIPTION % {
-    "filter_fields": image_filter_fields,
-    "media_type": "images",
-}
+image_search_description = SEARCH_DESCRIPTION.format(
+    filter_fields=image_filter_fields,
+    media_type="images",
+)
 
 search = custom_extend_schema(
     desc=image_search_description,
diff --git a/api/api/serializers/docs.py b/api/api/serializers/docs.py
index 631a8eab7..5c333f2b6 100644
--- a/api/api/serializers/docs.py
+++ b/api/api/serializers/docs.py
@@ -40,17 +40,17 @@ Should be used with `{TAG}`, `source` or `creator`+`source`"""
 
 EXCLUDED_SOURCE_HELP_TEXT = """
 A comma separated list of data sources to exclude from the search.
-Valid values are `source_name`s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/.
+Valid values are `source_name`s from the stats endpoint: {origin}/v1/{media_path}/stats/.
 """
 SOURCE_HELP_TEXT_COLLECTIONS_DISABLED = """
 A comma separated list of data sources; valid values are
-`source_name`s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/."""
+`source_name`s from the stats endpoint: {origin}/v1/{media_path}/stats/."""
 
 SOURCE = """
 For default search, a comma separated list of data sources.
 When the `collection` parameter is used, this parameter only accepts a single source.
 
-Valid values are `source_name`s from the stats endpoint: %(origin)s/v1/%(media_path)s/stats/.
+Valid values are `source_name`s from the stats endpoint: {origin}/v1/{media_path}/stats/.
 """
 
 SOURCE_HELP_TEXT = (
diff --git a/api/api/serializers/media_serializers.py b/api/api/serializers/media_serializers.py
index 3b4e94f7f..809562cba 100644
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -303,8 +303,8 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
             "media_path": media_path,
         }
 
-        self.fields["source"].help_text = SOURCE_HELP_TEXT % variables
-        self.fields["excluded_source"].help_text = EXCLUDED_SOURCE_HELP_TEXT % variables
+        self.fields["source"].help_text = SOURCE_HELP_TEXT.format(**variables)
+        self.fields["excluded_source"].help_text = EXCLUDED_SOURCE_HELP_TEXT.format(**variables)
 
     def is_request_anonymous(self):
         request = self.context.get("request")
```

Really want to emphasise that I'm not trying to be nit-picky here, and recognise that it's rare to block on something that may seem like preference, but `%` formatted strings are so tricky and unusual in new Python code that I think it is worth doing in this case for the reason's I've shared.

api/api/serializers/media_serializers.py
```diff
@@ -106,10 +120,15 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
...
         "q",
+        "source",
+        "excluded_source",
         "license",
         "license_type",
         "creator",
         "tags",
+        # Uncomment after Additional search views project is launched
+        # "collection",
+        # "tag",
```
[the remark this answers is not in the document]
obulat: I updated the comment to link to the GitHub [issue](https://github.com/WordPress/openverse/issues/3919) to remove the `unstable__` prefix. So, these will be uncommented _after_ the prefix is gone.

I am not sure what the best process for launching would be. This is one of the scenarios I can think of:

1. Switch the frontend `additional_search_views` flag on, without fully removing the flag, and set `SHOW_COLLECTION_DOCS` env variable in the API to True in prod. This means that the users on the frontend can see the additional search views as fully-launched, and the API users can start using them with the `unstable__` parameters.
2. Test how well both API and frontend work in prod for ~ 1 week. If needed, we can revert using the frontend flag and the API env variable.
3. If everything is successful, add the stable parameters to the frontend.
4. Stabilize the API parameters (remove the conditionals from the additional search views documentation and the `unstable__` prefix from `tag` and `collection` parameters)
5. Remove the additional frontend parameters and the feature flag.

These lines will be uncommented after step 4, then. So, they won't need the prefix. And the unstable versions will be seen after step 1, but will be removed after step 4.

api/api/serializers/media_serializers.py
```diff
@@ -132,28 +158,14 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
         help_text="A query string that should not exceed 200 characters in length",
         required=False,
     )
-    license = serializers.CharField(
-        label="licenses",
-        help_text=make_comma_separated_help_text(LICENSE_GROUPS["all"], "licenses"),
+    source = serializers.CharField(
+        label="provider",
```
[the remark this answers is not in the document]
obulat: I would really appreciate if you could create an issue for that. `provider` label is not very prominent in the documentation, but it is confusing. I never noticed it before 😆 
Screenshot 2024-03-27 at 9 58 30 AM

api/api/serializers/media_serializers.py
```diff
@@ -245,6 +290,22 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
...
+                "The media request serializer's `media_type` context variable must be set."
+            )
+        media_path = {"image": "images", "audio": "audio"}[self.media_type]
+        variables = {
+            "origin": settings.CANONICAL_ORIGIN,
+            "media_path": media_path,
+        }
+
+        self.fields["source"].help_text = SOURCE_HELP_TEXT % variables
+        self.fields["excluded_source"].help_text = EXCLUDED_SOURCE_HELP_TEXT % variables
```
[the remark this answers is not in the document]
obulat: Thank you so much for explanation and the patch! I didn't remember how to make string format lazy, so I found a quick fix that wasn't optimal.
I applied your patch.

sarayourfriend (approved):
LGTM! I'll create the issues for the field labels tomorrow. I need to log off for the day now :slightly_smiling_face:

PR #3890, opened 2024-03-09
Add point_alias DAG and add alias params to create_new_es_index DAGs

stacimc:
## Fixes

Fixes #3493 by @stacimc

## Description

This PR adds DAGs for pointing a `target_alias` at an existing elasticsearch index, with one DAG each for staging and production. Before the alias is applied, it will first be removed from the index to which it currently points (if applicable). There is also an option to delete this pre-existing index if desired. This was done by slightly extending the existing `point_alias` taskgroup from the `create_proportional_by_source_staging_index` DAG.

The `point_alias` task group is then used in a separate `point_<environment>_alias` DAG factory, and also added as a step to the `create_new_<environment>_es_index` DAGs to add functionality to those DAGs as well.

### Test that `create_proportional_by_source_staging_index` DAGs still work

1. Run the DAG locally with default params and verify it all passes. In elasticvue you should see a new index with a name like `audio-50-percent-proportional-20240308t235423` and the alias` audio-subset-by-source`. 
2. Run the DAG a second time and verify all passes. This time `audio-subset-by-source` should have been *removed* from the index created in the previous step, and applied to a new index with a later timestamp.

### Test the new DAG

1. Run `point_staging_alias` locally. Pass in an existing index and an alias that does not currently exist. Verify that the alias gets added to the selected index. `remove_alias_from_existing_index` and `delete_old_index` steps will be skipped.
2. Run again using the same alias, pointed to a different index. This time `remove_alias_from_existing_index` will not be skipped; the alias should be removed from the old index before being applied to the new one. `delete_old_index` is still skipped.
3. Run again using the same alias and another index but also enable `should_delete_old_index`. This time no tasks should be skipped, and the alias should not only be moved to the new index but the old index should be deleted entirely.

### Test create_new_<x>_es_index DAGs

1. Run `create_new_staging_es_index` and do NOT pass in a target_alias. It should pass, and all the `point_alias` steps should be skipped.
2. Run again, but this time pass in an alias that does not already exist. Verify that it is added to your new index, and `remove_alias_from_existing_index` and `delete_old_index` steps are skipped
3. Run again with the same alias. Verify that it is added to your new index and removed from the index created in the previous step.
4. Run again with the same alias, and `should_delete_old_index` enabled. Verify it is added to the new index, and the index created in step 3 is deleted.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
[...]

PR #3910, opened 2024-03-13
Add elasticsearch concurrency tags for Airflow

### LINKED ISSUE #3891: Improve support for "concurrency pools" in Elasticsearch DAGs
## Problem

We have a large number of DAGs that operate on Elasticsearch or the underlying databases. These DAGs must not run concurrently, so we have two reusable tasks (`prevent_concurrency_with_dags` and `wait_for_external_dags`) that each accept a list of dag_ids and either (a) immediately fail if any of those dags are running or (b) wait until none of those dags are running, respectively.

The problem is that every time a new elasticsearch DAG is added, all the other DAGs that affect that environment must also be updated to fail fast/wait on it. It is easy to miss dependencies, and indeed a few were missed already:

* `create_new_staging_es_index` DAG needs to prevent concurrency with `create_proportional_by_source_staging_index`
* `staging_database_restore` needs to wait on `create_proportional_by_source_staging_index`

## Description

⚠️ Messy description, jotting down some in-progress thoughts. ⚠️ 

I've prototyped some of this and think this would work:

* Create two new dag tags used to define "concurrency pools" for staging and production respectively. Something like "staging_elasticsearch_concurrency" and "production_elasticsearch_concurrency". Apply them to all the appropriate DAGs:
  * staging
    * staging_database_restore
    * recreate_full_staging_index
    * point_staging_alias
    * create_new_staging_es_index
    * create_proportional_by_source_staging_index
  * production
    * both data refreshes
    * both filtered index creation dags
    * create_new_production_es_index
    * point_production_index 
* Update the `prevent_concurrency_with_dags` and `wait_for_external_dags` task groups to be a single task (rather than dynamically mapping a task for each external dag id). Instead of taking a list of external_dag_ids, it takes the tag to check for
  * It then queries for Dags with that tag (`session.query(DagModel).filter(DagModel.tags.any(DagTag.name == tag)).all()` in order to get the list of external dags. It excludes the DAG from which the task was called
  * It should also throw an error if the DAG from which the task was called _is not itself in the pool_. This prevents you from forgetting to add it
  
(Note: actual implementation ended up being very similar to this idea but not quite)

## Alternatives

I prototyped this using tags, but as I write it out it seems obvious that we should literally use Airflow **pools** instead. I think that would work with the exact same approach.

stacimc:
## Fixes

Fixes #3891 by @stacimc 

Blocked, this needs to be rebased on #3890 when merged (missing latest changes from that PR)
TODO should also add an exception for relationship between staging_db_restore and the proportional index DAG.

## Description

We now have quite a few DAGs that all operate on Elasticsearch or the underlying databases, and therefore must not run concurrently. We've been using the `prevent_concurrency_with_dags` and `wait_for_external_dags` utils to accomplish this in each DAG, by passing in a hard-coded list of the external DAGs we want to either wait on or otherwise prevent concurrency with. Since this list is hard-coded for each DAG, it is extremely easy to miss dependencies especially as DAGs are added and changed. A few were missed in the implementation of the Search Relevancy project so far.

This PR introduces two new Tags that are used to identify DAGs as being part of a "concurrency group": `production_es_concurrency` and `staging_es_concurrency`. Each DAG in these groups have been given the appropriate tag (so for example, `create_new_staging_es_index` and `staging_database_restore` both have the `staging_es_concurrency` tag). The concurrency tasks have been updated to accept the name of the tag you want to select on, rather than a hard coded list, ie:

```
prevent_concurrency = prevent_concurrency_with_dags_with_tag(
    tag=STAGING_ES_CONCURRENCY_TAG,
)
```

^ This will select all the DAGs that have the given `tag`, *except for the running DAG/the DAG that called the task*, and fail immediately if any of them are running.

Screenshot 2024-03-13 at 10 35 10 AM

So now if you add a new ES dag, rather than having to manually update a bunch of lists of DAG dependencies on every single other ES dag, you just have to remember to add the appropriate `tag` to your new DAG. The concurrency util will raise an error if you forget, as well.

### Additional thoughts/alternatives 

[This comment](https://github.com/WordPress/openverse/issues/3891#issuecomment-1988961899) on the issue explains why it wasn't possible to use the existing feature of Airflow pools to do this. Tags also have the benefit of being very visible in the UI and easy to filter by.

## Testing Instructions

These are the intended concurrency groups:

  * `staging_es_concurrency`
    * staging_database_restore
    * recreate_full_staging_index
    * point_staging_es_alias
    * create_new_staging_es_index
    * create_proportional_by_source_staging_index
  * `production_es_concurrency`
    * both data refreshes
    * both filtered index creation dags
    * create_new_production_es_index
    * point_production_es_index 

For each DAG, run the DAG and check that the appropriate DAGs were waited on/prevented_concurrency_with. You can do this by looking at the logs for the `get_dags_in_<tag>_group` task, and then checking that the correct number of mapped tasks were made. (Example: `create_proportional_by_source_staging_index` is in the staging group which has 6 dags. It has a `prevent_concurrency_with_dag` task generated for every other DAG in the group except itself).

Airflow 2.9 is supposed to include the ability to prove meaningful names for dynamically mapped tasks (instead of just a numerical index), which will make this much easier to read at-a-glance once we upgrade!

Screenshot 2024-03-13 at 10 37 03 AM

Note: The data refreshes are a little special. They use the `SingleRunExternalDagSensor` to wait on each other, so in their `wait_for_es_dags` task they should only wait on the _other_ production ES Dags (5 dags total -- the two filtered index, create_new_production_es_index, and point_production_es_index).

### Test the actual concurrency utils still works

We should also test the actual concurrency for at least a few options. The easiest way to do this is to trigger them in sequence through the Airflow CLI. Make sure the relevant DAGs are turned on in Airflow, then run `just catalog/shell` and:

```
# For example, this would trigger the new staging index DAG and then immediately trigger staging db restore
# We would expect staging_database_restore to pass all of its other `wait` tasks, but poke a few times on the task that
# waits for `create_new_staging_es_index` and only continue once that DAG completes.
> airflow dags trigger create_new_staging_es_index && airflow dags trigger staging_database_restore
```

A general tip for remembering how the concurrency checks should work: scheduled DAGs (data refreshes, staging database restore) will wait if a conflicting DAG is running, and resume once it completes. DAGs that are only run manually (create_new_es_index, create_filtered_index, recreate_full_staging_index) fail immediately if a conflicting DAG is running.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/elasticsearch.py
+++ b/catalog/dags/common/elasticsearch.py
@@ -3,10 +3,11 @@ from datetime import timedelta
 from typing import Literal, Union
 
 from airflow.decorators import task, task_group
+from airflow.exceptions import AirflowSkipException
 from airflow.models.connection import Connection
 from airflow.providers.elasticsearch.hooks.elasticsearch import ElasticsearchPythonHook
 from airflow.sensors.base import PokeReturnValue
-from airflow.utils.trigger_rule import TriggerRule
+from elasticsearch.exceptions import NotFoundError
 
 from common.constants import REFRESH_POKE_INTERVAL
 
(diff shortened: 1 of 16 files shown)


stacimc:
Closing in favor of #3921 to disentangle merge issues

Issue #3633, opened 2024-01-07
Establish preliminary Django admin access control for content moderator users


sarayourfriend:
## Description

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#preliminary-access-control

AetherUnbound:
The code change for this was pushed in #3899, and I've added the `Content Moderator` group with the following permissions:

![image](https://github.com/WordPress/openverse/assets/10214785/9fa92f62-f3b3-4777-a62f-258e05f311f4)

I'm not sure if `view` on the media tables themselves need to be added - but we can always add that down the line.

PR #3914, opened 2024-03-14
Content moderator user preferences admin view

### LINKED ISSUE #3634: Add content moderator user preferences admin view
## Description

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#add-user-preferences-admin-view

AetherUnbound:
## Fixes

Fixes #3634 by @AetherUnbound

## Description

This PR adds a new `UserPreferences` ["profile" model](https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#extending-user) to the project, along with a new Django Admin view for editing a user's preferences. The admin view only ever displays the currently logged in user's preferences, and it has a custom form set up to set (or unset) a specific value within the `jsonb` column used for storing preferences.

In a bit of a follow up to #3633, I've also added some startup logic to create both an additional `moderator` user with staff access to the Django Admin site and a "Content Moderators" group which has the permissions we're expecting for that group. The upshot of this is that all of the testing below shouldn't require any additional user or group creation! Yay!

This was really tricky to figure out how to match the expectations of the IP while doing all this within the Django framework. I haven't done a _ton_ of Django work, so if there are more idiomatic ways to accomplish what I've done, please suggest them!

## Screenshots

**Moderator user view**
image

**Moderator preferences edit form**
image

## Testing Instructions

I added some tests specifically for the admin UI form - those should pass (and hopefully make sense).

1. Run `just api/dbshell`, then execute `select u.username, p.preferences from auth_user u left join api_userpreferences p on p.user_id = u.id;`. There should be 3 users and all preferences should be blank.
2. Log in to the admin site (http://localhost:50280/admin/) using `deploy`/`deploy`. Observe the new "USER PREFERENCES" section with the "My Preferences" entry.
3. Click on "My Preferences". Observe that only `deploy's preferences` shows up in the list.
4. Click on `deploy's preferences`. Observe that the form only includes a check box for "Blur images".
5. Log out and log in again using `moderator`/`deploy`. Observe that only the report/sensitivity tables and user preferences sections show up.
6. Follow the steps above to edit the user preferences (verifying only the moderator user shows up), only this time unchecking the "Blur images" box. Click save.
7. Run the select above once more. The `preferences` value for the `moderator` user should be `{"moderator": {"blur_images": false}}`
8. Edit the preferences once more time and re-enable blurring images. Run the select above and verify the preferences saved in the database are now `{"moderator": {"blur_images": true}}`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/admin/forms.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def _get_blur_images(self):
+        mod_preferences = self.instance.preferences.get("moderator", {})
+        return mod_preferences.get("blur_images", True)
+
+    @staticmethod
+    def _set_blur_images(instance, value):
+        mod_preferences = instance.preferences.get("moderator", {})
+        mod_preferences["blur_images"] = value
+        instance.preferences = instance.preferences | {"moderator": mod_preferences}
```
sarayourfriend, line 25: Theoretically, this is mixing concerns a bit between the form and the model. The logic of how to read and save data on the model should live in the model, with the form ideally able to do simple gets and sets on model attributes. That's also true for the correct default values of any part of the model's data, that should all live on the model class.

Because this is a "virtual" field, we need to implement it on the model with `@property`, but it lets us move the logic for getting and setting, including expected defaults, all onto the model, and reduce the form's concerns. There's also some unnecessary code in the `save` method of the form we can remove:

```diff
diff --git a/api/api/admin/forms.py b/api/api/admin/forms.py
index 48a43426c..2ac566773 100644
--- a/api/api/admin/forms.py
+++ b/api/api/admin/forms.py
@@ -12,22 +12,8 @@ class UserPreferencesAdminForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.initial["blur_images"] = self._get_blur_images()
-
-    def _get_blur_images(self):
-        mod_preferences = self.instance.preferences.get("moderator", {})
-        return mod_preferences.get("blur_images", True)
-
-    @staticmethod
-    def _set_blur_images(instance, value):
-        mod_preferences = instance.preferences.get("moderator", {})
-        mod_preferences["blur_images"] = value
-        instance.preferences = instance.preferences | {"moderator": mod_preferences}
+        self.initial["blur_images"] = self.instance.blur_images
 
     def save(self, commit=True):
-        instance: UserPreferences = super().save(commit=False)
-        blur_images = self.cleaned_data.get("blur_images", True)
-        self._set_blur_images(instance, blur_images)
-        if commit:
-            instance.save()
-        return instance
+        self.instance.blur_images = self.cleaned_data.get("blur_images")
+        return super().save(commit=commit)
diff --git a/api/api/models/moderation.py b/api/api/models/moderation.py
index bf4677fb8..fc818a239 100644
--- a/api/api/models/moderation.py
+++ b/api/api/models/moderation.py
@@ -11,6 +11,22 @@ class UserPreferences(models.Model):
     def __str__(self):
         return f"{self.user.username}'s preferences"
 
+    @property
+    def moderator(self):
+        return self.preferences.get("moderator", {})
+
+    @moderator.setter
+    def moderator(self, value):
+        self.preferences["moderator"] = value
+
+    @property
+    def blur_images(self):
+        return self.moderator.get("blur_images", True)
+
+    @blur_images.setter
+    def blur_images(self, value):
+        self.moderator |= {"blur_images": value}
+
 
 @receiver(post_save, sender=User)
 def create_or_update_user_profile(sender, instance, created, **kwargs):
```

It's all the same relative complexity, but moves it to the "right" place, with the goal of making it possible for any code to read `blur_images` from the model with the exact same behaviour _vis-a-vis_ the handling of retrival off the model's data and correct defaults. Otherwise, places that referenced the model would need to re-implement the same retrieval and default handling logic the form currently implements.

sarayourfriend (changes requested):
Looking really good! The approach you took makes sense coming from a low amount of Django background and less context on how code sharing in Django works. I've left a comment with a diff showing the specific reorganisation of the code that would align with Django's mental model. All of which ends up with the exact same behaviour (and almost the same code) that you've implemented, just in different locations to make it align with Django's separation of concerns between models and forms. That separation of concerns somewhat matches the separation between model and view/controller in the MVC pattern ([though Django's FAQs point out that Django complicates the VC part of MVC by having the framework do a lot of the typical "controller" stuff](https://docs.djangoproject.com/en/dev/faq/general/#django-appears-to-be-a-mvc-framework-but-you-call-the-controller-the-view-and-the-view-the-template-how-come-you-don-t-use-the-standard-names)).

Hope that helps!

sarayourfriend:
Oh, the only other thing, to make this more usable, it'd be good if the list view (`http://localhost:50280/admin/api/userpreferences/`) is skipped when clicking "my preferences" from the model list view. There's only ever one item in the list, the logged-in user's preferences, so it's an unnecessary interstitial page. I tried this out locally, and it looks like this works. The idea is to just redirect requests to the changelist view to the change view.

```diff
diff --git a/api/api/admin/__init__.py b/api/api/admin/__init__.py
index 3cb3c31cb..581a6961e 100644
--- a/api/api/admin/__init__.py
+++ b/api/api/admin/__init__.py
@@ -1,6 +1,8 @@
 from django.contrib import admin
 from django.contrib.auth.admin import GroupAdmin, UserAdmin
 from django.contrib.auth.models import Group, User
+from django.http.response import HttpResponseRedirect
+from django.urls import reverse
 
 from api.admin.forms import UserPreferencesAdminForm
 from api.admin.site import openverse_admin
@@ -122,3 +124,7 @@ class IndividualUserPreferencesAdmin(admin.ModelAdmin):
             # the changelist itself
             return True
         return obj.user == request.user
+
+    def changelist_view(self, request, extra_context=None):
+        obj = self.get_queryset(request).first()
+        return HttpResponseRedirect(reverse("admin:api_userpreferences_change", args=[obj.id]))
```

api/api/admin/forms.py
AetherUnbound: This makes a _ton_ of sense, thanks for taking the time to explain it to me!! It's interesting that there are so many ways to accomplish this, but my own infrequency in working with Django means that I'm not as clued into thinking this way by default. These changes seem a lot better, I've incorporated them 🚀

AetherUnbound:
Thanks so much for the feedback @sarayourfriend, and the extra bit of code to jump straight into the preferences rather than the list view! I wanted to try and make that the case but didn't even know it was possible. I've made those changes and rebased, this should be good to re-review 😄

api/api/models/moderation.py
```diff
@@ -0,0 +1,35 @@
...
+class UserPreferences(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    preferences = models.JSONField(default=dict)
+
+    def __str__(self):
+        return f"{self.user.username}'s preferences"
+
+    @property
+    def moderator(self):
+        return self.preferences.get("moderator", {})
```
sarayourfriend, line 16: I realised yesterday thinking about this PR that a small bug could come from this particular implementation of the `moderator` getter, caused by sloppy object reference handling (my bad!):

```py
assert "moderator" not in prefs.preferences
prefs.moderator["property"] = "whatever"
prefs.save()
```

`prefs.moderator` is the default case from `get`, rather than the actual object on `self.preferences`, and we haven't "bound" that default object to `self.preferences`. Therefore, this won't actually change the `preferences` object at all, because the moderator object won't be saved into the `preferences` object. The only reason there isn't a bug in the code we have already is because the `blur_images` setter uses `|=` which will call the setter for moderator. But if you assigned directly to `self.moderator["blur_images"]`, it wouldn't work.

We can fix this by:

```suggestion
        if "moderator" not in self.preferences:
            self.preferences["moderator"] = {}
        
        return self.preferences["moderator"]
```

That way, the return value of the moderator getter is a reference to the object on `self.preferences` and never a "free" object.

api/api/admin/__init__.py
```diff
@@ -134,3 +145,32 @@ class AccessTokenAdmin(admin.ModelAdmin):
...
+        return qs.filter(user=request.user)
+
+    def has_change_permission(self, request, obj=None):
+        if not obj:
+            # the changelist itself
+            return True
+        return obj.user == request.user
+
+    def changelist_view(self, request, extra_context=None):
+        obj = self.get_queryset(request).first()
```
sarayourfriend, line 173: I think this will cause an issue in live environments with the existing users, because they won't have had the `create` receiver triggered that will create a preferences instance for _new_ users. We could solve this with a data migration that creates a user preferences instance for all existing users, and add it to the `0058` migration.

sarayourfriend (changes requested):
Haven't tested the new code locally, but I realised a couple very small potential bugs while thinking about and re-reading the code.

api/api/migrations/0058_userpreferences.py
```diff
@@ -0,0 +1,34 @@
...
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def create_user_preferences(apps, schema_editor):
+    UserPreferences = apps.get_model("api", "UserPreferences")
+    User = apps.get_model("auth", "User")
+    for user in User.objects.all():
+        if not hasattr(user, "userpreferences"):
```
sarayourfriend, line 12: Just a nit (and mostly a tip for the future, really), but we would ideally do this kind of filtering at the ORM level, so that it's executed in the DB, rather than in Python. Calling `.all()` on a model manager will load literally every single instance of the model into memory. That might not matter in this case, because I don't think we have many User instances in our live environments, but at some future date it could.

In Django ORM you could do that a couple of ways, in order of best to worst options. You can definitely use the first one in this case, but for more complex cases, the others might be necessary. For all of these we would still want to use QuerySet slices to avoid loading all selected instances into memory: even a subset could be huge. These alternative queries reduce the selected rows, but don't necessarily solve the problem of loading too many instances into memory all at once.

So for this case, you'd just use `isnull` to do an outer join:

```py
User.objects.filter(userpreferences__isnull=True)
```

Which turns into this SQL:

```sql
SELECT
    "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined"
FROM
    "auth_user"
    LEFT OUTER JOIN
        "api_userpreferences" ON ("auth_user"."id" = "api_userpreferences"."user_id")
WHERE
    "api_userpreferences"."id" IS NULL
```

I don't think there's any reason _not_ to use that approach for this particular kind of one-to-one relationship.

**For other circumstances**, you can also do the following, to avoid filtering in Python.

With an inner-select and not-in (probably the simplest, but potentially also problematic):

```py
User.objects.exclude(pk__in=UserPreferences.objects.values_list("user", flat=True))
```

Which turns into this SQL:

```sql
SELECT
    "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined"
FROM
    "auth_user"
WHERE NOT (
    "auth_user"."id" IN (SELECT U0."user_id" FROM "api_userpreferences" U0)
)
```

With a join:

```py
from django.db.models import Exists, OuterRef

User.objects.filter(Exists(UserPreferences.objects.filter(user=OuterRef("pk"))))
```

which turns into this SQL:

```sql
SELECT
    "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined"
FROM
    "auth_user"
WHERE
    EXISTS (
        SELECT 1 AS "a" FROM "api_userpreferences" U0 WHERE U0."user_id" = ("auth_user"."id") LIMIT 1
    )
```

Any of these are ideally combined with QuerySet slicing to avoid loading the full list into memory. Any of these query sets can be used like so, to achieve chunking:

```py
qs = User.objects.whatever()
slice_size = 10

while (slice := qs[:slice_size]):
    for user in slice:
        # do whatever
```

That has the added benefit of being a noop if `qs` is empty. QueryString slices are already evaluated, so assigning it to `slice` saves any second query, and `bool(empty_qs)` evaluates to False, so the while loop combined with walrus operator works great here.

**But anyway, I don't think this will be a problem for this PR**. Each `ThrottledApplication` instance is _able_ to have a user (https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L102-L108) but it looks like that isn't something the library automatically enforces. I went ahead and queried production API database to be sure:

```
deploy@localhost:openledger> select count(*) from api_throttledapplication where user_id is not null;
+-------+
| count |
|-------|
| 0     |
+-------+
SELECT 1
Time: 0.227s
deploy@localhost:openledger> select count(*) from api_throttledapplication where user_id is null;
+-------+
| count |
|-------|
| 5031  |
+-------+
SELECT 1
Time: 0.233s
```

And more succinctly:

```
deploy@localhost:openledger> select count(*) from auth_user;
+-------+
| count |
|-------|
| 2     |
+-------+
SELECT 1
Time: 0.227s
```

So anyway, definitely not a problem in this particular case, but just wanted to share how to avoid this in the future in case it ever would be a problem, and give examples of how it can be achieved. Django ORM is very powerful but terribly complex :cold_sweat:

sarayourfriend (approved):
LGTM! I left one big comment but it is not a request for a change, which I've explained there. Just an "FYI" for the future. Apologies if it's all information you're familiar with, but the in-Python filtering on a `.all()` QuerySet pulled entirely into memory all at once is an easy and common mistake to make for folks new to Django, the ORM needing a "here be dragons" disclaimer and all.

sarayourfriend:
Oh, one thing we might want in the future, if we do get outside folks working on reports, is the ability to view a specific user's preferences from the User admin. It should be easy to do with [`InlineModelAdmin`](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#inlinemodeladmin-objects) and would make it possible for Openverse admin to "debug" preferences for a particular moderator if they ever became more complex, considering we don't have any "become user" functionality.

That's for a separate issue and only if we ever needed it. As far as we know, it isn't going to be a problem for the forseeable future.

api/api/migrations/0058_userpreferences.py
AetherUnbound: THANK YOU for taking the time to write this all out! I'll definitely be referencing it in the future 😄 I took the easiest approach here because I _knew_ we had very few user accounts at this point, so iterating over all of them would not be a problem. I appreciate the reminder about how this "giant hammer" approach could be problematic in the future!

api/api/models/moderation.py
```diff
@@ -0,0 +1,38 @@
+from django.contrib.auth.models import User
+from django.db import models
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+
+class UserPreferences(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
```
dhruvkb, line 8: Barring a few exceptions and old code, the API seems to conventionally use strings to reference target in relationships.

```suggestion
    user = models.OneToOneField(to="auth.User", on_delete=models.CASCADE)
```

api/api/migrations/0058_userpreferences.py
```diff
@@ -0,0 +1,34 @@
...
+            UserPreferences.objects.create(user=user)
+            user.userpreferences.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('api', '0057_alter_sensitiveaudio_options'),
```
dhruvkb, line 21: This change, combined with the rename of this migration file, might avoid the need for rebasing the PR.
```suggestion
        ('api', '0058_moderation_decision'),
```

dhruvkb (approved):
I have one request for a change and also a PR #4034 with some bigger changes towards the admin UI for the model.

api/api/migrations/0058_userpreferences.py
sarayourfriend: Please rebase the PR rather than trying to avoid it! There's no reason to avoid rebasing and it's the clearest and most consistent way to avoid shenanegans with migration files, which we should prioritise avoiding.

dhruvkb: Sorry about that. I misspoke (fixed now). I suggested the change to avoid having to delete and remake the migration, rebasing the PR is absolutely recommended to make sure everything works as expected against the current state of `main`.

AetherUnbound: Updated in 28f8b6b2d02ca529c6e8cb63082252f42146e838

api/api/models/moderation.py
AetherUnbound: Updated in 6762fae2db5a62a65136f9d9d74a73f0e0a64d49

AetherUnbound:
I _believe_ I've addressed all the issues! Will wait for CI to pass, then I'll merge this 😄

api/api/admin/__init__.py
```diff
@@ -127,3 +138,32 @@ class AccessTokenAdmin(admin.ModelAdmin):
...
+
+
+@admin.register(UserPreferences)
+class IndividualUserPreferencesAdmin(admin.ModelAdmin):
+    """
+    Model admin for showing user preferences. This should only ever show the
+    currently logged-in user's preferences
+    """
+
+    can_delete = False
```
dhruvkb, line 150: This line doesn't do anything. It can be removed.

```suggestion
```

api/api/admin/__init__.py
```diff
@@ -127,3 +138,32 @@ class AccessTokenAdmin(admin.ModelAdmin):
...
+
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        return qs.filter(user=request.user)
+
+    def has_change_permission(self, request, obj=None):
+        if not obj:
+            # the changelist itself
+            return True
+        return obj.user == request.user
```
dhruvkb, line 163: This is not needed, as any objects that are not associated with the current user have already been removed from the queryset and automatically do not have the change permission. Additionally we need to remove permissions to add and delete instances.

```suggestion
    def has_add_permission(self, request):
        return False

    def has_delete_permission(self, request, obj=None):
        return False
```

dhruvkb (changes requested):
Recommending some changes based on #4034.

AetherUnbound:
After the recent changes, the `moderator` user is not able to see user preferences. I'm going to try and figure out why.

api/api/admin/__init__.py
AetherUnbound: It looks like this is required @dhruvkb, see 1f64f2888be569ecb4650ab18269a699677c4e53.

dhruvkb (approved):
LGTM, thanks for addressing my feedback and correcting my mistake as well!

PR #3921, opened 2024-03-14
Add elasticsearch concurrency tags for Airflow

### LINKED ISSUE #3891: Improve support for "concurrency pools" in Elasticsearch DAGs

stacimc:
## Fixes

Fixes #3891 by @stacimc 

## Description

We now have quite a few DAGs that all operate on Elasticsearch or the underlying databases, and therefore must not run concurrently. We've been using the `prevent_concurrency_with_dags` and `wait_for_external_dags` utils to accomplish this in each DAG, by passing in a hard-coded list of the external DAGs we want to either wait on or otherwise prevent concurrency with. Since this list is hard-coded for each DAG, it is extremely easy to miss dependencies especially as DAGs are added and changed. A few were missed in the implementation of the Search Relevancy project so far.

This PR introduces ~two~ **three** new Tags that are used to identify DAGs as being part of a "concurrency group": `production_elasticsearch_concurrency`,  and `staging_elasticsearch_concurrency` for DAGs which affect the elasticsearch cluster, and `staging_api_db_concurrency` for DAGs which affect the API DB. Each DAG in these groups have been given the appropriate tag (the groups are listed below). The concurrency tasks have been updated to accept the name of the tag you want to select on, rather than a hard coded list, ie:

```
prevent_concurrency = prevent_concurrency_with_dags_with_tag(
    tag=STAGING_ES_CONCURRENCY_TAG,
)
```

^ This will select all the DAGs that have the given `tag`, *except for the running DAG/the DAG that called the task*, and fail immediately if any of them are running.

Screenshot 2024-03-13 at 10 35 10 AM

So now if you add a new ES dag, rather than having to manually update a bunch of lists of DAG dependencies on every single other ES dag, you just have to remember to add the appropriate `tag`(s) to your new DAG. The concurrency util will raise an error if you forget, as well.

### Additional thoughts/alternatives 

[This comment](https://github.com/WordPress/openverse/issues/3891#issuecomment-1988961899) on the issue explains why it wasn't possible to use the existing feature of Airflow pools to do this. Tags also have the benefit of being very visible in the UI and easy to filter by.

## Testing Instructions

These are the intended concurrency groups:

  * `staging_api_db_concurrency`
    * recreate_full_staging_index
    * staging_database_restore 
  * `staging_elasticsearch_concurrency`
    * recreate_full_staging_index
    * point_staging_es_alias
    * create_new_staging_es_index
    * create_proportional_by_source_staging_index
  * `production_elasticsearch_concurrency`
    * both data refreshes
    * both filtered index creation dags
    * create_new_production_es_index
    * point_production_es_index 

Note that `staging_database_restore` is **not** part of the `staging_elasticsearch_concurrency` group, because it only affects the API _database_ and therefore DAGs that only operate on the staging ES cluster (like `create_new_staging_es_index` don't need to block on it. Also note that `recreate_full_staging_index` is part of **both** the `staging_elasticsearch_concurrency` and `staging_api_db_concurrency` group, as it pulls records from the api db for reindexing.

For each DAG, run the DAG and check that the appropriate DAGs were waited on/prevented_concurrency_with. You can do this by looking at the logs for the `get_dags_in_<tag>_group` task, and then checking that the correct number of mapped tasks were made. (Example: `create_proportional_by_source_staging_index` is in the staging group which has 4 dags total. It has a `prevent_concurrency_with_dag` task generated for each of the 3 DAGs in the group excluding itself). Check the code to see special cases where some dependencies have been excluded for the data refresh.

Airflow 2.9 is supposed to include the ability to prove meaningful names for dynamically mapped tasks (instead of just a numerical index), which will make this much easier to read at-a-glance once we upgrade!

Screenshot 2024-03-13 at 10 37 03 AM

### Test the actual concurrency utils still works

We should also test the actual concurrency for at least a few options. The easiest way to do this is to trigger them in sequence through the Airflow CLI. Make sure the relevant DAGs are turned on in Airflow, then run `just catalog/shell` and:

```sh
# For example, this would trigger the recreate full staging index DAG and then immediately trigger staging db restore
# We would expect staging_database_restore to poke a few times while awaiting the `recreate_full_staging_index`
# DAG to complete, then resume once that DAG completes
> airflow dags trigger recreate_full_staging_index && airflow dags trigger staging_database_restore
```

A general tip for remembering how the concurrency checks should work: scheduled DAGs (data refreshes, staging database restore) will wait if a conflicting DAG is running, and resume once it completes. DAGs that are only run manually (create_new_es_index, create_filtered_index, recreate_full_staging_index) fail immediately if a conflicting DAG is running.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/common/sensors/utils.py
+++ b/catalog/dags/common/sensors/utils.py
@@ -2,14 +2,15 @@ from datetime import datetime
 
 from airflow.decorators import task, task_group
 from airflow.exceptions import AirflowSensorTimeout
-from airflow.models import DagRun
+from airflow.models import DagModel, DagRun, DagTag
 from airflow.sensors.external_task import ExternalTaskSensor
+from airflow.utils.session import provide_session
 from airflow.utils.state import State
 
 from common.constants import REFRESH_POKE_INTERVAL
 
 
-def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
+def _get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
     """
     Retrieve the most recent DAG run's execution date.
 
@@ -35,6 +36,18 @@ def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
     return []
 
 
+def _get_dags_with_tag(tag: str, excluded_dag_ids: list[str], session=None):
+    """Get a list of DAG ids with the given tag, optionally excluding certain ids."""
+    if not excluded_dag_ids:
+        excluded_dag_ids = []
+
+    dags = session.query(DagModel).filter(DagModel.tags.any(DagTag.name == tag)).all()
+
+    # Return just the ids, excluding excluded_dag_ids
+    ids = [dag.dag_id for dag in dags if dag.dag_id not in excluded_dag_ids]
+    return ids
+
+
 def wait_for_external_dag(external_dag_id: str, task_id: str | None = None):
     """
     Return a Sensor task which will wait if the given external DAG is
@@ -58,7 +71,7 @@ def wait_for_external_dag(external_dag_id: str, task_id: str | None = None):
         # Wait for the whole DAG, not just a part of it
         external_task_id=None,
         check_existence=False,
-        execution_date_fn=lambda _: get_most_recent_dag_run(external_dag_id),
+        execution_date_fn=lambda _: _get_most_recent_dag_run(external_dag_id),
         mode="reschedule",
         # Any "finished" state is sufficient for us to continue
         allowed_states=[State.SUCCESS, State.FAILED],
@@ -66,11 +79,18 @@ def wait_for_external_dag(external_dag_id: str, task_id: str | None = None):
 
 
 @task_group(group_id="wait_for_external_dags")
-def wait_for_external_dags(external_dag_ids: list[str]):
+@provide_session
+def wait_for_external_dags_with_tag(
+    tag: str, excluded_dag_ids: list[str], session=None, **context
+):
     """
-    Wait for all DAGs with the given external DAG ids to no longer be
-    in a running state before continuing.
+    Wait until all DAGs with the given `tag`, excluding those identified by the
+    `excluded_dag_ids`, are no longer in the running state before continuing.
     """
+    external_dag_ids = _get_dags_with_tag(
+        tag=tag, excluded_dag_ids=excluded_dag_ids, session=session
+    )
+
     for dag_id in external_dag_ids:
         wait_for_external_dag(dag_id)
 
@@ -81,7 +101,6 @@ def prevent_concurrency_with_dag(external_dag_id: str, **context):
     Prevent concurrency with the given external DAG, by failing
     immediately if that DAG is running.
     """
-
     wait_for_dag = wait_for_external_dag(
         external_dag_id=external_dag_id,
         task_id=f"check_for_running_{external_dag_id}",
@@ -91,6 +110,27 @@ def prevent_concurrency_with_dag(external_dag_id: str, **context):
         wait_for_dag.execute(context)
     except AirflowSensorTimeout:
         raise ValueError(f"Concurrency check with {external_dag_id} failed.")
+
+
+@task_group(group_id="prevent_concurrency_with_dags")
+@provide_session
+def prevent_concurrency_with_dags_with_tag(
+    tag: str, excluded_dag_ids: list[str], session=None, **context
+):
+    """
+    Prevent concurrency with any DAGs that have the given `tag`, excluding
+    those identified by the `excluded_dag_ids`. Concurrency is prevented by
+    failing the task immediately if any of the tagged DAGs are in the running
+    state.
+    """
+    external_dag_ids = _get_dags_with_tag(
+        tag=tag, excluded_dag_ids=excluded_dag_ids, session=session
+    )
+
+    for external_dag_id in external_dag_ids:
+        prevent_concurrency_with_dag.override(
+            task_id=f"prevent_concurrency_with_{external_dag_id}"
+        )(external_dag_id)
 
 
 @task(retries=0)
@@ -109,12 +149,3 @@ def is_concurrent_with_any(external_dag_ids: list[str], **context):
 
     # Explicit return None to clarify expectations
     return None
-
-
-@task_group(group_id="prevent_concurrency")
-def prevent_concurrency_with_dags(external_dag_ids: list[str]):
-    """Fail immediately if any of the given external dags are in progress."""
-    for dag_id in external_dag_ids:
-        prevent_concurrency_with_dag.override(
-            task_id=f"prevent_concurrency_with_{dag_id}"
-        )(dag_id)
--- a/catalog/dags/common/sensors/utils.py
+++ b/catalog/dags/common/sensors/utils.py
@@ -10,6 +10,9 @@ from airflow.utils.state import State
 from common.constants import REFRESH_POKE_INTERVAL
 
 
+THREE_DAYS = 60 * 60 * 24 * 3
+
+
 def _get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
     """
     Retrieve the most recent DAG run's execution date.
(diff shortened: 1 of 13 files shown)


stacimc:
Pinging @AetherUnbound as well since she asked if something like this was possible on a previous PR

catalog/dags/common/sensors/utils.py
```diff
@@ -35,9 +39,40 @@ def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
     return []
 
 
-def wait_for_external_dag(external_dag_id: str, task_id: str | None = None):
+@task
+def get_dags_with_concurrency_tag(
+    tag: str, excluded_dag_ids: list[str], session=None, **context
```
AetherUnbound, line 44: Is `**context` necessary here? Can we set `dag=None` directly and use the magic that Airflow provides for injecting that directly?

catalog/dags/common/sensors/utils.py
```diff
@@ -35,9 +39,40 @@ def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
...
+    dag_ids = [dag.dag_id for dag in dags]
+
+    running_dag_id = context["dag"].dag_id
+    if running_dag_id not in dag_ids:
+        raise ValueError(
+            f"The `{running_dag_id}` DAG tried preventing concurrency with the `{tag}`,"
+            " tag, but does not have the tag itself. To ensure that other DAGs with this"
+            f" tag will also avoid running concurrently with `{running_dag_id}`, it must"
+            f"have the `{tag}` tag applied."
+        )
```
AetherUnbound, line 60: Good call to check for!

catalog/dags/common/sensors/utils.py
```diff
@@ -35,9 +39,40 @@ def get_most_recent_dag_run(dag_id) -> list[datetime] | datetime:
...
+
+    # Return just the ids of DAGs to prevent concurrency with. This excludes the running dag id,
+    # and any supplied `excluded_dag_ids`
+    return [id for id in dag_ids if id not in [*excluded_dag_ids, running_dag_id]]
```
AetherUnbound, line 64: Nit: set could make this faster:
```suggestion
    return [id for id in dag_ids if id not in {*excluded_dag_ids, running_dag_id}]
```

AetherUnbound (approved):
This is fantastic! What a cool way to use tags 😄 🚀 - I tested everything locally and it worked as expected, even failed an ES index creation run when the data refresh was running. Only a few small things, otherwise this is good to go!

catalog/dags/common/sensors/utils.py
stacimc: You're right! Done 🚀

catalog/dags/database/staging_database_restore/staging_database_restore_dag.py
```diff
@@ -76,7 +77,8 @@
 def restore_staging_database():
     # If the `recreate_full_staging_index` DAG was manually triggered prior
     # to the database restoration starting, we should wait for it to
-    # finish.
+    # finish. It is not necessary to wait on any of the other ES DAGs as
+    # they do not directly affect the database.
     wait_for_recreate_full_staging_index = wait_for_external_dag(
         external_dag_id=RECREATE_STAGING_INDEX_DAG_ID,
     )
```
krysal, line 84: So because this is not using the `prevent_concurrency_with_dags_with_tag` utility, the DAG for restoring staging DB will not wait for other DAGs except for the one that is explicitly specified here, right?

krysal (commented):
This is a creative solution to potential concurrency conflicts between DAGs! Curiously, it also resembles the concept of pools.

I have some questions. First, what is the expected result of the test instructions showed? In my case, the `staging_database_restore` passes the concurrency test, while the `create_new_staging_es_index` fails it. The logic you mentioned suggests it should be the other way around, but it may be because the `staging_database_restore` DAG starts with the DAG being turned on.

Second, from what I understand, with the current code, all DAGs from the `staging_es_concurrency` tag wait for `staging_database_restore`, but this one only waits for `recreate_full_staging_index`, which is the only one with whom it really has no conflict. This sounds fine to me if we don't want to complicate things too much with many exceptions. An alternative I can think of is to create another **`staging_db_concurrency` tag** for DAGs working at the DB level. Both DAGs can be in that bag while `recreate_full_staging_index` retains the `staging_es_concurrency` tag. This keeps dependencies straight. What do you think of this approach?

stacimc:
> An alternative I can think of is to create another staging_db_concurrency tag for DAGs working at the DB level

It's funny you bring this up -- I just mentioned to @AetherUnbound earlier today that I was thinking of making an issue to do _exactly_ that in the future 😄 

The thinking was that for now, it is not actually a problem for any of these DAGs to be slightly over cautious in what DAGs they block on, and we're not introducing dependencies that weren't already hard-coded. Since you also brought it up, though, I'll just go ahead and implement it in this PR; it is the better long-term solution.

catalog/dags/elasticsearch_cluster/create_new_es_index/create_new_es_index_dag.py
```diff
@@ -137,7 +144,7 @@ def create_new_es_index_dag(config: CreateNewIndex):
         max_active_runs=1,
         catchup=False,
         doc_md=__doc__,
-        tags=["elasticsearch"],
+        tags=["elasticsearch", config.concurrency_tag],
```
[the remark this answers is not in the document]
stacimc: For sure! I went with `dag_config` as the new name instead, which I think is more clearly distinct from the `index_config` that's one of the DAG params :)

catalog/dags/elasticsearch_cluster/create_new_es_index/create_new_es_index_types.py
```diff
@@ -22,8 +16,8 @@ class CreateNewIndex:
 
     environment:         str representation of the environment in which to
                          create the new index
-    blocking_dags:       list of dags with which to prevent concurrency; the
-                         generated create_new_es_index dag will fail
+    concurrency_tag:     tag used to identify dags with which to prevent
```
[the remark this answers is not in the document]
stacimc: Good call, I think that makes it much clearer. Done!

catalog/dags/database/staging_database_restore/staging_database_restore_dag.py
stacimc: Accurate, but this has now been modified to instead use `wait_for_external_dags_with_tag` and the new staging db concurrency group 👍

stacimc:
All requested changes have been made, the biggest of which is simply splitting the staging concurrency groups out into separate DB and ES tags. The PR description has been updated, including the testing instructions -- as @krysal pointed out I had forgotten to update the testing instructions after changing the behavior of the `staging_database_restore` task, but it should now be accurate!

zackkrida:
@krysal will you have an opportunity to review this this week? If not, please go ahead and assign another reviewer. Thank you!

krysal (approved):
Thanks for the update. This looks great! 💯

PR #3927, opened 2024-03-14
Filter out duplicates from `raw_tags` in the catalog

### LINKED ISSUE #3926: Update `raw_tags` to avoid duplicates in the catalog
## Problem

We are not checking for possible duplicates of tags in provider script at the moment of ingestion. It is assumed that they do not exist from providers but it is better to be sure on our side.

## Description

Change the `raw_tags` field type to be a set instead of a list of strings.

## Additional context

Arises from #1566.

krysal:
## Fixes
Fixes  #3926 by @krysal 

## Description

This PR mainly aims to convert the `raw_tags` field to a set to ensure we're letting any duplicates pass. The number of files affected might be daunting but the commit history can help track changes.

This PR also:
- renames variables for denied tags
- adds terms to denied tags lists
- homogenizes the signatures of the `add_item` method of the `ImageStore` `AudioStore` classes

Something to consider while validating tags is sorting them right before enhancing them. This could allegedly be beneficial for "consistency between runs, saving on inserts into the DB later." The flicker DAG was doing it in addition to filtering out the duplicates.

---

I also identified the following provider DAG aren't collecting any tags for media:
- Auckland Museum
- Brooklyn Museum
- Cleveland Museum
- Europeana
- iNaturalist
- Museum Victoria
- Phylopic
- Science Museum
- SMK
- **Wikimedia**

## Testing Instructions

Run any number of DAGs you desire to test and confirm the tags are saved without duplicates. Run at least an audio and an image provider. Using the `AIRFLOW_VAR_INGESTION_LIMIT` variable to a low number, e.g. 100, is suggested for quick checks.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/common/storage/media.py
```diff
@@ -274,26 +281,28 @@ def _enrich_meta_data(meta_data, license_url, raw_license_url) -> dict:
...
-        else:
-            return [
-                self._format_raw_tag(tag)
-                for tag in raw_tags
-                if not self._tag_blacklisted(tag)
-            ]
+        if not raw_tags:
+            return
+
+        if not isinstance(raw_tags, set):
```
AetherUnbound, line 298: Instead of enforcing a `set` here (and having to change the logic in every provider), would it be easiest to just cast a `list` to `set` to remove the duplicates at this step while also accepting `set` as a valid type being passed into this function?

krysal: I was thinking of that option, yes, that would have been easier but at the same time, we still have a reasonably number of providers to allow a change like this so I went a step further and applied it to all the providers collecting tags. I think reducing fields to one type makes things easier in the long run, helping remember what kind of data we want here.

obulat (commented):
When the code to clean up was originally written, it was used to clean up the old TSVs  as well as the code from provider scripts. This is why there are checks to see if the tags are a dict or a list of strings. Now, we only use this with the `MediaStore`, so I _think_ we assume that the tags come as a list of strings.

With this assumption, we can simplify the tag enrichment and formatting:
1. convert the tags to a set, as is done here
2. use a set for denylist, and use the `difference`[^1] to remove the denylisted tags
3. enrich the tags, i.e, convert the strings to dictionaries of the shape `{ "name": tag, "provider": provider_name }`

What do you think, @krysal ? 

[^1]: https://docs.python.org/3.9/library/stdtypes.html#frozenset.difference

catalog/dags/common/storage/media.py
obulat: To minimize changes, I would rather keep the tags as a list of strings in the provider scripts, and convert them to a set here, in the `MediaStore` class.

I think that most of the duplicates were not supplied by the providers, but instead created by us when we incorrectly decoded the tags and saved both versions, or lower-cased the tags, and then saved both versions.

catalog/dags/common/storage/media.py
```diff
@@ -274,26 +281,28 @@ def _enrich_meta_data(meta_data, license_url, raw_license_url) -> dict:
...
+            raise TypeError(f"Expected set, got {type(raw_tags)}: {raw_tags}.")
+
+        return [
+            self._format_raw_tag(tag)
+            for tag in raw_tags
+            if not self._tag_denylisted(tag)
+        ]
 
     def _format_raw_tag(self, tag):
```
obulat, line 307:
```suggestion
    def _format_raw_tag(self, tag: str) -> dict:
        return {"name": tag, "provider": self.provider}
```

catalog/dags/common/storage/media.py
```diff
@@ -242,19 +249,19 @@ def _flush_buffer(self) -> int:
...
+    def _tag_denylisted(tag: str | dict) -> bool:
         """
-        Determine if the is banned or contains a banned substring.
+        Determine if the tag is banned or contains a banned substring.
 
         :param tag: the tag to be verified against the blacklist
         :return: true if tag is blacklisted, else returns false
         """
         if isinstance(tag, dict):  # check if the tag is already enriched
             tag = tag.get("name")
```
obulat, line 260:
```suggestion
```

catalog/dags/common/storage/media.py
obulat, line 307:
```suggestion
    def _format_raw_tag(self, tag):
        return {"name": tag, "provider": self.provider}
```

obulat: We can safely remove the `isinstance` checks since the provider scripts always provide a list of strings.

catalog/dags/common/storage/media.py
krysal: > we could very easily enforce that constraint in the MediaStore class

That is what the PR is doing.

I don't see the complexity in the change to a set; we have made more prominent changes in the past. As I said before, defining the type of fields clearly helps to decide what data goes in and it's easier to manage one data type instead of two. Also, sets aren't really more complex than a list.

catalog/dags/common/storage/media.py
krysal: I left this untouched precisely to minimize changes 😄 As this part of the code does not affect the rest of the operations, I thought it might be better to avoid it in this instance, but I can remove it if we agree on this path.

obulat: I thought an easier path to minimizing would be not touching the provider scripts, and handling all of the conversion here, in the `MediaStore` :)

obulat:
@krysal, I converted this PR to draft to prevent more pings.

AetherUnbound (changes requested):
I think that changing the type of `raw_tags` to a set and all the downstream changes necessary in the provider scripts to support that is not the appropriate approach here. It brings on a greater maintenance burden for each of the individual scripts when it's something we can centralize the logic for. Additionally, I worry that the error which is raised if a contributor adds a provider DAG which returns a list instead of a set may be more opaque and may make contribution harder as a result. I changed the NYPL DAG to return a list of tags instead of a set as an example, and this was the error it raised when testing locally:

```
[2024-03-20, 22:33:28 UTC] {taskinstance.py:2728} ERROR - Task failed with exception
providers.provider_api_scripts.provider_data_ingester.IngestionError: Expected set, got <class 'list'>: ['Branch libraries', 'Libraries'].
query_params: {"q": "CC_0", "field": "use_rtxt_s", "page": 1, "per_page": 250}
The above exception was the direct cause of the following exception:
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 439, in _execute_task
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 200, in execute
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 217, in execute_callable
  File "/opt/airflow/catalog/dags/providers/factory_utils.py", line 55, in pull_media_wrapper
    data = ingester.ingest_records()
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 276, in ingest_records
    raise error from ingestion_error
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 241, in ingest_records
    self.record_count += self.process_batch(batch)
  File "/opt/airflow/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py", line 473, in process_batch
    store.add_item(**record)
  File "/opt/airflow/catalog/dags/common/storage/image.py", line 146, in add_item
    image = self._get_image(**image_data)
  File "/opt/airflow/catalog/dags/common/storage/image.py", line 153, in _get_image
    image_metadata = self.clean_media_metadata(**kwargs)
  File "/opt/airflow/catalog/dags/common/storage/media.py", line 157, in clean_media_metadata
    media_data["tags"] = self._enrich_tags(media_data.pop("raw_tags", None))
  File "/opt/airflow/catalog/dags/common/storage/media.py", line 299, in _enrich_tags
    raise TypeError(f"Expected set, got {type(raw_tags)}: {raw_tags}.")
TypeError: Expected set, got <class 'list'>: ['Branch libraries', 'Libraries'].
```

Given that we _can_ just perform the de-duplication on the `MediaStore` class and prevent this from occurring entirely, I'd rather go with that approach.

krysal:
I refute that "it brings on a greater maintenance burden for each of the individual scripts," instead, it is changed to use the correct data type. Keeping two possible different types in fact _creates a greater maintenance burden_ down the line.  

> Additionally, I worry that the error which is raised if a contributor adds a provider DAG which returns a list instead of a set may be more opaque and may make contribution harder as a result. I changed the NYPL DAG to return a list of tags instead of a set as an example, and this was the error it raised when testing locally: ...

This is precisely what I'm proposing we should avoid, not using a `list` for `raw_tags`. Why would you want to continue to do that? @AetherUnbound you're one of the persons usually proposing the use of set over list whenever it's possible also for performance reasons. On the other hand, that error is pretty clear to me. How can it be more explicit? It mentions the problematic field and what is is expected to be. It seems to me that we are underestimating the contributors in this regard. Sets are not a complex data structure or something that comes from some obscure and esoteric third-party library.

If you see something else that _currently_ can be problematic with the use of `set` here I'd like to know and discuss it, for now I'd rather apply the changes already made here.

zackkrida:
@krysal and @AetherUnbound, I think this might be a good point in the conversation to solicit advice from @obulat and @stacimc on this particular issue concerning lists and sets. Since there is a clear difference of opinion here it might be useful to see where the majority opinion is held and go from there.

stacimc (commented):
I will say generally that I do not feel strongly about either option; I think the cons of each side are relatively minor.

My take is that if we allow two possible types in provider scripts we need only convert to a set in the MediaStore, which can be implemented very simply and I would _not_ anticipate to be any sort of maintenance burden down the line. If this work hadn’t already been done I would strongly advocate for that approach — but then again, the work to update provider scripts has been done here.

I do agree that the error raised by incorrectly using a list in a new provider script is fairly clear, but it is an error I expect people would encounter frequently and I don’t actually see a significant benefit to enforcing the type there. As far as performance goes I don’t see a reason why sets should be required in the providers, just before filtering out the denylisted terms — which would happen anyway if we did it in the MediaStore, and is consistent with other validation.

So TLDR my preference is to converting in the MediaStore, but I do not feel especially strongly because I don't think the maintenance burden is particularly bad either way (now that the initial work has already been completed).

krysal:
Thanks for sharing your thoughts, @stacimc.

> I do agree that the error raised by incorrectly using a list in a new provider script is fairly clear, it is an error I expect people would encounter frequently and I don’t actually see a significant benefit to enforcing the type there. As far as performance goes I don’t see a reason why sets should be required in the providers...

I was not exactly referring to performance as the _main_ benefit of using set, it is a slight advantage that we get for free this way; it's the constraint that we want **_no duplicates_** in tags. Having it enforced at the provider script level helps the DAG author think about what to put in there based on what the provider API returns. That is the main reason I'm trying to convey. It helps align the provider scripts with the data constraints.

stacimc:
> it's the constraint that we want no duplicates in tags.

We could require the providers to do _all_ the cleanup that we do in the MediaStore by the same logic, and then just raise errors in the MediaStore if it was done incorrectly; which isn't a bad thing, I just don't see a particular benefit when it's easy and convenient to apply the constraints in a centralized place. I would feel differently if there was any ambiguity in how the constraint should be applied.

I don't think it's a huge hurdle to developing provider scripts to enforce this there, but it is _a_ hurdle without significant benefit IMO and also a departure from how we use the MediaStore. If we were going to do more of this down the road Olga's suggestion for using a dataclass sounds promising to me, but I think that would require a separate issue and discussion. This discussion has raised some interesting points!

catalog/dags/providers/provider_api_scripts/nypl.py
```diff
@@ -138,7 +138,7 @@ def get_record_data(self, data):
                 "filetype": filetype,
                 "category": category,
                 "meta_data": metadata,
-                "raw_tags": NyplDataIngester._get_tags(mods) or None,
+                "raw_tags": self._get_tags(mods) or None,
```
obulat, line 141:
```suggestion
                "raw_tags": self._get_tags(mods),
```
`_get_tags` already returns set or None.

catalog/dags/providers/provider_api_scripts/nypl.py
```diff
@@ -223,20 +223,21 @@ def _get_creators(creatorinfo):
...
-        tags = []
-        if topics:
-            for topic in topics:
-                if isinstance(topic, list):
-                    tags.extend([t.get("$") for t in topic])
-                else:
-                    tags.append(topic.get("$"))
-        return [tag for tag in tags if tag]
+        if not topics:
+            return
```
obulat, line 233:
```suggestion
            return None
```
We usually have explicit `None` returns

catalog/dags/providers/provider_api_scripts/rawpixel.py
```diff
@@ -250,18 +244,10 @@ def _get_creator(data: dict) -> str | None:
...
-                for keyword in keywords
-                if not any(
-                    exclude_tag in keyword
-                    for exclude_tag in RawpixelDataIngester.tags_exclude_list
-                )
-            ]
-        return []
+    def _get_tags(metadata: dict) -> set | None:
+        if keywords := metadata.get("popular_keywords"):
+            return {k for k in keywords}
```
obulat, line 249:
```suggestion
            return set(keywords)
```
Would this be simpler?

catalog/dags/providers/provider_api_scripts/stocksnap.py
```diff
@@ -166,8 +165,8 @@ def _get_metadata(item):
         return metadata
 
     @staticmethod
-    def _get_tags(item):
-        return item.get("keywords")
+    def _get_tags(item: dict) -> set:
+        return {keyword for keyword in item.get("keywords")}
```
obulat, line 169: Is there a special reason for `keyword for keyword` here, can't we use `set(item.get("keywords", []))`? It's the same in a provider script above.

catalog/tests/dags/common/storage/test_media.py
```diff
@@ -513,11 +513,7 @@ def test_MediaStore_get_image_enriches_singleton_tags():
     image_store = image.ImageStore("test_provider")
 
     actual_image = image_store._get_image(
-        license_info=get_license_info(
-            license_="by-sa",
-            license_version="4.0",
-            license_url="https://license/url",
-        ),
+        license_info=BY_LICENSE_INFO,
```
obulat, line 516: Nice :)

obulat (changes requested):
The code changes look good.

My main request for change is the update to `_format_raw_tag` function

Since we are updating the shape of tags in the provider scripts, we should also update the `_format_raw_tag` to stop handling tags that are not a set of strings (I added a comment on `isinstance` checks inline.

catalog/dags/providers/provider_api_scripts/rawpixel.py
krysal: Indeed!

catalog/dags/providers/provider_api_scripts/stocksnap.py
krysal: Right, changed!

catalog/dags/common/storage/media.py
```diff
@@ -11,13 +11,18 @@
 
 logger = logging.getLogger(__name__)
 
-# Filter out tags that exactly match these terms. All terms should be
-# lowercase.
-TAG_BLACKLIST = {"no person", "squareformat"}
+# Filter out tags that exactly match these terms. All terms should be lowercase.
+TAG_DENYLIST = {
+    "flickriosapp:filter=flamingo",
```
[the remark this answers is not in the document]
krysal: Right, I missed these, updated!

AetherUnbound (changes requested):
> Understandably, we have different ways of approaching the problem. It is part of what it means to be a diverse team with different perspectives. It's unfortunate, though, that we disagree on the relationship between benefits and inconveniences, but we can still move forward.
> 
> Can I get some approval for this PR if we see no risk involved? It's beyond me why this has been delayed just for personal preferences until now.

To be clear, my rationale for disagreeing with this approach is not one of personal preferences. When we initially inherited the project, each provider script implemented nearly every step that was necessary for normalizing the data and preparing it for insertion into the database. Since then, we've taken several steps to centralize those efforts rather than have all that logic exist on the providers themselves, namely with the [`MediaStore`](https://github.com/WordPress/openverse/blob/ca82f843d5e8097fb2310cdaaa1c8fdf3550fdab/catalog/dags/common/storage/media.py#L42) and [`ProviderDataIngester`](https://github.com/WordPress/openverse/blob/6636dcfbb57abca19ef32027975f78548e10411f/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L67) classes. These efforts sought to move more of the shared normalization and functionality from the individual provider scripts into a common location. Doing this was a sacrifice to the simplicity of each script at the cost of more abstraction, but it allowed us an easier way to modify the behavior of _all_ provider scripts without having to affect each individual one.

The approach set out in this PR would be a step backwards from the ways we've worked to centralize the data normalization in the past. In the same way that we're requiring types for the tags be enforced on each provider script, we could require that each provider script provide valid and fully qualified URLs as well. Instead, we've [aggregated that logic in a singe place within the `MediaStore` class](https://github.com/WordPress/openverse/blob/ca82f843d5e8097fb2310cdaaa1c8fdf3550fdab/catalog/dags/common/storage/media.py#L93-L164) so that it _does not_ need to be present in each provider. Regardless of the accuracy of the type at the provider level, deduplicating tags seems like it's exactly the sort of normalization logic that should also exist at the `MediaStore` level along with all the other normalization steps we're taking there. I do not think that enforcing types at the provider level is a strong enough reason to motivate enforcing an individual validation everywhere that could be done in one place for all providers.

Perhaps I have the wrong impression or there are other elements of this I'm not considering. Do you mind sharing why you feel this validation differs significantly enough from the other centralized validation we're doing that it warrants being on every provider script?

krysal:
@AetherUnbound I understand better where your reasoning comes from, thank you for explaining it. It seems that we differ on that this change is making the provider scripts more complex. I find it simpler to make sure we use the correct data type that does the work we want (deduplicate) in its instantiation and even has performance benefits than checking for lists and doing type conversions later in the media store. URL validations are, of course, more complex, which we cannot easily solve with a change of type, and it's why there is a dedicated module for it. But I truly don't see the harm in changing to `set` when we could also save the CPU of conversions, maybe make people see this data type in action. The tags, a list of non-duplicate strings (which is also unlikely to change), are the perfect use of it.

I'm laser-focused on the case we have here, but I get the general preference to concentrate all the validation in media stores. Before implementing something else in future opportunities, I will consult with the team about possible options.

stacimc (changes requested):
> > We could require the providers to do all the cleanup that we do in the MediaStore by the same logic...
>
> I'm not advocating for that change so I see no point in discussing it. Agree that it can be interesting as a separate issue if you all wish.

I wasn't suggesting you were advocating for this, @krysal, and for what it's worth neither am I (I would even be opposed to that idea). I was trying to illustrate that what we're doing in this PR goes against the existing convention that we have for all of the other cleanup in the MediaStore (as @AetherUnbound expanded upon in her [later comment](https://github.com/WordPress/openverse/pull/3927#pullrequestreview-1972044911)).

> Can I get some approval for this PR if we see no risk involved? It's beyond me why this has been delayed just for personal preferences until now.

This was delayed primarily because there was an explicit ask earlier in the thread to get feedback for a majority opinion. I don't think the views expressed were simple style preferences or anything of that nature.

Reiterating my take for example: I do not think we should add (even minor) complexity to the provider scripts and break with our established conventions of normalizing in the MediaStore without a clear and significant benefit. I do not think deduplicating tags in the provider scripts has any meaningful performance improvement over doing it in the MediaStore, and I don't see particular benefit to provider script authors being made to think about tag duplication specifically, when that is so trivial to normalize at a higher level. So while I do not think the approach in this PR _breaks_ anything, I disagree with the approach.

Given that the [group consensus](https://github.com/WordPress/openverse/pull/3927#issuecomment-2025516251) that was asked for is in favor of deduplicating in the MediaStore, can you please change the implementation to match the established patterns?

krysal:
I'm sorry we don't agree on how we view this, but commit to the group consensus. It will be easier to make the changes in a new PR, so I'm closing this one.

PR #3947, opened 2024-03-20
Handle ClientResponseError before raising upstream status check

### LINKED ISSUE #3783: Thumbnail request fails even though the image accelerator returns a valid image
## Description

There is a problem with a Europeana image in that the API fails to process the `/thumb` request even when Photon (the image accelerator 3rd party part) returns it correctly. Sentry shows a bad request as a hint of the problem.

```
ClientResponseError: Bad Request
  File "api/utils/image_proxy/__init__.py", line 177, in get
    upstream_response.raise_for_status()
```

## Reproduction

1. Go to https://api.openverse.engineering/v1/images/c17f7ec8-d8ae-4e83-9440-5d1c35fb5d0f/thumb/
2. Observe the response is a `424 Failed Dependency`
3. Follow the Photon/Image accelerator link showed: https://i0.wp.com/imeji-mediathek.de/imeji/file/hFfmQSuYGYX2mJzI/92/07/d6/7a-35a9-4e68-a2f4-e1557938a545/0/original/9cda59f585ae70e2dc99d3f993197dbd.jpg?w=600&quality=80
4. See it working, making the first result unexpected and wrong.

## Additional context

[Sentry issue](https://openverse.sentry.io/share/issue/878b1fc728ed4c0cb483420de69014eb/)

firatbezir:
## Fixes
Fixes #3783 by @krysal 

## Description
This PR handles the ClientResponseError exception before raising the upstream status check. This ensures proper handling of errors and maintains the exception chain integrity.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


firatbezir:
Hi there, i try to understand why **linting check** was not satisfied event tough i run the comand regarding linting as seen from ss but still linting is red. i have had this experience before, could you please explain shortly if there is something from my end to fix tihs?
image

api/api/utils/image_proxy/__init__.py
```diff
@@ -169,37 +169,37 @@ async def get(
...
-            content_type=content_type,
+            content, 
+            status = status_code,
+            content_type = content_type,
+        )
+    except ClientResponseError as exc:
+        status = exc.status
+        do_not_wait_for(
+            _tally_client_response_errors(tallies, month, domain, status)
+        )
```
obulat, line 190:
```suggestion
        do_not_wait_for(_tally_client_response_errors(tallies, month, domain, status))
```

api/api/utils/image_proxy/__init__.py
```diff
@@ -169,37 +169,37 @@ async def get(
...
-        upstream_response.raise_for_status()
         status_code = upstream_response.status
         content_type = upstream_response.headers.get("Content-Type")
         logger.debug(
             "Image proxy response status: %s, content-type: %s",
             status_code,
             content_type,
         )
-        content = await upstream_response.content.read()
+        content = await upstream_response.read()
```
obulat, line 179: I'm not familiar with this part of the code. Why was `.content` removed here?

obulat (changes requested):
I added the changes that linter makes as "suggestions" here, @firatbezir .

I honestly don't know how to test this, so I'm going to wait for @krysal to give her opinion :)

obulat:
> I try to understand why **linting check** was not satisfied event though i run the comand regarding linting as seen from ss but still linting is red. i have had this experience before, could you please explain shortly if there is something from my end to fix tihs?

I'm not sure, but I think the "M" next to the file name means that the changes were not committed. If that's the case, you will need to commit the changes before running the linter.

firatbezir (commented):
i changed this and pushed again

api/api/utils/image_proxy/__init__.py
```diff
@@ -178,28 +177,27 @@ async def get(
             content_type,
         )
         content = await upstream_response.content.read()
+        upstream_response.raise_for_status()
```
sarayourfriend, line 180: What's the reason to move this from its previous location? If we're going to raise for the status, ideally it would happen before we spend cycles reading the response, at least, if not before everything else too. If we need to move it down here, can it go before the `content.read`?

sarayourfriend (changes requested):
Thanks for the PR!

Requesting changes mostly to ask for the `just` git submodule to be removed. Looks like maybe just an accidental commit during development.

Also requesting clarification on the reasoning behind moving the call to `raise_for_status`, and if it can go before the `read` call.

api/api/utils/image_proxy/__init__.py
firatbezir: Hi @sarayourfriend,

Thanks for taking a look at the PR and for your thoughtful comment!

To answer your question, the reason for moving the raise_for_status() call was to make sure that any errors are caught and handled properly before raising the upstream status check. This way, i thought we cuold've maintained the integrity of the exception chain.

I think your suggestion to move the raise_for_status() call before the read() call is a great one! It makes a lot of sense to check the status of the response before reading its content, as it would allow us to catch any issues with the response earlier on and prevent unnecessary read ops.

I'm going to make that change now and update the PR. Thanks again for your feedback and for helping to improve the code!

firatbezir:
> Thanks for the PR!
> 
> Requesting changes mostly to ask for the `just` git submodule to be removed. Looks like maybe just an accidental commit during development.
> 
> Also requesting clarification on the reasoning behind moving the call to `raise_for_status`, and if it can go before the `read` call.

Considering the changes i commiited regarding `raise_for_status`, i am not sure this satisfies the expectation from this issue. Could you please elaborate your comment regarding `just` ?

api/api/utils/image_proxy/__init__.py
```diff
@@ -169,37 +169,35 @@ async def get(
...
         status_code = upstream_response.status
         content_type = upstream_response.headers.get("Content-Type")
         logger.debug(
             "Image proxy response status: %s, content-type: %s",
             status_code,
             content_type,
         )
-        content = await upstream_response.content.read()
+        upstream_response.raise_for_status()
+        content = await upstream_response.read()
```
sarayourfriend, line 180: This is interesting. There is a slight difference in `upstream_response.content.read()` and `upstream_response.read()`. The former (the original version) streams the content, whereas the new version (just calling `.read` on the response), reads the content entirely into memory.

I am pretty sure in this particular case there is no difference, though, because we weren't actually streaming the response back to the requester. We were streaming the whole response into memory and then writing it to the response later on. I think this is left over from when I tried to use `StreamingHttpResponse` and couldn't get it to work reliably. This should be an okay change to make! But I wanted to note it here _just in case_ there ends up being an issue during deployment that could be explained by this very minor and easy-to-miss difference in the code. I really think this is a safe change, though, and is positive because it uses the API that matches directly how we are actually using the response.

sarayourfriend (commented):
This LGTM aside from the `just` submodule. Dhruv's clarification is correct :+1: Once that is removed, I'll happily approve this. If you have any thoughts on how to add a test for this change, please do, but I don't have any myself, so not blocking on the lack of tests for that reason.

BTW: to remove the submodule, you may need to use `git rm just` from the repository root, especially if you need to keep that `just` folder around for yourself for some reason... though I'd recommend moving it elsewhere if you can, to more easily avoid this in the future :slightly_smiling_face:

firatbezir:
i drafted this PR and updated the `init.py` and removed `just` submodule considering the reviews @sarayourfriend and @obulat have suggested. However, **CI + CD / Run tests for the API** still fails and  i don't have any idea why. If you have any other suggestion to fix this, i really appreciate.

sarayourfriend:
I was just re-reviewing this PR, and, now that I am reading the original issue again, I am realising that you've unfortunately found a bad combination of a poorly described issue and us maintainers not doing a good job making sure _we_ understood the issue before even starting to review the PR. We often take for granted each other's understanding of Openverse's code history and systems integrations and sometimes that leads to us failing to consider the underlying cause of a particular issue. In this case, that's led to us repeatedly reviewing this PR without catching what we could have early on: that this approach does not address the actual problem, and that the linked GitHub issue does not capture the problem either. The latter is due almost entirely to when the issue was created, and some significant changes to this code that happened since then, which I share below.

The issue describes a situation where these ClientResponseError exceptions were sent to Sentry, but that is impossible now. We absolutely always handle them in the exception handling, and never send them to Sentry.

I was confused about this, but I went back and looked at the Sentry issue, and found the last commit where this was happening: https://github.com/WordPress/openverse/blob/9b4f727d87e026aa2ca7fb9afcda0c05666ef9e5/api/api/utils/image_proxy/__init__.py#L191-L228

As you can see, in that code, we were still sending these exceptions to Sentry. I remember now that in [this commit](https://github.com/WordPress/openverse/commit/b99d89deb6f5e83a59969ecbed531afff79577ba) @AetherUnbound actually removed the code to report the exceptions to Sentry. They were mostly unactionable.

This PR does not get at the underlying problem of the issue, which is that particular Europeana thumbnail returning a 400 from our API endpoint, even though the Site Accelerator (Photon) link works perfectly fine. The GitHub issue does not describe this as the issue, and we shouldn't have let anyone other than staff take on such a thinly described task.

The fact remains that the particular thumbnail the issue links does not work, and this PR won't fix that. I'll update the issue with a better description of what the fix should be, but it is likely to be a much more complex issue than a small change to error handling, requiring some investigation of our Photon (Site Accelerator, as it is now called) integration, and why this particular request 400s when we make it from our server but not when hit directly. It might even have something to do with an infrastructure issue, which means I will have to mark it as staff only.

I really apologise for this confusion and for wasting your time on this PR @firatbezir. You and I have discussed before about how Openverse project needs better standards for issue writing and scoping when we are open to community contributions for a particular issue, and this is a good example of why. The issue is not sufficiently described for someone without expert knowledge of our systems and an understanding of this history of our codebase in this particular area to adaquately address.

I'm going to close this PR. If you want help finding an issue to work on, please let me know.

firatbezir:
Hi @sarayourfriend, actually this was the one great experience for me. Helped me understand lots of points that are needed to take a good care of such as `just` submodule, api testing etc. I really appreciate for the patience you, @obulat, and @dhruvk showed to let me capture the basic fundamentals of processing the issue of **openverse repository**. Before the selecting a new one, I would be a wise idea to check on with you guys. ✌️

Issue #3592, opened 2023-12-27
Dark mode


fcoveram:
## Description

Include a feature that changes the UI from light to a dark color scheme.

### Background

Dark mode is a product-wide setting that adapts the user interface to provide a comfortable experience in low-light environments and, in some cases, improve accessibility for those with visual impairments, as [pointed out by NN/g](https://www.nngroup.com/articles/dark-mode/).

As described in [Material Design](https://m2.material.io/design/color/dark-theme.html#usage) documentation, dark themes also reduce the luminance emitted by device screens and conserve battery power, as devices with OLED screens benefit from the ability to turn off black pixels at any time of day.

## Documents

- [ ] Project Proposal
- [ ] Implementation Plan(s)
- [ ] Design Proposal

## Milestones/Issues

TBD

fcoveram:
## Update 2024-03-25

* Project proposal document is done, #3913 
* Project status changed to "In RFC"

fcoveram:
# Update on 2024-05-01

* Frontend implementation plan is still in progress but very close to finish.
* Dark mode designs #3959 are done and the ticket was closed.
* #4155 was created to decide how to switch the site theme.
* #4157 was created to decide the feature introduction.
* #4232 was created requesting the addition of a new component for #4155
* In parallel, @fcoveram has been updating the Design Library with the new variables to have every Figma component in light and dark mode.

zackkrida:
I am going to lead this project moving forward.

The [implementation plan](https://github.com/WordPress/openverse/issues/3958) for the project is approved and merged.

Next, I will create issues and milestones for the project. Once created I will outline the workflows and link to the milestones in the project issue body. At that point I will move the project to "On Hold" while we work on the Nuxt 3 migration. This will not be worked on until that project is completed, or is again put on hold due to limiting circumstances.

zackkrida:
@fcoveram do you have an update yet on the following aspects of the designs?

- The finalized dark mode mockups
- The semantic color names to replace our existing colors ("yellow" to "complementary", for example)
- The mapping of dark mode colors to their light mode counterparts

zackkrida:
## Update 05-10-2024

All issues and milestones have been created and added to the parent issue ✨. The project is ready to be worked on but will be moved to "On Hold".

There is some remaining work around color names that I raised [here](https://github.com/WordPress/openverse/issues/4268#issuecomment-2103597982) and am awaiting reply.

fcoveram:
[...]

PR #3963, opened 2024-03-25
Dark Mode Frontend Implementation Plan

### LINKED ISSUE #3958: Implementation Plan: Dark Mode Frontend
## Problem

The #3592 project needs a frontend implementation plan to spec out how the work will be completed. 

## Description

This should, abstractly, include:

- How new colors will be managed in Tailwind config
- How any dark-mode specific UI will be built
- If feature flagging will be necessary

zackkrida:
## Fixes

Fixes #3958 by @zackkrida 

## Description

This is an implementation plan for adding a dark mode to the Openverse frontend as part of the #3592 project.

I have chosen @obulat and @sarayourfriend as reviewers for their expertise in implementing frontend features, TypeScript generally, and familiarity with the testing-related code in the frontend.

## Testing Instructions

Please read the IP using the decision-making process described below.

## Decision-making process

### Current round

This discussion is currently in the [**Decision round**](https://docs.openverse.org/meta/decision_making/process_description.html#decision-round). The deadline for this round is Friday, May 3rd.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


fcoveram:
Do you have a deadline in mind for this document @zackkrida? 

If helpful, I found the following Figma documentation/post that could ease moving the design to frontend.

* [Guide to variables in Figma](https://help.figma.com/hc/en-us/articles/15339657135383-Guide-to-variables-in-Figma#h_01H4S62F4ERWJQ6B4DZ7NBCJWC)
* [How Figma uses Dev Mode to ship high quality features, faster](https://www.figma.com/best-practices/how-figma-uses-dev-mode/)

zackkrida:
@fcoveram I'm going to finish this today. I will look at these resources. Thanks!

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,224 @@
...
+-->
+
+The following plan requires approved designs and semantic color names.
+
+1. Parallel Work Stream "A": Implement the new color palette
+   1. Create a `FORCE_DARK_MODE` feature flag which is disabled by default and
+      available in every environment. Add logic to add a `dark-mode` class to
+      the root HTML tag when this flag is enabled and a `light-mode` class when
+      disabled.
+   2. Rename all colors in the tailwind config and the frontend components to
```
obulat, line 145: This step also doesn't seem to be dependent on anything else and could be done in parallel.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+crucial to understanding this implementation. Quite simply and elegantly, the
+designs use a "palette swap" approach in which each color has a 1:1 replacement
+from light mode to dark mode.
+
+![Openverse.org dark mode color palette](/_static/dark_mode_palette_example.png)
+
+While we _will_ include easy mechanisms for exceptions to this rule, they do not
+appear to be necessary based on the designs. Implementing our dark mode, then,
+should allow component authors to write components _once_, using semantic color
+names, that will automatically switch between their light and dark mode
```
sarayourfriend, line 47: yay!

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+We will switch our color names defined in the tailwind configuration to use
+semantic names, for example replacing "pink" with "primary" and "yellow" with
+"complementary". Instead of hardcoding these colors in the Tailwind
+configuration, the tailwind configuration will reference CSS variables defined
+in our root css file. The value of the CSS variables will be switched based on a
+dark mode CSS class added to the HTML root when dark mode is enabled.
+
+Tailwind's built in `dark:` modifier can be used for any styles which need to
+override the default behavior or add dark-mode specific styles beyond the core
+palette swap.
```
sarayourfriend, line 61: I was surprised that the semantic colour names wasn't Tailwind's built-in dark mode handling!

Here are the Tailwind docs for darkmode, might be worth linking for context and ease of reference: https://tailwindcss.com/docs/dark-mode

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+
+Here are pretend, simplified examples of this setup.
+
+In our primary CSS file, we setup two lists of CSS variables with different
+values when the `.dark-mode` class is present:
```
sarayourfriend, line 66: To make Tailwind respect this for the `dark:` prefix, we'll need to update the config for manual toggling, right? https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Expected Outcomes
+
+<!-- List any succinct expected products from this implementation plan. -->
+
+- Users writing components will have a "dark mode compatible by default"
+  experience. By default, they will not need to think much about dark mode.
+- Color names in the codebase will be replaced with semantic names.
+- Frontend developers will have easy tools to visually test components in light
+  and dark mode.
```
sarayourfriend, line 129: Should this include expectations for visitors of the site? These are only devex outcomes. What about Openverse.org users?

```suggestion
- Users writing components will have a "dark mode compatible by default"
  experience. By default, they will not need to think much about dark mode.
- Color names in the codebase will be replaced with semantic names.
- Frontend developers will have easy tools to visually test components in light
  and dark mode.
- Visitors to the site will have a darkmode option available to them to toggle on/off
- Visitors with `prefers-color-scheme` set to dark will have dark mode automatically enabled
```

The toggle option I'm not sure is possible yet, because Francisco hasn't finished the designs for that interaction, but `prefers-color-scheme` is a particularly important one to consider seeing as we're using a custom selector to indicate when dark mode should be used.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+```css
+:root {
+  --color-primary: black;
+  --color-secondary: white;
+}
+
+.dark-mode {
+  --color-primary: white;
+  --color-secondary: black;
+}
```
sarayourfriend, line 77: Could potentially need to be more complex here, if we want to avoid a light-mode flash for users manually selecting dark mode.

The selector would need to be something like... `.dark-mode OR (prefers-color-scheme: dark AND NOT .light-mode)`, where `.light-mode` is only ever set for users who have manually toggled to light mode from `prefers-color-scheme: dark`.

An inverted selector needed for the default-light mode state as well... `.light-mode OR (prefers-color-scheme: light AND NOT .dark-mode)`.

I think this makes the tailwind darkmode prefix logic a bit complicated too :thinking:

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+        preference: "dark" | "light" | "system" // Defaults to "light"
+        systemValue: "light" | "dark" // Readonly representation of the system value
+      }
+      ```
+
+      The color mode should be stored in a cookie so that a previously-selected
+      user choice can be used when rendering via SSR and prevent a visual flash
+      of light mode styles for users who have selected dark mode. The system
+      value is read by matching the `'(prefers-color-scheme: dark)'` media
+      query.
```
sarayourfriend, line 198: How will our frontend caching affect users with this cookie set? Do we need to consider a `Vary` header on the response for these users?

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+
+   3. Behind the feature flag, add the new user interface element which toggles
+      dark mode (exact design TBD, but it will be comprised of existing UI
+      components). Default to "light" mode but support choosing between "dark",
+      "light", and "system".
```
sarayourfriend, line 203: Shouldn't it default to `system` for user's with `prefers-color-scheme: dark` set? Or does that come later?

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+
+> See the ["Rollback"](#rollback) section for details on how to revert this
+> deployment.
+
+1. Set the `DARK_MODE_UI_TOGGLE` feature flag to "on" in all environments.
+2. Deploy the production frontend.
+3. Test and verify the deploy was successful and that dark mode:
+   1. Looks correct
+   2. The toggle works correctly (chosen settings persist, the control works
+      with keyboard, etc.)
```
sarayourfriend, line 217: Shouldn't we do this verification step in staging, not production?

sarayourfriend (commented):
Looking great so far! I think there are some potential complications with the selector that need to be ironed out, and some other minor clarifications. I don't anticipate any blockers during the decision phase.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,263 @@
...
+
+      3. Visual Regression tests. Update
+         `frontend/test/playwright/utils/breakpoints.ts` so that each breakpoint
+         produces and expects a dark mode screenshot to pass as well as the
+         existing light mode screenshot. **This will also be a significant diff,
+         as at the time of writing it will create 293 new screenshots to review.
+         This is also the point of the process where @fcoveram and
+         @wordpress/openverse-frontend should review the full dark mode
+         appearance for correctness and sufficient color contrast (see the
+         ["Accessibility"](#accessibility) section for more details.**
```
obulat, line 178: Small detail question. Should the old snapshots be renamed to `<snapshot_name>_light` so that the new ones are similarly `_dark`? Or should the old ones retain their names, and only new ones  have an additional suffix? If the former, it will introduce a significant diff in old snapshots in addition to the dark mode ones, so it would be better to do in a separate PR for ease of review.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
obulat: After the problems we had during launch of additional search views, I think we should verify in staging, and then set the production flag to `enabled` (without ever setting it to `swtichable`)

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,224 @@
...
+   2. Rename all colors in the tailwind config and the frontend components to
+      use new, semantic names.
+   3. Replace "hardcoded" color values in the tailwind configuration file with
+      css variables defined in the "base" layer of the `tailwind.css` file.
+   4. Add the dark mode colors in the form of additional css variable
+      definitions, nested under the `.dark-mode` css class, in the "base" layer
+      of the `tailwind.css` file. **At this point, the full dark mode appearance
+      should be able to be tested manually by @fcoveram and
+      @wordpress/openverse-frontend for any inconsistiencies or problems**.
+   5. Create a `color-mode.ts` test utility for playwright that works comparably
```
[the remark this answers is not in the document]
zackkrida: I think I would prefer that the initial screenshots are as correct as possible, rather than commit screenshots with errors only to be replaced in subsequent PRs. Can you think of any benefits not to initially commit the screenshots once the components look right in a PR?

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: I'll definitely add this :+1:

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: I totally forgot to document this, yes. I'll update, and make sure the documentation is clear that this is only relevant for the "escape hatch" of `dark:` and not the core color changing mechanism.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Perfect, I'll commit this as-is.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: My original thought was that the `prefers-color-scheme` media query would be read in JS and used to set this css class. 

For example, a first-time user visits the site, chooses "dark mode" from the UI. On subsequent SSR visits their "dark mode" selection is read from the cookie and the class is added on the server. 

If we do need more advanced tailwind prefix logic, though, they have decent functionality to make it less clunky:

https://tailwindcss.com/docs/dark-mode#using-multiple-selectors

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: I need to think about this a bit more but it might be wise for us to disable frontend caching on the actual site pages, OR store the color theme preference in a cookie _and_ in localstorage simultaneously :thinking:

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: This is a bit of an open question IMO. Originally @fcoveram wanted to default to light mode so that the site wouldn't suddenly change to dark mode for users, but the more I think about it defaulting to system seems preferable and like a very clean way to _introduce_ the feature to the very users most likely to use it. This is also being discussed a bit in #4157.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Sounds good!

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: I don't have a strong preference either way.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
sarayourfriend: Reading it in JS would cause a flash on SSR though for users who's system-level preference is not the default, right? Or is the server able to read that media query for the client? The benefit of using the media query is that it entirely divorces the implementation from JS, meaning the styles the user sees are immediately reflective of their environment and choice (both), without any delay. SSR only comes into play to set a choice that overrides the media query.

We have this issue all over the place with JS-dependent media query options (where we use `v-if="breakpoint == 'md'"` type logic rather than CSS), and it creates rather messy JavaScript and our unfortunate reliance on cookies to SSR the correct breakpoint _on subsequent page loads_ but never the first (except for when we get lucky and the UA indicates a relevant platform, which might not continue to be reliable into the future with UA hints).

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
sarayourfriend: A cookie is the only option workable for SSR, I think. SSR has no access to local storage, we really should only rely on local storage for state that matters only after users are actively interacting with the page (recent searches, for example) or when we absolutely need to avoid GDPR problems.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
sarayourfriend: Renaming existing snapshots should just have file-rename diffs, not diffs of the actual content (I hope). Adding the colour scheme to all snapshots explicitly sounds like a good idea to keep things easy to understand and maximally clear.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
sarayourfriend: For this and https://github.com/WordPress/openverse/pull/3963#discussion_r1575399744, it would be helpful to map out the flow and end states for the different scenarios, maybe as Mermaid charts to help with visualisation. I just want to make sure we've covered the cases going in. I believe they are:

- SSR
  - First time visitor
    - prefers-color-scheme: dark -> dark render
    - prefers-color-scheme: light (or unset) -> light render
    - (_These visitors never have preferences set, so I think they're effectively the same as repeat visitors with unset preferences, but that's only true if unset is distinct from system, i.e., if we avoid setting the cookie to anything until the user interacts, vs creating the cookie with the default value of system on the first request; **this in particular would be good to specify, when the cookie is set and whether a fallback value is ever necessary for completely unset cookies**_)
  - Repeat visitor
    - prefers-color-scheme: dark -> dark render
      - unset
      - system
    - prefers-color-scheme: light (or unset) -> light render
      - unset
      - system
    - prefers-color-scheme: dark; manually selected light preference -> light render
    - prefers-color-scheme: light; manually selected dark preference -> dark render
- CSR (i.e., SPA navigation/setting toggles)
  - prefers-color-scheme: dark -> dark render
    - unset
    - system
  - prefers-color-scheme: light -> light render
    - unset
    - system
  - prefers-color-scheme: dark; manually selected light preference -> light render
  - prefers-color-scheme: light; manually selected dark preference -> dark render

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
fcoveram: > …defaulting to system seems preferable and like a very clean way to introduce the feature to the very users most likely to use it…

I revisited this option with other designer contributors, and it seems the best one. No matter the design we choose in #4157, we can safely go with this option.

zackkrida:
@sarayourfriend and @obulat any objection to me drafting the PR and moving on to the following updates?

- Update the old snapshot names to include the "light" theme
- Move verification and testing from production to staging
- Update the implementation to default to the user's system preference as discussed [here](https://github.com/WordPress/openverse/pull/3963#discussion_r1575849900). Choosing "dark" will set a `.dark-mode` class, choosing "light" will set a `light-mode` class, and choosing "system" (the default value) will do nothing.
- Update the tailwind config so that `darkMode` is set like this:
   ```ts
   /** @type {import('tailwindcss').Config} */
   module.exports = {
    darkMode: ['variant', [
      '@media (prefers-color-scheme: dark) { &:not(.light-mode *) }',
      '&:is(.dark-mode *)', // :is is so the specificity matches and there's not unexpected bahavior 
    ]],
  // ...
}
   ```
   This makes sure that we default to the user preference but that the user preference is overwritten if they have chosen light or dark mode in the UI toggle.
- Add a chart(s) for the different color scenarios (requested [here](https://github.com/WordPress/openverse/pull/3963#discussion_r1575448554))

sarayourfriend:
Yes, sounds good to me.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Added in 6295bb0a496428a19e22329f4dd4e1c863a60f17.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Added in 6295bb0a496428a19e22329f4dd4e1c863a60f17.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Updated, and included a strategy for bypassing the cache when this new cookie is present.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: I ultimately felt that we didn't need a chart but have tried to supply enough new documentation that the cases should be clear. If you still think this needs further clarification in the document let me know and I'll happily amend.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: We are now defaulting to system.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
zackkrida: Added in 6295bb0a496428a19e22329f4dd4e1c863a60f17

zackkrida:
@obulat and @sarayourfriend this proposal has been updated, and is ready for final review in the Decision round.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,326 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @obulat
+- [ ] @sarayourfriend
```
sarayourfriend, line 13:
```suggestion
- [x] @sarayourfriend
```

sarayourfriend (approved):
LGTM!

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,326 @@
...
+**Author**: @zackkrida
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @obulat
```
obulat, line 12:
```suggestion
- [x] @obulat
```

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,326 @@
...
+      modes, with "system" as the default. The "dark" and "light" options will
+      set a `.dark-mode` or `.light-mode` class on the HTML element of the site.
+      The default "system" choice does not add a class to the HTML element. When
+      there is no HTML `.{color}-mode` class present, the site will default to
+      the `prefers-color-scheme` media query value.
+      1. Add a `TOGGLE_COLOR_SCHEME` analytics event with a playload including
+         the color mode preference chosen by the user.
+      2. Update our Cloudflare static page caching rule for the frontend in with
+         a Cookie bypass rule (in pseudocode, something like `and not
+         http.cookie contains "openverse_color_scheme"))
```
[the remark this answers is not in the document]
obulat: Would this make all of the users who select the dark mode bypass cache?

obulat (approved):
LGTM!
Thank you for mentioning the cache rules, too.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
```diff
@@ -0,0 +1,326 @@
...
+      2. Add the dark mode colors as CSS variable definitions, nested under the
+         `.dark-mode` CSS class, in the "base" layer of the `tailwind.css` file.
+         This will not result in any visual changes, _except when
+         `FORCE_DARK_MODE` is enabled._
+
+      3. Visual Regression tests. Update
+         `frontend/test/playwright/utils/breakpoints.ts` so that each breakpoint
+         produces and expects a dark mode screenshot to pass as well as the
+         existing light mode screenshot. **This will also be a significant diff,
+         as at the time of writing it will create 293 new screenshots to review.
```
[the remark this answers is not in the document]
zackkrida: Definitely. Frankly the existing size is probably already cause to consider moving them elsewhere and coming up with a strategy. I'll create a standalone issue to address it.

documentation/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.md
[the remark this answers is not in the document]
zackkrida: @obulat yes, it would, but crucially only on the pages which are being cached: the homepage and all static pages. Also, due to SPA routing these, static pages are only served from the cache on the user's **initial page load**. Therefore, cached homepage and static page visits mark an overall small percentage of Openverse traffic.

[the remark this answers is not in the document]
sarayourfriend: Just wanted to add more 'flavour' to the issue, but even if we were worried about the caching issue, we could solve it by caching and segmenting the cache at the nginx level (or replace it with varnish, something more suitable for complex caching rules). We'd already be doing that if we moved those routes "statically" out of the app or cached them _at all_, and we could segment on as many relevant cookies as needed (hopefully not too many!).

Just say, there are solutions to this if it ever became a problem.

PR #3979, opened 2024-03-27
Reset the media store state if collection state changes

### LINKED ISSUE #3976: Staging: Provider filter is preserved on collection tag views
## Description

Part of the #410 project. New collection pages are preserving search filters when they shouldn't. For example, searching under a specific provider, selecting a single result, and clicking on a tag leads that tag view to only include results from the provider instead of all results.

This may apply to other filters that I am not aware of!

## Reproduction

1. Make sure the `additional_search_views` feature is enabled: https://staging.openverse.org/preferences
1. Search "rabbit" from StockSnap only: https://staging.openverse.org/search/image?q=rabbit&source=stocksnap
2. Click on a single result, then a tag
4. Observe that the tag view is filtered to only show StockSnap results. You can refresh and see different results.

obulat:
## Fixes

Fixes #3976 by @zackkrida

## Description

This PR adds `clearMedia` call to the search store's function setting the collection state. This makes the `setCollectionState` similar to `setSearchStateFromUrl`. To make sure that we don't clear media when non-collection state params change (for instance, if we add filters to the collection views, or if we add `page` parameter to the frontend URL to improve pagination), we first check if the `collectionParams` are different.

## Testing Instructions

Go through the following steps and check that the linked issue is fixed (step 5):

1. Run the app locally using `just frontend/run dev`
2. Make sure the additional_search_views feature is enabled: http://localhost:8443/preferences
3. Search "rabbit" from StockSnap only: http://localhost:8443/search/image?q=rabbit&source=stocksnap
4. Click on a single result, then a tag
5. Observe that the tag view is _not_ filtered to only show StockSnap results. You can refresh and see the same results.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/data/api-service.ts
+++ b/frontend/src/data/api-service.ts
@@ -122,16 +122,7 @@ export const createApiService = ({
     return config
   })
   client.interceptors.response.use(
-    (response) => {
-      const params = new URLSearchParams(response.config.params).toString()
-      console.log(
-        "> API response",
-        new URL(
-          `${response.config.baseURL}${response.config.url}${params ? "?" + params : ""}`
-        ).href
-      )
-      return response
-    },
+    (response) => response,
     (error) => {
       if (error.code === "ECONNABORTED") {
         error.message = `timeout of ${
--- a/frontend/src/data/api-service.ts
+++ b/frontend/src/data/api-service.ts
@@ -122,7 +122,16 @@ export const createApiService = ({
     return config
   })
   client.interceptors.response.use(
-    (response) => response,
+    (response) => {
+      const params = new URLSearchParams(response.config.params).toString()
+      console.log(
+        "> API response",
+        new URL(
+          `${response.config.baseURL}${response.config.url}${params ? "?" + params : ""}`
+        ).href
+      )
+      return response
+    },
     (error) => {
       if (error.code === "ECONNABORTED") {
         error.message = `timeout of ${
--- a/frontend/src/stores/search.ts
+++ b/frontend/src/stores/search.ts
@@ -332,9 +332,39 @@ export const useSearchStore = defineStore("search", {
 
       this.addRecentSearch(formattedTerm)
     },
+    isCollectionStateSame(
+      collectionParams: CollectionParams,
+      mediaType: SupportedMediaType
+    ) {
+      if (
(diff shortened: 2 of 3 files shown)


frontend/src/data/api-service.ts
```diff
@@ -122,7 +122,16 @@ export const createApiService = ({
...
+    (response) => {
+      const params = new URLSearchParams(response.config.params).toString()
+      console.log(
+        "> API response",
+        new URL(
+          `${response.config.baseURL}${response.config.url}${params ? "?" + params : ""}`
+        ).href
+      )
+      return response
+    },
```
obulat, line 134:
```suggestion
    (response) => response,
```

zackkrida (approved):
I don't love having to manage `isCollectionStateSame` as a custom approach to checking for deep equality of two objects, but I am not sure we have a simpler alternative at the moment and this appears to be working. LGTM! We should really move to using lodash or similar in the future IMO.

frontend/src/stores/search.ts
```diff
@@ -333,9 +333,39 @@ export const useSearchStore = defineStore("search", {
 
       this.addRecentSearch(formattedTerm)
     },
+    isCollectionStateSame(
+      collectionParams: CollectionParams,
+      mediaType: SupportedMediaType
+    ) {
```
AetherUnbound, line 339: Is there an easy way to test this new function?

AetherUnbound (commented):
I wanted to try and review this PR, but it doesn't have any testing instructions so I'm not sure how best to proceed 🙂 I do have one question.

frontend/src/stores/search.ts
obulat: I removed this function in favor of @wordpress/is-shallow-equal that we are already using in another component.

obulat:
> I wanted to try and review this PR, but it doesn't have any testing instructions so I'm not sure how best to proceed 🙂 I do have one question.

Sorry, I updated the instructions :)

AetherUnbound (approved):
Perfect, thanks for the testing instructions! Glad we can use an existing function here too 😄

Issue #3642, opened 2024-01-07
Drop `status` column from content report tables


sarayourfriend:
## Description

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#backfill-moderationdecision-and-drop-report-status

Blocked by ~#3641~ #4530 and the **re-run of the `backfillmoderationdecision` command**.

dhruvkb:
This is not directly related to this project but I wanted to suggest inheriting from `OpenLedgerModel` in the report models (currently they do not, and instead use a custom field for `created_at`) which would introduce a migration so it's possible to run both migrations simultaneously to reduce the downtime on that table.

sarayourfriend:
Good call! Unifying the date columns is a great idea, especially given how small the tables are at the moment. Would it require following the zero downtime migration policy? If so, we should create the issues for each part of the process, maybe as follow-up tasks to the project rather than part of it.

AetherUnbound:
This is still blocked by https://github.com/WordPress/openverse/pull/4530, but can be tackled after that is run!

krysal:
The backfill was run between Tuesday and Wednesday so this is unblocked.

PR #3989, opened 2024-03-31
Create the moderation decision model

### LINKED ISSUE #3636: Create `ModerationDecision` table
## Description

The plan for this has been modified slightly from the [description in the IP](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#id3) based on the subsequent [Bulk actions IP](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.html), so the full description (including updates) is written out below. Where it differs from the IP I have bolded the text so it is easier to see.

- Create the new ModerationDecision model with the following columns:
  - id: basic auto-incrementing id
  - created_on: the date the moderation decision was created, in other words, the date the decision was made
    - Provided by OpenLedgerModel
   - moderator_id: a foreign key relationship with moderator who made the decision
   - media_identifiers: a **many-to-many** relationship to the media the decision is related to, implemented by a join table.
     - Because deindexing media removes them from the API database, the identifiers needs to be non-constrained, so not a foreign key, just an indexed non-unique UUID column using the [`db_constrained` argument](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.ManyToManyField.db_constraint)
   - explanation: moderator notes explaining the decision; optional
   - action: slug column of the action taken
     - **`marked_sensitive`** (marked a work as sensitive)
     - `deindexed_copyright` (deindexed the work from Openverse search due to copyright)
     - `deindexed_sensitive` (deindexed the work from Openverse search due to sensitivity)
     - `rejected_reports` (the reviewed reports were rejected)
     - **`deduplicated_reports`** (the reviewed reports were marked as duplicates, in other words, re-reports that did not result in deindexing but were also accurate, for example, re-reports of a sensitive work that add information regarding sensitivity but do not require deindexing)
     - **`reversed_deindex`** (undo a previous deindexing decision)
     - **`reversed_mark_sensitive`** (undo a previous decision marking the record sensitive)
- Create an AbstractModerationDecision for each media type
- Add new nullable `decision_id` column to the Report models
- Remove all logic from the Report models’ `save` methods

Blocks https://github.com/WordPress/openverse/issues/3638, https://github.com/WordPress/openverse/issues/3641, https://github.com/WordPress/openverse/issues/3840, and https://github.com/WordPress/openverse/issues/3844

dhruvkb:
## Fixes

Fixes #3636 by @sarayourfriend 

## Description

This PR

- adds the moderation decision models (abstract and concrete for both media types) 
- adds a `decision` field on the report models
- deprecates the `status` field on the report models to use the decision model
  - usages of this field are replaced with the properties based on the `decision` field
  - this field can be removed in the future by the PR for #3642
- updates the admin UI for reports to account for changed fields
- refactors code to reduce duplication by moving common code to parent class
- fixes a validation issue in the `AbstractMediaReport` model (and update tests)

## Note

The model uses some different names to match the naming scheme used by other API models. These are small, positive changes and should not block merge.

## Testing Instructions

1. Observe the admin views for the reports and decisions.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/constants/moderation.py
```diff
@@ -0,0 +1,23 @@
...
+    MARKED_SENSITIVE = "mark", "Marked sensitive"
+
+    DEINDEXED_COPYRIGHT = "deidx_copyright", "Deindexed (copyright)"
+    DEINDEXED_SENSITIVE = "deidx_sensitive", "Deindexed (sensitive)"
+
+    REJECTED_REPORTS = "rej", "Rejected"
+    DEDUPLICATED_REPORTS = "dedup", "De-duplicated"
+
+    REVERSED_MARK_SENSITIVE = "rev_mark", "Reversed deindex"
+    REVERSED_DEINDEX = "rev_deidx", "Reversed mark sensitive"
```
sarayourfriend, line 19: What's the reason to abbreviate the slugs? We don't need to save space in these fields and it will make it much harder to remember how to query them from PGCLI. I'd much prefer descriptive, fully-spelled out slugs that are obvious and direct references to the "human-readable"/display terms. The last two also have their human-readable strings swapped.

```suggestion
    MARKED_SENSITIVE = "marked_sensitive", "Marked sensitive"

    DEINDEXED_COPYRIGHT = "deindexed_copyright", "Deindexed (copyright)"
    DEINDEXED_SENSITIVE = "deindexed_sensitive", "Deindexed (sensitive)"

    REJECTED_REPORTS = "rejected", "Rejected"
    DEDUPLICATED_REPORTS = "deduplicated", "De-duplicated"

    REVERSED_MARK_SENSITIVE = "reversed_marked_sensitive", "Reversed mark sensitive"
    REVERSED_DEINDEX = "reversed_deindexed", "Reversed deindex"
```

api/api/models/media.py
```diff
@@ -134,16 +135,13 @@ class AbstractMediaReport(models.Model):
...
+    deleted content. Subclasses must populate fields ``media_class`` and ``url_frag``.
     """
 
     media_class: type[models.Model] = None
     """the model class associated with this media type e.g. ``Image`` or ``Audio``"""
-    sensitive_class: type[models.Model] = None
-    """the class storing sensitive media e.g. ``SensitiveImage`` or ``SensitiveAudio``"""
-    deleted_class: type[models.Model] = None
-    """the class storing deleted media e.g. ``DeletedImage`` or ``DeletedAudio``"""
+    url_frag: str = None
```
sarayourfriend, line 143: Can we use the full name, here, rather than an abbreviation? Or else `url_path_part` might be more descriptive still.

I'm fairly certain it's unnecessary if we use `reverse` anyway, though. Even if we still need some reference to the plural media type, ideally this kind of metadata is defined on the `media_class` or in something like the `MediaTypeConfig` dataclasses we have in the test suite, as it's information shared by all related models.

api/api/models/media.py
```diff
@@ -183,60 +181,111 @@ class AbstractMediaReport(models.Model):
...
             )
 
-    def url(self, media_type):
-        url = f"{settings.CANONICAL_ORIGIN}v1/{media_type}/{self.media_obj.identifier}"
+    @property
+    def url(self):
+        origin = settings.CANONICAL_ORIGIN
+        if not origin.endswith("/"):
+            origin = f"{origin}/"
+        url = f"{origin}v1/{self.url_frag}/{self.media_obj.identifier}"
```
sarayourfriend, line 214: This should use reverse: https://docs.djangoproject.com/en/5.0/ref/urlresolvers/#reverse

Is there a need to construct this manually? This manual constructed version is also missing the trailing slash, which reverse should include based on our configuration.

We should also change this to `media_object_url` or something more descriptive of the target of the URL. It took me a while to register the fact that this isn't the url for the report, and I only realised because I ended up spending a couple minutes trying to understand what the report URL would be used for.

This also should not dot-through to `media_obj.identifier` as it will cause an unnecessary DB query. `media_obj_id` can be used instead to avoid the join.


```suggestion
    def media_obj_url(self):
        return reverse("image-detail", args=[self.media_obj_id])
```

This won't give an absolute URL, but the admin can (and should) generate that based on the request, using `request.get_absolute_uri` and pass in the `.media_obj_url` from the report instance.

If we don't do that, then if someone is on `api-production.openverse.org` and clicks on one of these urls, their session will get trashed. They'll need to re-login through Cloudflare access and then again at Django admin, and at that point probably have lost the `redirect_to` query parameter, but at the very least, their work was unnecessarily disrupted.

sarayourfriend (changes requested):
Looks pretty good so far, but I have two significant requests for changes:

1. Lets use full-names for the decision reason slugs, unless there's a really good reason not to that I am not seeing. If that is the case, then please document the reason in code.
2. We need to stick to standard Django tools for URL building, and should use [`request.build_absolute_uri`](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.build_absolute_uri) for consistent and correct behaviour for each requester.

api/api/constants/moderation.py
dhruvkb: There wasn't any real reason, I've switched to using full names in e32bfa0.

api/api/models/media.py
dhruvkb: I've updated the code to use `reverse` and generate only links with paths in 85e1678. That way the links will not point to a different domain and we won't have to worry about sessions.

localhost_50280_admin_api_imagereport__q=2e22

AetherUnbound (approved):
The model and associated changes all look good! I was able to run the app successfully locally.

api/api/models/media.py
```diff
@@ -183,60 +180,121 @@ class AbstractMediaReport(models.Model):
...
+    def is_moderated(self) -> bool:
         """
+        Determine if the report has been moderated and has an associated
+        decision. Also see ``is_pending``.
 
-        self.clean()
+        :return: whether the report is in the "moderated" state
+        """
 
+        return self.decision_id is not None
```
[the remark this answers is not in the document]
sarayourfriend: It seems duplicative of `not self.is_pending`. It's probably clearer for code needing this information to just reference `is_pending`?

This isn't used yet, so we could just leave it out for now until we need it ([following YAGNI principles](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it))

sarayourfriend (approved):
LGTM! Just one suggestion to leave out the `is_moderated` property until we need it (or maybe even just in preference of using `not is_pending` as a clearer and "only one obvious way").

api/api/models/media.py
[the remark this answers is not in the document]
dhruvkb: I only had them because it felt "more readable" (subjective) to write report.is_pending or report.is_moderated because Python's `not` comes before the noun. 

But I concede, it was not good practice to have two ways to do one thing and also YAGNI. Dropped in 6a806b9.

api/api/constants/moderation.py
```diff
@@ -0,0 +1,23 @@
...
+
+    REJECTED_REPORTS = "rejected_reports", "Rejected"
+    DEDUPLICATED_REPORTS = "deduplicated_reports", "De-duplicated"
+
+    REVERSED_MARK_SENSITIVE = "reversed_mark_sensitive", "Reversed mark sensitive"
+    REVERSED_DEINDEX = "reversed_deindex", "Reversed deindex"
+
+    @property
+    def is_reversal(self):
+        return self in [self.REVERSED_DEINDEX, self.REVERSED_MARK_SENSITIVE]
```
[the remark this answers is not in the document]
dhruvkb: Yes, thank you.

Issue #4005, opened 2024-04-02
Set cache control header on API responses


sarayourfriend:
## Problem

We do not set cache-control headers on our API responses at the origin, and instead rely on Cloudflare to set these.

While we can continue configuring specific cache-control settings, Cloudflare is able to interpret cache control headers from upstream responses and use those settings for edge TTL. That would simplify our Cloudflare configuration by removing the details of individual endpoints and allow that to live instead in the API code, where it probably belongs.

This wouldn't preclude us adding additional cache settings in Cloudflare if we ever wanted to.

## Description

Add the [`cache_control`](https://docs.djangoproject.com/en/5.0/topics/http/decorators/#django.views.decorators.cache.cache_control) decorator to the API views, with TTL as listed. The TTLs are taken from our Cloudflare configuration and for this issue we'll take for granted that they are correct, and defer modifications to these TTLs to a future issue if needed.

- Stats endpoint: 259200 (3 days)
- Thumbnail endpoint: 31536000 (1 year)
- Waveform endpoint: 31536000 (1 year)
- Default for all other endpoints: 2592000 (1 month)

Lastly, explicitly exclude the `auth_tokens` routes by adding [`never_cache`](https://docs.djangoproject.com/en/5.0/topics/http/decorators/#django.views.decorators.cache.never_cache). This again reduces configuration necessary in Cloudflare and moves it upstream into our code.

## Additional context

This is a supplement to https://github.com/WordPress/openverse-infrastructure/issues/777, removing the need to move many of the rules defined in our Cloudflare configuration to the the openverse.org zone. As such, I'm setting it to high priority and into the todo list and starting work on it ASAP as part of the openverse.org migration project. This doesn't block that issue because, as I've written in the issue, we'll just ignore the .engineering caching rules and add a single rule to instruct Cloudflare to use the upstream header.

sarayourfriend:
This will require upgrading to Django 5, due to the need for async support on `cache_control` decorator; or to backport support for async support in that decorator from Django 5 into 4.2.

sarayourfriend:
I'm going to close this issue as won't do, at least for now. It's out of scope of the original project and potentially way way more complicated than I originally imagined. Thanks @obulat for pointing out some the issues in the linked PR. Closing the PR as well.

Issue #4015, opened 2024-04-03
Update the API to Django 5


sarayourfriend:
## Problem

Django 5 is out and the initial bugs are sorted. It has some significant QoL and other improvements for async routes, and we should update to Django 5.

## Description

Review the Django 5 release notes and cross-reference breaking changes with our API codebase.

https://docs.djangoproject.com/en/5.0/releases/5.0/#backwards-incompatible-changes-in-5-0

At a glance, I don't think we are affected by the breaking changes, **but please take a close look at each item to confirm this**.

Update the API to Django 5 and make any necessary changes.

Remove any backported code that may exist to add Django 5 features before we've upgraded (e.g., cache_control backport for https://github.com/WordPress/openverse/issues/4005).

sarayourfriend:
@WordPress/openverse-maintainers would be great to get this issue going soon. 4.2 is an LTS release and supported until 2026, so we are not at risk of falling out of support. However, as noted in the issue description, 5 and 5.1 already have loads of improvements we would benefit from, including and especially for async routes.

5.1, for example, finally has support for connection pooling Postgres: https://docs.djangoproject.com/en/5.1/releases/5.1/#postgresql-connection-pools

It also displays "operation categories" for migrations, making it easier to understand what the effect of a migration will be, potentially helping avoid unexpected types of migrations (like ones that might not be zero-downtime), without relying on careful, manual PR-time review of instructions: https://docs.djangoproject.com/en/5.1/ref/migration-operations/#django.db.migrations.operations.base.OperationCategory

[5.1 has a few deprecated features, which may require changes after the fact to address](https://docs.djangoproject.com/en/5.1/releases/5.1/#features-deprecated-in-5-1). The main one we might need to address is deprecation of position arguments to `Model.save()` and `asave()`. We do not need to address those immediately upon upgrading, as they are only deprecated, not removed.

[5.0 added support for ASGI disconnects](https://docs.djangoproject.com/en/5.1/releases/5.0/#asynchronous-views), which would be nice to have in order to safely halt longer running requests in async views (e.g., aborting outbound aiohttp requests in the thumbnail route, killing audiowaveform processes in the waveform route, etc).

[5.1 added support for `ModelAdmin.list_display` to target boolean attributes, and `__` attributes for joining across models](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display), allowing us to delete this code (and especially the comment explaining the one-liner), for example:

[...]

PR #4016, opened 2024-04-03
Add cache control headers to API responses

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/4005 by @sarayourfriend 

## Description

Adds the `cache_control` decorator to our API views, or `never_cache` when appropriate.

Most of the complexity in this PR comes from needing to backport code from Django 5 to support async view functions in `never_cache` and `cache_control`, as well as an additional shim to work around a limitation of `method_decorator`.

Other than that, the main "actual" interesting code is the additional method on the search parameters serializer to determine whether the request is eligible for public caching. Public caches like Cloudflare will not cache responses at the edge if authentication headers are on the request, but will do so if you explicitly set `public` in the `cache-control` header. The PR sets `public` in the header for any requests that do not use search features that are limited to authenticated users only.

That is the only route (other than admin, of course) where authentication causes a difference in behaviour beyond rate limiting, at least that I can think of, so the other routes can all have `public` so that the edge always caches responses for those routes (given that authed and anon users are getting the same responses back in those cases).

## Testing Instructions

Check out the cache control configurations and confirm they make sense for each route.

Run the app locally using `just api/up` and test the views. In the rendered DRF page, you will see the cache-control header listed in the headers for the response, so it's easy to debug this visually.

Confirm the unit tests cover the new behaviours defined in the serializer method. I have avoided testing the configuration of the cache control for views where it was purely a question of configuration, rather than some business logic that needed to succeed like in the case of the variably public cache control of the search routes.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -13,7 +13,7 @@ from elasticsearch_dsl.response import Hit
 from api.constants import sensitivity
 from api.constants.licenses import LICENSE_GROUPS
 from api.constants.parameters import COLLECTION, TAG
-from api.constants.sorting import DESCENDING, RELEVANCE, SORT_DIRECTIONS, SORT_FIELDS
+from api.constants.sorting import DEFAULT_SORT_DIRECTION, DEFAULT_SORT_FIELD, SORT_DIRECTIONS, SORT_FIELDS
 from api.controllers import search_controller
 from api.models.media import AbstractMedia
 from api.serializers.base import BaseModelSerializer
@@ -235,7 +235,7 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
         help_text="The field which should be the basis for sorting results.",
         choices=SORT_FIELDS,
         required=False,
-        default=RELEVANCE,
+        default=DEFAULT_SORT_FIELD,
     )
     unstable__sort_dir = serializers.ChoiceField(
         source="sort_dir",
@@ -243,7 +243,7 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
         "`relevance`.",
         choices=SORT_DIRECTIONS,
         required=False,
-        default=DESCENDING,
+        default=DEFAULT_SORT_DIRECTION,
     )
     unstable__authority = serializers.BooleanField(
         label="authority",
@@ -427,10 +427,10 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
         return self._truncate(value)
 
     def validate_unstable__sort_by(self, value):
-        return RELEVANCE if self.is_request_anonymous() else value
+        return DEFAULT_SORT_FIELD if self.is_request_anonymous() else value
 
     def validate_unstable__sort_dir(self, value):
-        return DESCENDING if self.is_request_anonymous() else value
+        return DEFAULT_SORT_DIRECTION if self.is_request_anonymous() else value
 
     def validate_unstable__authority(self, value):
         return False if self.is_request_anonymous() else value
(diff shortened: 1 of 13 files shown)


api/api/utils/cache_control/__init__.py
```diff
@@ -0,0 +1,26 @@
+from functools import wraps
+
+from api.utils.cache_control.backports import cache_control, never_cache  # noqa: E501
+
+
+def async_method_cache_control(**cache_control_kwargs):
+    """
+    Decorate an async method with `cache_control`.
+
+    Only necessary for async view methods because Django's `method_decorator`
```
obulat, line 10: Should we add something like `TODO: Remove when migrating to Django 5, see https://github.com/WordPress/openverse/issues/4015`?

api/api/serializers/media_serializers.py
```diff
@@ -486,6 +486,34 @@ def validate(self, data):
...
+        For Openverse search, any search request that does not utilise auth-only features
+        can be publicly cached.
+        """
+        if self.is_request_anonymous():
+            return True
+
+        validated_data = self.validated_data
+        # Check if any auth-only features are used.
+        # Only allow public cache control if
+        return (
```
obulat, line 509: Might be more readable if we extract this as a function, something like `request_includes_auth_only_features`

api/api/views/media_views.py
```diff
@@ -202,7 +226,9 @@ def get_media_results(
 
     # Extra actions
 
+    # stats can always be cached
     @action(detail=False, serializer_class=ProviderSerializer, pagination_class=None)
+    @method_decorator(cache_control(max_age=ttl(days=3), public=True))
```
obulat, line 231: Would we reset cache in Cloudflare after adding a new provider/running data ingestion?

api/api/views/media_views.py
```diff
@@ -265,6 +294,8 @@ async def get_image_proxy_media_info(self) -> image_proxy.MediaInfo:
         ],
     )
 
+    # Thumbnails can always be cached
+    @async_method_cache_control(max_age=ttl(weeks=52), public=True)
```
obulat, line 298: What would happen if the thumbnail response is flaky, or returns an error only on the first request? Would the error be cached, or are errors never cached?

obulat (commented):
The changes look great, and I will re-review tomorrow after running this locally.

Do you think we can add some linting rule to check if the endpoint has a cache-control decorator?

api/api/views/media_views.py
sarayourfriend: Yes, but 3 days is, I think, a pretty acceptable latency in that respect, considering it already takes a week to get new data anyway :shrug: Cache purges in Cloudflare cost money if you do them often, so it's better to just find a value that works. We can adjust as needed.

api/api/views/media_views.py
sarayourfriend: That is an interesting question. I need to look deeper into this in Django and make sure I understand everything correctly before saying definitely.. but I _think_ that because an error response should be raised, it would break the view function's routine before the cache control header is set. So when we raise the upstream thumbnail failure, that will stop the rest of the view from happening and hand off to the error handler middleware instead, which should prevent the cache control headers from being set because the cache_control decorator sets them on the response after the view function is finished.

But, again, I need to confirm this by trying it out locally and seeing what the behaviour is.

If it does not behave that way, we have two options:

1. We can make write a wrapper to the `patch_cache_control_headers` function to only patch onto status codes in a range we've defined, either per route or following a global setting for the app.
2. We can close the cache control issue as won't do, and instead rely on heuristic caching in browsers and define only edge TTLs in Cloudflare for specific routes.

These are the status codes that are described as cacheable by default when cache control headers are not defined (e.g., if we don't do this PR): https://datatracker.ietf.org/doc/html/rfc7231#section-6.1

We could, instead of setting the cache-control header, set the last-modified header. Then browsers can use heuristic caching as described [here in the RFC for cache controls](https://datatracker.ietf.org/doc/html/rfc7234#section-4.2.2). We would define edge caching behaviour in Cloudflare _or_ we can set _both_ and instruct Cloudflare to use the cache control header for edge TTL and re-write response headers to exclude the cache control header sent to browsers. So then Cloudflare uses the cache-control for edge caching instructions (with limitations on the range we allow) and all responses have headers re-written to exclude cache-control when forwarded to the requester.

That blends the two approaches and allows us to control caching entirely from the application code, without having to touch Cloudflare cache rules for the API. The frontend really needs different caching heuristics altogether, like maybe only an hour or caching, because UI changes are independent of the underlying data changes possible from the API.

I need to confirm two things though:

1. Do the cache-control headers get set for error responses; in other words, does raising the error response cause the cache control decorator to be bypassed and never run. And do we handle errors consistently enough to rely on this (i.e., do we _raise_ rather than _return_ error responses).
2. Does Cloudflare's browser TTL bypass setting allow completely leaving off the cache-control header or does it just re-write it to match edge TTL?

Our default browser TTL is 4 hours, which you can see on a request like https://api.openverse.engineering/v1/, and is configured entirely by Cloudflare.

sarayourfriend:
Drafting to take a pause and consider the issues in this comment thread: https://github.com/WordPress/openverse/pull/4016#discussion_r1549583753

I was working with some assumptions about error handling, but I realised I had not confirmed they (they are really just assumptions), and in thinking and reading more about the cache-control header, there might be a better way to approach this.

sarayourfriend:
Closing as won't do, for now. We can revisit this in the future if we want to change the way we produce cache instructions: https://github.com/WordPress/openverse/issues/4005#issuecomment-2063045726

Issue #3876, opened 2024-03-05
Switch to Docker Compose v2


zackkrida:
## Problem

Docker recommends users use `docker compose` instead of `docker-compose` as of July 2023, when version 1 of the compose api stopped being supported. 

> Important
> 
> Docker's documentation refers to and describes Compose V2 functionality.
> 
> Effective July 2023, Compose V1 stopped receiving updates and is no longer in new Docker Desktop releases. Compose V2 has replaced it and is now integrated into all current Docker Desktop versions. For more information, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate).
>
> -- https://docs.docker.com/compose/

Most fresh installs of Docker no longer include the `docker-compose` alias. 

## Description

Update all of our justfiles and build steps to use `docker compose` instead of `docker-compose`. 
As a bonus, this would allow for predictability in service container names, which use a `-` instead of `_` separator in Compose V2[^1].

## Alternatives

Keep using old software :cry: 

[^1]: https://docs.docker.com/compose/migrate/#service-container-names

sarayourfriend:
Duplicate issue https://github.com/WordPress/openverse/issues/3973 has some additional information that could be helpful for completing this and cleaning up our code (e.g., massively reduce the complexity of `just ps` script)

obulat:
I updated the priority to critical because the CI is intermittently failing due to `docker-compose` not being available in GitHub CI runners.

AetherUnbound:
Resolved by #4017

sarayourfriend:
@AetherUnbound can you clarify what work that is? The basic part is done, separate issues could be helpful to break down further tasks.

AetherUnbound:
Specifically this piece from @obulat's PR description:

> This PR only replaces `docker-compose` in `justfile`s and documentation with `docker compose`.
> It does not simplify the `docker ps` script.
> @zackkrida mentions the replacement of underscores with dashes for container names in the issue: should this be done manually? Would it also affect production? 

If this piece is not considered important then we can close the issue.

sarayourfriend:
I think they are important, but should be separate issues. We're not using compose v1 anymore, so at the very least the title of the issue and the main portion of it is complete. The others are far less important "nice to have" clean ups, right?

dhruvkb:
> As a bonus, this would allow for predictability in service container names, which use a `-` instead of `_` separator in Compose V2[^1]

[...]

PR #4017, opened 2024-04-03
Replace docker-compose with docker compose in just scripts and docs

obulat:
## Fixes

Related to #3876 by @zackkrida

## Description

This PR only replaces `docker-compose` in `justfile`s and documentation with `docker compose`.
It does not simplify the `docker ps` script.
@zackkrida mentions the replacement of underscores with dashes for container names in the issue: should this be done manually? Would it also affect production? 

I think it's best to keep the scope of this PR small to fix the critical issue of CI failures due to `docker-compose` command not being found.

## Testing Instructions

The CI should pass.
All of the `just` commands that used to use `docker-compose` should work as usual.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "2.4"
+version: "3.4"
 
 # Common build configuration for Airflow
 # Extension field, see https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.4"
+version: "2.4"
 
 # Common build configuration for Airflow
 # Extension field, see https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields
--- a/documentation/frontend/reference/playwright_tests.md
+++ b/documentation/frontend/reference/playwright_tests.md
@@ -31,7 +31,7 @@ cross-platform browser differences from creating flaky test behavior. We run
 both end-to-end and visual-regression tests in the same container to save from
 having to run the Nuxt production build twice.
 
-Having docker and Compose V2 is a pre-requisite to running the playwright tests
+Having Docker and Compose V2 is a pre-requisite to running the Playwright tests
 locally. Please follow
 [the relevant instructions for your operating system for how to install docker and docker-compose](https://docs.docker.com/get-docker/).
 If you're on Windows 10 Home Edition, please note that you'll need to
--- a/documentation/frontend/reference/playwright_tests.md
+++ b/documentation/frontend/reference/playwright_tests.md
@@ -31,8 +31,8 @@ cross-platform browser differences from creating flaky test behavior. We run
 both end-to-end and visual-regression tests in the same container to save from
 having to run the Nuxt production build twice.
 
-Having docker and docker-compose is a pre-requisite to running the playwright
-tests locally. Please follow
+Having docker and Compose V2 is a pre-requisite to running the playwright tests
+locally. Please follow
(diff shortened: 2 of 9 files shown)


ingestion_server/test/integration_test.py
```diff
@@ -166,10 +166,11 @@ def _load_data(conn, table_names):
 
 
 def _compose_cmd(compose_path, cmd: list[str], **kwargs):
-    """Run a Docker Compose command"""
+    """Run a Docker compose command"""
```
dhruvkb, line 169: This was okay as it was.

documentation/frontend/reference/playwright_tests.md
```diff
@@ -31,8 +31,8 @@ cross-platform browser differences from creating flaky test behavior. We run
 both end-to-end and visual-regression tests in the same container to save from
 having to run the Nuxt production build twice.
 
-Having docker and docker-compose is a pre-requisite to running the playwright
-tests locally. Please follow
+Having docker and Compose V2 is a pre-requisite to running the playwright tests
```
dhruvkb, line 34:
```suggestion
Having Docker and Compose V2 is a pre-requisite to running the Playwright tests
```

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.4"
+version: "3.4"
```
dhruvkb, line 1: This is a big change. I'm not sure we should be switching to v3 without a thorough audit of the effects. Afaik, v2 is still supported with Compose v2 so I would prefer to not change this.

dhruvkb (changes requested):
The changes look good to me.

The underscore and dash change is a non-issue for us especially if we use the `docker compose exec` command with service names instead of container names.

I have more confidence in the changes because many folks on the team, myself included, have been using Compose v2 for a while and it has not affected anything for us.

I do have some nits about capitalisation but those are not blockers. The blocker is the v2 &rarr; v3 upgrade in `docker-compose.yml`.

docker-compose.yml
obulat: The comment in the same file links to documentation for v3 of compose files, so I decided to try fixing this. The JetBrains IDEs usually complain about some parts of this file (mainly, the extension fields), but when you update the version to 3.4, the red disappears. However, I re-checked and _extension fields_ were introduced in version 2.4[^1], so it's confusing.

I'll revert this change.
[^1]: https://docs.docker.com/compose/compose-file/compose-versioning/#version-24

obulat, line 1:
```suggestion
version: "2.4"
```

obulat: Revert the update of docker-compose.yml version

ingestion_server/test/integration_test.py
obulat, line 169:
```suggestion
    """Run a Docker Compose command"""
```

docker-compose.yml
dhruvkb: IIRC there was some issues with the resource limits when using version 3.x. But we can eventually switch up once we verify things are working normally. Can you create a an issue to do the audit and upgrade?

dhruvkb (approved):

docker-compose.yml
obulat: Done: #4020

AetherUnbound (approved):
Thanks for making this change! I encountered it yesterday but it was successful after a second run. Weird!

PR #4026, opened 2024-04-03
Implementation plan: Ingestion server removal

### LINKED ISSUE #3981: Implementation Plan for Ingestion Server removal
## Problem

Write the Implementation Plan for the ingestion server removal project.

## Description

As decided in the [project proposal](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240319-project_proposal.html), the single implementation plan should cover: 

* mechanism for spinning up indexer workers from Airflow, along with required infrastructure changes
* process for moving the rest of the data refresh into Airflow (data copy and cleaning, table/index promotion, etc)

stacimc:
## Fixes

Fixes #3981 by @stacimc

## Description

This PR adds the implementation plan for removing the ingestion server, and moving the data refresh into Airflow. 

## Reviewers

I have requested reviews from @sarayourfriend and @AetherUnbound. @sarayourfriend's deep knowledge of the infrastructure, and in particular the ongoing work to the catalog deployments, would be extremely beneficial. @AetherUnbound also has expertise in our infrastructure and catalog, and in orchestrating similar distributed tasks using Airflow.

## Testing Instructions

Please read the IP following the decision-making process defined below.

## Decision-making process

### Current round

This discussion is currently in the **Decision round**. The deadline for review of this round is April 15 2024. Note the deadline would typically be April 16, but is extended to account for my own AFK.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
+++ b/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
@@ -135,7 +135,17 @@ approaches:
 Maintaining the
 [EC2 worker instances](https://calculator.aws/#/createCalculator/ec2-enhancement),
 provided we continue to automatically stop the instances when not in use, would
-be cheaper than using [ECS](https://calculator.aws/#/createCalculator/Fargate).
+be cheaper than using
+[ECS](https://calculator.aws/#/createCalculator/Fargate)[^1], given equivalent
+resources.
+
+Since our existing indexer workers do not use all of the resources available, we
+can also consider reducing resources for the workers. For EC2, we are already
+using the cheapest possible offering that will comfortably accommodate our
+memory consumption. Since ECS is more flexible and allows scaling vCPU and RAM
+independently, it is possible that ECS could be cheaper than EC2 with careful
+adjustment of resources[^1].
+
 Notably both solutions are _signifcantly_ cheaper than the current
 implementation, because the bulk of the cost comes from constantly running the
 two m5.xlarge EC2 instances for the ingestion servers.
@@ -189,7 +199,7 @@ in local development to spin up a worker as a local Docker container.
 This is still relatively complex, however, because the worker Docker containers
 must be launched from Airflow, which is itself dockerized. This requires either
 using a Docker-in-Docker approach, which should typically be avoided, or
-modifying the configuration to allow Airflow to launch sibling containers[^1].
+modifying the configuration to allow Airflow to launch sibling containers[^2].
 Requirements differ across operating systems, yet this **must** be tested and
 maintained across platforms because the data refresh is used as a critical step
 in setting up our local development environment.
@@ -591,5 +601,10 @@ monitor the first production data refreshes closely.
 <!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
 
 [^1]:
+    See
+    https://github.com/WordPress/openverse/pull/4026#pullrequestreview-1978477921
+    for specific cost assessment, courtesy of @sarayourfriend
+
+[^2]:
 
 https://towardsdatascience.com/using-apache-airflow-dockeroperator-with-docker-compose-57d0217c8219
--- a/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
+++ b/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
@@ -114,8 +114,12 @@ of maintaining remote workers that can be run only when needed.
 
 Two other options were evaluated for this IP:
 
-- Simply move the 8 total existing EC2 worker instances into the catalog, and
-  connect to them directly from Airflow.
+- Keep the EC2 instances as they are now, but connect to them directly from
+  Airflow. Rather than managing the 8 total EC2 instances directly, we will
+  instead set up an EC2 Auto Scaling group for each environment, with an initial
+  desired capacity of 0. The data refresh DAGs will spin up instances by
+  increasing this desired capacity. (More details follow later in this
+  document.)
 - Remove the EC2 instances entirely in favor of a new ECS task definition using
   the `indexer-worker` image, which would remove all API code and contain only
   the reindexing script. Airflow would spin up the appropriate number of ECS
@@ -200,40 +204,38 @@ into the status of the task from Airflow.
 
 #### Impact on Deployments
 
-This is the greatest strength of the ECS approach. Each time an ECS task is spun
-up on Fargate, it will pull the Docker image configured in the task definition.
-By using the `latest` tag in this configuration we can ensure that the latest
-indexer-worker image is pulled each time, which means that any changes to the
-reindexing code will become live in production as soon as the new docker image
-is published after merging on `main`.
+This is a great strength of both approaches, each of which would eliminate the
+need for deployments when changes are made to the data refresh code (including
+changes to the indexer workers).
 
-For the EC2 approach, any changes to the reindexing script or its dependencies
-would require deploying the catalog (and by extension the indexer workers).
-However:
+Each time an ECS task is spun up on Fargate, it will pull the Docker image
+configured in the task definition. By using the `latest` tag in this
+configuration we can ensure that the latest indexer-worker image is pulled each
+time, which means that any changes to the reindexing code will become live in
+production as soon as the new docker image is published after merging on `main`.
 
-- No deployment would be required for changes to the vast majority of the data
-  refresh implementation, including the data copy, filtered index creation,
-  alias management, and all aspects of orchestrating the indexer workers.
-- _If_ deploying the indexer-worker becomes a frequent issue in the future, we
-  could iterate on the data refresh DAG to automatically update the workers. For
-  example we could use an Airflow
-  [SSHOperator](https://docs.aws.amazon.com/mwaa/latest/userguide/samples-ssh.html)
-  to connect to the instances and pull the latest image before triggering the
-  task. I think this is unlikely enough to be a problem that this does not need
-  to be a requirement of the initial implementation.
+The EC2 approach uses an ASG to achieve a similar result. Because the ASG will
+actually terminate (rather than stop) the instances when their work is complete
+and start _new_ instances for each data refresh, we can pull the Docker image
+with the `latest` tag in the `user_data` script so that the latest Docker image
+is pulled each time a new data refresh starts. By using AWS Systems Manager
+parameters instead of AMI IDs in the launch template, we can even use the ASG to
+automatically use new AMI IDs without needing to deploy a new launch template
+each time a system dependency is updated.
 
 ### Conclusion
 
-The ECS approach allows us to remove 100% of the server management code, and
-eliminates the need for deployments in essentially every case (unless the task
-definition itself changes). However it is more expensive, more difficult to
-implement, and requires a special secondary implementation for running locally
-that may be prone to cross-platform issues.
+Both approaches allow us to eliminate the need for deployments in almost all
+cases, except for when changes are made to the task definition and launch
+template respectively.
 
-The EC2 approach on the other hand is cheaper, quicker to implement, and gets us
-the vast majority of what we want in terms of removing complexity and reducing
-the need for deployments. Consequently I argue here that the EC2 approach is the
-one we should pursue.
+The main advantage of the ECS approach is that it allows us to remove 100% of
+the server management code. However, it is more difficult to implement, likely
+more expensive, and requires a special secondary implementation for running
+locally that may be prone to cross-platform issues. The EC2 approach on the
+other hand is cheaper, quicker to implement, and gets us the vast majority of
+what we want in terms of removing complexity. Consequently I argue here that the
+EC2 approach is the one we should pursue.
 
 ## Step-by-step plan
 
@@ -262,8 +264,8 @@ developed as separate DAGs alongside the current ones.
    indexer worker image locally.
 1. Add the distributed reindexing step to the DAG (excludes infrastructure
    work).
-1. Add the `catalog-indexer-worker` to the Terraform configuration for the
-   catalog, and add steps to deploy them in the related Ansible playbooks.
+1. Set up the necessary resources for the ASGs for staging and production in the
+   catalog Terraform configuration.
 1. Add all remaining steps to the data refresh DAGs:
    `Create and Populated Filtered Index`, `Reapply Constraints`,
    `Promote Table`, `Promote Index`, `Delete Index`.
@@ -336,9 +338,9 @@ server's work:
 
 ### Implement new catalog indexer worker
 
-In this step we will create the new catalog-indexer-worker Docker image, and add
-distributed reindexing to the new DAGs. This step does not include the
-infrastructure work to actually create the new EC2 instances.
+In this step we will create the new catalog-indexer-worker Docker image. This
+step does not include adding the orchestration steps to the DAG, or the
+infrastructure work to actually create the ASGs.
 
 First we will create a new `indexer-worker` directory under
 `catalog/dags/data_refresh`, which will contain the contents of the new indexer
@@ -390,22 +392,26 @@ refreshes locally and in production).
 
 In this step we will add tasks to the data refresh DAGs to orchestrate the
 distributed reindex. At the end of this step, it will be possible to run a
-distributed reindex locally. Because the infrastructure work to create the EC2
-instances is not complete, it can not be run on production yet. The following
-code can all be refactored from
+distributed reindex _locally_, but because the infrastructure work to create the
+ASGs is not complete, it can not be run on production yet. The following code
+can all be refactored from
 [`distributed_reindex_scheduler.py`](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/distributed_reindex_scheduler.py).
 
-- Use Airflow's
-  [EC2Hook::describe_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.describe_instances)
-  to get a list of internal URLs bound to the indexer workers for the
-  appropriate environment. In the local environment, this will instead return
-  the URL for the catalog-indexer-worker Docker container.
 - Use
-  [EC2Hook::start_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.start_instances)
-  to start the appropriate EC2 instances. This step should skip in a local
-  environment.
-- Use dynamic task mapping to generate a Sensor for each of the expected indexer
-  workers, and ping its `healthcheck` endpoint until it passes.
+  [`describe_auto_scaling_groups`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/describe_auto_scaling_groups.html)
+  and filter by tags to select the appropriate ASG for the desired environment.
+  (Skips in local env.)
+- Use
+  [`set_desired_capacity`](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/set_desired_capacity.html)
+  to increase the desired capacity of the ASG to the desired number of workers,
+  depending on the environment. This will cause the ASG to begin spinning up
+  instances. (Skips in local env.)
+- Use
+  [`describe_auto_scaling_groups`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/describe_auto_scaling_groups.html)
+  to poll the ASG until all instances have been started, and get the EC2
+  instance IDs. (Skips in local env.)
+- Use dynamic task mapping to generate a Sensor for each of the instance IDs,
+  and ping its `healthcheck` endpoint until it passes.
 - Use dynamic task mapping to distribute reindexing across the indexer workers
   by first calculating `start` and `end` indices that will split the records in
   the media table into even portions, depending on the number of workers
@@ -414,55 +420,44 @@ code can all be refactored from
     `end_index` it should handle
   - Use a Sensor to ping the worker's `task/{task_id}` endpoint until the task
     is complete, logging the progress as it goes
-  - Use the
-    [EC2Hook::stop_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.stop_instances)
-    to shut the instance down. It should use the
-    [`NONE_FAILED` TriggerRule](https://github.com/WordPress/openverse/issues/286)
-    to ensure that the instances are shut down, even if there are upstream
-    failures. This step should skip in a local environment.
+- Use
+  [`terminate_instance_in_auto_scaling_group`](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/terminate_instance_in_auto_scaling_group.html)
+  to terminate the instance. Make sure to set `ShouldDecrementDesiredCapacity`
+  to `True` to ensure that the ASG does not try to replace the instance. This
+  task should use the
+  [`NONE_FAILED` TriggerRule](https://github.com/WordPress/openverse/issues/286)
+  to ensure that the instances are terminated, even if there are upstream
+  failures. (Skips in local env.)
+- Finally, after all tasks have finished (regardless of success/failure), we
+  should have a cleanup task that calls `set_desired_capacity` to 0. Generally
+  this should be a no-op, but if an instance crashes during reindexing (rather
+  than simply failing during reindexing) the ASG will spin up a replacement and
+  Airflow will not automatically clean it up. This task ensures that any
+  dangling instances are terminated.
+
+```{note}
+It is not possible to retry a single indexer worker with this set up, because once a worker fails the instance is actually terminated (rather than simply stopped). If a task that triggers a reindex is cleared after an instance has been terminated, it will simply fail. The entire reindex must be restarted from the first step in this task group.
+
+However, there is a valuable tradoff to this approach: it ensures that all of the indexer workers in a data refresh are identical, while still allowing us to avoid manual deployments every time the indexer logic changes. For example, imagine some changes to the reindexing logic are merged to `main` while a data refresh is actively underway, and a new Docker image is published. If one indexer worker failed, and it were possible to retry **just** that indexer worker, it would use the new Docker image -- leading to inconsistency in the behavior of different workers within a single data refresh.
+```
 
 ### Create the Terraform and Ansible resources needed to deploy the new indexer workers
 
-In this step we will add the resources needed to actually deploy the new indexer
-workers. At the time that this IP is being written, work to
-[deploy Airflow with Ansible](https://github.com/WordPress/openverse-infrastructure/pull/829)
-is actively underway. This IP builds off of this implementation and is blocked
-by its success in production.
+In this step we will add the resources needed to actually configure the ASGs.
 
-The goals are to:
+Some important notes:
 
-- Add the Terraform resources for the indexer worker to
-  [`next/production/airflow.tf`](https://github.com/WordPress/openverse-infrastructure/blob/7a8fdd02dc448cfb9f2991f7b4ece1da7349209c/next/production/airflow.tf),
-  using the generic ec2-service module. The configuration can be copied from the
-  [current configuration](https://github.com/WordPress/openverse-infrastructure/tree/main/modules/services/ingestion-server)
-  in the ingestion-server.
-  [This PR](https://github.com/WordPress/openverse-infrastructure/pull/829) can
-  be used as a guide.
-- Update the Airflow
-  [Ansible playbooks](https://github.com/WordPress/openverse-infrastructure/blob/7a8fdd02dc448cfb9f2991f7b4ece1da7349209c/ansible/roles/airflow/tasks/airflow.yml)
-  to deploy the indexer workers alongside the catalog.
-
-The existing logic for deploying Airflow can be followed closely when adding the
-catalog-indexer-worker. The majority of this is already implemented and the
-existing conventions should be followed, including planning by default, checking
-whether a data refresh is currently in progress, skipping if no differences
-exist, and checking for positive confirmation before deployment. It is most
-useful to note what differs about the catalog-indexer-workers:
-
-- **Important**: when the compose stack is restarted for the indexer workers, it
-  is critical that the indexer workers must shut themselves down as soon as
-  setup is complete. They will be spun up automatically by the DAG when needed
-  and should not be left running.
 - Currently, the staging and production workers are split into separate
   environments (i.e., a staging deploy is used to deploy the 2 staging workers
   separately from the 6 production workers). It is more accurate to view all 8
   workers as production instances (i.e., part of the production catalog
   deployment), which merely _operate_ on different environments. As such all 8
-  should be part of the production catalog deploy, but each instance should have
-  an environment `tag` applied which indicates which environment it is intended
-  to be used for.
+  should be part of the production deployment, but two separate ASGs which are
+  tagged indicating their intended environment.
 - The playbooks must be updated to check if any of the four _new_ data refresh
   DAGs are running before deploying, as well.
+- The `user_data` script should be updated to pull the Docker image with the
+  `latest` tag.
 
 ### Add remaining steps to the Data Refresh DAGs
 
@@ -556,6 +551,11 @@ The alternative options of using an ECS approach or performing the reindex
 entirely in Airflow are discussed at length in the
 [Approach to the Distributed Reindex](#approach-to-the-distributed-reindex)
 section.
+
+It is also possible to use EC2 instances but manage them directly in Airflow
+using EC2 operators to start and stop the instances as needed. However, more
+infrastructure work is required in this approach, and we would require
+deployments whenever there are code changes in the indexer workers.
 
 ## Blockers
 
--- a/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
+++ b/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
@@ -600,6 +600,8 @@ monitor the first production data refreshes closely.
 
 <!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
 
+- [Search relevancy sandbox](https://github.com/WordPress/openverse/issues/392)
+
 [^1]:
     See
     https://github.com/WordPress/openverse/pull/4026#pullrequestreview-1978477921
--- a/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
+++ b/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
@@ -10,7 +10,7 @@
 <!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
 
 - [ ] @sarayourfriend
-- [X] @AetherUnbound
+- [x] @AetherUnbound
 
 ## Project links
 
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+  by first calculating `start` and `end` indices that will split the records in
+  the media table into even portions, depending on the number of workers
+  available in the given environment. Then:
+  - POST to each worker's `reindexing_task` endpoint the `start_index` and
+    `end_index` it should handle
+  - Use a Sensor to ping the worker's `task/{task_id}` endpoint until the task
+    is complete, logging the progress as it goes
+  - Use the
+    [EC2Hook#stop_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.stop_instances)
+    to shut the instance down. This step should skip in a local environment.
```
sarayourfriend, line 415: You mention using `ec2-service` later on, and in thinking further about the implications of that compared to creating EC2 instances directly, I think we have an opportunity to leverage the functionality of an auto-scaling group to trivially get somewhat automatically updating indexer workers, along with simplifying the DAGs set of tasks by removing most of the need for mapped tasks.

When managing EC2 instances directly (what we do currently), we start the instances once when we provision them, and then immediately shut them down. When the instances start again, a cron job is responsible for restarting the docker compose services[^1]. All of this is configured once in the user data script at instance creation. User data scripts run only on instance creation and never again (at least not automatically, I'm sure you could locate the script and run it manually if you wanted to).

[^1]: Which, by the way, really is not necessary if we just used the docker service as intended, with an appropriate restart policy, but that's neither here nor there

With ASGs, however, we have some interesting capabilities that would make deploying even a bumped version tag (if we did not use a new tag to point to the "release" we want). The critical thing to know about ASGs is that they do not rely on _stopped_ instances, but rather _terminated_ instances. Meaning the instances literally get created and destroyed as needed, rather than started and stopped. This is important because it means the user data script runs every time an ASG instance spins up. The user data script is defined in the launch template, which in turn means that a "deployment" for a correctly leveraged ASG should mean just updating the launch template and then replacing each running instance one-by-one (or in whatever pattern makes sense for the given service). If we ran the API or frontend in an ASG again, that's how we would do it, rather than needing to create an entirely new ASG each time.

For the indexer workers, what does this mean? It means we can actually basically ignore Ansible for them, and instead completely leverage Terraform and only Terraform, significantly simplifying their management. Rather than running an Ansible playbook that would one-by-one start and then SSH into each indexer worker, we would just push an update to the ASG launch template. The next time any instance starts in that ASG, it will use the new launch template, and therefore have whatever updates we pushed to it.

This somewhat blends our old approach with the new approach, and basically sets up the best part of ECS (the need to only define a task-definition) with the ASG (by using a launch template, instead of task definition).

Then, in Airflow, rather than starting and stopping individual EC2 instances, we just do the following:

1. Query for the relevant ASG for the environment we need workers for
2. Issue a `set-desired-capacity` command to the ASG to set it to the capacity used for the data refresh
3. Query the ASG using `describe-auto-scaling-group`, which returns the ASG health and a list of all the instances with health status, and wait until instances are available to accept tasks (some or all are healthy).
4. Continue with the non-ASG approach as described in the IP, just use the ASG's instances list instead of querying for the EC2 instances matching the tag (and of course skip starting the instances).
5. As each instance finishes everything the DAG needs it to do, tell the ASG to spin it down using [manual specific instance termination](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-scaling-manually.html#terminate-an-instance-aws-cli) which also allows us to decrement the desired capacity at the same time. This bascially tells the ASG: terminate this instance, and don't replace it.

This has some significant benefits over the EC2 approach:

1. Deployments will take literally seconds (updating a launch template is nearly instant)
2. Instances do not need to know how to terminate themselves, the DAG will do that for us
3. We can take a completely "hands-free" approach to the instances, thinking only in terms of launch templates and ASGs for the most part, eliminating the need for any new Ansible playbooks.
4. Additionally, we can use SSM parameters for selecting the AMI at launch time rather than using a static AMI ID, meaning we have one less place to worry about updating system dependencies (which is the main headache long-running EC2 instances carry, in my opinion).

In the future, we could even employ a target group with the internal load balancer and use target group stickiness configured with a custom app cookie, where each task responsible for one of the indexer workers generates a unique cookie value for the lifetime of the DAG run. Then the target group will automatically route requests to the same instance without the task needing to maintain a reference to the instance or us ever needing to even list the individual instances. That's definitely out of scope here, but wanted to share an example of what ASG's get us.

To summarise, rather than directly interacting with EC2, we can interact with the ASG. I don't think we can use AWS operators for this, there don't appear to be ones for ASGs, but boto3 supports them, so we just need to write our own interactions (like we do for s3). This eliminates the need for the Ansible playbook, like I said, so it will meaningfully reduce the amount of work done on the infrastructure side, in addition to reducing the number of mapped tasks in the DAG.

sarayourfriend (commented):
This is soooo exciting! I left a very long comment with a suggestion for an important difference we should ideally take to simplify deployments and maintenance, as well as potentially simplify the DAGs (depending on whether boto3 in a plain Python operator is considered "simpler" than the AWS Airflow operators).

I also took the time to write out the differences in hourly cost for EC2 and ECS, which you can include in the document as an appendix, if you'd like to record evidence for the cost difference. Based on the resource consumption numbers you mentioned, I also looked to see if we could reduce the allocated resources in either approach and compare the costs. EC2 would not work for us here, because anything with less vCPU cuts the RAM in half, with the relevant EC2 instances, and there's nothing smaller than the m5.xlarge that would keep enough RAM to be comfortably above the 52% of 16 GB consumption we saw.

ECS is more flexible, and we can scale down the amount of RAM allocated to each task, but because we still need at least 50% of

> **EC2 m5.xlarge**
> 0.192/hour
> 4 vCPU
> 16 GiB

ECS hourly costs:
- 0.04048/hour/vCPU
- 0.004445/hour/GB

> **ECS direct comparison:**
> 4 vCPU = 0.16192/hour
> 16 GB = 0.07112/hour
> total = 0.23304/hour

> **ECS lower RAM:**
> 4 vCPU = 0.16192/hour
> 10 GB = 0.04445/hour
> total = 0.20637/hour

> **ECS lower RAM and CPU:**
> 2 vCPU = 0.08096/hour
> 10 GB = 0.04445/hour
> total = 0.12541/hour

So there is _potentially_ a way to make ECS less expensive, simply by scaling down the resources we allocate for each task to try to achieve as high utilisation as possible from each one. However, we have the opportunity for much more significant cost savings if we re-write the indexer worker in a language that has tighter memory requirements than Python, by leveraging disk instead of RAM for task storage on the worker, or potentially even just by refactoring the existing Python code to be more consciencious of memory usage. If we can cut memory reliably below 50% consumption, a m4.large with 2 vCPU and 8 GB costs a mere  0.10/hour, cutting costs nearly in half from the m5.xlarge.

Even if ECS could be less expensive now, optimising the indexer workers to maximise resource utilisation with EC2 will always result in cheaper running costs than ECS, vCPU-per-vCPU and GB-per-GB. Planning this project with EC2 sets us up well for the opportunity of long-term savings through code optimisation. With the ASG approach I've outlined, I think it also brings virtually every single benefit we've come to enjoy from ECS to the EC2 based workers.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
sarayourfriend: Oh also, I forgot about this difference with ECS. We already have a process for automatically updating the ECS task definition when we cut a new release. We do not have that for launch templates, which are roughly the equivalent for EC2 instances (and the ASG approach). But, as you've said in the plan, we do not often update the code for the workers, so that isn't a big concern. Additionally, it would be good to keep updates to the launch templates manual (e.g., running `tf apply`) so that we make sure that we only do it when a data refresh is _not_ running.

stacimc: @sarayourfriend this is incredible 😍 I'm going to take some time to do a bit more background reading on this myself, but this is really exciting.

One possible correction is that while the indexer workers currently shut themselves down, in my proposed implementation the _DAG_ would send the `stop_instances` request. (However as you've noted that's only in the context of a data refresh; they need to shut themselves down after deploying. I also had not realized that we're currently using a cron job to restart the stack when the instances start 😮)

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+
+In this step we will add tasks to the data refresh DAGs to orchestrate the
+distributed reindex. At the end of this step, it will be possible to run a
+distributed reindex locally. Because the infrastructure work to create the EC2
+instances is not complete, it can not be run on production yet. The following
+code can all be refactored from
+[`distributed_reindex_scheduler.py`](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/distributed_reindex_scheduler.py).
+
+- Use Airflow's
+  [EC2Hook#describe_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.describe_instances)
```
AetherUnbound, line 395: This is a total bikeshed-y nit and feel free to ignore, but I prefer to use the [scope resolution operator](https://learn.microsoft.com/en-us/cpp/cpp/scope-resolution-operator?view=msvc-170) from C++ to denote member methods in classes (rather than `#` since that has a different meaning in Python). We also use it already in a few places: [1](https://github.com/WordPress/openverse/blob/ed267d80e8a67ef31d82f3857dce239fba45ae57/catalog/dags/common/sensors/utils.py#L36-L37), [2](https://github.com/WordPress/openverse/blob/ed267d80e8a67ef31d82f3857dce239fba45ae57/catalog/dags/data_refresh/create_filtered_index_dag.py#L99), [3](https://github.com/WordPress/openverse/blob/ed267d80e8a67ef31d82f3857dce239fba45ae57/catalog/dags/data_refresh/create_filtered_index_dag.py#L109). 

```suggestion
  [EC2Hook::describe_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.describe_instances)
```

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
AetherUnbound: +1-ing Staci's note - thank you for this @sarayourfriend! I hadn't thought about leveraging ASGs here for this, but it feels so _right_. Definitely on board with this approach!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+  by first calculating `start` and `end` indices that will split the records in
+  the media table into even portions, depending on the number of workers
+  available in the given environment. Then:
+  - POST to each worker's `reindexing_task` endpoint the `start_index` and
+    `end_index` it should handle
+  - Use a Sensor to ping the worker's `task/{task_id}` endpoint until the task
+    is complete, logging the progress as it goes
+  - Use the
+    [EC2Hook#stop_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.stop_instances)
+    to shut the instance down. This step should skip in a local environment.
```
AetherUnbound, line 415: It might be worth noting here that we'll want this to happen regardless of upstream success or failures, so we're not keeping resources alive unnecessarily (the `none_failed` trigger rule, I believe).

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+promotion) in staging. Once the data refresh has been moved to Airflow, we can
+add a
+[DAG param](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html)
+to the data refresh that allows skipping the initial `Copy Data` steps. This DAG
+will no longer be necessary and can then be deleted.
+
+### Update the load_sample_data scripts
+
+Our `load_sample_data.sh` scripts currently use `just` commands to ]run parts of
+the data
```
AetherUnbound, line 507:
```suggestion
Our `load_sample_data.sh` scripts currently use `just` commands to [run parts of
the data
```

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+- [Catalog Data Cleaning project](https://github.com/WordPress/openverse/issues/430),
+  which removes the Clean Data steps from the data refresh
+
+## Alternatives
+
+<!-- Describe any alternatives considered and why they were not chosen or recommended. -->
+
+The alternative options of using an ECS approach or performing the reindex
+entirely in Airflow are discussed at length in the [Approach to the Distributed
+Reindex][#approach-to-the-distributed-reindex] section.
```
AetherUnbound, line 550:
```suggestion
entirely in Airflow are discussed at length in the [Approach to the Distributed
Reindex](#approach-to-the-distributed-reindex) section.
```

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+There is minimal risk, as we will be able to test against production data in
+staging before running against production. However the data refresh does
+completely replace the production media tables and Elasticsearch indices, so
+there is always inherent risk that production data could be lost or malformed if
+something goes wrong. We will ensure production backups are available and
+monitor the first production data refreshes closely.
+
+## Prior art
+
+<!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
```
AetherUnbound, line 583: Should we remove this section? Or reference the search relevancy sandbox project here?

AetherUnbound (commented):
This is an _incredibly_ thorough implementation plan, thank you for taking the time to get this all "on paper"! I have a few questions and notes for clarification, and I'm really excited about Sara's suggestion for how to manage the EC2 instances in a way that will be more hands-off 🚀

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
sarayourfriend: Oh, thanks for the clarification there, Staci.

> they need to shut themselves down after deploying

That would no longer be necessary using the ASG, just to clarify :grin:

sarayourfriend: Wanted to clarify this as well:

> so that we make sure that we only do it [launch template updates] when a data refresh is not running

Updating the launch template does not necessarily automatically trigger a replacement of running instances that are using the old launch template. You _can_ configure ASGs to do that, but I'm pretty sure the default configuration does not automatically deploy the latest version of a launch template. The only potential issue is if for some reason one of the instances crashes (stops responding on healthchecks) and gets recreated, the ASG would use the latest launch template to do that, and if there's any potential for different versions of the indexer worker running at once to cause issues with completing the refresh, it'd be good to avoid that.

If that won't cause issues, then never mind, doesn't really matter, we can update the launch templates whenever we want.

zackkrida: @sarayourfriend perhaps to your point above, during a data refresh we could set the ASG to use a fixed version of the template rather than `latest`, and then switch the config back to `latest` after the refresh is complete. That should be possible with boto3.

stacimc:
Drafting to address comments :)

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
stacimc: > during a data refresh we could set the ASG to use a fixed version of the template rather than latest, and then switch the config back to latest after the refresh is complete.

Pinning the versions in Airflow is a really interesting idea 🤔 I think we could do something like this to pin/unpin the launch template version, but if the user_data script is always pulling the `latest` tag on the Docker image then code changes will be able to come through (for new instances) regardless, right? Although maybe those changes are okay to make mid-refresh.

One thing to note is that if an indexer worker fails for any reason, the Airflow `wait_for_reindexing` task will fail and won't automatically retry with a new worker instance. If the worker failed due to some issue with the reindexing, Airflow will terminate the instance to ensure it’s not left running (and prevent the ASG from making a new one). If the instance itself _crashes_ for some reason and the ASG tears it down, then the Airflow `wait_for_reindexing` task will still fail because it will try to poll a terminated instance. But the ASG will spin up a new one and the Airflow task will not know how to connect to resume, or to terminate the new instance gracefully. I think we’d need an extra Airflow task at the end to account for this 😓 But in either case, if a worker fails the DAG won’t be able to retry on a new instance automatically.

This discussion has made me realize that if we use an ASG instead of the more naive EC2 approach, we lose the ability to retry a single failed reindexer worker, because when a worker is terminated and a new one is spun up, there is no way to clear a single `trigger_reindex` task (because it will be trying to connect to the terminated instance). The best thing I have come up with is making `distributed_reindex` a reusable TaskGroup that accepts a list of worker params, and then having an entirely separate DAG that accepts worker params as a conf option and can be used to manually retry parts of a failed data refresh. This feels very convoluted.

stacimc: Although as we pointed out recently, if a single reindexing task fails and we're forced to retry _all of them_ (in this case by clearing the DAG at the task that sets the ASG desired capacity), it's not any _slower_ than retrying a single indexer worker because they run concurrently. The loss there is in the literal cost of unnecessarily running the other workers for however many hours the reindexing takes.

stacimc: Sorry to post back-to-back 😅 but -- the inability to retry a single reindexing task has an advantage in that it actually fixes the problem of ensuring that all the indexer workers are running the exact same code and have the exact same configuration. The trade-off is just the cost of running the successful workers in scenarios where the configuration _hadn't_ changed, which is perhaps fair.

This would allow us to use `latest` for both the launch template version and docker tag.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+The critical data refresh process is orchestrated by Airflow via the
+`<media_type>_data_refresh` DAGs, but the bulk of the work is merely triggered
+by Airflow on a remote ingestion server. Due to the reasons laid out in the
+Project Proposal, we will be moving all operational pieces of the data refresh
+into Airflow itself, and removing the ingestion server entirely.
+
+The work which must be moved into Airflow can be split into these abstract
+steps, which will be discussed separately in this IP:
+
+- **Copy Data**: A FDW extension is used to connect the API database to the
```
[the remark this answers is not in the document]
stacimc: Sure!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+  upstream (catalog) database. The entire contents of the upstream media table
+  are copied into a new temp table in the API database. This temp table will
+  later replace the main media table in the API.
+- **Clean Data**: Data clean up which includes removing denylisted tags and
+  cleaning URLs. This step will become unnecessary when the
+  [Catalog Data Cleaning project](https://github.com/WordPress/openverse/issues/430)
+  is completed, which moves all data cleaning steps into ingestion.
```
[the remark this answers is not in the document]
stacimc: _Much_ improved, thanks

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+  Postgres (e.g.
+  [`es_helpers.py`](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/es_helpers.py))
+
+The
+[`Dockerfile`](https://github.com/WordPress/openverse/blob/main/ingestion_server/Dockerfile)
+can be copied from the existing ingestion server. It should be updated to
+reference the new file structure, and to expose only a single port, which should
+be distinguished from the ports currently in use by the ingestion server (8001
+and 8002). Other necessary files, including `env.docker`, `.dockerignore`,
+`Pipfile`, and `gunicorn.conf.py` can all be copied in from the existing
```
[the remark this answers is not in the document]
stacimc: Thanks for this, added a note to the dependencies section as well!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+ingestion server as well.
+
+Finally we will update the monorepo's root
+[`docker-compose.yml`](https://github.com/WordPress/openverse/blob/05ff48d05f2163104151c5589cf352a156bc6a97/docker-compose.yml)
+to add a new `catalog-indexer-worker` service. Its build context should point to
+the nested `data_refresh/indexer_worker` directory, and it should map the
+exposed port to enable the API to be reached by the catalog.
+
+When this work is complete, it should be possible to run `just catalog/ipython`
```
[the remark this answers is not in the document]
stacimc: Yep, mixed my suggestions :)

stacimc:
Updates are up and I've re-requested reviews. Particularly interested in folks' thoughts on the discussion toward the end of [this thread](https://github.com/WordPress/openverse/pull/4026#discussion_r1550825307) and the conclusions I came to here :)

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -196,40 +213,38 @@ into the status of the task from Airflow.
...
-  to be a requirement of the initial implementation.
+This is a great strength of both approaches, each of which would eliminate the
+need for deployments when changes are made to the data refresh code (including
+changes to the indexer workers).
+
+Each time an ECS task is spun up on Fargate, it will pull the Docker image
+configured in the task definition. By using the `latest` tag in this
+configuration we can ensure that the latest indexer-worker image is pulled each
+time, which means that any changes to the reindexing code will become live in
+production as soon as the new docker image is published after merging on `main`.
```
AetherUnbound, line 224: Nit for consistency (`Docker` is used below):
```suggestion
production as soon as the new Docker image is published after merging on `main`.
```

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -410,52 +429,44 @@ code can all be refactored from
...
     is complete, logging the progress as it goes
-  - Use the
-    [EC2Hook#stop_instances](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/ec2/index.html#airflow.providers.amazon.aws.hooks.ec2.EC2Hook.stop_instances)
-    to shut the instance down. This step should skip in a local environment.
+- Use
+  [`terminate_instance_in_auto_scaling_group`](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/terminate_instance_in_auto_scaling_group.html)
+  to terminate the instance. Make sure to set `ShouldDecrementDesiredCapacity`
+  to `True` to ensure that the ASG does not try to replace the instance. This
+  task should use the
+  [`NONE_FAILED` TriggerRule](https://airflow.apache.org/docs/apache-airflow/1.10.9/concepts.html#trigger-rules)
```
AetherUnbound, line 437: Two notes here:
1. I think we want `NONE_SKIPPED` instead? `NONE_FAILED` would mean those tasks _wouldn't_ run on failures, which I think we want for the step.
2. That link appears to be for Airflow v1.10 😅 

```suggestion
  [`NONE_SKIPPED` TriggerRule](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules)
```

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,610 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @sarayourfriend
+- [ ] @AetherUnbound
```
AetherUnbound, line 13:
```suggestion
- [X] @AetherUnbound
```

AetherUnbound (approved):
Thanks for the revisions, I think all my concerns are assuaged! I'm so stoked for this 🤩

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
sarayourfriend: We should generate AMIs for the indexer workers with the Docker image they should run bundled into them: https://github.com/WordPress/openverse-infrastructure/issues/843#issuecomment-2058094706

The launch template would always point to the latest AMI, and a new AMI would only be generated for production manually (we can also make it manual for staging if we want). We can configure the ASG to replace instances using the launch template they were running before, rather than using the latest launch template version (or pin the ASG to a specific launch template version rather than `$Latest`, and manually bump it when data refresh isn't running for that environment if a new AMI exists, maybe even just using a DAG that runs and checks if data refresh is running before bumping the launch template to the new AMI version).

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,586 @@
...
+The ECS approach allows us to remove 100% of the server management code, and
+eliminates the need for deployments in essentially every case (unless the task
+definition itself changes). However it is more expensive, more difficult to
+implement, and requires a special secondary implementation for running locally
+that may be prone to cross-platform issues.
+
+The EC2 approach on the other hand is cheaper, quicker to implement, and gets us
+the vast majority of what we want in terms of removing complexity and reducing
+the need for deployments. Consequently I argue here that the EC2 approach is the
+one we should pursue.
```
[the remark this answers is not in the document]
sarayourfriend: I think the inability to realistically test the data refresh locally, needing an _entirely separate process_ locally, is pretty significant.

> seemingly-unnecessary webserver aspects

In what sense are they unnecessary? ECS still runs a server, it eliminates a small amount of deployment complexity, but it isn't complexity we don't have elsewhere and can solve to make as or almost as simple as our ECS deployments (which would be good for making Airflow, Kibana, Elasticsearch, etc all simpler and easier to understand and maintain).

The difference in long-term cost is significant, with a bit of up-front infrastructure work that will build off of and benefit the work we've done for other services. Why avoid EC2 if we can and regardless _need_ to make it simple to use for the sake of our other services? We have a low level of understanding of EC2, but we need to improve that, and as we have done, everything about the process improves and eliminates the relative complexity of it compared to ECS.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,610 @@
...
+
+Some important notes:
+
+- Currently, the staging and production workers are split into separate
+  environments (i.e., a staging deploy is used to deploy the 2 staging workers
+  separately from the 6 production workers). It is more accurate to view all 8
+  workers as production instances (i.e., part of the production catalog
+  deployment), which merely _operate_ on different environments. As such all 8
+  should be part of the production deployment, but two separate ASGs which are
+  tagged indicating their intended environment.
```
sarayourfriend, line 465: I don't really follow this logic of putting both in the production root module.

The staging workers should use staging code, staging secrets, and reference other staging resources. The production workers should do that for production.

Why break the pattern for Airflow? I think the distinction between what is a staging and what is a production resource is bit confused here. Airflow is somewhat external to our deployment environment concept, but insofar as the ingestion workers should use code versioned to match e.g., the API's index mapping expectations, shouldn't they align with the application-level they refer to?

Following this logic would kind of make both Elasticsearch instances go into the production root module. Airflow is kind of like Kibana, which also references a staging resource, but exists in the production root module.

Airflow and Kibana are external to concepts of staged deployments, and cross things over in ways that make it no longer make sense to speak of whether they are "staging" or "production". It could be helpful to get rid of the idea of a deployment "stage" name for services like this that don't have staged deployments.

The indexer workers _do_ have staged deployments, as far as I know, and if they don't then we don't need two separate ASGs, we can just use a single ASG with a max capacity capable of handling both staging and production index at the same time, rather than having separate ASGs :shrug:

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,610 @@
...
+- The playbooks must be updated to check if any of the four _new_ data refresh
+  DAGs are running before deploying, as well.
+- The `user_data` script should be updated to pull the Docker image with the
+  `latest` tag.
```
sarayourfriend, line 469: Just a nit to note that this should change if we use the AMI approach, which I think we should, and I aim to have resolved so as not to block the infrastructure work of this IP. It's a conceptual/foundational piece of our ASG-based deployment puzzle that needs to be solved one way or another. It doesn't _need_ to block this project, so you don't need to change from this described approach here using docker to pull the latest image, and the AMI approach builds on that. But just in case we manage to sort it out before we get to this point of the plan, it's worth noting the difference.

sarayourfriend (approved):
This LGTM.

I've left a couple comments describing how the AMI approach I talk about [in this comment in the infrastructure repository](https://github.com/WordPress/openverse-infrastructure/issues/843#issuecomment-2058094706) would solve some of the edge cases in this PR introduced by using the "always pull latest" docker image approach. None of that blocks this plan though, we can move forward with it as is, and the AMI approach would supplant it when eventually implemented. Of course, it could be worth noting the AMI approach at least in general, so that if it is available before the implementation actually reaches the step of creating the ASG, that it should be used instead.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -0,0 +1,610 @@
...
+  [`set_desired_capacity`](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/set_desired_capacity.html)
+  to increase the desired capacity of the ASG to the desired number of workers,
+  depending on the environment. This will cause the ASG to begin spinning up
+  instances. (Skips in local env.)
+- Use
+  [`describe_auto_scaling_groups`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/describe_auto_scaling_groups.html)
+  to poll the ASG until all instances have been started, and get the EC2
+  instance IDs. (Skips in local env.)
+- Use dynamic task mapping to generate a Sensor for each of the instance IDs,
+  and ping its `healthcheck` endpoint until it passes.
```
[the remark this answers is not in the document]
stacimc: Good call. I was thinking about pinging the healthcheck for the api, rather than the instance itself -- but I had not considered that it's possible to define custom health checks, which looks like it should be possible. That would make this possible.

I was thinking about the local healthcheck endpoint and I think it's okay to not have that, since locally we'll have the container running when the stack comes up (rather than starting/stopping it to try to simulate the asg or anything like that).

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
stacimc: The thought behind this is that there is not intended to ever be any difference between the staging vs prod indexer workers, except the connection variables; to test updates we don't deploy staging code for the indexer workers, we run the data refresh against the staging environment. (This is basically also true today: we have not historically made use of the staging ingestion server, so the addition of staging_data_refresh DAGs in this project is actually a huge improvement!)

The advantage of having two ASGs is just that it makes it very simple to allow staging and production data refreshes to run simultaneously, without worrying about them interfering with each other. For example, one of the edge cases I raised in the IP is if an instance crashes during reindexing and the ASG replaces it with a new instance; we add a cleanup task that sets the desired capacity to 0 after reindexing is complete to ensure there cannot possibly be any leftover instances. But we cannot do this if the ASG might also be supporting a staging data refresh simultaneously.

Since we don't expect to have to deploy these often I don't think it's unreasonable to require them to be deployed separately, but it also doesn't seem advantageous to me. I wouldn't be strongly opposed to separating the deployments out if that's preferred, though.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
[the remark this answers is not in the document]
AetherUnbound: > seemingly-unnecessary webserver aspects

The indexer workers run their own webserver in order to accept tasks and issue progress on them. That code could be completely obviated if the containers which were to run on ECS received the range they needed to index from Airflow and then ran directly on that chunk. The containers could spin up, use the range supplied via CLI to run, and then complete and exit. This would remove the need for maintaining _any_ webserver aspects for the indexer workers, since the containers would only be running as long as it takes to index everything.

To be clear, I'm comfortable with the current approach proposed here nonetheless.

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
sarayourfriend: Gotcha, that's a good way of thinking about it. I wonder if we can think of them as "pools" rather than staging/production. In that sense, I think it's good to keep "environment" as defined in the terraform code to production for both, and indicate the purpose in the service name, "staging-indexer-worker-pool", "production-indexer-worker-pool" (or whatever you think makes sense, just indicating the purpose there rather than by changing the environment label). I realise now that could be what you meant by "tagged indicating their environment" but I interpreted it as changing the "environment" tag, which is a semantically important tag in Ansible and in how our infra resources are organised in relation to other supporting resources.

It might seem like just a nit-pick, but the production environment label relates to things like the load balancer selected, and I'd be massively confused if I saw infra stuff with environment labelled "staging" attached to the production internal load balancer (in a hypothetical future where we used that approach with the sticky cookies I mentioned, or just to make it easier to reference the instances overall).

sarayourfriend: Might need to take a step back here and defer the decision either way. It'll depend on what is easiest to manage within our Ansible playbooks, and whether we'd use separate AMIs for staging and production indexer workers or if the target environment is a runtime variable :thinking:

stacimc: >  I realise now that could be what you meant by "tagged indicating their environment" but I interpreted it as changing the "environment" tag, which is a semantically important tag in Ansible and in how our infra resources are organised in relation to other supporting resources.

That is what I meant, but I didn't realize I was overloading terminology -- thank you for pointing that out, I will absolutely clarify that.

> It might seem like just a nit-pick

Not at all, it's an important distinction and I was also worried that it might be confusing. I really love the '-pool' suggestion for the tags, that's a small difference that makes it much clearer!

Issue #3945, opened 2024-03-20
Log when source query parameter contains invalid values


obulat:
## Problem

In the [API `source` parameter validation improvements](https://github.com/WordPress/openverse/discussions/3895) discussion we decided to log when any of the sources in the API request parameters is invalid.

## Description

Log whenever one of the sources in the search request is invalid. This is the code that needs to be updated:

https://github.com/WordPress/openverse/blob/fc38553d04cde7586ce680543757d77a15be9a41/api/api/serializers/media_serializers.py#L655-L656

We can add the following code:
```py
sources = value.lower().split(",")
original_len = len(sources)
valid_sources = [source for source in sources if source in allowed_sources]
if original_len > len(sources):
    invalid_sources = set(sources).difference(set(valid_sources))
    logger.warn(f"Invalid sources in search query: {invalid_sources}")
```

## Alternatives

We could also save the information about requests and invalid sources in Redis if that would be easier to query.

manojks1999:
assign

PR #4031, opened 2024-04-04
Add warning to search response when source parameter has mixed validity

### LINKED ISSUE #4030: Improve invalid source parameter handling
## Problem

See https://github.com/WordPress/openverse/discussions/3895#discussioncomment-8995192 for background of this discussion.

## Description

We'll make the following changes:
- Remove the warning logging when invalid sources are encountered.
- Add a `warnings` list to the search response when a partially invalid source list is passed.
- Raise a validation error when all sources passed are invalid, when not a single valid source is passed (e.g., all the sources are mispelled or disabled)

## Additional context

Marking high priority for the logging improvements, because we don't need to fill our logging with these messages now that we have the answer we wanted.

sarayourfriend:
## Fixes

Fixes #4030 by @sarayourfriend 
Closes https://github.com/WordPress/openverse/discussions/3895 by @obulat

## Description

This PR:
- Removes the logging we added in #3945 for investingating the potential solutions in the linked discussion
- Adds a "warnings" on the search response when the request has mixed validity in the source parameter
- Raises a validation error when none of the sources are valid. See the linked discussion for @obulat's reasoning for why this is a necessary and acceptable breakage. I've left a comment in the code explaining it as well.

I am happy with the PR in its current state, but I went back and forth about whether the new `warnings` key should default to an empty list or to not being present at all when there are no warnings. I went with the empty list because it's the simplest and most consistent. But I could easily argue that having it not be present at all is also totally reasonable. I could also see wanting to prefix the key to `_warnings` or nest it in some `meta` object on the response. I'm open to any suggestions on this.

I originally started with a more complex idea for this PR, to have a middleware that would add warnings to the responses based on a list of warnings set onto the request context. That would make it so any endpoint could (theoretically) easily set a warning on any request, so long as it had some way of accessing the request. This was inspired by Django's messages utility used to flash warnings in rendered HTML pages. DRF does not have an equivalent, and existing libraries for it do something totally different then what I wanted. All of that is way more complex than we need for this specific issue, though, so I chose to go for a more direct approach. If we find other use cases for the warnings, we can evaluate whether a more generic solution is appropriate.

## Testing Instructions

Evaluate the changes and confirm the tests sufficiently cover the new cases. I've gone with additional integration tests rather than testing at the serializer level because there are three places that need to work together (i.e., to be "integrated") for this to work, so unit testing just the serializer would be insufficient, and would duplicate any meaningful testing at the integration level.

Run the application locally using `just api/up` and visit the search endpoint. Evaluate the following scenarios:
- No parameters: empty warnings list
- Only valid parameters: empty warnings list
- Mixed validity of the source parameter: the new warning
- Only bad parameters: 400 response

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -1,17 +1,21 @@
 import logging
 from collections import namedtuple
+from typing import TypedDict
 
 from django.conf import settings
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.core.validators import MaxValueValidator
+from django.urls import reverse
 from rest_framework import serializers
 from rest_framework.exceptions import NotAuthenticated, ValidationError
+from rest_framework.request import Request
 
 from drf_spectacular.utils import extend_schema_serializer
 from elasticsearch_dsl.response import Hit
 
 from api.constants import sensitivity
 from api.constants.licenses import LICENSE_GROUPS
+from api.constants.media_types import MediaType
 from api.constants.parameters import COLLECTION, TAG
 from api.constants.sorting import DESCENDING, RELEVANCE, SORT_DIRECTIONS, SORT_FIELDS
 from api.controllers import search_controller
@@ -294,8 +298,16 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
         required=False,
     )
 
+    class Context(TypedDict, total=True):
+        warnings: list[dict]
+        media_type: MediaType
+        request: Request
+
+    context: Context
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.context["warnings"] = []
         self.media_type = self.context.get("media_type")
         if not self.media_type:
             raise ValueError(
@@ -398,15 +410,31 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
                 )
             return value
         else:
-            sources = value.lower().split(",")
-            valid_sources = set(
-                [source for source in sources if source in allowed_sources]
-            )
-            if len(sources) > len(valid_sources):
-                invalid_sources = set(sources).difference(valid_sources)
-                logger.warning(
-                    f"Invalid sources in search query: {invalid_sources}; sources query: '{value}'"
+            sources = set(value.lower().split(","))
+            valid_sources = {source for source in sources if source in allowed_sources}
+            if not valid_sources:
+                # Raise only if there are _no_ valid sources selected
+                # If the requester passed only `mispelled_museum_name1,mispelled_musesum_name2`
+                # the request cannot move forward, as all the top responses will likely be from Flickr
+                # which provides radically different responses than most other providers.
+                # If even one source is valid, it won't be a problem, in which case we'll issue a warning
+                raise serializers.ValidationError(
+                    f"Invalid source parameter '{value}'. No valid sources selected. "
+                    f"Refer to the source list for valid options: {sources_list}."
                 )
+            elif invalid_sources := (sources - valid_sources):
+                self.context["warnings"].append(
+                    {
+                        "code": "partially invalid source parameter",
+                        "message": "The source parameter was partially invalid.",
+                        "invalid_sources": invalid_sources,
+                        "referenced_sources": valid_sources,
+                        "available_sources": self.context["request"].build_absolute_uri(
+                            reverse(f"{self.media_type}-stats")
+                        ),
+                    }
+                )
+
             return ",".join(valid_sources)
 
     def validate_excluded_source(self, input_sources):
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -423,15 +423,18 @@ class MediaSearchRequestSerializer(PaginatedRequestSerializer):
                     f"Refer to the source list for valid options: {sources_list}."
                 )
             elif invalid_sources := (sources - valid_sources):
+                available_sources_uri = self.context["request"].build_absolute_uri(
+                    reverse(f"{self.media_type}-stats")
+                )
                 self.context["warnings"].append(
                     {
                         "code": "partially invalid source parameter",
-                        "message": "The source parameter was partially invalid.",
-                        "invalid_sources": invalid_sources,
-                        "referenced_sources": valid_sources,
-                        "available_sources": self.context["request"].build_absolute_uri(
-                            reverse(f"{self.media_type}-stats")
+                        "message": (
+                            "The source parameter included non-existent sources. "
+                            f"For a list of available sources, see {available_sources_uri}"
                         ),
+                        "invalid_sources": invalid_sources,
+                        "valid_sources": valid_sources,
                     }
                 )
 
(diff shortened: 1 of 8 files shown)


api/api/serializers/media_serializers.py
```diff
@@ -398,15 +410,31 @@ def validate_source(self, value):
...
+                    f"Invalid source parameter '{value}'. No valid sources selected. "
+                    f"Refer to the source list for valid options: {sources_list}."
+                )
+            elif invalid_sources := (sources - valid_sources):
+                self.context["warnings"].append(
+                    {
+                        "code": "partially invalid source parameter",
+                        "message": "The source parameter was partially invalid.",
+                        "invalid_sources": invalid_sources,
+                        "referenced_sources": valid_sources,
```
dhruvkb, line 431: The name "valid" feels clearer to me than "referenced".

```suggestion
                        "valid_sources": valid_sources,
```

api/test/integration/test_media_integration.py
```diff
@@ -329,6 +329,57 @@ def test_detail_view_for_invalid_uuids_returns_not_found(
     assert res.status_code == 404
 
 
+def test_search_with_only_valid_sources_produces_no_warning(media_type, api_client):
+    stats = api_client.get(f"/v1/{media_type.path}/stats/")
+    assert stats.status_code == 200
+
+    search = api_client.get(
+        f"/v1/{media_type.path}/",
+        {"source": ",".join([s["source_name"] for s in stats.json()])},
```
dhruvkb, line 338: We have some hardcoded provider names in the `media_type` fixture. By not making another request we can make the tests independent of the `/stats` endpoint.

api/test/integration/test_media_integration.py
```diff
@@ -329,6 +329,57 @@ def test_detail_view_for_invalid_uuids_returns_not_found(
     assert res.status_code == 404
 
 
+def test_search_with_only_valid_sources_produces_no_warning(media_type, api_client):
```
dhruvkb, line 332: Nit: This is specialised case of the next test function which could be parametrized (in terms of invalid sources being an empty or populated set) to include this scenario.

dhruvkb (approved):
LGTM! I have strong feelings about [this one comment](https://github.com/WordPress/openverse/pull/4031#discussion_r1550890713) (the 2nd one below) and would definitely like to see that fixed but it felt wrong to block a PR that works over a change in the tests.

api/test/integration/test_media_integration.py
sarayourfriend: I thought of that too, but I liked having separate and descriptive names for them, rather than doing an if/else in the test to do the assertions for each one.

api/api/serializers/media_serializers.py
sarayourfriend: I found juggling the list of valid and available sources confusing. Usually when I see an error or warning about invalid values, the _valid_ values listed are all the possible valid values, if that makes sense? I'm torn, so I'll wait and see what the other reviewer says, and change it if they want it changed as well, if that's okay.

api/test/integration/test_media_integration.py
sarayourfriend: Done: https://github.com/WordPress/openverse/pull/4031/commits/caddd9eab6bce3dd81d8155872d62ee28e881163

Agree we should refactor that. Ideally we would unify those media type configurations with the ones that pre-existed them used in other tests too, rather than having two separate ones with slightly different names for things.

api/api/serializers/media_serializers.py
dhruvkb: "valid", "invalid" and "available"/"all" potentially. But yeah, let's allow one more review to see what they think.

stacimc: `referenced` did not feel clear to me either, but I see what Sara's saying as well. 

Catching up on the linked discussions, Dhruv mentioned `it seems like that is the most common way of handling cases where the input is partially acceptable and can generate a valid response`. Are you referencing something in particular you can link to? I'm not familiar with this type of response.

Along those lines, is the shape of this response following some established convention, or could it be changed? For example, do we need to explicitly list which of the provided sources were valid at all, or can we just have `invalid_sources` and `available_sources`? Or could this information all be spelled out in the "message" instead of in separate named fields?

sarayourfriend: > Or could this information all be spelled out in the "message" instead of in separate named fields?

I actually originally had the warnings just be a list of strings, but I found it hard to do a meaningful test without just duplicating the string almost word-for-word in the test case :cold_sweat: On top of that, because of using sets instead of lists, the order of sources in the strings was non-deterministic, making it even hard to test against a simple string.

We should use whatever format here we want. I've included in the documentation for the response field that it is meant to be human readable rather than read by a machine, and that the contents of each dict are not stable.

Maybe `discarded_sources`, `kept_sources`, and `available_sources`? :shrug: whatever folks want here, happy to change it, I am not attached to any specific language, even if I found something or other personally confusing. I think it will get the idea across that something isn't right about the parameter on the request and that the developer needs to take a closer look at it.

Which also makes me wonder whether the warnings should go _first_ in the JSON, rather than at the end? On a page of 20 results, I don't know whether it's easier to miss at the front or end of the document.

dhruvkb: > Are you referencing something in particular you can link to?

I didn't keep a record of my search when looking for a good pattern but I went through my browser history and found these references.

- https://www.mediawiki.org/wiki/API:Errors_and_warnings
- https://discuss.jsonapi.org/t/multi-status-responses-partial-success-in-particular/30/3

To be clear, this is not an established convention. It's the simplest, backwards compatible way I could think of to stick with a 200 OK status code but also convey problems in their input to the user.

stacimc: Gotcha! And thanks for the links -- I wanted to make sure I wasn't suggesting deviating from some widely accepted pattern :)

> I actually originally had the warnings just be a list of strings, but I found it hard to do a meaningful test without just duplicating the string almost word-for-word in the test case 😰 On top of that, because of using sets instead of lists, the order of sources in the strings was non-deterministic, making it even hard to test against a simple string.

Dang, that makes sense. One final suggestion -- what if we moved just the link to the available sources into the `message`? So the `warning` could be something like:

```
        {
            "code": "partially invalid source parameter",
            "message": "The source parameter was partially invalid. For a list of available sources, see http://localhost:50280/v1/images/stats",
            "invalid_sources": [
                "foo"
            ],
            "valid_sources": [
                "flickr"
            ]
        }
 ```

I think that would fix the problem with testing but make it a little clearer.

> Which also makes me wonder whether the warnings should go first in the JSON, rather than at the end? On a page of 20 results, I don't know whether it's easier to miss at the front or end of the document.

+1 for putting it first in the JSON, now you mention it.

stacimc (approved):
This all tests well for me, approved! I did +1 to your suggestion to move the warnings up in the JSON and had one more suggestion for the names, but not a blocker -- up to you :)

sarayourfriend:
I've moved the warning to the top, made it only appear when there is actually a warning (otherwise it's kind of ominous looking on requests that don't have issues, plus it's a few bytes over the wire that we can save on those responses). I also updated the warning dict to match Staci's suggestion.

PR #4032, opened 2024-04-04
Remove potentially problematic `do_not_wait_for`

sarayourfriend:
## Fixes

Related to the recent API outage we had.

## Description

This is the main place in our code where we create asyncio tasks that could potentially be the tasks in question in the "Task was destroyed" errors related to the recent outage.

I don't know for sure that these are the problem, but the approach with `do_not_wait_for` is not necessary for our thumbnails endpoint to be sufficiently performant.

Even if this doesn't address whatever the problem was, I'd call this a code quality improvement by simplify the code and removing something potentially arcane and confusing that isn't strictly necessary.

## Testing Instructions

CI should pass. Everything changed here is tested in unit tests.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/asyncio.py
+++ b/api/api/utils/asyncio.py
@@ -1,36 +1,6 @@
-import asyncio
-import logging
-from collections.abc import Awaitable
-
 from rest_framework.generics import get_object_or_404
 
 from asgiref.sync import sync_to_async
-
-
-parent_logger = logging.getLogger(__name__)
-
-
-_do_not_wait_for_logger = parent_logger.getChild("do_not_wait_for")
-
-
-def do_not_wait_for(awaitable: Awaitable) -> None:
-    """
-    Consume an awaitable without waiting for it to finish.
-
-    This allows us to call an async function that we don't care about
-    the result of, without needing to wait for it to complete. This is
-    useful, for example, if some operation creates a side effect that
-    isn't necessary for the response we're processing (e.g., Redis tallying).
-    """
-    try:
-        loop = asyncio.get_running_loop()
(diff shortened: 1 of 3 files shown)


dhruvkb (approved):
Am I understanding correctly that by changing `do_not_wait_for` to `await`, we are losing that functionality that allows us to not care for the outcome of a async task?

This PR LGTM, and I'm all for deleting code that could be the cause of a problem but I want to know if we're trading performance for it.

sarayourfriend:
Yes, you're correct in that understanding. From a performance perspective, this probably accounts for milliseconds of difference in the response time of our longest route. By percentage, I don't think this could be a meaningful increase in the time it takes to process the request. It therefore also isn't necessarily a meaningful performance _improvement_ either (another good reason to remove it).

We are technically reducing the performance of the endpoint _strictly_ speaking, but there are better ways to defer tasks to "sometime in the near future" than what we've done here (like a message queue or maybe async Django signals) if we want to revisit this again (or if we _need_ to, if this PR turns out to cause it's own problems).

dhruvkb:
💯 For sure, I'm sure the performance implication of the change would be extremely small, I just wanted to confirm if I was understanding correctly. I still can't fully wrap my heads around the async parts of the API so thanks for clarifying.

AetherUnbound (approved):
LGTM! A good call regardless, as you mention 🙂

PR #4043, opened 2024-04-05
Turn on additional search views frontend

### LINKED ISSUE #4035: Switch the `additional_search_views` flag on in staging and prod
## Problem

The additional search views are ready to test in prod.

## Description

To enable this feature in prod, we should set the flag `switchable` and `on` by default. This would enable us to quickly check if any feature works differently with/without the additional search views in prod.

## Alternatives

We could make the flag enabled by default, but would then be unable to compare the two states quickly (we would need to redeploy for that)

obulat:
## Fixes

Fixes #4035 by @obulat

## Description

This PR makes the `additional_search_views` flag switchable in prod, and turns it on by default, effectively launching the additional search views.
The test setup and checks had to be updated because the single result pages have slightly changed.

## Testing Instructions

Build and run the app using `just frontend/run build` and `just frontend/run start`
Go to `https://localhost:8443/search?q=cat` and select a search result (image/audio). The new page with link buttons and tags as links should be displayed by default.
You should also be able to open the source/creator/tag pages, without setting the preferences previously.
If you go to `https://localhost:8443/preferences`, you can turn `additional_search_views` off to view the site without the additional views.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/feat/feature-flags.json
+++ b/frontend/feat/feature-flags.json
@@ -12,10 +12,10 @@
     "additional_search_views": {
       "status": {
         "staging": "switchable",
-        "production": "disabled"
+        "production": "switchable"
       },
       "description": "Toggle additional search views for tag/creator/source.",
-      "defaultState": "off",
+      "defaultState": "on",
       "storage": "cookie"
     },
     "additional_search_types": {
--- a/frontend/src/components/VMediaInfo/VMediaTags.vue
+++ b/frontend/src/components/VMediaInfo/VMediaTags.vue
@@ -58,6 +58,7 @@ import type { Tag } from "~/types/media"
 import type { SupportedMediaType } from "~/constants/media"
 import { useFeatureFlagStore } from "~/stores/feature-flag"
 import { useSearchStore } from "~/stores/search"
+import { useI18n } from "~/composables/use-i18n"
 
 import { focusElement } from "~/utils/focus-management"
 
@@ -87,7 +88,8 @@ export default defineComponent({
 
     const searchStore = useSearchStore()
     const featureFlagStore = useFeatureFlagStore()
-    const { app, $sendCustomEvent } = useContext()
+    const { $sendCustomEvent } = useContext()
+    const i18n = useI18n()
 
     const additionalSearchViews = computed(() =>
       featureFlagStore.isOn("additional_search_views")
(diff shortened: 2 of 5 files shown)


zackkrida (approved):
Very nice. LGTM!

frontend/feat/feature-flags.json
```diff
@@ -12,10 +12,10 @@
     "additional_search_views": {
       "status": {
         "staging": "switchable",
-        "production": "disabled"
+        "production": "switchable"
       },
       "description": "Toggle additional search views for tag/creator/source.",
-      "defaultState": "off",
```
sarayourfriend, line 18: I'm a little confused about what the "right" way to toggle the flag on in production is.

We set it to switchable, but default it to on. That means it's basically on in all environments, but an individual could switch it off by visiting `/preferences`, right? In practice that's irrelevant to everyone except devs because `/preferences` isn't linked anywhere.

So this feature flag setting could be simpified by: replacing the staging entry with a single production entry, set to `enabled`, and removing `defaultState` altogether. Thinking of this in stages, should we do this in two steps, first set production to switchable, default state off, and staging to enabled? And then confirm it in staging as a globally enabled preference (rather than a toggled one), and then update `production` to enabled? It would be nice if we could control these using environment variables so that it doesn't take an entire new release for each stage to take effect.

sarayourfriend (approved):
This LGTM, I've just left a suggestion for how to either simplify the feature flag state, or an alternative staged deployment of the globally enabled feature (i.e., enable it globally in staging first, confirm it, then globally in production).

This is all well tested in e2e tests so I'm not too worried about it, and the consequence of a small bug in this feature that we haven't caught is a non-issue, from my perspective, but it's something we should consider for future features.

dhruvkb (approved):

frontend/feat/feature-flags.json
zackkrida: In production it seems that "switchable" is essentially "enabled" but with a convenience method for devs to see the disabled state. 

Abstractly, I think the flow for launching frontend features should be:

- Enable the feature by default in staging, switchable but off by default in prod
- Test it in both of those contexts
- Enable it in prod

I think we could use a doc outlining how to prep and ship user-facing frontend (or API, for that matter) features. I'll create issues this week.

PR #4044, opened 2024-04-05
Add `SEARCH_RESPONSE_TIME` event

### LINKED ISSUE #2471: Add `SEARCH_RESPONSE_TIME` analytics event
## Problem

While we know how long our server takes to respond to a request, we don't know how long it takes for clients to receive the response. It would be nice if we could implement this in our analytics code by adding an event `SEARCH_RESPONSE_TIME` to run client-side only that sends the time it took for the response. It should only do so for queries that aren't cached in the local browser cache.

To help visualise the part we don't have information on, here is a diagram:

```
client -> Cloudflare -> ELB -> nginx -> django -> nginx -> ELB -> Cloudflare -> client
```

The only timing we have visibility into is `Cloudflare -> ... -> Cloudflare` and any self-to-self timing in between (nginx to nginx, ELB to ELB). We never know the `client -> ... -> client` timings with our current measurement tools. Client timings have two manifestations as well: Cloudflare cache hits and misses. Therefore, we also need to record whether the Cloudflare cache header records a hit or miss so we can disambiguate and see which part is taking the longest (is it Cloudflare's response or everything after Cloudflare?).

This is motivated by a discovery that many search requests in our frontend can take longer than 1 second to respond for me locally, but this isn't apparent from the metrics we have for the nodes after Cloudflare in the diagram above.

## Description

Refer to the [frontend analytics guide](https://docs.openverse.org/frontend/guides/analytics.html) for more information on how to use and set up analytics for development.

When making a search request, create a date object before sending the request. If the `Date` header on the response is after the request start date, then the response is not cached in the local client. This is because the `Date` header does not change if the response is retrieved from local cache (which we don't care about the timings for). It _does_ change, however, on any outbound request, even ones that hit the Cloudflare cache.

When the request comes back, if the `Date` header is after the time we started making the request, then get the elapsed time between the request start and when the response was received. Also pull out the Cloudflare cache status header `cf-cache-status`. Send the following analytics payload:

```ts
type Events = {
  // ...

  /**
   * Time client-side search responses. Gives us observability into
   * real user experience of search timings.
   *
   * Payload:
   *   - `cfCacheStatus`: Whether the request hit Cloudflare or went all the way to our servers
   *   - `cfRayIATA`: The IATA location identifier at the end of the `cf-ray` header. Indicates the data centre the request passed through. This gives us an idea of approximate distance from our API servers without revealing more precise request location information.
   *   - `elapsedSeconds`: How many seconds it took to receive a response for the request
   *   - `queryString`: The full query string including additional filters (i.e., not just the search term)
   */
  SEARCH_RESPONSE_TIMING: {
    cfCacheStatus: "HIT" | "MISS"
    cfRayIATA: string
    elapsedTime: number
    queryString: string
  }
}
```

## Alternatives

We could implement a bona fide [RUM](https://en.wikipedia.org/wiki/Real_user_monitoring) library that measures this stuff more generally, including client-side render timings, etc. But those can often cost a tremendous amount of money and I think we can get the information we need for this using Plausible. We might even be able to implement deeper analysis like client side render timings in Plausible as well to get RUM-level data about site performance, but we're a ways from being able to use that information anyway, I think.

### LINKED ISSUE #3775: `SEARCH_TIME_EVENT` is unusable due to plausible and CORS limitations
## Description

`vue-plausible` only provides client-side integration, and we cannot send events from the server:

https://github.com/moritzsternemann/vue-plausible/blob/56d05277e4082183c700e9f0df4e82b711a00f07/src/nuxt-module.ts#L22-L27

_Note: this has never been the problem before because we never recorded server-side events._

On the other hand, on the client side, due to CORS limitations, we cannot get the required headers (`["date", "cf-cache-status", "cf-ray"]`). Browser limits JS access to only `"cache-control", "content-type" and "last-modified"`[^1].

## Reproduction

1. Open a server-side rendered search page on staging: https://staging.openverse.org/search/?q=cat
2. See "Whoops, error". In the console, there is a "t is undefined" error logged.
3. See also that you can search client-side, but the search time event is not logged.

## Additional context

The immediate fix for this is to revert the commit in question. This fix is critical because the app's server-side rendered search became unusable.

To correctly send the search time event, we the following two issues need to be opened, with a lower importance label.

### Server
- [ ] #3776

### Client

- [x] #3777

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#access-control-expose-headers

obulat:
## Fixes

Fixes #3775 by @obulat
Fixes #2471 by @sarayourfriend 

## Description

This PR re-adds the old PR. 

It is not possible to call `useContext` in the media service because there is no Vue instance available there. This is why I rewrote the code to return the parameters to the store, together with the response data. Then, in the store, we use `this.$nuxt.$sendCustomEvent` to send the event with the returned parameters.

## Testing Instructions

Try making searches and viewing tag/source/creator pages. You should see the search time events logged in the console.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/data/media-service.ts
+++ b/frontend/src/data/media-service.ts
@@ -6,8 +6,13 @@ import type {
 import type { ApiService } from "~/data/api-service"
 import type { DetailFromMediaType, Media } from "~/types/media"
 import { AUDIO, type SupportedMediaType } from "~/constants/media"
+import type { Events } from "~/types/analytics"
 
 import type { AxiosResponse } from "axios"
+
+export type SearchTimeEvent =
+  | Events["AUDIO_SEARCH_RESPONSE_TIME"]
+  | Events["IMAGE_SEARCH_RESPONSE_TIME"]
 
 export interface MediaResult<
   T extends Media | Media[] | Record<string, Media>,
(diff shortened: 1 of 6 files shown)


sarayourfriend (approved):
Awesome! Now that Plausible supports filtering on more than one custom property, we could combine these into a single event with a `mediaType` attribute, but up to you.

sarayourfriend (approved):
LGTM!

dhruvkb (approved):

PR #4053, opened 2024-04-08
Update catalog to non-legacy environment names and point to new airflow.openverse.org

### LINKED ISSUE #737: Unify environment names
## Description

We have several different, inconsistent names we use for different environments (see WordPress/openverse-api#474 and WordPress/openverse-api#475). We should unify these across all services in the API repository, and ideally make them consistent with the environment names in [WordPress/openverse-frontend](https://github.com/WordPress/openverse-frontend). This may also require adjusting some settings in [WordPress/openverse-catalog](https://github.com/WordPress/openverse-catalog) and the infrastructure repo.

## Additional context

I think ideally we would have:

 - Any dev environment: `development`
 - Our pre-production environment: `staging`
 - Our production environment: `production`
 - Any environment where testing is being performed: `testing`

 That seems to coincide with the values we have for the frontend. 

## Resolution

sarayourfriend:
## Fixes

Fixes #737 by @AetherUnbound 
Related to #2037

## Description

Update the catalog code to use `production` instead of `prod` and never use `dev` under any circumstances to refer to a deployment environment (references to databases are impossible to fix and must remain). "local" is used to indicate a local development environment. "staging" is essentially unused in catalog runtime code, except for when referring to other resources, and those already use the contemporary, non-legacy names.

Realistically this only affects the slack messaging tool, which skip in non-production environments, and pr review reminders, which do the same. This PR updates both to use `production` now.

Because this is related to the new Airflow deployment, which uses the unified environment names, I've also gone ahead and added the documentation updates for catalog deployment in this PR. They're small documentation changes and it was easier to group this PR in with that.

## Testing Instructions

Read the updated documentation and confirm it all works.

CI should pass, all existing unit tests should pass. The updated code should make sense when reviewed.

I wasn't able to run the tests locally, neither on this branch or on main. Something is going on for me locally where the tests take ages (like literally up to a minute or more _per test_), so I'm relying on CI to run the tests for me. I am able to run the catalog Airflow instance itself though, and things look to be working fine there. Use `just catalog/up` and visit localhost:9090 to verify Airflow runs without DAG import errors.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/env.template
+++ b/catalog/env.template
@@ -20,7 +20,7 @@ AIRFLOW__WEBSERVER__SECRET_KEY=sample-secret-key=
 # Executor to use
 AIRFLOW__CORE__EXECUTOR=LocalExecutor
 # Environment this instance is being run in
-AIRFLOW_VAR_ENVIRONMENT=dev
+AIRFLOW_VAR_ENVIRONMENT=local
 
 ########################################################################################
 # API Keys
@@ -34,26 +34,26 @@ AIRFLOW_VAR_ENVIRONMENT=dev
 # Connection/Variable info
 ########################################################################################
 # Airflow primary metadata database
-# Change the following line in prod to use the appropriate DB
+# Change the following line in production to use the appropriate DB
 AIRFLOW__DATABASE__SQL_ALCHEMY_CONN=postgresql+psycopg2://airflow:airflow@upstream_db:5432/airflow
 # Remote logging connection ID
 # Replace "access_key" and "secret+key" with the real values. Secret key must be URL-encoded
 AIRFLOW_CONN_AWS_DEFAULT=aws://test_key:test_secret@?region_name=us-east-1&endpoint_url=http%3A%2F%2Fs3%3A5000
 
-# Catalog DB connection. Change the following line in prod to use the appropriate DB
+# Catalog DB connection. Change the following line in production to use the appropriate DB
 AIRFLOW_CONN_POSTGRES_OPENLEDGER_UPSTREAM=postgres://deploy:deploy@upstream_db:5432/openledger
 AIRFLOW_CONN_POSTGRES_OPENLEDGER_TESTING=postgres://deploy:deploy@upstream_db:5432/openledger
 AIRFLOW_CONN_POSTGRES_OPENLEDGER_API_STAGING=postgres://deploy:deploy@db:5432/openledger
 TEST_CONN_ID=postgres_openledger_testing
 
-# Elasticsearch connections. Change the following line in prod to use the appropriate URLs.
+# Elasticsearch connections. Change the following line in production to use the appropriate URLs.
 AIRFLOW_CONN_ELASTICSEARCH_HTTP_PRODUCTION=http://es:9200
 AIRFLOW_CONN_ELASTICSEARCH_HTTP_STAGING=http://es:9200
 
 # AWS CloudWatch connection. Change the following line to toggle alarms during a Data Refresh.
 # AIRFLOW_CONN_AWS_CLOUDWATCH=aws://<key>:<secret>@?region_name=us-east-1
 
-# API DB connection. Change the following line in prod to use the appropriate DB
+# API DB connection. Change the following line in production to use the appropriate DB
 AIRFLOW_CONN_POSTGRES_OPENLEDGER_API=postgres://deploy:deploy@db:5432/openledger
 
 # Slack webhook connection info (note that these values are modified by the Docker entrypoint)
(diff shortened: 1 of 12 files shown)


catalog/env.template
```diff
@@ -107,7 +107,7 @@ OPENVERSE_BUCKET=openverse-storage
 # Seconds to wait before poking for availability of the data refresh pool when running a data_refresh
 # DAG. Used to shorten the time for testing purposes.
 DATA_REFRESH_POKE_INTERVAL=5
-# The pool to use for the data refresh tasks. Defaults to `data_refresh` in production but
+# The pool to use for the data refresh tasks. Defaults to `data_refresh` in productionuction but
```
obulat, line 110:
```suggestion
# The pool to use for the data refresh tasks. Defaults to `data_refresh` in production but
```

catalog/pytest.ini
```diff
@@ -18,6 +18,8 @@ addopts =
     --disable-socket
     --allow-unix-socket
 
+log_cli = True
```
obulat, line 21: Is this left-over from debugging the tests?
Tests run fine locally for me, but with this, the output does not look correct (the progress bar is not displayed as expected)

sarayourfriend: Left over, yes, I'll remove it.

```suggestion
```

documentation/catalog/guides/deploy.md
```diff
@@ -25,25 +24,34 @@ the app is built and tagged, deploy production:
...
+
+- If _any_ DAGs are running, the playbook will not apply the changes and will
+  let you know that. If this happens, visit Airflow and confirm the list of
+  running DAGs. If they can be stopped, stop them. If they need to be waited
+  for, wait until they are done, then run the playbook again. If you must deploy
+  and cannot wait for the DAGs to finish (or, if they are deferred and cannot
+  finish), run the playbook with `-e airflow_force=true` to ignore the running
+  DAGs check.
+- See the [setup](#setup) section above for more information about when to
+  decide if it is okay to deploy when DAGs are running.
```
AetherUnbound, line 42: Should we also have information here about how to apply changes to the actual EC2 instances (e.g. if launch template changes).

AetherUnbound (commented):
This looks good, only one question regarding additional documentation for the ASG/EC2 update case during deployment.

Also, since our production Airflow uses `prod` now, we'll need to update this variable too: https://github.com/WordPress/openverse-infrastructure/blob/7a9f8b4e6693463dce1e0afe2dfd6c4d61064aca/ansible/roles/airflow/templates/compose.yml.j2#L11

I believe we can add the variable in the UI and that will take precedence. So we can do that as soon as this PR is merged to ensure no running DAGs are affected by the `prod`/`production` difference, then remove it once the infrastructure repo is updated.

sarayourfriend:
> I believe we can add the variable in the UI and that will take precedence. So we can do that as soon as this PR is merged to ensure no running DAGs are affected by the prod/production difference, then remove it once the infrastructure repo is updated.

I was trying to understand variable precedence in Airflow. I'm pretty sure environment variables take precendence over UI. When I set `ENVIRONMENT` to `notlocal` in my local UI, and then run `airflow variables get ENVIRONMENT`, I get the value defined in the environment file.

This is for connections, but I think variables follow the same logic: https://forum.astronomer.io/t/what-precedence-does-airflow-determine-when-searching-for-a-single-connection/630

AetherUnbound:
Yes I believe you're right, when I've looked into it before I recall seeing that the environment variables were higher than the UI (which is unfortunate because they can't be changed as rapidly!). We'll definitely need to coordinate on that end before merging & deploying.

sarayourfriend:
Should we switch over to setting variables using airflow CLI instead of environment variables, so that we can have that flexibility? It's easy enough with the Ansible playbook.

sarayourfriend:
@AetherUnbound It seems like you've looked at this PR a bit, are you able to review it at all?

AetherUnbound:
> Should we switch over to setting variables using airflow CLI instead of environment variables, so that we can have that flexibility? It's easy enough with the Ansible playbook.

Ahh, that's right we could! That would certainly make the experience much more uniform!

AetherUnbound (approved):
LGTM given the deployment constraints!

sarayourfriend:
Related infrastructure PR: https://github.com/WordPress/openverse-infrastructure/pull/848

sarayourfriend:
I haven't deployed this because Freesound is running. @WordPress/openverse-catalog and in particular @stacimc, I'm pretty sure I should _not_ stop it, but it's been running for 8 days :thinking: It is ingesting a lot of records though, after failing a bunch of previous runs. Is that expected?

stacimc:
@sarayourfriend I've just manually failed that Freesound run in order to load the data we've got so far, so that we can deploy the catalog. You can go ahead with this as well, I'll wait to resume the DAG until you're done!

stacimc:
To clarify my previous message, we had not yet deployed this (or the routine catalog deploy). We will need to do both; Freesound is paused so we can go ahead whenever. I will handle deploying from here 👍

PR #4065, opened 2024-04-08
Round decimal record difference percent to three decimals after zeros

### LINKED ISSUE #1581: Truncate data refresh percent change report to 3 digits after the decimal
## Description

PR WordPress/openverse-catalog#636 added a count difference reporting message after the data refresh has been run. Part of this count is the percent change, which is given as a float. The difficulty with printing a straight float is that the numbers after the decimal can vary wildly. Here are a few examples:

> Record count difference for audio: 174,290 → 817,198
> Change: +642,908 (+368.8725687073269% Δ)

> Record count difference for image: 563,561,088 → 564,775,616
> Change: +1,214,528 (+0.21550955625949816% Δ)

> Record count difference for image: 579,553,984 → 579,554,880
> Change: +896 (+0.0001546016462204149% Δ)

We have a wide range of potential digits (in the hundreds to hundredths of a percent) and it can be hard to capture the level of information we would like for all cases. I think it would be best for us to have **3 significant digits after the decimal or closest zero, rounded up**. This would mean for the above three cases the output would be:

- `368.8725687073269` -> `368.873`
- `0.21550955625949816` -> `0.216`
- `0.0001546016462204149` -> `0.000155`

## Alternatives

Leave as is, or do some basic string truncation which could leave us with values like `0.000%`.

## Implementation

mattfergoda:
## Fixes

Fixes #1581 by @AetherUnbound 

## Description

Round the data refresh percent change report to 3 digits after the decimal, ignoring zeros. E.g. 100.012365 -> 100.0124.

## Testing Instructions

`just catalog/dags/test`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_refresh/reporting.py
+++ b/catalog/dags/data_refresh/reporting.py
@@ -23,7 +23,7 @@ def report_record_difference(before: dict, after: dict, media_type: str, dag_id:
     total_after = sum(after.values())
     count_diff = total_after - total_before
     if total_before > 0:
-        percent_diff = (count_diff / total_before) * 100
+        percent_diff = count_diff / total_before
     else:
         percent_diff = float("inf")
     breakdown_diff = {k: after.get(k, 0) - before.get(k, 0) for k in all_keys}
@@ -40,7 +40,7 @@ def report_record_difference(before: dict, after: dict, media_type: str, dag_id:
 Data refresh for {media_type} complete! :tada:
 _Note: All values are retrieved from elasticsearch_
 *Record count difference for `{media_type}`*: {total_before:,} → {total_after:,}
-*Change*: {count_diff:+,} ({percent_diff:+}% Δ)
+*Change*: {count_diff:+,} ({percent_diff:+%} Δ)
 *Breakdown of changes*:
 {breakdown_message}
 """
--- a/catalog/tests/dags/data_refresh/test_reporting.py
+++ b/catalog/tests/dags/data_refresh/test_reporting.py
@@ -1,6 +1,9 @@
 import pytest
 
-from data_refresh.reporting import report_record_difference, report_status
+from data_refresh.reporting import (
+    report_record_difference,
+    report_status,
+)
 
 
 @pytest.mark.parametrize(
@@ -9,23 +12,23 @@ from data_refresh.reporting import report_record_difference, report_status
         [
             {"src1": 1, "src2": 19},
             {"src1": 2, "src2": 38},
-            ["20 → 40", "+20 (+100.0%", "`src1`:+1", "`src2`:+19"],
+            ["20 → 40", "+20 (+100.000000%", "`src1`:+1", "`src2`:+19"],
         ],
         [
             {"src1": 1, "src2": 19},
             {"src1": 3, "src2": 57, "src3": 20},
-            ["20 → 80", "+60 (+300.0%", "`src1`:+2", "`src2`:+38", "`src3`:+20"],
+            ["20 → 80", "+60 (+300.000000%", "`src1`:+2", "`src2`:+38", "`src3`:+20"],
         ],
         [
             {"src1": 4, "src2": 21},
             {"src1": 4},
             # Unchanged source count shouldn't show up
-            ["25 → 4", "-21 (-84.0%", "`src2`:-21"],
+            ["25 → 4", "-21 (-84.000000%", "`src2`:-21"],
         ],
         [
             {"src1": 4000, "src2": 20},
             {"src1": 2000, "src2": 10},
-            ["4,020 → 2,010", "-2,010 (-50.0%", "`src1`:-2,000", "`src2`:-10"],
+            ["4,020 → 2,010", "-2,010 (-50.000000%", "`src1`:-2,000", "`src2`:-10"],
         ],
         [
             {},
@@ -35,7 +38,12 @@ from data_refresh.reporting import report_record_difference, report_status
         [
             {"src1": 10, "src2": 10},
             {},
-            ["20 → 0", "-20 (-100.0%", "`src1`:-10", "`src2`:-10"],
+            ["20 → 0", "-20 (-100.000000%", "`src1`:-10", "`src2`:-10"],
+        ],
+        [
+            {"src1": 5000000000},
+            {"src1": 4938271605},
+            ["5,000,000,000 → 4,938,271,605", "-61,728,395 (-1.234568%"],
         ],
         [{"src1": 4}, {"src1": 4}, ["Sources not listed had no change in count"]],
         [{}, {}, ["Both indices missing? No breakdown to show"]],


catalog/dags/data_refresh/reporting.py
```diff
@@ -30,9 +30,29 @@ def report_record_difference(before: str, after: str, media_type: str, dag_id: s
...
+    if num % 1 == 0:
+        return int(num)
+
+    # Round decimal part to 3 sig figs. Leave int part untouched.
+    [int_part, dec_part] = str(num).split(".")
+
+    # Have to awkwardly prepend the decimal part with '0.' to get correct
+    # rounding behavior, then slice '0.' off before concatenating
+    # the final number.
+    return float(int_part + "." + "{:.{n}g}".format(float("0." + dec_part), n=n)[2:])
```
sarayourfriend, line 58: This is over complicating things a bit, I think for the sake of always having 3 significant digits. Using some fancy [string format spec](https://docs.python.org/3/library/string.html#format-specification-mini-language), the whole thing can be simplified, so long as we don't care that `0.00005646%` will get truncated to just `0.00006%`... and frankly, at such small percentages, I really don't think it matters to have accuracy past 1/100, it just isn't interesting or useful information to have such abstract and small percentages.

I've left a suggestion on the line above for format spec to use.

catalog/dags/data_refresh/reporting.py
```diff
@@ -30,9 +30,29 @@ def report_record_difference(before: str, after: str, media_type: str, dag_id: s
 Data refresh for {media_type} complete! :tada:
 _Note: All values are row estimates and are not (but nearly) exact_
 *Record count difference for `{media_type}`*: {before:,} → {after:,}
-*Change*: {count_diff:+,} ({percent_diff:+}% Δ)
+*Change*: {count_diff:+,} ({_round_sig_figs(percent_diff, 3):+}% Δ)
```
sarayourfriend, line 33: As mentioned below, the `_round_sig_figs` can be removed if we simplify things using just format spec significant digits instructions.

If we use `g`, the general format, this will use scientific notation in some cases:

```
>>> "{n:+3g}".format(n=0.0000512346888129371512345)
'+5.12347e-05'
```

So, I think it would be better to remove the `* 100` on line 24, and use the `%` spec instead:

```
>>> "{n:+3%}".format(n=0.000000512346888129371512345)
'+0.000051%'
```

```suggestion
*Change*: {count_diff:+,} (percent_diff:+3%} Δ)
```

If you apply that suggestion, please update line 24 as well.

sarayourfriend (changes requested):
Thanks for the PR @mattfergoda! I'm pretty sure there is a simpler way to accomplish this, as long as we don't get too caught up with always wanting three digits on very small decimals. For our context, I'd say that level of precision is not worth the code required to achieve it, when using format-spec will give us all the information we could possibly actually need and use.

catalog/dags/data_refresh/reporting.py
AetherUnbound: @sarayourfriend I was talking about this with @mattfergoda in person - many times for the data refresh, we do have changes that are quite small by percentage but still register as a +/- half a million record difference. Things like `0.0654%` or even `0.00481%`. For me, and perhaps this is just my own perspective, truncating those down completely does lose some of the specificity we have. Maybe it doesn't matter, but whatever change we make I'd like to _at least_ be sure we have a few digits available a couple of decimal points out.

catalog/dags/data_refresh/reporting.py
AetherUnbound: FWIW the `3` here does not seem to have any effect 🤔 
```python
[nav] In [3]: "{n:+3%}".format(n=0.001000512346888129371512345)
Out[3]: '+0.100051%'

[nav] In [4]: "{n:+3%}".format(n=0.001000512340888129371512345)
Out[4]: '+0.100051%'

[nav] In [5]: "{n:+3%}".format(n=0.0000512340888129371512345)
Out[5]: '+0.005123%'

[nav] In [6]: "{n:+2%}".format(n=0.0000512340888129371512345)
Out[6]: '+0.005123%'

[nav] In [7]: "{n:+1%}".format(n=0.0000512340888129371512345)
Out[7]: '+0.005123%'

[nav] In [8]: "{n:+%}".format(n=0.0000512340888129371512345)
Out[8]: '+0.005123%'

```

mattfergoda: If `"{n:+3%}".format(num)` is the agreed way forward, I'm happy to make this change and update the tests.

catalog/dags/data_refresh/reporting.py
sarayourfriend: What do you mean by "truncating them down completely"? As I wrote:

> 0.00005646% will get truncated to just 0.00006%

In what way is this a meaningful difference in the information, when the actual count is also present?

catalog/dags/data_refresh/reporting.py
sarayourfriend: The number is apparently ignored for `%`, TIL. It does have an effect on general formatting, and I converted to `%` from that, not realising it wouldn't make a difference.

```
>>> "{n:+3g}".format(n=52.212331223456234512)
'+52.2123'
>>> "{n:+3g}".format(n=0.001000512346888129371512345)
'+0.00100051'
>>> "{n:+3g}".format(n=0.6888129371512345)
'+0.688813'
>>> "{n:+3g}".format(n=0.000000000000000000000000888129371512345)
'+8.88129e-25'
```

catalog/dags/data_refresh/reporting.py
AetherUnbound: Those are all fair points! We can go with the much simpler implementation with that in mind 🙂

catalog/dags/data_refresh/reporting.py
AetherUnbound: Sara's right, I think this approach will be way more straightforward than anything I'm suggesting. If you're comfortable with making those changes Matt, please proceed!

mattfergoda:
@AetherUnbound @sarayourfriend I simplified the rounding logic based on the above conversation and updated the tests accordingly. I added a parametrized test case to the existing ones where the decimal value is truncated.

sarayourfriend (commented):
@mattfergoda, thanks for updating the PR. Left one comment, but it's unclear to me whether a change is needed until CI runs. Can you rebase the PR for that?

AetherUnbound (approved):
Thank you for the contribution @mattfergoda, and for your patience while iterating on this issue! It looks like GitHub is having an outage at the moment which is preventing CI from running, but this LGTM.

AetherUnbound:
Ah, it looks like this is conflicting with #4067 - it seems easiest to squash the commits and rebase, in the interest of time I'll go ahead and take care of that.

sarayourfriend (approved):
LGMT! Thanks again @mattfergoda :+1:

Issue #3548, opened 2023-12-18
Create a DAG to suggest creation of `ContentProvider` models for new sources


obulat:
## Problem

As stated in #3393, we can have some sources that don't have a `ContentProvider` record in the API and therefore are not listed on the `openverse.org/sources` or in the filters list. We need to have a way to check for missing providers regularly.
 
## Description

The stats endpoint checks the Redis cache for the value of `sources-<mediaType>` key, and if it's not available, it sends an ES request for source aggregation and caches it.

https://github.com/WordPress/openverse/blob/b004dc620897d832fefa01c97a4fe16206adc713/api/api/controllers/search_controller.py#L565

The DAG should get the cached sources (or send the ES request for sources, if cache is not available), and compare it to the value that's returned by the `/images/stats` endpoint (the one that only returns sources that have a `ContentProvider` and does not have "Hide Content" set to true).

This DAG should send a Slack notification with the details of missing `ContentProvider`s so that the maintainers can add them.

PR #4067, opened 2024-04-08
Get change in record counts during data refresh from Elasticsearch

### LINKED ISSUE #1404: Report data refresh count change by provider
## Description

PR WordPress/openverse-catalog#636 added record difference reporting for changes before & after the data refresh. These stats are very useful, but it may also be helpful to get more specific information on _which providers_ contributed to the change. In addition to including the total change, we could report the number of new records per-provider.

This query would need to be updated:

https://github.com/WordPress/openverse-catalog/blob/d4dbf4d0617aeee9610adadbbee12be641174c0b/openverse_catalog/dags/data_refresh/dag_factory.py#L184-L196

Unfortunately this more optimized query would make grouping by provider impossible. We'd need to go with a query like the following, which is _not_ optimized in the same way:

```sql
SELECT provider, count(*)
FROM {data_refresh.media_type}
GROUP BY 1;
```

It's worth noting that this query runs against the API database, _not_ the catalog DB, and we do have an index on `provider` in that database. The other good news is that this step gets run at the start of the data refresh DAG, concurrently with the matview refresh (which for our larger table, takes about 12 hours). The rest of the steps are dependent on this task, but even if the query takes several hours it will complete before the matview refresh and not block any downstream tasks. The final reporting of the record counts will take longer after the refresh is complete, but the improved information from the report is probably worth it :slightly_smiling_face:

## Additional context

On a code level, I'd consider this an easier issue to dive into. The difficult comes when we want to test this locally, as one also must set up the API stack in order to trigger the data refresh. If you're interested in taking on this issue please let us know! We're happy to help you walk through the steps to get the API set up :slightly_smiling_face:

## Implementation

mjtung:
## Fixes

Fixes #1404 by @AetherUnbound 

## Description

**Change to reporting output**
In addition to the current reporting message for data refresh, prints out a "breakdown of changes", which gives the number of records changed grouped by each source. Here is a sample of the output (the final 3 lines are new, the others remain the same as before):
```
Data refresh for media complete! :tada:
_Note: All values are now from elasticsearch_
*Record count difference for `media`*: 20 → 40
*Change*: +20 (+100.0% Δ)
*Breakdown of changes*:
src1:+1
src2:+19
```
**Logic change to produce the change to reporting output**
Submits a query to Elasticsearch to get the breakdown of the `doc_count` per `source`, instead of using an SQL query to get data from the postgres db. Then the result is parsed into a dictionary and passed into the reporting task in the DAG to print the message. The Elasticsearch query is submitted before and after the data refresh.

## Testing Instructions

1. Run Airflow DAG `audio_data_refresh`
2. Check logs for task `report_record_counts` in this DAG to see that it prints a breakdown as above.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (commented):
Thank you for your contribution, @mjtung! 

I don't have time to test this PR locally today, so I can't approve it, but I wanted to comment to say that this is an awesome PR. The code is really well-documented and well-tested.

I have a question about the aggregation query. In the API's `/stats` endpoint, we use the `size` parameters twice for this query, should we add it here, too?

https://github.com/WordPress/openverse/blob/3c6224f3c4bb8a17d29cec60e3ead667b1204acd/api/api/controllers/search_controller.py#L577-L592

By the way, I think this PR will also help resolve #3548 in the future. It would probably be easy to compare the results from this message with the `sources` returned by the `/stats` endpoint to suggest new `ContentProvider` sources instead of creating a separate DAG for #3548, @AetherUnbound , what do you think?

catalog/dags/common/elasticsearch.py
```diff
@@ -78,6 +78,29 @@ def remove_excluded_index_settings(index_config):
...
+    """
+    es_hook = ElasticsearchPythonHook(hosts=[es_host])
+    body = {"aggs": {"unique_sources": {"terms": {"field": "source"}}}}
+
+    # Unfornately the ElasticsearchPythonHook's search function doesn't work with aggs, because it returns the "hits" key only
+    # See source code at https://airflow.apache.org/docs/apache-airflow-providers-elasticsearch/stable/_modules/airflow/providers/elasticsearch/hooks/elasticsearch.html#ElasticsearchPythonHook
+    # Therefore using get_conn to call the search from the ES client
+    es_client = es_hook.get_conn
+    # es_result object looks like: [{'key': 'flickr', 'doc_count': 2500}, {'key': 'stocksnap', 'doc_count': 2500}]
+    es_result = es_client.search(body=body, index=index)
```
AetherUnbound, line 95: (RE: missing index at the start)
```suggestion
    try:
        es_result = es_client.search(body=body, index=index)
    except NotFoundError as err:
        logging.warning(f"Index {index} does not appear to exist: {err}")
        return {}
```

catalog/dags/data_refresh/reporting.py
```diff
@@ -17,21 +17,23 @@ def report_status(media_type: str, message: str, dag_id: str):
...
-    # Note for formatting:
-    # '+' - number will always have a sign in front of it
-    # ',' - number is comma separated
-    # '.' - number is a float
+def report_record_difference(before: dict, after: dict, media_type: str, dag_id: str):
+    all_keys = before.keys() | after.keys()
+    total_before = sum(before.values())
+    total_after = sum(after.values())
+    count_diff = total_after - total_before
+    percent_diff = (count_diff / total_before) * 100
```
AetherUnbound, line 25: Note that when we don't have an index, this fails:
```
[2024-04-13, 01:00:27 UTC] {taskinstance.py:2731} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 439, in _execute_task
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 200, in execute
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 217, in execute_callable
  File "/opt/airflow/catalog/dags/data_refresh/reporting.py", line 25, in report_record_difference
    percent_diff = (count_diff / total_before) * 100
ZeroDivisionError: division by zero
```

I was able to use the following locally with success, but feel free to take a different approach:
```suggestion
    if total_before == 0:
        percent_diff = float("inf")
    else:
        percent_diff = (count_diff / total_before) * 100
```

catalog/dags/data_refresh/dag_factory.py
```diff
@@ -75,27 +74,12 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
...
+        es_host = es.get_es_host(environment=PRODUCTION)
 
         # Get the current number of records in the target API table
-        before_record_count = PGExecuteQueryOperator(
-            task_id="get_before_record_count",
-            conn_id=OPENLEDGER_API_CONN_ID,
-            sql=count_sql,
-            handler=single_value,
-            return_last=True,
+        before_record_count = es.get_record_count_group_by_sources(
```
AetherUnbound, line 80: By default, the function name is used as a task ID (and since we call this twice, Airflow helpfully adds `__1` to the end of the second call). We can make the task names a bit more descriptive in the UI by using the following:
```suggestion
        before_record_count = es.get_record_count_group_by_sources.override(
            task_id="get_before_record_count"
        )(
```

catalog/dags/data_refresh/dag_factory.py
```diff
@@ -105,25 +89,18 @@ def create_data_refresh_dag(data_refresh: DataRefresh, external_dag_ids: Sequenc
         )
 
         # Get the final number of records in the API table after the refresh
-        after_record_count = PGExecuteQueryOperator(
-            task_id="get_after_record_count",
-            conn_id=OPENLEDGER_API_CONN_ID,
-            sql=count_sql,
-            handler=single_value,
-            return_last=True,
+        after_record_count = es.get_record_count_group_by_sources(
```
AetherUnbound, line 92: Same note here, except that in the case where we don't have an index, the skipping of the delete index step makes this skip too. We can correct that with a trigger rule:

```suggestion
        after_record_count = es.get_record_count_group_by_sources.override(
            task_id="get_after_record_count",
            trigger_rule=TriggerRule.NONE_FAILED_OR_SKIPPED,
        )(
```

catalog/dags/data_refresh/dag_factory.py
krysal, line 80: Let's conserve the `task_id` of the counting tasks.

```suggestion
        before_record_count = es.get_record_count_group_by_sources.override(
            task_id="get_before_record_count"
        )(
```

catalog/dags/data_refresh/dag_factory.py
krysal, line 92:
```suggestion
        after_record_count = es.get_record_count_group_by_sources.override(
            task_id="get_after_record_count"
        )(
```

AetherUnbound (dismissed):
This is an EXCELLENT change, thanks so much @mjtung! One issue that I encountered was the case where we actually _don't_ have an index to start with. This is a rare case, but it's happened before when we migrated to a new version of Elasticsearch. I've pointed out some places that would need to be changed as a result, and some errors that come up along with it.

I was able to simulate this case with the following:
1. Run `just api/init` to start up the API/ingestion server and run the first data refresh
2. Use Elasticvue pointed at the local cluster to delete all indices after the above finished
3. Run the catalog with `just c`
4. Kick off an audio data refresh

Lastly, the skipping of the final trigger delete task _also_ causes the `get_after_record_counts` task to run immediately, before the `point_alias` step (or most of the data refresh) actually completes.  I've also suggested a change that should help prevent that:

```diff
diff --git a/catalog/dags/data_refresh/create_filtered_index.py b/catalog/dags/data_refresh/create_filtered_index.py
index a77c2d53e..869181ccf 100644
--- a/catalog/dags/data_refresh/create_filtered_index.py
+++ b/catalog/dags/data_refresh/create_filtered_index.py
@@ -133,6 +133,7 @@ def create_filtered_index_creation_task_groups(
                     get_current_index_if_exists.task_id, "return_value"
                 ),
             },
+            trigger_rule=TriggerRule.NONE_FAILED,
         )
 
         index_readiness_check >> do_point_alias
```

Almost all of this would have been uncovered on `main` if we did the same thing, so this is definitely not the result of your changes 😅 but good to tackle it now anyway!

catalog/dags/data_refresh/reporting.py
```diff
@@ -17,21 +17,23 @@ def report_status(media_type: str, message: str, dag_id: str):
...
+    breakdown_diff = {k: after.get(k, 0) - before.get(k, 0) for k in all_keys}
+    breakdown_message = "\n".join(f"{k}:{v:+,}" for k, v in breakdown_diff.items())
```
AetherUnbound, line 27: One other formatting suggestion:

```suggestion
    breakdown_message = "\n".join(f"`{k}`: {v:+,}" for k, v in breakdown_diff.items())
```

catalog/tests/dags/data_refresh/test_reporting.py
```diff
@@ -6,10 +6,26 @@
...
+        [
+            {"src1": 1, "src2": 19},
+            {"src1": 2, "src2": 38},
+            ["20 → 40", "+20 (+100.0%", "src1:+1", "src2:+19"],
+        ],
+        [
+            {"src1": 1, "src2": 19},
+            {"src1": 3, "src2": 57, "src3": 20},
+            ["20 → 80", "+60 (+300.0%", "src1:+2", "src2:+38", "src3:+20"],
+        ],
```
krysal, line 18: Nice that it handles well new sources 💯

krysal (approved):
I tried it locally, and it works wonderfully! Excellent job here, @mjtung 👏  I left some suggestions, but they're only about the form, and since Olga still has to review it, I'm not blocking on them.

Again, great work!

catalog/dags/data_refresh/dag_factory.py
krysal: Perhaps `NONE_FAILED` would be better because, locally, we skip the tasks related to Cloudwatch alarms, but it would still be useful to have the report.

AetherUnbound: It looks like `NONE_FAILED_OR_SKIPPED` is actually deprecated, so I agree I think `NONE_FAILED` is best (editing the suggestion for it)

mjtung:
> I was able to simulate this case with the following:
> 
> 1. Run `just api/init` to start up the API/ingestion server and run the first data refresh
> 2. Use Elasticvue pointed at the local cluster to delete all indices after the above finished
> 3. Run the catalog with `just c`
> 4. Kick off an audio data refresh

Thanks - any tips please on how to connect Elasticvue to the cluster running locally in docker?

AetherUnbound:
> Thanks - any tips please on how to connect Elasticvue to the cluster running locally in docker?

You should be able to get [Elasticvue here](https://elasticvue.com/) (I highly recommend the browser extension version so you don't have to deal with any cross-origin issues), then connect to the cluster using the URL http://localhost:50292!

mjtung:
> You should be able to get [Elasticvue here](https://elasticvue.com/) (I highly recommend the browser extension version so you don't have to deal with any cross-origin issues), then connect to the cluster using the URL http://localhost:50292!

Great thanks, that worked! I then found the port number 50292 in the `quickstart.md` file; any chance this could be added to somewhere on the website like https://docs.openverse.org/catalog/guides/quickstart.html? I couldn't find much info on Elasticsearch there.

I think I have addressed all the requested changes now.

obulat:
> Great thanks, that worked! I then found the port number 50292 in the `quickstart.md` file; any chance this could be added to somewhere on the website like https://docs.openverse.org/catalog/guides/quickstart.html? I couldn't find much info on Elasticsearch there.

We can open an issue for more prominent documenting of port numbers. In fact, we should mention the console output you see when you run `just up`: it runs `just ps` at the end which prints out all of the container ports:
```
================================================================================
                                 Service Ports                                  
================================================================================
webserver (catalog):
-  http://0.0.0.0:9090  (→ 8080)
nginx (nginx):
-  http://0.0.0.0:50270 (→ 8080)
ingestion_server (ingestion_server):
-  http://0.0.0.0:50281 (→ 8001)
web (api):
-  http://0.0.0.0:50280 (→ 50280)
plausible (plausible/analytics):
-  http://0.0.0.0:50288 (→ 8000)
s3 (minio/minio):
-  http://0.0.0.0:5010  (→ 5000)
-  http://0.0.0.0:5011  (→ 5001)
frontend_nginx (frontend_nginx):
-  http://0.0.0.0:50290 (→ 8080)
upstream_db (upstream_db):
-  http://0.0.0.0:50255 (→ 5432)
es (docker.elastic.co/elasticsearch/elasticsearch):
-  http://0.0.0.0:50292 (→ 9200)
cache (redis):
-  http://0.0.0.0:50263 (→ 6379)
db (postgres):
-  http://0.0.0.0:50254 (→ 5432)
================================================================================
```

obulat (approved):
I tested these changes locally, removing the index before data refresh and removing data from the catalog - it works perfectly!
Thank you so much for your contribution, @mjtung 🎉

AetherUnbound:
@mjtung I wanted to follow up and say that this week's audio data refresh message went out, and the new source-specific changes look great!!

image

PR #4068, opened 2024-04-08
Add verbose logging option to `ProviderDataIngester` 

### LINKED ISSUE #1420: Add verbose logging option to `ProviderDataIngester`
## Description

When troubleshooting a provider API script using the new `ProviderDataIngester` class, it can be difficult sometimes to tell at which step in the execution an issue is happening or what the data is at that step.

It would be useful to have logging at a high level of verbosity which could be used locally in test scenarios to see what the values are at each step in the process (e.g. `get_next_query_params`, `get_batch`, `get_batch_data`, `process_batch`, etc.

## Implementation

nicoepp:
To log for a certain DAG set SHOULD_VERBOSE_LOG to a list containing that dag id.

For example, set the SHOULD_VERBOSE_LOG Airflow variable to: ["smk_workflow"]

## Fixes

Fixes #1420 by @AetherUnbound

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -195,6 +195,16 @@ def __init__(
...
 
+        environment = Variable.get(
+            "ENVIRONMENT", default_var='dev'
+        )
+
+        should_verbose_log = self.dag_id in Variable.get(
+            "SHOULD_VERBOSE_LOG", default_var=[], deserialize_json=True
+        )
+
+        self._should_verbose_log = environment != 'prod' or should_verbose_log
```
sarayourfriend, line 206:
```suggestion
        self._should_verbose_log = environment == "local" or should_verbose_log
```

Small change to be in line with #4053.

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -195,6 +195,16 @@ def __init__(
 
         self.ingestion_errors: list[IngestionError] = []  # Keep track of skipped errors
 
+        environment = Variable.get(
+            "ENVIRONMENT", default_var='dev'
```
sarayourfriend, line 199:
```suggestion
            "ENVIRONMENT", default_var="local"
```

sarayourfriend (changes requested):
Thanks for the PR @nicoepp! I've left two requested changes to align with #4053, which would have conflicts with this branch.

Also: can you run linting locally? The CI linting check failed. You can run it with `just lint` from the root of the repository.

Thanks again :slightly_smiling_face:

catalog/dags/common/slack.py
```diff
@@ -396,15 +396,15 @@ def on_failure_callback(context: dict) -> None:
             log.info("Forgoing Slack alert due to upstream failures")
             return
         exception_message = f"""
-*Exception*: {exception}
-*Exception Type*: `{exception.__class__.__module__}.{exception.__class__.__name__}`
+*Exception Type*: `{exception.__class__.__module__}.{exception.__class__.__name__}
+*Exception*: {exception}`
```
sarayourfriend, line 400: Looks like the trailing `` ` `` got lost in the mix here.

```suggestion
*Exception Type*: `{exception.__class__.__module__}.{exception.__class__.__name__}`
*Exception*: {exception}
```

nicoepp:
Thanks for you feedback @sarayourfriend! I pushed some commits and I'm running the linter locally now but I gotta go now. I will hopefully push the linting changes tonight.

stacimc (changes requested):
Thanks very much @nicoepp! I've left a few suggestions to clarify the (admittedly confusing 😄) distinction between terms like  "record"/"batch item" in the logs. In addition, I think we should add at least one more verbose log to log the full result of `get_batch_data`.

Since that is _quite_ the verbose log, I've also asked that we not enable this logging by default locally, treating this more as an optional development tool. (CC @AetherUnbound to give you a chance to disagree with me if that's not what you had in mind! 😄)

nicoepp:
I'm playing around with the idea to log the full result of `get_batch_data` as suggested by @stacimc. I'm running into problems logging the full list of dictionaries of the returned batch. The chrome tab where I start the DAG in Airflow and where I try to view the log starts becoming unresponsive and then crashes since the log gets very very big! 

I modified my code now to just log the first 5 and the last 5 dictionaries in each list returned by `get_batch_data`. See attached screenshot. Would that work for you guys? 

image

stacimc:
> I modified my code now to just log the first 5 and the last 5 dictionaries in each list returned by get_batch_data. See attached screenshot. Would that work for you guys?

Absolutely perfect, even just the first five is probably fine for our needs!

nicoepp:
Here is the full log file (from the screenshot) for better reference:
[dag_id=smk_workflow_run_id=manual__2024-04-10T20_03_28.log](https://github.com/WordPress/openverse/files/14937378/dag_id.smk_workflow_run_id.manual__2024-04-10T20_03_28.log)

nicoepp:
Ok. Thanks for the feedback. I'm only logging the first 5 dictionaries of the batch for now. 

I also pushed a commit making the verbose logging always controlled by the Airflow variable. No more checking the "ENVIRONMENT" variable while deciding to log or not.

The variable value has to look like this:
image

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -401,6 +407,7 @@ def get_batch(self, query_params: dict) -> tuple[list | None, bool]:
 
         # Build a list of records from the response
         batch = self.get_batch_data(response_json)
+        self._verbose_log("Batch data:", batch)
```
stacimc, line 410:
```suggestion
        self._verbose_log(
            f"Got batch with {len(batch)} items. The first items are:",
            batch[:5] if batch else None
        )
```

Can we update this message with the actual batch len, to make sure that it's not misleading that only 5 are actually logged?

We could also do the list slicing here, and then have `_verbose_log` log **all** the items in `data` to make it more flexible for future use cases.

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -512,3 +527,10 @@ def _commit_records(self) -> int:
             total += store.commit()
         logger.info(f"Committed {total} records")
         return total
+
+    def _verbose_log(self, msg: str, data=[]):
+        if self._should_verbose_log:
+            logger.info(msg)
+            if data:
+                for item in data[:5]:
+                    logger.info(f"\t{item}")
```
stacimc, line 536: I love the added formatting, much visually cleaner! However instead of logging each line separately, I think we should do this all in one log statement by instead appending the items from `data` to the `msg` before logging. Something like:
```
if data:
    msg += "\n".join([f"\t{item}" for item in data])
```

stacimc (changes requested):
This looks great! Just a few very small comments and then we should ship this 💯

stacimc:
I've also just noticed the catalog test failures -- that's because we are mocking all the results of `Variable.get` in the provider data ingester tests and are now getting an unexpected new call. @nicoepp you'll need to update [test_provider_data_ingester.py](https://github.com/WordPress/openverse/blob/1ecea3c3fc0853553845ecde10b6498b38fc3f6a/catalog/tests/dags/providers/provider_api_scripts/test_provider_data_ingester.py#L109), everywhere you see a line like this:

```diff
        MockVariable.get.side_effect = [
            20,  # ingestion_limit
            {},  # skipped_ingestion_errors
+           [],  # should_verbose_log 
        ]
```

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
nicoepp: I'm running into several problems changing the code as suggested:

It seems like batch is not always a list, it sometimes can be a dict which makes the tests fail since a dict cannot be sliced. This was not a problem if we did the slicing inside the `_verbose_log` method because we would first check if the `Variable` was set and the tests never activate verbose logging. See following screenshot showing the test that is mocking a dict batch instead of a list batch.

image

Also, the `len(batch)` fails since batch can be `None`. So, code that could handle all this would look something like this:
```python
    # ...

    # Build a list of records from the response
    batch = self.get_batch_data(response_json)

    batch_to_log = batch if batch and not isinstance(batch, dict) else []
    self._verbose_log(
        f"Got batch with {len(batch_to_log)} items. The first items are:",
        batch_to_log[:5]
    )

    # Optionally, apply some logic to the response to determine whether
    # ingestion should continue or if should be short-circuited. By default
    # this will return True and ingestion continues.
    should_continue = self.get_should_continue(response_json)

    return batch, should_continue
```

This code isn't the nicest looking IMO. Is this what we want? Ideas on how to improve it?

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -195,6 +195,10 @@ def __init__(
 
         self.ingestion_errors: list[IngestionError] = []  # Keep track of skipped errors
 
+        self._should_verbose_log = self.dag_id in Variable.get(
+            "SHOULD_VERBOSE_LOG", default_var=[], deserialize_json=True
+        )
```
krysal, line 200: The name of the variable makes me think it should be a boolean. Do you have reasons to make `[]` the default?

```suggestion
        self._should_verbose_log = self.dag_id in Variable.get(
            "SHOULD_VERBOSE_LOG", default_var=False
        )
```

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
```diff
@@ -512,3 +527,10 @@ def _commit_records(self) -> int:
             total += store.commit()
         logger.info(f"Committed {total} records")
         return total
+
+    def _verbose_log(self, msg: str, data=[]):
```
krysal, line 531: `list` in Python is mutable, so [it's tricky to use them as default for params](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). We want to avoid confusing situations so it's better to make the default here `None` (or remove the default).
```suggestion
    def _verbose_log(self, msg: str, data: list = None):
```

krysal (commented):
Nico, I'm converting this to a draft. Please let us know when this is ready for review. Thanks!

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
nicoepp: Yes, @AetherUnbound told me at the PyCascades sprint she wanted to be able to activate the verbose logging per DAG (i.e. dag_id). This is the name we came up with. What naming would you propose?

sarayourfriend: @krysal are you able to answer @nicoepp's question here?

krysal: @nicoepp That's fine then! I wasn't sure how it would be used, but from what you say it makes sense.

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
nicoepp: @stacimc I pushed a commit now that tries to address this with code that is a bit nicer looking. E.g. for the exceptional case where batch is a `dict` in the tests I thought it was better to do the handling of that and the batch slicing inside the `_verbose_log` method. So, I added an optional param to indicate if you want to slice the batch and by how much before you print it. 

Can you take a look and tell what you think?

catalog/dags/providers/provider_api_scripts/provider_data_ingester.py
nicoepp: I pushed code that does all this in one log statement now

AetherUnbound:
I've made a few changes to this PR but I think it's ready to undraft:
- I changed the setup behavior so the ingester _always_ logs verbosely locally
- I changed the places where the verbose logging was outputting dictionaries to have it use `json.dumps` so we can copy those lines and parse them as JSON easily (for future runs, analysis, etc)

sarayourfriend:
@AetherUnbound just a heads up some of the tests are failing. It looks like maybe in some cases we're expecting the logging to always be a single parsable JSON document? I haven't looked deeply at it, but just a guess based on the changes you described and the test logs. I can look more into this if you'd like help sorting this PR out, just let me know.

AetherUnbound:
Ahh, thanks for the heads up @sarayourfriend - I'll fix the tests tomorrow!

stacimc (approved):
LGTM! I still think it would be nice for there to be _some_ way of disabling this locally with the variables, but that's absolutely not a blocker. Something we can adjust later if it does end up becoming necessary :)

sarayourfriend (approved):
:+1:

PR #4069, opened 2024-04-08
Update print_ps() to use Docker Compose ps JSON output

### LINKED ISSUE #4045: Simplify `just ps` script using compose v2's `docker compose ps` JSON output
## Current Situation

The Python script that `just ps` uses was written to be compatible with Docker compose v1. That meant it could not take advantage of compose v2's JSON output from `docker compose ps`, which added some complexity to the script.

https://github.com/WordPress/openverse/blob/bdb8f0ad1dd0b2ae504f708b4105fdadc10e8988/utilities/ps.py

This issue is open to community contributors. In order to complete this issue, you should have an intermediate understanding of Python, and basic familiarity with Docker compose. You must also have the entire Openverse docker compose stack running, for which you should follow https://docs.openverse.org/general/general_setup.html and the setup guides for each of the API, frontend, catalog, and ingestion server. You will need this to test the output of `just ps`. Please include sample output of the modified script in the PR description.

## Suggested Improvement

Now that [we've moved to compose v2](https://github.com/WordPress/openverse/issues/3876), we should clean up this script, simplifying it and deleting code that parsed the complex table output. Check out https://github.com/WordPress/openverse/pull/1160 where we originally introduced the script for background on its current implementation. 

This function in particular should be much simpler using `--format json` output.

https://github.com/WordPress/openverse/blob/bdb8f0ad1dd0b2ae504f708b4105fdadc10e8988/utilities/ps.py#L66-L92

Use `just dc ps --format json` instead of `docker ps` (no compose). That will also eliminate the need to filter services using the network name heuristic.

## Benefit

Remove unnecessary code, simplify the `just ps` script. The best code is no code.

zaharoian:
## Fixes

Fixes #4045  by @sarayourfriend 

## Description

Instead of using `docker ps` and creating a custom tabulated table, `docker compose ps --format json` is used to parse the output.

### Output Variations
There are slight adjustments with the `docker compose ps` returned format:
- Containers were originally sorted by recency, they are now returned in alphabetical order
- Container image names originally excluded `docker.io/` (Docker's public repository) but included 3rd party hostnames, now each container name contains hostnames unless it is a `openverse` generated container.

### Current output:
Screenshot 2024-04-08 at 2 47 07 PM

### Output with changes:
Screenshot 2024-04-08 at 2 48 12 PM

## Testing Instructions
1. Checkout this PR and run `just up`.
2. See the list of services being printed automatically after `up` is complete.
3. Run `just ps`.
4. See the list of services

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/utilities/ps.py
+++ b/utilities/ps.py
@@ -81,7 +81,7 @@ def parse_ps() -> list[Service]:
         for port_config in container_ps["Publishers"]:
             published_port = port_config["PublishedPort"]
             if published_port:
-                bindings.add((port_config["TargetPort"], published_port))
+                bindings.add((published_port, port_config["TargetPort"]))
         if bindings:
             services.append(Service(service_name, image_name, bindings))
 
--- a/utilities/ps.py
+++ b/utilities/ps.py
@@ -1,3 +1,4 @@
+import json
 import subprocess
 import traceback
 from dataclasses import dataclass
@@ -27,20 +28,19 @@ class Service:
 
 def get_ps() -> list[str]:
     """
-    Invoke Docker "ps" with a filter to get the configuration for all services. The
-    info is returned as a list of plain text lines.
+    Invoke Docker Compose "ps" with the JSON format flag and
+    return a list of lines of JSON output.
 
     :return: the output printed by the subprocess to STDOUT
     """
 
     proc = subprocess.run(
         [
-            "docker",
+            "just",
+            "dc",
             "ps",
-            "--filter",
-            "network=openverse_default",
             "--format",
-            "{{.Names}}\t{{.Image}}\t{{.Ports}}\n",
+            "json",
         ],
         capture_output=True,
         text=True,
@@ -54,8 +54,9 @@ def get_ps() -> list[str]:
 
 def parse_ps() -> list[Service]:
     """
-    Convert the output lines given by Docker "ps" into a list of services and
-    their port mappings.
+    Convert the output lines given by Docker Compose "ps" into a list of
+    services and their port mappings.
+
     :return: a list of running services with their port
     """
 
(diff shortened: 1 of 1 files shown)


utilities/ps.py
```diff
@@ -64,32 +67,23 @@ def parse_ps() -> list[Service]:
...
-                continue
-            # Ports are of the form 0.0.0.0:50230->3000/tcp or :::50230->3000/tcp
-            host_port, container_port = (
-                port_config.split(":")[-1].split("/")[0].split("->")
-            )
-            bindings.add((host_port, container_port))
+        for port_config in container_ps["Publishers"]:
+            published_port = port_config["PublishedPort"]
+            if published_port:
+                bindings.add((port_config["TargetPort"], published_port))
```
sarayourfriend, line 84: This is reversed, we want the host port with the host address, with the target port pointed to.

Instead of:

```
cache (docker.io/redis):
-  http://0.0.0.0:6379  (→ 50263)
```

We want:

```
cache (docker.io/redis):
-  http://0.0.0.0:50263 (→ 6379)
```

For that, just reverse the two entries in the tuple on this line. This works perfectly for me locally.

```suggestion
                bindings.add((published_port, port_config["TargetPort"]))
```

sarayourfriend (changes requested):
LGTM other than one small issue. Works perfectly otherwise, nice work and thanks for the contribution :pray:

sarayourfriend (approved):
Thanks again!@

utilities/ps.py
```diff
@@ -27,20 +30,19 @@ def print(self):
...
+    return a list of lines of JSON output.
 
     :return: the output printed by the subprocess to STDOUT
     """
 
     proc = subprocess.run(
         [
-            "docker",
+            "just",
+            "dc",
```
dhruvkb, line 42: This creates a dependency on `just` that's not present in the original script. I don't mind, because `just` must be present to invoke the `ps` recipe anyway.
```suggestion
            "docker",
            "compose",
```

dhruvkb (approved):
LGTM! Thanks for the contribution @zaharoian!

utilities/ps.py
sarayourfriend: In what meaningful circumstance would `just` not be present?

If you don't use `just dc` you'll also need to pass the compose file and setup the environment arguments that `just` sets. Directly using `docker compose` with our stack don't work so well, in my experience.

If there's a real situation in which the `ps` script is used and `just` wouldn't be present, then sure, but I can't think of one :shrug: We rely on `just` for the most basic operations of Openverse development and it's a hard, non-negotiable requirement for working in our codebase, on the level of having system Python and bash available.

sarayourfriend: I'm going to merge with the two approvals now, but @dhruvkb feel free to open a follow-up issue/PR if you think avoiding `just` is a must for this script.

sarayourfriend:
Thanks again for the contribution @zaharoian!

PR #4072, opened 2024-04-08
Do not reproduce reingestion workflow documentation

### LINKED ISSUE #3207: Do not produce reingestion workflow documentation if it matches regular workflow documentation
## Description

Presently, our [DAG documentation generation](https://github.com/WordPress/openverse/blob/efbe123d9c860dab1c46878b04b320acb5fbb28d/catalog/utilities/dag_doc_gen/dag_doc_generation.py) script will generate documentation for all DAGs that define a `doc_md` on the DAG:

https://github.com/WordPress/openverse/blob/efbe123d9c860dab1c46878b04b320acb5fbb28d/catalog/utilities/dag_doc_gen/dag_doc_generation.py#L122

For some DAGs (namely reingestion workflows, where we reuse the same DAG machinery over different windows), this produces the exact same documentation multiple times. This is unnecessarily redundant, here's an example from Flickr:

- https://docs.openverse.org/catalog/reference/DAGs.html#flickr-workflow
- https://docs.openverse.org/catalog/reference/DAGs.html#flickr-reingestion-workflow

We should record the DAG doc text while iterating over the DAGs and skip producing documentation for any DAGs that have already matched a previous example exactly.

The easiest way to do this might be to record the doc markdown in a mapping from `doc markdown: dag ID`, then check for the doc markdown's presence in that mapping before producing the string. This is complicated by the fact that we want the original workflow (not the reingestion one) to be the DAG which receives the documentation, not the reingestion workflow.

We could also instead add some logic when coming across DAGs with `reingestion` in the DAG ID to find and check the original DAG and see if the docstrings match. If they don't match, then continue with processing as normal, otherwise skip.

mattfergoda:
## Fixes

Fixes #3207 by @AetherUnbound 

## Description

`dag_doc_generation.py` no longer produces documentation for DAGs with markdown that is identical to what it has seen before (iterating in alphabetical order by type). For example, "Audio" DAGs have docs generated, "Image" DAGs do not since they come after "Audio" in the alphabet. Same for "Prod" vs. "Staging" and "Provider" vs. "Provider Reingestion." In the `# Provider` section, it will still generate a row for these duplicate DAG docs, but will link to the common doc.

## Testing Instructions

All tests pass.  
Run the documentation server locally and make sure the updated DAG documentation makes sense.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/utilities/dag_doc_gen/dag_doc_generation.py
+++ b/catalog/utilities/dag_doc_gen/dag_doc_generation.py
@@ -146,7 +146,10 @@ def get_dags_info(dags: DagMapping) -> list[DagInfo]:
 
 
 def generate_type_subsection(
-    name: str, dags_info: list[DagInfo], is_provider: bool
+    name: str, 
+    dags_info: list[DagInfo], 
+    is_provider: bool, 
+    dag_by_doc_md: dict[str, str]
 ) -> str:
     """Generate the documentation for a "DAGs by type" subsection."""
     log.info(f"Building subsection for '{name}'")
@@ -171,7 +174,7 @@ def generate_type_subsection(
         # If we have documentation for the DAG, we'll want to link to it within the
         # markdown, so we reference it using the heading text (the DAG ID)
         if dag.doc:
-            dag_id = f"[{dag_id}](#{dag.dag_id})"
+            dag_id = f"[{dag_id}](#{dag_by_doc_md[dag.doc]})"
         text += f"\n| {dag_id} | `{dag.schedule}` |"
         if is_provider:
             text += f" `{dag.dated}` | {', '.join(dag.provider_workflow.media_types)} |"
@@ -210,7 +213,11 @@ def generate_dag_doc(dag_folder: Path = DAG_FOLDER) -> str:
     for dag in dags_info:
         dags_by_type[dag.type_].append(dag)
 
+    dag_by_doc_md: dict[str, str] = {}
     for type_, dags in sorted(dags_by_type.items()):
+        for dag in dags:
+            if dag.doc not in dag_by_doc_md:
+                dag_by_doc_md[dag.doc] = dag.dag_id
         # Create a more human-readable name
         name = type_.replace("_", " ").replace("-", " ").title()
         # Special case for provider tables since they have extra information
@@ -219,7 +226,9 @@ def generate_dag_doc(dag_folder: Path = DAG_FOLDER) -> str:
         # sub-list as part of a table of contents, but defer adding the sub-lists until
         # all are generated.
         text += f" 1. [{name}](#{type_.replace('_', '-')})\n"
-        dag_types.append(generate_type_subsection(name, dags, is_provider))
+        dag_types.append(
+            generate_type_subsection(name, dags, is_provider, dag_by_doc_md)
+        )
 
     text += "\n" + "\n\n".join(dag_types)
 
@@ -232,8 +241,10 @@ def generate_dag_doc(dag_folder: Path = DAG_FOLDER) -> str:
         # Similar to the DAGs-by-type section, we add the reference to a table of
         # contents first, and then defer adding all the generated individual docs until
         # the very end.
-        text += f" 1. [`{dag.dag_id}`](#{dag.dag_id})\n"
-        dag_docs.append(generate_single_documentation(dag))
+        text += f" 1. [`{dag.dag_id}`](#{dag_by_doc_md[dag.doc]})\n"
+        # Only generate unique docs.
+        if dag.dag_id == dag_by_doc_md[dag.doc]:
+            dag_docs.append(generate_single_documentation(dag))
 
     text += "\n" + "".join(dag_docs)
 
--- a/catalog/utilities/dag_doc_gen/dag_doc_generation.py
+++ b/catalog/utilities/dag_doc_gen/dag_doc_generation.py
@@ -146,10 +146,10 @@ def get_dags_info(dags: DagMapping) -> list[DagInfo]:
 
 
 def generate_type_subsection(
-    name: str, 
-    dags_info: list[DagInfo], 
-    is_provider: bool, 
-    dag_by_doc_md: dict[str, str]
+    name: str,
+    dags_info: list[DagInfo],
+    is_provider: bool,
+    dag_by_doc_md: dict[str, str],
 ) -> str:
     """Generate the documentation for a "DAGs by type" subsection."""
     log.info(f"Building subsection for '{name}'")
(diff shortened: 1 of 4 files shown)


sarayourfriend (changes requested):
Thanks very much for this PR @mattfergoda, it's a very interesting problem you've chosen to tackle, and potentially a bit more complex than the issue may have conveyed.

```
[`image_data_refresh`](#audio_data_refresh)  
```

This line in the new generated DAG docs indicates to me that this approach will not work. The method you're using to track duplicated docs is great, but I think we need to do a little extra work to identify the "generic" name of two matching DAGs. `image_data_refresh` and `audio_data_refresh` should have a generic name something like `{image,audio}_data_refresh` or maybe just `data_refresh`.

Whatever format we want to choose, we can use the underscore-as-separator heuristic to find the segments, and identify which ones match. Here is a small example script with a process for doing so:

```py
dag_ids = [
    ("image_data_refresh", "audio_data_refresh"),
    ("recreate_image_popularity_calculation", "recreate_audio_popularity_calculation"),
    ("create_filtered_audio_index", "create_filtered_image_index"),
    ("staging_elasticsearch_cluster_healthcheck", "production_elasticsearch_cluster_healthcheck"),
]
for a, b in dag_ids:
    a_segments = a.split("_")
    b_segments = b.split("_")
    combined_name_segments = []
    for i, segment in enumerate(a_segments):
        if segment == b_segments[i]:
            combined_name_segments.append(segment)
        else:
            combined_name_segments.append(f"{{{a_segments[i]},{b_segments[i]}}}")

    print("_".join(combined_name_segments))
```

And the output:

```
{image,audio}_data_refresh
recreate_{image,audio}_popularity_calculation
create_filtered_{audio,image}_index
{staging,production}_elasticsearch_cluster_healthcheck
```

We'd need to keep a running list of all the matching IDs and instead of just ignoring subsequent ones, update the combined name of the DAG as we go.

We could also use an additional heuristic to replace things like `{image,audio}` with a reference to media type, so that it's resilient to future additional media types. Same with `{staging,production}`, that can be simplified to a reference to `deployment environment` or the like. **It could also be simpler to leave the shared segments out altogether, so that `create_filtered_{audio,image}_index` becomes `create_filtered_index`**. That is probably my preferred way to do it, it reads the best to me of these options.

Regardless of the approach taken, you'll need to make changes to existing references to the heading anchors in the documentation site (or elsewhere) so that those links don't end up broken.

We'd also need to build in the `reingestion` heuristic, which @AetherUnbound pointed out in the issue as a complicating factor, and isn't handled in the PR yet at all:

> This is complicated by the fact that we want the original workflow (not the reingestion one) to be the DAG which receives the documentation, not the reingestion workflow.

mattfergoda:
Thanks for the thorough feedback @sarayourfriend. It does sound like this will require a deeper dive. Given some other obligations I have at the moment, I'm not sure when I'll have a chance to get to this.

AetherUnbound:
@mattfergoda totally understandable! Would you be alright with us taking over on this to get it across the finish line?

mattfergoda:
@AetherUnbound of course not!

AetherUnbound:
@sarayourfriend I've added some logic for determining which single DAG documentation to reference for a given DAG, based on some heuristics we have on the names. I'm going to add a test specifically for that, but this should be ready to go!

catalog/utilities/dag_doc_gen/dag_doc_generation.py
```diff
@@ -57,6 +57,23 @@
...
+    "audio": "{media_type}",
+    "production": "{environment}",
+    "staging": "{environment}",
+}
+# DAG IDs to ignore when collapsing reference documentation for a mapped term
+DAG_IDS_TO_IGNORE_COLLAPSE = {
+    "recreate_full_staging_index",
+    "staging_database_restore",
+    "create_proportional_by_source_staging_index",
+}
```
sarayourfriend, line 73: Interesting! If we mapped DAG IDs all in one iterative go, rather than embedding the mapping into the existing loop, we could obviate the need for this specific listing. Basically, it seems like the criteria for this is list is whether a DAG, when its name is mapped to a generic name, actually shares a mapped name with any other DAG. If not, it's unique, and the name shouldn't have terms collapsed. Something like this pseudo code...

```python
mapped_id_for_dag_ids = dict()
dag_ids_for_mapped_id = dict(list)

for dag_id in all_dag_ids:
    mapped_id = determine_mapped_dag_id(dag_id)
    mapped_id_for_dag_ids[dag_id] = mapped_id
    dag_ids_for_mapped_id[mapped_id].append(dag_id)

for dag_id in all_dag_ids:
    # When determining if the docs should be merged
    mapped_id = mapped_id_for_dag_ids[dag_id]
    should_use_mapped_id = len(mapped_dag_ids[mapped_id]) > 1
```

If `dag_ids_for_mapped_id` only has one entry for the mapped DAG ID, then we'll just use the original name. Then we don't need to manually maintain this list of exceptions for environment or single media-type specific DAGs.

documentation/catalog/reference/DAGs.md
```diff
@@ -32,36 +32,36 @@ The following are DAGs grouped by their primary tag:
...
-| [`audio_data_refresh`](#audio_data_refresh)                   | `0 0 * * 1`       |
-| [`create_filtered_audio_index`](#create_filtered_audio_index) | `None`            |
-| [`create_filtered_image_index`](#create_filtered_image_index) | `None`            |
-| [`image_data_refresh`](#image_data_refresh)                   | `0 0 * * 1`       |
+| DAG ID                                                             | Schedule Interval |
+| ------------------------------------------------------------------ | ----------------- |
+| [`audio_data_refresh`](#media_type_data_refresh)                   | `0 0 * * 1`       |
+| [`create_filtered_audio_index`](#create_filtered_media_type_index) | `None`            |
+| [`create_filtered_image_index`](#create_filtered_media_type_index) | `None`            |
+| [`image_data_refresh`](#media_type_data_refresh)                   | `0 0 * * 1`       |
```
sarayourfriend, line 40: It'd be nice if these were sorted by the mapped DAG ID, followed by the original DAG ID, so that related DAGs are always together. I guess in this version the DAGs are grouped by media type, but the difference in ordering for the audio and image DAGs reads a bit lazy/careless to me (especially imagining not having the context of this document being automatically generated).

sarayourfriend (approved):
Awesome! I've left two rather nit-picky comments that I think should be addressed, but can be done as follow up issues if you prefer. If that is the case, please create issues for them.

catalog/utilities/dag_doc_gen/dag_doc_generation.py
AetherUnbound: I tried playing around with this a bit, and there's not an approach that I like that doesn't involve a larger refactoring of the code. I think we'll stick with the current approach because there are only a few exclusions, and this documentation is run as part of our CI/CD anyway (so we'll know when the doc is generated whether this needs to be adjusted for new DAGs).

AetherUnbound (approved):
Although I collaborated on the code, I think things look good and I'd like to get this merged! 😄

Issue #4101, opened 2024-04-12
Stocksnap 403 forbidden error during ingestion


AetherUnbound:
## Airflow log link

> **Note**: _Airflow is currently only accessible to maintainers & those given
> access. If you would like access to Airflow, please reach out to a member of
> @WordPress/openverse-maintainers_.

https://airflow.openverse.org/log?execution_date=2024-03-01T00%3A00%3A00%2B00%3A00&task_id=ingest_data.pull_image_data&dag_id=stocksnap_workflow&map_index=-1

## Description

The Stocksnap DAG encountered an error during ingestion:

```
[2024-04-01, 23:09:02 UTC] {requester.py:85} ERROR - Error with the request for URL: https://cdn.stocksnap.io/img-thumbs/960w/LZITOLMWL6.jpg
[2024-04-01, 23:09:02 UTC] {requester.py:86} INFO - HTTPError: 403 Client Error: Forbidden for url: https://cdn.stocksnap.io/img-thumbs/960w/LZITOLMWL6.jpg
[2024-04-01, 23:09:02 UTC] {requester.py:89} INFO - Using headers {'User-Agent': 'Openverse/0.1 (https://openverse.org; openverse@wordpress.org)', 'Accept': 'application/json'}
[2024-04-01, 23:09:02 UTC] {media.py:233} INFO - Writing 68 lines from buffer to disk.
[2024-04-01, 23:09:02 UTC] {provider_data_ingester.py:513} INFO - Committed 31168 records
[2024-04-01, 23:09:02 UTC] {taskinstance.py:2728} ERROR - Task failed with exception
providers.provider_api_scripts.provider_data_ingester.IngestionError: 403 Client Error: Forbidden for url: https://cdn.stocksnap.io/img-thumbs/960w/LZITOLMWL6.jpg
query_params: {}
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 439, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
    return execute_callable(context=context, **execute_callable_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 200, in execute
    return_value = self.execute_callable()
[...]

PR #4102, opened 2024-04-12
Fill out query params for Stocksnap DAG, allow restarts

AetherUnbound:
## Fixes

Related to #4101

## Description

This PR adds some machinery around the `query_params` handling for the Stocksnap DAG so that we can leverage the options we already have available for query parameters for manipulating the page(s) that the Stocksnap DAG operates on. This doesn't address the error described in the above issue, but it should make it much easier to reproduce (since we should now receive a specific page for where the run failed rather than having to wait 9 hours!).

## Testing Instructions

1. Run the Stocksnap DAG locally, with a sensible ingestion limit. This should behave normally. Use `just catalog/pgcli` and `SELECT count(*) from images;` to verify that the ingestion limit records were loaded.
1. Trigger a new Stocksnap run with the `initial_query_params` set to `{"page": 100}`. Use the same commands as above to verify that there are now twice as many records (e.g. new data and not the same page was loaded).
1. Add a `raise ValueError("Whoops")` in the `get_record_data` function of the Stocksnap ingester, and check that the error logs show which page failed.
1. (Optional) Use the query params list (e.g. `[{"page": 5}, {"page": 10}]`) to trigger a new DAG to only run over a few different sets of parameters.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
@@ -39,7 +39,7 @@ def test_get_media_type():
 
 
 def test_endpoint_with_initialized_page_counter():
-    expect_result = "https://stocksnap.io/api/load-photos/date/desc/0"
+    expect_result = "https://stocksnap.io/api/load-photos/date/desc/1"
     actual_result = stocksnap.endpoint
     assert expect_result == actual_result
 
@@ -123,7 +123,9 @@ def test_endpoint_increment():
 
 def test_initial_params_respected():
     stocksnap = StockSnapDataIngester(conf={"initial_query_params": {"page": 5}})
+    query_params = stocksnap._get_query_params(None)
     assert stocksnap._page_counter == 5
+    assert query_params == {"page": 5}
 
 
 def test_get_record_data_returns_none_when_media_data_none():
--- a/catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
@@ -114,11 +114,16 @@ def test_process_batch(filesize_mock):
 
 
 def test_endpoint_increment():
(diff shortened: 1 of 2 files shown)


obulat (approved):
LGTM!

catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
```diff
@@ -39,7 +39,7 @@ def test_get_media_type():
 
 
 def test_endpoint_with_initialized_page_counter():
-    expect_result = "https://stocksnap.io/api/load-photos/date/desc/0"
+    expect_result = "https://stocksnap.io/api/load-photos/date/desc/1"
```
sarayourfriend, line 42: This is surpising... does this difference matter? Is page 0 a meaningful page that we need to retrieve, different from page 1?

sarayourfriend (approved):
Requesting clarification on the changed test; approving in case page 0 is meaningless and we are intentionally changing this to skip that page now.

catalog/tests/dags/providers/provider_api_scripts/test_stocksnap.py
AetherUnbound: So this is actually a more semantically appropriate tests - the way the class was set up previously is that it would produce `0` if that attribute was accessed directly after construction, but when it was _actually_ called it would be after the first `get_next_query_parameters`, which would always increment the value and set it to `1`. We want page `1` and need to start there, this change just more appropriately reflects that as the starting page (rather than relying on the expected ingestion flow which would increment it before actually calling the endpoint).

sarayourfriend: Great!

Issue #286, opened 2022-08-04
Implementation Plan: Switch Python package management away from Pipenv


sarayourfriend:
## Problem

Pipenv was a good start, but it has some issues that are difficult to work around:

1. It is slow to lock and resolve versions
2. It is difficult (maybe impossible) to atomically update dependencies on an individual basis

## Description

It would be great to have a concise RFC to investigate what it would take to switch from Pipenv to Poetry or PDM. The RFC should make a recommendation for one or the other and describe the trade-offs between them.

Please evaluate the effort required for such in the [API](https://github.com/WordPress/openverse-api) and [openverse "meta"](https://github.com/WordPress/openverse) repositories.

Please also evaluate the effect on any Dockerfiles and other deployment artifacts.

## Additional context

WordPress Make Slack conversation here: https://wordpress.slack.com/archives/C02012JB00N/p1659363874533649

You can register for a Make Slack account here: https://chat.wordpress.org

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
I've just learned about `pip-tools` recently and it seems like it could be a very good option: https://github.com/jazzband/pip-tools. [pypi itself uses it, for example](https://github.com/pypi/warehouse/pull/11912).

It follows existing Python conventions closely, so may be more usable for the catalogue (which relies on requirements.txt files in a way that it can't use Pipenv for now). However, it's just some tools around pip, so doesn't manage virtual environments. That's an advantage Poetry and PDM have (and Pipenv as well), the venv management is transparent and you don't have to think about it.

I'd be nice to document the trade offs here.

dhruvkb:
We have not prioritised this so far, which makes sense because of the large scale disruption to at least two parts of the stack (the API and ingestion server) and many other smaller sub-projects. However, two developments suggest we should be considering it with increased priority.

- Pipenv has been a major component of the cause that led to service disruption because of its tendency to update unrelated packages during an update. It has been noted that other package managers (notably Poetry) are not subject to this problem.

- The ingestion server is soon on its way out #3925, which will leave the API as the only Pipenv based stack in the repository. Other smaller Python sub-projects (which are not critical to Openverse services) can switch to a different project manager to match the API or stick to Pipenv which serves them well.

dhruvkb:
[...]

PR #4107, opened 2024-04-14
Switch API package management from Pipenv to PDM

dhruvkb:
## Related

Related to #286. Additional background can be read in [this comment](https://github.com/WordPress/openverse/issues/286#issuecomment-2053604034).

## Description

This PR replaces Pipenv with PDM. While PDM has monorepo support, we have opted not to use this feature in this PR. Instead we will use a path dependency to depend on subpackages located at `packages/` in the monorepo. To see this in action, the license and attribution related functionality in the API has been separated into its own package at `packages/ov-attribution`.

The `ov-attribution` subpackage is listed as a dependency as well as an editable dev-dependency of the API. This means that in dev mode, the package will not actually be copied into the virtualenv but rather only point to `../packages/ov-attribution`. When installed with `--no-editable` it will be copied into the virtualenv. Thanks @sarayourfriend for suggesting this and making usage of PDM possible.

## Testing Instructions

1. Install PDM. I recommend using the `pipx` approach, I am too.
2. Build the new Docker image with `just dc build web`.
3. Explore the editable local package by changing something in `py-packages/openverse-attribution` to see the change immediately in the API.
6. Run the tests. (Or don't, CI has you covered.)
   - Test `openverse-attribution` with:
   ```
   cd py-packages/openverse-attribution && pdm install && pdm run pytest`
   ```
   - Test the API with `just api/test`.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
I'm testing this out locally and tried the instructions - I made some changes to `packages/ov-attribution/ov_attribution/attribution.py` and the API did not reload 🤔 In fact, it took rebuilding the web image to get the changes to appear. Is that intended?

dhruvkb:
It should have reflected immediately because the `ov-attribution` package is mounted into the API image during development so changes should reflect immediately. I suspect it's because the Django API is not reloading when the package changes because it's not watching it and not automatically reloading. I'll try to fix it.

sarayourfriend (commented):
Just suggestions so far, I haven't done a deep dive yet.

One other nit, is I personally think it's a bit nicer to use the full `openverse` name in package names, rather than abbreviate to `ov`, which is a somewhat ambiguous abbreviation we only rarely use, and only internally as far as I know... unlike e.g., WP for WordPress, which has both the benefit of the name capitilising both letters and being pretty old and well recognised. `ov` on the other hand, already has a fair number of references on PyPI: https://pypi.org/search/?q=ov-&o=. `openverse`, has exactly one package named with it :slightly_smiling_face: https://pypi.org/search/?q=openverse

Just a nit, though, and totally changeable at any future date before we publish the package anyhow, so no need to spend time on it for this PR. There are much more exciting and interesting things to think about!

One other thing that comes to mind is whether overloading `packages` as both Python and JavaScript packages is a good idea. I wonder particularly whether we couldn't have two separate directories for JS and Python packages instead? That way we can keep the nice glob patterns, and there's no risk of easily confusing what's going on? I can't think of any good names that don't make me cringe from their clumsiness (`python-packages`, `javascript-packages`, verbose, clear, but ugh?).

By the way, take a look at the differences between src and flat layout in the PyPA docs: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/#src-layout-vs-flat-layout (only relevant for packages, not for the API).

Using the src layout comes with some pretty nice benefits, especially if we're cross-referencing the package locally (namely, making sure only the stuff that's allowed to be imported is imported and that the package used in tests is the same one users will install).

[PDM's build backend auto-detects the src layout, so it doesn't require any additional configuration](https://backend.pdm-project.org/build_config/#the-src-layout).

dhruvkb:
I feel like the `openverse_` prefix is too verbose. But doable.

Maybe `py_packages` and `js_packages`? They're shorter than full names and equal in length to each other. But no solid opinions from me on that (a terrible set of names would be `🐍_packages/` and `🟨_packages/`; this is probably the only set I would strongly oppose haha). I wanted to keep the number of root folders in the repo to a minimum but I agree that Python and JS packages should not be in one folder as that breeds confusion. 

I'll also look into the `src/` layout. I felt nice about the symmetry of `api/api` and `ov-attribution/ov_attribution` but if the `src/` layout is better, we'll use that 👍.

api/justfile
```diff
@@ -14,8 +14,8 @@ NO_COLOR := "\\033[0m"
 # Version #
 ###########
 
-export API_PY_VERSION := `grep 'python_version' Pipfile | awk -F' = ' '{print $2}' | sed 's/"//g'`
-export PGCLI_VERSION := `grep 'pgcli' Pipfile | awk -F' = ' '{print $2}' | sed 's/\~//g; s/"//g; s/=//g'`
+export API_PY_VERSION := `grep 'requires-python' pyproject.toml | awk -F'"' '{print $2}' | sed 's/^==//g; s/\.\*$//g'`
```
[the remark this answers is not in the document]
dhruvkb: @sarayourfriend we don't necessarily require PDM to be installed on the host machine, epecially for API development. That's why these commands only use standard tools.

AFAIK GNU `sed` differs from BSD `sed` when the `-i` flag is involved. This usage is cross-platform.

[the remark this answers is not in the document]
sarayourfriend: Gotcha! Might be worth leaving a comment in the justfile then.

sarayourfriend:
> Maybe py_packages and js_packages

Works for me!

.github/renovate.json
```diff
@@ -73,7 +73,8 @@
         "api/**",
         "catalog/**",
         "ingestion_server/**",
-        "documentation/**"
+        "documentation/**",
+        "py_packages/openverse-attribution/**"
```
sarayourfriend, line 77: This is 100% a complete and total nit and personal preference but the asymmetry of `py_packages/openverse-attribution` (and similarly `js_packages/eslint-plugin`) really bugs me (for no good technical reason I could possible reasonably argue for, I acknowledge LOL!)

```suggestion
        "py-packages/openverse-attribution/**"
```

sarayourfriend (commented):
This is so exciting! Really looking forward to a world without pipenv, however we get there.

.github/renovate.json
dhruvkb: Oooh good eye. It's definitely not a personal preference, a mix of underscores and hyphens reads quite badly. Now that you've pointed it out, I would also be completely bothered by it all the time.

AetherUnbound:
@dhruvkb should this be drafted while we review #4128? I'm a bit confused at which I should look at first?

dhruvkb:
@AetherUnbound I undrafted it because I felt quite confident that PDM would be our pick and also so I could also receive feedback on the actual implementation of the migration like that from @sarayourfriend regarding using an `src/` layout and naming the dirs `py-packages`.

Now that #4128 is merged, feel free to review this PR as deeply as you like. Thanks!

api/Dockerfile
```diff
@@ -26,23 +26,25 @@ ENV PYTHONUNBUFFERED=1
...
+# - Install PDM to install Python dependencies
 RUN apt-get update \
   && apt-get install -y python3-dev git \
   && rm -rf /var/lib/apt/lists/* \
-  && python -m venv /venv \
-  && pip install --upgrade pipenv
+  && pip install pdm~=2.14
+
+# Copy subpackages from additional build-context 'packages'
+COPY --from=packages openverse-attribution ./py-packages/openverse-attribution
```
AetherUnbound, line 38: I wonder here if there's a way for us to "wildcard" this 🤔 As in, not have to be explicit about every package.

api/pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+description = "API for the Openverse project"
+readme = "README.md"
+license = {text = "MIT"}
+authors = [
+  {name = "Openverse", email = "openverse@wordpress.org" },
+]
+
+requires-python = "==3.11.*"
+dependencies = [
+  "adrf >=0.1.4, <0.2",
```
AetherUnbound, line 13: I'm surprised that we're unable to use the `~=` notation for these versions 😮 I had assumed we'd be able to!

Also, is the fact that these aren't exact problematic? I know we'll have the lock file, is that what Renovate/dependabot will update directly?

api/pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+# These packages override their main counterparts with an editable installation.
+overrides = [
+  "-e openverse-attribution @ file:///${PROJECT_ROOT}/../py-packages/openverse-attribution"
+]
+dev = [
+  "ipython >=8.22.1, <9",
+  "pgcli >=3.5.0, <4",
+  "remote-pdb >=2.1.0, <3",
+]
+test = [
```
AetherUnbound, line 55: Is there a functional difference between `dev` and `test` here, or is the separation only semantic?

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
```diff
@@ -0,0 +1,62 @@
+import re
+
+from openverse_attribution.license import License
+
+
+def get_attribution_text(
```
AetherUnbound, line 6: FWIW I think this function should _also_ support the `License` enum as an argument, rather than potentially having to re-enumerate all of the other fields.

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
```diff
@@ -0,0 +1,62 @@
...
+    :param license_version: the version of the license, if known
+    :param license_url: the URL to the license, to override the default
+    :return: the plain-text English language attribution
+    """
+
+    lic = License(license_slug)
+
+    title = f'"{title}"' if title else "This work"
+
+    attribution_template = "{title} {creator} {marked-licensed} {license}. {view-legal}"
```
AetherUnbound, line 37: Nit: This could probably be a module-level constant.

AetherUnbound (commented):
This is great! I have a few comments/questions, and per the discussion in #4165, I think we should have `packages/python/` instead of `py-packages/` at the top level (to mirror what we have in `automation/`).

api/pyproject.toml
dhruvkb: This allows us to selectively build smaller and smaller images from dev (all deps) to CI (main + test) to prod (main only).

api/Dockerfile
dhruvkb: It is possible to make this a wildcard, I only didn't do that because we have only one package right now (and that too, is more of a PoC than a real library worth distributing).

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
dhruvkb: I did not get what you mean by this. Did you mean this diff to the function parameters?

```diff
- license_slug: str,
+ license: License,
```

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
dhruvkb: That's true, but since this is very implementation specific, I did not want it to be exposed to the outside world and it's much cleaner imo if the module only exports one item: function `get_attribution_text`.

api/pyproject.toml
dhruvkb: Renovate and Dependabot usually update lockfiles when a dependency is updated. They might/might not adjust the lower bound of the values in the `pyproject.toml` as well (they sometimes updated `Pipfile`, sometimes not), I do not know.

`~=` is supported (afaik PDM supports [all specifiers from PEP 440](https://peps.python.org/pep-0440/#version-specifiers)) I felt that the behaviour of `~=` is confusing because it changes depending on how many parts are specified in the version.

Consider `adrf`. Here `~=0.1.4` would be equivalent to `>=0.1.4,==0.1.*` which is what we want (v0 means that there could be breaking changes even on minor versions) so all good there. Then consider `aiohttp`. Here `~=3.9.3` would be equivalent to `>=3.9.3,==3.9.*` which is not what we want (for post v1, we want minor updates too), so that would need us to write `~=3.9` which then means we lose the lower bound as that is equivalent to `>=3.9,==3.*`.

Also personally `>=,<` is very easy to understand as it corresponds to my understanding of left-inclusive, right-exclusive ranges.

Again no strong thoughts, happy to use `~=` if favoured by the team.

obulat (changes requested):
Awesome! Can't wait to see the PRs that Renovate will open.

I tried changing the attribution package code, and it immediately showed up in the API response.

My request for change is to add the testing of the `openverse-attribution` to the CI tests.

dhruvkb:
@obulat to keep this PR small (already 35 files changed) I made a separate issue to cover CI changes #4167.

obulat (approved):
Thank you for all the work on selecting the package manager and implementing the changes, @dhruvkb . Makes sense to split the changes into several PRs.
Can we create a milestone for all the pdm-related accompanying issues?

dhruvkb:
@obulat good call, [done](https://github.com/WordPress/openverse/milestone/29).

sarayourfriend:
> I think we should have packages/python/ instead of py-packages/ at the top level (to mirror what we have in automation/).

I love this, Madison! :100:

api/pyproject.toml
AetherUnbound: Gotcha! I hadn't seen the test install happening anywhere, so I wasn't sure.

api/pyproject.toml
AetherUnbound: Ahh, that makes a lot of sense. Thanks for the explanation, I'm definitely okay with the more explicit approach you've provided here 🙂

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
AetherUnbound: Yes, potentially something like `license: License | str` as the first parameter was what I was thinking! It would be a bit of overloading, but it feels odd not being able to use the library's enum for this function.

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
AetherUnbound: That's understandable, but we could still make it module-level without necessarily exposing it (e.g. use an underscore as the prefix, like `_ATTRIBUTION_TEMPLATE`). That way it's clear for the library developers without necessarily being an exported attribute.

api/Dockerfile
```diff
@@ -26,23 +26,25 @@ ENV PYTHONUNBUFFERED=1
...
 # - Install system packages needed for building Python dependencies
 # - Create a virtualenv inside `/venv`
-# - Install Pipenv to install Python dependencies
+# - Install PDM to install Python dependencies
 RUN apt-get update \
   && apt-get install -y python3-dev git \
   && rm -rf /var/lib/apt/lists/* \
-  && python -m venv /venv \
-  && pip install --upgrade pipenv
+  && pip install pdm~=2.14
```
krysal, line 35: Is there a reason the version should be manually specified here? `2.15.0` is already out for example.

api/Dockerfile
```diff
@@ -26,23 +26,25 @@ ENV PYTHONUNBUFFERED=1
...
+# Copy subpackages from additional build-context 'packages'
+COPY --from=packages openverse-attribution ./py-packages/openverse-attribution
+
+# Copy the Python project manifest and PDM lockfile into the container
+COPY pyproject.toml pdm.lock ./
 
-# Copy the Pipenv files into the container
-COPY Pipfile Pipfile.lock ./
+# Pass additional arguments when installing Python packages with PDM
+ARG PDM_INSTALL_ARGS
```
krysal, line 44: Nit:
```suggestion
ARG PDM_INSTALL_ARGS='--no-editable'
```

api/Dockerfile
```diff
@@ -26,23 +26,25 @@ ENV PYTHONUNBUFFERED=1
...
 
-# Install Python dependencies system-wide (uses the active virtualenv)
-RUN pipenv install --system --deploy --dev
+# Install Python dependencies into a new virtualenv
+RUN pdm install --check --frozen-lockfile ${PDM_INSTALL_ARGS:-'--no-editable'}
```
krysal, line 47: Nit (cont)
```suggestion
RUN pdm install --check --frozen-lockfile ${PDM_INSTALL_ARGS}
```

krysal (commented):
This is exciting! The transition looks smoother than I anticipated ✨  I haven't reviewed all the PR, but I noticed some things I want to comment on.

api/Dockerfile
sarayourfriend: Best not to wildcard it either, if we can avoid it; it would be nice for it to be "automatic" if new packages are added (I'm sure they will be, eh?) but we can look into ways to derive it from the pyproject.toml rather than a catch-all wild card. There might eventually be dev dependencies in the packages, or packages only used by catalog and ingestion worker (for example) that we would not want copied into the context of the API container (and vice-versa re ingestion workers).

api/Dockerfile
```diff
@@ -96,6 +98,7 @@ RUN useradd --create-home opener \
 USER opener
 
 # Copy code into the final image
+COPY --chown=opener --from=packages openverse-attribution /py-packages/openverse-attribution/
```
sarayourfriend, line 101: This shouldn't be necessary, I thought, because it should be copied into the `.venv` for production (non-editable). Unless this is explicitly to support the development-mode of this target? I wish it were possible to conditionally copy, if that is the reason.

Bit of a nit, though, we can iterate on this without issue.

Edit (after reading the compose file again): Even in development mode, it shouldn't be necessary, because we map the python packages directory in the compose file, so the editable path reference should still be valid.

api/api/models/media.py
```diff
@@ -8,12 +8,12 @@
 from django.utils.html import format_html
 
 from elasticsearch import Elasticsearch, NotFoundError
+from openverse_attribution.attribution import get_attribution_text
```
sarayourfriend, line 11: I know we've gone back and forth on the name, but it ocurred to me this doesn't even need openverse specified in the package name. Nothing about it is actually Openverse specific. It is, however, CC-specific! `cc_attribution` would be a good name for it.

Just a nit though, don't bother spending time on this, we can bikeshed the name if we ever get around to publishing it.

api/pyproject.toml
sarayourfriend: Agreed `>=` is a lot clearer. `~=` always struck me as a bit weird and annoying to have to think so hard about every time I read it.

Most projects I've seen using pyproject use `>=` or `==`.

documentation/packages/openverse_attribution/index.md
```diff
@@ -0,0 +1,79 @@
+# `openverse-attribution`
```
sarayourfriend, line 1: Actually, occurring to me that it'd be important for package README's to be in the package itself, maybe just copied to documentation during the build step, otherwise it isn't available for inclusion in the build output of the package. npm and pypi require a README (`twine check` will fail on build outputs that don't have it, and you won't be allowed to upload it).

Not an issue for this PR, but something we will need to address in the future when we publish packages from the packages directories.

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
sarayourfriend: +1 to your point, Madison.

However, let's please create an issue to sort these implementation details out elsewhere. There are a number of things we'd want to make sure are nailed down tight about this library before it was used outside our project, and it would be good to have this PR not focused on this, but the package management side of things specifically.

The main point of including it in this PR, as far as I know, is to prove the concept of shareable monorepo Python packages, not to define the actual official and publicly available API for a published attribution generation library.

py-packages/openverse-attribution/src/openverse_attribution/attribution.py
sarayourfriend: My comment in this thread is also relevant here: https://github.com/WordPress/openverse/pull/4107/files#r1574945150

We can sort these details out later, they shouldn't block this PR, the implementation details here are out of scope for the project and don't have lasting implications we can't sort out later.

sarayourfriend:
> Install PDM. I recommend using the pipx approach, I am too.

FWIW, this isn't the official recommendation from PDM, and I think I've run into weird virtual environment issues sometimes when using the pipx approach that I never had happen when using the project's own recommendation: https://pdm-project.org/en/latest/#recommended-installation-method

If you're not having issues, no worried, but if anyone does run into things, I was able to get everything working without a hitch by avoiding pipx for this.

api/Dockerfile
sarayourfriend: Only reason is select a specific version is for lockfile compataibility stuff with PDM, similar to pnpm. FWIW `~=` will resolve to 2.15.0 I think? But if we're going to use a specific version of PDM, we should document and parameterise it like we do for pnpm :thinking: Ideally we're just using the latest PDM though, however "unreproducible" that technically could be... If we're very worried about it, we should use `https://tracker.debian.org/pkg/pdm` instead, but we didn't pin pipenv before, so I think it might not be too necessary for pdm either.

sarayourfriend (approved):
This LGMT, Dhruv, keeping in mind my opinion that implementation details of the `openverse-attribution` package are out of scope for this PR and should be pushed to a separate issue dedicated that package.

I'm approving this personally now despite the `py-packages` directory, because I don't think it's strictly necessary to block this PR on the location of those files, and it'll be trivial to move them after the related JS package PR is merged.

One thing that could be nice is to have a fast-follow to implement a just recipe like the `p` recipe, that runs `pdm` for all Python packages, optionally following a pattern. If we don't use pdm workspaces, I suppose we can just use `find` to identify `pyproject.toml` files and use PDM's` -P` flag to point to them and run the command in each one. To that end, we should have consistent script names as well (cf. `test:unit`, `types` for the JS packages).

That, to me, is a nice to have in a follow up issue though, and not necessary for this first PR, which is already interesting enough as it is. I agree with many of the things others have brought up as potential avenues to explore in follow up issues, but do not believe any to be blockers for this specific PR, which we'll want to get out and able to build upon for the rest of the -> PDM migration. Feel free to correct me @krysal and @AetherUnbound, just leave a request changes, but in the spirit of expedient decision making, it'd be great to be able to move forward with this without accidentally expanding the scope, especially for anything that don't have lasting harmful implications for the project or its team members (cf. [the definition of "blocker" in our decision making process](https://docs.openverse.org/meta/decision_making/process_description.html#terms).

Not trying to override anyone else's intuition, just want to encourage making sure we carefully evaluate and respect a defined scope for the PR.

api/Dockerfile
dhruvkb: `~=2.14` means `>=2.14,==2.*` so it gives us both the benefit of any new minor updates and patches that come out but prevents updating to a v3 that might break compatibility. Python's `pyproject.toml` doesn't have an equivalent of Node's `engines` so I don't think making this a parameter will work.

sarayourfriend: It could be encoded in the `pyproject.toml` at the monorepo root, as just a comment (like we've done for the catalog's Python version), but I'm not very worried at all about this.

dhruvkb:
Since these changes are working and the idea has approval, I will merge this PR. Not dismissing any of the comments from the reviews, I'll make separate issues for each of them and add them to the milestone.

api/justfile
```diff
@@ -14,13 +14,16 @@ NO_COLOR := "\\033[0m"
 # Version #
 ###########
 
-export API_PY_VERSION := `grep 'python_version' Pipfile | awk -F' = ' '{print $2}' | sed 's/"//g'`
-export PGCLI_VERSION := `grep 'pgcli' Pipfile | awk -F' = ' '{print $2}' | sed 's/\~//g; s/"//g; s/=//g'`
+# These commands use standard GNU tools instead of `pdm` because we do not
+# require PDM on the host machine for API development.
```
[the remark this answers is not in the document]
dhruvkb: This statement still holds. If one is working on the API (and even if they are making changes to the attribution package), they do not need PDM on the host (because the package will be mounted into the API container directly and be usable via the API).

PDM is only needed on the host if one is purely working outside the context of the API, like the attribution package directly.

PR #4142, opened 2024-04-17
Decode and deduplicate tags in the API

obulat:
## Fixes

Fixes #4125 by @obulat
Fixes #1927 by @sarayourfriend 

## Description

This PR copies the string decoding code from the frontend to the API to decode the incorrectly-encoded tag names. It also adds a step to check that the decoded characters can be URI-encoded. This is done so that the tag names can be used as the URL query parameter in the frontend, to fix #4125.

A follow-up to this PR should remove this code from the frontend. It should only be done after the API changes are deployed.

## Testing Instructions

Check that the tests run and test the edgecases.

Try adding the invalid tags to sample data using this patch:
```
Subject: [PATCH] Decode and deduplicate tags in the API
---
Index: sample_data/sample_image.csv
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/sample_data/sample_image.csv b/sample_data/sample_image.csv
--- a/sample_data/sample_image.csv	(revision 06904ebc6c4c1747bdaf18b22e781b1ccbadd209)
+++ b/sample_data/sample_image.csv	(date 1713333760883)
@@ -5,7 +5,7 @@
[...]

Issue #3912, opened 2024-03-13
Upload Ingestion Server's TSV files to AWS S3


krysal:
## Description

Corresponds to the [first step](https://docs.openverse.org/projects/proposals/data_normalization/20240227-implementation_plan_catalog_data_cleaning.html#save-tsv-files-of-cleaned-data-to-aws-s3) of the Data Cleaning project. We should have the files for cleaned tags after this is done.

## Additional context

This is part of https://github.com/WordPress/openverse/issues/430.

Issue #3415, opened 2023-11-29
Use the `batched_update` DAG with stored CSVs to update Catalog URLs


obulat:
## Problem

We have generated some CSVs with `identifier` and another column that we need to use to update the Catalog media table, but we don't have a way to efficiently run the media table updates.

## Description

The batched update DAG is reusable DAG which can be used to perform an arbitrary batched update on a Catalog media table, while handling deadlocking and timeout concerns.
During the cleanup process in data refresh, we generate the CSVs that contain the item `identifier` and the cleaned up version of another column (`title`, `url`, `foreign_landing_url`, `creator_url` and ~`tags`~). We need a DAG that is similar to the batched update DAG, but can use a CSV table for selecting the items that need to be updated.

It is important that this work does not _delete_ any tags. The tag column, while present in the CSVs, should _not_ be used.

## Additional context

The CSV files are saved in the docker container of the ingestion server when we run data refresh.

zackkrida:
This was [discussed](https://make.wordpress.org/openverse/2024/05/20/openverse-monthly-priorities-meeting-2024-06-05/#comment-282) in today's priorities meeting. We decided to move forward with this PR but _only_ apply updates to URLs. We do not want to delete any data (tags in this case). 

The accuracy threshold filtering of existing will now be preserved and iterated on as part of the [Removal of the ingestion server](https://github.com/WordPress/openverse/issues/3925#top) [#3925](https://core.trac.wordpress.org/ticket/3925) project.

PR #4124, opened 2024-04-16
Modify `add_license_url` DAG for more specific null check

krysal:
## Fixes

Fixes #3885 by @krysal

## Description

This PR modifies the existing DAG submitted by @obulat to correct all the rows without a `license_url` in their `meta_data` field. Yesterday, I queried upstream DB and the distribution of rows to fix was the following:

```sql
openledger> SELECT license, license_version, count(identifier)
 FROM image WHERE meta_data->>'license_url' IS NULL
 GROUP BY license, license_version;
+----------+-----------------+----------+
| license  | license_version | count    |
|----------+-----------------+----------|
| by       | 2.0             | 18013988 |
| by       | 3.0             | 109      |
| by       | 4.0             | 20280    |
| by-nc    | 2.0             | 12378526 |
| by-nc    | 3.0             | 84       |
| by-nc    | 4.0             | 34596    |
| by-nc-nd | 2.0             | 23425098 |
| by-nc-nd | 3.0             | 292      |
| by-nc-nd | 4.0             | 15547    |
| by-nc-sa | 2.0             | 26487161 |
| by-nc-sa | 3.0             | 9229     |
| by-nc-sa | 4.0             | 9575     |
| by-nd    | 2.0             | 5184389  |
| by-nd    | 3.0             | 112      |
| by-nd    | 4.0             | 175      |
| by-sa    | 2.0             | 9843818  |
| by-sa    | 3.0             | 137      |
| by-sa    | 4.0             | 15819    |
| cc0      | 1.0             | 884224   |
| pdm      | 1.0             | 1452034  |
| pdm      | 4.0             | 43       |
+----------+-----------------+----------+
SELECT 21
Time: 2181.062s (36 minutes 21 seconds), executed in: 2181.030s (36 minutes 21 seconds)
```

This version does not interact with AWS S3; given the low number of license-version combinations, it's not necessary. Instead, if some failed pairs are found, it's notified at the end.

## Testing Instructions

Spin up the catalog stack, `just catalog/up`, and remove the targeted field from some rows. You could try the next update statement on the sample data.

```sql
UPDATE image SET meta_data = '{"foo": "bar"}'::jsonb WHERE identifier IN (
	'd74da171-0347-435c-a32a-74a4526d3950',
	'bc4120e9-1ed3-4499-9a10-69d5bebb03e4',
	'22c54bc1-df4e-4bd7-ab19-9e649aacada1',
	'093344ca-9aa7-4f96-b40a-1fd493c41a15'
);
```

[...]

PR #4143, opened 2024-04-17
Decode and deduplicate tags during data refresh

### LINKED ISSUE #4125: Incorrectly decoded tag names cause URIError: URI malformed
## Description
Some tags in the database were encoded incorrectly, and we decode them in the frontend:
https://github.com/WordPress/openverse/blob/b2f3fd6a38b473d00b50db61a83f625e05042a6c/frontend/src/utils/decode-data.ts#L14-L24

Some combinations are incorrect and are converted to combinations that cannot be URI encoded. When we try to use them as part of the tag collection url, we get an unhandled gray screen exception.

For instance, the `udadd` in `ciudaddelassiencias` is incorrectly decoded. 

## Sentry Issue
[OPENVERSE-FRONTEND-1VB](https://openverse.sentry.io/issues/5199166395/?referrer=github_integration)

```
URIError: URI malformed
  at encodeURIComponent (<anonymous>)
  at Array.map (<anonymous>)
  at router.resolve (./.nuxt/router.js:12294:12)
  at resolveRoute (./.nuxt/nuxt-i18n/plugin.routing.js:68:39)
  at localePath (./.nuxt/nuxt-i18n/plugin.routing.js:12:39)
...
(45 additional frame(s) were not displayed)
```

obulat:
## Fixes

Fixes #4125 by @obulat

## Description

This is an alternative to #4142 that decodes and deduplicates tags in the cleanup process during ingestion instead of the API serializer. I opened it as a reply to @sarayourfriend's comment in #4142.

This might make the cleanup process longer than usual, so I would prefer to add these changes once we have the process of saving the updated data to a TSV(#3912) and updating the upstream catalog with it (#3415), from Data normalization project.

It is important to fix the encoding quickly because it can cause a gray Nuxt error screen for pages that contain tags with character sequences that cannot be URI-encoded (such as `udadd` from "ciudaddelassiencias").

## Update
I realized that we do need to update the incorrect decoding of combinations such as `udada` in the frontend in any case to fix all strings, not only the tags. That's why I opened #4144.

So, the question is now this: should we add tag decoding/deduplicating to the cleanup process during data refresh now, or only after the data normalization process is merged, @WordPress/openverse-catalog ? I'm afraid that this update could make the data refresh process too long, and think that it might be better to finish the data normalization project, and only add this cleanup after the other cleanup processes are removed.

## Outdated alternatives, for the record
_There are three alternatives to fix this:_
_1. Add a fix to the frontend fixes. We already do some decoding in the frontend, so we can add one step. Then, after the data normalization project pieces are in place, we can add the decoding/deduplicating of tags to the tag cleanup (this PR). After updating the upstream catalog data, we can remove the decoding/deduplicating from both the frontend and the ingestion._
_2. Add the decoding and deduplicating of tags to the API serializer. Then, remove the frontend decoding/deduplicating. After the data normalization process is ready, add the decoding/deduplicating of tags to the tag cleanup (this PR). After updating the upstream catalog data, we can remove the decoding/deduplicating from both the API and the ingestion._
_3. Only add the cleanup to the ingestion server. If the data cleanup process does not become too much longer, use it until the data normalization pieces are ready, and remove the cleanup steps from the frontend.~~_

_I would prefer the alternative number 3, but I'm a bit hesitant because I don't know how much longer data refresh will become.
After writing these steps, I think I'd prefer alternative number 1 because it has fewer changes (we already do the cleanup in the frontend).
What do you think?__

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -21,6 +21,7 @@ from psycopg2.extras import DictCursor, Json
 
 from ingestion_server.db_helpers import database_connect
 from ingestion_server.indexer import DB_BUFFER_SIZE
+from ingestion_server.strings import decode_data, deduplicate_tags
 
 
 # Number of records to buffer in memory at once
@@ -117,6 +118,22 @@ class CleanupFunctions:
             return None
 
     @staticmethod
+    def deduplicate_tags(tags: list[dict]) -> list[dict]:
+        """
+        Remove tags that have the same name and provider.
+        Not comparing accuracy here: if the tags have different accuracy values,
+        they will also have different provider values (e.g., `clarifai` vs `flickr`).
+        """
+        seen = set()
+        unique_tags = []
+        for i, tag in enumerate(tags):
+            tag_tuple = (tag["name"], tag.get("provider"))
+            if tag_tuple not in seen:
+                seen.add(tag_tuple)
+                unique_tags.append(tag)
+        return unique_tags
+
+    @staticmethod
     def cleanup_tags(tags):
         """
         Delete denylisted and low-accuracy tags.
@@ -129,19 +146,25 @@ class CleanupFunctions:
         if not tags:
             return None
         for tag in tags:
-            below_threshold = False
             if "accuracy" in tag and float(tag["accuracy"]) < TAG_MIN_CONFIDENCE:
-                below_threshold = True
-            if "name" in tag and isinstance(tag["name"], str):
-                lower_tag = tag["name"].lower()
-                should_filter = _tag_denylisted(lower_tag) or below_threshold
-            else:
-                log.warning(f'Filtering malformed tag "{tag}" in "{tags}"')
-                should_filter = True
-            if should_filter:
                 update_required = True
-            else:
-                tag_output.append(tag)
+                continue
+            if "name" not in tag or not isinstance(tag["name"], str):
+                update_required = True
+                continue
+            decoded_tag_name = decode_data(tag["name"])
+            if _tag_denylisted(decoded_tag_name.lower()):
+                update_required = True
+                continue
+            if decoded_tag_name != tag["name"]:
+                update_required = True
+                tag["name"] = decoded_tag_name
+            tag_output.append(tag)
+
+        deduplicated_tags = deduplicate_tags(tag_output)
+        if len(deduplicated_tags) != len(tag_output):
+            update_required = True
+            tag_output = deduplicated_tags
 
         if update_required:
             fragment = Json(tag_output)
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -21,7 +21,7 @@ from psycopg2.extras import DictCursor, Json
 
 from ingestion_server.db_helpers import database_connect
 from ingestion_server.indexer import DB_BUFFER_SIZE
-from ingestion_server.strings import decode_data, deduplicate_tags
+from ingestion_server.strings import decode_data
 
 
 # Number of records to buffer in memory at once
@@ -161,7 +161,7 @@ class CleanupFunctions:
                 tag["name"] = decoded_tag_name
             tag_output.append(tag)
 
-        deduplicated_tags = deduplicate_tags(tag_output)
+        deduplicated_tags = CleanupFunctions.deduplicate_tags(tag_output)
         if len(deduplicated_tags) != len(tag_output):
             update_required = True
             tag_output = deduplicated_tags
--- a/ingestion_server/ingestion_server/strings.py
+++ b/ingestion_server/ingestion_server/strings.py
@@ -1,0 +1,47 @@
+import logging
+import re
+from urllib.parse import quote, unquote
+
+
+DOUBLE_BACKSLASH_ESCAPE = re.compile(
+    r"\\(x)([\da-f]{2})|\\(u)([\da-f]{4})", re.IGNORECASE
+)
+NO_BACKSLASH_ESCAPE = re.compile(r"(u)([\da-f]{4})", re.IGNORECASE)
+
+
+def convert_grp(grp: str) -> str | None:
+    """
+    Convert a hex value into a character. Return None if the conversion results in
+    a character that cannot be used as a URI component.
+    """
+    try:
+        converted = chr(int(grp, 16))
+        # Decoded strings should be usable as URI components
+        quote(converted)
+        return converted
+    except UnicodeEncodeError:
+        return None
+
+
+def decode_data(data: str | None = "") -> str:
+    if not data:
+        return ""
+
+    def replace_func(match):
+        """Replace the matched group with the converted character if possible, otherwise return the original string."""
+        prefix, grp = match.groups()
+        if converted := convert_grp(grp):
+            return converted
+        return f"{prefix}{grp}"
+
+    # Handle characters encoded with double backslashes
+    if DOUBLE_BACKSLASH_ESCAPE.search(data):
+        try:
+            decoded_data = data.encode().decode("unicode_escape")
+            data = decoded_data
+        except (UnicodeDecodeError, UnicodeEncodeError):
+            logging.debug(f"Failed to decode data with double backslash: {data}")
+    # Handle characters encoded without backslashes
+    data = re.sub(NO_BACKSLASH_ESCAPE, replace_func, data)
+
+    return unquote(data)


sarayourfriend:
It occurs to me after reading what you said about the data refresh, at the point at which we are adding this to clean up, is it actually _faster_ to do the clean up in the catalog once and for all?

A hot-fix on the frontend is, I think, preferred. These searches are behind unstable parameters, so I don't feel like we need to fix them in the API now (it's an edge-case bug, okay, that's fine, the method is labelled as potentially weird/changing/etc).

If the frontend hot-fix holds us over to be able to do the clean up in the catalog, then it wouldn't affect data refresh timing, right?

I can't remember if this is a very hard task, though, doing an iterative cleanup of records in the catalog. Would it take longer to do in the catalog than in a single data refresh? I guess that's my main question.

Curious to hear from @WordPress/openverse-catalog, but based on what you've said, I think putting the simpler fix into the frontend, where we already decode anyway (and maybe pass the tag list through `Set` to deduplicate?) is a better stop-gap if we can then just do this at the catalog level, even if it takes a couple of weeks for a DAG to do the iterations.

obulat:
> It occurs to me after reading what you said about the data refresh, at the point at which we are adding this to clean up, is it actually _faster_ to do the clean up in the catalog once and for all?

Cleaning up the catalog is very difficult because we have so much data. Even selecting items based on fields other than the indexed fields can be extremely slow. That's why @krysal is working on the data normalization project: it will save the data from the cleanup process during ingestion in the shape of `identifier,<cleaned field>`. Selecting using `identifier` and updating them should be faster because `identifier` is indexed in the catalog.
Once the 2 pieces (saving the cleaned data from the ingestion and updating the catalog using the saved TSV from step 1) are ready, we can do a faster catalog clean up once and for all. Then we can remove the hot-fixes from the frontend and the API. 


> A hot-fix on the frontend is, I think, preferred. These searches are behind unstable parameters, so I don't feel like we need to fix them in the API now (it's an edge-case bug, okay, that's fine, the method is labelled as potentially weird/changing/etc).

I looked through the code again, and agree with you now. I opened the frontend PR #4144, and will close the API one. Thank you for reviewing and improving it! I've added some changes you suggested to the frontend fix.

> If the frontend hot-fix holds us over to be able to do the clean up in the catalog, then it wouldn't affect data refresh timing, right?

I don't understand what "hot-fix holds us over" mean here, sorry. The data refresh timing will not be affected by the frontend hot fix, but the fix in _this_ PR **will**. That's why we I think we should not add these changes until we get the data normalization working - after which we will only need to run this cleanup once during data refresh.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -17,6 +17,7 @@
 
 from ingestion_server.db_helpers import database_connect
 from ingestion_server.indexer import DB_BUFFER_SIZE
+from ingestion_server.strings import decode_data, deduplicate_tags
```
obulat, line 20:
```suggestion
from ingestion_server.strings import decode_data
```

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -125,19 +142,25 @@ def cleanup_tags(tags):
...
+            decoded_tag_name = decode_data(tag["name"])
+            if _tag_denylisted(decoded_tag_name.lower()):
+                update_required = True
+                continue
+            if decoded_tag_name != tag["name"]:
+                update_required = True
+                tag["name"] = decoded_tag_name
+            tag_output.append(tag)
+
+        deduplicated_tags = deduplicate_tags(tag_output)
```
obulat, line 160:
```suggestion
        deduplicated_tags = CleanupFunctions.deduplicate_tags(tag_output)
```

sarayourfriend:
> I don't understand what "hot-fix holds us over" mean here, sorry. The data refresh timing will not be affected by the frontend hot fix, but the fix in this PR will. That's why we I think we should not add these changes until we get the data normalization working - after which we will only need to run this cleanup once during data refresh.

I'm not familiar with the details of the data normalisation project, sorry if my ignorance here is causing confusing suggestions. I just mean that the frontend fix is, to me, an acceptable temporary solution provided we are able to fix this at the catalog level once, rather than doing it on every data refresh.

> Cleaning up the catalog is very difficult because we have so much data

The data refresh deals with the same amount of data, doesn't it? And scrolls through using just the identifier. Like I said, I'm not familiar with the data normalisation project, but why isn't it possible to iterate through all records in the catalog (rather than trying to _select_ the relevant records) and apply a transformation in place, using the identifier as the key field (which is indexed)? Maybe this question is covered in the data normalisation project, so I will take some time to read up on it today or early next week.

krysal:
@obulat I'll prioritize #3912, and it will be my next issue to work on.

@sarayourfriend The root of the problem is that the catalog DB doesn't have indices like the API DB, so it's much slower to find and update rows. It's possible to do it still, but it takes much longer when we don't have the identifiers of the rows we want to update in advance.

sarayourfriend:
> It's possible to do it still, but it takes much longer when we don't have the identifiers of the rows we want to update in advance.

Understood. But the ingestion server approach iterates through every record every data refresh as it is. Is the data refresh able to do this faster than a DAG, for some reason? Even if it's slow for some other reason, doing it once in the catalog is surely less time and CPU cycles even in the near term than doing it every single data refresh?

obulat:
> Understood. But the ingestion server approach iterates through every record every data refresh as it is. Is the data refresh able to do this faster than a DAG, for some reason?

I believe that we don't have to worry about writes and database locks with data refresh, whereas the catalog database should always be writeable by DAGs

AetherUnbound:
@obulat and @krysal - is this still something we want to accomplish in this manner in the near term? Should we prioritize this before the removal of the ingestion server?

stacimc:
Jumping in here, so apologies if I'm missing a lot of context -- is this still broadly the plan for deploying these changes:

> we can first add this step to the data refresh process' cleanup step, save the cleaned up data, run the csv batched update dag to add this data to the catalog, and then remove this cleanup from the data refresh.

(Taken from this [comment](https://github.com/WordPress/openverse/pull/4142#issuecomment-2064638713).)

Just checking that the intention is for this to be included for one data refresh _and then removed_, (which I think addresses some of @sarayourfriend's concern). This is also necessary in order to not break with our stated goals of removing cleanup steps from the data refresh in the Ingestion Server Removal project, to @AetherUnbound's point about prioritizing this quickly.

krysal:
> @obulat and @krysal - is this still something we want to accomplish in this manner in the near term? Should we prioritize this before the removal of the ingestion server?

Yes, and yes, for the reasons that @stacimc stated, but to clarify, we expect _all_ the clean steps in the ingestion server to be unnecessary eventually after [this milestone](https://github.com/WordPress/openverse/milestone/23) is done. @obulat has also warned us about adding this additional step before more progress is made because the data refresh will potentially take much longer to complete; this is why the PR is in draft mode.

sarayourfriend:
> I believe that we don't have to worry about writes and database locks with data refresh, whereas the catalog database should always be writeable by DAGs

I don't want to derail this, but this has come up so many times, and I'm very confused by some of the decisions that have been made. I've stayed out of the discussion for lack of familiarity with the catalog database, but I recently learned about the batched update DAG, and am more confused than ever about the reasons and rationale I've heard for avoiding updates to data in the catalog database, or why there is so much complexity in these types of operations on the data. I do not understand where the problem is coming from, what causes locking to the table, or how any locking that does exist could be a problem, if the batched update DAG works (which we know it does).

My understanding is that the batched update DAG iterates through the catalog database in batches, and applies transformations to rows. It does not lock a table to update a row (keyed by an indexed field, such as the primary key), otherwise the batched update DAG would never work. Furthermore, Postgres _doesn't even have locks on writes_, and you have to explicitly request a lock if you want one. Even when a row is locked, it is only locked to other writers, not to queryers (https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS).

Provided we aren't making changes to the columns or data types, what is the specific limitation we are running up against when making changes to the catalog database that cause so much time and effort spent on what appear to be, on the face, trivial data transformation tasks?

The existence of this limitation has been referenced many times, but I've never seen a clear explanation for why iterating through the catalog database and updating rows one by one would cause any problems. I've recently learned about the batched update DAG, and am left even more confused, because it clearly _is_ possible to iterate through the catalog database and make updates to rows. Sometimes locks are referenced, sometimes a lack of indexes are referenced, but both of these seem to misunderstand the nature of the problem. There are no indexes that can cause the types of problems described, and there is no such problem of locking in Postgres that would affect this particular problem (iterating through each row individually or in batches, and making changes to the row, using the primary key as the identifier).

How much complexity are we adding to which parts of our system, and how many things are we jumping through hoops to accomplish that could be a simple Python operator that takes a row and gives back the cleaned version, run by a DAG that goes through each work one-by-one. Even if it took a week for each pass (which I doubt, considering the data refresh doesn't take nearly that long, but we could go slower to avoid hogging resources on the catalog box), I don't see where the limitation is coming from that prevents this from working.

The batched update DAG seems to prove that it's possible, and that the overhead of it is non-existent or negligible. What am I missing? What prevents us from moving these in-Python transformations from the ingestion server into a Python operator?

I am deeply confused by the way the conversations and decisions about catalog data transformations and the cleanup steps have been made so far, and the limitations that are referenced offhand. I do not wish to derail anything, or to imply that anything is being intentionally ignored or that this problem is easy or obvious. However, I am worried that there have been grave and deep misunderstandings of the technologies and problems we're working with, and that in fact those misunderstandings are even in conflict with actual things we can and do already do (batched update DAG). I am in particular worried that these misunderstandings have cost us dearly in project planning and implementation time, and have also caused difficulties and misunderstandings in the communication about these problems and the tools we're working with. I, at least, certainly feel uncertain in my understanding of what is happening, and feel like the explanations and rationales I am reading (and the way I have interpreted them) is in direct conflict with what I know to be true about the tools we have, and the problems we're trying to solve.

To be clear, to me the fundamental problem is: how do we iterate through all the catalog data and make selective transformations of some of the rows. This framing assumes that _all rows are iterated through every time_, rather than the theoretically more efficient version that only iterates through the rows that need transformation. _However, that inefficiency is also solvable_ because we have indexed versions of our data where we can retrieve immutable keyed information (the identifier), so long as the queryable aspects are available in either the API database or Elasticsearch.

obulat:
> To be clear, to me the fundamental problem is: how do we iterate through all the catalog data and make selective transformations of some of the rows. This framing assumes that _all rows are iterated through every time_, rather than the theoretically more efficient version that only iterates through the rows that need transformation. _However, that inefficiency is also solvable_ because we have indexed versions of our data where we can retrieve immutable keyed information (the identifier), so long as the queryable aspects are available in either the API database or Elasticsearch.

Initially, using the `cleanup` process for getting the list of `identifier`s and the cleaned up data was thought as a workaround for the problem that a naive `UPDATE` query on the production catalog database is prohibitively expensive.
When running an `UPDATE` query, Postgres usually adds a new row with updated data and makes all of the indexes point to the new row in the table. This means a lot of work and extra space for every row we need to update.
Since we were iterating over every image in the clean up process anyway, it seemed a good point to save the cleaned up data.

However, now that we have the batched update DAG, we can iterate over each item from Flickr (our biggest provider) in 8-10 days. The batched update DAG does not suffer from the index update problem because it uses a temporary table that has no indexes in it. Using Postgres Python functions, we can probably create good `SELECT` and `UPDATE` queries for solving the issue of this PR. This was the conclusion of our [Public meeting discussion in Slack](https://wordpress.slack.com/archives/C02012JB00N/p1717429486019369).

Closing this PR in favor of a better solution that uses `batched_update` DAG.

krysal:
@sarayourfriend, I'll add to he context my learning from last week.

>My understanding is that the batched update DAG iterates through the catalog database in batches, and applies transformations to rows. It does not lock a table to update a row (keyed by an indexed field, such as the primary key), otherwise the batched update DAG would never work...

That's partially correct. The batched update DAG first scan the database to select the subset of rows that needs to be modified, save them in an auxiliary table and then use it to do the batched updates. I believe we haven't executed a `batched_update` DAG run over all the catalog rows.

> Provided we aren't making changes to the columns or data types, what is the specific limitation we are running up against when making changes to the catalog database that cause so much time and effort spent on what appear to be, on the face, trivial data transformation tasks?

> The existence of this limitation has been referenced many times, but I've never seen a clear explanation for why iterating through the catalog database and updating rows one by one would cause any problems. I've recently learned about the batched update DAG, and am left even more confused, because it clearly _is_ possible to iterate through the catalog database and make updates to rows.

I tried a similar DAG to do batched updates for the `license_url` (the [previous version](https://github.com/WordPress/openverse/pull/4124)), skipping the creation of the auxiliary table, and it consistently kept timing out unfortunately (either by executing the SQL directly or via a DAG). If you see the query at that time (linked below), you'll notice it's pretty similar but finding which rows to update every time is a heavy task apparently. That is my conclusion from the experiment. So the step to collect identifiers of the rows previously is unavoidable.

https://github.com/WordPress/openverse/blob/feb8e003fe34db0fa744c7120a99abef839bf0e3/catalog/dags/maintenance/add_license_url.py#L115-L128

https://github.com/WordPress/openverse/blob/9acbc9f76e6996b9456c1b89e7c0cabdcaa9c04e/catalog/dags/database/batched_update/constants.py#L33-L41

> How much complexity are we adding to which parts of our system, and how many things are we jumping through hoops to accomplish that could be a simple Python operator that takes a row and gives back the cleaned version, run by a DAG that goes through each work one-by-one.

> The batched update DAG seems to prove that it's possible, and that the overhead of it is non-existent or negligible. What am I missing? What prevents us from moving these in-Python transformations from the ingestion server into a Python operator?

The DAG uses SQL to do the transformations so a limitation to consider is what we can do in the language, although we are evaluating using [custom functions](https://www.postgresql.org/docs/current/plpython-funcs.html) with a touch of python in the database too.

The [initial plan](https://docs.openverse.org/projects/proposals/data_normalization/20240227-implementation_plan_catalog_data_cleaning.html) contemplated the use of the ingestion server to take advantage of the code parallelization and save the identifiers of the rows to modify, but given how cumbersome it is to modify it, it has been decided today to try relying more heavily on the SQL side.

sarayourfriend:
I think I am still missing a piece of the puzzle.

> That is my conclusion from the experiment. So the step to collect identifiers of the rows previously is unavoidable.

If collecting the identifiers is just "all the identifiers", does that work? It's the approach I implemented for #4429 but I still feel like I must be missing something, because no one seems to be suggesting this approach anywhere, and we're spending a lot of time working on unindexed subsets of the data. The DAG I wrote uses the batched update DAG to select _all_ rows for the temporary identifiers table. Staci suggested splitting operations up by provider in the Make Slack, which makes sense to me because the provider is indexed (as part of the mulitcolumn btree unique index of provider and foreign identifier, and it is the leading column and so still efficient to query against even individually).

What I don't get is the insistence of operating batched updates on only working on a subset of the data (aside from splitting it along logical parts like the provider). Like in https://github.com/WordPress/openverse/issues/1566#issuecomment-2038338095, why did we need to query for rows that had duplicate fields to begin with? Is that really faster than just `SET tags = (SELECT ...deduplicated_tags...)` on every row?

I also don't see exactly what the relationship of adding Python to our database in this case. Is that just to make the transformation more flexible/easier to reason about? It surely won't do anything to make the batched update more efficient across a larger set of data? If we use the same strategy as the existing batched update DAG to create the table to track which rows needed transforming still, but then just queried those rows out into Python, transformed them there, and then reinserted them, is that materially different than doing the transformation in SQL? Or doing it in Python, but in a weird Postgres context?

I still feel like I'm missing something that prevents what appears to me to be a simpler solution.

I guess it boils down to what the performance characteristics are of: `UPDATE <media> SET <whatever transformation> WHERE identifier in (<chunked static list of identifiers>);`, and if you did that for all the rows of the database. Or if you did `SELECT * FROM <media> WHERE identifier in (<chunked static list of identifiers>)` and then iterated over those rows in Python to build batched update queries in Python of the form `UPDATE <media> SET <values> FROM (values (<identifier> <...transformed data>), (<identifier> <...transformed data>)) AS t(identifier, ...) WHERE identifier = t.identifier;`, which has the added benefit that you can also exclude untransformed rows from the built update query to skip touching them with noop changes.

Assume it is chunked but that the goal is to touch every row, and that we use the temporary identifier table like the batched update DAG does currently. Is this feasible? If not, why, specifically?

AetherUnbound:
I think some of the hesitancy we have had around this is that we haven't done a batched update for _the entire table_ yet: the most we've done is all Flickr images for the popularity update, and that consistently takes around 9 days. Assuming the updates are linear in time, an update made to the entire database (of which Flickr constitutes around 60% of) would take around 15 days. This is long, but not infeasible.

I think, where possible, we _should_ try and filter down the initially selected rows. For instance, with #4199, I suspect the issue is only in Flickr, and so selecting and updating the rows for Flickr would be best. But we can also add some logic to the batched update DAG which will discard any rows from the table that haven't been updated for cases where we just need to select all the data because we're uncertain about it. We could add an extra row in the temporary table which is `false` initially, then gets set to `true` if we end up changing the value. Then all `false` values can be deleted from the temp table before the batched update takes place.

I agree that the batched update, given how many considerations it makes around updates and how we've seen it prove to be significantly faster than most conventional cases, should be our go-to for handling these cases.

stacimc:
I strongly prefer using the batched update DAG whenever possible, even if it _does_ take slightly longer -- 15 days would be acceptable IMO, and hopefully we can get much shorter than that. Among other things, the advantage of not interfering with the timing of the data refresh is huge.

If the batched update DAG is missing needed features I would also much prefer we invest in adding functionality to that DAG, for example by automating chunking of batched updates by provider or by adding @AetherUnbound's [suggestion](https://github.com/WordPress/openverse/pull/4143#issuecomment-2146354476).

sarayourfriend:
Great, thanks. This directly and clearly answers my questions!

PR #4163, opened 2024-04-19
Save cleaned data of Ingestion Server to AWS S3

### LINKED ISSUE #3912: Upload Ingestion Server's TSV files to AWS S3

krysal:
## Fixes

Part of #3912 by @krysal

## Description

This PR changes the Ingestion Server's cleanup step to save the cleaned data to an S3 bucket once they reach a certain number of rows. I chose 10M given that from the previously saved files, the one with most lines has +9M, so we know around that quantity is manageable in memory and produce fewer file parts.

```sh
wc -l creator_url.tsv && wc -l foreign_landing_url.tsv &&  wc -l url.tsv 
 9035462 creator_url.tsv
 8963194 foreign_landing_url.tsv
   57390 url.tsv
```

S3 does not support appending to files; you can only replace them, so the file for each column to modify is split into chunks and uploaded as they are generated.

This is the first part to upload the files, additionally, the server instance will need to have permission over the bucket, as @sarayourfriend pointed out.

### Using Airflow UI

To test this locally with MinIO, create the `openverse-catalog` bucket if it's not automatically created at start. Go to http://localhost:5011/ (username: `test_user` & password: `test_secret`).

```
just api/up && just catalog/up
```

Then make some rows of the `image` table in the catalog _dirty_ by removing the protocol in one or several of the URLs (`url`, `creator_url`, or `foreign_landing_url`) or modify the tags for low accuracy. Run the data refresh from the Airflow UI, wait for it to finish and check the bucket in MinIO.

http://localhost:5011/browser/openverse-catalog/

The data refresh should continue even if the upload to S3 fails for whatever reason. Try shooting down the S3 container and clearing the `ingest_upstream` step in the DAG to confirms it continues despite the failure in the upload.

Set the `CLEANUP_BUFFER_SIZE` environment variable to a low number, like 5 (having more rows to clean), to test uploading several files for the same field.

### Using init script

Alternatively, follow instructions provided by @AetherUnbound in [this comment](#pullrequestreview-2016077387).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
> When running locally, I got the bucket does not exist error. Do you think it's better to reuse an existing bucket, or to add the new one to the minio docker template:
> BUCKETS_TO_CREATE=openverse-storage,openverse-airflow-logs

I checked in S3 and we've used `s3://${OPENVERSE_BUCKET}/shared/data-refresh-cleaned-data/` for previous versions of these files, so I think that would be okay to use!

> upload the files to an S3 bucket once they reach a certain size (chose 10 MB kind of arbitrarily, to make them manageable)

Having checked S3, it looks like the `creator_url.tsv` and `foreign_landing_url.tsv` (the largest ones) are 700-800MB (IIRC `tags.tsv` was too large to even save and upload). We have 16GB of storage available on the primary ingestion server and 8GB available on each indexer worker. Given that a 10MB limit would produce 80 files for `creator_url.tsv` for example, I think setting the limit to something like 1GB might make more sense!

ingestion_server/env.template
```diff
@@ -2,7 +2,7 @@ PYTHONUNBUFFERED="0"
 
 #AWS_REGION="us-east-1"
 
-#ENVIRONMENT="local"
+ENVIRONMENT="local"
```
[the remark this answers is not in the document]
AetherUnbound: Agreed, I ran into this while trying to test this locally myself.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -272,24 +272,72 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+        if os.path.getsize(f"{field}.tsv") >= self.disk_buffer_size or force_upload:
+            self._upload_to_s3(field)
+
+    def _upload_to_s3(self, field: str):
+        part_number = self.buffer[field]["part"]
+        log.info(f"Uploading file part {part_number} of `{field}` to S3...")
+        s3_file_name = f"shared/data-refresh-cleaned-data/{field}_{part_number}.tsv"
+        self.s3_bucket.upload_file(f"{field}.tsv", s3_file_name)
+        self.buffer[field]["part"] += 1
+        os.remove(f"{field}.tsv")
```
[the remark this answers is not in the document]
AetherUnbound: Ah, case in point - I ran `j down -v` then removed some of the URL schemes from our `sample_image.csv` file and ran `j api/init`. Contradicting what I said [in the PR to split out the docker-compose files](https://github.com/WordPress/openverse/pull/4114#pullrequestreview-2002247833) (it was accurate at the time! 😅), it looks like the ingestion server now also should depend on Minio. So either we need to move around the docker dependencies and declare the Minio services as dependencies for the ingestion server too, or allow this step to fail (either always or at least in the local case).

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -272,24 +272,72 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
-    total_time = end_time - start_time
-    log.info(f"Finished saving cleaned data in {total_time},\n{cleanup_counts}")
-    return cleanup_counts
+    buffer = {
+        field: {"part": 1, "rows": []}
+        for field in _cleanup_config["tables"]["image"]["sources"]["*"]["fields"]
+    }
+
+    def __init__(self):
+        bucket_name = config("AWS_S3_BUCKET", default="openverse-catalog")
```
AetherUnbound, line 288: As @obulat noted, while we use `openverse-catalog` in production, the default bucket locally is `openverse-storage`:

https://github.com/WordPress/openverse/blob/51fd23537b695114b203ace195041eaec0d7f8b4/catalog/env.template#L106-L105

AetherUnbound (changes requested):
I'm excited about this! There are a few things that will need to be adjusted, in addition to my comments above about file size before upload.

Additionally, the testing instructions appear to no longer be accurate, since that initialization based on the environment appears to be a part of the code now.

Lastly, I adjusted the default bucket name and removed some HTTP schemes from the sample data as I noted below. I was able to get the upload to work, but the CSV which was uploaded had multiple copies of each identifier (the screenshot below has the rows sorted just to show the duplications, they were not in that order in the originally produced file). It appears that some of the logic is causing duplicate file writing.
image

ingestion_server/env.template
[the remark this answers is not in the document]
krysal: I was thinking about it, and the thing is, often, I have a conflict with default env vars for the catalog. I'm used to using production values as default, but we prefer dev/testing values there. I guess using `local` is not so bad here, and we want to get rid of the ingestion server so that they won't last long anyway.

ingestion_server/ingestion_server/cleanup.py
krysal: Gotcha, I added the existing bucket in production to `BUCKETS_TO_CREATE`. This way we have fewer differences between environments. I don't know why we would want to conserve `openverse-storage`. Let me know if there is a reason for this default bucket to be different between environments.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -272,24 +272,72 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+    disk_buffer_size = 10 * 1024 * 1024
 
-    end_time = time.time()
-    total_time = end_time - start_time
-    log.info(f"Finished saving cleaned data in {total_time},\n{cleanup_counts}")
-    return cleanup_counts
+    buffer = {
+        field: {"part": 1, "rows": []}
+        for field in _cleanup_config["tables"]["image"]["sources"]["*"]["fields"]
+    }
```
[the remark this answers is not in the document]
krysal: Love this suggestion! I didn't think of it before. It can't be a NamedTuple because I'm modifying the `part` and `rows` but a dataclass does the job perfectly! Done 💯

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -272,24 +272,72 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+
+    def _save_to_disk(self, field: str, force_upload=False):
+        log.info(f"Saving {len(self.buffer[field])} rows of `{field}` to local file.")
+        with open(f"{field}.tsv", "a") as f:
+            csv_writer = csv.writer(f, delimiter="\t")
+            csv_writer.writerows(self.buffer[field]["rows"])
+            # Clean memory buffer of saved rows
+            self.buffer[field]["rows"] = []
+
+        if os.path.getsize(f"{field}.tsv") >= self.disk_buffer_size or force_upload:
```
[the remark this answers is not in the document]
krysal: Right about the variable, added ✔️ 

I think the directory is indifferent since it has been working well this way, but if you prefer a specific location, we can discuss it.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -272,24 +272,72 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+            self._upload_to_s3(field)
+
+    def _upload_to_s3(self, field: str):
+        part_number = self.buffer[field]["part"]
+        log.info(f"Uploading file part {part_number} of `{field}` to S3...")
+        s3_file_name = f"shared/data-refresh-cleaned-data/{field}_{part_number}.tsv"
```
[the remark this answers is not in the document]
krysal: Good note. I believe adding the date to the filename, as @obulat suggested in a previous opportunity, will solve that issue. It makes clearer when the files were created and avoids dealing with risky file deletions in S3.

ingestion_server/ingestion_server/cleanup.py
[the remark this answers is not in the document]
krysal: Right, the process should continue, ignoring the upload. Thanks for reminding me of this case!

krysal:
@obulat @AetherUnbound Thanks for your valuable suggestions. I was too quick to mark this as ready before but I'm still glad I did!

@AetherUnbound  Regarding the file size, I set it to 850 MB, so only one file is created for each field but tags. 1 GB is on the big side for downloading without a high-speed connection. About the duplicated rows in the file, I'm not sure what could have happened there; I don't see those results in my tests 🤔 Could you try again? I applied the rest of changes suggested.

ingestion_server/ingestion_server/cleanup.py
obulat: I would prefer to replace the local `openverse-storage` with `openverse-catalog` here:
https://github.com/WordPress/openverse/blob/0ed4071577d846ff222342e5c296b723f5e7bb7e/catalog/env.template#L106
What do you think, @AetherUnbound ?
Otherwise, now locally we have 2 different buckets: one keeps the cleanup data and another - the data from provider scripts.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -266,30 +267,96 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+            try:
+                self._upload_to_s3(field)
+            except Exception as e:
+                log.error(f"Error uploading {field} to S3: {e}")
+
+    def _upload_to_s3(self, field: str):
+        part_number = self.buffer[field].part
+        log.info(f"Uploading file part {part_number} of `{field}` to S3...")
+        s3_file_name = f"{self.s3_path}/{self.date}_{field}_{part_number}.tsv"
+        self.s3_bucket.upload_file(f"{field}.tsv", s3_file_name)
```
obulat, line 340: Maybe extract the tsv file name here, too?
```suggestion
        tsv_file_name = f"{field}.tsv"
        self.s3_bucket.upload_file(tsv_file_name, s3_file_name)
```

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -383,7 +451,8 @@ def clean_image_data(table):
     conn.commit()
     iter_cur.close()
     conn.close()
-    end_time = time.time()
+    data_uploader.flush()
+    end_time = time.perf_counter()
```
obulat, line 455: Nit: it would be nice to format `cleanup_time` in the log: `{cleanup_time:.3f}`

ingestion_server/ingestion_server/cleanup.py
obulat: If we do decide to use `openverse-catalog` everywhere locally, then it's best to fallback to the `OPENVERSE_BUCKET` env variable

obulat (approved):
This works great now, @krysal !

I left several comments for improvements (non-blocking).

ingestion_server/ingestion_server/cleanup.py
krysal: Done.

ingestion_server/ingestion_server/cleanup.py
krysal: Done 👌

ingestion_server/ingestion_server/cleanup.py
krysal: I changed it to use `OPENVERSE_BUCKET`!

AetherUnbound: It looks like `openverse-catalog` is still being added to the `BUCKETS_TO_CREATE` list, shouldn't that be removed if we're deferring to `OPENVERSE_BUCKET`?

AetherUnbound (changes requested):
This is unfortunately still producing duplicate rows for me locally 😕 here's everything I did to get this working:

1. Add `OPENVERSE_BUCKET=openverse-storage` because my existing minio environment file didn't have the new bucket, and the new bucket defaults to `openverse-catalog`
2. Remove `https://` from the `foreign_landing_url` for the first 10 records of `sample_image.csv`
3. Run `just down -v`
4. Run `just c` to start the catalog (and get S3 running)
5. Run `just api/init` (this performs the image data refresh for us)
6. Download the produced TSV from localhost:5011

Although the logs say 10 records were written for `foreign_landing_url`, the CSV I download has 40 records (4 full copies of each 10 rows).

krysal:
@AetherUnbound After following those instructions, I also get some duplicates. It's weird, but I think it might be related to the retries for the image data refresh process locally.

https://github.com/WordPress/openverse/blob/feb8e003fe34db0fa744c7120a99abef839bf0e3/load_sample_data.sh#L130-L138

I added a specific folder for the output files, which gets recreated every time the process begins. I don't get more duplicates with these changes; I hope it works for you too! Let me know.

AetherUnbound:
Looks like tests for the API & ingestion server are failing on this

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -298,6 +299,12 @@ def __init__(self):
         self.s3 = self._get_s3_resource()
         self.s3_bucket = self.s3.Bucket(bucket_name)
         self.date = time.strftime("%Y-%m-%d")
+        self._recreate_local_dir()
+
+    def _recreate_local_dir(self):
+        if os.path.exists(self.local_path):
+            os.system(f"rm -rf {self.local_path}")
+        os.makedirs(self.local_path)
```
AetherUnbound, line 307: We should _definitely_ be preferring something like [`pathlib`](https://docs.python.org/3/library/pathlib.html) over using `os` calls and particularly `os.system` directly. If `local_path` is initialized as a `pathlib.Path` object, this becomes:

```suggestion
    def _recreate_local_dir(self):
        self.local_path.unlink(missing_ok=True)
        self.local_path.mkdir()
```

AetherUnbound: Ah! Seems like [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree) might work for that case 🙂

ingestion_server/ingestion_server/cleanup.py
krysal: @AetherUnbound `OPENVERSE_BUCKET` defaults to "openverse-catalog" in the Ingestion Server.

ingestion_server/ingestion_server/cleanup.py
krysal: Yeah, I applied exactly that ✅

sarayourfriend:
> Looking at this code line in the same Ingestion Server makes me guess the credentials are taken from the environment variables and nothing else is needed in production to connect to S3, but I can't guarantee it. Let me know your thoughts on this approach.

The client authenticates using the instance profile. The instance profile is _not_ configured with permissions for s3, so this change will fail when deployed unless those permissions are added to the ingestion server IAM policy. You'll need to add those new policies to this Terraform resource block: https://github.com/WordPress/openverse-infrastructure/blob/610a8207a50c08e38955bdf069ee0e721ab630f1/modules/services/ingestion-server/iam.tf#L26

Please also add the new bucket in Terraform ensure we control and own the bucket. There's a trend of scanning GitHub for bucket name references and hijacking ones that aren't reserved.

krysal:
@sarayourfriend That is very helpful, thank you!

> If there is a new bucket, please also add the new bucket in Terraform and ensure we control and own the bucket. There's a trend of scanning GitHub for bucket name references and hijacking ones that aren't reserved.

I'm using the existing bucket so there shouldn't be surprises with it :) Good to know that either way!

---

@AetherUnbound I resorted to simplify the file management given the previous approach was producing strange results in integration tests. This is ready for a re-review now.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -266,30 +267,89 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+            "s3", region_name=config("AWS_REGION", default="us-east-1")
+        )
+
+    def _upload_to_s3(self, field: str):
+        part_number = self.buffer[field].part
+        log.info(f"Uploading file part {part_number} of `{field}` to S3...")
+        s3_file_name = f"{self.s3_path}/{self.date}_{field}_{part_number}.tsv"
+        tsv_file = f"{field}.tsv"
+        with open(tsv_file, "w") as f:
+            csv_writer = csv.writer(f, delimiter="\t")
```
AetherUnbound, line 327: It looks like this current set of settings will cause the `url` column (at least when I tested it) to be quoted. Since we're using tab delimiting, we shouldn't need to worry about quoting fields. It may also make ingestion into Postgres harder down the line too, if we have to try and unquote the fields before inserting them into the database after reading the file! I'd recommend the following[^1]:

```suggestion
            csv_writer = csv.writer(f, delimiter="\t", quoting=csv.QUOTE_NONE)
```

[^1]: https://docs.python.org/3/library/csv.html#csv.QUOTE_NONE

AetherUnbound (approved):
This looks good, I was able to test it with the ingestion altering method I had used in the past. The output had no duplicate rows this time! I have one non-blocking suggestion.

krysal:
It turns out the quotes are needed for `tags`, or the file creation will raise errors. I undid said change, and locally, the tests passed, so I need to investigate why they aren't working in CI.

krysal:
Okay, so I just moved the instructions regarding whether the file upload is skipped or not, so there is not much change, which is interesting in relation to tests... Anyway, I'm merging this now!

PR #4169, opened 2024-04-19
Update media moderation view

### LINKED ISSUE #3637: Augment media admin view with moderation information
## Description

See the following for implementation description and rationale. **This issue covers only the point 1 in the linked section.** https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#id4

Blocks https://github.com/WordPress/openverse/issues/3638

obulat:
## Fixes

Fixes #3637 by @sarayourfriend

## Description

This PR updates the reporting view in line with the implementation plan requirements.

The media moderation view is split into 3 sections:
- current report details
- media item details
- other reports for the media item

The IP says to use `TabularInline` for the other reports, however, it's not possible to do so because a `TabularInline` can be used for displaying the child models on the parent admin page[^1]. The other reports for the media item, on the other hand, display the children of the parent of the admin page model (Child (report) -> Parent (media) -> Children (reports)). That's why I used a separate function to build a table of other reports.

[^1]: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#inlinemodeladmin-objects
## Screenshots
image

## Testing Instructions

Run the app using `just up`.
Go to `http://localhost:50280/admin` and log in using the local credentials (user: `deploy` user, password: `deploy`)
Add several reports for images and audio. Add several reports for the same item.
Then click on change report. You should see the updated form that contains a blurred image or an audio, as well as other parts of the report.
Check that add report/change report/view report views work correctly.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/admin/__init__.py
```diff
@@ -47,32 +57,214 @@ class MediaReportAdmin(admin.ModelAdmin):
...
+        if obj.media_obj.creator and obj.media_obj.creator_url:
+            return mark_safe(
+                f"{title} by {create_link(obj.media_obj.creator_url, obj.media_obj.creator)}"
+            )
+        else:
+            creator = obj.media_obj.creator or obj.media_obj.creator_url
+            creator_string = f" by {creator}" if creator else ""
+            return f"{title}{creator_string}"
+
+    def media_display(self, obj):
```
AetherUnbound, line 105: Should this be an `@abstractmethod`?

AetherUnbound (changes requested):
This is fantastic! I was able to test this locally and have things behave exactly as expected. I have a few comments about how we're setting the domain that might require changes.

api/api/admin/__init__.py
```diff
@@ -47,32 +57,214 @@ class MediaReportAdmin(admin.ModelAdmin):
...
+    media_type = "image"
+
+    @admin.display(description="Image")
+    def media_display(self, obj):
+        """
+        Display a blurred image with a clickable overlay.
+        Use the image thumbnail if available, else replace with the direct image url.
+        """
+        if obj.media_obj.url:
+            thumb_url = f"https://api.openverse.engineering{obj.media_url()}thumb/"
```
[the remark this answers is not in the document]
sarayourfriend: It should use [`request.build_absolute_uri`](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.build_absolute_uri) and [`reverse`](https://docs.djangoproject.com/en/5.0/ref/urlresolvers/#reverse), actually.

api/api/admin/__init__.py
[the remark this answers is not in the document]
sarayourfriend: Please, 100% split this into multiple files. The only requirement for Django admin is that all files are _seen_ at some point, so really just that they need to be imported into this module (rather than them pulling exports from here and never explicitly getting loaded).

But yes, these new classes should be split into relevantly named modules.

api/api/admin/__init__.py
```diff
@@ -47,32 +57,214 @@ class MediaReportAdmin(admin.ModelAdmin):
...
+        if not reports:
+            return ""
+        result = "<table><thead><tr><th>Date</th><th>Report reason</th><th>Status</th><th>Report link</th></tr></thead><tbody>"
+        for report in reports:
+            report_link_html = f'<a href="{reverse("admin:api_imagereport_change", args=[report.id])}">{report.id}</a>'
+            created_at = report.created_at.strftime("%Y-%m-%d %H:%M:%S")
+            report_row = f"<tr><td>{created_at}</td><td>{report.reason}</td><td>{report.status}</td><td>{report_link_html}</td></tr>"
+            result += report_row
+        result += "</tbody></table>"
+        return mark_safe(result)
```
sarayourfriend, line 165: This is really not the right way to do things in Django. Most important issue being that this string concatenation approach does not get any benefit from Django's template cache. A particularly important benefit, regarding maintainability, is the separation of view from model, which is essential for making the data model an understandable and distinctly separate thing from the presentation. Extracting all the presentation logic in the template forces us to write admin pages that cleanly silo the data model into template context, rather than creating functions like this where the two are meshed together such that it makes it unnecessarily harder to understand. That's even the more important benefit, in our context, is to avoid this approach because it is so much harder to read, reason about, and maintain than Django templates.

Please update the change_form template you've added in this PR to render the table in the template. It's also worth checking out Django's form building tools, if you're not familiar with them, as there are canonical ways to do just about everything: https://docs.djangoproject.com/en/5.0/topics/forms/#the-template and https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#overriding-admin-templates (the latter of which you've almost definitely seen, given the template override you've already added, but including it just in case and for other reviewers who might not be familiar with Django, like Madison).

api/api/admin/__init__.py
```diff
@@ -47,32 +57,214 @@ class MediaReportAdmin(admin.ModelAdmin):
...
+            return mark_safe(
+                f"{title} by {create_link(obj.media_obj.creator_url, obj.media_obj.creator)}"
+            )
+        else:
+            creator = obj.media_obj.creator or obj.media_obj.creator_url
+            creator_string = f" by {creator}" if creator else ""
+            return f"{title}{creator_string}"
+
+    def media_display(self, obj):
+        pass
```
sarayourfriend, line 106: At the very least it should raise `NotImplementedError` (referring to @AetherUnbound's comment regarding this being abstract).

```suggestion
    @abc.abstractmethod
    def media_display(self, obj):
        raise NotImplementedError("Subclasses must implement `media_display`")
```

api/api/admin/__init__.py
```diff
@@ -47,32 +57,214 @@ class MediaReportAdmin(admin.ModelAdmin):
...
+    def media_display(self, obj):
+        """
+        Display a blurred image with a clickable overlay.
+        Use the image thumbnail if available, else replace with the direct image url.
+        """
+        if obj.media_obj.url:
+            thumb_url = f"https://api.openverse.engineering{obj.media_url()}thumb/"
+            return mark_safe(
+                f'<div class="container"><img src="{thumb_url}" alt="Media Image" class="blur-image" height="300" '
+                f'onclick="toggleBlur(this)" onerror="replace(this, \'{obj.media_obj.url}\')" style="cursor: pointer;" />'
```
sarayourfriend, line 250: What is replace doing here in the `onerror`? The implementation of the function doesn't clear anything up to me. Why would the `img` element's `src` ever need to change to point to a non-image URL?

sarayourfriend (changes requested):
Requesting changes to move the HTML building into the templates, rather than using string concatenation.

Things look good from a functionality perspective, so it's really just moving the HTML code into a place that is easier to read, reason about, and maintain, plus is the canonical "Django-way".

I've left a couple other clarifying comments about some things I didn't understand about the implementation, and +1'd some of Madison's comments, but overall looking very good.

api/api/admin/__init__.py
obulat: By default, we use the thumbnail url. However, especially locally, this URL might not contain an image: some of the sample data identifiers do not exist in the production database. So, for local testing, we try loading the thumbnail url as the image src. If that cannot be loaded due to image not being available at that URL, we try loading the image's direct URL, which most likely exists. 
This behavior mirrors the behavior of the frontend single result image, where we try to load the direct image URL, and if that is not available - we load the thumbnail.

We should update the sample data to only contain data that exists in production, even if the links are dead - at least we should have the items for all identifiers in the sample data - but that's a different issue.

obulat:
Drafting this PR to address @sarayourfriend 's change requests.

api/api/admin/__init__.py
sarayourfriend: Ohhhh I see, I totally misread `obj.media_obj.url`, and interpreted it as the API route for the image rather than the upstream image URL.

Okay, so I understand this now, thanks for clarifying. The other question is then... do we need to do anything to make sure this works as expected for both images and audio (audio thumbnail doesn't proxy `.url`, I think? :thinking:)

api/api/admin/media_report.py
```diff
@@ -0,0 +1,199 @@
...
+        if not self.media_type or not obj:
+            return []
+
+        reports = (
+            self.model.objects.filter(media_obj__identifier=obj.media_obj.identifier)
+            .exclude(id=obj.id)
+            .order_by("created_at")
+        )
+        for report in reports:
+            report.href = reverse("admin:api_imagereport_change", args=[report.id])
```
AetherUnbound, line 77: I believe this will also need to be reversed by media type:

```suggestion
            report.href = reverse(f"admin:api_{self.media_type}report_change", args=[report.id])
```

api/api/admin/media_report.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if obj.media_obj.tags:
+            tags_by_provider = {}
+            for tag in obj.media_obj.tags:
+                tags_by_provider.setdefault(tag["provider"], []).append(tag["name"])
+            additional_data["tags"] = tags_by_provider
+        if self.media_type == "image":
+            additional_data["thumb_url"] = (
+                f"https://api.openverse.engineering{obj.media_url()}thumb/"
+            )
```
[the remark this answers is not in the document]
sarayourfriend: If done in the template, it's a bit less fussy to just use the `url` filter: https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#url

api/api/admin/media_report.py
```diff
@@ -0,0 +1,199 @@
...
+    @admin.display(description="URLs")
+    def landing_urls(self, obj):
+        if not self.media_type or not obj:
+            return "N/A"
+        openverse_url = (
+            f"https://openverse.org/{self.media_type}/{obj.media_obj.identifier}"
+        )
+        return mark_safe(
+            f"<p>{create_link(obj.media_obj.foreign_landing_url, obj.media_obj.provider)}, "
+            f'{create_link(openverse_url, "openverse.org")}'
```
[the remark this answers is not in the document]
obulat: I completely forgot to remove this, it's not used anymore. Fixed!

obulat:
I refactored the changes to extract the data in the python code, and render it only in the template. The urls are also built in the template.

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -87,7 +83,11 @@ <h2>Media details</h2>
       <div class="flex-container">
         <label>Tags</label>
         <div class="readonly">
-          <p>{{ tags }}</div>
+          {% for provider, provider_tags in tags.items %}
+            <p><em>{{ provider }}</em></p>
+            <p> {% for tag in provider_tags %}{{ tag }}, {% endfor %}</p>
```
AetherUnbound, line 88: Nit: This will also introduce a trailing comma on the last item. I think we could use `forloop.last` as the check for this? 

image

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -0,0 +1,159 @@
...
+          <td>{{ report.created_at }}</td>
+          <td>{{ report.reason }}</td>
+          <td>{{ report.status }}</td>
+          <td><a href="{% url 'admin:api_'|add:media_type|add:'report_change' report.id %}">{{ report.id }}</a></td>
+        </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+  {% else %}
+    <p>No other reports for <a href="{{reverse_media_url}}">{{ identifier }}</a> found.</p>
```
AetherUnbound, line 134: When I click on this for a report, I get redirected to the admin home page with this message 🤔 

image

The URL it's trying to go to is something like http://localhost:50280/admin/api/image/de37812d-2daa-49f6-91ab-e87e780ca653/change/, is it possibly because the `/change/` URL isn't available for images? Should something else be used here instead?

AetherUnbound (commented):
Two small notes, this is very close!

api/api/templates/admin/api/media_report/change_form.html
sarayourfriend: It should be done with the join filter: https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std-templatefilter-join

```suggestion
            <p>{{ provider_tags | join:', ' }}</p>
```

api/api/admin/media_report.py
```diff
@@ -0,0 +1,170 @@
...
+            "title": obj.media_obj.title,
+            "provider": obj.media_obj.provider,
+            "source": obj.media_obj.source,
+            "creator": obj.media_obj.creator or obj.media_obj.creator_url,
+            "creator_url": obj.media_obj.creator_url,
+            "url": obj.media_obj.url,
+            "tags": tags_by_provider,
+            "reverse_media_url": reverse(
+                f"admin:api_{self.media_type}_change", args=[obj.media_obj.identifier]
+            ),
```
sarayourfriend, line 126: Considering how many of these are mappings of properties on `obj.media_obj`, it might be clearer (especially in the template) to pass "media_obj" as a context variable instead. It would make it clearer in the template, for example, where `url`, `creator`, `source`, etc are coming from.

```suggestion
        additional_data = {
            "other_reports": self.get_other_reports(obj),
            "media_obj": obj.media_obj,
            "tags": tags_by_provider,
            "reverse_media_url": reverse(
                f"admin:api_{self.media_type}_change", args=[obj.media_obj.identifier]
            ),
```

api/api/admin/media_report.py
```diff
@@ -0,0 +1,170 @@
...
+            "foreign_landing_url": obj.media_obj.foreign_landing_url,
+            "description": obj.media_obj.meta_data.get("description", ""),
+            "title": obj.media_obj.title,
+            "provider": obj.media_obj.provider,
+            "source": obj.media_obj.source,
+            "creator": obj.media_obj.creator or obj.media_obj.creator_url,
+            "creator_url": obj.media_obj.creator_url,
+            "url": obj.media_obj.url,
+            "tags": tags_by_provider,
+            "reverse_media_url": reverse(
```
sarayourfriend, line 124: Suggesting a less ambiguous/indirect name:

```suggestion
            "media_admin": reverse(
```

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -0,0 +1,159 @@
...
+      <img src="{% url media_type|add:'-thumb' identifier=identifier %}" alt="Media Image" class="blur-image" height="300"
+           onclick="toggleBlur(this)"
+           onerror="this.onerror=null;this.src='{{ url }}';"
+           style="cursor: pointer;">
+      <p>Show content</p>
+    </div>
+
+  {% elif media_type == 'audio' %}
+    <audio controls>
+      <source src="{{ url }}" type="audio/mpeg">
```
sarayourfriend, line 60: Is audio/mpeg safe? What if it's any of the other codecs supported ([webm, for example](https://api.openverse.engineering/v1/audio/?extension=webm))? I think the browser can be left to deduce the correct type on its own :thinking:. Is `type` a required attribute for audio `source` element?

```suggestion
      <source src="{{ url }}" type="audio/mpeg">
```

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -0,0 +1,159 @@
...
+      Your browser does not support the audio element.
+    </audio>
+  {% endif %}
+
+  <div class="form-row field-created_at">
+      <div class="flex-container">
+        <label>Title and creator</label>
+        <div class="readonly">
+          <p>{{ title }} {% if creator_url %}by <a href="{{creator_url}}">{{creator}}</a> {% else %} {{creator}} {% endif %} </p></div>
```
sarayourfriend, line 69: Can we lean on the attribution generation here to make things easier? It already resolves the creator_url/creator resolution, printing the title, source, etc.

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -0,0 +1,159 @@
...
+      </div>
+  </div>
+  <div class="form-row field-created_at">
+      <div class="flex-container">
+        <label>Media links</label>
+        <div class="readonly">
+          <p><a href="{{foreign_landing_url}}">Source ({{source}})</a></p>
+          <p><a href="https://openverse.org/{{media_type}}/{{identifier}}">openverse.org</a></p>
+          <p><a href="{% url media_type|add:'-detail' identifier=identifier %}">openverse API</a></p>
```
sarayourfriend, line 78:
```suggestion
          <p><a href="{% url media_type|add:'-detail' identifier=identifier %}">Openverse API</a></p>
```

sarayourfriend (approved):
I've left some comments, and I am approving despite the issues @AetherUnbound identified. My sense about the context using `media_obj` being clearer is strong, but it isn't a blocker, and we can refactor this after the fact without issue. I don't want to prolong this PR's review on these problems.

The problems Madison identified are indeed problems, but I feel comfortable saying we can address them in fast follow issues, rather than in this PR, and it won't cause harm to the future issues in this milestone that depend on this baseline implementation existing. In the interest of unblocking that work, and recognising this PR has already been long in review, I'd prefer you created follow-up issues for those, and we merged this either as-is or with only minor modifications (up to you how much).

@AetherUnbound deferring to you on how much you feel are definite blockers vs things that can be fixed in fast-follows.

api/api/templates/admin/api/media_report/change_form.html
obulat: Removed, doesn't cause any warnings without it.

api/api/templates/admin/api/media_report/change_form.html
obulat: Apparently, we don't have a function to get the html attribution, only the text one that does not add creator and work urls.
For some reason, I was under the impression that all the items mentioned in the IP need to be on separate lines. Now, I updated the form to display the media details more compactly. For attribution, I created a new sub-template.

api/api/admin/media_report.py
obulat: Thank you for this suggestion, it's much cleaner now!

api/api/admin/media_report.py
obulat: I removed this link entirely because the link this creates is incorrect, and there's no need to add the link anyway.

obulat:
I moved as much as possible to the templates, simplified the context object passed to the template, removed the link to the reverse image admin page when there were no reports (it wasn't in the IP, and the link to media change form is incorrect).

Here's an updated form:

ImageReport object _3_ _ Change image report _ Django site admin

sarayourfriend:
Looks awesome, @obulat

AetherUnbound (approved):
Fantastic, I'm with @sarayourfriend on the other changes, those can be done in follow up if/as necessary. Thanks so much Olga, this looks excellent!

api/api/templates/admin/api/media_report/change_form.html
sarayourfriend: That'll be the database `id`, primary key column.

PR #4176, opened 2024-04-22
Don't show sources if their count has not changed

AetherUnbound:
## Description

Follow-up to #4067

Here's the report from the most recent audio data refresh:

image

The current logic is set to show all sources, regardless of whether there was a change or not. This PR changes the logic so only sources that have increased or decreased in size are shown in the final report. While this doesn't have a huge effect on audio data refreshes, it will for image data refreshes since we have significantly more sources there.

## Testing Instructions

Updated tests should pass. Alternatively, run the data refresh locally after `just api/init` and see that no sources are listed in the completion message because none of the counts changed. 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/tests/dags/data_refresh/test_reporting.py
+++ b/catalog/tests/dags/data_refresh/test_reporting.py
@@ -19,6 +19,7 @@ from data_refresh.reporting import report_record_difference, report_status
         [
             {"src1": 4, "src2": 21},
             {"src1": 4},
+            # Unchanged source count shouldn't show up
             ["25 → 4", "-21 (-84.0%", "`src2`:-21"],
         ],
         [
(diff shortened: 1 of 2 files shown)


catalog/tests/dags/data_refresh/test_reporting.py
```diff
@@ -19,7 +19,7 @@
         [
             {"src1": 4, "src2": 21},
             {"src1": 4},
-            ["25 → 4", "-21 (-84.0%", "`src1`:+0", "`src2`:-21"],
+            ["25 → 4", "-21 (-84.0%", "`src2`:-21"],
```
krysal, line 22: It would be nice to add a note that unchanged sources are omitted.

krysal (approved):
Nice.

dhruvkb (commented):
+1 for @krysal's suggestions. Do we need to handle a case where all sources are unchanged? That seems unlikely to ever occur at all.

AetherUnbound:
> Do we need to handle a case where all sources are unchanged? That seems unlikely to ever occur at all.

Only in local testing I would imagine! I don't think it's worth having logic for that case, personally. And I've made the message about unchanged sources ambiguous enough that it should cover the case where all sources are absent too

obulat (approved):
LGTM!

PR #4197, opened 2024-04-24
Convert OperationalError into unhealthy health check response

### LINKED ISSUE #3746: OperationalError: [Errno -2] Name or service not known
Sentry Issue: [OPENVERSE-API-CP](https://openverse.sentry.io/issues/4933375855/?referrer=github_integration)

```
OperationalError: [Errno -2] Name or service not known
(2 additional frame(s) were not displayed)
...
  File "django/db/backends/base/base.py", line 270, in connect
    self.connection = self.get_new_connection(conn_params)
  File "django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
  File "django/db/backends/postgresql/base.py", line 275, in get_new_connection
    connection = self.Database.connect(**conn_params)
  File "psycopg/connection.py", line 729, in connect
    attempts = conninfo_attempts(params)
  File "psycopg/_conninfo_attempts.py", line 45, in conninfo_attempts
    raise e.OperationalError(str(last_exc))

OperationalError: [Errno -2] Name or service not known
(15 additional frame(s) were not displayed)
...
  File "api/views/health_views.py", line 55, in get
    self._check_db()
  File "api/views/health_views.py", line 35, in _check_db
    connection.ensure_connection()
```

AetherUnbound:
## Fixes

Fixes #3746

## Description

This PR captures `OperationalError`s that are emitted when the database is no longer available and the `connections.ensure_connection()` function is called as part of the health check for the API. Rather than raise an uncaught exception, this instead converts the exception into an HTTP 503 response (similar to if the Elasticsearch cluster is not in a healthy state).

## Testing Instructions

Tests should pass, but additionally, you can observe this operationally:

1. Run `just a` to start the API stack
1. Run `docker stop openverse-db-1`
1. Visit http://localhost:50280/healthcheck/ and observe that a 503 instead of a generic 500 is returned

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/health_views.py
+++ b/api/api/views/health_views.py
@@ -36,9 +36,7 @@ class HealthCheck(APIView):
         try:
             connection.ensure_connection()
         except OperationalError as err:
-            raise HealthCheckException(
-                code="postgres", detail=f"Database unavailable: {err}"
-            )
+            raise HealthCheckException(f"postgres: {err}")
 
     @staticmethod
     def _check_es() -> None:
@@ -50,12 +48,10 @@ class HealthCheck(APIView):
         es_health = settings.ES.cluster.health(timeout="5s")
 
         if es_health["timed_out"]:
-            raise HealthCheckException(code="elasticsearch", detail="es_timed_out")
+            raise HealthCheckException("elasticsearch: es_timed_out")
 
         if (es_status := es_health["status"]) != "green":
-            raise HealthCheckException(
-                code="elasticsearch", detail=f"es_status_{es_status}"
-            )
+            raise HealthCheckException(f"elasticsearch: es_status_{es_status}")
 
     def get(self, request: Request):
         if "check_es" in request.query_params:
--- a/api/api/views/health_views.py
+++ b/api/api/views/health_views.py
@@ -6,10 +6,16 @@ from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+from psycopg import OperationalError
+
 from api.utils.throttle import ExemptOAuth2IdRateThrottle, HealthcheckAnonRateThrottle
 
 
 class ElasticsearchHealthcheckException(APIException):
+    status_code = status.HTTP_503_SERVICE_UNAVAILABLE
+
+
+class DatabaseHealthCheckException(APIException):
     status_code = status.HTTP_503_SERVICE_UNAVAILABLE
 
 
@@ -32,7 +38,10 @@ class HealthCheck(APIView):
 
         Returns nothing if everything is OK, throws error otherwise.
         """
-        connection.ensure_connection()
+        try:
+            connection.ensure_connection()
+        except OperationalError as err:
+            raise DatabaseHealthCheckException(str(err))
 
     @staticmethod
     def _check_es() -> None:
--- a/api/api/views/health_views.py
+++ b/api/api/views/health_views.py
@@ -10,11 +10,7 @@ from rest_framework.views import APIView
 from api.utils.throttle import ExemptOAuth2IdRateThrottle, HealthcheckAnonRateThrottle
 
 
-class ElasticsearchHealthcheckException(APIException):
-    status_code = status.HTTP_503_SERVICE_UNAVAILABLE
-
-
-class DatabaseHealthCheckException(APIException):
+class HealthCheckException(APIException):
     status_code = status.HTTP_503_SERVICE_UNAVAILABLE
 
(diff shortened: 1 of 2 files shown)


api/api/views/health_views.py
```diff
@@ -13,6 +14,10 @@ class ElasticsearchHealthcheckException(APIException):
     status_code = status.HTTP_503_SERVICE_UNAVAILABLE
 
 
+class DatabaseHealthCheckException(APIException):
```
sarayourfriend, line 17: A small change to ask to combine these, and use a different error code to distinguish between the two. The benefit of this is also in the test, where we can actually assert the response message aligns with the expected failure. Right now, the test could absolutely just not work due to a small misconfiguration or misunderstanding of the implementation details of a particular check, and we'd never know (if another check failed), because they all just check for 503, which has at least two meanings from this view.

Bonus points if we can remove the inconsistently named ES class's "Healthcheck" (vs "HealthCheck" to match the view).

sarayourfriend (changes requested):
LGTM but a request to unify the response interface, which will bear out in more dependable/less implementation dependent tests.

api/api/views/health_views.py
AetherUnbound: When you say different error code, does that mean different HTTP status? If so, which one would would you recommend for each? Also d'oh, apologies about the misnaming!

sarayourfriend: > Also d'oh, apologies about the misnaming!

You misunderstood, I think you got the naming _right_, at least in that it actually matches the view class's name. I don't know which is "correct" HealthCheck or Healthcheck (I think of it as one word but my spell checker doesn't agree :shrug:) but consistency with the view is better, I think.

> When you say different error code, does that mean different HTTP status?

No, sorry, I mean the `APIException` class's `code` attribute: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L106

It can be set in the constructor. So then we'd just have a generic `HealthCheckException`, and when raising it, pass a meaningful code and/or detail to the constructor:

```py
raise HealthCheckException(code='elasticsearch', detail='Yellow cluster health')

raise HealthCheckException(code='postgres', detail='Unreachable')
```

And so forth. Main thing for this PR would be to unify them and distinguish based on the code, not necessarily that the whole method should be updated to use it correctly. Up to you how far to go with it.

AetherUnbound: Ahhh, thank you for the explanation! That makes sense, I'll go ahead with that change 🙂

AetherUnbound: Hmm, what's interesting is that the provided `code` value doesn't show up in the response sent back to the client, only `detail` 🤔 I'll have to look into why that's the case.

AetherUnbound:
@sarayourfriend I've made the change you suggested, but it still doesn't appear that `code` is showing up in the actually response JSON which gets returned from the API 😕 I've tried poking around in the DRF innards, but from what I can tell `code` is being set correctly.

api/api/views/health_views.py
sarayourfriend: Maybe just needs to be in the detail then. `code` might genuinely just be for internal server side use.

sarayourfriend:
Code might be just for internal server side use. If only detail is exposed in the response then it needs to go there.

dhruvkb (approved):
LGTM! 

Although unlikely to occur, this `detail` (`str`) cannot support multiple dependency services failing simultaneously.

AetherUnbound:
> Although unlikely to occur, this `detail` (`str`) cannot support multiple dependency services failing simultaneously.

Ahh, you're right. Should we make an issue for this? It could be a good contribution opportunity 🙂

sarayourfriend (approved):
Fantastic!

sarayourfriend:
> Ahh, you're right. Should we make an issue for this? It could be a good contribution opportunity 🙂

Agreed it's a different issue, it's an existing problem in the implementation.

My take though, is that it doesn't matter. If the API is failing at _these_ points and we don't have monitors that let us know e.g., Elasticsearch is down, then we're really letting things slide as far as monitoring and observability go.

This endpoint is, to me, most useful to confirm API configuration, not to monitor Postgres and ES, and in that sense, I really don't think it matters if it doesn't give you _all_ the potential problems at once.

Issue #3885, opened 2024-03-07
Backfill `license_url` field for images where it's null in the meta_data


krysal:
## Problem

As a requirement for #703, we must fill the `metadata.license_url` field for images without it. At the moment of writing this issue, the number amounts to 97.7 million in the upstream DB.

```sql
SELECT COUNT(identifier) FROM image WHERE meta_data->>'license_url' IS NULL;
+----------+
| count    |
|----------|
| 97775299 |
+----------+
SELECT 1
Time: 2732.486s (45 minutes 32 seconds), executed in: 2732.465s (45 minutes 32 seconds)
```

## Description

This potentially could be solved by a one-off DAG that gets the license URL computed from the `license` and `license_version` fields. @obulat made one DAG previously for filling one of the cases when the license URL is null, see #1005.

## Additional context

See previous work at #1565.

krysal:
Reopening this since it's still being work out.

krysal:
The DAG to backfill was triggered yesterday at 14:59:16 UTC and ended successfully after ~8 hours. This is completed. However, we will keep the DAG for some time to see if https://github.com/WordPress/openverse/issues/4318 keeps repeating.

PR #4198, opened 2024-04-24
Warn on `license_url` computation in the API

### LINKED ISSUE #703: Remove `license_url` computation
## Problem

We are currently using code to compute `license_url` for some images that do not have it in the database. This code should become unnecessary when we normalize data in the database.
This issue is blocked by https://github.com/WordPress/openverse-catalog/issues/511

## Description

We should remove the code that computes the `license_url` as it should be available in the `meta_data.license_url` field for all media items in the database.

krysal:
## Fixes

Related to #703 by @obulat

## Description

In the interest of closing this PR, I changed the approach only to warn when the license is computed, as suggested by @obulat. The #703 issue will addressed when we confirm #4318 is not an issue anymore.

---

### Old description
My understanding is that the computation from `openverse_attribution.license.License` can be removed but we still need the property.

This PR should not be merged until the updated `add_license_url` run is completed and it's confirmed the `image` table in the Catalog does not have more `NULL` values for this field. 

If you are wondering about `audio` I confirmed all items have this field set to something else than `NULL`.

```sql
openledger> SELECT COUNT(identifier) FROM audio WHERE meta_data->>'license_url' IS NULL;
+-------+
| count |
|-------|
| 0     |
+-------+
SELECT 1
Time: 17.712s (17 seconds), executed in: 17.695s (17 seconds)
```

## Testing Instructions

Run the API and remove the license_url of an image. Then, visit its media page and check the logs.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/models/media.py
+++ b/api/api/models/media.py
@@ -127,8 +127,9 @@ class AbstractMedia(
             return url
 
         logger.warning(
-            f"The {self.__class__.__name__.lower()} with identifier={self.identifier} "
-            f"has no `license_url` in `meta_data`."
+            "Media item missing `license_url` in `meta_data`",
+            media_class=self.__class__.__name__,
+            identifier=self.identifier,
         )
         try:
             return License(self.license.lower(), self.license_version).url
--- a/api/api/models/media.py
+++ b/api/api/models/media.py
@@ -125,6 +125,11 @@ class AbstractMedia(
 
         if self.meta_data and (url := self.meta_data.get("license_url")):
             return url
+
+        logger.warning(
+            f"The {self.__class__.__name__.lower()} with identifier={self.identifier} "
+            f"has no `license_url` in `meta_data`."
+        )
         try:
             return License(self.license.lower(), self.license_version).url
         except ValueError:


sarayourfriend:
@krysal what is the status of this PR and this particular blocker? Is the PR still blocked? Should it be labelled as such?

> This PR should not be merged until the updated add_license_url run is completed and it's confirmed the image table in the Catalog does not have more NULL values for this field.

obulat:
@krysal, what if instead of removing the computation completely we leave it in but add a warning log. Then, if there are no warnings in a longish period of time (a month?), we can remove the computation?
This would not be blocked by the DAG, and could be merged now.

krysal:
@sarayourfriend It's blocked pending completion of #3885, which I thought would be a relatively quick task, but some problems arose.

> @krysal, what if instead of removing the computation completely we leave it in but add a warning log. Then, if there are no warnings in a longish period of time (a month?), we can remove the computation?
This would not be blocked by the DAG, and could be merged now.

That is a good idea! :) I'll adapt the PR for that 👍

AetherUnbound:
@krysal I'm going to add the blocked label for this PR while work on #3885 is being done

stacimc (approved):
This looks fine to me, I was able to see the log after removing the meta_data from a record locally. It's been a minute since I've looked at this -- can you remind me what cases we're trying to catch here? I thought the catalog should be enforcing setting the licenses now, and that incorrect cases had already been backfilled, but maybe I'm misremembering. 

If there isn't one already, can we make an issue for checking the production logs so that it does not get forgotten?

api/api/models/media.py
```diff
@@ -125,6 +125,11 @@ def license_url(self) -> str | None:
 
         if self.meta_data and (url := self.meta_data.get("license_url")):
             return url
+
+        logger.warning(
+            f"The {self.__class__.__name__.lower()} with identifier={self.identifier} "
+            f"has no `license_url` in `meta_data`."
+        )
```
dhruvkb, line 132: Because we have structured logging, we can use that to get output that's more parseable than interpolated text.

```suggestion
        logger.warning(
            "Media item missing `license_url` in `meta_data`",
            class=self.__class__.__name__.lower(),
            identifier=self.identifier,
        )
```

dhruvkb (approved):
LGTM! Added a suggestion to make logging more structured but it's not a blocker in my view.

api/api/models/media.py
krysal: This is a fantastic suggestion ✨  Thanks, @dhruvkb!

krysal:
> This looks fine to me, I was able to see the log after removing the meta_data from a record locally. It's been a minute since I've looked at this -- can you remind me what cases we're trying to catch here? I thought the catalog should be enforcing setting the licenses now, and that incorrect cases had already been backfilled, but maybe I'm misremembering.
> 
> If there isn't one already, can we make an issue for checking the production logs so that it does not get forgotten?

@stacimc I'm also hoping we don't see more incorrect cases but I created #4318 to wait and confirm that. There is also more explanation of the reasoning.

PR #4220, opened 2024-04-29
Add "sa" as valid license and map "publicdomain" to "pdm"

dhruvkb:
## Description

Some files specify the license as "publicdomain". Our standard slug for public domain content is "pdm".

Some files in the API have licenses like "publicdomain" and "sa". This PR adds these licenses to the `openverse-attribution` package. This PR also adds some exception handling around the use of `License` (it should not be necessary anymore, but still).

A more comprehensive update for `License` is #4250, this PR is the baseline fix that will enable us to deploy the API again.

## Testing Instructions

0. Change some rows in the DB to have "publicdomain" or "sa" as the license.
1. Go to their detail page.
2. You should not get 500 errors.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/serializers/media_serializers.py
+++ b/api/api/serializers/media_serializers.py
@@ -741,9 +741,11 @@ class MediaSerializer(BaseModelSerializer):
         output["license"] = output["license"].lower()
 
         if output.get("license_url") is None:
-            output["license_url"] = License(output["license"]).url(
-                output["license_version"]
-            )
+            try:
+                lic = License(output["license"])
+                output["license_url"] = lic.url(output["license_version"])
+            except ValueError:
+                pass
 
         # Ensure URLs have scheme
         url_fields = ["url", "creator_url", "foreign_landing_url"]
--- a/api/api/utils/watermark.py
+++ b/api/api/utils/watermark.py
@@ -199,7 +199,7 @@ def _print_attribution_on_image(img: Image.Image, image_info):
         image_info["title"],
         image_info["creator"],
         image_info["license_version"],
-        license_url=False,
+        False,
     )
     text = _fit_in_width(text, font, new_width)
     attribution_height = _get_attribution_height(text, font)
--- a/api/api/utils/watermark.py
+++ b/api/api/utils/watermark.py
@@ -9,7 +9,7 @@ from rest_framework import status
 from rest_framework.exceptions import APIException
 
 import requests
-from openverse_attribution.attribution import get_attribution_text
+from openverse_attribution.license import License
 from PIL import Image, ImageDraw, ImageFont, UnidentifiedImageError
 from sentry_sdk import capture_exception
 
@@ -174,6 +174,11 @@ def _print_attribution_on_image(img: Image.Image, image_info):
     :return: return the framed and attributed image
     """
 
+    try:
+        lic = License(image_info["license"])
+    except ValueError:
+        return img
+
     width, height = img.size
     smaller_dimension = _smaller_dimension(width, height)
 
@@ -190,8 +195,7 @@ def _print_attribution_on_image(img: Image.Image, image_info):
 
     font = ImageFont.truetype(_get_font_path(), size=font_size)
 
-    text = get_attribution_text(
-        image_info["license"],
+    text = lic.attribution(
         image_info["title"],
         image_info["creator"],
         image_info["license_version"],
--- a/api/api/models/media.py
+++ b/api/api/models/media.py
@@ -8,7 +8,6 @@ from django.urls import reverse
 from django.utils.html import format_html
 
 from elasticsearch import Elasticsearch, NotFoundError
-from openverse_attribution.attribution import get_attribution_text
 from openverse_attribution.license import License
 
 from api.constants.moderation import DecisionAction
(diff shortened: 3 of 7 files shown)


dhruvkb:
We also do not handle "sa". The closest to that is "by-sa".

api/api/utils/watermark.py
```diff
@@ -190,8 +190,13 @@ def _print_attribution_on_image(img: Image.Image, image_info):
 
     font = ImageFont.truetype(_get_font_path(), size=font_size)
 
+    lic = image_info["license"]
```
obulat, line 193: Maybe add a "normalize license" function somewhere in utils that would convert these to the expected values? Then it will be easier to remove it once the catalog data has been fixed.

I would also prefer a try/catch around the `License` calls for now to ensure that we don't return 500 if there are any other problems with the production data.

obulat: We could probably add a very permissive implementation of something like `get_License` function for now, watch the logs for errors, and open an issue to make the function stricter once we know that the data is normalized.
This would allow us to deploy the API with this fix without having to worry about rollback.

api/api/utils/watermark.py
```diff
@@ -190,8 +192,9 @@ def _print_attribution_on_image(img: Image.Image, image_info):
 
     font = ImageFont.truetype(_get_font_path(), size=font_size)
 
+    lic_slug, _ = get_license_slug_enum(image_info["license"])
```
obulat, line 195: Should this also handle the case when `None` is returned?

obulat: If we return the image as is, this would be confusing to the user. Maybe return an error? 
Admittedly, this endpoint is deprecated, and we probably never have such a case. But if we do, with the current approach, we will return 500.

api/api/utils/license.py
```diff
@@ -0,0 +1,26 @@
+import logging
+
+from openverse_attribution.license import License
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_license_slug_enum(slug: str) -> tuple[str, License] | None:
```
stacimc, line 9: Could we use a namedtuple here to make it clearer how this is being used in the serializers?

api/api/utils/license.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Get the corrected license slug, and ``License`` enum for the given,
+    potentially invalid, license slug.
+
+    :param slug: the potentially invalid license slug
+    :return: the valid license slug
+    """
+
+    known_incorrect_licenses = {"publicdomain": "pdm", "sa": "by-sa"}
```
stacimc, line 18: This could go in `api/licenses/constants.py`.

api/api/serializers/media_serializers.py
```diff
@@ -740,10 +740,10 @@ def to_representation(self, *args, **kwargs):
         # Ensure license is lowercase
         output["license"] = output["license"].lower()
 
-        if output.get("license_url") is None:
-            output["license_url"] = License(output["license"]).url(
-                output["license_version"]
-            )
+        if not output.get("license_url") and (
+            lic_slug_enum := get_license_slug_enum(output["license"])
+        ):
```
stacimc, line 745: What happens if `get_license_slug_enum` returns None?

stacimc (changes requested):
Is this a problem that should be addressed in the catalog where we're setting those [slugs](https://github.com/WordPress/openverse/blob/main/catalog/dags/common/licenses/constants.py)?

I'm also curious about mapping `sa` to `by-sa`. In the catalog we set these for https://creativecommons.org/licenses/sa/1.0/ and https://creativecommons.org/licenses/sa/2.0/jp, both of which I get results for (although there's a notice that they're retired). I'm not really familiar with how this works, but is it accurate to map that to `by-sa`? Should we just not ingest those records, instead?

dhruvkb:
Thanks for catching and explaining this @stacimc because mapping "sa" to "by-sa" would not have been right! I have updated the code in 179182f to treat "sa" accurately as a valid, but deprecated license.

api/api/utils/license.py
dhruvkb: After 179182f, this is no longer needed as there is only the case of "publicdomain" to account for.

obulat (changes requested):
@dhruvkb, I thought this could be a quick fix that would enable us to deploy the API without reverting the pdm changes. However, after @stacimc's note, I realize that we should carefully review the handling of licenses to ensure that we don't accidentally change the values.
Catalog has a very comprehensive list of license pair and license url correspondences, and `publicdomain` there is a valid license (mark) value: https://github.com/WordPress/openverse/blob/2051936d08b31c024873051fd3b7725635579e47/catalog/dags/common/licenses/constants.py
I think we should move it to the attribution package, and update it to accept all possible values.

obulat (approved):
LGTM!

zackkrida (approved):
LGTM for the sake of our deployments. I would generally prefer we use `license` instead of `lic` in the model and serializer but that's a style nitpick and not blocking.

dhruvkb:
`license` shadows built-in name.

PR #4225, opened 2024-04-29
Expand documentation on API media properties

### LINKED ISSUE #4024: Write docstrings to document API fields
## Description

For fields with complications, or edge cases, or noteworthy details, write descriptions in the API model docstrings that can be parsed by the documentation generation script from #4023.

These docstrings should go into a lot more detail about the nature and shape of data in the `JSONField`-type fields.

Also document what the API expects from thee catalog, what it wrangles and generates automatically.

AetherUnbound:
## Fixes

Fixes #4024 by @dhruvkb

## Description

This PR fills out documentation information for nearly all the API fields. Where possible, I've added more information about what should be contained in certain `JSONField`s.

I also added some automatic linking of related model docstrings to the generated documentation output. It seemed prudent to leverage that over repeating the documentation as help text which may go out of date.

A migration has been added for this change, but per `sqlmigrate api 0060`, it's all no-ops:

```
BEGIN;
--
-- Alter field alt_files on audio
--
-- (no-op)
--
-- Alter field last_synced_with_source on audio
--
-- (no-op)
--
-- Alter field meta_data on audio
--
-- (no-op)
--
-- Alter field removed_from_source on audio
--
-- (no-op)
--
-- Alter field tags on audio
--
-- (no-op)
--
-- Alter field view_count on audio
--
-- (no-op)
--
-- Alter field watermarked on audio
--
-- (no-op)
--
-- Alter field last_synced_with_source on image
--
-- (no-op)
--
-- Alter field meta_data on image
--
-- (no-op)
--
-- Alter field removed_from_source on image
--
-- (no-op)
--
-- Alter field tags on image
--
-- (no-op)
--
-- Alter field view_count on image
--
-- (no-op)
--
-- Alter field watermarked on image
--
-- (no-op)
COMMIT;
```

## Testing Instructions

Check out the rendered document and make sure all the new additions make sense and are correct!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/models/audio.py
```diff
@@ -1,3 +1,5 @@
+from textwrap import dedent as d
```
sarayourfriend, line 1: Kind of a nit, but the `d` alias is a bit obscure and roundabout compared to using the real name. What's the reason to alias this, does it aid in reading the code? I could see it being nice when writing, but after that, it's just a bit more confusing. Without the context of having just read this import, it looks kind of random in the code itself, and in the code itself I'd almost assume it was some kind of translation/gettext utility based on how similar single-letter string functions work in that space (including in our frontend).

api/api/models/audio.py
```diff
@@ -192,7 +194,16 @@ class Audio(AudioFileMixin, AbstractMedia):
...
+        help_text=d("""
+        JSON object containing information on alternative audio files. Each object
+        is expected to contain:
+
+        - `url`: URL reference to the file
+        - `filesize`: File size in bytes
+        - `filetype`: Extension of the file
+        - `bit_rate`: Bitrate of the file in bits/second
+        - `sample_rate`: Sample rate of the file in bits/second
+        """),
```
sarayourfriend, line 206: It's a shame dedent is needed for `"""` strings.

I'm accustomed to seeing the use of implicit string concatenation in these fields (I think we do elsewhere), making the extra utility unnecessary (and therefore, not needing to remember or think about it). It also doesn't introduce the surprising newline after "Each object", which will render differently depending on context. Markdown will ignore it, but other contexts do not necessarily.

No real preference here either way, trade-offs in both cases to be sure.

api/api/models/media.py
```diff
@@ -65,19 +70,37 @@ class AbstractMedia(
...
+        db_index=True,
+        help_text="The date the media was last updated from the upstream source.",
+    )
+    removed_from_source = models.BooleanField(
+        default=False,
+        help_text="Whether or not the media has been removed from the upstream source.",
+    )
+
+    view_count = models.IntegerField(
+        blank=True, null=True, default=0, help_text="Vestigial field, purpose unknown."
```
sarayourfriend, line 85: Should we leave it out then? It's odd to include such a thing in public documentation, especially in a versioned API...

api/api/models/media.py
```diff
@@ -65,19 +70,37 @@ class AbstractMedia(
...
+    last_synced_with_source = models.DateTimeField(
         blank=True,
         null=True,
-        default=0,
+        db_index=True,
+        help_text="The date the media was last updated from the upstream source.",
+    )
+    removed_from_source = models.BooleanField(
+        default=False,
+        help_text="Whether or not the media has been removed from the upstream source.",
```
sarayourfriend, line 81: Editorial nit:

```suggestion
        help_text="Whether the media has been removed from the upstream source.",
```

sarayourfriend (commented):
I've included some editorial nits. Take them or leave them, none are critical in any sense, mostly just oriented towards simplifying the text.

api/api/models/image.py
```diff
@@ -108,6 +108,13 @@ class Meta:
 
 
 class ImageReport(AbstractMediaReport):
+    """
+    Represents images that have been reported by users.
```
sarayourfriend, line 112: Editorial nit, though stronger on this one. The issue of the current phrasing is a suggestion that the model represents something about an image (or work generally, extrapolated to audio report), whereas it represents the user-submitted report of an image. The same issue exists on audio report.

```suggestion
    User-submitted report of an image.
```

sarayourfriend (commented):
I question the use of "Represents" as a necessary preamble to model doc strings. I suspect this is to appease the documentation linter, but these aren't imperative classes, they're declarative data representations. As such, the imperative tone is unnecessarily wordy and even strangely abstract. I'd strongly recommend dropping that conceit from model doc strings and instead stating in simple terms what the model _is_ rather than what it _does_. "Represents user-submitted reports of audio tracks" becomes "User-submitted reports of audio tracks".

api/api/models/media.py
```diff
@@ -88,7 +111,18 @@ class AbstractMedia(
         help_text="The top-level classification of this media file.",
     )
 
-    meta_data = models.JSONField(blank=True, null=True)
+    meta_data = models.JSONField(
+        blank=True,
+        null=True,
+        help_text=d(
+            """
```
sarayourfriend, line 118: Consistency nit.

```suggestion
        help_text=d("""
```

api/api/models/media.py
```diff
@@ -65,19 +70,37 @@ class AbstractMedia(
...
+    view_count = models.IntegerField(
+        blank=True, null=True, default=0, help_text="Vestigial field, purpose unknown."
     )
 
     tags = models.JSONField(
         blank=True,
         null=True,
-        help_text="Tags with detailed metadata, such as accuracy.",
+        help_text=d(
+            """
```
sarayourfriend, line 92: Consistency nit.

```suggestion
        help_text=d("""
```

This text is not necessary, in my opinion, and duplicates the documentation of the tag schema. Note the "tags" schema embedded below.

20240430_09h07m43s_grim

If we actually want to improve this field's documentation, we should remove the suggestion of particular individual properties from this help text, and instead augment whatever improvements we wish to represent on the Tag schema.

sarayourfriend (commented):
Okay, this is my actual final review. Only the following issues are blockers, the rest are soft suggestions or merely editorial nits:

- We should not duplicate documentation between referent fields and the thing referred. For example, "tags" should not duplicate the documentation on the tag schema and especially should not concern itself with the properties of the tag. The tag's documentation bears this responsibility. For a given work, "tags" should be documented only as "The tags given to a work" (as an example).
- Do not alias `dedent` to `d`. It hurts readability. I think the instances it is used, however, would be removed by addressing the previous item.
- Drop the imperative tone from model doc strings, preferring instead to document what a model _is_ rather than what it _does_, in particular because what it _does_ is implicitly "representation", due to the fact that it is a data model. They do not have operative meaning, only representational, despite the fact that individual methods _might_ be operative, in particular, `save`. Those methods of course make sense to have the imperative tone, but on a data model class it's at best strangely abstract and at worst complicates documentation for the sake of a "style" that shouldn't be universally applied.
  - This will probably require modifying the linting rules. However, I consider it a blocker because the additions in this PR are wide-spread, and addressing this after the fact is additional work we can avoid now. On top of which, it is worse documentation that we would produce if we ignored the linter in this case and used direct terms relevant for the context.

sarayourfriend (changes requested):
Meant to "request changes", seems like I can't get review submission right at all on this one... sorry!

api/api/models/media.py
AetherUnbound: When you say leave it out, do you mean not include any documentation for it? Since this documentation is also for us as maintainers, I think having this kind of information is useful for us too.

api/api/models/audio.py
```diff
@@ -192,7 +194,16 @@ class Audio(AudioFileMixin, AbstractMedia):
...
-        help_text="JSON describing alternative files for this audio.",
+        help_text=d("""
+        JSON object containing information on alternative audio files. Each object
+        is expected to contain:
+
+        - `url`: URL reference to the file
+        - `filesize`: File size in bytes
+        - `filetype`: Extension of the file
+        - `bit_rate`: Bitrate of the file in bits/second
+        - `sample_rate`: Sample rate of the file in bits/second
```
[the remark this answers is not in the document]
AetherUnbound: Part of the reason for documenting this here is that it will show up in the [API Media Properties page](https://docs.openverse.org/meta/media_properties/api.html). Since this is on the model and not the serializer, I'm not sure it'd show up on the OpenAPI documentation. I'll have to do some experimentation with that.

api/api/models/media.py
AetherUnbound: Ah, I've been using `black` to format files which was forcing this change (and it's valid under `ruff format` too). Swapped to using the latter locally and corrected this!

AetherUnbound:
@sarayourfriend I believe I've addressed most of your feedback, except around the JSON fields help text. This PR is centered around adding documentation that's most helpful _for maintainers_, which will be rolled up into the [API media properties](https://docs.openverse.org/meta/media_properties/api.html) page. This, as far as I can tell, is _not_ the documentation that's surfaced on our OpenAPI documentation, as that information comes from the serializers themselves. I think having different intentions behind each kind of documentation is important - the serializer help text is what gets surfaced to users, and the model help text is what's useful for maintainers in discerning what's actually present in the database. This PR's purview is the latter case.

api/api/models/audio.py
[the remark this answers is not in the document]
sarayourfriend: Oops, I totally missed that! You're right. I still don't think it should be documented here and that it should be a reference to some shared Tag documentation, but ignore the part I said about the OpenAPI documentation. It won't have any side-effect there.

api/api/models/media.py
sarayourfriend: Yes, sorry, I had the wrong context in mind for this. Resolving.

sarayourfriend (approved):
LGTM. As mentioned in the thread, I had the wrong downstream context in mind when reviewing these changes, and didn't realise it wouldn't change the OpenAPI schema docs.

I still wish there was a better and clearer way to document the contents of the JSON field, but I guess that's what the JSON field is always going to be anyway: a blob of whatever gets put there, rather than a particular protocol/dataclass/etc.

Looks to me as is now :+1: Apologies for the confusion from my misunderstanding of the downstream changes from this.

api/api/management/commands/documentmedia.py
```diff
@@ -293,6 +299,11 @@ def generate_notes(model: str, fields: list[FieldInfo]) -> tuple[str, set[str]]:
         if not field.is_relation and field.value_info.help_text:
             field_output += f"**Help text:** {field.value_info.help_text}\n\n"
             record = True
+        if field.is_relation and field.relation_info.doc:
+            field_output += (
+                f"**Model docstring:** {dedent(field.relation_info.doc)}\n\n"
```
dhruvkb, line 304: Would be nice to know which model this docstring is for without going back to the table to get that info.

```suggestion
                f"**`{field.relation_info.to}` docstring:** {dedent(field.relation_info.doc)}\n\n"
```

dhruvkb (approved):
LGTM, thanks for filling out all the help texts!

api/api/management/commands/documentmedia.py
AetherUnbound: Great idea, done in [d11bee1](https://github.com/WordPress/openverse/pull/4225/commits/d11bee165402e5da1d1cb85e8ee9b4d0e04f7dfa)!

PR #4231, opened 2024-04-30
Initialize Provider store data, UI store and flags store from cookies in a plugin

### LINKED ISSUE #4230: Remove the `ssrRef` from provider store
## Problem

`ssrRef` from `@nuxtjs/composition-api` does not have a very good analog in Nuxt 3. In fact, it's not really needed in the store.

## Description

The store state should be initiated once per user request in a plugin instead of the `middleware` that runs on _every_ navigation.
An `init-stores` plugin can initialize the state of the provider store (fetching data from `/stats` endpoint), as well as setting the UI store state from cookies.

## Additional context

Priority set to high because this is necessary for Nuxt 3 migration.

obulat:
## Fixes

Fixes #4230 by @obulat

## Description

_One of the main purposes of this PR is to remove the `ssrRef` from the provider store because it's not available in Nuxt 3.
While removing that, I also wanted to simplify the code we run on each page navigation by moving what's possible to the app start on the server (the `init-stores` plugin)._

This PR adds an `init-stores` plugin that initializes the store state at the app's start, instead of sometimes doing it at every page navigation in `middleware`.

Now, all of the `initFromCookies` calls (for the UI store and the flags store) are done in the `init-stores` plugin.
The flags store's `initFromQuery` is still called from the `middleware` in case the query changes between requests.

This plugin also initializes the providers' data. ~~It also keeps track of the date when the providers were updated and if more than 60 minutes have passed since the last update, refetches the provider data. This is the way we implemented this before, but I'm not sure if it's necessary: I don't think we have users that stay on the site for longer than 60 minutes without refreshing the page. Once you refresh the page, the `init-stores` plugin runs and you get the up-to-date provider data anyways. @zackkrida, do you think we should still keep the update-if-60-minutes-passed functionality for provider data?~~ The providers data is initialized once per request, instead of updating it every 60 minutes.

Some of the data that was saved in `env` property in `nuxt.config` was moved to `publicRuntimeConfig` as `env` is not used in Nuxt 3.

## Update
This PR also adds the `sourceName` and `providerName` properties to the media item when the item is fetched (for single item and the search results) instead of when it's requested. This allows to simplify code using the media: we no longer need to call the provider store in component or `_addProviderName` in the single result store.

To fix the Playwright tests in this PR, I also had to update the stats tape, split the cookies that are saved in `preparePageForTests` between the `features` and `sessionFeatures` (e.g. `fetch_sensitive`) cookies, and add waits to some of the tests that were taking the snapshots before the page rendered.

## Testing Instructions

The app should work as expected; there should be no changes. There will be no calls to `initFromCookies` (ui store and flags store), but you would have to add a log in the app to see that.

The CI should pass, too.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
If we aren't storing providers in a global variable so that it's cached on the SSR side of things, we don't need to ever check the last time it was fetched. We can rely on browser cache to not needlessly resend the request over and over. The only reason to manually manage that is for the SSR side, not for clients. If we were storing the list of providers in a module-level variable rather than recreating it for every request in the plugin function as part of hydrating the store from the cookies, then it would be a different story. But the current approach does not benefit at all from manually managing the cached list of providers, and browsers will do a better job of it (generally). That being said, it absolutely shouldn't get requested more than once per SPA load. If we want caching on the SSR side of things, then we need to change the way the values are stored.

If `getProviders` was updated to something like the following, all the additional complexity of "fetch if needed" can be removed without changing a single thing about the current performance or request frequency of the providers list:

```ts
async getProviders() {
  if (!(this.providers[AUDIO].length && this.providers[IMAGE].length)) {
    await this.fetchMediaProviders()
  }
  return this.providers
}
```

With that, the middleware can be removed, `lastUpdated` can be removed, and all the `needsUpdate` code with it.

obulat:
Thank you for confirming my thoughts that we only need to fetch providers once per SPA load, and for giving more context.

> If `getProviders` was updated to something like the following, all the additional complexity of "fetch if needed" can be removed without changing a single thing about the current performance or request frequency of the providers list:

There are several calls for provider store functions. We would need to update all of them, which will require making them async and adding the necessary related changes. I would prefer to add fewer changes: only fetch the data in the plugin.

sarayourfriend:
> only fetch the data in the plugin.

Makes sense to me!

frontend/src/middleware/middleware.ts
zackkrida, line 1: Since this is now a single-use middleware for the feature flag initialization only, I think we should rename it to feature-flags.ts.

frontend/src/components/VContentSwitcher/VSearchTypeButton.vue
```diff
@@ -9,7 +9,7 @@
     v-bind="$attrs"
     @click="$emit('click')"
   >
-    <VIcon :name="searchType" class="h-6 w-6" />
+    <VIcon :name="searchType" />
```
zackkrida, line 12: Just noting for other reviewer(s), this is safe to remove as it is the default value.

documentation/meta/media_properties/frontend.md
zackkrida, line 1: Why the checkmark removals here?

frontend/src/stores/provider.ts
```diff
@@ -29,6 +25,7 @@ export interface ProviderState {
     audio: string[]
     image: string[]
   }
+  lastUpdated: number
```
zackkrida, line 28: This can be removed I believe?

zackkrida (changes requested):
This nearly LGTM with the exception of one blocking change: I think we should rename `frontend/src/middleware/middleware.ts` to be more explicitly a `feature-flag.ts` middleware, now that it serves a singular purpose.

The test changes introduced seem like they must have been frustrating to have to write! Do you have a sense of why thy were introduced in this PR? I am curious because I would expect these changes to not alter page rendering timing, or make them a bit _faster_ if it did.

frontend/test/storybook/visual-regression/v-footer.spec.ts
```diff
@@ -26,6 +27,7 @@ test.describe("VFooter", () => {
           await page.goto(pageUrl(dir, footerKind))
           if (dir === "rtl") {
             await page.locator("#language").selectOption({ value: "ar" })
+            await sleep(500)
```
sarayourfriend, line 30: Rather than sleeping an arbitrary amount, is there a meaningful locator we `waitFor`? https://playwright.dev/docs/api/class-locator#locator-wait-for

obulat: In the search type test failures, you can see that the button didn't display the icon. When you look into the HTML markup, you can see the `svg` in there, however, the inner `use` doesn't have the required SVG. I couldn't find a good way to wait for _that_ svg.

obulat:
> The test changes introduced seem like they must have been frustrating to have to write! Do you have a sense of why thy were introduced in this PR? I am curious because I would expect these changes to not alter page rendering timing, or make them a bit _faster_ if it did.

They were! :) I'm not sure about the reason, but I think that the page rendering became faster, so the assets and images don't get enough time to render now.

frontend/test/storybook/visual-regression/v-footer.spec.ts
sarayourfriend: I see... if the delay in loading is due to a network request, I wonder if we can [wait for the network request](https://playwright.dev/docs/api/class-page#page-wait-for-response)? I'll give this (and other similar APIs) a try tomorrow when I am working, if you don't get to it first.

My main concern with adding the sleeps is that they are inherently flaky, because if the page suddenly got _faster_, these tests would again start to fail for the same reason. On top of that, I'd have a hard time knowing how to communicate why these sleeps are in the code and when we would consider it correct to use sleeps. Of course, if there's no other way, then that's what it is. I suppose in the worst case, if we wanted to avoid sleeps at all cost, we could wait for a snapshot of the SVG element to match :confused:; If what we're concerned about is that the SVG is _visually_ loaded, that might even be the most "intelligible" way, in that it gets at the precise thing we care about waiting for.

frontend/test/storybook/visual-regression/v-filter-tab.spec.ts-snapshots/filter-tab-focused-hovered-0-linux.png
obulat, line 1: These snapshots were incorrect previously: the other tab was selected and focused

sarayourfriend (approved):
This is so exciting! I am extremely pleased at the solution you found in #4272 and the degree to which it simplified this PR. Great work, Olga!

zackkrida (approved):
This looks great now; lgtm!

PR #4238, opened 2024-04-30
Change search query approach to include only available providers

### LINKED ISSUE #4076: Exclude media from sources without `ContentProvider` record from search
## Problem

When we add a new provider, we don't immediately add a `ContentProvider` record to the API. This means that the media is immediately available in search results.

## Description

When building an excluded source query, we should add the sources that don't have a `ContentProvider` model to the excluded sources. To do that, we can get all sources using:

https://github.com/WordPress/openverse/blob/d377c69fa360c9e32ec74272cd0d69f8ff8fe04c/api/api/controllers/search_controller.py#L549

This is where the query is built:

https://github.com/WordPress/openverse/blob/d377c69fa360c9e32ec74272cd0d69f8ff8fe04c/api/api/controllers/search_controller.py#L188-L193

## Alternatives

An alternative would be to add a step to the data refresh that would add new `ContentProvider` records for all the sources that were added during the data refresh: #3548

## Additional context

#2487 is related in that it added a filter for the single results: if "Hide content" for their provider is set to `True`, the result is not returned.

krysal:
## Fixes

Fixes #4076 by @obulat 

## Description

This PR proposes an alternative solution to the one proposed in the issue: The search controller should limit the results queried to those associated with existing and not hidden providers. IMO, this is a simpler approach that takes advantage of the existing code structure. It fits the "filtered provider" concept in the sense that it queries for valid providers instead of skipping the "excluded." What I don't know is if we want to update the value of the `FILTERED_PROVIDERS_CACHE_VERSION` variable in this case.

## Testing Instructions

1. Spin up the API, `just a`, and go to http://localhost:50280/v1/images/ to confirm the search is working with the usual providers
2. Hide one of the image providers via Django admin, e.g. StockSnap.
3. Confirm the results for the hidden provider are not returned http://localhost:50280/v1/images/
4. Show the results of StockSnap again and delete the `ContentProvider` entry for Flickr.
5. Confirm the results for the deleted provider are not returned http://localhost:50280/v1/images/

## Checklist

- [x] My pull request has a descriptive title (not a vague title like Update index.md`).
- [x] I ran the DAG documentation generator (only applicable for catalog PRs).

[best_practices]:

<details>

```
Version 1.1
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129
    indicated in the file; or
    in the file; or
    it.
```

</details>

### DIFF
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -7,7 +7,7 @@ from elasticsearch_dsl.response import Hit
 from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
 from api.controllers.search_controller import (
     _post_process_results,
-    get_enabled_providers_query,
+    get_enabled_sources_query,
 )
 
 
@@ -55,8 +55,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
             related_query["should"].append(Q("terms", tags__name__keyword=tags))
 
     # Exclude the dynamically disabled sources.
-    if enabled_providers_query := get_enabled_providers_query():
-        related_query["filter"].append(enabled_providers_query)
+    if enabled_sources_query := get_enabled_sources_query():
+        related_query["filter"].append(enabled_sources_query)
     # Exclude the current item and mature content.
     related_query["must_not"].extend(
         [Q("term", mature=True), Q("term", identifier=uuid)]
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -7,7 +7,7 @@ from elasticsearch_dsl.response import Hit
 from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
 from api.controllers.search_controller import (
     _post_process_results,
-    get_excluded_providers_query,
+    get_filtered_providers_query,
 )
 
 
@@ -55,8 +55,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
             related_query["should"].append(Q("terms", tags__name__keyword=tags))
 
     # Exclude the dynamically disabled sources.
-    if excluded_providers_query := get_excluded_providers_query():
-        related_query["must_not"].append(excluded_providers_query)
+    if filtered_providers_query := get_filtered_providers_query():
+        related_query["must"].append(filtered_providers_query)
     # Exclude the current item and mature content.
     related_query["must_not"].extend(
         [Q("term", mature=True), Q("term", identifier=uuid)]
--- a/api/api/controllers/elasticsearch/related.py
+++ b/api/api/controllers/elasticsearch/related.py
@@ -7,7 +7,7 @@ from elasticsearch_dsl.response import Hit
 from api.controllers.elasticsearch.helpers import get_es_response, get_query_slice
 from api.controllers.search_controller import (
     _post_process_results,
-    get_filtered_providers_query,
+    get_enabled_providers_query,
 )
 
 
@@ -36,7 +36,7 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
     tags = getattr(item_hit, "tags", None)
     creator = getattr(item_hit, "creator", None)
 
-    related_query = {"must_not": [], "must": [], "should": []}
+    related_query = {"filter": [], "must_not": [], "should": []}
 
     if not title and not tags:
         if not creator:
@@ -55,8 +55,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
             related_query["should"].append(Q("terms", tags__name__keyword=tags))
 
     # Exclude the dynamically disabled sources.
-    if filtered_providers_query := get_filtered_providers_query():
-        related_query["must"].append(filtered_providers_query)
+    if enabled_providers_query := get_enabled_providers_query():
+        related_query["filter"].append(enabled_providers_query)
     # Exclude the current item and mature content.
     related_query["must_not"].extend(
         [Q("term", mature=True), Q("term", identifier=uuid)]
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -41,12 +41,8 @@ logger = structlog.get_logger(__name__)
 NESTING_THRESHOLD = config("POST_PROCESS_NESTING_THRESHOLD", cast=int, default=5)
 SOURCE_CACHE_TIMEOUT = 60 * 60 * 4  # 4 hours
 FILTER_CACHE_TIMEOUT = 30
-FILTERED_PROVIDERS_CACHE_KEY = "filtered_providers"
-FILTERED_PROVIDERS_CACHE_VERSION = 2
-THUMBNAIL = "thumbnail"
-URL = "url"
-PROVIDER = "provider"
-QUERY_SPECIAL_CHARACTER_ERROR = "Unescaped special characters are not allowed."
+ENABLED_SOURCES_CACHE_KEY = "enabled_sources"
+ENABLED_SOURCES_CACHE_VERSION = 2
 DEFAULT_BOOST = 10000
 DEFAULT_SEARCH_FIELDS = ["title", "description", "tags.name"]
 DEFAULT_SQS_FLAGS = "AND|NOT|PHRASE|WHITESPACE"
@@ -162,26 +158,29 @@ def _post_process_results(
     return results[:page_size]
 
 
-def get_enabled_providers_query() -> Q | None:
+def get_enabled_sources_query() -> Q | None:
     """
-    Get a query that only includes enabled providers.
+    Get a query that only includes enabled sources.
 
-    To exclude a provider, set ``filter_content`` to ``True`` in the
+    To exclude a source, set ``filter_content`` to ``True`` in the
     ``ContentProvider`` model in Django admin.
     The list of ``provider_identifier``s is cached in Redis with
-    `:FILTERED_PROVIDERS_CACHE_VERSION:FILTERED_PROVIDERS_CACHE_KEY` key.
+    `:ENABLED_SOURCES_CACHE_VERSION:ENABLED_SOURCES_CACHE_KEY` key.
     """
 
     try:
-        enabled_providers = cache.get(
-            key=FILTERED_PROVIDERS_CACHE_KEY, version=FILTERED_PROVIDERS_CACHE_VERSION
+        enabled_sources = cache.get(
+            key=ENABLED_SOURCES_CACHE_KEY, version=ENABLED_SOURCES_CACHE_VERSION
         )
     except ConnectionError:
-        logger.warning("Redis connect failed, cannot get cached filtered providers.")
-        enabled_providers = None
+        logger.warning("Redis connect failed, cannot get cached enabled sources.")
+        enabled_sources = None
 
-    if not enabled_providers:
-        enabled_providers = list(
+    if not enabled_sources:
+        # `ContentProvider` currently only handles _sources_, not providers.
+        # TODO: This is a legacy naming convention that should be updated.
+        # https://github.com/WordPress/openverse/issues/4346
+        enabled_sources = list(
             models.ContentProvider.objects.filter(filter_content=False).values_list(
                 "provider_identifier", flat=True
             )
@@ -189,16 +188,16 @@ def get_enabled_providers_query() -> Q | None:
 
         try:
             cache.set(
-                key=FILTERED_PROVIDERS_CACHE_KEY,
-                version=FILTERED_PROVIDERS_CACHE_VERSION,
+                key=ENABLED_SOURCES_CACHE_KEY,
+                version=ENABLED_SOURCES_CACHE_VERSION,
                 timeout=FILTER_CACHE_TIMEOUT,
-                value=enabled_providers,
+                value=enabled_sources,
             )
         except ConnectionError:
-            logger.warning("Redis connect failed, cannot cache filtered providers.")
+            logger.warning("Redis connect failed, cannot cache enabled sources.")
 
-    if enabled_providers:
-        return Q("terms", provider=enabled_providers)
+    if enabled_sources:
+        return Q("terms", source=enabled_sources)
     return None
 
 
@@ -286,8 +285,8 @@ def build_search_query(
     if not search_params.validated_data["include_sensitive_results"]:
         search_queries["must_not"].append(Q("term", mature=True))
     # Exclude dynamically disabled sources (see Redis cache)
-    if enabled_providers_query := get_enabled_providers_query():
-        search_queries["filter"].append(enabled_providers_query)
+    if enabled_sources_query := get_enabled_sources_query():
+        search_queries["filter"].append(enabled_sources_query)
 
     # Search either by generic multimatch or by "advanced search" with
     # individual field-level queries specified.
@@ -397,8 +396,8 @@ def build_collection_query(
     if not include_sensitive_by_params:
         search_query["must_not"].append({"term": {"mature": True}})
 
-    if enabled_providers_query := get_enabled_providers_query():
-        search_query["filter"].append(enabled_providers_query)
+    if enabled_sources_query := get_enabled_sources_query():
+        search_query["filter"].append(enabled_sources_query)
 
     return Q("bool", **search_query)
 
--- a/api/api/controllers/search_controller.py
+++ b/api/api/controllers/search_controller.py
@@ -11,7 +11,6 @@ import structlog
 from decouple import config
 from elasticsearch.exceptions import NotFoundError
 from elasticsearch_dsl import Q, Search
-from elasticsearch_dsl.query import EMPTY_QUERY
 from elasticsearch_dsl.response import Hit, Response
 from redis.exceptions import ConnectionError
 
@@ -162,9 +161,10 @@ def _post_process_results(
     return results[:page_size]
 
 
-def get_excluded_providers_query() -> Q | None:
+def get_filtered_providers_query() -> Q | None:
     """
-    Hide data sources from the catalog dynamically.
+    Get a query for filtering out items from hidden content providers or without any content provider.
+
     To exclude a provider, set ``filter_content`` to ``True`` in the
     ``ContentProvider`` model in Django admin.
     The list of ``provider_identifier``s is cached in Redis with
@@ -181,7 +181,7 @@ def get_excluded_providers_query() -> Q | None:
 
     if not filtered_providers:
         filtered_providers = list(
-            models.ContentProvider.objects.filter(filter_content=True).values_list(
+            models.ContentProvider.objects.filter(filter_content=False).values_list(
                 "provider_identifier", flat=True
             )
         )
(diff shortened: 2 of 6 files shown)


api/api/controllers/search_controller.py
```diff
@@ -166,16 +165,16 @@ def _post_process_results(
     return results[:page_size]
 
 
-def get_excluded_providers_query() -> Q | None:
+def get_filtered_providers_query() -> Q | None:
```
sarayourfriend, line 168: Nit, but "filtered" reads to me basically the same as "excluded". "enabled", "available", or something along those lines would have a more obviously positive, "these are the providers we serve results from" connotation.

api/api/controllers/search_controller.py
```diff
@@ -166,16 +165,16 @@ def _post_process_results(
...
     """
-    Hide data sources from the catalog dynamically.
+    Get a query for filtering out items from hidden content providers or without any content provider.
```
sarayourfriend, line 170: As a supplement to my other comment, this would be clearer, maybe with an optional additional line that clarifies how we decide if a provider is "enabled", if the code is for some reason insufficient.

```suggestion
    Get a query that only includes enabled providers.
```

api/api/controllers/search_controller.py
```diff
@@ -291,8 +290,8 @@ def build_search_query(
     if not search_params.validated_data["include_sensitive_results"]:
         search_queries["must_not"].append(Q("term", mature=True))
     # Exclude dynamically disabled sources (see Redis cache)
-    if excluded_providers_query := get_excluded_providers_query():
-        search_queries["must_not"].append(excluded_providers_query)
+    if filtered_providers_query := get_filtered_providers_query():
+        search_queries["must"].append(filtered_providers_query)
```
sarayourfriend, line 294: I think this should use `filter` instead of `must`. `must` will affect the scoring of the documents, whereas `filter` doesn't. Documentation here about that difference: https://www.elastic.co/guide/en/elasticsearch/reference/8.10/query-dsl-bool-query.html

```suggestion
        search_queries["filter"].append(filtered_providers_query)
```

This might also influence the terminology change I suggested. If we are using "filter" for the ES portion, maybe "filtered providers" is accurate (though it still reads with a negative/exclusion connotation to me, might just be my personal reading not shared by others).

api/api/controllers/search_controller.py
```diff
@@ -401,8 +395,8 @@ def build_collection_query(
     if not include_sensitive_by_params:
         search_query["must_not"].append({"term": {"mature": True}})
 
-    if excluded_providers_query := get_excluded_providers_query():
-        search_query["must_not"].append(excluded_providers_query)
+    if filtered_providers_query := get_filtered_providers_query():
+        search_query["must"].append(filtered_providers_query)
```
sarayourfriend, line 399: Same as above, with respect to must vs filter.

```suggestion
        search_query["filter"].append(filtered_providers_query)
```

api/test/unit/controllers/elasticsearch/test_related.py
```diff
@@ -82,8 +82,10 @@ def test_related_media(
         "from": 0,
         "query": {
             "bool": {
+                "must": [
```
sarayourfriend, line 85: As above.

```suggestion
                "filter": [
```

api/test/unit/controllers/test_search_controller_search_query.py
```diff
@@ -283,9 +250,10 @@ def test_create_search_query_empty_with_dynamically_excluded_providers(
     assert actual_query_clauses == {
         "must_not": [
             {"term": {"mature": True}},
-            {"terms": {"provider": [excluded_providers_cache]}},
         ],
-        "must": [{"match_all": {}}],
+        "must": [
```
sarayourfriend, line 254: As above

```suggestion
        "filter": [
```

sarayourfriend (commented):
I like this approach, it makes perfect sense to me and I agree it is much clearer than the multiple steps of reasoning required for the other approach.

I've left one comment regarding the terminology used, but that's not as important to me as clarifying whether we need to use `must` or `filter`. Based on my understanding, I think `filter` is the right choice, but if it isn't, then we should document why we use `must`, considering its effect on document scoring (and conversely, we should document why we use `filter`, if we do, namely so that it _doesn't_ affect scoring).

api/api/controllers/search_controller.py
obulat: Yes, this query should use `filter` instead of `must` because it's more performant, and we don't need the relevancy score for providers. It's a bit unintuitive that the opposite of `must_not` query is the `filter` query, not `must` query.

api/api/controllers/search_controller.py
```diff
@@ -341,12 +340,6 @@ def build_search_query(
     if settings.USE_RANK_FEATURES:
         search_queries["should"].extend(create_ranking_queries(search_params))
 
-    # If there are no `must` query clauses, only the results that match
```
obulat, line 344: We'll have to re-add this after changing the providers query back from `must`.

api/api/controllers/elasticsearch/related.py
```diff
@@ -55,8 +55,8 @@ def related_media(uuid: str, index: str, filter_dead: bool) -> list[Hit]:
             related_query["should"].append(Q("terms", tags__name__keyword=tags))
 
     # Exclude the dynamically disabled sources.
-    if excluded_providers_query := get_excluded_providers_query():
-        related_query["must_not"].append(excluded_providers_query)
+    if filtered_providers_query := get_filtered_providers_query():
+        related_query["must"].append(filtered_providers_query)
```
obulat, line 59:
```suggestion
        related_query["filter"].append(filtered_providers_query)
```

obulat (commented):
This is such a smart approach! 
I agree with Sara on the naming, something with the meaning of `enabled` or `included` would be better.

api/api/controllers/search_controller.py
krysal: Thanks for letting me know about the `filter` clause. I wasn't aware of it, but reading the documentation, it totally makes sense.

krysal:
I'm happy to change the terminology and, of course, the clause for the query. I had similar thoughts regarding the names, but it seemed to me that they could fit with the existing environment variables, and I didn't want to give it much thought since it could also be decided in the review 😄 

This is back to draft in the meantime.

sarayourfriend (commented):
@krysal @obulat do y'all know whether the source/provider distinction is potentially getting muddled in this new approach? I can't remember if `ContentProvider` only has true "providers" and not sources. If it also has sources, do we need to be careful about how it's used in relation to the `source` filters (both the include and exclude parameters)?

This LGTM otherwise, but I wanted to make sure I understood where the distinction was before approving.

krysal:
@sarayourfriend `ContentProvider` contains sources, and I don't see any issue with the functionalities. I tried hiding Freesound locally, for example, and I requested audio exclusively from it, `http://localhost:50280/v1/audio/?source=freesound`, and got zero rows as expected (since the source is disabled). Let me know if you or @obulat find a conflict.

sarayourfriend:
The problem with the implementation, if there is one (if sources are in `ContentProvider` in addition to providers) is we only use the entries in `ContentProvider` to filter on the `provider` field of the ES documents, not `source`, whereas the query parameters refer to `source`.

krysal (commented):
> The problem with the implementation, if there is one (if sources are in ContentProvider in addition to providers) is we only use the entries in ContentProvider to filter on the provider field of the ES documents, not source, whereas the query parameters refer to source.
>
> To be clear, if this is a problem now, then it was a problem before with the exclusions, but will be worse now because we wouldn't be including all the sources, only those which happen to also be providers.

I SEE, wow, I think you're right. We're mixing **sources** with **providers**, I hadn't paid much attention to it because I thought this was covered since the "new" parameters work 😅

> If ContentProvider really does have sources, not just providers, then my hunch is that we actually have a complex issue here when it comes to providers like smithsonian with many sources. Do we have an individual content provider entry for all smithsonian sources?

I don't think we have those; some sources are too small (<100 items), so we decided to skip them. But the ugly problem is that there shouldn't be a `provider` entry equivalent to all `source` entries (not sure if I'm explaining myself here, but hopefully so!). The provider indicates where we obtain the data from, despite which the source site is.

There could be a simpler solution tho, proposed below.

api/api/controllers/search_controller.py
```diff
@@ -191,13 +192,13 @@ def get_excluded_providers_query() -> Q | None:
...
-                value=filtered_providers,
+                value=enabled_providers,
             )
         except ConnectionError:
             logger.warning("Redis connect failed, cannot cache filtered providers.")
 
-    if filtered_providers:
-        return Q("terms", provider=filtered_providers)
+    if enabled_providers:
+        return Q("terms", provider=enabled_providers)
```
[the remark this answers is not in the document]
sarayourfriend: That's along the lines of what I'm thinking... but _if and only if_ all active sources have a `ContentProvider` entry.

Which is to say, _exclusion_ was certainly safer for sources (though the code was already problematic in that we couldn't disable a source).

Clarifying whether `ContentProvider` is _just providers_ or if it's also sources (and if not, then how _sources_ factor into the data model) would be really helpful.

If it's just providers, then this code would work fine, if all we want is to _include providers that have a `ContentProvider` and are not disabled_. However, if we also need to be able to selectively enable/disable _sources_, then either: (a) we need to overload the conception of `ContentProvider`, and probably stick to only explicitly _excluding_ sources rather than including them based on `ContentProvider` (otherwise we'd have to make a `ContentProvider` for all the Smithsonian sources); or, (b) we need to create a separate `Source` model to track sources, and then probably only track source _exclusions_, considering how many sources there are and how tedious it would be to manage that by hand... though that tediousness is maybe just this first go-round and we could automatically create sources during the data refresh when a new one is detected.

All to say: this is pretty complicated _but_ so long as we are not currently _relying on `ContentProvider` to exclude a **source**_  then it should work. If that's the case, then great, we can move forward with the PR as is, and then circle back to this question of how to (and whether to) manage source inclusion/exclusion. Otherwise, we probably need to stick to exclusion.

We definitely have sources in `ContentProvider` (finnish_satakunnan_museum and finnish_heritage_agency, the Smithsonian sources, others too), but I can't at a glance tell which are excluded (need to do a db query for that).

These are the only hidden ones:

```
deploy@localhost:openledger> select provider_name from content_provider where filter_content;
+---------------------+
| provider_name       |
|---------------------|
| Flora-On            |
| ccMixter            |
| Science Museum – UK |
+---------------------+
SELECT 3
Time: 0.222s
```

All of these are "provider-sources", so it should be okay _for now_. But if we make this change to the code, we really need to circle back and reconsider the meaning of `ContentProvider` and the provider/source relationship in that model.

It would probably be a good idea to have separate models, `Provider` and `Source`, where `Source` has a foreign key relationship to `Provider`, rather than overloading the meaning of `ContentProvider`. Even just reasoning through the limitations of the current approach to query building and the meaning of these models would be so much easier if that distinction was clear in absolute terms.

sarayourfriend (changes requested):
Just requesting changes to clarify the PR is under discussion while we decide what steps we want to take now.

api/api/controllers/search_controller.py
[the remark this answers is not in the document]
obulat: Originally, we only had `source`. Then, the `provider` field was added to denote the provider of the metadata about the image [^1]. The PR that added the `provider` field, didn't touch the `ContentProvider` model, and this model wasn't updated later.

`ContentProvider` currently only handles _sources_, not _providers_. You can check this if you go to https://openverse.org/sources - you'll see NASA in the table. This uses the `stats` endpoint that queries `sources`, not `providers`.

All this is to say that I think we _can_ proceed with this PR, and exactly because we rely on `ContentProvider` to filter sources. I'll review the code again to check that this is true.

> All to say: this is pretty complicated but so long as we are not currently relying on ContentProvider to exclude a source then it should work. If that's the case, then great, we can move forward with the PR as is, and then circle back to this question of how to (and whether to) manage source inclusion/exclusion. Otherwise, we probably need to stick to exclusion.

Could we open a new issue (maybe even a Project proposal request?) to 
- Rename the `ContentProvider` model to `ContentSource`
- Create a new `ContentProvider` model for providers
- surface provider stats in the API
- use this view to add provider links to the frontend single result page
There is some additional info in the cc-archive issues and PRs (https://github.com/cc-archive/cccatalog-api/pull/548/files#r444933371, https://github.com/cc-archive/cccatalog-api/pull/560#issue-654914001)

By the way, some of the issues mention that a source can appear in several providers. I think we have a project for implementing this, this would reduce the number of duplicates we have, but we would need to find a way to handle a work that has several providers. 

[^1]: https://github.com/cc-archive/cccatalog-api/issues/531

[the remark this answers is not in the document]
sarayourfriend: > All this is to say that I think we can proceed with this PR, and exactly because we rely on ContentProvider to filter sources. I'll review the code again to check that this is true.

If `ContentProvider` is _only_ for sources then we need to change the filter here to `source`, not `provider`.

obulat (changes requested):
My request for changes, in light of the sources/provider discussion, is to replace `provider` with `source` everywhere it's necessary, and add a comment on the `ContentProvider` model to say that it's actually a content source, not provider (until we rename the model)

api/api/controllers/search_controller.py
krysal: > > All this is to say that I think we can proceed with this PR, and exactly because we rely on ContentProvider to filter sources. I'll review the code again to check that this is true.
> 
> If `ContentProvider` is _only_ for sources then we need to change the filter here to `source`, not `provider`.

This is exactly what I proposed in the first comment of this thread, which it is... deleted?? I don't understand what happened 😳 Anyway, I 100% agree that the current naming is a mess and very confusing between what actually is a source and what is a provider. I'll apply the minimal changes required to fulfill the issue requirement and keep the PR simple. As @obulat suggests, the renaming sure warrants its own issue, maybe even a small project.

> Originally, we only had source. Then, the provider field was added to denote the provider of the metadata about the image [1](https://github.com/WordPress/openverse/pull/4238#user-content-fn-1-12b80993ab3e016fe9624dcd6bf5e692).

That is a curious fact. I thought the `provider` field came before the `source` given that the model's name is `ContentProvider`.

krysal, line 201:
```suggestion
    if enabled_providers:
        return Q("terms", source=enabled_sources)
```

api/api/controllers/search_controller.py
```diff
@@ -41,12 +41,8 @@
...
 SOURCE_CACHE_TIMEOUT = 60 * 60 * 4  # 4 hours
 FILTER_CACHE_TIMEOUT = 30
-FILTERED_PROVIDERS_CACHE_KEY = "filtered_providers"
-FILTERED_PROVIDERS_CACHE_VERSION = 2
-THUMBNAIL = "thumbnail"
-URL = "url"
-PROVIDER = "provider"
-QUERY_SPECIAL_CHARACTER_ERROR = "Unescaped special characters are not allowed."
+ENABLED_SOURCES_CACHE_KEY = "enabled_sources"
+ENABLED_SOURCES_CACHE_VERSION = 2
```
obulat, line 45: Thank you for removing the unused constants from [5 years ago](https://github.com/WordPress/openverse/commit/effee2205d1b57d22c90077f8f935cba5e7d1025) :)

api/test/unit/controllers/test_search_controller.py
```diff
@@ -838,39 +838,39 @@ def _delete_all_results_but_first(_, __, results, ___):
...
     cache = request.getfixturevalue(cache_name)
 
     if is_cache_reachable:
         cache.set(
-            key=FILTERED_PROVIDERS_CACHE_KEY,
+            key=ENABLED_SOURCES_CACHE_KEY,
             version=2,
             timeout=30,
-            value=[f"provider{i + 1}" for i in range(excluded_count)],
+            value=[f"source{i + 1}" for i in range(excluded_count)],
```
obulat, line 853: Should we also rename `excluded_count` to `enabled_count`?

obulat (commented):
I'll approve after testing this PR locally.

obulat (approved):
Everything works well locally 🎉 

I think we do need to change the variable name in [api/test/unit/controllers/test_search_controller.py](https://github.com/WordPress/openverse/pull/4238/files/f9b0d0edc001958265e488c00acd5e1985b8c2b6#diff-3f72d9dc992edb46dcea0b7136c9ecaa0f888653a2c4c6261ba0116de468eef8)

api/test/unit/controllers/test_search_controller.py
krysal: Sure, it doesn't make sense with `excluded`. Good eye!

api/api/controllers/search_controller.py
krysal: I didn't even know they were so old, wow! What is strange is that there should be a pre-commit lint step that fails due to unused variables 🤔  Not sure what happened with that...

krysal:
I had to rebase it with `main` because the API image wasn't being built correctly. Other than that, the variable in the test is updated ✔️

sarayourfriend:
I'll re-review this today :+1:

api/api/controllers/search_controller.py
sarayourfriend: You can't reliably lint for unused variables declared at the module scope in Python, because all top-level variables are automatically exported. It'd be like trying to lint an unused `export const whatever;` in JavaScript. It can't be done unless you statically analyse all references to the module. I don't know if mypy supports something like that, but certainly ruff never could.

api/api/controllers/search_controller.py
sarayourfriend: If we only care about sources and not providers, then we should "just" (in quotes because maybe it isn't trivial to do) rename ContentProvider to `ContentSource` (although, I don't know why the "content" prefix is necessary) and the variables/serializers/documentation to clarify it (e.g. in the stats endpoint).

We don't need to add structure that is never used, so if `ContentProvider` only ever really means to reference the sources, then we can just forget about representing providers as an exclusive category at the API database level altogether, at least until we actually have a reason to do that.

> This uses the stats endpoint that queries sources, not providers

@obulat the stats endpoint queries `ContentProvider`, so I don't think this is in any way obvious through any of the current implementation, on the name level, certainly not by just reading the API code... and my intuition would be to assume there was a confusion between source and provider, rather than the difference in reference on the frontend to all references in the API were intentional.

Just want to clarify that this _is_ messy and needs to be cleaned up, otherwise we're relying on latent, undocumented understandings of how the usage of that particular model changed, without ever having made the code reflect that, not even at the public documentation level:

https://api.openverse.org/v1/#tag/images/operation/images_stats

The documentation there now says that `display_name` is "The name of content provider, e.g. Flickr", contrasted to `source_name`, which is "The source of the media, e.g. flickr". The difference is unexplained, and there's no way to know that `ContentProvider` actually represents sources based on how the code uses it.

https://github.com/WordPress/openverse/blob/c92d484cd0f356cb71e8fd7fc15d4a45d6523d0f/api/api/views/media_views.py#L206-L217

Note that the serializer, model, and variable name of the retrieved data are all provider, not source.

Really just wanting to make it clear that this _isn't_ clear in the code, and I think we should consider treating resolving that with greater urgency than we have thus far. The source/provider distinction is terribly tedious, and anything we do to clarify it even for ourselves is worth the effort.

sarayourfriend: Thanks for opening the issue, by the way, @krysal! For visibility, it is here:

https://github.com/WordPress/openverse/issues/4346

sarayourfriend (approved):
LGTM!

api/api/controllers/search_controller.py
krysal: I agree 100% with you @sarayourfriend, so I raised the issue's priority.

Issue #3798, opened 2024-02-15
Use `pytest-pook` plugin to clean up pook fixtures and usage in tests


sarayourfriend:
## Current Situation

The API tests have several different patterns they follow to manage pook usage within tests. Some use `pook.use` as a context manager. Some use `pook.on`. And some others use the `pook_on` fixture.

## Suggested Improvement

All of these can be safely replaced with `pytest-pook`'s `@pytest.mark.pook` on the test. The mark is compatible with parametrize and all other pytest plugins by default, and does not interfere with other test decorators.

The plugin also has the additional benefit of two checks: 

1. It checks to make sure that pook was used in tests that are marked with it. This prevents the possibility of thinking a test was using pook, but that it wasn't (for example, if relying on a complex fixture to create pook mocks or an inner function).
2. It checks to make sure all mocked requests are exhausted at the end of the test, the benefit of which is explained in the plugin docs: https://git.sr.ht/~sara/pytest-pook#dangling-or-unused-pending-mocks. This behaviour can also be turned off for individual tests that intentionally rely on this approach.

To implement this change: 
1. Add `pytest-pook` as a new development dependency in the API, catalog, and ingestion server test suites
2. In each of those applications, identify each test that uses pook and replace `pook.use`, `pook.on`, or `pook_on` usage with the pytest plugin
3. Remove `pook_on` and `pook_off` fixtures (they should now no longer be used)

## Benefit

Unify the approach to pook context management in our test suites so that contributors do not have to choose from the various different approaches. Instead, there will be a single default choice that works for all cases.

## Additional context

I'm the author of `pytest-pook` and it is also used in pook's own test suite: https://github.com/h2non/pook/blob/master/tests/unit/interceptors/aiohttp_test.py#L15

Sukhshum2628:
Hi, I'd like to work on this if it's still open.

I've gone through the issue description and the approach is clear — 
add `pytest-pook` as a dev dependency in the API, catalog, and 
ingestion server, then replace `pook.use`, `pook.on`, and `pook_on` 
fixture usage with `@pytest.mark.pook` across those three test suites, 
and finally remove the now-unused `pook_on` and `pook_off` fixtures.

Before I start, I wanted to confirm:
1. Should this be a single PR covering all three applications, or 
   would you prefer separate PRs per application to keep diffs smaller?
2. Is there a preferred way to handle tests that intentionally leave 
[...]

PR #4249, opened 2024-05-01
Cache repeated thumbnail failures within configured TTL

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/550 by @sarayourfriend
Related to #3798 (doesn't close that issue because there are still places we'll want to clean up direct pook context management where it can be avoided, and `.matches` assertions as well)

## Description

I've added a function to wrap the `image_proxy.get` function, because the process for caching was complex enough that trying to put it into `image_proxy.get` itself would get very messy. The new function decorates `image_proxy.get` and short-cuts requests with a cached failure response when the upstream request has failed enough times in a row within a window of time.

The approach I've chosen works under the assumption that thumbnails are more likely to be flaky, than permanently broken, because dead link filtering should prevent completely unavailable images from appearing in search, which is the main place thumbnail requests would come from or become accessible from (whether on the frontend or through referencing the URLs returned by an API search request).

If we decided _not_ to work with that assumption, then we could consider doing away with a TTL on the cached failures, instead just permanently caching those thumbnails as a failure. I can see an argument for doing that, but it's a bit more extreme, and this at least gets us to a place where we're more intelligently anticipating potential failures, without making it hard for us to experiment with other approaches in the future. In fact, if we wanted to just test a hard-failure, without a TTL, we could just set a very large TTL, and see whether whatever we're hoping happens in that case starts happening.

Because I've gone forward with the assumption that thumbnail requests are more likely to be flaky rather than outright persistent failures _in most cases_, I've also chosen to decrement the failure count whenever a successful response happens. This is easier than the approach suggested in the issue, which was to create separate started and completed tallies. The approach suggested in the issue was trying to solve a problem where a thumbnail request could fail in such a way that our application never even got to return a failed response (I think because of an OOM or worker timeout, something like that). This was a bigger issue pre-ASGI. However, trying to handle that is far more complex because we'd need to take into consideration that multiple requests could happen concurrently (across different or even the same workers) for the same thumbnail. Imagine we set the "uncompleted request" tolerance to 5. If a thumbnail request came in, uncached, from 6 different users within a short period of time (100s of ms from each other), then the 6th request would see 5 "uncompleted" requests in the tallies, for requests that could succeed (they haven't _necessarily_ failed). There's no way to disambiguate that situation from 5 "actually" uncompleted requests. In that case, the 6th requester would get a "cached" failure, even though the thumbnail would have succeeded. An easy suggestion is to increase the tolerance. However, that's not workable, because there isn't a tolerance that makes sense for the normal case of an unpopular but always failing thumbnail _and also prevents popular but successful_ thumbnails from looking like they've got a bunch of failures to concurrent requests.

As far as I know, the problem where our workers are crashing is not something we're dealing with now, so I've chosen to go with a simpler approach that handles a more common case, where a thumbnail fails in a way we can reliably track, and if it does that too much too quickly, well then assume it's going to keep failing, at least for however long we configure the TTL ("cache window") to be.

**Aside from all that**, I simplified the `test_image_proxy` module in the following ways:

- Introduce `pytest-pook` (#3798) and remove the various `mock.matches` assertions, which the pytest plugin already checks
- Remove the custom async wrapper `photon_get` fixture, and just use `async_to_sync` instead

I also moved the pytest.ini configuration into pyproject.toml to consolidate configuration there.

## Testing Instructions

Check out the code and the unit tests. This is hard to actually test locally unless you modify a work to point to a known bad upstream URL, in which case you can reproduce it by requesting the failing thumbnail more times than the configured tolerance (refer to the new setting) and seeing that you eventually get the bypass error response, rather than the upstream error response.

Otherwise, I think I've covered the new wrapper in the unit tests, but please give feedback if there's a way to improve the clarity of the test or if I missed something.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/utils/image_proxy/__init__.py
```diff
@@ -120,9 +124,87 @@ def _tally_client_response_errors(tallies, month: str, domain: str, status: int)
         logger.warning("Redis connect failed, thumbnail HTTP errors not tallied.")
 
 
+# thmbfail == THuMBnail FAILures; this key path will exist for every thumbnail requested, so it needs to be space efficient
```
zackkrida, line 127: I thought this was one of those "mocking spongebob" memes with the weird capitalization before I realized you were capitalizing the letters in the "thmbfail" acronym 

8owr7l

api/api/utils/image_proxy/__init__.py
```diff
@@ -120,9 +124,87 @@ def _tally_client_response_errors(tallies, month: str, domain: str, status: int)
...
+
+def _cache_repeated_failures(_get):
+    """
+    Wrap ``image_proxy.get`` to cache repeated upstream failures
+    and avoid re-requesting images likely to fail.
+
+    Do this by storing a count for each media identifier of requested thumbnails.
+    When the upstream request succeeds, increment the key. When it fails, decrement it.
+
+    If the count goes below -2, then the thumbnail has failed twice within cache frame.
```
zackkrida, line 139:
```suggestion
    If the count goes below -2, then the thumbnail has failed twice within the cache frame.
```

super nit

api/conf/settings/thumbnails.py
```diff
@@ -15,3 +17,15 @@
...
+
+# The length of time to cache repeated thumbnail failures
+THUMBNAIL_FAILURE_CACHE_WINDOW_SECONDS = config(
+    "THUMBNAIL_FAILURE_CACHE_WINDOW_SECONDS",
+    default=int(timedelta(days=2).total_seconds()),
+    cast=int,
+)
+
+# The number of times to try a thumbnail request before caching a failure
+THUMBNAIL_FAILURE_CACHE_TOLERANCE = config(
```
zackkrida, line 29: super nit: `THUMBNAIL_FAILURE_CACHE_TOLERANCE` is probably clear enough but `THUMBNAIL_FAILURE_CACHE_TOLERANCE_TRIES` would be even more explicit.

api/pyproject.toml
```diff
@@ -60,7 +60,26 @@ test = [
...
+[tool.pytest.ini_options]
+DJANGO_SETTINGS_MODULE = "conf.settings"
+
+pythonpath = "."
+
+filterwarnings = [
+  # Ignore warnings related to unverified HTTPS requests.
+  # Reason: This warning is suppressed to avoid raising warnings when making HTTP requests
+  # to servers with invalid or self-signed SSL certificates. It allows the tests to proceed
+  # without being interrupted by these warnings.
```
zackkrida, line 78: Great docs :+1:

api/test/unit/utils/test_image_proxy.py
```diff
@@ -67,25 +69,17 @@ def auth_key():
...
-            return res
-        finally:
-            tasks = asyncio.all_tasks(session_loop)
-            for task in tasks:
-                session_loop.run_until_complete(task)
-
-    yield do
+# Do not convert tests to async, because of this issue:
+# https://github.com/pytest-dev/pytest-django/issues/580
+# While the transaction workaround technically works, its
```
zackkrida, line 74:
```suggestion
# While the transaction workaround technically works, it is
```

super nit

api/api/utils/image_proxy/__init__.py
```diff
@@ -120,9 +124,87 @@ def _tally_client_response_errors(tallies, month: str, domain: str, status: int)
...
+FAILURE_CACHE_KEY_TEMPLATE = "thmbfail:{ident}"
+
+
+def _cache_repeated_failures(_get):
+    """
+    Wrap ``image_proxy.get`` to cache repeated upstream failures
+    and avoid re-requesting images likely to fail.
+
+    Do this by storing a count for each media identifier of requested thumbnails.
+    When the upstream request succeeds, increment the key. When it fails, decrement it.
```
stacimc, line 137: As I understand it, isn't this the opposite of what we do? We track the failure_count, and increment when it fails/decrement when it succeeds.

stacimc (approved):
Excellent! Well documented, tests are super clear, and the rationale makes sense to me. Approving to unblock, although I do think we should fix my one comment about the docstring to avoid confusion.

api/api/utils/image_proxy/__init__.py
sarayourfriend: I'm not familiar with that meme :sweat_smile:

api/api/utils/image_proxy/__init__.py
sarayourfriend: Oops, yes! I originally implemented a slightly different approach and forgot to update this docstring. I'll push a commit to address this and the edits @zackkrida proposed.

api/pyproject.toml
sarayourfriend: To clarify, I copied these from the pytest.ini, didn't write them myself!

Issue #4182, opened 2024-04-23
Refine `openverse-attribution` as a standalone pacakge


dhruvkb:
## Problem

We can consider `openverse-attribution` to be its own standalone library, our first Python library that can be published to PyPI from the monorepo. There are a number of things that need to be done before that is feasible.

## Description

There are some higher-level discussions.

1. Devise a naming system. Will it be called `openverse-attribution` or something more generally suitable like `cc-attribution` ([comment](https://github.com/WordPress/openverse/pull/4107#discussion_r1575460567))?
2. Will the `README.md` file be descriptive and duplicated between the docs site and the library? Will it be copied into the package during build? Will it be a shell `README.md` that only points to the docs ([comment](https://github.com/WordPress/openverse/pull/4107#discussion_r1575463968))?

There are also some code refactors.

1. Use `License` enum as the argument type for param `license(_slug?)` in `get_attribution_text` ([comment](https://github.com/WordPress/openverse/pull/4107#discussion_r1573024182)).
2. Make `attribution_template` a module level constant `_ATTRIBUTION_TEMPLATE` ([comment](https://github.com/WordPress/openverse/pull/4107#discussion_r1573024555)).

## Additional context

Currently the package is local and only used by the API. Some of these concerns will get real once we're set on publishing the package to PyPI.

PR #4250, opened 2024-05-01
Update `openverse-attribution` with new features and improvements

dhruvkb:
## Fixes

Partially relates to #4182.

## Description

This PR

- updates `openverse-attribution` to add lots of new features
  - support for versions and jurisdictions
  - stricter validation for slug, version and jurisdiction params
  - ability to infer missing version and jurisdiction
- refactors `openverse-attribution` for a clearer API
  - Old `License` enum renamed to `LicenseName`.
  - Added dataclass `License` wrapping `name` (`LicenseName`) and optionally `version` (`str`) and `jurisdiction` (`str`).
  - all functionality is exposed via the `License` class
- fixes problems present in `openverse-attribution`
  - several license names that were missing were added such as SA, ND-NC and DevNations
  - added support for license aliases e.g. 'zero' (i.e. CC0) and 'mark'/'publicdomain' (i.e. PDM)
- updates usage of `openverse-attribution` in the API


packages/python/openverse-attribution/src/openverse_attribution/license.py
```diff
@@ -1,128 +1,271 @@
...
+    jur: str | None
 
-    # Public domain dedication
-    CC0 = "cc0"
+    def __init__(
+        self,
+        slug: str,
+        ver: str | None = None,
+        jur: str | None = None,
+    ):
```
sarayourfriend, line 25: This constructor is a little confusing. I can pass a jurisdiction, even if the license doesn't support it :thinking: for example:

```py
License(slug="pdm", ver="4.0", jur="jp")
```

This is something meaningless about this constructor, but it would silently pass.

I think there is a way to validate this without needing to use this imperative code, either, and instead rely on the `all_licenses` to construct ccREL formatted references for all known licenses and marks. Flattening `all_licenses.json` into a set of strings like `{"by-nc/3.0/pl", "by-nc/2.0/jp", "by/4.0", ...}` and then either taking those strings directly as arguments (which would make it easier to support additional license namespaces in the future, without a significant API change) or as parts, in the current implementation, but construct the string and check for it in the flatted all_licenses.

We can also alias version with `*` to the latest jurisdiction, so `by-nc/*/jp` would map to `by-nc/2.0/jp` so that if the version isn't provided, fall back to `*` and it still gets mapped to the right place.

This would make most of this code declarative, a bit easier to follow/reason about, and would make supporting additional license namespaces trivial. To do that, we'd add a new `namespace` parameter in the vein of RDF namespaces, so libreart would have `https://artlibre.org/licence`, CC would have `https://creativecommons.org/licenses`, and so forth. That would just key to the mapping used, rather than needing to change the logic of this function.

It's worth also taking a look at some prior art, and the API of the library for parsing spdx signifiers: https://pypi.org/project/spdx-lookup/. In particular, the `by_id` and `by_name` methods, which are essentially similar to what I'm suggesting as far as using a single string argument. Validation is much easier this way, especially if we consider multiple license namespaces beyond CC.

packages/python/openverse-attribution/src/openverse_attribution/license_name.py
```diff
@@ -0,0 +1,131 @@
...
+
+        return self in {
+            LicenseName.PDM,
+            LicenseName.CERTIFICATION,
+            LicenseName.PUBLICDOMAIN,
+            LicenseName.CC0,
+        }
+
+    @property
+    def allowed_ver_jur(self) -> list[tuple[str, str]]:
```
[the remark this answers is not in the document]
dhruvkb: These are such long words 😢 and so repeatedly used.

packages/python/openverse-attribution/src/openverse_attribution/data/all_licenses.json
```diff
@@ -0,0 +1,161 @@
+{
```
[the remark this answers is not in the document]
dhruvkb: No I generated this by scraping and parsing these two webpages.

- https://creativecommons.org/licenses/list.en
- https://creativecommons.org/publicdomain/list.en

packages/python/openverse-attribution/src/openverse_attribution/license.py
dhruvkb: > This is something meaningless about this constructor, but it would silently pass.

It does not. It accepts partial info (for when version and jurisdiction are unknown) and completes it if possible. If the information is invalid it will not pass.

Screenshot 2024-05-02 at 8 36 38 AM

> We can also alias version with * to the latest jurisdiction, so by-nc/*/jp would map to by-nc/2.0/jp so that if the version isn't provided, fall back to * and it still gets mapped to the right place.

The constructor already works like this. You can skip the fields you do not know. 

If only one match fits, it will populate it...

Screenshot 2024-05-02 at 8 41 21 AM

... else it will use the latest match as a `fallback`.

Screenshot 2024-05-02 at 8 39 25 AM

packages/python/openverse-attribution/src/openverse_attribution/license.py
```diff
@@ -1,128 +1,271 @@
...
 
-    # Deprecated CC licenses
-    SA = "sa"
-    SAMPLING = "sampling+"
-    NC_SAMPLING = "nc-sampling+"
+@dataclass
+class License:
+    name: LicenseName
+    ver: str | None
+    jur: str | None
```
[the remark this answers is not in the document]
dhruvkb: Resolved in 19d0f04.

packages/python/openverse-attribution/src/openverse_attribution/license_name.py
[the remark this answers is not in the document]
dhruvkb: Resolved in f1c6fa6.

packages/python/openverse-attribution/src/openverse_attribution/license_name.py
```diff
@@ -0,0 +1,131 @@
...
+            LicenseName.NC,
+            LicenseName.ND,
+            LicenseName.NC_SA,
+            LicenseName.ND_NC,
+            LicenseName.SAMPLING,
+            LicenseName.SAMPLING_PLUS,
+            LicenseName.NC_SAMPLING_PLUS,
+            LicenseName.DEVNATIONS,
+            LicenseName.CERTIFICATION,
+        }
```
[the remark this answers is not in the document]
dhruvkb: Yes! Done in 9d7f91b.

packages/python/openverse-attribution/src/openverse_attribution/license.py
sarayourfriend: Apologies for misunderstanding the validation! Your reasoning sounds good to me :+1:

sarayourfriend (approved):
LGTM. I think there are opportunities to make this library more flexible and if we got to the point of publishing it, I think we should. But you're right, this package as it is, and maybe for forever, is meant to share logic across parts of Openverse codebase, not solve a general problem.

Sorry I got sidetracked with that. LGTM!

dhruvkb:
💯 We should definitely look into making this more generalised and also consider open sourcing it. But until that time, we should take up smaller incremental improvements like this one.

packages/python/openverse-attribution/src/openverse_attribution/license.py
```diff
@@ -1,128 +1,278 @@
...
-    PDM = "pdm"
-    PUBLIC_DOMAIN = "publicdomain"
+        This function validates the license, version and jurisdiction. If some
+        fields are not provided, it makes an attempt to deduce them.
 
-    def name(self, version: str | None = None) -> str:
+        The only exception is 'publicdomain' for which the version and
+        jurisdiction fields are ignored.
+
+        Use an empty string as ``jur`` to specify a generic, universal or
```
obulat, line 35:
```suggestion
        Use an empty string as ``jurisdiction`` to specify a generic, universal or
```

packages/python/openverse-attribution/src/openverse_attribution/license.py
```diff
@@ -1,128 +1,278 @@
...
+        self.fallback_jur = None
+
+        if self.name is LicenseName.PUBLICDOMAIN:
+            self.ver = None
+            self.jur = None
+            return
+
+        # Validate version against known versions.
+        if ver:
+            if ver not in all_licenses.keys():
```
obulat, line 65: Nit: These two conditions could be combined into one.

obulat (approved):
So many comprehensive changes ⭐ 

In the catalog, we have a function to get the license information based on the license name-license version pair, and/or the licence URL. I think we should add the ability to get `License` from a URL, without providing the license name-version pair. It should probably done as a separate PR, though.

packages/python/openverse-attribution/src/openverse_attribution/license.py
```diff
@@ -1,128 +1,278 @@
...
         """
-        Get the full name of the license.
-        The ``version`` parameter is disregarded and the version is always 1.0
-        for CC0, PDM and deprecated licenses. If not provided, the version is
-        omitted for all other licenses.
 
-        :param version: the version number of the license
-        :return: the full name of the license
+        # Shorten long variable names
+        ver = version
```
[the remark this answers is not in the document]
dhruvkb: I was enjoying the similarities in both the length (3 letters) and the pronunciations.

packages/python/openverse-attribution/src/openverse_attribution/license.py
dhruvkb: Fixed in cf8bc73.

packages/python/openverse-attribution/src/openverse_attribution/license.py
dhruvkb: Fixed in 9db7bb5.

PR #4254, opened 2024-05-02
Alter Django Admin media views to surface content reports

### LINKED ISSUE #3635: Create new content report Django admin table view
## Description

We will scrap the existing reports view in the Django Admin UI. Instead of showing a list of reports, we’ll show a list of reported media, sorted by the total number of pending reports, and then by the age of the oldest pending report. This will have the effect of sorting re-reported works to the top of the list, under the assumption that a work with multiple pending reports is probably higher priority, because a decision to report it was taken multiple times. By default, we will only show works with “pending” reports (reports where `decision_id = NULL`). This creates a queue to work through progressively, from top to bottom. Remove any manual sorting options for now. We can add those back if we get feedback that moderators want them, but for now they just clutter and confuse the interface, and make the foundational feature more complex.

Said another way:
- Create a new view based on the media object default model view but sort media by pending report count and the oldest report’s age
- Exclude media items that have no reports
- Filter to only show media with pending reports by default, but allow showing all reported media

## Context

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#id2

AetherUnbound:
## Fixes

Fixes #3635 by @sarayourfriend

## Description

This PR modifies the base media list views to accomplish the following:
- Exclude media items that have no reports
- Add columns for total report count, pending report count, oldest report date, and pending report links
- Automatically filter by records with pending reports, but allow viewing all reported records

This work was admittedly pretty hairy :sweat_smile: I'm hoping that the changes I've made here make sense, I'm certainly pleased that I was able to get all the functionality set up correctly! I'm happy to make anything more idiomatically "Django" where it makes sense to.

I also exposed the decision views so we could more easily add decisions for media, at least for testing. I'm sure that will change, and I'm honestly a little unclear on all of the moving parts here and how decisions will _actually_ be applied. This view currently links to each individual report, but since decisions can be applied to multiple reports _and_ multiple media records, I'm just not sure what all the actions will look like. That to say, this view might change a little once that's built out, which is okay!

### Screenshots
**Pending only filtered (default)**
Screenshot 2024-05-01_17-01-53

**All reported records**
Screenshot 2024-05-01_17-02-02

## Testing Instructions

The easiest way to do this is the following:

1. Run `just down -v && just api/init`
2. Run `just api/dbshell` and execute the following:

```sql
COPY public.api_imagedecision (id, created_on, updated_on, notes, action, moderator_id) FROM stdin;
1	2024-05-01 23:27:09.884471+00	2024-05-01 23:27:09.884487+00		rejected_reports	3
2	2024-05-01 23:32:51.386486+00	2024-05-01 23:32:51.386501+00		marked_sensitive	3
\.


COPY public.api_imagedecision_media_objs (id, imagedecision_id, image_id) FROM stdin;
1	1	893
2	2	111
\.


COPY public.nsfw_reports (id, identifier, reason, description, status, created_at, decision_id) FROM stdin;
1	2e22c7e7-58b7-486c-9df8-16e84c27478f	mature		pending_review	2024-04-26 20:44:31.271278+00	\N
2	2e22c7e7-58b7-486c-9df8-16e84c27478f	dmca		pending_review	2024-04-26 20:44:36.178746+00	\N
4	2e22c7e7-58b7-486c-9df8-16e84c27478f	mature		pending_review	2024-04-26 20:44:46.869982+00	\N
5	c296b69b-c8d7-4fc3-b820-a83de6138e0e	dmca		pending_review	2024-04-26 20:44:52.034395+00	\N
6	de37812d-2daa-49f6-91ab-e87e780ca653	dmca	asdfsadfsadfsadfsadfsadf	pending_review	2024-04-26 20:45:31.272964+00	\N
7	de37812d-2daa-49f6-91ab-e87e780ca653	dmca		pending_review	2024-05-01 21:12:53.177529+00	\N
8	c296b69b-c8d7-4fc3-b820-a83de6138e0e	dmca		pending_review	2024-05-01 21:21:10.442789+00	\N
9	ba711b67-2cbe-487a-9feb-19d70accf912	dmca		pending_review	2024-05-01 21:28:45.869087+00	\N
3	2e22c7e7-58b7-486c-9df8-16e84c27478f	other		pending_review	2024-04-26 20:44:42.274724+00	1
10	052d2cfc-fdcb-40de-994c-40eadef753fc	dmca		pending_review	2024-05-01 21:31:04.020163+00	2
\.
```

Then you should be able to go to http://localhost:50280/admin/api/image/ and see the views in the screenshots above!

The actual `image_id`s in the `api_imagedecision_media_objs` table might be non-deterministic, I'm not sure. It makes me wonder too if we should alter that table to use `identifier` in its table reference rather than `{media}_id`. It's what we do for the sensitive/deleted objects tables as well: https://github.com/WordPress/openverse/blob/0db784ebc4d4937c925dcb14c1465bafb3e20c69/api/api/models/image.py#L97-L96. CC @dhruvkb as you might be the best informed on that.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django.contrib import admin
 from django.contrib.admin.views.main import ChangeList
-from django.db.models import Count
+from django.db.models import Count, F, Min
 
 import structlog
 from elasticsearch import NotFoundError
@@ -31,12 +31,23 @@ class PredeterminedOrderChangelist(ChangeList):
 
 
 class MediaListAdmin(admin.ModelAdmin):
-    list_display = ("identifier", "report_count")
+    list_display = (
+        "identifier",
+        "report_count",
+        "pending_report_count",
+        "oldest_report_date",
+    )
     search_fields = ("identifier",)
     # Ordering is not set here, see get_queryset
 
     def report_count(self, obj):
         return obj.report_count
+
+    def pending_report_count(self, obj):
+        return obj.pending_report_count
+
+    def oldest_report_date(self, obj):
+        return obj.oldest_report_date
 
     def get_queryset(self, request):
         qs = super().get_queryset(request)
@@ -47,7 +58,14 @@ class MediaListAdmin(admin.ModelAdmin):
         qs = qs.filter(media_reports__isnull=False)
         # Annotate and order by report count
         qs = qs.annotate(report_count=Count("media_reports"))
-        qs = qs.order_by("-report_count")
+        # Show total pending reports by subtracting the number of reports
+        # from the number of reports that have decisions
+        qs = qs.annotate(
+            pending_report_count=F("report_count")
+            - Count("media_reports__decision__pk")
+        )
+        qs = qs.annotate(oldest_report_date=Min("media_reports__created_at"))
+        qs = qs.order_by("-report_count", "-pending_report_count", "oldest_report_date")
         return qs
 
     def get_changelist(self, request, **kwargs):
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -57,6 +57,9 @@ class MediaListAdmin(admin.ModelAdmin):
         "pending_reports_links",
     )
     list_filter = (PendingRecordCountFilter,)
+    # Disable link display for images
+    list_display_links = None
+    # Allow autocomplete to work from other referenced fields
     search_fields = ("identifier",)
     media_type = None
     # Ordering is not set here, see get_queryset
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -40,9 +40,9 @@ class MediaListAdmin(admin.ModelAdmin):
     def get_queryset(self, request):
         qs = super().get_queryset(request)
         # Filter down to only instances with reports
-        qs = qs.filter(media_report__isnull=False)
+        qs = qs.filter(media_reports__isnull=False)
         # Annotate and order by report count
-        qs = qs.annotate(report_count=Count("media_report"))
+        qs = qs.annotate(report_count=Count("media_reports"))
         qs = qs.order_by("-report_count")
         return qs
 
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -80,7 +80,8 @@ class MediaListAdmin(admin.ModelAdmin):
         return obj.oldest_report_date
 
     def pending_reports_links(self, obj):
-        pending_reports = obj.media_reports.filter(decision__isnull=True)
+        reports = getattr(obj, f"{self.media_type}_report")
+        pending_reports = reports.filter(decision__isnull=True)
         data = []
         for report in pending_reports.all():
             url = reverse(
@@ -95,16 +96,18 @@ class MediaListAdmin(admin.ModelAdmin):
         if "autocomplete" in request.path:
             return qs
         # Filter down to only instances with reports
-        qs = qs.filter(media_reports__isnull=False)
+        qs = qs.filter(**{f"{self.media_type}_report__isnull": False})
         # Annotate and order by report count
-        qs = qs.annotate(total_report_count=Count("media_reports"))
+        qs = qs.annotate(total_report_count=Count(f"{self.media_type}_report"))
         # Show total pending reports by subtracting the number of reports
         # from the number of reports that have decisions
         qs = qs.annotate(
             pending_report_count=F("total_report_count")
-            - Count("media_reports__decision__pk")
+            - Count(f"{self.media_type}_report__decision__pk")
         )
-        qs = qs.annotate(oldest_report_date=Min("media_reports__created_at"))
+        qs = qs.annotate(
+            oldest_report_date=Min(f"{self.media_type}_report__created_at")
+        )
         qs = qs.order_by(
             "-total_report_count", "-pending_report_count", "oldest_report_date"
         )
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -32,6 +32,7 @@ class PredeterminedOrderChangelist(ChangeList):
 
 class MediaListAdmin(admin.ModelAdmin):
     list_display = ("identifier", "report_count")
+    search_fields = ("identifier",)
     # Ordering is not set here, see get_queryset
 
     def report_count(self, obj):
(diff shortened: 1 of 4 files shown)


api/api/admin/media_report.py
```diff
@@ -10,6 +15,106 @@
...
+    def _get_default_ordering(self):
+        return []
+
+
+class PendingRecordCountFilter(admin.SimpleListFilter):
+    title = "pending record count"
+    parameter_name = "pending_record_count"
+
+    def choices(self, changelist):
+        """Override in order to get the default to be "pending" rather than "all"."""
```
sarayourfriend, line 39: Small nit:

```suggestion
        """Set default to "pending" rather than "all"."""
```

api/api/admin/media_report.py
```diff
@@ -10,6 +15,106 @@
...
+            )
+            data.append(format_html('<a href="{}">Report {}</a>', url, report.id))
+        return mark_safe(", ".join(data))
+
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        # Return all available image if this is for an autocomplete request
+        if "autocomplete" in request.path:
+            return qs
+        # Filter down to only instances with reports
```
sarayourfriend, line 96: Small whitespace nit, just helps break up the method and make it read a bit easier.

```suggestion

        # Filter down to only instances with reports
```

api/api/admin/media_report.py
```diff
@@ -10,6 +15,106 @@
...
+    def pending_reports_links(self, obj):
+        reports = getattr(obj, f"{self.media_type}_report")
+        pending_reports = reports.filter(decision__isnull=True)
+        data = []
+        for report in pending_reports.all():
+            url = reverse(
+                f"admin:api_{self.media_type}report_change", args=(report.id,)
+            )
+            data.append(format_html('<a href="{}">Report {}</a>', url, report.id))
+        return mark_safe(", ".join(data))
```
sarayourfriend, line 89: Ditto, just a nit for read/scan-ability.
```suggestion

        return mark_safe(", ".join(data))
```

sarayourfriend:
> The actual image_ids in the api_imagedecision_media_objs table might be non-deterministic, I'm not sure. It makes me wonder too if we should alter that table to use identifier in its table reference rather than {media}_id. It's what we do for the sensitive/deleted objects tables as well: https://github.com/WordPress/openverse/blob/0db784ebc4d4937c925dcb14c1465bafb3e20c69/api/api/models/image.py#L97-L96. CC @dhruvkb as you might be the best informed on that.

Yes, this needs to be to the identifier, not the `id`, otherwise you're right it will drift. This was missed as a requirement in the IP, though we should treat it as an assumed requirement. It's really frustrating to have these differences, I wish we could find a way to not have two identifiers, especially one that is ephemeral (kind of defeats the purpose of an identifier, really).

The fix is a two line change to add `to_field` on each media type's Decision model set to `identifier`:

```diff
diff --git a/api/api/models/audio.py b/api/api/models/audio.py
index 917be115f..163bb4d58 100644
--- a/api/api/models/audio.py
+++ b/api/api/models/audio.py
@@ -325,6 +325,7 @@ class AudioDecision(AbstractMediaDecision):
 
     media_objs = models.ManyToManyField(
         to="Audio",
+        to_field="identifier",
         db_constraint=False,
         help_text="The audio items being moderated.",
     )
diff --git a/api/api/models/image.py b/api/api/models/image.py
index d0ad5323a..2f349ab29 100644
--- a/api/api/models/image.py
+++ b/api/api/models/image.py
@@ -138,6 +138,7 @@ class ImageDecision(AbstractMediaDecision):
 
     media_objs = models.ManyToManyField(
         to="Image",
+        to_field="identifier",
         db_constraint=False,
         help_text="The image items being moderated.",
     )
```

Best to either add that in this PR or open an issue as a fast follow.

Similarly, (and this is _definintely_ for a separate issue), it'd be nice to set `related_name` on all of these report, decision, etc, so that they use the same name for each model, excluding the media type. That would make things like this unncessary:

```py
            - Count(f"{self.media_type}_report__decision__pk")
```

That would just need to be `Count("report__decision__pk")` if `related_name="reports", related_query_name="report"` was set on each media type's implementation. Neither would cause a SQL migration, as far as I am aware (these are virtual fields) but would make writing isomorphic functions for all of this stuff a lot more straightforward (no need to pass in `media_type` everywhere in this code).

sarayourfriend (approved):
This LGTM as is. I've left a comment describing the fix for the identifier issue you found. Either add that to this PR or create an issue as a fast-follow. Fixing that will block rolling the feature out, and it should be fixed as soon as possible to avoid introducing more code that will reference the incorrect identifier in queries.

dhruvkb (approved):
LGTM, seconding @sarayourfriend's comments about using `identifier` field instead of `id`.

AetherUnbound:
@sarayourfriend It looks like `to_field` is not available for Fields:

```
web-1  |   File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
web-1  |   File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
web-1  |   File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
web-1  |   File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
web-1  |   File "<frozen importlib._bootstrap_external>", line 940, in exec_module
web-1  |   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
web-1  |   File "/api/api/models/__init__.py", line 2, in <module>
web-1  |     from api.models.audio import (
web-1  |   File "/api/api/models/audio.py", line 321, in <module>
web-1  |     class AudioDecision(AbstractMediaDecision):
web-1  |   File "/api/api/models/audio.py", line 326, in AudioDecision
web-1  |     media_objs = models.ManyToManyField(
web-1  |                  ^^^^^^^^^^^^^^^^^^^^^^^
web-1  |   File "/.venv/lib/python3.11/site-packages/django/db/models/fields/related.py", line 1370, in __init__
web-1  |     super().__init__(
web-1  |   File "/.venv/lib/python3.11/site-packages/django/db/models/fields/related.py", line 108, in __init__
web-1  |     super().__init__(**kwargs)
web-1  | TypeError: Field.__init__() got an unexpected keyword argument 'to_field'
```

Since this is approved and the way to get reference fields working correctly isn't immediately obvious to me, I'm going to go ahead and merge this and create an issue for following up on the field name.

Issue #3435, opened 2023-12-01
Nuxt structured logging


sarayourfriend:
## Problem

See the milestone description for the rationale: https://github.com/WordPress/openverse/milestone/20

## Description

Add [the Node.js logging library Pino](https://github.com/pinojs/pino) to the Nuxt application. Update our existing logging to use Pino. Make sure Pino is only included in the client bundle in the development build (we currently silence all logs in production builds so we do not need to include the library in the client bundle).

sarayourfriend:
Nuxt 3 uses consola, which means it will already be in the bundle, so we should use it instead of Pino.

PR #4257, opened 2024-05-02
Migrate to Nuxt 3 

### LINKED ISSUE #411: Nuxt 3 migration
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2023-06-05        | TBD        | @obulat           | TBD              |

## Description

Vue 2 is reaching end-of-life at the end of 2023[^1]. We need to update to Nuxt 3 to use Vue 3.

Node 16, which is required _to build_ the frontend app (Nuxt 2 applications can theoretically _run_ on Node >16) reaches end-of-life on September 11th, 2023.

## Documents

- [x] #2311
- [x] #2324

## Issues

- [x] PR in Nuxt 2.16.xx: https://github.com/nuxt/nuxt/pull/19518
- [x] #815 
- [x] #2232
- [x] #2234

## Prior Art

Several members of the team have attempted the migration previously but without success. This is one of such attempts: https://github.com/WordPress/openverse-frontend/pull/313
The Nuxt team has said that working on simplifying the migration is one of their priorities for 2023, so hopefully, it will be easier this time.

[^1]: Vue 2 LTS: https://v2.vuejs.org/lts/

### LINKED ISSUE #3106: Unable to select a new image from the home page after clicking one and returning using browser-back
## Description
The bug in the website is that on the home page of the website, there are 9 randomly generated pictures and when we click on any one of them, it opens the whole page related to that image but when we go back and again click another image it again shows the details of the previously opened image or we can say of the first image opened.

## Screenshots
Here is the screen recording of the bug.
[](https://bit.ly/3PGOwQX)

## Additional context
1. Deletion of the section is not required as it is quite good thinking to add such a section the case is just that if improved it will be a great experience for the users.

zackkrida:
## Fixes
Fixes #411
Fixes #3106

## Description
Originally, this PR description contained information on how to deploy Nuxt 3 to staging. This was incorporated in the PR. To view them, look at the history of this comment.

Information described here:
- the most important items to review in this PR
- the large changes in patterns and structures
- smallish changes that were required to migrate to Nuxt 3
- enhancements that were not necessary for Nuxt 3 migration, but were useful and easy to make here

## The most important items to review

- `/server` directory contains the server routes (healthcheck and robots), as well as the server-side Sentry and server-side logger that uses `Consola`.
- `/data/api-service` - the code used for requesting data from Openverse API.
- `app.vue` - the single entry point to the frontend app.
- `nuxt.config.ts` - static configuration file for the app, plugins and modules.
- `pages/search.vue` and `middleware/search.ts` - the page component and the middleware responsible for fetching and displaying search results
- `pages/collection.vue` and `middleware/collection.ts` - the page component and the middleware responsible for for fetching and displaying collection results
- `pages/image/[id]/index.vue`, `pages/audio/[id]/index.vue` and `middleware/single-result.ts` - the page components and the middleware responsible for fetching and displaying the single result pages
- `Dockerfile`

Infra-related questions:
- Does this PR set up all the necessary env variables for the app to run (when we run `just start` in production)? Probably should be solved in an accompanying infra PR (if any changes needed)

### API service
All the various `service`s were replaced with a single `api-service` that has only the methods that we use. The extra methods that were not used by the frontend (e.g., `put`, `update`, `delete`) were removed.
The API service uses `useRuntimeConfig` for the API URL, and uses the API token from `useNuxtApp`.

### app settings and `runtimeConfig`
_Added 2024-07-02_
Previously, we often used the `env` variables, and manually added the `env` variables to the `nuxt.config`'s `env` property. Here, we moved `env` properties to the public `runtimeConfig`.
With Nuxt 3, we use `runtimeConfig` to keep track of app configuration such as the API URL. Before I updated Nuxt 3 to the latest version, calling `useRuntimeConfig` or `useNuxtApp` would often throw an error because it would not have access to the Nuxt context. The call to `useNuxtApp` would be nested in several layers of function calls (`useFetch` -> store's `fetchMedia` -> api-service's `fetch` function). However, with Nuxt 3.12, all of these errors disappeared, so I removed the `try/catch` guards from these calls.

### `env` variable changes
_Added 2024-07-15_
Nuxt 3 sets up the `runtimeConfig` values using the env variables provided when running the app. The convention is to use a variable with `NUXT_` prefix for private variables and `NUXT_PUBLIC_` prefix for public variables. Currently, `nuxt.config` sets some of the variable defaults for local development. 
It will be necessary to provide the following env variables to the production build:
```
# --------------------------------- #
# REQUIRED PRODUCTION ENV VARIABLES
# --------------------------------- #
# NUXT_PUBLIC_SENTRY_DSN=https://b6466b74788a4a2f8a7912eea912beb7@o787041.ingest.sentry.io/5799642
# NUXT_PUBLIC_SENTRY_ENVIRONMENT=production
# SITE_URL=https://openverse.org
# PLAUSIBLE_SITE_URL=https://openverse.org
# NUXT_PUBLIC_API_URL=https://api.openverse.org/
# NUXT_API_CLIENT_ID=
# NUXT_API_CLIENT_SECRET=
``` 

### Fetching data in pages
The `useFetch` from `@nuxtjs/composition-api` was replaced with Nuxt 3 function  `useAsyncData` in the `search`, `collection` and `single-result` pages use the for fetching the data.

All of the server-side fetching for search, collection and single result pages is done in the middleware.
I tried moving the fetching to the `useAsyncData` to unify and simplify it. However, this causes client-server mismatch because the fetching is done in the sub-page, after the surrounding components such as header and footer have already been rendered. This means that the result count is always set to 0 in the header.
When hydrating, the client expects it to be the real number because the store has the updated data, and the header is being rendered with that. We could wrap the changing items with `<ClientOnly>` to prevent the mismatch, but it's much easier to fetch the data in the middleware, before the app starts rendering, and have the correct data rendered on the server everywhere.

### Single entry point, `app.vue`
Nuxt 3 introduces a new single entry point, `app.vue`. This is used to add the functionality that should work whenever a user gets a server-rendered page, wether it's a homepage, a search page, a single result page or a static page. 
Previously, we used the global middleware for the functions that do not require the app to be mounted and do not require a window access. For the functions that need the app to be mounted (such as updating the layout values with the cookies or syncing values with local storage), we added them to every layout file.

Layout files became much simpler because most functionality was moved to `app.vue`

Note: the only page not using the `app.vue` is `error.vue`. We must duplicate the i18n properties and the layout update from cookies there.

### Trailing slashes
After @sarayourfriend's [comment](https://github.com/WordPress/openverse/pull/4257#discussion_r1670622082) on trailing slashes, I reviewed our approach to trailing slashes in the frontend. Previously, our approach was slightly inconsistent. The only route that had a trailing slash was the "All media" search path, which had a trailing slash between the path and the query (`https://openverse.org/search/?q=cat`). All other paths did not have trailing slashes.
In this Nuxt 3 branch, the trailing slash from the search route was removed, so now no routes have trailing slashes.

### Modals
Vue 3 and Nuxt 3 natively support teleports, so the modals were updated to use the native support. We no longer need a modal target because Nuxt automatically creates a `#teleports` element.
[8655ced](https://github.com/vuejs/core/commit/8655ced480ea0fe453ff5fe445cecf97b91ec260) in Vue version 3.4.32 broke the way `VInputModal` was rendering the recent searches modal on mobile on SSR. This is why I rewrote the `VHeaderMobile` component to use a simpler component with `VModalContent`.

### Plausible types don't allow `null`
 The new Plausible plugin uses `plausible-tracker` JS library, and the event properties can only be `string`, `number` or `boolean` [there](https://github.com/plausible/plausible-tracker/blob/ab75723ad10660cbaee3718d1b0a670e2dfd717d/src/lib/request.ts#L29). So, all the `null` values in analytics events were replaced with the string `"null"` in this PR.

### import.meta
`process.*` was replaced with `import.meta` [Link to Nuxt docs](https://nuxt.com/docs/api/advanced/import-meta) and [JS docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta)

### i18n keys
`i18n` now checks for the correctness of the keys. So, if there are strings with non-ASCII keys (e.g., "text of the {translated keys containing non-ascii letters}"), the build fails. This is why I added a quick fix for this. It should be improved later (and the translations need to be updated in GlotPress)

### script setup for pages
All pages were converted to `script setup` because it's impossible to set the `layout` and the `middleware` in a component that uses `defineComponent` imported from `vue`.
The new Vue 3 convention is to have `script` tag at the top of the component. In this PR, we keep to the previous convention when the top tag was `template`, followed by `script`.
In the follow-up issues, we should switch the order of `script` and `template` for components using `script setup`. We should also open an issue to convert the components to `script setup` syntax, which can be labeled as `help wanted`.

### SVG sprite

SVG sprite module was vendored in because it did not work with our folder structure, where we have a `/src` top folder.

### ua-parse

`ua-parse` was removed because the library it uses does not have ESM, and we don't use `UA` anywhere.

### splitting attrs to classes and other attrs
In Vue 2, the `class` and `style` were not passed to the child component as part of `$attrs`, but instead were added to the outermost element of the child component. This broke the styles of the components that used `inheritAttrs: false` (`VSearchBar`, `VSelectField`, `VItem`).

### Media item preprocessing
We already pre-process the API responses to add necessary properties (`frontendMediaType`) or to fix encoding/decoding problems.
I added the extraction of the filetype from the main media URL. This way, we don't have to replace the filetype on the client-side after the image/audio have been loaded. This removes the flickering and also the occasional client-server mismatch.

### UUID check
Added a `validateUUID` check to the single result page to directly show an error page if the path `id` parameter is not a valid UUID instead of attempting to fetch it.
### DevEx improvement
The server API requests are logged in the console to simplify debugging.

### Replacing ad-hoc implementations with Vueuse
`use-window-scroll` composable was replaced by the composable from `vueuse`

### Homepage images issue
Fixed the [issue](https://github.com/WordPress/openverse/issues/3106) when the image on a single result page was not updated if you selected one from the homepage, went back and selected another one. 

### `v-for` keys
In `v-for` elements I replaced indexes with a better key option.

## Disabled or not working items
- Storybook (waiting for module update)
- Prometheus
- ua-parse
- Storybook tests

## Follow-up TODOs
- Add structured logging to both the Nitro and Nuxt parts of the app (#3435)
- Move the `script` tag to the top of the components that use `script setup` syntax
- Convert all components to `script setup` syntax (this should be a meta issue, and can be `help wanted`/`good first issue`)


frontend/test/unit/test-utils/render-suspended.ts
```diff
@@ -0,0 +1,185 @@
+// Copied from https://github.com/nuxt/test-utils/blob/bc86d23b2cf92aa3c3f19d52944452e4e5c27d63/src/runtime-utils/render.ts
```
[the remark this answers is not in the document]
obulat: I did try the updated version of this file (with the changes I added), but it didn't work, the tests failed to render the components. So, I decided to keep it as is for now.

sarayourfriend:
> We need to provide the env variables to the process when starting the app that was built for production. How do we do that?

@obulat can you clarify what you mean? Do you mean that we need to define production variables at build time? Or that we need to add new environment variables to the production execution environment?

sarayourfriend:
> We could wrap the changing items with <ClientOnly> to prevent the mismatch, but it's much easier to fetch the data in the middleware, before the app starts rendering, and have the correct data rendered on the server everywhere

That sounds great to me, I agree it's a better architecture. There are many different parts of our app that rely on the same data, and for the search pages it makes sense for that to go top-down, from the abstract "start of the request" to the real rendering of the page.

> Vue 3 and Nuxt 3 natively support teleports, so the modals were updated to use the native support. We no longer need a modal target because Nuxt automatically creates a #teleports element.

Awesome!

frontend/nuxt.config.ts
```diff
@@ -48,49 +28,14 @@ const favicons = [
...
-    },
-    {
-      rel: "search",
-      type: "application/opensearchdescription+xml",
-      title: "Openverse",
-      href: "/opensearch.xml",
-    },
-  ],
-}
+const isProd = import.meta.env.NODE_ENV === "production"
```
sarayourfriend, line 31: Can we change the name of this variable? `isProd` makes it sound like the deployment environment, but it's actually the type of runtime. Maybe `isDistributableBuild`, or something? This confusion is evident in this file, even, where sentry's `environment` is set to `environment: isProd ? "production" : "local",`, which is incorrect. It should match `deploymentEnv: process.env.DEPLOYMENT_ENV ?? "local",`.

frontend/nuxt.config.ts
```diff
@@ -106,80 +51,89 @@ const openverseLocales = [
...
+      // Can be overwritten by NUXT_PUBLIC_API_URL env variable
+      apiUrl,
+      providerUpdateFrequency: 3600000,
+      savedSearchCount: 4,
+      sentry: {
+        dsn: isProdNotPlaywright
+          ? "https://b6466b74788a4a2f8a7912eea912beb7@o787041.ingest.sentry.io/5799642"
+          : "",
+        environment: isProd ? "production" : "local",
+        release: "",
```
sarayourfriend, line 106:
```suggestion
        environment: process.env.DEPLOYMENT_ENV ?? "local",
        release: process.env.SEMANTIC_RELEASE,
```

frontend/nuxt.config.ts
```diff
@@ -106,80 +51,89 @@ const openverseLocales = [
...
+  routeRules: {
+    "/photos/**": { redirect: { to: "/image/**", statusCode: 301 } },
+    "/meta-search": { redirect: { to: "/about", statusCode: 301 } },
+    "/external-sources": { redirect: { to: "/about", statusCode: 301 } },
+  },
+  tailwindcss: {
+    cssPath: "~/styles/tailwind.css",
   },
   i18n: {
     baseUrl: "https://openverse.org",
```
sarayourfriend, line 139: I know this isn't changed in the diff, but why doesn't this match the actual deployment URL? Shouldn't this be `staging.openverse.org` for staging, or does it not matter?

frontend/server/utils/logger.ts
```diff
@@ -0,0 +1,9 @@
+import { consola } from "consola"
+
+const logger = consola.withTag("Openverse")
+// In production, `info` on the server and silent on the client.
+// In other environments, `debug`.
+logger.level =
+  process.env.NODE_ENV === "production" ? (import.meta.server ? 3 : -999) : 4
```
sarayourfriend, line 7: This is, again, a confusing use of 'environments'. Does it mean the build type or the actual deployment environment?

Regardless, ideally the log level should be configurable by an environment variable like it is for the API. We should be able to run staging or production with increased logging if we need to investigate something.

frontend/src/composables/use-match-routes.ts
```diff
@@ -29,10 +30,10 @@ export const useMatchRoute = (
     return routes.includes(route.split("__")[0])
   }
 
-  const matches = ref(routeNameMatches(route.value?.name))
+  const matches = ref(routeNameMatches(String(route?.name)))
```
sarayourfriend, line 33: Just curious, is `String()` needed for TypeScript, to actually convert it to a string, or something else? Why not `.toString()`? Or when is it not a string to begin with?

frontend/nuxt.config.ts
```diff
@@ -106,80 +51,89 @@ const openverseLocales = [
...
-    "portal-vue/nuxt",
+    "@pinia/nuxt",
     "@nuxtjs/i18n",
-    "@nuxtjs/proxy",
-    "@nuxtjs/redirect-module",
-    "@nuxtjs/sentry",
-    "cookie-universal-nuxt",
-    "vue-plausible",
-    "~/modules/prometheus.ts",
-    // Sitemap must be last to ensure that even routes created by other modules are added
```
sarayourfriend, line 173: Should we keep this comment?

frontend/src/composables/use-sprite.ts
```diff
@@ -0,0 +1,42 @@
...
+
+function generateName(name: string) {
+  return name
+    .toLowerCase()
+    .replace(/\.svg$/, "")
+    .replace(/[^a-z0-9-:]/g, "-")
+    .replace(/:/g, "--")
+}
+
+export const useSprite = (name: string) => {
```
sarayourfriend, line 20: Is this the module you said was vendored in? Where does it come from? Is it https://github.com/nuxt-modules/svg-sprite? We should include a source, credit, attribution, and link to an issue with a plan to remove the need for vendoring it.

frontend/src/constants/media.ts
```diff
@@ -68,4 +68,4 @@ export const contentStatus = Object.freeze({
 } as const)
 
 export const searchPath = (type: SearchType) =>
-  `/search/${type === ALL_MEDIA ? "" : type}`
+  `/search${type === ALL_MEDIA ? "" : "/" + type}`
```
sarayourfriend, line 71: So does nuxt 3 not have trailing slashes now? We went around and around trying to always use or not use trailing slashes. If we're changing it again, that should definitely be made clear.

frontend/src/data/api-service.ts
```diff
@@ -120,100 +170,66 @@ export const createApiService = ({
     return config
   })
   client.interceptors.response.use(
-    (response) => response,
+    (response) => {
+      logger.debug(
+        ">>>",
```
sarayourfriend, line 175: What is this `>>>` used in the debug logs in this module? Is that a Consola thing?

frontend/src/error.vue
```diff
@@ -0,0 +1,70 @@
...
+        <NuxtLayout name="default">
+          <VFourOhFour class="flex-grow" :error="error" />
+        </NuxtLayout>
+        <div id="modal"></div>
+      </Body>
+    </Html>
+  </div>
+</template>
+
+<script setup lang="ts">
```
sarayourfriend, line 32: Most examples of script setup I see have it above the template. Here's Vue's docs first example they show you of an SFC: https://vuejs.org/guide/introduction.html#single-file-components

Is breaking that convention intentional, or just matching our existing code (and old Vue style), or will we eventually move towards it when converting things to script setup?

frontend/server/utils/logger.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export { logger }
```
sarayourfriend, line 9: I've noticed a mix of logger and console being used. When are we supposed to use one or the other?

frontend/src/plugins/api-token.server.ts
```diff
@@ -1,11 +1,13 @@
+import { defineNuxtPlugin, useRuntimeConfig } from "#imports"
```
sarayourfriend, line 1: Should this module move to server/plugins?

sarayourfriend (commented):
I reviewed as much as I could this sitting, but I've lost energy and need to stop for today. I will pick back up the review next week.

I've left some comments. So far nothing much jumps out at me as potentially problematic, most of my questions are clarifications and asking about consistency decisions and other things. I don't expect the merged code will be 100% consistent or in the absolute best possible shape (it's simply far too many changes and far too much code for that to be possible), but even if we don't make the changes _now_, it would be good to have a record of things like where plugins _should_ ideally live, how the logger should work and be used, etc.

sarayourfriend (commented):
@obulat our conversation just now about the language-header based caching reminded me: we need to make sure the Nginx configuration will still work for Nuxt 2, especially since the changes from https://github.com/WordPress/openverse/pull/4516.


I'll make a point to look into that next week if you don't have time to review it.

Update: follow up here: https://github.com/WordPress/openverse/issues/411#issuecomment-2201946777

frontend/src/plugins/api-token.server.ts
obulat: It probably should move to `server/plugins`, or to a separate module for the API interaction, but I think it's best to defer this to a separate follow-up issue. It is working as is, so I'd prefer not to touch it yet :)

frontend/server/utils/logger.ts
obulat: Initially, I thought we can't use `logger` from `/server` in the `nuxt` `/src` part of the app, but I forgot about it and logger seems to work in the `/src`, too. I'll try moving the logger to `/shared` top-level folder to show that it's used both by the Nitro server and the Nuxt app.

frontend/src/error.vue
obulat: This was done to match our existing code, to minimize the number of changes. It would probably increase the cognitive load when some components have template on top and others - script. We should probably convert everything to `script setup` and _then_ change the order. On the other hand, we could also move the script to the top to make it easier to distinguish between the `script setup` and `script` components at a glance. What do you think?

frontend/src/data/api-service.ts
obulat: I used it so that the API request is easy to see in the logs. Do you think it's not necessary, or there's a better way to display this?
I have often had problems when debugging the server rendering: has the request been fired, have we received a response, or was there some other problem during the API request/response phase? It's easy to debug the browser-based requests, but for there usually is no information about the server requests, so I decided to add it here.

frontend/src/constants/media.ts
obulat: Thank you for flagging this, I'll re-add the trailing slash using `site.trailingSlash` setting in `nuxt.config`: https://nuxtseo.com/nuxt-seo/guides/trailing-slashes#enabling-trailing-slashes for consistency

frontend/src/composables/use-match-routes.ts
obulat: Route now can be a string or a Symbol. And you're right, I'll replace this with `.toString()`.

frontend/server/utils/logger.ts
obulat: All of the `process.env.NODE_ENV === "production"` checks were using the `isProd` variable, which was set to `process.env.NODE_ENV === "production"`:
https://github.com/WordPress/openverse/blob/97ff97dd7201542dc6103c8c6091f4045181b88a/frontend/src/utils/node-env.ts#L3

Your comments show that these checks were previously incorrect. I'll review them.

frontend/nuxt.config.ts
obulat: The docs say that this should not be used if we set the VueRouter's base URL (at least in Nuxt 3):

> The fallback base URL to use as a prefix for alternate URLs in hreflang tags. By default VueRouter's base URL will be used and only if that is not available, fallback URL will be used.

However, in staging, all `hreflang` alternate links do use `https://openverse.org`. I'll check what happens with Nuxt 3 if we remove this setting

frontend/src/error.vue
sarayourfriend: Up to you! I think it'd be good to match the typical expectations someone working with Vue would have, which I suppose would be the format the Vue documentation uses (personally I find it a bit easier to reason about too, for some reason). I don't think consistency between the old options SFC and the new script setup SFC is important, or at least not _more_ important than consistency with Vue community's expectations and documentation's examples styles.

Does not need to be something addressed in this PR though, can definitely be a follow up issue, if preferred. Moreso it would be good to decide and set a specific expectation so we're all working towards the same goal.

frontend/src/data/api-service.ts
sarayourfriend: No, that's great, I'm moreso asking just about the `>>>` specifically. I suppose it's just there to add a prefix to the logs? I haven't tried the PR to see whether the logs have an implicit format (timestamp prefix, etc). I'm just a little wary of adding haphazard logging, even debug logs. I'd be thrilled to have logs of outbound requests from SSR in production even, but the `>>>` isn't a parseable prefix :sweat_smile: 

We have this issue to use structured logging via Consola: https://github.com/WordPress/openverse/issues/3435 we can address the issue then. It would be better not to add things like `>>>` to begin with (if indeed it isn't a special signifier for Consola) but it's not an important issue. We can resolve that in the issue.

frontend/src/error.vue
obulat: I'll add the notes on the script setup components order to the PR description. It seems best to create a follow up set of issues for moving the script to the top of the file, and to convert components to script setup.

frontend/src/constants/media.ts
obulat: I was wrong about this.

Our current setup is inconsistent: we only add the trailing slash to the "All media" search endpoint (so, the URL is `https://openverse.org/search/?q=cat` with a slash between the path and the query), and all other endpoints do not have a trailing slash (so, image search URL is `https://openverse.org/search/image?q=cat, with no slash between the path and the query).

In this branch, we remove trailing slashes from the All media search URL, so now we have no trailing slashes in the frontend.

I think it should be okay to change the URL structure for the search route because we disable robots on it and make it not searchable by search engines.

frontend/src/composables/use-match-routes.ts
obulat: Added a `getRouteNameString` function, and replaced all occurrences of `String()` with `toString`.

frontend/src/data/api-service.ts
obulat: Removed `>>>`.
We can replace the `src`'s logger with a structured logging Consola setup in a follow-up PR.

frontend/src/composables/use-sprite.ts
obulat: I added the comments to `use-sprite`, as well as `VIcon` and `VSvg`

frontend/nuxt.config.ts
obulat: The docs for nuxt sitemap don't say anything about placing the module last, so I removed it.

frontend/nuxt.config.ts
```diff
@@ -191,160 +145,20 @@ const config: NuxtConfig = {
...
-        viewports: VIEWPORTS,
-      },
-    },
+    trailingSlash: false,
+    vueI18n: "./src/vue-i18n",
   },
   plausible: {
+    enabled: !isTest,
+    logIgnoredEvents: !isProductionBuild,
     trackLocalhost: !isProdNotPlaywright,
```
zackkrida, line 159:
```suggestion
    ignoredHostnames: ["localhost", "staging.openverse.org"]
```

It looks like this `trackLocalhost` field is deprecated now for a new one:

```
ignoredHostnames?: string[] | undefined
Hostnames to ignore when tracking events.

@default

['localhost']
```

frontend/justfile
```diff
@@ -49,6 +50,18 @@ run-img tag="openverse-frontend:local":
...
+# You must set the CR_PAT env variable to your GitHub token before running this script to enable ghcr login.
+# export CR_PAT=YOUR_TOKEN
+push-to-nuxt-preview github_username:
+    echo $CR_PAT | docker login ghcr.io -u {{github_username}} --password-stdin
+    just build-img ghcr.io/wordpress/openverse-frontend:nuxt-preview
+    docker build -t ghcr.io/wordpress/openverse-frontend_nginx:nuxt-preview -f ./Dockerfile.nginx .
+    docker push ghcr.io/wordpress/openverse-frontend:nuxt-preview
+    docker push ghcr.io/wordpress/openverse-frontend_nginx:nuxt-preview
+    gh workflow run -R WordPress/openverse-infrastructure 'Deployment - production-frontend-nuxt-3' -f tag=nuxt-preview
+
```
zackkrida, line 64: Just a TODO reminder: We should remove this before merging.

frontend/server/routes/robots.ts
zackkrida, line 1: This is not working at `/robots.txt`, only `/robots`. PR to fix here:

https://github.com/WordPress/openverse/pull/4659

frontend/server/plugins/sentry.ts
```diff
@@ -0,0 +1,35 @@
...
+  const {
+    public: { sentry },
+  } = useRuntimeConfig()
+
+  if (import.meta.dev) {
+    // Should not need this, but without it Dev builds fail with "The requested module 'vue' does not provide an export named 'computed'"
+    // @see https://github.com/getsentry/sentry-javascript/issues/12490
+    // @ts-expect-error - globalThis is not defined here, and this is only used in dev mode.
+    globalThis._sentryEsmLoaderHookRegistered = true
+  }
```
zackkrida, line 18: Good news, this appears to be unnecessary with @sentry/node "^8.20.0"!

https://github.com/getsentry/sentry-javascript/issues/12490#issuecomment-2214961942

I tested locally with `j f` and it did not show any error. Is that what the problem was, or did it fail on dev _builds_ specifically, rather than the dev local server?

frontend/src/components/VHomeGallery/VHomeGallery.vue
```diff
@@ -13,8 +13,7 @@
       }"
     >
       <ClientOnly>
-        <Component
-          :is="prefersReducedMotion ? 'div' : 'Transition'"
+        <Transition
           v-for="(image, idx) in imageList"
```
zackkrida, line 17: This diff will fix the animations:

```diff
diff --git a/frontend/src/components/VHomeGallery/VHomeGallery.vue b/frontend/src/components/VHomeGallery/VHomeGallery.vue
index 1e8cc48b1..c6c4aa866 100644
--- a/frontend/src/components/VHomeGallery/VHomeGallery.vue
+++ b/frontend/src/components/VHomeGallery/VHomeGallery.vue
@@ -17,8 +17,8 @@
           v-for="(image, idx) in imageList"
           :key="idx"
           enter-active-class="transition-opacity delay-[var(--delay)] duration-500"
+          enter-from-class="opacity-0"
           leave-active-class="transition-opacity delay-[var(--delay)] duration-500"
-          enter-class="opacity-0"
           leave-to-class="opacity-0"
           mode="out-in"
           appear
```

zackkrida:
@obulat I've had the opportunity to review this for several hours and I've made some suggestions. I think the homepage animation and robots.txt concerns are worth blocking over. I've also provided fixes for both :smile: 

This is phenomenal work, and once my comments have been addressed we should push to our nuxt preview domain once more. Then, on Monday or a bit later we can conduct a lengthy test wherein a large percentage or all traffic is routed to the Nuxt 3 code. That would give us much assurance about merging and _finally_ deploying this PR.

I am confident the work here will give us a stable foundation for iterating on the frontend, cleaning up legacy code patterns, and migrating to Nuxt 4. Amazing work!

frontend/server/plugins/sentry.ts
obulat: Problem is indeed fixed, thank you!

frontend/src/utils/og.ts
```diff
@@ -16,7 +16,13 @@ export const createDetailPageMeta = ({
...
 }) => {
   const head: { meta: Meta } = { meta: [] }
-  const meta: Meta = []
+  const meta: Meta = [
+    {
+      key: "robots",
+      name: "robots",
+      content: "all",
+    },
+  ]
```
zackkrida, line 25: Is this necessary? I thought by default the seo robots plugin was adding the correct robots meta tags to pages and that we could avoid setting these values outside of the nuxt config. That was the paradigm I tried to embrace in #4659, at least.

frontend/package.json
```diff
@@ -56,94 +59,59 @@
...
+    "node-html-parser": "^6.1.13",
+    "npm-run-all2": "^6.2.2",
+    "nuxt": "3.12.4",
+    "rimraf": "^6.0.1",
+    "talkback": "^4.2.0",
+    "typescript": "5.5.3",
+    "vitest": "^2.0.4",
+    "vitest-dom": "^0.1.1",
+    "vue": "3.4.33",
+    "vue-router": "^4.4.0",
```
zackkrida, line 113: Just noting: This release includes a partial fix for the memory leak issue! https://github.com/vuejs/router/pull/2137

frontend/src/utils/og.ts
obulat: I noticed that I incorrectly reversed the `seo` Playwright test changes from the single result fix PR. So, when fixing this, I re-added this value here.
Then I tested the set up again, and realized that the robots setup used a trailing slash for `disallow` routes and therefore was creating incorrect meta tags and robots settings. Updated in the latest commit. I'm not sure what the difference between the value `all` and the default value used by robots module is.

obulat (approved):
🎉

Issue #4166, opened 2024-04-19
Convert usages of Pipenv to PDM


dhruvkb:
## Description

We have opted to use PDM as our package manager for the API and various other Python libraries in the monorepo (with the API migration happening in #4107). We still using Pipenv in many other places and in the interest of consistency between our toolchains, we would like to migrate all these projects over to PDM as well (potentially also reorganise them into the new Python packages directory).

- [x] `automations/python/` - #4625
- [x] `documentation/` - @dhruvkb #4266 
- [-] `ingestion_server/`  - moved to #4658
- [x] `utilities/provider_tallies/` - #4624
- [x] `utilities/dead_links/` - #4623
- [x] `utilities/project_planning/`- #4622
- [x] `utilities/generate_test_locales/` - #4626

## More

We also have many references to Pipenv, specifically `Pipfile` and `Pipfile.lock` scattered across the codebase. Once these projects are no longer using Pipenv, these references will also need to be updated.

- [x] `documentation/general/general_setup.md`
- [x] `documentation/automations/guides/quickstart.md` - #4625
- [-] `documentation/meta/maintenance/elasticsearch_cluster.md`
- [-] `ingestion_server/Dockerfile` - moved to #4658
- [-] `ingestion_server/.dockerignore` - moved to #4658
- [-] `ingestion_server/justfile` - moved to #4658
- [-] `.pre-commit-config.yaml` - moved to #4658
- [-] `.github/actions/setup-env/action.yml` - moved to #4658

## Additional context

The goal is to make a number of small atomic PRs, that migrate only one project and its references per PR, so that they can be quickly verified to work and merged. Please comment below which tasks from above you will be working before starting work on them.

malik-aditya-r:
Hey @WordPress/openverse-maintainers, If this issue is still available I'd like to work on it.

dhruvkb:
Yes @malik-aditya-r, it is still available. Please mention which of the above you will be migrating and make small PRs for each target. I'll assign this to you.

malik-aditya-r:
Hey @dhruvkb Thanks for assigning me the task. I'll like to start with automations/python and once that is successfully done move on to utilities/* and others.
I'll start by going through the PR #4107 And will reach out if any questions arise.

obulat:
@dhruvkb, it seems that there are only 2 non-ingestion-server items that need to be fixed here.

[...]

PR #4262, opened 2024-05-03
Automatically convert URLs to hyperlinks using `linkify` extension

### LINKED ISSUE #4208: Automatically warn or convert unformatted URLs in the documentation site
## Description

Context: #4173 and #4195 

We issued a one-off change to update all links in the [DAGs documentation](https://docs.openverse.org/catalog/reference/DAGs.html) page, but it's likely that we will encounter the issue of unclickable links in documentation in the future.

We should build out either a Sphinx plugin or a Vale linting rule to handle these cases, and either warn when a link is unformatted (e.g. `http://{link}` instead of the correct `<http://{link}>`) or automatically convert those links into the formatted version.

@sarayourfriend may be able to provide context on the Vale approach.

For the Sphinx approach, we could use something like our existing extensions which automatically [link GitHub usernames](https://github.com/WordPress/openverse/blob/331b576ad8dd99dc00c6a777a3b3addc6dbd3225/documentation/_ext/link_usernames.py) or [GitHub issues](https://github.com/WordPress/openverse/blob/331b576ad8dd99dc00c6a777a3b3addc6dbd3225/documentation/_ext/link_issues.py). We would have a similar extension which would rewrite unformatted links to the formatted version in the produced HTML files.

## Alternatives

Leave this without a check and occasionally audit the project to ensure all links in the documentation are clickable.

madewithkode:
## Fixes

Fixes #4208 by @AetherUnbound 

## Description
Automatically convert unformatted urls in docstrings to hyperlinks using the `linkify` MyST extension.

## Testing Instructions

This can be observed by: 
- Removing Sphinx's existing hyperlinking syntax from any url in a DAG's docstring. This is the leading `<` and trailing `>` wrapping URLs.
- Note the affected DAG docstring
- Regenerate the documentation file(`DAG.md`) by running `just catalog/generate-docs`
- Run `just documentation/live` and navigate to (http://localhost:50230/catalog/reference/DAGs.html)[http://localhost:50230/catalog/reference/DAGs.html]

You should now still see the docstring link you had earlier removed the hyperlinking syntax from still correctly render as hyperlink due to MyST's `linkify` extension.
P.S: This also applies for Airflow docs.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/Pipfile
+++ b/documentation/Pipfile
@@ -10,7 +10,7 @@ sphinx-copybutton = "~=0.5"
 sphinx-reredirects = "~=0.1"
 sphinx-notfound-page = "~=1.0"
 furo = ">=2023.8.19"
-myst-parser = "~=2.0"
+myst-parser = {version = "~=2.0", extras = ["linkify"]}
 sphinxcontrib-mermaid = "~=0.9"
 
 [requires]
--- a/documentation/Pipfile.lock
+++ b/documentation/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "281119330664f17cb2e7a558ce16e12aa1e4dbd656d8017ffa797b0d50490b2d"
+            "sha256": "e2c81a082ff316c0afe8188d001eef072a8ea4e405da183cec04b3bb4621a465"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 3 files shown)


sarayourfriend (approved):
LGTM! I was going to comment to add a specific version pin in Pipfile for the new dependency, but #4166 with soon change this to PDM anyway, and for the docs the fact that the build works is sufficient to convince me things are fine.

documentation/Pipfile
```diff
@@ -12,6 +12,7 @@ sphinx-notfound-page = "~=1.0"
 furo = ">=2023.8.19"
 myst-parser = "~=2.0"
```
dhruvkb, line 13: This is a clearer way to establish where `linkify-it-py` is being used imo.

```suggestion
myst-parser = {version = "~=2.0", extras = ["linkify"]}
```

This change will necessitate running `pipenv lock` again.

dhruvkb (changes requested):
LGTM! However, I have a recommendation regarding how the linkify package is being defined in the `Pipfile`.

Also noting that almost every package in `Pipfile.lock` has been updated 😬. Can't wait for #4166 to land.

documentation/Pipfile
madewithkode: Hi @dhruvkb thanks for your insight on this, would declaring `linkify` as a myst-parser dependency as suggested require me to remove the `linkify-it-py = "*"` dependency declaration from the pipfile entirely or it can remain alongside the new change?

dhruvkb: You can remove the explicit declaration from line 15. By using the `extras`, pipenv will automatically add `linkify-it-py` as a sub-dependency of `myst-parser`.

madewithkode: Thank you, the PR has been updated.

dhruvkb (approved):
LGTM, thanks for the contribution and for addressing my feedback @madewithkode!

PR #4263, opened 2024-05-03
Add structured logging for the API

### LINKED ISSUE #3432: API structured logging
## Problem

See the milestone description for the rationale: https://github.com/WordPress/openverse/milestone/20

## Description

Add `django-structlog` to the API configuration: https://github.com/jrobichaud/django-structlog. When doing this, we can also remove the request ID middleware, as it'll no longer be necessary, `django-structlog` has that built in.

We should keep the response headers added in #3368, even though `django-structlog` logs some of those, because they're still useful in Cloudflare (we can reference response headers in WAF rules, for example).

For ease of reading in local development, add a switch to default to flat line style logs when `ENVIRONMENT == "local"`.

dhruvkb:
## Fixes

Fixes #3432 by @sarayourfriend

## Description

This PR
- adds structured logging support for Django using 
- refactors the logging setup to remove cruft and dead code
- removes `django-log-request-id` as that functionality is included in `django-structlog`
- updates all loggers to use `structlog`

Please don't be daunted by the file count, most of these are `logging` &rarr; `structlog` migrations. These changes was required to make the tests pass, else I would've made them a separate PR. The main files to review are the 5 or so changed files inside `api/conf/`.

The PR is marked as 🟧 high-priority because it blocks all the work on monitoring metrics associated with content moderation. 

## Testing Instructions

0. Check out the branch and run the dev server: `just api/up`.
1. Tail the server logs: `just logs web`.
2. Access some endpoints, perform some tasks and observe these points.
   - All logs for a particular request have the ID of that request (no separate package required).
   - Each request has a `request_started` event logging the URL, HTTP method and user agent.
   - Each request also has a `request_finished` event logging the URL, HTTP method and response status code.
   - There is no need to `getChild` logger as each log entry records the logger name, file name, func name and even line where it was called.
   - Logs with only a text message become JSON like `{"event": "<message>"}`.
   - Other loggers (like those for Uvicorn and Elasticsearch) are not affected and continue to be unstructured.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (approved):
The logs now look awesome, and you can follow a request through all of the log statements.

api/api/controllers/search_controller.py
```diff
@@ -560,19 +554,19 @@ def get_sources(index):
     except ValueError:
         cache_fetch_failed = True
         sources = None
-        log.warning("Source cache fetch failed due to corruption")
+        logger.warning("Source cache fetch failed due to corruption")
     except ConnectionError:
         cache_fetch_failed = True
         sources = None
-        log.warning("Redis connect failed, cannot get cached sources.")
+        logger.warning("Redis connect failed, cannot get cached sources.")
```
sarayourfriend, line 561: I wonder if it would be nice to change these to structured logs too, maybe in a future issue. Something like `logger.warning("Redis connect failed", when="get-cached-sources")`, and so forth. Ideally these logs never matter though, and would never be meaningful to query, so probably not worth it!

api/conf/settings/logging.py
```diff
@@ -43,81 +41,92 @@ def health_check_filter(record: LogRecord) -> bool:
...
-            "format": "[{server_time}] {message}",
+        "console": {
+            "format": "[{asctime} - {name} - {lineno:>3}][{levelname}] {message}",
             "style": "{",
         },
-        "console": {
-            "format": "[%(asctime)s - %(name)s - %(lineno)3d][%(levelname)s] [%(request_id)s] %(message)s",  # noqa: E501
+        "json": {
+            "()": structlog.stdlib.ProcessorFormatter,
+            "processor": structlog.processors.JSONRenderer(),
```
sarayourfriend, line 50: Suggestion for how to handle changing the log format for local. Along with the suggestion below for these lines (and updating references to the `json` formatter to `structured`, we'd need to add:

```py
LOG_PROCESSOR = config("LOG_PROCESSOR", default="logfmt" if ENVIRONMENT == "local" else "json")
```

Which uses the right default per environment, and allows easily switching to json format locally if needed by setting `LOG_PROCESSOR=json` in the local .env file.

```suggestion
        "structured": {
            "()": structlog.stdlib.ProcessorFormatter,
            "processor": structlog.processors.JSONRenderer() if LOG_PROCESSOR == "json" else structlog.processors.LogfmtRenderer(),
```

sarayourfriend (changes requested):
Awesome! Two small blockers:

- The issue requests `For ease of reading in local development, add a switch to default to flat line style logs when ENVIRONMENT == "local".`, and indeed, the JSON logs are very hard to read with my human eyes. I've left a suggestion for implementing this, but would be happy with anything that makes it configurable, with sensible defaults per environment.
- We have a [documentation website page about logging](https://docs.openverse.org/general/logging.html), that doesn't necessarily need to be completely re-written now (some parts of the app still use that approach, until they are covered by the other structured logging issues), but at least a reference to structlog in the API for now would be good to put there in case it is referenced.

These are small blockers, but I feel they're blockers because the first is an important devex problem that the issue anticipated and requested; the second is a documentation issue that if left unaddressed would contradict the state of the code after this PR is merged.

api/api/controllers/search_controller.py
dhruvkb: Yes, we should definitely update existing logs to be more "structured" but that, like you said, will have to be a separate issue and PR.

api/conf/settings/logging.py
dhruvkb: I have updated to use the `ConsoleRenderer` during development. I feel like that'll be helpful.

Screenshot 2024-05-07 at 5 45 36 AM

sarayourfriend: Perfect!

sarayourfriend (approved):
Excellent!

PR #4272, opened 2024-05-06
Storybook test changes to prevent flakiness from `sleep`

obulat:
_Note: a large portion of the changed files are the deleted no-longer used snapshots._

## Description
This PR extracted all the storybook testing changes from #4231 to make that PR easier to review.

This PR cleans up some of the old snapshots that are no longer used.
It also adds some waiting for responses and visibility checks before taking snapshots to make sure the page is fully loaded. On main, these changes are not necessary, but in #4231 the tests are no longer passing without these preparation.

### Filter tabs
The `focused hovered` snapshots were incorrect: the filters tab was hovered but not focused.

### Footer RTL
The footer RTL snapshots were incorrect: they used an RTL layout but English text.
E.g., snapshot `frontend/test/storybook/visual-regression/v-footer.spec.ts-snapshots/footer-content-rtl-md-linux.png`


frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
```diff
@@ -8,43 +8,57 @@ const pressedUrl = (showLabel: boolean) =>
...
         )
       })
 
       test(`hovered ${buttonName}`, async ({ page }) => {
-        await page.goto(defaultUrl(showLabel))
+        await goAndWaitForSvg(page, defaultUrl(showLabel))
         await page.hover(searchTypeButtonLocator)
+
         await expectSnapshot(
           `v-search-type-button-${buttonName}-hovered `,
```
obulat, line 43:
```suggestion
           `v-search-type-button-${buttonName}-hovered`,
```

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
```diff
@@ -8,43 +8,57 @@ const pressedUrl = (showLabel: boolean) =>
...
       })
 
       test(`pressed hovered ${buttonName}`, async ({ page }) => {
-        await page.goto(pressedUrl(showLabel))
+        await goAndWaitForSvg(page, pressedUrl(showLabel))
+
         await page.hover(searchTypeButtonLocator)
+
         await expectSnapshot(
           `v-search-type-button-${buttonName}-pressed-hovered `,
```
obulat, line 63:
```suggestion
           `v-search-type-button-${buttonName}-pressed-hovered`,
```

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
```diff
@@ -8,45 +8,59 @@ const pressedUrl = (showLabel: boolean) =>
...
       })
 
       test(`hovered ${buttonName}`, async ({ page }) => {
-        await page.goto(defaultUrl(showLabel))
+        await goAndWaitForSvg(page, defaultUrl(showLabel))
         await page.hover(searchTypeButtonLocator)
+
         await expectSnapshot(
-          `v-search-type-button-${buttonName}-hovered `,
+           `v-search-type-button-${buttonName}-hovered`,
```
obulat, line 43:
```suggestion
          `v-search-type-button-${buttonName}-hovered`,
```

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
```diff
@@ -8,45 +8,59 @@ const pressedUrl = (showLabel: boolean) =>
...
 
       test(`pressed hovered ${buttonName}`, async ({ page }) => {
-        await page.goto(pressedUrl(showLabel))
+        await goAndWaitForSvg(page, pressedUrl(showLabel))
+
         await page.hover(searchTypeButtonLocator)
+
         await expectSnapshot(
-          `v-search-type-button-${buttonName}-pressed-hovered `,
+           `v-search-type-button-${buttonName}-pressed-hovered`,
```
obulat, line 63:
```suggestion
          `v-search-type-button-${buttonName}-pressed-hovered`,
```

frontend/test/storybook/visual-regression/v-filter-tab.spec.ts
```diff
@@ -11,51 +11,82 @@ test.describe.configure({ mode: "parallel" })
...
      * If the tab is focused, it will always be selected (it's not manually activated)
      */
-    test(`focused,${appliedFilterCount} filters`, async ({ page }) => {
-      await gotoWithArgs(page, { appliedFilterCount, isSelected: true })
+    test(`focused, ${appliedFilterCount} filters`, async ({ page }) => {
+      await goAndWaitForSvg(page, { appliedFilterCount, isSelected: true })
       await focusFiltersTab(page)
 
-      expect(await page.locator(wrapper).screenshot()).toMatchSnapshot(
+      await expect(page.locator(wrapper)).toHaveScreenshot(
```
sarayourfriend, line 55: `await expect(...).toHaveScreenshot()`, nice!! That makes _so much sense_, to be able to wait for the screenshot as we expect it.

Does this also auto-update screenshots with `-u`, the way `toMatchSnapshot` does?

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
```diff
@@ -8,45 +8,59 @@ const pressedUrl = (showLabel: boolean) =>
...
 const searchTypeButtonLocator = 'button[aria-haspopup="dialog"]'
+const getIcon = (page: Page) =>
+  page.locator('button[aria-haspopup="dialog"] svg').first()
 
 test.describe.configure({ mode: "parallel" })
 
+const goAndWaitForSvg = async (page: Page, url: string) => {
+  const response = page.waitForResponse(/\.svg/)
+  await page.goto(url)
+  await response
```
sarayourfriend, line 19: Can this also be expressed (perhaps more simply) as:

```suggestion
  await Promise.all([
    page.waitForResponse(/\.svg/),
    page.goto(url),
  ])
```

It isn't immediately clear why `waitForResponse` needs to be called before but awaited after `page.goto`, even though I can sort of _guess_ why it is... but it seems abstract regardless. A comment here would help. Is it just that `waitForResponse` needs to be called before `page.goto` because it starts "capturing" responses until it is awaited, or something like that?

sarayourfriend (approved):
LGTM! I left one request for clarification and/or a suggested change that might be clearer, but not a blocker. Being able to wait for a particular screenshot is :100:!

frontend/test/storybook/visual-regression/v-filter-tab.spec.ts
obulat: I thought so too when I realized that it retries the screenshots. However, when I tried it, I saw that it makes the tests run much longer (~10 seconds instead of ~2 seconds), and, more importantly, doesn't always fix the problem. The icon-only search type button was still failing.

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
obulat: This is the way that the Playwright docs suggest using `waitForResponse`. I went down the rabbit hole to find the reason for it.

Previously, the docs were using `Promise.all` as you suggest. The reasons for it are really well described in [this comment](https://github.com/playwright-community/eslint-plugin-playwright/issues/199#issuecomment-1950272527). However, they found at least 2 possible ways of incorrectly using the `waitForResponse`:
- changing the order of the calls (putting the `goto` before `waitForResponse`) would sometimes make it so that the response would come before the navigation, so this await would hang forever.
- users would put `await` before one of the calls inside the `Promise.all` call
This is why all docs [were rewritten](https://github.com/microsoft/playwright/pull/19154) to use the setup that I used in this PR.

@sarayourfriend, do you think it's better to use the idiomatic style used in Playwright docs, or replace it with `Promise.all`?

sarayourfriend: > users would put await before one of the calls inside the Promise.all call

:sweat_smile: I wish that would just raise a type error but... I guess it can't. I suppose a linter could prevent this kind of error, though.

Let's go with the idiomatic style. If it's documented that way at the library level, that's more important.

frontend/test/storybook/utils/response.ts
```diff
@@ -0,0 +1,18 @@
+import { type Page } from "@playwright/test"
+
+/**
+ * Wait for a response to a URL that matches a regex.
+ * The pattern of setting up a wait for response without `await`,
+ * and awaiting the response after navigating to the URL is used
+ * to prevent race conditions.
+ * @see https://playwright.dev/docs/api/class-page#page-wait-for-request.
```
obulat, line 8: Very detailed explanations of the race conditions: https://github.com/microsoft/playwright/issues/5470#issuecomment-1285640689
They do offer `Promise.all` for this, but later Playwright docs moved to this setup.

frontend/test/storybook/visual-regression/v-search-type-button.spec.ts
zackkrida, line 1: Changes look good here :+1:

frontend/test/storybook/visual-regression/v-header-internal.spec.ts
```diff
@@ -1,51 +1,69 @@
...
+      breakpoints.describeEachMobile(({ expectSnapshot }) => {
+        test(`mobil header with open modal`, async ({ page }) => {
+          await page.goto(pageUrl(dir))
+
+          // Ensure fonts are loaded before taking the snapshot
+          const responsePromise = page.waitForResponse(/var\.woff2/)
+          await page.locator('button[aria-haspopup="dialog"]').click()
+          await responsePromise
+          // Mouse stays over the button, so the close button is hovered.
+          // To prevent this, move the mouse away.
```
zackkrida, line 62: Great :+1:

zackkrida (approved):
It appears CI is passing now and the described changes make sense. LGTM!

PR #4276, opened 2024-05-07
Update Science museum urls

### LINKED ISSUE #4261: Some Science Museum records continue to have invalid URLs
## Description

#4105 updated the Science Museum DAG after some provider API changes. Among other things the URLs for all our records needed to be updated, as all production records were now invalid. The difference generally looks like:

```
# Correct url
f"https://coimages.sciencemuseumgroup.org.uk/{url}"

# Old, invalid url. Note the /images/
f"https://coimages.sciencemuseumgroup.org.uk/images/{url}
```

After fixing this in the DAG we were able to run the DAG fully in production, skipping only 5 batches due to an error on the provider side (see #4013). However **27,810** records out of **122,194** total production Science Museum records still have the incorrectly formatted URL.

```sql
openledger> select count(*) from image where provider='sciencemuseum' and url ilike '%/images/%'
 ;
+-------+
| count |
|-------|
| 27810 |
+-------+

openledger> select count(*) from image where provider='sciencemuseum';
+--------+
| count  |
|--------|
| 122194 |
+--------+
```

I selected the first 10 of these and manually tested them. All urls failed as they are currently formatted. Only four of them were fixed by simply removing the `/images/` part of the path. Therefore I don't think it's sufficient to just run a batched_update where we alter the paths that way.

This also leads me to fear that some of the URLs which _were_ updated by the DAG may still be invalid.

We need to investigate further to see why some aren't working, to understand if there's a better way to build the URLs, and then to update our production data.

stacimc:
## Fixes

Fixes #4261 by @stacimc 

## Description

Hopefully this should be the final change needed to get Science Museum back in shape after the API updates in #4105.

The problem is that even though we're now formatting URLs properly:
* We have about 20k records that are not being reingested (possibly because we have to skip some batches due to an error with the provider API), and whose URLs are therefore not reformatted
* Even when the URLs _are_ reformatted, some portion of those continue to 403

We don't want to ingest records with broken links. So this PR does two things:

* Updates the Science Museum DAG to discard records whose URL cannot be reached
* Adds a one-time maintenance DAG which will be used to update and validate all of our existing records, including ones that are not being picked up by ingestion

Screenshot 2024-05-09 at 12 11 36 PM

This DAG will go back and correctly format all existing urls, and then check to see if the url is reachable. Since I don't know how many records will turn out to be unreachable, I have not opted to make the DAG automatically delete these records, just record their identifiers in a new table. Once the DAG runs, we'll be able to see how many are invalid and either investigate further or manually delete them using the recorded ids.

## Testing Instructions

Run the Science Museum DAG and ensure that it still works. Let it ingest a few hundred records and then mark it as a success.

To test the maintenance DAG we will need to make sure we have at least one record which is invalid. You can do this manually by running `just catalog/pgcli` and then:

```sql
-- Update a record's URL to a url that is incorrectly formatted (contains '/images/'), AND
-- is known to 403 even when formatted correctly.
UPDATE image
SET url='https://coimages.sciencemuseumgroup.org.uk/images/206/868/large_cd0568_011_091005_2009_43_2_NEC_mobile_phone.jpg'
WHERE identifier IN (
   SELECT identifier
   FROM image
   WHERE provider='sciencemuseum'
   LIMIT 1
);
```

Then run the `update_science_museum_urls` DAG and verify that everything passes and one record id is added to the `science_museum_invalid_ids` table.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/maintenance/update_science_museum_urls.py
+++ b/catalog/dags/maintenance/update_science_museum_urls.py
@@ -1,0 +1,164 @@
+"""
+# Update Science Museum URLs
+
+One-time maintenance DAG to update Science Museum records to have valid URLs. See https://github.com/WordPress/openverse/issues/4261.
+
+For each Science Museum record, this DAG:
+
+* updates the url to the new format, excluding `/images/` in the path if it exists
+* validates whether the url . If not, the record ID is added to an `invalid_science_musem_ids` table.
+
+Once complete, we can use the `invalid_science_museum_ids` to identify records to delete. They are not automatically deleted by this DAG, in order to give us an opportunity to first see how many there are.
+"""
+
+import logging
+from datetime import timedelta
+from textwrap import dedent
+
+from airflow.decorators import dag, task
+from airflow.exceptions import AirflowSkipException
+from airflow.models.abstractoperator import AbstractOperator
+from airflow.models.param import Param
+
+from common import slack
+from common.constants import DAG_DEFAULT_ARGS, POSTGRES_CONN_ID
+from common.sql import RETURN_ROW_COUNT, PostgresHook
+from common.urls import rewrite_redirected_url
+
+
+logger = logging.getLogger(__name__)
+
+
+DAG_ID = "update_science_museum_urls"
+INVALID_IDS_TABLE = "science_museum_invalid_ids"
+
+UPDATE_URLS_QUERY = dedent(
+    """
+    UPDATE image
+    SET url=REPLACE(url, '/images', '')
+    WHERE provider='sciencemuseum' AND url ILIKE '%/images/%';
+    """
+)
+CREATE_TABLE_QUERY = dedent(
+    f"""
+    CREATE TABLE IF NOT EXISTS {INVALID_IDS_TABLE}
+    (identifier uuid);
+    """
+)
+
+
+@task
+def run_sql(
+    sql: str,
+    handler: callable = None,
+    postgres_conn_id: str = POSTGRES_CONN_ID,
+    task: AbstractOperator = None,
+):
+    postgres = PostgresHook(
+        postgres_conn_id=postgres_conn_id,
+        default_statement_timeout=PostgresHook.get_execution_timeout(task),
+    )
+
+    return postgres.run(sql, handler=handler)
+
+
+@task
+def get_batches(
+    batch_size: int,
+    postgres_conn_id: str = POSTGRES_CONN_ID,
+    task: AbstractOperator = None,
+):
+    postgres = PostgresHook(
+        postgres_conn_id=postgres_conn_id,
+        default_statement_timeout=PostgresHook.get_execution_timeout(task),
+    )
+    records = postgres.get_records(
+        "SELECT identifier, url FROM image where provider='sciencemuseum';"
+    )
+
+    return [records[i : i + batch_size] for i in range(0, len(records), batch_size)]
+
+
+@task
+def process_batch(records):
+    invalid_ids = []
+    for identifier, url in records:
+        # Failed to reach URL
+        if rewrite_redirected_url(url) is None:
+            invalid_ids.append(identifier)
+
+    return invalid_ids
+
+
+@task
+def record_invalid_ids(invalid_ids):
+    # Chain together
+    ids_to_record = sum(invalid_ids, [])
+
+    if not ids_to_record:
+        raise AirflowSkipException("No invalid urls found!")
+
+    values = ", ".join([f"('{id}')" for id in ids_to_record])
+    query = dedent(
+        f"""
+        INSERT INTO {INVALID_IDS_TABLE} (identifier)
+        VALUES {values}
+        """
+    )
+
+    return run_sql.function(sql=query, handler=RETURN_ROW_COUNT)
+
+
+@task
+def notify_slack(invalid_count: int):
+    slack.send_message(
+        f"Detected {invalid_count} invalid Science Museum urls.", dag_id=DAG_ID
+    )
+
+
+@dag(
+    dag_id=DAG_ID,
+    schedule=None,
+    catchup=False,
+    tags=["maintenance"],
+    doc_md=__doc__,
+    default_args={
+        **DAG_DEFAULT_ARGS,
+        "retries": 0,
+        "execution_timeout": timedelta(days=2),
+    },
+    render_template_as_native_obj=True,
+    params={
+        "batch_size": Param(
+            default=1_000,
+            type="integer",
+            description="The number of records to update per batch.",
+        ),
+    },
+)
+def update_science_museum_urls():
+    # Update all URLs to have the correct format
+    update = run_sql.override(task_id="update_url_format")(sql=UPDATE_URLS_QUERY)
+
+    # Create table to track ids of records that still have an
+    # invalid/unreachable url
+    create_table = run_sql.override(task_id="create_invalid_id_table")(
+        sql=CREATE_TABLE_QUERY
+    )
+
+    # Split records into batches
+    batches = get_batches(batch_size="{{ params.batch_size }}")
+
+    # Verify each record's url
+    process = process_batch.expand(records=batches)
+
+    # Record the identifiers of records with invalid urls
+    record = record_invalid_ids(process)
+
+    # Report the number of invalid records to Slack
+    notify = notify_slack(record)
+
+    update >> create_table >> process >> record >> notify
+
+
+update_science_museum_urls()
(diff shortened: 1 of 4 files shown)


sarayourfriend:
I'm unassigning myself from review of this. I think I only got pinged because of the docs change, but it's just catalog, so left Krystle and Madison in to reflect that. If that doesn't work for y'all for any reason, let me know and I can review this PR.

krysal (approved):
LGTM! The reasoning behind the changes makes sense.

catalog/dags/maintenance/update_science_museum_urls.py
```diff
@@ -0,0 +1,164 @@
...
+        if rewrite_redirected_url(url) is None:
+            invalid_ids.append(identifier)
+
+    return invalid_ids
+
+
+@task
+def record_invalid_ids(invalid_ids):
+    # Chain together
+    ids_to_record = sum(invalid_ids, [])
```
AetherUnbound, line 96: Nit: This tripped me up a little bit, as I wasn't quite sure what was happening here! I looked into it, and apparently for larger lists, `itertools.chain` is the more appropriate way to go: https://stackoverflow.com/a/41772165

catalog/dags/maintenance/update_science_museum_urls.py
```diff
@@ -0,0 +1,164 @@
...
+    slack.send_message(
+        f"Detected {invalid_count} invalid Science Museum urls.", dag_id=DAG_ID
+    )
+
+
+@dag(
+    dag_id=DAG_ID,
+    schedule=None,
+    catchup=False,
+    tags=["maintenance"],
```
AetherUnbound, line 123: `maintenance` feels like an odd tasks for a data consistency issue, given the other DAGs we've tagged with it... 🤔 The `add_license_url` DAG has the `data_normalization` tag, maybe we should use that? Or `data_consistency`?

catalog/dags/maintenance/update_science_museum_urls.py
```diff
@@ -0,0 +1,164 @@
...
+    # Verify each record's url
+    process = process_batch.expand(records=batches)
+
+    # Record the identifiers of records with invalid urls
+    record = record_invalid_ids(process)
+
+    # Report the number of invalid records to Slack
+    notify = notify_slack(record)
+
+    update >> create_table >> process >> record >> notify
```
AetherUnbound, line 161: FWIW the last few steps here should be redundant, since the output of the previous step is used as input for the next one, creating an implicit relationship:
```suggestion
    update >> create_table >> process
```

catalog/dags/providers/provider_api_scripts/science_museum.py
```diff
@@ -166,9 +167,13 @@ def _get_creator_info(attributes):
...
-        if url.startswith("http"):
-            return url
-        return f"https://coimages.sciencemuseumgroup.org.uk/{url}"
+
+        # Will return None if url 403s
+        return rewrite_redirected_url(
+            url
+            if url.startswith("http")
+            else f"https://coimages.sciencemuseumgroup.org.uk/{url}"
+        )
```
AetherUnbound, line 176: Ooo, great call! Although, this may have a side-effect as denoted below.

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
```diff
@@ -166,7 +166,7 @@ def test_get_record_data_success(object_data):
     expected_image_data = {
         "foreign_identifier": "i458349",
         "foreign_landing_url": "https://collection.sciencemuseumgroup.org.uk/objects/co84905/small-votive-organ-of-unknown-type-votive-viscera",
-        "url": "https://coimages.sciencemuseumgroup.org.uk/458/349/large_a659676__0001_.jpg",
+        "url": "https://coimages.sciencemuseumgroup.org.uk/458/349/large_a659676__0001_.jpg/",
```
AetherUnbound, line 169: This adding of a trailing slash is a little concerning - I tried to access the new URL and with the trailing slash I got a 403, but without the slash I was able to view the image. We may need to adjust the logic (both in the ingestion DAG and the cleanup one) to account for this.

AetherUnbound (changes requested):
Thank you for jumping on this, and it's great that the DAG for fixing it is so simple! I have a few notes, including some concern about the URLs produced for testing. I was able to the ingestion DAG just fine locally, along with the cleanup DAG (both with records to clean up and without).

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
stacimc: It does not add a slash; this is done by our mocks for the [tests](https://github.com/WordPress/openverse/blob/main/catalog/tests/dags/conftest.py#L146) but does not appear in our data. Good eye, though :) I can add a comment at least or mock the entire `rewrite_redirected_url`.

catalog/dags/providers/provider_api_scripts/science_museum.py
stacimc: Responded below!

catalog/dags/maintenance/update_science_museum_urls.py
stacimc: I know, I just much prefer being explicit if it's not too gnarly to do so! I can remove if we'd prefer to be concise.

catalog/dags/maintenance/update_science_museum_urls.py
stacimc: I actually changed it to `maintenance` from `data_normalization` because I wasn't sure if it was wrong to associate it with that specific project 😆 But I think you're right, that's probably a bitter fit than `maintenance`. Works for me, will update.

catalog/dags/maintenance/update_science_museum_urls.py
```diff
@@ -0,0 +1,164 @@
...
+):
+    postgres = PostgresHook(
+        postgres_conn_id=postgres_conn_id,
+        default_statement_timeout=PostgresHook.get_execution_timeout(task),
+    )
+    records = postgres.get_records(
+        "SELECT identifier, url FROM image where provider='sciencemuseum';"
+    )
+
+    return [records[i : i + batch_size] for i in range(0, len(records), batch_size)]
```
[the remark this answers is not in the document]
stacimc: True! We could also reduce the batch size (Science Museum doesn't have that many records, something like 120k). Do you have a preference?

catalog/dags/maintenance/update_science_museum_urls.py
```diff
@@ -0,0 +1,164 @@
...
+CREATE_TABLE_QUERY = dedent(
+    f"""
+    CREATE TABLE IF NOT EXISTS {INVALID_IDS_TABLE}
+    (identifier uuid);
+    """
+)
+
+
+@task
+def run_sql(
```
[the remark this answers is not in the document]
stacimc: I did think that, and I had it in a TODO at one point. Ultimately there are some minor fiddly differences, and this is a DAG that we intend to run once and then delete so I did not think it warranted the effort to make a utility version that supports all use cases/refactor any other DAGs. I do absolutely think that's something we should do as we find more use cases, though. I'll make an issue!

catalog/dags/maintenance/update_science_museum_urls.py
stacimc: I was _hoping_ it would be quite a short list! But you're right, let's be safe 🚀

catalog/dags/maintenance/update_science_museum_urls.py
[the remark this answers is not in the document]
AetherUnbound: 120k still seems like a lot to put in XComs 😅 I think I'd def prefer the temp table approach!

catalog/dags/maintenance/update_science_museum_urls.py
AetherUnbound: That's fair, and makes sense!

catalog/tests/dags/providers/provider_api_scripts/test_science_museum.py
AetherUnbound: Whew, that's a relief!

catalog/dags/maintenance/update_science_museum_urls.py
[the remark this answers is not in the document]
stacimc: I meant it wouldn't take too long to do very small batches because the total isn't so high, not that 120k is itself not much -- particularly since this can all be deleted as soon as it's done. But I've gone ahead and put them in a temp table.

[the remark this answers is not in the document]
AetherUnbound: Ah, got it! And you're right, I hadn't even thought about potentially deleting the XComs after the run 🤦🏼‍♀️ I'm not used to the one-off DAG paradigm haha.

AetherUnbound (approved):
LGTM!! Thanks for making those changes 🚀

PR #4280, opened 2024-05-07
Expose provider in the API tags response

### LINKED ISSUE #4273: Expose provider information in the tags
## Description

Related to #431

In order to differentiate the source of various tags, both creator- and machine-generated, we plan to expose the provider information to the tags block in the API results. 

This can be done by adding a new field, `unstable__provider`, to the `TagSerializer` here (note that we'll need to set `source="provider"`, since the field in the database is `provider`):

https://github.com/WordPress/openverse/blob/3ed38fc4b138af2f6ac03fcc065ec633d6905d73/api/api/serializers/media_serializers.py#L442

We are using `unstable__provider` here so that we have the flexibility to remove or change the parameter name down the line without affecting API versioning. The parameter itself can be stabilized after this project is complete.

## Additional context

See: https://docs.openverse.org/projects/proposals/rekognition_data/20240423-implementation_plan_machine_generated_tags_api.html#expose-provider-in-tags

AetherUnbound:
## Fixes

Fixes #4273 by @AetherUnbound

## Description

This PR exposes the `provider` field within the tags attribute that is present in the database under the key `unstable__provider`. The core of this change is in `media_serializers.py`, most of the other changes in this PR revolve around making sure the field shows up in tests and documentation.

One thing that was unexpected was having to set `allow_null=True` on the serializer. It seems like there are some cases where we're converting results directly from Elasticsearch into serialized results rather than retrieving the full data from the database. We could alter the ingestion server to ingest this new field, but I don't think it's necessary (and I think we can leave this as nullable here, given the explanation I've provided).

## Testing Instructions

Tests should pass! You can also do the following:

1. `just down -v`
1. `just api/init`
1. Visit http://localhost:50280/v1/images/ and see that each of the tags arrays also sends an `unstable__provider` key
1. Visit http://localhost:50280/v1/images/?q=crim and observe a result that has tags from multiple providers

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (approved):
Nice that the change is so small! Thank you for the explanation on `allow_null`. I wonder if this is only this way in the local data, or in prod, too?

AetherUnbound:
> Thank you for the explanation on `allow_null`. I wonder if this is only this way in the local data, or in prod, too?

Because it has to do with what we index into Elasticsearch, I assume it would affect production as well. The local data is "complete" in the sense that all of the tags have a provider in the actual database. If anything, I would imagine production to be less compliant on that front!

api/api/serializers/media_serializers.py
```diff
@@ -598,6 +598,16 @@ class TagSerializer(serializers.Serializer):
...
         "tags have a null accuracy field.",
     )
+    unstable__provider = serializers.CharField(
+        label="provider",
+        source="provider",
+        # Provider is present in the database but not in Elasticsearch, so it may
+        # not always be present during serialization
+        allow_null=True,
+        help_text="The source of the tag. When this field matches the provider for the "
+        "record, the tag itself came from the upstream provider. Otherwise, the tag "
```
zackkrida, line 608: > 
```suggestion
        "record, the tag originated from the upstream provider. Otherwise, the tag "
```

Nit

zackkrida (approved):
Looks great!

PR #4283, opened 2024-05-07
Replace sample secret key for plausible with a more obviously safe value

### LINKED ISSUE #4217: Replace the sample string in a secret with a more obviously safe value
## Description

The local Plausible setup we use a (probably default) `SECRET_KEY_BASE`. This is not a problem because we don't use this key in production, but to prevent false security flagging, we should replace the value with a safe-looking one.

## Solution
Replace the current value `MN00EYOjQegVd5Z2NkMoSk3o1cklN/37YINtjYtM3yxsoXUhRNnvgWu3pgnEibKR126HmlGeHu7KHpPaYj/Awg==` with  "local_secret_key_base":

https://github.com/WordPress/openverse/blob/b55d6d521ee71b717f73ccf225b1de3bf35e5df9/docker/plausible/env.docker#L2

madewithkode:
## Fixes

Fixes #4217 by @obulat 

## Description
Replace sample secret key for plausible with a more obviously safe value.

## Testing Instructions

This change can be observed by taking a look at the value of `SCRET_KEY_BASE` environment variable in `docker/plausible/env.docker`

The local development `sha256` key hash for plausible using this new value has also been updated in `setup_plausible.sh` 

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/setup_plausible.sh
+++ b/setup_plausible.sh
@@ -8,18 +8,21 @@ docker compose exec -T "$PLAUSIBLE_SERVICE_NAME" \
   /app/bin/plausible rpc \
   'Plausible.Auth.User.new(%{name: "Deploy", email: "deploy@example.com", password: "deploy", password_confirmation: "deploy"}) |> Plausible.Repo.insert'
 
-# Create an API key with '{sites:provision:*}' scope
-# API key: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa (that's 64 times 'a')
-# Key hash: SHA256(`SECRET_KEY_BASE` + API key)
+# load necessary environment variables into shell session
+# Note: API_KEY with '{sites:provision:*}' scope and SECRET_KEY_BASE
+# are both gotten from env file.
+# shellcheck disable=SC1091
+source docker/plausible/env.docker
+
 docker compose exec -T "$PLAUSIBLE_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy -d plausible <<-EOF
 	INSERT INTO api_keys
 	  (id, user_id, name, key_prefix, key_hash, inserted_at, updated_at, scopes, hourly_request_limit)
 	VALUES
-	  (1, 1, 'Development', 'aaaaaa', '332015ffc9f0e1f475c0fadb1e1a14d2eb09774249f71961f5a2c477efc0a0fc', now(), now(), '{sites:provision:*}', 1000)
+	  (1, 1, 'Development', 'aaaaaa', encode(digest('$SECRET_KEY_BASE$API_KEY', 'sha256'), 'hex'), now(), now(), '{sites:provision:*}', 1000)
   ON CONFLICT (id) DO NOTHING
 	EOF"
 
-authorization_header="Authorization: Bearer aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+authorization_header="Authorization: Bearer $API_KEY"
 local_plausible="http://localhost:50288"
 
 # Create site using API key
--- a/docker/plausible/env.docker
+++ b/docker/plausible/env.docker
@@ -1,4 +1,5 @@
 BASE_URL=http://localhost:50288
-SECRET_KEY_BASE=MN00EYOjQegVd5Z2NkMoSk3o1cklN/37YINtjYtM3yxsoXUhRNnvgWu3pgnEibKR126HmlGeHu7KHpPaYj/Awg==
+API_KEY=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+SECRET_KEY_BASE=this_is_a_sample_secret_key_longer_than_64_bytes_this_is_a_sample_secret_key_longer_than_64_bytes
 DATABASE_URL=postgres://deploy:deploy@plausible_db:5432/plausible
 CLICKHOUSE_DATABASE_URL=http://plausible_ch:8123/plausible_ch


dhruvkb (approved):
Looks good to me, thanks for the contribution @madewithkode!

AetherUnbound (changes requested):
Ah dear, I tried running `just down -v && just up` locally, but the `plausible` container seemed like it had failed. A peek at the logs showed this:

```
plausible-1  | ERROR! Config provider Config.Reader failed with:
plausible-1  | ** (RuntimeError) SECRET_KEY_BASE must be at least 32 bytes long. See https://plausible.io/docs/self-hosting-configuration#server
plausible-1  |     /app/releases/0.0.1/runtime.exs:47: (file)
plausible-1  |     (elixir 1.14.3) src/elixir.erl:309: anonymous fn/4 in :elixir.eval_external_handler/1
plausible-1  |     (stdlib 4.2) erl_eval.erl:748: :erl_eval.do_apply/7
plausible-1  |     (stdlib 4.2) erl_eval.erl:136: :erl_eval.exprs/6
plausible-1  |     (elixir 1.14.3) src/elixir.erl:294: :elixir.eval_forms/4
plausible-1  |     (elixir 1.14.3) lib/module/parallel_checker.ex:110: Module.ParallelChecker.verify/1
plausible-1  |     (elixir 1.14.3) lib/code.ex:425: Code.validated_eval_string/3
plausible-1  |     (elixir 1.14.3) lib/config.ex:288: Config.__eval__!/3
```

It looks like we might need a much longer secret key, even if it's going to be an obviously safe value. Does that seem correct to you @dhruvkb?

madewithkode:
Hi @AetherUnbound, great find!
Since the requirement is a string >=32 bytes long, How about something like this: `this_is_a_sample_secret_key_longer_than_32_bytes`
cc @dhruvkb

AetherUnbound:
Love it! Let's do that!

madewithkode:
@AetherUnbound @dhruvkb So trying with `this_is_a_sample_secret_key_longer_than_32_bytes`  errored out with the error below, apparently the cookie signing still required secret key to be > 32 bytes. So I just resorted to repeating  the string (`this_is_a_sample_secret_key_longer_than_64_bytes_this_is_a_sample_secret_key_longer_than_64_bytes`) and everything now works fine.
`plausible-1  | Server: localhost:50288 (http)
plausible-1  | Request: GET /
plausible-1  | ** (exit) an exception was raised:
plausible-1  |     ** (ArgumentError) cookie store expects conn.secret_key_base to be at least 64 bytes
plausible-1  |         (plug 1.14.2) lib/plug/session/cookie.ex:185: Plug.Session.COOKIE.validate_secret_key_base/1
plausible-1  |         (plug 1.14.2) lib/plug/session/cookie.ex:177: Plug.Session.COOKIE.derive/3
plausible-1  |         (plug 1.14.2) lib/plug/session/cookie.ex:202: Plug.Session.COOKIE.read_raw_cookie/2
plausible-1  |         (elixir 1.14.3) lib/enum.ex:4239: Enum.find_value_list/3
plausible-1  |         (plug 1.14.2) lib/plug/session/cookie.ex:85: Plug.Session.COOKIE.get/3
plausible-1  |         (plug 1.14.2) lib/plug/session.ex:78: anonymous fn/5 in Plug.Session.fetch_session/1
plausible-1  |         (plausible 0.0.1) PlausibleWeb.Router.browser/2
plausible-1  |         (plausible 0.0.1) lib/plausible_web/router.ex:1: PlausibleWeb.Router.__pipe_through6__/1
plausible-1  | 00:10:11.182 [error] #PID<0.3613.0> running PlausibleWeb.Endpoint (connection #PID<0.3605.0>, stream id 6) terminated`

setup_plausible.sh
```diff
@@ -15,7 +15,7 @@ docker compose exec -T "$PLAUSIBLE_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy
 	INSERT INTO api_keys
 	  (id, user_id, name, key_prefix, key_hash, inserted_at, updated_at, scopes, hourly_request_limit)
 	VALUES
-	  (1, 1, 'Development', 'aaaaaa', '332015ffc9f0e1f475c0fadb1e1a14d2eb09774249f71961f5a2c477efc0a0fc', now(), now(), '{sites:provision:*}', 1000)
+	  (1, 1, 'Development', 'aaaaaa', '68e5212e2a10fb0b622250e2189f86051da4b908c6e82f343ee3260c05cf7082', now(), now(), '{sites:provision:*}', 1000)
```
sarayourfriend, line 18: It would be pretty nice if we could generate this in this script, instead of needing to maintain it by hand.

If you set `api_key` and then `source docker/plausible/env.docker`, I think you could replace the key literal here with a call to [Postgres's crypto functions](https://www.postgresql.org/docs/current/pgcrypto.html):

```suggestion
	  (1, 1, 'Development', 'aaaaaa', encode(digest('$SECRET_KEY_BASE$api_key', 'sha1'), 'hex'), now(), now(), '{sites:provision:*}', 1000)
```

Not strictly necessary for this PR, but it would make changes to `SECRET_KEY_BASE` a lot more "arbitrary" and require less work both when testing a specific value or changing the default value again (if ever needed in the future).

sarayourfriend (approved):
LGTM! Left a suggestion for a simplification in case of further changes, and to self-document how that hash is calculated (instead of needing literal documentation), but not a requirement.

setup_plausible.sh
madewithkode: Hi @sarayourfriend This makes sense, let me update.

madewithkode: This has now been updated, thank you for the  suggestion @sarayourfriend .

AetherUnbound (approved):
Awesome, glad this worked out!!

PR #4285, opened 2024-05-07
Make API image smaller by not including dev dependencies

### LINKED ISSUE #1008: Reduce size of API production image
## Problem

Development python packages are included in the final production image for the API <s>and the Ingestion Server</s>. Those aren't needed there and represent a risk; we should be able to remove them easily following a strategy similar to the Catalog with a [docker-compose.yml](https://github.com/WordPress/openverse-catalog/blob/main/docker-compose.yml) / [docker-compose.override.yml](https://github.com/WordPress/openverse-catalog/blob/main/docker-compose.override.yml) files.

## Description

Exclude `dev` packages from the **[openverse-api](https://github.com/WordPress/openverse/pkgs/container/openverse-api)** <s>and [**openverse-ingestion_server**](https://github.com/WordPress/openverse/pkgs/container/openverse-ingestion_server)</s> images.

dhruvkb:
## Fixes

Fixes #1008 by @krysal

## Description

This PR separates the API image into `api` and `api_dev`, the former of which is published to GHCR and does not include dev-dependencies, while the latter is used in CI and includes dev-dependencies.

It also updates the `set_matrix_images.py` code to enforce consistency in the outputs so that fallbacks and null-handling is not needed in `ci.yml`.

## Testing Instructions

1. See that CI builds both `api` and `api_dev`.
1. See that CI uses `api_dev` to run the tests (proven by the CI passing, the prod image does not have pytest).
1. See that CI publishes `api` (cannot test without merging the PR unfortunately).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -233,14 +233,14 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: ${{ matrix.context }}
-          target: ${{ matrix.target || '' }}
+          target: ${{ matrix.target }}
           push: false
           tags: openverse-${{ matrix.image }}
           file: ${{ matrix.file }}
           cache-from: type=gha,scope=${{ matrix.image }}
           cache-to: type=gha,scope=${{ matrix.image }}
           outputs: type=docker,dest=/tmp/${{ matrix.image }}.tar
-          build-contexts: ${{ matrix.build-contexts || '' }}
+          build-contexts: ${{ matrix.build-contexts }}
           build-args: |
             SEMANTIC_VERSION=${{ needs.get-image-tag.outputs.image_tag }}
             CATALOG_PY_VERSION=${{ steps.prepare-build-args.outputs.catalog_py_version }}
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1023,11 +1023,12 @@ jobs:
 
       - name: Load and tag image `${{ matrix.image }}` (latest & sha)
         run: |
-          docker tag openverse-${{ matrix.image }} \
+          docker image tag openverse-${{ matrix.image }} \
             ghcr.io/wordpress/openverse-${{ matrix.image }}:latest
-          docker tag openverse-${{ matrix.image }} \
+          docker image tag openverse-${{ matrix.image }} \
             ghcr.io/wordpress/openverse-${{ matrix.image }}:${{ needs.get-image-tag.outputs.image_tag }}
-          docker push --all-tags ghcr.io/wordpress/openverse-${{ matrix.image }}
+          docker image push --all-tags \
+            ghcr.io/wordpress/openverse-${{ matrix.image }}
 
   ##############
   # Deployment #
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -399,6 +399,7 @@ jobs:
     needs:
       - get-changes
       - build-images
+      - get-image-tag
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


dhruvkb:
It finally works, the tests are passing, and the diff was fairly minimal, but the difference is not as much as I had hoped.

image

I don't know if it's worth proceeding with this.

sarayourfriend:
It's a 7.5% reduction, I'd say that's pretty meaningful.

dhruvkb:
In the CI+CD workflow, we build all the images we would need in `build-images` (this includes `api` and `api_dev` now) and only publish a subset of these images in `publish-images` (this never includes `api-dev`). Images are only built once and then these artifacts are repeatedly downloaded by `load-img` where needed. We build `api_dev` only when the API needs to be tested. For all other scenarios, we only build `api`.

The issue with building a layered image is that we do not include PDM in the final image (just the `.venv/` created in `builder`) so it's not possible to install a PDM lockfile on top of the final image. Additionally installing packages in the `entrypoint` or `cmd` intuitively "feels" wrong.

dhruvkb:
If we build a dev API image in the only place it's used, we can skip the artifact upload, download and cache steps and at the cost of some DRYness, reduce the artifact volume. I'm interested in your thoughts on 2d9dca2.

.github/workflows/ci_cd.yml
```diff
@@ -417,7 +418,34 @@ jobs:
...
+
+      - name: Build API dev image
+        uses: docker/build-push-action@v5
+        with:
+          context: api
+          target: api
+          push: false
+          load: true
+          tags: openverse-api
+          cache-from: type=gha,scope=api_dev
```
sarayourfriend, line 442: I'm not 100% sure how this works, but if we cache on `scope=api`, would that allow us to reuse the layers from the existing `api` build, if it's there? :thinking: 

A small improvement if it's even possible, so no worries if it isn't or if doing that is very complex.

sarayourfriend:
It looks good to me. Am I correct in understanding that the basic difference is we wouldn't upload/load the `api_dev` image outside that particular API test job, and therefore remove the duplication of an API image loading in every other step?

sarayourfriend (approved):
I am fine with this increase in complexity. If there are additional improvements we can make towards reusing the layers through shared caching or other approaches, that's fine. Even ignoring the _size_ improvements, it is a basic security improvement not to include dependencies that aren't required for the deployed application. In that sense, this is a win to me regardless of whether it ultimately increases the volume of images/bytes-on-disk/bytes-on-wire.

.github/workflows/ci_cd.yml
dhruvkb: I was worried that layers from `api-dev` that differ from those in the `api` (especially after the difference of `--dev`/`--prod` in the PDM install step) would overwrite the cache and vice versa making both image builds slower by introducing cache misses.

Anyway, this separate cache automatically speeds up the builds across CI runs so it's still very fast.

dhruvkb:
Yes @sarayourfriend. The `api_dev` image is now only built in the API test job and not cached, uploaded as an artifact or downloaded anywhere.

The production image `api` will be built, uploaded as an artifact, downloaded for use in other jobs and also published to GHCR.

stacimc (approved):
Interesting discussion in this thread! I agree with the conclusions made here and support the change. Code LGTM. Do you think there's anywhere we should document this change?

dhruvkb:
Yes, there is some documentation about the CI + CD workflow that needs updating. I'll update the docs, push to this PR and merge it.

PR #4291, opened 2024-05-08
Display generated tags separately

### LINKED ISSUE #4192: Displaying machine-generated content
Design for #4039, as part of "Incorporate Rekognition data into the catalog" project (#431).

## Description
To display machine-generated tags in the media details page, the idea that convinces me most is separating the media info section into two tabs: Source and Generated.

Mockups of media details page

Here is a prototype of how it could work.

https://github.com/WordPress/openverse/assets/895819/c57c1fcc-6811-42b0-9713-24468ff916a0

Placing the content in a separated tab would help us with the following:
* Set a new and clear area for every machine-generated content related to the media.
* Display a larger amount of items.
* Being more transparent and precise about how content is generated.
* Introduce a new component ("Section tab", name TBD) that could be used when recommending related media based on other criteria ("by the same author", "in the same source", etc). An idea already discussed a few times.

While testing the ideas, I had two doubts that condition the design decision:
1. How many tags will be displayed? If the list is big, we can include the action "see more" already design for the tag area.
2. How does a machine-generated tag relate to tags pulled in from sources? In the prototype I included a scenario where we also need to distinguish between source and machine-generated in the tag view.

## Mockups
Figma: [Machine-generated tags](https://www.figma.com/file/BUZ0VjjzGigUuICDlhEttG/Machine-generated-tags?type=design&node-id=0-1&mode=design)

obulat:
## Fixes
Fixes #4192 by @fcoveram 

## Description
This PR adds a separate display of machine-generated tags and tags provided by sources on the single result page following the [Figma mockups](https://www.figma.com/file/BUZ0VjjzGigUuICDlhEttG/Machine-generated-tags?type=design&node-id=22-13656&mode=dev).

Based on @fcoveram's [suggestion](https://github.com/WordPress/openverse/pull/4291#issuecomment-2136714826), if the item only has source tags, the tags section doesn't have a visible heading.

~~I started to write the implementation plan for the frontend feature, but realized that this can be done in a single PR. As the project lead, do you think we need a frontend IP, @AetherUnbound?~~

This PR is blocked on #4379 because it has a link to the "About tags" page, but the page is not ready.
Another blocker is the text for this link. Currently, the link says "What is this?"; suggestions for the text are welcome.

~~The PR can be reviewed but should not be merged before #4379 is done.~~
#4379 was merged into this PR. Please also review the wording here.

## Testing instructions
Run the app using `just frontend/run dev` (running `dev:only` would not update the English strings, so you would need to run `dev` at least once)
Go to http://localhost:8443/image/a487f4eb-ce05-43e1-acae-73c4ab090cc9 to see how machine-generated and source tags are displayed.


zackkrida:
@obulat this looks great with the exception of an (existing) spacing issue that becomes more apparent when we add a text label above the tags. 

image

Observe that the "Source" and "Source Tags" text labels are not in alignment. "Source" is 0.25rem to the right of "Source Tags". 

I think we just need to remove horizontal padding from the `dd` and `dt` elements, and that this will fix the pages to match the mockups.

AetherUnbound:
> I started to write the implementation plan for the frontend feature, but realized that this can be done in a single PR. As the project lead, do you think we need a frontend IP, @AetherUnbound?

I think an IP would still be good, even just as a decision log for [some of the questions/points that were made in the project proposal](https://docs.openverse.org/projects/proposals/rekognition_data/20240320-project_proposal_rekognition_data.html#frontend). It would not have to be a big document IMO!

frontend/src/components/VMediaInfo/VCollapsibleTagSection.vue
```diff
@@ -0,0 +1,229 @@
+<template>
+  <div class="-my-1.5px">
+    <ul
+      ref="tagsContainerRef"
+      class="flex flex-wrap gap-3 overflow-y-hidden p-1.5px"
+      :class="heightClass"
+    >
+      <li v-for="tag in visibleTags" :key="tag">
+        <VTag :href="localizedTagPath(tag)">{{ tag }}</VTag>
+      </li>
```
AetherUnbound, line 10: Since we have `accuracy` and `provider` information available on the tags, one of the thoughts I had when thinking about how these results would manifest in the frontend was how we might be able to convey that information too. I [mentioned it in the initial design issue](https://github.com/WordPress/openverse/issues/4192#issuecomment-2077689928), but I'm wondering - could the individual tags (at least for the generated section) maybe also have `Source: <provider>, Accuracy: <accuracy>%` tooltips available on them?

AetherUnbound (commented):
This looks good when I run it locally! One question about some other information we have available for the generated tags.

frontend/src/components/VMediaInfo/VCollapsibleTagSection.vue
zackkrida: I like the idea of a tooltip. That should probably be decided on in the IP though 🤔

This PR is simple enough as is, and improves the _current_ frontend (we disclose the clarifai tags) that maybe we should get it merged as is? Alternatively, we draft it and wait for the IP.

AetherUnbound: I think drafting and waiting for the IP is the best move IMO!

fcoveram:
To @zackkrida's point

> Observe that the "Source" and "Source Tags" text labels are not in alignment. "Source" is 0.25rem to the right of "Source Tags".

If I recall correctly, we intentionally added this spacing to avoid cropping links' focus ring. In [this](https://private-user-images.githubusercontent.com/6351754/328933428-4cf544c7-e920-466c-978c-8fffe32c8a1e.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTYyNzcxNDIsIm5iZiI6MTcxNjI3Njg0MiwicGF0aCI6Ii82MzUxNzU0LzMyODkzMzQyOC00Y2Y1NDRjNy1lOTIwLTQ2NmMtOTc4Yy04ZmZmZTMyYzhhMWUucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUyMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MjFUMDczNDAyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MjczNGE1ZWYwNTM2MWJjZmU0MWJkMzdiOTcyNmJhOWYyYzVhNGMwYjE1MGU3ODAwNGFiZjcxNGFkODE2ZmMyNCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.-OPkvq8FexnvM1kEi7avl4jBxbNFYAb9_r5LQ1gL_QY) case, the `flickr` label.

We can add the same spacing in this section to make it look align. Or perhaps, and correct me if I'm wrong, investigate how to fix this site-wise issue and stop us from adding empty spacings that could break the layout.

frontend/src/components/VMediaInfo/VCollapsibleTagSection.vue
fcoveram: I don't fully understand the idea. Would you mind sharing an example or making a quick draw?

AetherUnbound: Forgive the poor design quality, but this is the sort of thing I had in mind when I wrote this comment: 

tag_tooltip

I imagine there might be some way to include this information in an accessible way, like perhaps as part of the item's label? But this felt like a good way to still surface that information without visual overload.

fcoveram: Very clear mockup. Thanks @AetherUnbound

obulat:
> We can add the same spacing in this section to make it look align. Or perhaps, and correct me if I'm wrong, investigate how to fix this site-wise issue and stop us from adding empty spacings that could break the layout.

I added a negative margin to the "Source" or "Provider" (whichever comes first) to fix this. You can see in the snapshots that they are aligned now.

Now, I looked at the snapshots again, and I feel like the tags headings are fine when there are both source and generated tags. However, when there are only source tags, the "Source tags" heading is not very clear. Why are they called "source" tags? And since the "What it is?" link is only available on the Generated tags, it is not possible to navigate to the explanation page. What do you think, @fcoveram?

fcoveram:
I think we can safely remove the "Source tags" heading when there are no machine-generated tags. The pill set in the media details area is common content in stock and other similar services and does not need further explanation.

obulat:
> I think we can safely remove the "Source tags" heading when there are no machine-generated tags. The pill set in the media details area is common content in stock and other similar services and does not need further explanation.

Updated in [15cb517](https://github.com/WordPress/openverse/pull/4291/commits/15cb517500405c48c170b0bedb03a663d242f9b6). I also added a screen-reader-only "Tags" heading.

zackkrida:
> Currently, the link says "What is this?"; suggestions for the text are welcome.

@obulat I think the classic "Learn more" would be appropriate here.

frontend/src/components/VMediaInfo/VCollapsibleTagSection.vue
zackkrida: @fcoveram could you turn this idea into a new issue with a refined mockup?

zackkrida (approved):
LGTM. I'll write the copy for the page next week. I made sure to test that collapsible tags are still working properly; no issues there.

frontend/src/locales/scripts/en.json5
```diff
@@ -820,6 +826,13 @@
     audioInfo: "Audio track information",
     tags: {
       title: "Tags",
+      generated: {
+        heading: "Generated tags",
+        pageTitle: "What is this?",
```
obulat, line 831:
```suggestion
        pageTitle: "Learn more",
```

zackkrida:
Hey @fcoveram could you look at this again and let us know what you think? Thanks :smile:

fcoveram:
I will take a look at this tomorrow morning 🙏 sorry for the delay

fcoveram (approved):
All looks good 🚀 🌟

PR #4295, opened 2024-05-09
Initialize feature flag state on the server

### LINKED ISSUE #4223: Initialize the feature flag store state on server request
## Problem

Currently, we check the `DEPLOYMENT_ENV` value every time we access the feature flag state.
This is probably unnecessary because the env variable will not be changed, or the changed variable will not be read by the Nuxt app, during the run of the app.

## Description

We should initialize the flag status (enabled/disabled/switchable) and the default state when initializing the feature flag store. This will make the flag state (on/off) lookup much easier and faster.

## Alternatives

If the state of the feature flags was kept somewhere online, we could update a flag status/state without having to deploy the app. Then we would need to fetch this file every time we call the feature flag status/state getter, and would need to re-evaluate every time, even on the client.

obulat:
## Fixes
Fixes #4223 by @obulat

## Description

This PR sets the flag states on the server init based on the `DEPLOYMENT_ENV` env variable, instead of checking the state every time we access the flag.

The filtering of flags by switchable or by feature groups was moved to the store, too, to simplify the component code.

## Testing instructions

The CI checks preferences and sensitive toggle, so the CI passing is a good indicator.
Go to `/preferences` and see that the features are displayed correctly and can be toggled.

### DIFF
--- a/frontend/src/pages/preferences.vue
+++ b/frontend/src/pages/preferences.vue
@@ -8,8 +8,8 @@
     it doesn't count as a user preference.
     -->
     <div
-      v-for="(group, groupIndex) in featureData.groups"
-      :key="groupIndex"
+      v-for="group in featureGroups"
+      :key="group.title"
       class="not-prose border-b border-dark-charcoal-20 py-6 last-of-type:border-b-0"
     >
       <h2 class="label-bold mb-2">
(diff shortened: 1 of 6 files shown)


frontend/test/unit/specs/stores/search-store.spec.js
```diff
@@ -491,12 +491,12 @@ describe("Search Store", () => {
...
       async ({ searchType, nextSearchType, expectedFilterCount }) => {
-        const searchStore = useSearchStore()
-        searchStore.setSearchType(searchType)
-
         const featureFlagStore = useFeatureFlagStore()
         featureFlagStore.toggleFeature("additional_search_types", "on")
 
+        const searchStore = useSearchStore()
+        searchStore.setSearchType(searchType)
+
```
dhruvkb, line 499: Can you explain if the order is important here, and if so, can you document it in the file?

dhruvkb (approved):
Great refactor, works as expected.

frontend/src/pages/preferences.vue
```diff
@@ -8,8 +8,8 @@
     it doesn't count as a user preference.
     -->
     <div
-      v-for="(group, groupIndex) in featureData.groups"
-      :key="groupIndex"
```
sarayourfriend, line 12: Oof! Good catch to change this to a stable value :cold_sweat:

frontend/test/unit/specs/stores/feature-flag-store.spec.js
```diff
@@ -164,13 +167,13 @@ describe("Feature flag store", () => {
...
-          status: { staging: "switchable" },
-        })
-      ).toEqual(flagStatus)
+      const initialValue = process.env.DEPLOYMENT_ENV
+      process.env.DEPLOYMENT_ENV = environment
+      const actualStatus = getFlagStatus({
+        status: { staging: "switchable" },
+      })
+      expect(actualStatus).toEqual(flagStatus)
+      process.env.DEPLOYMENT_ENV = initialValue
```
sarayourfriend, line 176: Modifying `process.env` needs to be in a `beforeTest` and `afterTest` hook (or whatever the Playwright equivalent is). If `expect` fails and raises, it will never be set back to its initial value.

frontend/test/unit/specs/stores/feature-flag-store.spec.js
```diff
@@ -143,16 +143,19 @@ describe("Feature flag store", () => {
...
       const featureFlagStore = useFeatureFlagStore()
-      featureFlagStore.$nuxt.$config.deploymentEnv = environment
+
       expect(featureFlagStore.featureState("feat_env_specific")).toEqual(
         featureState
       )
       expect(featureFlagStore.isOn("feat_env_specific")).toEqual(
         featureState === "on"
       )
+      process.env.DEPLOYMENT_ENV = initialValue
```
sarayourfriend, line 158: Here as well, regarding use of the before/after hooks to ensure proper cleanup in all cases.

sarayourfriend (changes requested):
LGTM, but one small blocker on the test setup/cleanup when modifying `process.env`. It's a blocker because the environment for those tests will leak into others if they themselves fail, potentially causing false-positives/negatives which would be confusing when debugging the specific failure of either of those two tests.

frontend/test/unit/specs/stores/feature-flag-store.spec.js
obulat: TIL, thank you! I'll update

obulat: Thank you for explaining the reason!

sarayourfriend (approved):
LGTM! Glad for the comprehensive tests :tada:

PR #4297, opened 2024-05-09
Set up airflow variable defaults with descriptions automatically

### LINKED ISSUE #4202: Set up Airflow Variable defaults with descriptions automatically
## Description

We have a number of Airflow Variables which act as levers for some of our DAGs, allowing us to tweak settings for certain DAGs without having to restart the stack. At present, the documentation for them is spread out and not easily accessible:

- [`SILENCED_SLACK_NOTIFICATIONS`](https://github.com/WordPress/openverse/blob/d500b7764c411f7d228ae12c57dce519c8709610/catalog/dags/common/slack.py#L246-L247)
- [`SKIPPED_INGESTION_ERRORS`](https://github.com/WordPress/openverse/blob/6636dcfbb57abca19ef32027975f78548e10411f/catalog/dags/providers/provider_api_scripts/provider_data_ingester.py#L189-L194)
- [`CONFIGURATION_OVERRIDES`](https://github.com/WordPress/openverse/blob/2cffcb9f8da6961e84a00854a3cd472fd0f9dad8/catalog/dags/providers/provider_workflows.py#L217-L220)

For this (and any other Variables that it might seem useful for), we should set up a JSON file with a default value and an adequate description for both what the shape of the expected values are and how this affects the related DAGs. This can effectively serve as documentation for the Variables above.

We can then use [`airflow variables import`](https://airflow.apache.org/docs/apache-airflow/stable/cli-and-env-variables-ref.html#import_repeat4) (with the `--action-on-existing-key` set to `skip`) as part of our [`entrypoint.sh`](https://github.com/WordPress/openverse/blob/1cf6c6ad555a0afca4ed766fe7fa9480ceb38173/catalog/entrypoint.sh) to ensure it runs both locally _and_ in production on every startup. In production, if the keys already exists, we'll just need to add the descriptions to them once since we won't be overwriting the values.

madewithkode:
## Fixes

Fixes #4202 by @AetherUnbound  

### Set up airflow variable defaults with descriptions automatically.

This change adds a `variables.tsv ` file inside `/catalog`, this file contains tab seperated values of variables, their default values and their description. 
It also adds new logic in `/catalog/entrypoint.sh` to read this file and use the contents to automatically set corresponding airflow variables. This logic is smart enough to not override existing airflow variables with any duplicates found in the file, but it goes ahead to add inexistent ones.

## Testing Instructions

This change can be observed by:
- Spin up the  `openverse_catalog` container(airflow) - `just c`
- navigate to `http://localhost:9090/variable/list/` you should be able to see all variables from `variables.tsv` alongside their descriptions now automatically set.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


madewithkode:
Hi @AetherUnbound here's a draft PR tackling this issue, kindly take a look and let me know if I'm missing something.

AetherUnbound (commented):
This is a fantastic start! I hate to say it yet again, but unfortunately this looks a little bit more complicated than I initially thought 😆 

The `entrypoint.sh` file runs for _every_ Airflow container on _every_ start. When I tried running `just down -v && just c && just logs`, all the containers failed to start because they were all trying to run `airflow variables list` and unable to read from the (not yet initialized!) database. That database initialization happens during the `exec /entrypoint "$@"` line for the `scheduler` container only, and so we get into a situation where we're trying to modify the database when it hasn't been initialized yet, and we can' initialize the database because we're trying to modify it first!

Thankfully, Airflow gives us some tools for this. What I was able to do was wrap all of the variable adding logic in the following:

```bash
if [[ "$*" == "webserver" ]]; then
  # Wait for the database to initialize, will time out if not
  airflow db check-migrations
  # ... the rest of the logic you added

fi
```

What this does is:
1. Check that we're only executing this command on the `webserver` container (by checking the command passed in, accessible via `$*`, which should be `webserver` for the webserver)
2. Wait for the scheduler container to finish the migrations (by running `airflow db check-migrations`, which will wait to 60s for migrations to be applied)

This ensures that we're only running the variable adding command on one container, and _only after_ the database has already been set up!

We also need one more change to the catalog's Dockerfile:

```diff
diff --git a/catalog/Dockerfile b/catalog/Dockerfile
index 64ad1837d..7a164fd52 100644
--- a/catalog/Dockerfile
+++ b/catalog/Dockerfile
@@ -69,5 +69,6 @@ ARG CONSTRAINTS_FILE="https://raw.githubusercontent.com/apache/airflow/constrain
 RUN pip install -r ${REQUIREMENTS_FILE} -c ${CONSTRAINTS_FILE}
 
 COPY entrypoint.sh /opt/airflow/entrypoint.sh
+COPY variables.tsv /opt/airflow/variables.tsv
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "/opt/airflow/entrypoint.sh"]
```

This will bake the `variables.tsv` file into the image, so it should also be available in our production image when we deploy (our docker mounts differ between development and production). 

With all that in place, I was able to get this working great! Let me know if you'd like any assistance making those changes 🙂

madewithkode:
Hi @AetherUnbound thanks for all the insights.
PR has been updated and I can confirm that everything works quite as expected locally. This was quite fun as it was my first time writing a bash script to solve an actual production issue. 😅

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+  existing_variables=()
+
+  # Iterate through each variable and add it to $existing_variables
+  while IFS= read -r variable; do
+    # Append the current variable to the array
+    existing_variables+=("$variable")
+  done <<<"$output"
+
+  if $found_existing_vars; then
+    echoe -e "Found the following existing variables(The values of these will not be overwritten):\n"
```
AetherUnbound, line 90:
```suggestion
    echo -e "Found the following existing variables (the values of these will not be overwritten):\n"
```

AetherUnbound (commented):
This is really close! One more change and it should be ready 😄

catalog/entrypoint.sh
madewithkode: Thanks for spotting that! Updated Now

AetherUnbound (approved):
Looks great, thanks for this awesome addition and for addressing all the feedback! 😄 🚀

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+      airflow variables set --description "$column3" "$column1" "$column2"
+    fi
+  done <"variables.tsv"
+
+  # Print the new variables list
+  new_varibles_list=$(airflow variables list -o plain | tail -n +2)
+  echo -e "The following variables are now set:\n"
+  echo "$new_varibles_list"
+
+  # if the last line in variables.tsv did not correctly terminate
```
obulat, line 124:
```suggestion
  # If the last line in variables.tsv did not correctly terminate
```

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+  # Wait for the database to initialize, will time out if not
+  airflow db check-migrations
+  # Set up Airflow Variable defaults with descriptions automatically
+  header "SETTING VARIABLE DEFAULTS"
+
+  # List all existing airflow variables, ignoring the first descriptive "key"
+  output=$(airflow variables list -o plain | tail -n +2)
+  found_existing_vars=true
+
+  # if there are no existing variable, print this notification and continue
```
obulat, line 74:
```suggestion
  # If there are no existing variables, print this notification and continue
```

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+  existing_variables=()
+
+  # Iterate through each variable and add it to $existing_variables
+  while IFS= read -r variable; do
+    # Append the current variable to the array
+    existing_variables+=("$variable")
+  done <<<"$output"
+
+  if $found_existing_vars; then
+    echo -e "Found the following existing variables(the values of these will not be overwritten):\n"
```
obulat, line 90:
```suggestion
    echo -e "Found the following existing variables (the values of these will not be overwritten):\n"
```

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+    for variable in "${existing_variables[@]}"; do
+      echo "$variable"
+    done
+  fi
+
+  # now iterate through each row of variables.tsv and and only
```
obulat, line 96:
```suggestion
  # Now, iterate through each row of variables.tsv and and only
```

catalog/entrypoint.sh
```diff
@@ -61,4 +61,74 @@ while read -r var_string; do
...
+  # run airflow variables set --description <description> <key> <value>
+  # if the key doesn't already exist in the database i.e not found in
```
obulat, line 98:
```suggestion
  # if the key doesn't already exist in the database, i.e it is not found in
```

obulat (approved):
Very nice DX improvement, @madewithkode ! Thank you for your contribution.
I've added several spelling suggestions.

Issue #3564, opened 2023-12-20
Add a terms of service page to the frontend


AetherUnbound:
## Description

We presently have terms of service defined for our API here: https://docs.openverse.org/terms_of_service.html

We recently received an external request for information about the terms of service for our frontend. While the question came in the context of scraping our data from the frontend, it would be useful for us to establish a document similar to the one for our API which specify the terms of use we're expecting there as well. Many of the same conditions could apply in the cases where it makes sense to.

## Additional context

Might be related to #2171

AetherUnbound:
It may be easiest for us to make the existing ToS more generic and then link to it from the About page rather than setting up a separate page for it (and having to manage two documents, plus translations).

sarayourfriend:
> then link to it from the About page

Might be good to check this, but ToS might not be binding if a link to it isn't prominent on every page it could pertain to... I read some stuff recently about this with respect to licencing. Definitely should check with lawyers if we can, just to be sure, but would like to avoid a case where someone has plausible deniability because they didn't check for the ToS in a specific sub-page of the site.

AetherUnbound:
Ah, that's a very good point. I'll poke around and see if I can get an answer for this!

AetherUnbound:
It seems like from a few sources[^1][^2], linking the ToS in our footer or some other location where it's visible on each page is the best bet.

[^1]: https://www.contractscounsel.com/t/us/website-terms-of-service
[^2]: https://www.termsfeed.com/blog/where-place-terms-conditions/#Where_To_Put_Terms_And_Conditions_On_Your_Website

dhruvkb:
We might also show a frontend banner that we've added the ToS and link to it (like we do for analytics and incomplete translations), but that makes a total of 3 banners for a new user which is potentially bad UX and might need @fcoveram to chime in.

sarayourfriend:
I don't think a new banner is necessary. So long as we have a ToS, if we find someone scraping we have something to point to saying why we blocked them. Even if they didn't see it, that doesn't mean the ToS isn't in effect, right? Most users will _not_ care about the ToS, so putting in front of everyone feels very heavy.

dhruvkb:
I think it's important to communicate changes to the ToS to existing users. IANAL but I would likely be against a service changing the terms of service that I originally agreed to without letting me know they had changed.

[...]

Issue #3559, opened 2023-12-20
Solicit ongoing feedback from Openverse.org users


sarayourfriend:
### Summary

Develop at least one low-effort way to receive ongoing feedback from users about Openverse’s usability, both things that work and things that don’t work.

### Background

Openverse users currently have three main methods to deliver feedback:

1. The Google form in the Openverse.org website, behind the "Feedback" link in the homepage header. This form is unmonitored and to my knowledge not a single Openverse maintainer regularly reviews the information from the form. The form's fields are also poorly structured and do not do a good job capturing the various reasons why someone could be giving feedback. The form gets a lot of spam.
2. GitHub issues. GitHub requires a separate account, potentially one completely irrelevant to a lot of Openverse users. It is also a specialist website, with terminology specific to the development of software, rather than the management of feedback about software features. We almost never get feedback about Openverse in this way other than from relatively technical folks or people already involved with the project.
3. WordPress's communication channels like the Make blog or Make chat. We sometimes receive helpful feedback here, especially from within the WordPress community, but rarely from folks outside the existing WordPress community.

We also sometimes receive email, though these are usually emails for/from integrations rather than from regular users. Some users also use the content report form to give feedback, but that's not ideal because it's not the right place for feedback to go. It is also not a well-structured form for feedback, as its goal is for reporting media on Openverse, not for feedback about Openverse itself.

Of these, only the unmonitored, spammy Google form is presented to users in any kind of meaningful or direct way as a method of feedback delivery. We never solicit feedback directly from users (for example, by prompting regular users to give feedback).

There are a variety of people who have an interest in the maintenance and improvement of Openverse. These are non-exclusive categories (individuals can be in more than one):
* Users, within which there are at least the following important subdivisions:
  * Educators
  * Students
    * Primary
    * Secondary
    * Tertiary
  * Young people generally
  * Older people
  * People with low technological literacy (people for whom using computers or the web is not “intuitive”)
  * Bloggers
  * WordPress editors
  * People in the various regions of the world
  * People who speak languages other than English
* Creators whose works are included in Openverse
* Providers/sources
* Openverse maintainers
  * Designers
  * Translators
  * Coders

Very few (if any) of our available feedback mechanisms are well suited to these categories of people. The goal of this project is to change that by adding some way of soliciting feedback directly from users of Openverse.org, following some heuristic to decide when to prompt for feedback. The project also includes the creation of the feedback repository, whether it's reusing one of the existing ones, or creating/using a new tool or approach.

[...]

PR #4302, opened 2024-05-10
Implementation Plan: Machine-generated tags on the frontend

### LINKED ISSUE #4039: Implementation Plan: Determine and design how machine-generated tags will be displayed/conveyed in the Frontend
## Description

Implementation plan for #431

Context: https://docs.openverse.org/projects/proposals/rekognition_data/20240320-project_proposal_rekognition_data.html#frontend

Related to #4038, draft an implementation plan for determining how machine-generated labels will be conveyed in the frontend. We [already have some machine-generated tags](https://github.com/WordPress/openverse/pull/3948#discussion_r1552301581) in the API, and so we will see the affects of this IP immediately (irrespective of #4038, though possibly augmented by the use of `tags.provider` in addition to `tags.accuracy`).

Similar to the API, it may also be useful to share the label accuracy with users (either visually or with extra content on mouse hover) along with its provider (for cases where we may have multiples of the same machine-generated tags from different sources). It would be beneficial to have a page much like our [sensitive content explanation](https://openverse.org/sensitive-content) (either similarly available in the frontend or on our documentation website) that describes the nature of the machine generated labels, the means by which they were determined, and how to report an insensitive label.

obulat:
## Due date:

2024-05-24

## Assigned reviewers

- [x] @AetherUnbound
- [x] @zackkrida 

Both of you were chosen for your familiarity with the project and the frontend. Input from @fcoveram is also necessary from the design point of view: whether the proposed approach is feasible and good for the UX.

## Description

Resolves #4039

## Current round

The deadline for review of this round is **2024-05-28**


documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+This component will handle filtering the tags into two sets: machine-generated
+and source-provided. It will also display the headings and the link to the
+"About tags" page. A new child component, `VCollapsibleTagSection`, will handle
+displaying of the tags in several rows, collapsing them if there are more than
+three rows of tags.
+
+Currently, we deduplicate all tags to only display unique tags. When we display
+the tags by generation type, there will probably be duplicates between the
+source and the generated tags. However, each section individually will still be
+normalized to prevent duplicates.
```
AetherUnbound, line 76: Ahh, this is a great way to handle that!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+
+The generated tags section should split the tags by different providers into
+subsections, and should use the provider of the tag as the section label. For
+instance, if there are tags by both `recognition` and `clarifai`, there will be
+two sub-sections in the Generated tags section: tags generated by `recognition`
+and tags generated by `clarifai`. In the future, if there are more providers,
+there could be more tag sections. The tags in each subsection will be ordered by
+their accuracy.
```
AetherUnbound, line 84: Great call, I really like this approach.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+flags are not necessary.
+
+### Infrastructure
+
+<!-- Describe any infrastructure that will need to be provisioned or modified. In particular, identify associated potential cost changes. -->
+
+No infrastructure changes will be necessary, we will only update the single
+result pages of the Nuxt frontend.
+
+## Alternatives
```
AetherUnbound, line 116: Thank you for including the alternatives!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+
+<!-- Describe any alternatives considered and why they were not chosen or recommended. -->
+
+The tags will be displayed in two sections: "Source tags" and "Generated tags".
+
+For the machine-generated tags, we want to add a link to the page that describes
```
AetherUnbound, line 122: It looks like this is missing a bit of information at the end.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+   the fullest information to the users. The second option provides the accuracy
+   and provider information to the user, however it is not very accessible.
+   Different browsers and screen readers treat the tooltips differently. We have
+   enough room on the page to display this data in full, using the third option:
+   labelling all the tags generated by a single provider with a "by provider"
+   heading, and adding the accuracy to the tag's inner content. This way, the
+   user will see the provider and the accuracy of the tag without any additional
+   actions. As a bonus, if a tag is generated by multiple providers, the user
+   will see that different providers generate the same tag, and will be able to
+   see the difference in accuracy between the providers.
```
AetherUnbound, line 144: Nit for formatting:
```suggestion
The first option is the simplest to implement, however, it fails to provide
the fullest information to the users. The second option provides the accuracy
and provider information to the user, however it is not very accessible.
Different browsers and screen readers treat the tooltips differently. We have
enough room on the page to display this data in full, using the third option:
labelling all the tags generated by a single provider with a "by provider"
heading, and adding the accuracy to the tag's inner content. This way, the
user will see the provider and the accuracy of the tag without any additional
actions. As a bonus, if a tag is generated by multiple providers, the user
will see that different providers generate the same tag, and will be able to
see the difference in accuracy between the providers.
```

Great points here though!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+ready:
+[Displaying machine-generated content](https://github.com/WordPress/openverse/issues/4192)
+If we decide to go with the third alternative option, we will need to add the
+provider headings and the accuracy to the tags.
+
+## Blockers
+
+<!-- What hard blockers exist that prevent further work on this project? -->
+
+There are no blockers for this work.
```
AetherUnbound, line 160: A little more context here might be helpful:
```suggestion
Since we already have machine-generated tags available in the API, there are no blockers for this work.
```

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+accessibility of the tag list. Both the sighted and the screen reader users will
+understand what the list is for. It is always desirable to have visible label
+next to a list of elements.
+
+The display of generated tags' provider and accuracy is, on the other hand, a
+challenging problem from accessibility standpoint. Tooltips are often described
+as inaccessible for mobile users who do not have a hover action. MDN's section
+on tooltip best practices suggests providing "clear labels and sufficient body
+text" instead of tooltips that hide important information. We could add the tag
+accuracy to the However, if we decide to add the provider to the generated tags,
```
AetherUnbound, line 176: Another place where it looks like this is missing info

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,199 @@
...
+important to split the text into shorter chunks to make them easier to handle
+for translators. The eslint rule will make sure that this is done properly.
+
+## Risks
+
+<!-- What risks are we taking with this solution? Are there risks that once taken can’t be undone?-->
+
+## Prior art
+
+<!-- Include links to documents and resources that you used when coming up with your solution. Credit people who have contributed to the solution that you wish to acknowledge. -->
```
AetherUnbound, line 199: I think if we don't have any information for these sections, we can remove them.

AetherUnbound (commented):
Thanks so much for codifying this all into a document! I support the suggested approach broadly, just a few clarification pieces before a decision is made here.

zackkrida:
@obulat I see there are a few places with missing text still, I can review this more deeply after those are addressed. I am also positive about the general approach outlined so far :)

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
obulat: Thank you for pointing this out. This was actually extra that should have been removed.

obulat:
> @obulat I see there are a few places with missing text still, I can review this more deeply after those are addressed. I am also positive about the general approach outlined so far :)

@zackkrida, I updated the proposal, and it's ready for review

zackkrida:
I will review this today.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -169,12 +168,18 @@ understand what the list is for. It is always desirable to have visible label
...
+who do not have a hover action. MDN's section on tooltip best practices suggests
+providing "clear labels and sufficient body text" instead of tooltips that hide
+important information.
+
+The third alternative option, where the provider is displayed once per the list
+of all the generated tags generated by it, and accuracy is displayed on the tag
+itself, is the most accessible option. The user will see (or hear) the provider
+and the accuracy of the tag without any additional actions. We need to make sure
+that the tag name and its accuracy are separated (such as with an invisible,
+`sr-only`, dot) to make the screen readers add a pause between them.
```
AetherUnbound, line 182: Great call-out!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,196 @@
...
+
+<!-- See the implementation plan guide for more information: https://github.com/WordPress/openverse/tree/19791f51c063d0979112f4b9f4eeace04c8cf5ff/docs/projects#implementation-plans-status-in-rfc -->
+<!-- This template is exhaustive and may include sections which aren't relevant to your project. Feel free to remove any sections which would not be useful to have. -->
+
+## Reviewers
+
+<!-- Choose two people at your discretion who make sense to review this based on their existing expertise. Check in to make sure folks aren't currently reviewing more than one other proposal or RFC. -->
+
+- [ ] @zackkrida
+- [ ] @AetherUnbound
```
AetherUnbound, line 13:
```suggestion
- [x] @AetherUnbound
```

AetherUnbound (approved):
Excellent, no blocking objections! This is so exciting!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
fcoveram: What is the need behind this decision? I don’t envision users needing to distinguish between providers to make a better search decision. Showing the provider assumes that users know the difference between them, like how data is processed and therefore the output.

I would go with something simpler for this release, and once we reach users when #3559 begins, inquiry how users relate to machine-generated metadata during their searches.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,196 @@
...
+There is no need to create a separate component for machine-generated tags, the
+slot in the `VTag` component will enable us to display the additional accuracy
+information.
+
+### Add "About tags" description page
+
+We should add a page to explain the users the difference between where the
+Openverse tags come from, the difference between the source and generated tags,
+and a short description of the machine-generated tag providers. This page should
+be linked from the Generated tags section.
```
fcoveram, line 95: I agree with adding content to explain tag are generated, but I disagree with adding a 10th page in the content header.

The current header is quite long, and adding an extra link in the current set doesn't provide enough information about what "tags" is. Instead, I would address the content need as proposed [here](https://github.com/WordPress/openverse/issues/3564#issuecomment-1916380597) for #3564.

The page could be very simple: reuse the current grid and page style, and add a fixed TOC on the left to jump between sections. Since the section will be linked from the search page, the information is introduced with more context about what Openverse is about and how relates to sources and content.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,196 @@
...
+fullest information to the users. The second option provides the accuracy and
+provider information to the user, however it is not very accessible. Different
+browsers and screen readers treat the tooltips differently. We have enough room
+on the page to display this data in full, using the third option: labelling all
+the tags generated by a single provider with a "by provider" heading, and adding
+the accuracy to the tag's inner content. This way, the user will see the
+provider and the accuracy of the tag without any additional actions. As a bonus,
+if a tag is generated by multiple providers, the user will see that different
+providers generate the same tag, and will be able to see the difference in
+accuracy between the providers.
```
fcoveram, line 142: I'm drawn to go with option 1.

> …it fails to provide the fullest information to the users.

This part assumes that users need this information in the media details page for a better search experience. Since we don't have users data about how they relate to machine-generated information, I would go with the simplest solution.

Providing all information is certainly necessary, but it needs to be displayed in context to not load with unnecessary information that risks confusing visitors. Describing how tags are generated in an internal page meets this need as it allows more technical description.

fcoveram:
Very clear document. I left comments related to showing how tags are generated.

In short, I agree with the content need and embrace the "giving all the details" approach, but I think we can introduce the info with more context more smoothly.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
obulat: I wrote "similar to the sensitive content page" several times, and deleted it. I guess I should have left it in.

What I meant here was we should have a page explaining what the machine-generated tags are, but not add it to the header. The sensitive media page is just like that: you can only navigate it by typing the address in the URL bar, or by opening a sensitive media item and clicking on a link from that page. Do you think this would work, @fcoveram ?

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
obulat: So, you think that showing that the tag is machine-generated or created by the source is enough, @fcoveram, and additional information for each tag would be confusing? All that we will display now on the single result page is the tag name.

fcoveram: > So, you think that showing that the tag is machine-generated or created by the source is enough, and additional information for each tag would be confusing?

Yes to both questions.

> All that we will display now on the single result page is the tag name.

Perhaps next to the "Generated tags" sentence, we can include a link saying "how does this work?" (TBD) that takes you to a "machine-generated content" section in the About page. Here is a quick mockup.

media details page. generated tags section

zackkrida: I agree, great discussion here!

obulat: I've updated #4291 to match this suggestion:

Screenshot 2024-05-20 at 7 28 45 PM

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
zackkrida: I think it'd be best not to sub-divide by provider; later, we could follow-up and display that information some other way in the future.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,196 @@
...
+two subsections in the Generated tags section: tags generated by `recognition`
+and tags generated by `clarifai`. In the future, if there are more providers,
+there could be more tag sections. The tags in each subsection will be ordered by
+their accuracy.
+
+There is no need to create a separate component for machine-generated tags, the
+slot in the `VTag` component will enable us to display the additional accuracy
+information.
+
+### Add "About tags" description page
```
zackkrida, line 90: Good :+1: at some point in the future we'll probably need to start thinking about the page hierarchy on openverse.org and what should be on docs.openverse.org or openverse.org. Also, if we thinks docs.openverse.org is appropriate for _users_ or only contributors and API consumers.

zackkrida (approved):
Agreed we should skip disambiguating providers from each other on first pass; otherwise LGTM.

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
AetherUnbound: This works! I trust @fcoveram's perspective regarding the design approach and what might be best for users here 😄 I think we could find better terminology to use there but that can be discussed in the issues/PRs!

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -0,0 +1,196 @@
...
+tag. The tag collection view will not be affected as we will not show different
+results for a machine-generated tag than the source tag.
+
+## Expected Outcomes
+
+<!-- List any succinct expected products from this implementation plan. -->
+
+When this work is completed, the single result page will display
+machine-generated tags separately from the source tags. For machine-generated
+tags, we will also display the tag provider and its accuracy.
```
fcoveram, line 38: Can someone share an example how this would look and work? I don't fully understand the idea and it sounds like a design change is necessary. Please correct me if I'm wrong.

obulat:
Drafting to update

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
```diff
@@ -117,31 +113,33 @@ result pages of the Nuxt frontend.
...
+   screen readers treat the tooltips differently.
+2. Display one section per each machine-generated tag provider, and label
+   sections with relevant heading, e.g. "Machine-Generated Tags: Rekognition".
+   In each section, list the tags in the order of their accuracy. Each tag would
+   display the tag name and the tag accuracy. This way, the user will see the
+   data without additional actions such as hovering. We have enough room on the
+   page to display this data in full. As a bonus, if a tag is generated by
+   multiple providers, the user will see that different providers generate the
+   same tag, and will be able to see the difference in accuracy between the
+   providers.
```
AetherUnbound, line 142: Similar to point 1, do you mind adding a note here on why this was _not_ an ideal choice right now?

documentation/projects/proposals/rekognition_data/20240510-implementation_plan_machine_generated_tags_frontend.md
AetherUnbound: Since we're deciding _not_ to go with this approach, do you mind changing the language (specifically `We need to make sure...`) to indicate that, _if we were to choose this approach down the line_, this is a consideration we'd need to take?

AetherUnbound (approved):
Two small notes, but no blocking objections!

zackkrida (commented):
Agreed with @AetherUnbound's suggestions. LGTM

PR #4315, opened 2024-05-10
Convert longer media `varchar` fields to `text` in the API

### LINKED ISSUE #4311: Convert longer media `varchar` fields to `text` in the API
## Description

Related #4279, see that issue for the rationale.

Alter the `Image` and `Audio` models in the API to convert the following fields from `character varying` to `text` in the underlying database:

- `foreign_identifier`
- `foreign_landing_url`
- `url`
- `thumbnail`
- `creator`
- `creator_url`
- `title`
- `audio_set_foreign_identifier` (audio only)

The resulting migration should look like this:

```sql
ALTER TABLE audio ALTER COLUMN title TYPE TEXT USING title :: TEXT;
```

Postgres uses the same underlying storage methods for both these types, so the conversion from one to the other should be extremely fast. We'll want to verify how long the migration takes in staging to ensure it will not cause any downtime.

AetherUnbound:
## Fixes

Fixes #4311 by @AetherUnbound

## Description

This PR converts the listed fields from an underlying `character varying` type to the `text` type. In some cases, URL validation was used for fields. I created a custom field type which mirrors `URLField` except that it inherits from `TextField` rather than `CharField`.

Below is the SQL migration for this change. There are several indices that are dropped, but note that these are only the `_like` indices that optimize for `varchar_pattern_ops`. Since we perform all our searching with Elasticsearch and still have the standard indices, I think it's totally fine to have them be removed (and may even speed up certain operations).

```sql
BEGIN;
--
-- Alter field audio_set_foreign_identifier on audio
--
ALTER TABLE "audio" ALTER COLUMN "audio_set_foreign_identifier" TYPE text USING "audio_set_foreign_identifier"::text;
--
-- Alter field creator on audio
--
ALTER TABLE "audio" ALTER COLUMN "creator" TYPE text USING "creator"::text;
--
-- Alter field creator_url on audio
--
ALTER TABLE "audio" ALTER COLUMN "creator_url" TYPE text USING "creator_url"::text;
--
-- Alter field foreign_identifier on audio
--
DROP INDEX IF EXISTS "audio_foreign_identifier_617f66ad_like";
ALTER TABLE "audio" ALTER COLUMN "foreign_identifier" TYPE text USING "foreign_identifier"::text;
--
-- Alter field foreign_landing_url on audio
--
ALTER TABLE "audio" ALTER COLUMN "foreign_landing_url" TYPE text USING "foreign_landing_url"::text;
--
-- Alter field thumbnail on audio
--
ALTER TABLE "audio" ALTER COLUMN "thumbnail" TYPE text USING "thumbnail"::text;
--
-- Alter field title on audio
--
ALTER TABLE "audio" ALTER COLUMN "title" TYPE text USING "title"::text;
--
-- Alter field url on audio
--
DROP INDEX IF EXISTS "audio_url_b6a832d3_like";
ALTER TABLE "audio" ALTER COLUMN "url" TYPE text USING "url"::text;
--
-- Alter field creator on audioset
--
ALTER TABLE "audioset" ALTER COLUMN "creator" TYPE text USING "creator"::text;
--
-- Alter field creator_url on audioset
--
ALTER TABLE "audioset" ALTER COLUMN "creator_url" TYPE text USING "creator_url"::text;
--
-- Alter field foreign_identifier on audioset
--
DROP INDEX IF EXISTS "audioset_foreign_identifier_ef0c8e77_like";
ALTER TABLE "audioset" ALTER COLUMN "foreign_identifier" TYPE text USING "foreign_identifier"::text;
--
-- Alter field foreign_landing_url on audioset
--
ALTER TABLE "audioset" ALTER COLUMN "foreign_landing_url" TYPE text USING "foreign_landing_url"::text;
--
-- Alter field thumbnail on audioset
--
ALTER TABLE "audioset" ALTER COLUMN "thumbnail" TYPE text USING "thumbnail"::text;
--
-- Alter field title on audioset
--
ALTER TABLE "audioset" ALTER COLUMN "title" TYPE text USING "title"::text;
--
-- Alter field url on audioset
--
DROP INDEX IF EXISTS "audioset_url_144aed53_like";
ALTER TABLE "audioset" ALTER COLUMN "url" TYPE text USING "url"::text;
--
-- Alter field creator on image
--
ALTER TABLE "image" ALTER COLUMN "creator" TYPE text USING "creator"::text;
--
-- Alter field creator_url on image
--
ALTER TABLE "image" ALTER COLUMN "creator_url" TYPE text USING "creator_url"::text;
--
-- Alter field foreign_identifier on image
--
DROP INDEX IF EXISTS "image_foreign_identifier_4c72d3ee_like";
ALTER TABLE "image" ALTER COLUMN "foreign_identifier" TYPE text USING "foreign_identifier"::text;
--
-- Alter field foreign_landing_url on image
--
ALTER TABLE "image" ALTER COLUMN "foreign_landing_url" TYPE text USING "foreign_landing_url"::text;
--
-- Alter field thumbnail on image
--
ALTER TABLE "image" ALTER COLUMN "thumbnail" TYPE text USING "thumbnail"::text;
--
-- Alter field title on image
--
ALTER TABLE "image" ALTER COLUMN "title" TYPE text USING "title"::text;
--
-- Alter field url on image
--
DROP INDEX IF EXISTS "image_url_c6aabda2_like";
ALTER TABLE "image" ALTER COLUMN "url" TYPE text USING "url"::text;
COMMIT;
```

For instance, here's the indices on the `audio` table:

```
Indexes:
    "audio_pkey" PRIMARY KEY, btree (id)
    "audio_identifier_key" UNIQUE, btree (identifier)
    "audio_url_key" UNIQUE, btree (url)
    "unique_provider_audio" UNIQUE, btree (foreign_identifier, provider)
    "audio_category_ceb7d386" btree (category)
    "audio_category_ceb7d386_like" btree (category varchar_pattern_ops)
    "audio_foreign_identifier_617f66ad" btree (foreign_identifier)
    "audio_foreign_identifier_617f66ad_like" btree (foreign_identifier varchar_pattern_ops)
    "audio_genres_e34cc474" btree (genres)
    "audio_last_synced_with_source_94c4a383" btree (last_synced_with_source)
    "audio_provider_8fe1eb54" btree (provider)
    "audio_provider_8fe1eb54_like" btree (provider varchar_pattern_ops)
    "audio_source_e9ccc813" btree (source)
    "audio_source_e9ccc813_like" btree (source varchar_pattern_ops)
    "audio_url_b6a832d3_like" btree (url varchar_pattern_ops)
```

Only the following indices would be removed:
- `audio_foreign_identifier_617f66ad_like`
- `audio_url_b6a832d3_like`

Each of these has a standard `btree` index which would remain after the operation.

## Testing Instructions

CI should pass, since this is essentially a semantic underlying database change.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/models/fields.py
```diff
@@ -0,0 +1,21 @@
+from django import forms
+from django.core import validators
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class URLTextField(models.TextField):
+    """URL field which uses the underlying Postgres TEXT column type."""
+
+    default_validators = [validators.URLValidator()]
```
sarayourfriend, line 10: _(Up front context) I know we currently use a URL field for some of these (though this would be a new detail for `creator_url`, I think), so this is only relevant feedback for this PR due to the need to add new code to continue supporting it. Just wanting to clarify up front I know this isn't a decision you're making about the model, and that my intention is to ask whether we need this at all, and if not, then to give feedback that this code is unnecessary and can be removed._

---

This validator is the only thing that differentiates the URL field from a regular text field. I question whether we need this. Not blocking because it's a trivial thing to remove in the future or as a fast-follow if we like (would be a no-SQL migration). I'd prefer we didn't add this code, however. These write-time Django validators are irrelevant for our domain and usage of the ORM because we never write to these tables with the Django ORM (except in tests). Data validation either has to happen in the catalogue, data refresh, or not at all, but definitely not in Django write-time validators.

The change requested would be to use a `TextField` and forego write-time validators as a code-quality improvement and clarification of the intention of these models and the domain they're actually concerned with (i.e. _reading_).

sarayourfriend (commented):
> Since we perform all our searching with Elasticsearch and still have the standard indices, I think it's totally fine to have them be removed (and may even speed up certain operations).

:100:. This was my thought exactly when I read those changes in the migration.

LGTM. I've left one remark regarding the URL fields. As clarified there, it can be a fast follow or a separate issue, if you prefer to avoid bike shedding the decision here. In my opinion it isn't controversial to say we don't need write-time validators on models that are never written to using Django ORM, but I also respect that could be seen as a significant change (however much I don't see it to be, given the usage of the models). All of that to say, just want to reiterate it is not a blocker.

sarayourfriend (approved):

api/api/models/fields.py
```diff
@@ -0,0 +1,21 @@
...
+    description = _("URL")
+
+    def formfield(self, **kwargs):
+        # As with CharField, this will cause URL validation to be performed
+        # twice.
```
krysal, line 15: Twice meaning it would happen in the form and at the database level?

api/api/models/fields.py
krysal: That's an interesting point. I can't think of a use case right now, but I also think it doesn't hurt to leave the validation for this kind of field. I have no strong opinion on either side.

krysal (approved):
LGTM. It's surprisingly nice how seamless the change is, and good to know the unique constraints are not affected.

api/api/models/fields.py
sarayourfriend: @AetherUnbound Just checking, were you able to create the issue?

@krysal I guess generally there is a rationale of removing code that isn't used. All code is a liability, either as a vulnerability or increased maintenance cost (the latter being most relevant and in fact exemplified here), so as much of it as we don't need is a good idea to remove.

AetherUnbound: Created in https://github.com/WordPress/openverse/issues/4320

PR #4334, opened 2024-05-15
Add 'revoked' field to ThrottledApplication to enable easily revoking access to client applications violating openverse TOS

### LINKED ISSUE #4321: Add ability to revoke access to specific Openverse API registered client applications
## Problem

Our main tool to fight ToS violations at the moment is to employ Cloudflare rules targeting particular request/response headers. This works well, but isn't necessarily intuitive, and we can do more in this regard. We have no way built into Django admin to revoke access to a client application.

## Description

~Add a new throttle scope "revoked" that simply disallows access. Either raise a 401 in the throttle itself, or probably better, augment `conf.oauth2_extensions.OAuth2Authentication` to check the scope and raise DRF's unauthorised exception when the throttle scope is "revoked".~

We'll be adding a new boolean `revoked` to the throttled application instead. See context in this comment: https://github.com/WordPress/openverse/pull/4334#issuecomment-2113303369

## Alternatives

A workaround is to set the throttled application back to unverified. The registrant would not receive a new email to re-verify, so this effectively revokes access, as unverified applications receive 401s as well. However, this is only a workaround, and is indirect (there's no way to make this clear in Django admin, nowhere near as clear as an explicit throttle scope). As such, it is not a satisfactory long-term solution.

madewithkode:
## Fixes

Fixes #4321  by @sarayourfriend 

## Description

Add 'revoked' field to `ThrottledApplication` to enable easily revoking access to client applications violating openverse TOS.

## Testing Instructions

- Create a `ThrottledApplication` on Django admin and go ahead to use the generated credentials to authenticate and make request(s) to the API. This should work as expected and return results.

- Head back to your ThrottledApplication on Django admin, edit this application by checking the 'revoked' field and save. Try making requests to the API and you should get a `401` `PermissionDenied`.

Additionally, a unit test exists for this at `api/test/integration/test_auth.py::test_revoked_application_access`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
> Also, how are we able to apply this new rate limiting rule without also explicitly including it in the throttle_classes of the relevant views/viewsets, I recall you mentioned this can somehow be done dynamically through Django admin but essentially I’d like to know how it works under the hood.

The questions you're running into are good ones, and I think expose the fact that maybe a _throttle_ isn't the right place to handle _authorisation_. The use of a throttle was a bit of a hack, and actually I'm realising that a new throttle class isn't necessary at all, because the `revoked` `rate_limit_model` would never actually allow any request _at the start of the request_ and throttle classes would never run. That authentication class runs before throttling (which makes sense, because throttling depends on authentication to identify the rate limit).

For what it's worth, if we did need the throttle class to run, it would need to be added to the `DEFAULT_THROTTLE_CLASSES` tuple here: 

https://github.com/WordPress/openverse/blob/575f529a1aed2c01efce6abd3684a4cd98dc375f/api/conf/settings/rest_framework.py#L43-L53

We don't define specific throttle classes on any routes except on a few routes that are more restrictive like thumbnails.

_However_, with your questions in mind, it really seems to me a throttle scope isn't the right choice, and a new boolean field on `ThrottledApplication`, checked in the same authentication method we were checking the rate limit model, would be clearer. There wouldn't be these dangling questions of "how does this throttle class get used" for a throttle class that indeed, never would.

> I kept running into some weird recursion issues whenever I added the rate-limit-checking condition in conf/oauth2_extensions.py/OAuth2Authentication/authenticate() not sure why that was happening or how to fix yet.

This makes sense, I hadn't realised it would happen, but I see now that it's unavoidable with the exact conditional check I suggested.

The problem is that `request.auth` is a property that runs the authentication schemes for the application.

https://github.com/encode/django-rest-framework/blob/HEAD/rest_framework/request.py#L247-L256

https://github.com/encode/django-rest-framework/blob/HEAD/rest_framework/request.py#L377-L394

Those are the two code places to keep in mind.

So the authentication runs the first time some code accesses the `request.auth` property. The authentication routine includes calling into `OAuth2Authentication::authenticate` to do the authentication. But then in there, I suggested we access `request.auth`... This access `request.auth` _in the middle_ (as it were) of `request.auth` already trying to resolve!

We can get around this by knowing that `authenticate` returns a tuple of `user, auth`, that each get assigned to the respective properties on `request`. So in the `authenticate` method, `request.auth is result[1]` would eventually be true. Instead of keeping `result` as an anonymous result, we can unpack the tuple knowing that the rest framework authentication API is stable. Then instead of `request.auth.application`, use `auth.application`.

```diff
diff --git a/api/conf/oauth2_extensions.py b/api/conf/oauth2_extensions.py
index 2ece1b497..4ebe41a45 100644
--- a/api/conf/oauth2_extensions.py
+++ b/api/conf/oauth2_extensions.py
@@ -11,7 +11,7 @@ class OAuth2Authentication(BaseOAuth2Authentication):
     keyword = "Bearer"
 
     def authenticate(self, request):
-        result = super().authenticate(request)
+        user, auth = super().authenticate(request)
         if getattr(request, "oauth2_error", None):
             # oauth2_error is only defined on requests that had errors
             # it will be undefined or empty for anonymous requests and
@@ -19,7 +19,11 @@ class OAuth2Authentication(BaseOAuth2Authentication):
             # `request` is mutated by `super().authenticate`
             raise AuthenticationFailed()
 
-        return result
+        if application := getattr(auth, "application", None):
+            if application.revoked:
+                raise PermissionDenied()
+
+        return user, auth
 
 
 class OAuth2OpenApiAuthenticationExtension(TokenScheme):
```

---

To summarise: your questions are excellent and expose a deep problem with my suggested implementation. Instead of a new throttle class, let's add a boolean `revoked` to `ThrottledApplication`, and check that in the `authenticate` method. I think the Django admin for `ThrottledApplication` will automatically allow making changes to `revoked`, but if not, here's where that model's admin is declared:

https://github.com/WordPress/openverse/blob/0a7d31bcff87f8d46f9378cf2ce2bda4ebc0692f/api/api/admin/__init__.py#L69-L87

Sorry for leading you astray with my implementation suggestion! Please let me know if this new approach makes sense, and thank you very much for asking these questions. I wouldn't have realised the problem with my suggested implementation otherwise.

madewithkode:
Hi @sarayourfriend, Once again, thank you for this detailed insights and explanations. Glad to hear my question helped make things more clearer. And yes, your new suggested approach is very clear and intuitive.

madewithkode:
Hi @sarayourfriend all suggested changes have been implemented and I can confirm that things work as expected locally. Would go ahead and undraft this PR.

api/conf/oauth2_extensions.py
```diff
@@ -11,15 +14,23 @@ class OAuth2Authentication(BaseOAuth2Authentication):
...
     def authenticate(self, request):
-        result = super().authenticate(request)
+        user_auth_tuple = super().authenticate(request)
         if getattr(request, "oauth2_error", None):
             # oauth2_error is only defined on requests that had errors
             # it will be undefined or empty for anonymous requests and
             # requests with valid credentials
             # `request` is mutated by `super().authenticate`
             raise AuthenticationFailed()
+        # if this is an authed request, check and
```
sarayourfriend, line 24: Very minor nit-pick for extra whitespace to help visually break the function up.

```suggestion

        # if this is an authed request, check and
```

sarayourfriend (approved):
Awesome @madewithkode! This LGTM, fantastic work getting this turned around after my initial misguidance :tada: :rock:. Thank you very much for the contribution :pray:

api/conf/oauth2_extensions.py
madewithkode: @sarayourfriend Updated!

zackkrida (approved):
LGTM, works great locally! @sarayourfriend I wanted to ask a question but feel free to merge this. Should we send a different response to the user other than

```json
{
    "detail": "You do not have permission to perform this action."
}
```

I can see this response leading to folks generating new credentials. Maybe something like

```json
{
    "detail": "Your registered application has been revoked for violating Openverse's Terms of Service. Learn more here: https://docs.openverse.org/terms_of_service.html."
}
```

sarayourfriend:
> I can see this response leading to folks generating new credentials. Maybe something like

To be fair, either response would provoke a bad actor to do so; this is really just a signal of "we're aware of your behaviour" more than anything.

I don't think we necessarily need to give the reason why the request was forbidden, and we should discuss internally what the policy would be if someone asked us directly about a specific client application.

sarayourfriend:
@madewithkode this just needs a rebase to resolve the conflict. I'll do that later during my workday today (it's 10 AM here for me) if you don't get to it first, then I'll merge afterwards.

Thanks again@

madewithkode:
@sarayourfriend rebased and pushed the changes.

sarayourfriend:
Thanks! Sorry I wasn't able to get to it yesterday in the end. Merged :rocket:

Issue #4256, opened 2024-05-02
Prettier's pre-commit plugin is no longer maintained


sarayourfriend:
## Current Situation

See the note here: https://github.com/pre-commit/mirrors-prettier?tab=readme-ov-file#archived

## Suggested Improvement

Find a new way to run prettier on pre-commit (probably locally?)

## Benefit

Not using an unmaintained piece of software

dhruvkb:
Seeing `pre-commit`'s maintainer's general comments across GitHub I don't think we'll even get an explanation beyond "prettier made some changes that breaks plugins entirely".

sarayourfriend:
Maybe worth considering a change to https://github.com/typicode/husky and https://github.com/lint-staged/lint-staged instead of pre-commit.

If we need to run prettier locally anyway, `pnpm` (and therefore node) is suddenly a hard and fast dependency (because prettier is required for documentation linting). Switching to lint-staged would require some up-front time to convert our checks, however, it would subsequently simplify our management of the tool by removing the need to install a separate tool (and manage that tool using unique means, i.e., downloading the .pyz). husky and lint-staged would just be a monorepo-level package.json dependency, and only a `pnpm install` away from being configured. That simplifies our CI scripts too, no need to manage pre-commit there either.

The best part of pre-commit are the isolated environments for each hook, and we can certainly do that ourselves if we like. But combined with a PDM dependencies configuration at the root pyproject.toml, we could just as easily run all the tools we use in the monorepo with either `pnpm run`, `docker run`, or `pdm run`. We can use `docker run` anytime we really do need an isolated environment rather than running directly in the repository.

@dhruvkb as our repository devex tooling aficionado... what do you think?

sarayourfriend:
I've unassigned myself from this because I haven't had time to directly work on it since the sprint to get `ov` working.

I still like the idea to start running prettier directly in the execution environment. We'll want to make sure it's always available in `ov`, but because we already rely on `pnpm install` to have ESLint available to run in CI linting, we'll definitely have prettier available if we add it to the root package.json.

@dhruvkb any opinions here?

dhruvkb:
I agree with your most recent comment and think that going the way of ESLint would be ideal. It'll be an extension of a practice that we've already determined to work well.

We can add Prettier and its plugins as dev dependencies in the root `package.json` and then invoke it as a `language: system` hook, similar to what we do for ESLint:

[...]

PR #4343, opened 2024-05-16
Dockerfy the Openverse development environment

### LINKED ISSUE #2068: Make linting more contributor-friendly
## Problem

Linting is one of the first steps for a new contributor. When a contributor tries to push their changes, the linting step ensures that any common pitfalls, style differences etc. get caught and reported.

## Description

Our lint setup is extremely complicated. It needs

- Node.js
  - including Node.js dependencies
- Docker
- just

This makes it harder for a new contributor with a partial or incomplete setup to be able to work on the project. It's even more problematic if these technologies aren't even directly relevant to the task at hand, for example Docker and Node.js are never used in the documentation process at all but needed just for linting!

Alternatively, some might opt not to lint at all and instead push changes and let CI (or another maintainer) lint it for them. This is not optimal.

## Solution

The solution is to make client-side linting as free of these dependencies as possible and relegate some of the more contrived linting steps (like frontend unit tests) to a separate lint step that only runs on the CI side.

Alternatively we must improve our documentation and mark all these technologies as required. This will make the inital barrier to entry but the end result will inevitably be a smoother dev workflow as these technologies are always good to know for any developer.

## Alternatives

The alternative is to document that users need not lint locally and can let CI run the linting for them. This has the drawback of losing autofix and adds the manual effort of comprehending the output (often cryptic to a newbie) and making the required changes (often the output does not specified what's needed to be done).

sarayourfriend:
## Fixes

Fixes #2068 
Fixes https://github.com/WordPress/openverse/issues/4137
Fixes https://github.com/WordPress/openverse/issues/4404

## Description

Dockerfy the Openverse development environment. This PR makes `bash`, `git`, and `docker` the only dependencies (on the host) required to work on Openverse (though there are probably some gaps in the tools installed in the container that need to be fixed for all workflows). bash and git are installed by default on almost all systems that aren't geared exclusively towards experts (who won't have trouble installing bash or git).

The PR accomplishes this by passing the docker socket through to an alpine-based container with all the development dependencies installed (or ready to be installed). That's the main thing that makes this work, because then docker-compose _et al_ are able to run _inside_ the container.

I originally tried this with alpine, and got very close, but the assumptions that exist in alpine are tedious and difficult to work with in a development environment. Alpine does things rather differently than most other Linux distributions to keep things small. This, of course, comes with a trade-off of not matching the typical experience, neither for the most common Linux distributions nor for macOS. For example, coreutils is not included in alpine linux, so many utilities (including e.g., timeout which we use) are busybox implementations or at least not the GNU implementation, and therefore suffer from slight differences. coreutils can be installed, but other things like user management and permissions are still quite different.

In the end, it was _much_ easier to get things working with an archlinux base. That produces an image that is well over a GB in size, however, but mostly because of the `pacman -Syyu` system update to get the latest versions of all packages in the base installation. You can see the real base image is only ~150 MB compressed (https://hub.docker.com/layers/library/archlinux/latest/images/sha256-85dc960fa1b01560091e6de62b09c4ad99c35cf818f6a7e2b2118a57f712bcb7?context=explore), so users of the image would only download that base image every so often, and just need to download the updated dependencies when running `docker build` (via `ov build`).

However, Arch's support for ARM comes from an entirely separate off-shoot project, rather than upstream Arch, and as such the archlinux docker image does not work on m-model macs. I switched to Fedora after Dhruv had issues, and Fedora _does_ support ARM in it's upstream, so hopefully it will work! The only thing not available in Fedora's repositories are the docker dependencies, but it's very easy to install the repositories for Docker with dnf and so that was trivial to solve. Everything else (including `just`) is there and ready to go!

If we go this route, then Openverse development setup would go like this:

1. Make sure you have bash, git and docker installed
2. Clone the repository and cd into it
3. Run `./ov init`

Then run any development related commands using the `ov` script. Make it easier to use by adding it to somewhere on your path using `ln -s /path/to/ov <somewhere>`. If `~/.local/bin` is on your path, for example, `ln -s /path/to/ov ~/.local/bin/ov` works well. Otherwise, just make sure you're in the Openverse directory when running `./ov`.

`ov just` will list all available just recipes

`ov just api/up` will start the API

`ov just p frontend test:unit` will run the frontend unit tests

`ov just p frontend dev` will run the frontend app

And so on and so forth.

`ov bash` opens up a shell in the container for you to poke around (and especially to debug things that aren't working in this setup ^_^)

## Testing Instructions

Try out the steps I've shared above. See what works and what doesn't. If there are dependencies missing, let's add them! If this is a good way to go, I'll update the documentation. I plan to use `ov` myself if this moves forward, which means it will be much more maintainable than a script none of us will probably ever run on a regular basis.

Also try adding new aliases in `.ovprofile` and see that they work. The example included by default is meant to illustrate how associative arrays should be used in bash.

We can use this in CI as well!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
I've taken a try at basic documentation for this in the docs site... but part of me feels like this would warrant a complete overhaul of the structure of that documentation. So much of it could basically be _deleted_ if this dockerfied development environment really works.

I'd be keen to hear what @dhruvkb and @AetherUnbound think about that, and @krysal who's thought a lot about documentation as well.

obulat (commented):
I started testing this PR locally, and used my Windows machine's WSL to both have a fresh install and to see what it's like. I'm going to continue trying, but I just wanted to leave a note on how long it took to run `./ov`: 13370 seconds (2 hours 40 minutes). My internet download speed was measured at 77 Mbps, so it's not the download that took so long. I wonder if WSL or specifically my WSL setup are causing this to take so long.

documentation/general/quickstart.md
```diff
@@ -6,6 +6,42 @@ This guide covers the steps to get the Openverse stack running locally on your
...
+
+The `ov` script Openverse uses to run command in our development environment can
+also bootstrap the project on your local system, as long as you have bash, git,
+and docker installed. If not,
+[refer to this section for information on how to get those installed](#TODO).
+
+Once you have bash, git, and docker, you can run:
+
+```shell
+sh -c "sh <(curl -fsSL https://raw.githubusercontent.com/WordPress/openverse/main/ov)"
```
obulat, line 26: This command causes an error on Mac:
```
sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `sh <(curl -fsSL https://raw.githubusercontent.com/WordPress/openverse/main/ov)'
```

ov
```diff
@@ -0,0 +1,141 @@
...
+
+# https://stackoverflow.com/a/1482133
+OPENVERSE_PROJECT="$(dirname "$(readlink -f -- "$0")")"
+export OPENVERSE_PROJECT
+
+dev_env="$OPENVERSE_PROJECT"/docker/dev_env
+
+is_yes() {
+  [[ ${1^^} == "Y" || ${1^^} == "YES" ]] && echo "true" || echo "false"
+}
```
obulat, line 13: bash -c "bash <(curl -fsSL https://raw.githubusercontent.com/WordPress/openverse/e9378ee25f5643124aad6cc785abe2126b6d1148/ov)" prints everything out and stops with this:
```
Would you like to bootstrap the Openverse development environment? (nY): y
/dev/fd/63: line 12: ${1^^}: bad substitution
```

sarayourfriend:
Let's ignore the bootstrapping part for now and consider it an optional additional feature to this. Instead, clone the repository, and run the ov script directly as the MVP. I'll change the instructions and implementation later.

docker/dev_env/Dockerfile
AetherUnbound, line 1: Not sure where the best place to include this would be, but it would be lovely if we could have `alias j=just` somewhere 😄

AetherUnbound (approved):
This is fantastic! I just have a few small requests/requests for clarification, but this already works great on my machine. I haven't tried to hook PyCharm up to use it as a development environment, though I was able to start the stack and run some DAGs 🚀 

I am getting the following whenever I run `ov`, it'd probably be best to suppress these:

```
$ ov just down
INFO: PDM 2.15.1 is installed, while 2.15.3 is available.
Please run `pipx upgrade pdm` to upgrade.
Run `pdm config check_update false` to disable the check.
just dc down 
...
```

docker/dev_env/Dockerfile
sarayourfriend: Yes, sure, just add it to `entrypoint.sh`!

We could also add a line to source an optional `.bashrc`/`.profile` type file at the root of the repository (which would be git ignored), so it's easy to add whatever aliases an individual wanted available in the `ov` environment without needing to necessarily add it to everyone's environment and without waiting for a PR.

docker/dev_env/run.sh
```diff
@@ -0,0 +1,57 @@
...
+  run_args+=(
+    --env PNPM_HOME="$pnpm_home"
+    -v "$pnpm_home:$pnpm_home:rw,z"
+  )
+fi
+
+# Share the PDM cache with the container, if it's available locally
+# --quiet so PDM doesn't repeatedly fill the console with update messages
+# if they're enabled
+if [ "$(pdm --quiet config install.cache)" == "True" ]; then
```
sarayourfriend, line 49: This comment and the use of `--quiet` in both PDM evocations are meant to address those PDM update messages.

Can you play around and see why `--quiet` isn't doing that for you? It seemed to work for me, but maybe I missed some edge case.

docker/dev_env/Dockerfile
sarayourfriend: Okay, I've added two things here.

First, `docker/dev_env/bash_profile.sh` can hold shared, team-wide aliases like `j`. Maybe some of the aliases in the just file could be moved there (or _aliased_ there :sweat_smile:) so `ov s` would be the same as `ov just s`, and so forth.

Additionally, `.ovprofile` in the repo root (git ignored, created by `ov init`) serves to hold user-specific environment configurations so no one needs to wait on a PR to get merged for their favourite alias to be available right away.

docker/dev_env/run.sh
```diff
@@ -0,0 +1,57 @@
...
+  # and others don't get confused about where files are supposed to be
+  -v "$OPENVERSE_PROJECT:$OPENVERSE_PROJECT:rw,z"
+  --workdir "$OPENVERSE_PROJECT"
+  # Save the home directory of the container so we can reuse it each time
+  --mount "type=volume,src=openverse-dev-env,target=$opener_home"
+  # Expose the host's docker socket to the container so the container can run docker/compose etc
+  -v /var/run/docker.sock:/var/run/docker.sock
+)
+
+if [ -t 0 ]; then
```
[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/WordPress/openverse/pull/4343/commits/2b7143a0089925654011b557a6c47877183c11e8

docker/dev_env/run.sh
sarayourfriend: Also, for what it's worth, you should upgrade PDM :grin:

docker/dev_env/Dockerfile
AetherUnbound: Thanks Sara, I think that's a great approach! I tried this out though, and unfortunately the file wouldn't actually be sourced unless changed `true || source "$profile"` to just `source "$profile"`. Any other configuration would cause `ov j` to fail because it couldn't be find 😕

docker/dev_env/run.sh
AetherUnbound: @sarayourfriend It looks like the output is not suppressed even when I run the command outside `ov` in my own shell 🤔 

```bash
$ pdm --quiet config install.cache
False
INFO: PDM 2.15.1 is installed, while 2.15.3 is available.
Please run `pipx upgrade pdm` to upgrade.
Run `pdm config check_update false` to disable the check.
```

AetherUnbound:
@sarayourfriend I've added two commits to tackle each of the problems I mentioned in the comments. Please feel free to drop either one of them if the change I made breaks some other conventions or expectations!

docker/dev_env/run.sh
sarayourfriend: Ah! I'd run into this before with the `-v` flag for debugging errors. Thanks for sorting it out.

docker/dev_env/Dockerfile
sarayourfriend: Thanks for fixing that! I had a heck of a time getting shellcheck to stop complaining on that, and must have forgotten to test the last change that finally got it to quiet down. Glad it was easy to fix, and the source=/dev/null still worked. `true ||` doesn't even make sense to me now... I must have been very tired when I added that :facepalm:

AetherUnbound: It is exactly what the Stack Overflow post recommends, which is the strangest thing!! But yea, on second look I'm not sure why it would work at all 😅

justfile
```diff
@@ -62,6 +62,33 @@ install:
     just node-install
     just py-install
 
+@install-hooks:
```
dhruvkb, line 65: Please document the recipes.

docker/dev_env/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# - git: required by some python package; contributors should use git on their host
+# - make: required by some pre-commit hooks for node-gyp
+# - perl: used in linting
+# - bash: used for scripts
+# - python: language runtime
+# - python-pipx: python CLI app installer
+# - docker*: used to interact with host docker socket
+# - nodejs, npm: language runtime and dependency manager (system npm required by node-based pre-commit hooks)
+# - just: command runner
+# - n: nodejs distribution manager
```
dhruvkb, line 16: `n` is something I have not used before. I am always used `nvm` for managing Node versions. I don't have any specific reasons to favour one over the other, but just noting it down as something that differs from my personal toolchain.

documentation/general/quickstart.md
```diff
@@ -6,6 +6,63 @@ This guide covers the steps to get the Openverse stack running locally on your
...
+
+With just these three dependencies available on your computer, you can use the
+`ov` script to run the entire Openverse development environment without worrying
+about other more specific dependencies.
+
+[Follow these instructions to clone the repository](#clone-the-repository). Then
+run the `ov init` script from the repository's root directory:
+
+```bash
+ov init
```
dhruvkb, line 54: Calling `ov` needs the `./` prefix?

```suggestion
./ov init?
```

Screenshot 2024-05-28 at 1 23 49 PM

dhruvkb (commented):
Why has the `automations/python/Pipfile.lock` file changed in the PR? Was it accidental?

I got hit by this when trying to start.
```
docker: no matching manifest for linux/arm64/v8 in the manifest list entries.
```

Maybe this approach will not work on Apple Silicon M-series Macs?

dhruvkb:
I managed to build the image using the `docker` command directly instead of `ov build`...

```bash
docker build --platform=linux/amd64 -t openverse-dev-env:latest docker/dev_env
```

...but it fails in `run.sh` because `getent` is different from the Linux implementation.

```bash
❯ ./ov just install-hooks
/Users/dhruvkb/Developer/ov/openverse/docker/dev_env/run.sh: line 5: getent: command not found
```

docker/dev_env/entrypoint.sh
```diff
@@ -0,0 +1,76 @@
...
+    fi
+
+    # SC1090 cannot be disabled directly
+    # Workaround from https://github.com/koalaman/shellcheck/issues/2038#issuecomment-680513830
+    # shellcheck source=/dev/null
+    source "$profile"
+  fi
+done
+
+"$@"
```
dhruvkb, line 76: I'm basing this on the API `run.sh` entrypoint-file but I think `exec` is needed here?
```suggestion
exec "$@"
```

sarayourfriend: It isn't strictly necessary, the array syntax (`$@`) expands correctly in bash for this to work. It would be necessary if it was `$*` instead.

Fine to change though, doesn't hurt anything.

sarayourfriend:
> ...but it fails in run.sh because getent is different from the Linux implementation.

Can you figure out what the process is for macOS to find the group? It's absolutely necessary for permissions to pass the docker group. I don't have any way to test this on macOS so I'll need help from someone with a mac to debug these issues if this is going to work.

sarayourfriend:
The upstream arch project doesn't support ARM, and the Arch for ARM project has separate repositories, so it won't "swap out" easily for ARM based laptops. Drat! I'll try using Fedora...

sarayourfriend:
I've got changes working to use Fedora locally. I'll push those in a moment, and then follow up with adding code from https://stackoverflow.com/a/10910180 for getting the docker group in macOS. @dhruvkb I'll need to you test that for me though, and potentially tinker with it if it doesn't work right away.

dhruvkb:
For sure, I'm excited to review it as soon as you push changes.

sarayourfriend:
Okay, it's ready for you, @dhruvkb

docker/dev_env/entrypoint.sh
dhruvkb: I have read that `exec` is needed for `SIG*` to be passed correctly because it replaces the process with the contents of `"$@"`. It's some nuance that I don't fully understand but let's change it just for consistency, if nothing else.

sarayourfriend: Ahhh, good to know that! I'll have to consider that for other places.

I've already applied your suggestion in the latest, so we'll be good to go :slightly_smiling_face:

sarayourfriend:
@AetherUnbound I rerequested review from you because I changed the base image to Fedora from Arch. I'll update the PR description shortly. Arch doesn't support ARM, which was causing some of the problems Dhruv saw on mac.

You'll need to make sure to run `ov clean` so that you don't keep the old data volume from the arch version, it won't work otherwise! `ov clean && ov init`, and then whatever testing you'd like to do inside the environment, should all still work.

@dhruvkb you may also need `ov clean`, heads up.

docker/dev_env/Dockerfile
sarayourfriend: I don't know why, but `nvm` (and relatedly `pyenv`) have always driven me to tears of frustration with how they work... and `n` just does the thing I expect it to do every time. `nvm` for some reason _always_ wants to compile Node from scratch on my computer! `pyenv` does the same thing for some reason? I think I tried `asdf` and had a similar thing happen with Python, Node, or something else. `nvm`'s magic of wanting to auto-detect and change Node versions for you was endlessly frustrating... I guess I have had a bad experience with it :sweat_smile: Some of that might come from ignorance or may have been fixed in the many years since I've bothered with it.

I don't know, happy to use whatever here really, but in my recent experience `n` always downloads the binary distribution and "just works" (including `n install auto`, which I never figured out with nvm). If `nvm` works better than `n` for some reason, happy to switch to it. With this `ov`/docker based approach, it theoretically shouldn't matter, as long as someone gets it working right it should "just work" for everyone.

documentation/general/quickstart.md
dhruvkb: I missed the step to `ov link`. Sorry.

docker/dev_env/run.sh
```diff
@@ -0,0 +1,79 @@
...
+  --user "$UID:$host_docker_gid"
+  --network host
+  # Bind the repo to the same exact location inside the container so that pre-commit
+  # and others don't get confused about where files are supposed to be
+  -v "$OPENVERSE_PROJECT:$OPENVERSE_PROJECT:rw,z"
+  --workdir "$OPENVERSE_PROJECT"
+  # Save the home directory of the container so we can reuse it each time
+  --mount "type=volume,src=openverse-dev-env,target=$opener_home"
+  # Expose the host's docker socket to the container so the container can run docker/compose etc
+  -v /var/run/docker.sock:/var/run/docker.sock
```
dhruvkb, line 42: This is genius. Before I read this, I was thinking this would be a Docker-in-Docker type scenario but this is far simpler and so much more sensible.

ov
```diff
@@ -0,0 +1,95 @@
...
+  docker build -t openverse-dev-env:latest "$dev_env"
+  ;;
+
+clean)
+  docker volume rm openverse-dev-env
+  ;;
+
+link)
+  # Assumes XDG-like bin directory... there's probably a better way? :thinking:
+  ln -s "$_self" "${2:$HOME/.local/bin/ov}"
```
dhruvkb, line 78: Missed a hyphen, fix available in #4401.

```suggestion
  ln -s "$_self" "${2:-$HOME/.local/bin/ov}"
```

docker/dev_env/run.sh
```diff
@@ -0,0 +1,79 @@
...
+set -e
+
+# Kudos https://stackoverflow.com/a/10910180 for macOS group ID reading
+case "$OSTYPE" in
+linux*)
+  host_docker_gid="$(getent group docker | cut -d: -f3)"
+  ;;
+
+darwin*)
+  host_docker_gid="$(dscl . -read /Groups/docker | awk '($1 == "PrimaryGroupID:") { print $2 }')"
```
dhruvkb, line 12: There is no `docker` group in macOS. Most of the permission-related stuff is handled by Docker Desktop or Orbstack. #4401 takes a userless approach that works well on a Mac but can, understandably, be against what I assume were your goals for this PR.

docker/dev_env/entrypoint.sh
```diff
@@ -0,0 +1,76 @@
+#! /usr/bin/env bash
+
+set -e
+
+if [ "$UID" = "0" ]; then
+  printf "Do not run this container with the root user! Use the 'run.sh' script or pass --user to docker run.\n"
+  exit 1
+fi
```
dhruvkb, line 8: This bit had to be dropped because the Mac-compatible implementation cannot supply a reasonable `--user` argument to Docker.

dhruvkb:
Switching to Fedora solved the ARM platform issues that Arch had. Also feels good to interact with Fedora again.

I've made a kind-of follow-up PR in #4401 with all the changes I had to make to get this setup to work locally. I'll tag a couple of other Mac users to try that out but some of the changes (specially the dropping of user management) are rather drastic so I'm not entirely sure if that is the best way to add Mac compatibility.

AetherUnbound (commented):
I've tried running this locally, and the new Fedora image works great. However, it looks like the change from `"$@"` to `exec "$@"` has broken the function setups in `bash_profile.sh` - `ov j` no longer works 😢 I tried looking at a few SO posts for this (https://unix.stackexchange.com/questions/50692/executing-user-defined-function-in-a-find-exec-call, https://stackoverflow.com/questions/4321456/find-exec-a-shell-function-in-linux), but it doesn't seem like there's an easy way to get around it. In fact, I'm not sure the `ovprofile` function/aliasing will work _at all_ with `exec` 😕 

Still though, that's fine in my mind. I can create a local alias for `ovj=ov just` or something and that will work find! Not approving here yet because I think we should either try to find a workaround for that or remove the custom `.ovprofile` steps if they're not able to work at all.

sarayourfriend:
Thanks for looking into those changes @dhruvkb, I'll try it out! Hopefully it works for both host types... it seems like it should? :thinking: 

One of the problems I was running into on Linux though is that because the filesystem is bound, if the user inside the container doesn't have the same group permissions as the host user, it will write files as `root` (because that's the default user inside the Fedora container). We have this issue with our regular compose stack too, which is why we pass the user in `exec` and other places (and the places we don't do not write files).

For example, when running `pdm install` inside the container, pdm will write a `.pdm-python` file inside the project directory. If the container user is not the same UID (and preferably in a group with the same GID as the host), it will write is as root (UID:GID 0:0) and cause a permissions issue.

It's trivial to optionally pass `--user` for just linux hosts though, so it'll be easy to solve if it is a problem.

docker/dev_env/run.sh
```diff
@@ -0,0 +1,79 @@
...
+
+suppress_output() {
+  "$@" 2>/dev/null 1>/dev/null
+}
+
+opener_home="/home/opener"
+
+run_args=(
+  -i
+  --rm
```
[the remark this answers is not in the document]
sarayourfriend: The issue is if you don't --rm, then you need to variably `exec` instead of `run`. So we'd have to `docker ps` in `run.sh` to check to see if the container is already available, and if not, then use `run` instead. I wonder if that actually improves start up speed?

docker/dev_env/run.sh
sarayourfriend: I learned about it from https://distrobox.it/, which is pretty nifty (and a similar tool comes installed by default on Fedora, or at least the immutable versions)

sarayourfriend:
@AetherUnbound with the exec and alias issue, I have another idea. Instead of writing bash functions, we could use a bash associative array to map `$_cmd` to an alias, and combine that with a user-defined bash array in `.ovprofile`, then read that from `ov` itself. Basically, take advantage of the fact that we control the full entrypoint, and don't strictly need to modify the environment inside the container. And then we can configure `entrypoint` to read the same arrays and populate a .bashrc of aliases or something too, so that the same aliases are available if you do `ov bash`?

We can start with just making them available at the `ov` script level, and explore options for customising/expanding the in-container bash environment later...

docker/dev_env/Dockerfile
```diff
@@ -42,16 +45,18 @@ RUN dnf -y install dnf-plugins-core \
     gcc \
     just \
     which \
-    pipx \
-    nodejs \
+    nodejs npm \
+    python3.12 pipx \
```
dhruvkb, line 49: I had removed `npm` (as it comes automatically with `nodejs`) and `python3.12` (as it is preinstalled on the system). I don't mind having them listed here but they don't really add anything.

docker/dev_env/Dockerfile
```diff
@@ -42,16 +45,18 @@ RUN dnf -y install dnf-plugins-core \
...
     docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin \
   && pipx install \
     pdm pipenv  \
     pre-commit \
-  && npm install -g \
+    && npm install -g \
```
dhruvkb, line 54: Was this additional indentation accidental or intentional?

docker/dev_env/run.sh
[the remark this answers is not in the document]
dhruvkb: Maybe not. Let's leave it like this for now.

docker/dev_env/Dockerfile
```diff
@@ -1,21 +1,24 @@
-FROM docker.io/library/fedora:latest
+FROM docker.io/library/fedora:latest AS base-system
```
dhruvkb, line 1: Are you considering multi-stage builds for the dev-env image?

docker/dev_env/Dockerfile
sarayourfriend: I figured that was why you removed them, I wanted to keep them as explicit dependencies for clarity of intention.

docker/dev_env/Dockerfile
sarayourfriend: Unintentional, I'll fix this.

docker/dev_env/Dockerfile
sarayourfriend: I was, because I was getting frustrated with the `ENV` declarations causing the `dnf` runs to not be cached, but I'll remove this.

One thought I had was if the differences between macos and Linux hosts continue to be an issue, maybe having separate targets for each would help.

docker/dev_env/shared_aliases.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Place common shared utility functions here
+# Functions, not aliases, must be used. Aliases are not expanded in scripts
+# and as such will not be available in `entrypoint.sh` when the command
+# passed to the docker run is evaluated. That means that `j` below, for example,
+# wouldn't work if using `ov j`. However, as a function, it does work.
+
+# Personal aliases and utilities should go in an .ovprofile file at the monorepo root
+
+declare -A shared_aliases
```
dhruvkb, line 11: `declare -A` is a feature that was introduced in Bash newer than the one on macOS (version 3.2.57(1)-release). This doesn't work for me.

sarayourfriend: Just adding a note here, we tried switching to zsh, and that worked great, except that shellcheck doesn't support it.

We only use the associative bash arrays for aliases, and I think we could easily support it another way (e.g., using Python to expand aliases). I've removed the alias support for now @AetherUnbound and we can try to accommodate that in a separate PR.

sarayourfriend:
Dhruv just pointed out to me that I haven't actually addressed #4256 yet in this PR. I'll unlink that from this PR and in a separate PR we can switch to running Prettier using the system pre-commit hook instead of mirrors-prettier

sarayourfriend:
So far the follow-up issues I'll open for this are:

- Explore using `ov` in CI
- Add support for alias expansion to `dev_env/entrypoint.sh` (without using bash associative arrays)

Anything else @dhruvkb @AetherUnbound

docker/dev_env/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+    nodejs npm \
+    python3.12 pipx \
+    docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin \
+  && pipx install \
+    pdm pipenv  \
+    pre-commit \
+  && npm install -g \
+    n \
+    corepack \
+    && corepack enable
```
dhruvkb, line 56: Missed one.
```suggestion
  && corepack enable
```

docker/dev_env/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+RUN bash -c 'chmod -Rv 0777 /opt'
+
+# Avoid overwriting `.venv`'s from the host
+ENV PDM_VENV_IN_PROJECT="False"
```
dhruvkb, line 61: I just realised what does this do? Isn't this also how `node_modules` work? If overwriting is a problem, won't it be a problem for Node?

justfile
```diff
@@ -62,6 +62,10 @@ install:
     just node-install
     just py-install
 
+# Install `ov`-based git hooks
+@install-hooks:
+    bash -c "cp ./docker/dev_env/hooks/* ./.git/hooks"
```
dhruvkb, line 67: We also have to install pre-push hooks here.

ov
```diff
@@ -0,0 +1,94 @@
...
+    scratch:
+
+        ov clean && ov init
+
+  ov link [path]
+    Create a symlink from the path arg to the ov script. Path defaults
+    to the XDG-like ~/.local/bin/ov
+
+  ov hook HOOK
+    Run git hooks through pre-commit inside the development toolkit container
```
dhruvkb, line 40: Nit:
```suggestion
    Run Git hooks through pre-commit inside the development toolkit container
```

dhruvkb (changes requested):

justfile
sarayourfriend: Done in https://github.com/WordPress/openverse/pull/4343/commits/63237098756a2c6d934519e58f80a8a2a92a4eb2

docker/dev_env/run.sh
```diff
@@ -0,0 +1,74 @@
+#! /usr/bin/env bash
+
+set -e
+
+volume_name="openverse-dev-env"
+
+if ! docker volume inspect openverse-dev-env 2>/dev/null 1>/dev/null; then
```
dhruvkb, line 7: I always get confused, is `[ ... ]` not needed here? Also `&>`.

```suggestion
if ! docker volume inspect openverse-dev-env &>/dev/null; then
```

sarayourfriend: The square brackets aren't needed when testing the return code of the command, which is what we're doing here. `[` is actually an alias for `test` and `]` is just a conventional way to close it (but actually `;` does the closing if you put `then` on the same line). `[[` is a special bash extension separate from `[`.

If we used `[..]` it would run the contents through `test`, which we don't want here. Instead we just use `!` to assert whether the exit code is zero or non-zero.

sarayourfriend: I've switched to &> in https://github.com/WordPress/openverse/pull/4343/commits/59d705bac0d73642761888966109141f8b395afe

docker/dev_env/Dockerfile
dhruvkb: Explained by @sarayourfriend.

dhruvkb (approved):
Tried a bunch of things in the container, things are working perfectly!

sarayourfriend:
@AetherUnbound I've merged this based on your previous approval of the overall approach, and based on it working for Dhruv and myself as example linux and macOS hosts.

Both of us expect there to be further improvements or changes that need to be made to make the solution truly universal, but I wanted to get this particular PR merged because it's already gotten pretty complex!

This doesn't change anything about the way the repository works and the documentation isn't there yet to suggest community contributors to use it, so I think it's safe to merge this as an "alpha" version, the "MVP" of this idea, and to continue trying it out individually and making improvements as we go.

Aliases aren't working for now, but I'll open an issue to try adding them back in as a special feature!

AetherUnbound:
That sounds great, thanks for going ahead and merging it! I'm super excited about this!

PR #4349, opened 2024-05-16
Prevent exposing Django Admin features referencing media tables in prod

### LINKED ISSUE #4344: Prevent Django Admin default queries on primary media tables in production
## Description

On several occasions now, we've run into production resource issues with our database due to Django Admin performing some default query against our (massive) primary media tables (e.g. #970).

Recent changes to the reporting admin (specifically #4254) have added new views that are also selecting the primary media tables in order to generate the UI. These are incredibly useful for local testing, but are actively harmful for production. ~~We should alter the logic for these fields so that the automatic completion only occurs locally and does not occur in production. This will help prevent issues from occurring from simply visiting pages in Django Admin in production.~~

Going forward (per @sarayourfriend's comment https://github.com/WordPress/openverse/issues/4344#issuecomment-2146265927), we should create a custom object manager for the media models that ensures the following things:

- `ORDER BY` is not used on the base media tables
- A `LIMIT` is always defined for queries on the media tables

This would prevent us from erroneously adding naive queries that could impact production in the future, even if we add the functionality for the query in a naive way.

AetherUnbound:
## Fixes

Fixes #4344 by @AetherUnbound

## Description

This PR modifies the Django Admin registrations for various media report related views in order to prevent exposing access to the primary media tables in production. Queries that are made to these tables are often naive and, for such a large table, cause significant issues. However, they're extremely useful to have when running the application locally in order to easily create/reference media while we build out the content safety reporting features.

This will necessarily reduce our available features in production, but because of the nature of the failure which caused this, we couldn't have used those features anyway!

## Testing Instructions

Tests should pass. You can also verify this locally by:

1. `just api/init`
1. Visit http://localhost:50280/admin/ and observe that `Image Decision` shows up in the list of available models for site administration
1. Add `ENVIRONMENT=production` and `DJANGO_SECRET_KEY=blahblahblah` to `api/.env`
1. Run `just down && just a` to restart the API stack
1. Visit http://localhost:50280/admin/ and observe that the decision views are no longer present

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/admin/__init__.py
+++ b/api/api/admin/__init__.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import admin
 from django.contrib.auth.admin import GroupAdmin, UserAdmin
 from django.contrib.auth.models import Group, User
@@ -55,8 +56,9 @@ for klass in [
     admin.site.register(klass, MediaSubreportAdmin)
 
 # Temporary addition of model admin for decisions while this view gets built
-admin.site.register(ImageDecision, admin.ModelAdmin)
-admin.site.register(AudioDecision, admin.ModelAdmin)
+if settings.ENVIRONMENT != "production":
+    admin.site.register(ImageDecision, admin.ModelAdmin)
+    admin.site.register(AudioDecision, admin.ModelAdmin)
 
 
 @admin.register(ContentProvider)
--- a/api/api/admin/media_report.py
+++ b/api/api/admin/media_report.py
@@ -1,3 +1,5 @@
+from typing import Sequence
+
 from django.conf import settings
 from django.contrib import admin
 from django.contrib.admin.views.main import ChangeList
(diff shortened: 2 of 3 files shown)


dhruvkb (approved):
It's unfortunate that these models have to be removed from the admin but also understandable. Thanks for fixing this so quickly, LGTM!

AetherUnbound:
> It's unfortunate that these models have to be removed from the admin but also understandable. Thanks for fixing this so quickly, LGTM!

Agreed, but only for prod! Hopefully we'll have the official features for actually handling this well available soon 😄

sarayourfriend:
I don't understand why this is the solution? Is the problem not the media object identifier auto-complete query on the Django admin page? 

https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields

There's no way we need to entirely remove these pages from Django admin, that's just not how this kind of thing works. There is something about the admin configuration causing it to make the query, it isn't an _inherent feature_ of those page, but something we can change.

Was this meant as a temporary solution? I'm confused about what the long-term intention is here. @AetherUnbound can you clarify on this? This seems a fine stop-gap if we don't know what is causing it, but it's under no circumstances an acceptable long-term solution.

AetherUnbound:
This is indeed a temporary solution. The `<Media>Decision` admin views were added in a very naive way by me in #4254 as a means of creating these records locally easily. The naivete of the implementation led to, I believe, some bad queries in production when trying to render the page. In my mind, it's fine to disable those views in production until we have the actual decision-creation mechanism appropriately set up, since the current exposure of those tables was for testing anyway. My assumption was that future tasks in the [moderation basic functionality milestone](https://github.com/WordPress/openverse/milestone/22) would actually expose the necessary views there in a more appropriate way.

The problem is _also_ the media object identifier auto-complete query, which is why those were production deferred on the other models. I also deferred the search fields because those are used by the auto-complete query and (again, my assumption) cause unoptimized queries as well.

PR #4372, opened 2024-05-22
Reduce permissions of default authentication scope

### LINKED ISSUE #4324: Reduce permissions of default authentication scope
## Problem

While the Openverse ToS disallows scraping, our current authentication scheme and API parameter management is not conducive for preventing the easiest classes of abuse. Right now it is trivial for someone to authenticate at the most basic level and scrape the Openverse API. Even banning a single client application is easy to get around.

## Description

My proposed solution is to reduce the allowance of page size and page depth for the default authentication scope.

To do this, add a field to `ThrottledApplication` similar to `rate_limit_model` named `privileges`[^alternative-names]. The field should be an [`ArrayField`](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/fields/#arrayfield) with a `base_field` of `CharField` with `choices` set to a list of privileges. The choices to start with should be `increased_page_size` and `increased_query_depth`. The default of the field should be an empty list (i.e., no special privileges).

> This data model gives us the flexibility to add or remove different types of privileges for a given `ThrottledApplication`, without needing to automatically opt everyone with any privilege into all privileges or to create a confusing hierarchy of combinations. If we used a boolean or a single CharField, then `True` would need to imply all expanded parameter allowances, or each individual choice would need to represent one or a mix of the choices. It's much cleaner to use the ArrayField. `ThrottledApplication` is pulled in full for every request with Bearer auth, so the list will be immediately available for the serializer to check permissions on the request.

[^alternative-names]: I also considered `api_scopes`, `query_scope` or `parameter_allowances` but rejected them. `api_scopes` has too much potential for confusion with OAuth scopes, which isn't really the question here. The latter two have an inherently narrow application to just search query. This field could represent access to other non-search endpoints, so a non-search specific name is more appropriate.

Make the field editable in Django admin so maintainers can add or remove privileges for individual client apps.

Once the new field is available, update the serializer to check the relevant entries in the new field. If the request parameters exceed any given privilege, add a warning to the request following the example of the source parameter warning:

https://github.com/WordPress/openverse/blob/7dd355000f4d1d4bd597e3256b1f804d43f13f18/api/api/serializers/media_serializers.py#L407-L417

The warning text should link to a section of the API documentation that explains the privileges and how to request increased access. This would be new documentation, and the implementing PR must add it. Unauthenticated requests should continue to return an error response as they do today.

We also need to be able to evaluate existing usage patterns and ensure that any integrations known to be safe and relying on these features that require a higher level of trust will continue to work. We should also ensure that existing requests, regardless of trust level, do not 403. Instead, I propose that we clamp parameters to the maximum value of their application scope, and along with the warning mentioned above, monitor the frequency of requests exceeding granted privileges on a per-application basis. A new log line for each request that exceeds privileges that includes the ungranted privilege and the client application name would be sufficient for analysis after deployment.

When a request exceeds the privileges, clamp the parameter down to the anonymous maximum.

### Specifics for each parameter

Based on the table below, I've made suggestions for behaviour for each specific parameter

<details><summary>The table was pulled from logs analysis using this query analysing the last month of API nginx logs for paginated requests</summary>

```
  filter @logStream like "nginx"
| parse request /(?<httpMethod>(GET|POST|PUT|DELETE)) \/v1\/(?<mediaType>(images|audio))\/((?<resource>[0-9a-z\-\/]*)\/)?(?<queryParams>\?.*)? /
| parse queryParams /page=(?<page>\d+)/
| parse queryParams /page_size=(?<pageSize>\d+)/
| fields isempty(resource) and not isempty(mediaType) as isSearch, page * least(20, pageSize) as paginationDepth
| filter isSearch and page > 1 and status >= 200 and status <= 300
| stats count() as paginatedQueryCount,
  max(greatest(1, page)) as maxPage,
  min(least(1, page)) as minPage,
  max(greatest(20, pageSize)) as maxPageSize,
  min(least(20, pageSize)) as minPageSize,
  avg(least(20, pageSize)) as averagePageSize
  by client_application_name

```
</details>

For privacy, I've passed the client application names through sha1

| client_application_name | paginatedQueryCount | maxPage | minPage | maxPageSize | minPageSize | averagePageSize |
| --- | --- | --- | --- | --- | --- | --- |
| <no auth> | 666581 | 100 | 1 | 20 | 1 | 19.992 |
| 5dceabe8e24ce99c33a4e2c82b44e7d7c04df303 | 12167 | 20 | 1 | 24 | 20 | 20 |
| 235e9ac0fe5d168ed2c8245440cbce8c3deca272 | 153 | 11 | 1 | 50 | 20 | 20 |
| 09323e87079447d33d3564d1111d060693310236 | 246 | 12 | 1 | 25 | 20 | 20 |
| 46dc23b46b51e47c710bdae407469947973b7e85 | 106 | 11 | 1 | 500 | 1 | 18.8962 |
| 76e45af6b84391eb0a74170a0c281a9f33bcf3e3 | 447 | 20 | 1 | 20 | 20 | 20 |
| da0e53723b9999ce1d916823b2af6f732d458bb6 | 190 | 20 | 1 | 24 | 20 | 20 |
| 101c63b2f5d652671f29898455f17da436074a2b | 123 | 20 | 1 | 500 | 10 | 16.9106 |
| 3ea411a77707d2551bee90c10260d4f2ab1bad0c | 87 | 6 | 1 | 25 | 20 | 20 |
| 912a6241f10c3faa0460b71511259e1f15109b7d | 19 | 6 | 1 | 25 | 20 | 20 |
| 2a3aec6e30b424572565e331fa566fa1615d249b | 17 | 5 | 1 | 25 | 20 | 20 |
| 3afac6d593a6457ca512f6664e135174a03f7c1f | 2 | 2 | 1 | 25 | 20 | 20 |
| 84023988640fd26e8f592a42160a2cfb536aa739 | 41 | 5 | 1 | 500 | 20 | 20 |
| 0854151fc8312612081a246d55ef8b3fe731cc61 | 17 | 3 | 1 | 100 | 20 | 20 |
| b8d84862bbaed24eba05fa85cc77373f75cac31b | 4 | 5 | 1 | 20 | 20 | 20 |
| a3213452eee0959445e8828c50637f9fd043ab18 | 5 | 2 | 1 | 20 | 20 | 20 |
| 8690a4e435333974f3d632c25351560abcac546e | 6 | 3 | 1 | 25 | 20 | 20 |
| b1c7ddb28dcbf9fe1c873c57fce8147db97d1c8a | 1 | 3 | 1 | 20 | 20 | 20 |
| 096fdac8d366831ded024ebd947369febb34472e | 2 | 2 | 1 | 20 | 20 | 20 |
| 9040e29a872a36dc4b6a5efc0e86e65328947469 | 1 | 2 | 1 | 30 | 20 | 20 |
| 2e2eb310033a0b99997c59a981197678d7b82991 | 10 | 5 | 1 | 20 | 20 | 20 |

Of these, the two with a maxPageSize of 500 are known to be unwanted requesters, and almost certainly scrapers.

#### `page_size`

Authenticated requesters should get a default maximum of 50. This is based on usage from known integrations where integrations we are aware of and have relationships with cap out their page size at 50. There is one integration that made over 100 requests in the last month that went up to 50 page size. I don't see a problem with that, and it seems as good a number as any to create the somewhat arbitrary cut-off of default privileges granted to authenticated requests.

#### `page`

If the page size maxes out at 50, page maxing out at 20 would be 1000 images. I don't necessarily feel that's excessive, but I think we should reduce the default privilege to max out at 10 pages of requests, if not 5. Most requesters do not go above 5, and all requesters should be respecting the page_count parameter on responses anyway, so reducing the maximum will not cause integration problems. Openverse search relevancy beyond the first few pages is not that useful in most cases anyway.

I ran another query to get the average page, and most are below 5 on average, even for large integrations.

I propose we start with 5 given the average case will not be affected, and given what we know about Openverse search relevancy. That creates a maximum result count of 250.

## Alternatives

Continue to have extremely broad and generous API allowances that make ToS abuse trivial and impossible to reasonably regulate. We can just "ignore" this issue. If we do, I question whether our ToS would ever be enforceable or how transparent our application of the policy really could ever be.

sarayourfriend:
## Fixes

Fixes #4324 by @sarayourfriend 

## Description

This PR adds a new `Privilege` concept to the Openverse API, and allows us to "upgrade" access to fine-grained aspects of the API for certain requesters. I've used this to add further gradations to the limits on page size and pagination depth in line with the issue's requirements.

I've left comments on the PR to help explain some of the changes. The line count looks big, but it really is mostly a lot of changes to unit tests and reorganisation of the admin site configuration. The unit tests had some severe duplication of the `media_type_config` fixture, which is now fixed, and cleans the test data up a bit. This makes the tests consistent across the integration and unit tests, and easier to reason about and work with (no more context switching _which media type fixture is this one?_...). I ended up touching these because the changes in the serializer at first assumed the `request` was always available. This works in practice, but not in tests, because we actually do not always instantiate the serializer with a fully and correctly configured request object. That's pretty frustrating, but fixing it would require even _more_ changes outside the scope of this PR, so for now the serializer shamefully allows the request to be missing so that tests will not fail. I've fixed the issue for some tests, but really we should have a helper method for the tests to use that creates the serializer with particular data and an correctly configured authed or anonymous request. Or, even better, move more tests towards the integration level, rather than unit level, and forget about these implementation details of the serializer in the tests (where they really don't matter). Anyway, that's all a lot more work and this is messy enough as it is.

Otherwise, the bulk of the real changes are the new api.constants.privileges module and its data, along with the changes to the media serializer.

## Testing Instructions

Run the app with `just api/test` and make anonymous requests, and see that limits are enforced. Page size should be limited to 20, and you shouldn't be able to go deeper than 240 results (12 pages of 20 size).

Make authenticated requests, and confirm the behaviour matches the configuration and your understanding of it. Do the same with a privileged app, by visiting the Django admin and adding the privileges on the ThrottledApplication. Enable the privilege by checking the checkbox for that privilege, and save the throttled application.

Review the new unit tests in `test_auth` and confirm they cover the requirements.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Something (presumably caused by my deduplication of the `media_type` fixture in favour of the more complete, older, and widely used `media_type_config`), is causing test state to leak between tests, and the query assertions are failing due to the enabled sources query augmentation.

Very frustrating! I can't see at all how the test configuration is causing that to happen, and it only happens on a full test run, so it's purely down to some luck of ordering. Very annoying. I'm trying to find a way to completely squash the redis data on each subsequent test run so there's no chance of either the real redis being used nor of the data leaking between tests.

api/conf/settings/spectacular.py
```diff
@@ -50,19 +50,17 @@
             "name": "auth",
             "description": dedent(
                 """
-                The API has rate-limiting and throttling in place to prevent
```
sarayourfriend, line 53: I chose not to list out all the details of what the limits are for throttling or privileged pagination. The throttle limits were wrong, sort of, in that those are the limits for _some_ views, but thumbnails and some others have different limits. I don't think we actually need to publish the configured limits explicitly, and doing it like this (when we know they are configured via environment variables elsewhere anyway) inevitably leads to inaccurate information and drifting documentation.

Instead, I've changed this to describe the general idea of Openverse's API having intentional limits to requesters, how we communicate those limits (kept from the existing docs), and letting the reader know they can request higher limits if they wish.

api/test/fixtures/media_type_config.py
```diff
@@ -0,0 +1,161 @@
+from dataclasses import dataclass
```
sarayourfriend, line 1: This is just moved from `api/test/unit/conftest.py` and deduplicates the much more recent and less complete `media_type` fixture removed from `api/test/integration`.

api/test/fixtures/rest_framework.py
```diff
@@ -13,3 +16,25 @@ def request_factory() -> APIRequestFactory():
...
+    return token
+
+
+@pytest.fixture
+def authed_request(access_token, request_factory):
+    request = request_factory.get(
+        "/", HTTP_AUTHORIZATION=f"Bearer {access_token.token}"
+    )
+
+    return APIView().initialize_request(request)
```
sarayourfriend, line 35: Fixtures in this module are just moved around, but I've also fixed an issue in this `authed_request` fixture where it was causing `request.auth` to be a token string, rather than the actual authenticated access token object that it would be in normal circumstances. It was using `force_authenticate` which causes `auth` to be just a string, and to bypass the authentication class. We don't need to do that, we have a real token, so we can just rely on the normal authentication configuration. That's a good thing! It means we're integrating the whole authentication flow into these tests and ensuring it works in all scenarios that we have tests for.

api/test/integration/test_auth.py
```diff
@@ -249,34 +250,6 @@ def test_authority_authed(
...
-        "/v1/images/", query_params, HTTP_AUTHORIZATION=f"Bearer {token}"
-    )
-
-    assert res.status_code == 200
-
-    query_params = {"page_size": 500}
-    res = api_client.get(
-        "/v1/images/", query_params, HTTP_AUTHORIZATION=f"Bearer {token}"
-    )
-    assert res.status_code == 200
```
sarayourfriend, line 277: These tests are superseded by the new ones at the end of this module.

api/test/integration/test_dead_link_filter.py
```diff
@@ -198,11 +197,3 @@ def no_duplicates(xs):
...
     # No results should be repeated so we should have no duplicate ids
     assert no_duplicates(ids)
-
-
-@pytest.mark.django_db
-def test_max_page_count(api_client):
-    response = api_client.get(
-        "/v1/images/", {"page": settings.MAX_PAGINATION_DEPTH + 1}
-    )
-    assert response.status_code == 400
```
sarayourfriend, line 208: This test is replaced by the new `test_auth` tests.

api/test/unit/conftest.py
```diff
@@ -1,40 +1,6 @@
-from dataclasses import dataclass
```
sarayourfriend, line 1: All of these removed lines are just moved to api/test/fixtures/media_type_config.py

obulat:
I tried testing this PR locally, but I get 
```
django.db.utils.ProgrammingError: column api_throttledapplication.privileges does not exist
LINE 1: ..."verified", "api_throttledapplication"."revoked", "api_throt...
``` 
error in tests and when I try to open the ThrottledApplication admin page. I assumed that I needed to run `just api/dj migrate`, but this gives me the following error:
```
  Applying api.0062_decision_through_tables...Traceback (most recent call last):

django.db.utils.ProgrammingError: relation "api_imagedecisionthrough" already exists
```
What should I do?

It would also be easier to review this PR if the admin splitting was extracted to a separate PR.

api/api/constants/privilege.py
```diff
@@ -0,0 +1,72 @@
+import typing
```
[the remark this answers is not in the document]
obulat: What about `AccessLevel`?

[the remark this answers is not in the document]
stacimc: I like `AccessLevel`!

sarayourfriend:
@obulat you need to run `just api/init`, I though?

api/api/constants/privilege.py
```diff
@@ -0,0 +1,72 @@
...
+    "page_size",
+    anonymous=20,
+    authenticated=50,
+    # Max out privileged page size at the maximum authenticated
+    # pagination depth, otherwise privileged page size limit can
+    # contradict pagination depth limit.
+    privileged=240,
+)
+
+PAGINATION_DEPTH = Privilege(
```
stacimc, line 62: I found this to be confusing, because by the variable name I expected it to be the number of pages and then for the result count to be calculated as PAGE_SIZE * PAGINATION_DEPTH in the `clamp` functions later. Alternatively, if the Privilege were named 'RESULT_COUNT` instead that might also be clearer?

stacimc (commented):
The code, including tests, looks great to me @sarayourfriend. I really like the refactoring here, and the effort in comments and the commit history to make that easier to review 😅 

I have been having trouble actually testing but just confirmed it's an issue with my local env and _not_ this branch. Submitting my comments so far, but I'm going to continue looking at this now!

api/api/constants/privilege.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Privileges granted to applications upon approved request to Openverse maintainers.
+
+    Maintainers review requests for increased privileges on a per-case basis.
+
+    Distinct from ``rate_limit_model`` which only affects access rates rather than privileges.
+    """
+
+    slug: str
+    anonymous: typing.Any
```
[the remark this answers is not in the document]
sarayourfriend: In practice, they're currently `int`, but I resume they could be anything. I'll try and see if Python will support a Generic parameter here :thinking:

api/api/constants/privilege.py
sarayourfriend: Sure! I struggled to name this particular one as well. "Result count" is fine by me :+1:

api/api/constants/privilege.py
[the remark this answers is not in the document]
sarayourfriend: Okay, I've gone with AccessLevel as a clearer name for `Level`, but when I tried it for `Privilege`, it still didn't make semantic sense to me, for similar reasons to `Privilege`. Access level feels like an aspect of what's being described here, rather than the thing itself.

Instead, I came up with `RestrictedFeature`, which I think perfectly captures the thing being described. Let me know what y'all think.

api/api/constants/privilege.py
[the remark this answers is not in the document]
sarayourfriend: A generic appears to work! :tada: Well, whether it would actually pass mypy, I have no idea, but the code runs still and I don't get any errors with my informal type information from my IDE.

api/api/constants/privilege.py
sarayourfriend: I've gone with "query depth" to disambiguate from the result count of the query response. Does that help? I noticed we do use `max_result_count` in the tallying code, so I'm fine using it if query depth still isn't clear enough, just let me know.

sarayourfriend: Done in https://github.com/WordPress/openverse/pull/4372/commits/8d64c03117994a8bfe721002207a818ba2fac787

api/api/serializers/media_serializers.py
```diff
@@ -82,19 +81,73 @@ def validate_page_size(self, value):
...
+        _, max_depth = privilege.PAGINATION_DEPTH.request_level(
+            self.context.get("request")
+        )
+
+        page_size = self.data["page_size"]
+        max_possible_page_count = max_depth / page_size
+
+        if real_page_count > max_possible_page_count:
+            return floor(max_possible_page_count)
+
```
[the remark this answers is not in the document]
sarayourfriend: Let me know if the changes to names fixes this for you. Otherwise, I can try to think of a different way to express this.

api/api/constants/privilege.py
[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/WordPress/openverse/pull/4372/commits/8d64c03117994a8bfe721002207a818ba2fac787

stacimc (approved):
Thanks for your patience @sarayourfriend! Finally got it working and this tested well for me. I tried authenticated & unauthenticated, with setting the privileges individually and both at once. Everything worked beautifully  :)

obulat (approved):
The code looks great and works well locally (`just recreate` worked, thank you!).

My main change request is for form clarification. It can be done in this PR, or in a follow-up one.

As a maintainer trying to restrict a violating app's access level, what do I need to do? When I look at the form, without the knowledge about the underlying code, I'm not sure if I should be changing the `rate_limit_model` or the `privileges`, and which privileges should I restrict to fix the problem. 
We could add a documentation page about it, but I think for the maintainer trying to fix an API traffic problem, it would be nice to have an instruction inside the form itself. Something like "Unselect max_page_size for an app that is violating Openverse TOS by scraping content" in the privileges label.

PR #4374, opened 2024-05-23
Enable soft-locking functionality for media reports to assist simultaneous moderators

### LINKED ISSUE #3639: Soft lock moderation actions for works in review by a moderator
## Description

See the following for implementation description and rationale. **This issue covers only the point 3 in the linked section.** https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#id4

Blocked by https://github.com/WordPress/openverse/issues/3637 and https://github.com/WordPress/openverse/issues/3636

dhruvkb:
## Fixes

Fixes #3639 by @sarayourfriend 

## Description

This PR
1. creates a lock manager to keep lock management utilities separate from the admin
2. highlights reports that are under an active soft-lock in the list view
3. softly locks and unlocks works, using two new admin endpoints at `/lock` and `/unlock`, in response to events (including but not limited to loading the change page, switching between tabs or closing the browser)
4. shows a message to any subsequent moderator when they open a work that's soft-locked by someone else
5. times-out soft-locks after 5 minutes (this functionality may be redundant, even harmful, considering point 3 above)

## Testing Instructions

This is best tested using two browsers, or two profiles in separate windows of the same browser, or one regular and one incognito session in the same profile of the same browser.

1. In client A, login with admin creds `deploy` and `deploy`.
2. In client B, login with moderator creds `moderator` and `deploy`. (This was weird when I first learned about it, I expected the password to be the same as the username.)

### Test list highlights

...continued

3. In both clients open the list view. You should see no highlighted rows.
4. In client A, click on a report ID and open the change page.
5. Refresh client B to see that row highlighted in yellow.
6. In client A, click on multiple report IDs and open them in separate tabs.
7. Refresh client B to see all those rows highlighted in yellow.

### Test other moderator message

...continued

8. In client B, click on the report ID in the highlighted row and open the change page. You should see a message that "deploy" is also looking at the same report.
9. Refresh client A. You should see a message that "moderator" is also looking at the same report.

### Test page visibility changes

...continued

10. Close client A to a different tab.
11. Refresh client B, the message about the another moderator should disappear within 10 seconds.
12. Reopen client A to the report tab.
13. Refresh client B, the message should reappear immediately.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/templates/admin/api/media_report/change_form.html
+++ b/api/api/templates/admin/api/media_report/change_form.html
@@ -147,4 +147,42 @@
     img.src = url;
   }
 </script>
+
+<script>
+  // This script tag manages the soft-locking mechanism for the report.
+
+  function toggleLock(url) {
+    // Toggle the soft-lock for the current user-report pair.
+    fetch(url, {
+      method: 'POST',
+      keepalive: true,
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRFToken': document.querySelector('[name=csrfmiddlewaretoken]').value
+      },
+    })
+      .then(res => console.log(res))
+      .catch(err => console.error(err));
+  }
+
+  function softLock() {
+    toggleLock('{% url "admin:api_"|add:media_type|add:"report_lock" object_id=object_id %}')
+  }
+
+  function softUnlock() {
+    toggleLock('{% url "admin:api_"|add:media_type|add:"report_unlock" object_id=object_id %}')
+  }
+
+  // On first load, set the soft-lock.
+  document.addEventListener("DOMContentLoaded", softLock)
+  // Subsequent visibility changes (including unload and other ways to
+  // close the page) will toggle the soft-lock.
+  document.addEventListener("visibilitychange", function() {
+    if (document.visibilityState === 'visible') {
+      softLock()
+    } else {
+      softUnlock()
+    }
+  })
+</script>
 {% endblock %}
--- a/api/api/templates/admin/api/media_report/change_form.html
+++ b/api/api/templates/admin/api/media_report/change_form.html
@@ -151,38 +151,26 @@
 <script>
   // This script tag manages the soft-locking mechanism for the report.
 
-  function toggleLock(url) {
-    // Toggle the soft-lock for the current user-report pair.
-    fetch(url, {
+  function softLock() {
+    fetch('{% url "admin:api_"|add:media_type|add:"report_lock" object_id=object_id %}', {
       method: 'POST',
-      keepalive: true,
+      keepalive: true, // This makes the request equivalent to a beacon.
       headers: {
         'Content-Type': 'application/json',
         'X-CSRFToken': document.querySelector('[name=csrfmiddlewaretoken]').value
       },
     })
-      .then(res => console.log(res))
-      .catch(err => console.error(err));
   }
 
-  function softLock() {
-    toggleLock('{% url "admin:api_"|add:media_type|add:"report_lock" object_id=object_id %}')
+  /**
+   * Activate the soft-lock and set an loop to keep it alive. Ensure
+   * that the loop interval is smaller than the TTL of the soft-lock.
+   */
+  function loadHandler() {
+    softLock()
+    setInterval(softLock, 5000)
   }
 
-  function softUnlock() {
-    toggleLock('{% url "admin:api_"|add:media_type|add:"report_unlock" object_id=object_id %}')
-  }
-
-  // On first load, set the soft-lock.
-  document.addEventListener("DOMContentLoaded", softLock)
-  // Subsequent visibility changes (including unload and other ways to
-  // close the page) will toggle the soft-lock.
-  document.addEventListener("visibilitychange", function() {
-    if (document.visibilityState === 'visible') {
-      softLock()
-    } else {
-      softUnlock()
-    }
-  })
+  document.addEventListener("DOMContentLoaded", loadHandler)
 </script>
 {% endblock %}
--- a/api/api/utils/moderation_lock.py
+++ b/api/api/utils/moderation_lock.py
@@ -9,6 +9,7 @@ from api.models.moderation import get_moderators
 
 
 LOCK_PREFIX = "moderation_lock"
+TTL = 10  # seconds
 
 logger = structlog.get_logger(__name__)
 
@@ -61,7 +62,7 @@ class LockManager:
             return
 
         object = f"{self.media_type}:{object_id}"
-        expiration = int(time.time()) + 5 * 60  # 5 minutes from now
+        expiration = int(time.time()) + TTL
 
         logger.info("Adding lock", object=object, user=username, expiration=expiration)
         self.redis.zadd(f"{LOCK_PREFIX}:{username}", {object: expiration})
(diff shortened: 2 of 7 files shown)


api/api/utils/moderation_lock.py
```diff
@@ -1,34 +1,41 @@
...
 class LockManager:
     def __init__(self, media_type):
         self.media_type = media_type
+        redis: Redis = django_redis.get_redis_connection("default")
+        try:
+            redis.ping()
+            self.redis = redis
+        except ConnectionError:
+            logger.error("Redis connection failed")
+            self.redis = None
```
sarayourfriend, line 25: Just a nit, but can you move this into a method `get_redis` that retrieves and tests it when used, rather than in the constructor? Otherwise, this constructor makes a blocking network call, which is rather unexpected.

This is more important once we move to use an async Redis client, because constructors cannot be asynchronous. In that case, each of the methods that use redis would already be asynchronous anyway.

Just a nit though, not strictly necessary for this PR, but something we know we'll need to refactor one day.

api/api/templates/admin/api/media_report/change_form.html
```diff
@@ -147,4 +147,42 @@ <h2>Other reports</h2>
...
+
+  function softUnlock() {
+    toggleLock('{% url "admin:api_"|add:media_type|add:"report_unlock" object_id=object_id %}')
+  }
+
+  // On first load, set the soft-lock.
+  document.addEventListener("DOMContentLoaded", softLock)
+  // Subsequent visibility changes (including unload and other ways to
+  // close the page) will toggle the soft-lock.
+  document.addEventListener("visibilitychange", function() {
```
sarayourfriend, line 180: This also includes switching tabs, even if the tab stays open. I think that's probably fine for the first pass... but we should make sure it's documented so the moderators understand that nuance here.

If there's a way to account for that in the condition, maybe some property on the event, it'd be good to do so, so that it stays soft locked if the moderator, for example, just opened a tab to review the creator's works on the source, or something like that.

With what we know of how this process will work and who will be doing it (specifically how many people will be doing it), I don't think this is a big deal. If it's easy to account for tab switching, that's great, but if it's a big hassle or introduces other complexities and nuances as a side-effect, then it isn't worth it for this first iteration.

api/api/utils/moderation_lock.py
```diff
@@ -0,0 +1,122 @@
...
+            pipe = self.redis.pipeline()
+            for key in self.redis.keys(pattern):
+                for value, score in self.redis.zrange(key, 0, -1, withscores=True):
+                    if score <= now:
+                        logger.info("Deleting expired lock", key=key, value=value)
+                        pipe.zrem(key, value)
+            pipe.execute()
+
+        for prefix in PREFIXES:
+            _prune(f"{prefix}:*")
```
sarayourfriend, line 41: Two issues here, one rather minor, and the other very much not minor.

First, the minor issue: is there a reason to avoid in-lining the closure's body as the loop body, rather than using a closure? The closure doesn't capture anything other than `self`, as far as I can tell, and using a loop would end up with virtually the same code, but in a more typical style. The utility of the closure here isn't obvious to me in this case, so it at least deserves an explanatory comment.

```suggestion
        pipe = self.redis.pipeline()
        for prefix in PREFIXES:
            pattern = f"{prefix}:*"
            now = int(time.time())
            for key in self.redis.keys(pattern):
                for value, score in self.redis.zrange(key, 0, -1, withscores=True):
                    if score <= now:
                        logger.info("Deleting expired lock", key=key, value=value)
                        pipe.zrem(key, value)

        pipe.execute()
```

That has the added (free) benefit of using the same pipeline, which is a small but nice improvement.

The other very significant issue, which is a blocker for this PR, is the use of `keys`, [which is O(n) where `n` is the number of keys in the database total (not just under the prefix)](https://redis.io/docs/latest/commands/keys/), and must not under any circumstances be used in our production code, as it will grind Redis to a halt. `keys` is Redis's version of a full-table scan.

Instead, we need to build a set of known keys and iterate through them. Luckily, the number of keys under the MODERATOR_LOCK_PREFIX is known, and easily retrievable:

```py
def get_moderator_locks(self):
    admin_usernames = get_user_model().objects.filter(...).values_list('username', flat=True)

    moderator_locked_works = redis.mget([f"{MODERATOR_LOCK_PREFIX}:{username}" for username in admin_usernames])

def prune():
    moderator_locks = self.get_moderator_locks()

    # do the pruning...
```

With this in mind, because we can never safely select all the `REPORT_LOCK_PREFIX`, I think we should ditch the report -> moderator mapping, and only use the moderator -> report mapping. The number of possible moderators is very small, so retrieving the set of locked works for each moderator is trivial (and literally infinitely less time-complex than the other way around).

Coupled with the polling approach I'm suggesting in the PR review comment, this should also simplify the overall approach.

Removing the report lock prefix does, strictly speaking, hurt reads when creating the table. However, the size of the table is limited, and realistically the search space of _locked works_ inside the list of works locked by each moderator will be very small, usually 0, and when not 0, then something like 1 and 2. It's trivial to create a set of those IDs and annotate the report list in Django ORM using that set at query time, and use it to represent the row. In fact, that's even _more_ efficient than the current approach of making a separate Redis request for each row of the table, like we are now.

sarayourfriend (changes requested):
Requesting changes due to the `keys` issue. Otherwise, this LGTM, and generally I am keen on the HTTP request approach! At first, I was sceptical of using HTTP requests to modify the lock, but I see how that accomplishes some interesting things, namely allowing us to detect navigation away even if the moderator closes the Django admin without navigating to another admin page first. It's cool to release the lock in this way.

However, in addition to the `keys` issue, outlined in my other comment, I think if we used a polling-send approach, we can also fix the tab switching issue, make it possible for a moderator to have a lock on multiple reports (if they have them open), and remove the need to ever manually release the locks.

Basically, this approach would use a very short expiration (lets say 10 seconds), and have the JavaScript send a polling request every 5 seconds. Each request sets the expiration for that work a further 10 seconds into the future. We can prune at this point as well. Setting a lock on one work does not release locks on other works for that same moderator, we only prune expired locks.

I want to emphasise that I think the HTTP request approach is really cool, and it's not one I would have thought of, I think. The use of `keys` and the improvements to be had by using a poll-send-expire approach instead of a lock/release approach notwithstanding, I was ready to "LGTM, let's merge this despite some nitpicks" on this PR!

api/api/utils/moderation_lock.py
dhruvkb: I've refactored this in 2e7c59e so that `self.redis` is no longer a thing in `LockManager`.

api/api/templates/admin/api/media_report/change_form.html
dhruvkb: In my first implementation, I was using an event just before a page is unloaded, but those approaches are fickle and unreliable. Then I thought that a lock should actually be lifted if the moderator switches to a different tab (considering all the media information is present in the report page itself) but you make a valid point.

I've switched to your polling based approach with short TTL so that even a background tab preserves the lock as long as it remains open.

dhruvkb:
Closing in favour of a much more holistic implementation in #4386.

PR #4387, opened 2024-05-26
Make miscellaneous improvements to the API developer experience

dhruvkb:
## Description

This PR
- disables caching of templates during development (otherwise template changes do not reflect unless the server is reloaded)
- fixes the verbose name of the API (so that "Api" is no longer used)
- fixes the verbose name of model fields with the acronym URL (so that "url" or "Url" are no longer used)

While the addition of the verbose names does lead to a new migration, the migration is a no-op in SQL.

<details>
<summary>SQL</summary>

```sql
BEGIN;
--
-- Alter field creator_url on audio
--
-- (no-op)
--
-- Alter field foreign_landing_url on audio
--
-- (no-op)
--
-- Alter field url on audio
--
-- (no-op)
--
-- Alter field creator_url on audioset
--
-- (no-op)
--
-- Alter field foreign_landing_url on audioset
--
-- (no-op)
--
-- Alter field url on audioset
--
-- (no-op)
--
-- Alter field creator_url on image
--
-- (no-op)
--
-- Alter field foreign_landing_url on image
--
-- (no-op)
--
-- Alter field url on image
--
-- (no-op)
COMMIT;
```
</details>

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/conf/settings/base.py
```diff
@@ -40,13 +40,18 @@
 
 ROOT_URLCONF = "conf.urls"
 
+loaders = [
+    "django.template.loaders.filesystem.Loader",
```
sarayourfriend, line 44: Do you know if we strictly need the `filesystem.Loader`? `DIRS` is set to just the `api` app's template directory, which makes me think our usage of `filesystem.Loader` is duplicative of `app_directories.Loader`, which already sources templates from `<app_dir>/templates`.  I think we can rely entirely on the `app_directories.Loader`, in that case, right? Maybe we can leave this loader out entirely?

sarayourfriend (commented):
Our usage of uvicorn reload instead of Django's development mode causes some of Django's normal dev-mode features not to work by default, I guess this is one of them. Did you happen to try configuring uvicorn's reload watcher pattern instead of changing Django's template caching behaviour?

https://www.uvicorn.org/settings/#reloading-with-watchfiles

It'd be nice to actually keep template caching locally when the templates aren't changing, so that at least we would theoretically see an issue if template caching caused an issue when it ran. If we tell uvicorn to also reload when templates change, then the Django configuration wouldn't need to change between prod and local. This seems to work for me locally:

```diff
diff --git a/api/run.py b/api/run.py
index 0265ad680..521625007 100644
--- a/api/run.py
+++ b/api/run.py
@@ -18,6 +18,8 @@ if __name__ == "__main__":
             ".",  # default, API directory
             "../packages/python/",
         ],
+        # *.py is default, add *.html so template changes also cause reloads
+        reload_includes=["*.py", "*.html"],
         log_level="debug",
         access_log=False,
     )
```

That should let us avoid needing to mess with the template loader configuration at all.

Can you confirm whether you've tried this, and if it doesn't work can you document in the PR why? This LGTM to me otherwise, and happy to approve it, but wanted to make sure there wasn't a potentially simpler solution that would also avoid introducing a meaningful difference in the way templates are processed locally vs in production.

dhruvkb:
This PR is exactly what Django used to do before version 4.1. In subsequent versions, they enabled caching of templates locally, and this PR just reverses their change. 

We could configure Uvicorn to track HTML files and reload if they've changed but I didn't even try that because this worked.

sarayourfriend:
> This PR is exactly what Django used to do before version 4.1. In subsequent versions, they enabled caching of templates locally, and this PR just reverses their change.

Good to know! In that case, the difference between local and production I suppose we don't need to worry about (though they changed it for some reason, I guess?). If the uvicorn watcher approach is simpler, then great, otherwise this works too :+1:

dhruvkb:
The change to reload Uvicorn when templates works well and is much more minimal compared to changing the loaders, so I've updated the PR.

One thing to note is that disabling caching by changing the loaders (the previous implementation of this PR) was faster during development as it did not require a full reload of Uvicorn. It would continue the Python process and merely read the latest contents of the template file from disk each time.

api/run.py
```diff
@@ -18,6 +18,8 @@
             ".",  # default, API directory
             "../packages/python/",
         ],
+        # Also reload Uvicorn when templates change.
+        reload_includes=["*.py", "*.html"],
```
obulat, line 22: Thank you for finding this, it's really helpful for working on admin!

obulat (approved):
Nice improvements, thank you for finding them!

api/run.py
dhruvkb: 💯 That is precisely why I made the PR! I got incredibly tired of `touch`-ing random `.py` files.

sarayourfriend (approved):
LGTM!

PR #4396, opened 2024-05-28
Improve accessibility labels for filters tab and button

### LINKED ISSUE #957: Better accessible name for Filters button and tab
## Problem

#826 added a counter to the filters button and filters tab. The labels for these filters appear in two places:
- [`VFilterButton` on wider viewports](https://github.com/WordPress/openverse/blob/ea17972522d59a56dec2f08e8a2fa68230992dcc/frontend/src/components/VHeader/VFilterButton.vue) (explicitly generated by the [aria-label here](https://github.com/WordPress/openverse/blob/ea17972522d59a56dec2f08e8a2fa68230992dcc/frontend/src/components/VHeader/VFilterButton.vue#L57-L59), currently of the form `{count} Filters` in English)
- [`VFilterTab` on smaller viewports](https://github.com/WordPress/openverse/blob/bee799971e740105a213668f6d1b35c8813ab24c/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue) (implicitly generated by the textual content of the tab, either `Filters` or `{count} Filters` depending on whether or not filters are applied)

### Screenshots

**Wider viewports**:

image

image

**Smaller viewports**:

image

image

## Description

Two things will need to be addressed here:
1. Unifying the accessibility information available for both viewport sizes. Having an explicit aria-label is useful rather than relying on the textual content, since the order of the number and the word describing `Filters` could be different in other languages.
2. Determining if `{count} Filters` is indeed the most accessible language here.

On the latter point, visually the button/tab conveys the notion that the filters are available under this button (if none are currently applied) or it shows the number of filters currently applied to the search along with the impression that the button is a controller for the filters pane. What is the best way to convey this meaning via text? Would something like `Filters: {count} enabled` or `Filters: {count} applied` be appropriate? Does `{count} Filters` work as-is?

AetherUnbound:
## Fixes

Fixes #957 by @obulat

## Description

This PR adds a common `aria-label` to both the `VFilterTab` and `VFilterButton` (based on feedback provided in #957) which more appropriately conveys in text format what sighted users are expected to understand about the action component. The new label takes the form (in English) of `Filters ({number})`.

I've opted to reduce the set of changes to just this, since the testing translation changes were pretty unrelated and this makes the changeset much smaller.

## Testing Instructions

1. Run the app locally using `just frontend/run dev`
2. Visit http://localhost:8443/search/?q=cats&license=pdm,cc0,by using a wide viewport
3. Right click on the filters button, and click "Inspect Accessibility Properties" (for firefox, not sure what this looks like in Chrome)
4. Verify that the text present for the button is `Filters (3)`
5. Change to a smaller viewport for mobile devices and reload the page
6. Open up the filters drawer with the button next to search
7. Right click on the filter tab and click "Inspect Accessibility Properties"
8. Verify that the text present for the button is `Filters (3)`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
+++ b/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
@@ -1,14 +1,18 @@
 <template>
-  <VTab id="filters" size="medium" class="gap-x-2">
+  <VTab id="filters" size="medium" class="gap-x-2" :aria-label="ariaLabel">
     <VFilterIconOrCounter :applied-filter-count="appliedFilterCount" />
     <h2 class="label-regular">{{ $t("filters.title") }}</h2>
   </VTab>
 </template>
 <script lang="ts">
+import { computed, defineComponent } from "vue"
+
+import { useI18n } from "~/composables/use-i18n"
+
 import VFilterIconOrCounter from "~/components/VHeader/VFilterIconOrCounter.vue"
 import VTab from "~/components/VTabs/VTab.vue"
 
-export default {
+export default defineComponent({
   name: "VFilterTab",
   components: { VFilterIconOrCounter, VTab },
   props: {
@@ -17,5 +21,12 @@ export default {
       default: 0,
     },
   },
-}
+  setup(props) {
+    const i18n = useI18n()
+    const ariaLabel = computed(() =>
+      i18n.tc("header.filterButton.withCount", props.appliedFilterCount)
+    )
+    return { ariaLabel }
+  },
+})
 </script>
--- a/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
+++ b/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
@@ -1,18 +1,19 @@
 <template>
-  <VTab id="filters" size="medium" class="gap-x-2" :aria-label="ariaLabel">
+  <VTab
+    id="filters"
+    size="medium"
+    class="gap-x-2"
+    :aria-label="$tc('header.filterButton.withCount', appliedFilterCount)"
+  >
     <VFilterIconOrCounter :applied-filter-count="appliedFilterCount" />
     <h2 class="label-regular">{{ $t("filters.title") }}</h2>
   </VTab>
 </template>
 <script lang="ts">
-import { computed, defineComponent } from "vue"
-
-import { useI18n } from "~/composables/use-i18n"
-
 import VFilterIconOrCounter from "~/components/VHeader/VFilterIconOrCounter.vue"
 import VTab from "~/components/VTabs/VTab.vue"
 
-export default defineComponent({
+export default {
   name: "VFilterTab",
   components: { VFilterIconOrCounter, VTab },
   props: {
@@ -21,12 +22,5 @@ export default defineComponent({
       default: 0,
     },
   },
-  setup(props) {
-    const i18n = useI18n()
-    const ariaLabel = computed(() =>
-      i18n.tc("header.filterButton.withCount", props.appliedFilterCount)
-    )
-    return { ariaLabel }
-  },
-})
+}
 </script>
--- a/frontend/src/locales/scripts/en.json5
+++ b/frontend/src/locales/scripts/en.json5
@@ -52,7 +52,7 @@
     loading: "Loading...",
     filterButton: {
       simple: "Filters",
-      withCount: "{count} Filter|{count} Filters",
+      withCount: "Filter ({count})|Filters ({count})",
     },
     seeResults: "See results",
     backButton: "Go back",
--- a/frontend/src/locales/scripts/en.json5
+++ b/frontend/src/locales/scripts/en.json5
@@ -52,6 +52,7 @@
     loading: "Loading...",
     filterButton: {
       simple: "Filters",
+      /** Used as the accessible label for the button or tab that opens the filters sidebar. Count refers the number of filters the user has enabled, not the number of available filters */
       withCount: "Filters ({count})",
     },
     seeResults: "See results",
--- a/frontend/src/locales/scripts/en.json5
+++ b/frontend/src/locales/scripts/en.json5
@@ -52,7 +52,7 @@
     loading: "Loading...",
     filterButton: {
       simple: "Filters",
-      withCount: "Filter ({count})|Filters ({count})",
+      withCount: "Filters ({count})",
     },
     seeResults: "See results",
     backButton: "Go back",
--- a/frontend/test/locales/ar.json
+++ b/frontend/test/locales/ar.json
@@ -52,7 +52,7 @@
     "loading": "جار التحميل...",
     "filterButton": {
       "simple": "مرشحات",
-      "withCount": "{count} عامل التصفية | {count} عوامل التصفية"
+      "withCount": "({count}) مرشحات|({count}) مرشح"
     },
     "seeResults": "انظر النتائج",
     "backButton": "عُد",
--- a/frontend/test/locales/ar.json
+++ b/frontend/test/locales/ar.json
@@ -52,7 +52,7 @@
     "loading": "جار التحميل...",
     "filterButton": {
       "simple": "مرشحات",
-      "withCount": "({count}) مرشحات|({count}) مرشح"
+      "withCount": "({count}) مرشح"
     },
     "seeResults": "انظر النتائج",
     "backButton": "عُد",
--- a/frontend/test/playwright/utils/navigation.ts
+++ b/frontend/test/playwright/utils/navigation.ts
@@ -49,12 +49,11 @@ export const searchTypeNames = {
  */
 export const openContentSettingsTab = async (
   page: Page,
-  tab: "searchTypes" | "filters" = "searchTypes",
-  dir: LanguageDirection = "ltr"
+  tab: "searchTypes" | "filters" = "searchTypes"
 ) => {
-  const tabKey = tab === "searchTypes" ? "searchType.heading" : "filters.title"
+  const tabId = tab === "searchTypes" ? "content-settings" : "filters"
 
-  await page.getByRole("tab", { name: t(tabKey, dir) }).click()
+  await page.locator(`#tab-${tabId}`).click()
 }
 
 /**
@@ -105,7 +104,7 @@ export const setContentSwitcherState = async (
       await buttonLocator.isEnabled()
       await buttonLocator.click()
     }
-    return openContentSettingsTab(page, contentSwitcherKind, dir)
+    return openContentSettingsTab(page, contentSwitcherKind)
   } else if (isPressed) {
     await closeContentSettingsModal(page, dir)
   }
--- a/frontend/test/playwright/utils/navigation.ts
+++ b/frontend/test/playwright/utils/navigation.ts
@@ -49,11 +49,18 @@ export const searchTypeNames = {
  */
 export const openContentSettingsTab = async (
   page: Page,
-  tab: "searchTypes" | "filters" = "searchTypes"
+  tab: "searchTypes" | "filters" = "searchTypes",
+  dir: LanguageDirection = "ltr"
 ) => {
-  const tabId = tab === "searchTypes" ? "content-settings" : "filters"
+  // Use a hard-coded Regex to match the dynamic Filters tab name
+  const tabName =
+    tab === "searchTypes"
+      ? t("searchType.heading", dir)
+      : dir === "ltr"
+        ? /filter/i
+        : /مرش/
 
-  await page.locator(`#tab-${tabId}`).click()
+  await page.getByRole("tab", { name: tabName }).click()
 }
 
 /**
@@ -104,7 +111,7 @@ export const setContentSwitcherState = async (
       await buttonLocator.isEnabled()
       await buttonLocator.click()
     }
-    return openContentSettingsTab(page, contentSwitcherKind)
+    return openContentSettingsTab(page, contentSwitcherKind, dir)
   } else if (isPressed) {
     await closeContentSettingsModal(page, dir)
   }


frontend/src/locales/scripts/en.json5
```diff
@@ -50,9 +50,9 @@
     aboutTab: "About",
     resourcesTab: "Resources",
     loading: "Loading...",
-    filterButton: {
+    filters: {
```
sarayourfriend, line 53: Just a note that by changing the name of this key, it will cause the translations for `filterButton.simple` to also go away and need to be translated again. It's better to keep the key the same, if we can, to reduce churn on the translators. This also has an immediate effect on translations in production, any languages that had `filterButton.simple` translated will no longer have it translated.

Ideally, we don't change the keys, if avoidable, and in this case it is (even if it's annoying to have the overly specific name).

Suggesting we keep the key the same for these reasons (even though it wasn't a great name to start with):

```suggestion
    filterButton: {
```

frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
```diff
@@ -1,14 +1,18 @@
 <template>
-  <VTab id="filters" size="medium" class="gap-x-2">
+  <VTab id="filters" size="medium" class="gap-x-2" :aria-label="ariaLabel">
```
sarayourfriend, line 2: Kind of a nit, but in the interest of avoiding unnecessary lines of code, you can use `$tc` in template contexts and avoid needing to create the computed, call `useI18n`, add the `setup`, etc.

```diff
diff --git a/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue b/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
index eb9bda1d7..7ae0385ee 100644
--- a/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
+++ b/frontend/src/components/VHeader/VHeaderMobile/VFilterTab.vue
@@ -1,13 +1,16 @@
 <template>
-  <VTab id="filters" size="medium" class="gap-x-2" :aria-label="ariaLabel">
+  <VTab
+    id="filters"
+    size="medium"
+    class="gap-x-2"
+    :aria-label="$tc('header.filters.withCount', appliedFilterCount)"
+  >
     <VFilterIconOrCounter :applied-filter-count="appliedFilterCount" />
     <h2 class="label-regular">{{ $t("filters.title") }}</h2>
   </VTab>
 </template>
 <script lang="ts">
-import { computed, defineComponent } from "vue"
-
-import { useI18n } from "~/composables/use-i18n"
+import { defineComponent } from "vue"
 
 import VFilterIconOrCounter from "~/components/VHeader/VFilterIconOrCounter.vue"
 import VTab from "~/components/VTabs/VTab.vue"
@@ -21,12 +24,5 @@ export default defineComponent({
       default: 0,
     },
   },
-  setup(props) {
-    const i18n = useI18n()
-    const ariaLabel = computed(() =>
-      i18n.tc("header.filters.withCount", props.appliedFilterCount)
-    )
-    return { ariaLabel }
-  },
 })
 </script>
```

sarayourfriend (changes requested):
LGTM! Only blocker is to not change the translation key (set it back to `filterButton`), but otherwise this works great :grin:

sarayourfriend:
> do we need to follow the steps of the [upload portion of the i18n documentation](https://docs.openverse.org/frontend/reference/i18n.html#steps) before/after this PR is merged?

Nope, it's handled automatically in this GitHub workflow when merging to main: https://github.com/WordPress/openverse/actions/workflows/generate_pot.yml

After merging, no need for any further intervention.

obulat (changes requested):
I looked at the visual regression test failures, but didn't understand what the problem is yet. As for the e2e tests, this patch should fix some of them:

```diff
Subject: [PATCH] Add catalog media properties docs

Signed-off-by: Olga Bulat <obulat@gmail.com>
---
Index: frontend/test/playwright/utils/navigation.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/frontend/test/playwright/utils/navigation.ts b/frontend/test/playwright/utils/navigation.ts
--- a/frontend/test/playwright/utils/navigation.ts	(revision d5123efe3d891dad612c8581d1fd36f166b0d9a5)
+++ b/frontend/test/playwright/utils/navigation.ts	(date 1716920676271)
@@ -50,11 +50,10 @@
 export const openContentSettingsTab = async (
   page: Page,
   tab: "searchTypes" | "filters" = "searchTypes",
-  dir: LanguageDirection = "ltr"
 ) => {
-  const tabKey = tab === "searchTypes" ? "searchType.heading" : "filters.title"
+  const tabId = tab === "searchTypes" ? "content-settings" : "filters"
 
-  await page.getByRole("tab", { name: t(tabKey, dir) }).click()
+  await page.locator(`#tab-${tabId}`).click()
 }
 
 /**
@@ -105,7 +104,7 @@
       await buttonLocator.isEnabled()
       await buttonLocator.click()
     }
-    return openContentSettingsTab(page, contentSwitcherKind, dir)
+    return openContentSettingsTab(page, contentSwitcherKind)
   } else if (isPressed) {
     await closeContentSettingsModal(page, dir)
   }

```

frontend/test/playwright/utils/navigation.ts
```diff
@@ -49,12 +49,11 @@ export const searchTypeNames = {
...
   page: Page,
-  tab: "searchTypes" | "filters" = "searchTypes",
-  dir: LanguageDirection = "ltr"
+  tab: "searchTypes" | "filters" = "searchTypes"
 ) => {
-  const tabKey = tab === "searchTypes" ? "searchType.heading" : "filters.title"
+  const tabId = tab === "searchTypes" ? "content-settings" : "filters"
 
-  await page.getByRole("tab", { name: t(tabKey, dir) }).click()
+  await page.locator(`#tab-${tabId}`).click()
```
sarayourfriend, line 56: Generally it's best practice to avoid element id-based locators in tests. They don't mirror human behaviour at all, regardless of what accessible technologies are in use. What's changed about the implementation that prevents selecting on the role and name any more? Is it because of the dynamic issue count? If so, we should use a regex to select based on a substring... you'll just need to copy in the RTL regex as distinct from the LTR (`/filters/` for LTR, `/التصفية/` for RTL, these regexes appear to work for me locally in Node).

sarayourfriend (approved):
This LGTM. I've left one comment regarding the use of an element-id selector in the test, which goes against best practices. I'm comfortable with you fixing it in a fast-follow PR to this one rather than blocking this PR now, but it does need to be fixed.

frontend/test/playwright/utils/navigation.ts
obulat: Our i18n setup in Playwright tests is quite basic, and does not have support for `$tc`. 
We do not want to use a hard-coded regex for the selector because the string might change without the changes to the key.
If we continue using the current test `t` function for this locator, it would return `Filter ({count})|Filters ({count})"` for English name of the tab. We could use some string manipulation and remove the part after the first `(`, but then the usefulness of this kind of selector is very questionable. Besides, for RTL, the order of words might be different, and we could be removing the part of the string that is actually useful (if the parentheses are in the beginning of the phrase).

My justification for this change was that this is a helper function used for opening the content settings modal. We could make sure that we use the role- and name- based selectors in the tests that specifically test the filters button and the way it opens the modal, and keep the `id`-based selector for the helper function used in other tests where this is not the behavior under test.

That said, I wonder if we really need the pluralization in the filters tab label ("Filter (1)"/"Filters (n)" distinction). Always using "Filters" would simplify the code, but would it look correct for an English speaker? "Filters (1)" and "Filters (5)"?

obulat: If we decide not to use pluralization for Filters, then this PR is ready. Otherwise, we will also need to update the `frontend/test/locales/ar.json` file with test values for Arabic (Google suggests "({count}) مرشح|مرشحات ({count})"), which will also probably cause the need to update RTL snapshots

sarayourfriend: There are a couple of things here:

1. My understanding is that pluralization might not be necessary because it's a stable name of the tab, not a reference to the number of filters applied. It's "<The> Filters <tab> ({count} <filters applied>)", in a sense?
2. Whether we pluralise or not is irrelevant to the question of a regex. The regex intentionally doesn't match the entire string and only the part that would be stable (and pulled from our stable testing environment documents, which are predictable). We don't need to update the testing translations to literally reflect the actual translation. The testing translations need to be _plausible_, and ideally somewhat reflect the length of the string in the tested language. So long as this tab is the filters tab, the search string would never change.
3. Navigation utilities, perhaps most of all, should follow the best practices for navigating the app "using the tools a human would have" because they're implicitly testing that getting around the application works for a human (as close as we can get). End-to-end tests are not unit tests, they are integration tests. I don't think they should take artificial short-cuts.

obulat: > 2\. Whether we pluralise or not is irrelevant to the question of a regex. The regex intentionally doesn't match the entire string and only the part that would be stable (and pulled from our stable testing environment documents, which are predictable). We don't need to update the testing translations to literally reflect the actual translation. The testing translations need to be _plausible_, and ideally somewhat reflect the length of the string in the tested language. So long as this tab is the filters tab, the search string would never change.

This would require a hard-coded regex instead of using the i18n keys.
I feel like there are 2 best practices that are in conflict here:
- use role and name based selectors to mirror the user behavior
- use the i18n keys and `t` function instead of hard-coding strings.

If we use the i18n key, we cannot easily create a regex that would correctly match the tab's accessible name because it would return `Filter ({count})|Filters ({count})"`, and we need to match `"Filter (1)"` or `"Filters (\d)"`. Even if we split the return value by a space, it would work for English("Filter" would match any variant), but I'm not sure it would work for Arabic (because the singular form might be very different from the plural one).

If we use the hard-coded regex, then we are going back on "not using hard-coded strings in tests".
Is there something I'm not seeing?

sarayourfriend: No, you're right, on the face of it using a regex is in conflict with avoiding hard coded text. But when avoiding hard coded text (which is a preference to make changing tests easier) conflicts with a best practice focused on ensuring the application is accessible, I have to think the accessibility preference is more important. Neither of these are absolute best practices or preferences, but certainly the one intended to ensure the accessibility of the application should be considered more important than one meant to make the development experience of test writing slightly simpler.

obulat (commented):
I realized that we don't use a pluralized (i.e., 1 filter/2 filters) for the visible label, it's always "Filters". So, we could probably use the single "Filters ({count})" for the aria-label as well.

In light of the discussion with @sarayourfriend, I created a patch that reverts the locator changes, and updates the filter locator to fix the tests:
```diff
Subject: [PATCH] Test fixes
---
Index: frontend/test/playwright/utils/navigation.ts
<+>UTF-8
===================================================================
diff --git a/frontend/test/playwright/utils/navigation.ts b/frontend/test/playwright/utils/navigation.ts
--- a/frontend/test/playwright/utils/navigation.ts
+++ b/frontend/test/playwright/utils/navigation.ts
@@ -49,11 +49,13 @@
  */
 export const openContentSettingsTab = async (
   page: Page,
-  tab: "searchTypes" | "filters" = "searchTypes"
+  tab: "searchTypes" | "filters" = "searchTypes",
+  dir: LanguageDirection = "ltr"
 ) => {
-  const tabId = tab === "searchTypes" ? "content-settings" : "filters"
+  // Use a hard-coded Regex to match the dynamic Filters tab name
+  const tabName = tab === "searchTypes" ? t("searchType.heading", dir) : dir === "ltr" ? /filter/i : /مرش/
 
-  await page.locator(`#tab-${tabId}`).click()
+  await page.getByRole("tab", { name: tabName }).click()
 }
 
 /**
@@ -104,7 +106,7 @@
       await buttonLocator.isEnabled()
       await buttonLocator.click()
     }
-    return openContentSettingsTab(page, contentSwitcherKind)
+    return openContentSettingsTab(page, contentSwitcherKind, dir)
   } else if (isPressed) {
     await closeContentSettingsModal(page, dir)
   }
Index: frontend/test/locales/ar.json
<+>UTF-8
===================================================================
diff --git a/frontend/test/locales/ar.json b/frontend/test/locales/ar.json
--- a/frontend/test/locales/ar.json
+++ b/frontend/test/locales/ar.json
@@ -52,7 +52,7 @@
     "loading": "جار التحميل...",
     "filterButton": {
       "simple": "مرشحات",
-      "withCount": "{count} عامل التصفية | {count} عوامل التصفية"
+      "withCount": "({count}) مرشحات|({count}) مرشح"
     },
     "seeResults": "انظر النتائج",
     "backButton": "عُد",

```

AetherUnbound:
Thank you @obulat for the patch! I've applied it and simplified the English translation. Let me know if anything else is needed for this!

frontend/src/locales/scripts/en.json5
```diff
@@ -52,7 +52,7 @@
     loading: "Loading...",
     filterButton: {
       simple: "Filters",
-      withCount: "{count} Filter|{count} Filters",
+      withCount: "Filters ({count})",
```
sarayourfriend, line 55: @AetherUnbound as per the earlier thread, it'd be good to include a translator note to describe the context of this string, otherwise it might not make sense what's going on here. To do so, you just need to add a JSDoc-style comment on the object property (made possible because json5 supports comments):

```suggestion
      /** Used as the accessible label for the button or tab that opens the filters sidebar. Count refers the number of filters the user has enabled, not the number of available filters */
      withCount: "Filters ({count})",
```

Something like that should be good enough... it still feels obscure if I imagine not knowing the precise context of this, but I'm not sure how detailed or wordy this should be. Maybe worth asking about it after merging in the i18n channel to see if folks have input on how better to handle this for the translators.

sarayourfriend (approved):
LGTM! Just a tip for adding a translators note and maybe a question to follow up with the GlotPress folks to make sure we're doing the best thing for them, but I don't think it needs to block this PR.

frontend/test/locales/ar.json
```diff
@@ -52,7 +52,7 @@
     "loading": "جار التحميل...",
     "filterButton": {
       "simple": "مرشحات",
-      "withCount": "{count} عامل التصفية | {count} عوامل التصفية"
+      "withCount": "({count}) مرشحات|({count}) مرشح"
```
obulat, line 55:
```suggestion
      "withCount": "({count}) مرشح"
```

obulat (approved):
I updated the Arabic text to also only include the plural version.
Congratulations on your big frontend PR ✨

PR #4408, opened 2024-05-30
Add rudimentary support for aliases in ov

### LINKED ISSUE #4407: Add support for aliases in `ov`
## Problem

We tried adding aliases in https://github.com/WordPress/openverse/pull/4343 using associative bash arrays. However, macOS's bash version is stuck on an older version that doesn't support associative arrays. So I removed that from the PR and decided to try it a different way in a future PR as a separate issue.

## Description

Create some method of expanding aliases when running `ov`. For example, `ov j` should expand to `ov just`.

It should be possible with one-dimensional arrays, and treating odd-or-0 index as the alias and the following even index as the expansion. This would be preferred, as it's probably less overhead than something else (like the Python script).

Another option I thought of was to use Python to process argv in the entrypoint and expand the aliases there. The entrypoint could end with `python ./docker/dev_env/expand_aliases.py "$@"`, and that script could import `$OPENVERSE_PROJECT/.ov_aliases.py` using a method like https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly, and then do something like:

```py
#... import personal_aliases from $OPENVERSE_PROJECT/.ov_aliases.py

shared_aliases = {
    "j": "just",
}

all_aliases = shared_aliases | personal_aliases

args = sys.argv[1:]
if args[0] in all_aliases:
  args[0] = all_aliases[args[0]]

sys.execv(args)
```

sarayourfriend:
## Fixes

Fixes #4407 by @sarayourfriend 
cc @dhruvkb and @AetherUnbound who will be the most immediately interested in this

## Description

This approach is similar to the associative array approach, but worse because we cannot easily look up an alias by its name, and there's no non-trivial way for aliases to stack (i.e., refer to another alias). We could do that by not breaking, and continuing to replace the first entry in the `args` array each time an expansion is encountered, but that would make alias expansion and cross-reference dependent on the order in which they're defined.

This approach is also susceptible to coding errors, if the arrays are not an even number of entries, and it could be hard to debug that.

The Python alternative I've described in #4407 might be better in that regard. Implementing cross-referencing aliases would be trivial in Python (because we can use dicts) and there wouldn't be an ordering issue. It's also much harder to introduce a difficult to debug or non-obvious coding error with that approach.

I'll leave this as a draft and try the Python version to see which is better.

## Testing Instructions

Try `ov j`, `ov nuxt`, and `ov welcome`. Define your own aliases. See if you understand the approach, does it feel easy to use, or is it too complicated?

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker/dev_env/shared_aliases.sh
+++ b/docker/dev_env/shared_aliases.sh
@@ -1,0 +1,18 @@
+#! /usr/bin/env bash
+
+# Aliases are defined as a one-dimensional bash array
+# where the odd-or-0 index entries are the aliases, and
+# the following even index is the expansion. We cannot use an
+# associative array because macOS's bash version is frozen and
+# doesn't support them.
+# **MULTI-WORD EXPANSIONS MUST BE QUOTED**, plain spaces
+# are the separator of bash array entries. See the `nuxt` alias as an example below.
+# Aliases are also currently limited in that they cannot reference each other or stack.
+
+declare -a shared_aliases
+export shared_aliases
+
+shared_aliases+=(
+  j 'just'
+  nuxt 'just p frontend dev'
+)
--- a/justfile
+++ b/justfile
@@ -66,6 +66,41 @@ install:
 @install-hooks:
     bash -c "cp ./docker/dev_env/hooks/* ./.git/hooks"
 
+# Create an `.ovprofile` as a starting point for development environment customisation. Does not make changes if the file already exists.
+init-ovprofile:
+    #! /usr/bin/env bash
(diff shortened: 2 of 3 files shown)


sarayourfriend:
Closing in favour of https://github.com/WordPress/openverse/pull/4409 because it fixes some of the problems with this approach with no disadvantages, as far as I can see.

PR #4409, opened 2024-05-30
Add support for aliases to `ov`

### LINKED ISSUE #4407: Add support for aliases in `ov`

sarayourfriend:
## Fixes

Fixes #4407  by @sarayourfriend 

cc @AetherUnbound and @dhruvkb who are probably the most interested with this.

## Description

An alternative to https://github.com/WordPress/openverse/pull/4408 using a Python-based approach.

This approach has all the features of the bash approach in https://github.com/WordPress/openverse/pull/4408, but none of the disadvantages. It shares the same overall problems like the aliases only being available in the entrypoint. They aren't available inside `just` recipes themselves, for example. I think that's fine though!

## Testing Instructions

Try `ov j`, `ov nuxt`, and `ov welcome`. See if it works and makes sense!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker/dev_env/expand_aliases.py
+++ b/docker/dev_env/expand_aliases.py
@@ -1,0 +1,28 @@
+import json
+import os
+import sys
+from pathlib import Path
+
+
+SHARED_ALIASES = {
+    "j": ["just"],
+    "nuxt": ["j", "p", "frontend", "dev"],
+}
+
+
+def expand_aliases(args: list[str]):
+    ov_aliases = Path(os.getenv("OPENVERSE_PROJECT")) / ".ov_aliases.json"
+    aliases = SHARED_ALIASES
+    if ov_aliases.is_file():
+        aliases |= json.loads(ov_aliases.read_text())
+
+    args = sys.argv[1:]
+
+    while args[0] in aliases:
+        args = aliases.pop(args[0]) + args[1:]
+
+    print(*args)
+
+
+if __name__ == "__main__":
+    expand_aliases(sys.argv[1:])
--- a/ov
+++ b/ov
@@ -57,7 +57,7 @@ fi
 case "$_cmd" in
 init)
   "$_self" build
-  "$_self" just install-hooks
+  "$_self" 'just install-hooks && just init-ov-aliases'
   ;;
 
 build)
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ logs/
 
 # Environment
 .env
-.ovprofile
+.ov_aliases.json
 
 # Python environments
 env/
--- a/docker/dev_env/entrypoint.sh
+++ b/docker/dev_env/entrypoint.sh
@@ -50,4 +50,5 @@ if [ -n "$PDM_CACHE_DIR" ]; then
   pdm config install.cache on
 fi
 
-bash -c "$*"
+expanded_args=$(python3 "$OPENVERSE_PROJECT"/docker/dev_env/expand_aliases.py "$@")
+bash -c "$expanded_args"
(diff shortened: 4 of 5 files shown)


ov
```diff
@@ -57,7 +57,7 @@ fi
 case "$_cmd" in
 init)
   "$_self" build
-  "$_self" just install-hooks
+  "$_self" 'just install-hooks && just init-ov-aliases'
```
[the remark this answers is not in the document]
sarayourfriend: Oops, yes it will, thanks!

sarayourfriend:
> I also needed to run ov build && ov init first

Ah, of course.

Hopefully the entrypoint doesn't need to change that often, but if we find it does for some reason, we can change it to a stub script that just passes through to a mounted script that doesn't need to be copied into the container. Something like

```sh
# entrypoint.sh

"$OPENVERSE_PROJECT"/docker/dev_env/real_entrypoint.sh "$@"
```

Where `real_entrypoint.sh` has everything that's going on here. It's another layer of indirection, which is kind of annoying, but would make iteration easier.

docker/dev_env/expand_aliases.py
```diff
@@ -0,0 +1,28 @@
+import json
+import os
+import sys
+from pathlib import Path
+
+
+SHARED_ALIASES = {
+    "j": ["just"],
+    "nuxt": ["j", "p", "frontend", "dev"],
```
dhruvkb, line 9: This feels like preferential treatment for the frontend 😄.

dhruvkb (approved):
LGTM. I would have preferred a more _standard_ way to define aliases than using a Python script, but this works well so it's okay with me.

docker/dev_env/expand_aliases.py
sarayourfriend: It was just the first useful one I could think of that clearly illustrated multi-term expansion :sweat_smile:

dhruvkb:
@krysal you have to link the `ov` script to a directory that's on your `$PATH`. On a Mac, `/usr/local/bin` is typically on the `$PATH` so this should work.

```bash
sudo ln -s <path to repo>/ov /usr/local/bin/ov
```

sarayourfriend:
Oh, that's no good! The hooks shouldn't rely on `ov` being in the path, that's definitely a bug. They should reference the script directly, I guess by relative path? I'm not sure what the CWD of a git hook is but I'm sure it's easy to find out. Issue here to fix that :https://github.com/WordPress/openverse/issues/4422

PR #4415, opened 2024-05-31
Add `backfillmoderationdecision` management command

### LINKED ISSUE #3641: Create `ModerationDecision` backfill management command
## Description

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#backfill-moderationdecision-and-drop-report-status

This issue only covers the creation and execution of the management command in staging and then production.

Blocked by #3636 and #3638.

krysal:
## Fixes

Fixes #3641 by @sarayourfriend 

## Description

This PR adds the command as described in the implementation plan section for [Deprecating and removing report status](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#deprecating-and-removing-report-status). It takes as options the media type, the username of the moderator (in production it should be `opener` but in local we use `deploy`) and a flag for whether to run the command or just get the count of report to process.

Verify that the mapping is according to the specification.

## Testing Instructions

To test manually, create some reports, mark them with a `reason` other than "pending" and run the command.

```
just dj backfillmoderationdecision [--dry-run | --no-dry-run] [--media-type {image,audio}] --moderator deploy
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/management/commands/backfillmoderationdecision.py
+++ b/api/api/management/commands/backfillmoderationdecision.py
@@ -1,0 +1,101 @@
+import argparse
+
+from django.contrib.auth import get_user_model
+
+from django_tqdm import BaseCommand
+
+from api.models import AudioDecision, AudioReport, ImageDecision, ImageReport
+from api.models.media import DMCA, MATURE_FILTERED, NO_ACTION, PENDING
+
+
+class Command(BaseCommand):
+    help = "Back-fill the moderation decision table for a given media type."
+
+    @staticmethod
+    def add_arguments(parser):
+        parser.add_argument(
+            "--dry-run",
+            help="Count reports to process, and don't do anything else.",
+            type=bool,
+            default=True,
+            action=argparse.BooleanOptionalAction,
+        )
+        parser.add_argument(
+            "--media-type",
+            help="The media type to back-fill moderation decisions.",
+            type=str,
+            default="image",
+            choices=["image", "audio"],
+        )
+        parser.add_argument(
+            "--moderator",
+            help="The username of the moderator to attribute the decisions to.",
+            type=str,
+            default="opener",
+        )
+
+    def handle(self, *args, **options):
+        dry = options["dry_run"]
+        username = options["moderator"]
+        media_type = options["media_type"]
+
+        MediaReport = ImageReport
+        MediaDecision = ImageDecision
+        if media_type == "audio":
+            MediaReport = AudioReport
+            MediaDecision = AudioDecision
+
+        non_pending_reports = MediaReport.objects.filter(decision=None).exclude(
+            status=PENDING
+        )
+        count_to_process = non_pending_reports.count()
+
+        if dry:
+            self.info(
+                f"{count_to_process} {media_type} reports to back-fill. "
+                f"This is a dry run, exiting without making changes."
+            )
+            return
+
+        if not count_to_process:
+            self.info("No reports to process.")
+            return
+
+        t = self.tqdm(total=count_to_process)
+
+        User = get_user_model()
+        try:
+            moderator = User.objects.get(username=username)
+        except User.DoesNotExist:
+            t.error(f"User '{username}' not found.")
+            return
+
+        for report in non_pending_reports:
+            decision = MediaDecision.objects.create(
+                action=self.get_action(report),
+                moderator=moderator,
+                notes="__backfilled_from_report_status",
+            )
+            report.decision = decision
+            report.save()
+            t.update(1)
+
+        t.info(
+            self.style.SUCCESS(
+                f"Created {count_to_process} {media_type} moderation decisions from existing reports."
+            )
+        )
+
+    @staticmethod
+    def get_action(report):
+        if report.status == MATURE_FILTERED:
+            return "confirmed_sensitive"
+
+        if report.status == NO_ACTION:
+            return "rejected_reports"
+
+        # Cases with status = DEINDEXED
+        if report.reason == DMCA:
+            return "deindexed_copyright"
+
+        return "deindexed_sensitive"  # For reasons MATURE and OTHER
--- a/api/api/management/commands/backfillmoderationdecision.py
+++ b/api/api/management/commands/backfillmoderationdecision.py
@@ -4,13 +4,14 @@ from django.contrib.auth import get_user_model
 
 from django_tqdm import BaseCommand
 
+from api.constants.moderation import DecisionAction
 from api.models import AudioDecision, AudioReport, ImageDecision, ImageReport
 from api.models.media import DMCA, MATURE_FILTERED, NO_ACTION, PENDING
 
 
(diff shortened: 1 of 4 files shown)


api/api/management/commands/backfillmoderationdecision.py
```diff
@@ -0,0 +1,101 @@
...
+        t = self.tqdm(total=count_to_process)
+
+        User = get_user_model()
+        try:
+            moderator = User.objects.get(username=username)
+        except User.DoesNotExist:
+            t.error(f"User '{username}' not found.")
+            return
+
+        for report in non_pending_reports:
```
sarayourfriend, line 73: This will pull every single non pending report into memory, as starting the iteration will evaluate the query set.

How many of these exist in production? If it's anything over ~100 I think we should chunk this.

Regardless of whether we chunk it up, we should use [`bulk_create`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-create) rather than `.save()` in each Python loop iteration which creates `N` queries. It will return the decision models in order of iteration, so we can reliably zip them with the chunk of reports to set the decision on each report. Postgres supports returning the keys so the returned models will have PKs.

sarayourfriend (commented):
Left a comment with just some fiddly ORM details, but if the number of non-pending reports in production is low enough, it probably won't even matter.

stacimc (commented):
In my tests, for both Image and Audio reports with the status "mature_filtered" did not get their action set to "confirmed_sensitive" as I expected (both had an empty action) 🤔   Are you able to reproduce that, @krysal?

api/api/management/commands/backfillmoderationdecision.py
krysal: Good suggestion! There are _only_ 135 non-pending reports, but it's a good practice to make these things more performance. I'll change it to do a bulk_create.

```sql
deploy@localhost:openledger> SELECT COUNT(*) FROM nsfw_reports WHERE status <> 'pending_review';
+-------+
| count |
|-------|
| 135   |
+-------+
SELECT 1
Time: 0.079s
```

krysal:
@stacimc I do get the "confirmed_sensitive" action for a "mature_filtered" report. What output for the command do you get?

api/api/management/commands/backfillmoderationdecision.py
krysal: Done in [4d97b12](https://github.com/WordPress/openverse/pull/4415/commits/4d97b1223c42f1ea0096f63480dbfdfb448091a8).

sarayourfriend:
@krysal I don't see any MediaDecisionThrough being created when I run this. The decisions are created, but not the through table records. I thought this would happen automatically, but it doesn't. What's more, it isn't possible to set it directly on the media, it needs to be handled through `image.imagedecision_set` on the media (or equivalent for audio), or on the decision with `decision.media_objs.set()`. But, we can also bulk create it anyway, if we have to do it manually, and this appears to work for me locally:

```diff
diff --git a/api/api/management/commands/backfillmoderationdecision.py b/api/api/management/commands/backfillmoderationdecision.py
index f88e40769..e9340604a 100644
--- a/api/api/management/commands/backfillmoderationdecision.py
+++ b/api/api/management/commands/backfillmoderationdecision.py
@@ -5,7 +5,7 @@ from django.contrib.auth import get_user_model
 from django_tqdm import BaseCommand
 
 from api.constants.moderation import DecisionAction
-from api.models import AudioDecision, AudioReport, ImageDecision, ImageReport
+from api.models import AudioDecision, AudioReport, ImageDecision, ImageReport, ImageDecisionThrough, AudioDecisionThrough
 from api.models.media import DMCA, MATURE_FILTERED, NO_ACTION, PENDING
 
 
@@ -43,9 +43,11 @@ class Command(BaseCommand):
 
         MediaReport = ImageReport
         MediaDecision = ImageDecision
+        MediaDecisionThrough = ImageDecisionThrough
         if media_type == "audio":
             MediaReport = AudioReport
             MediaDecision = AudioDecision
+            MediaDecisionThrough = AudioDecisionThrough
 
         non_pending_reports = MediaReport.objects.filter(decision=None).exclude(
             status=PENDING
@@ -83,6 +85,15 @@ class Command(BaseCommand):
             for report, decision in zip(reports_chunk, decisions):
                 report.decision = decision
             MediaReport.objects.bulk_update(reports_chunk, ["decision"])
+            MediaDecisionThrough.objects.bulk_create(
+                [
+                    MediaDecisionThrough(
+                        media_obj_id=report.media_obj_id,
+                        decision_id=report.decision_id,
+                    )
+                    for report in reports_chunk
+                ]
+            )
             t.update(1)
 
         t.info(
```

sarayourfriend (changes requested):
LGTM except for the issue with the decision through table, which I've described in a comment. Otherwise this is working fine when I test it, I did not have the issue Staci saw and explicitly made sure a decision was created in that case. However, I wonder if Staci is seeing an issue in Django admin caused by the lack of the decision through? Not sure what the downstream effects are _currently_ of a missing decision through row for a media object.

dhruvkb:
I'll hold off on reviewing this PR till you've had a chance to address @sarayourfriend's requests for changes.

The relationship between the media object and the decision can be created more succinctly using this

```python
decision.media_objs.add(media_obj)
```

where `decision` is the `ImageDecision`/`AudioDecision` instance and `media_obj` is the `Image`/`Audio` instance respectively. There is code very similar to the one you've created in the `media_report.py` file (maybe there is scope to extract them both into a separate util, but that might be me prematurely DRY-ing)

https://github.com/WordPress/openverse/blob/2da7be13b1c5dbca3d551b992fbbb90ec9be4cb9/api/api/admin/media_report.py#L443-L468

sarayourfriend:
The problem with doing it through the decision is that you have to go one-by-one, there's no way to make it a bulk operation; hence why I suggested `bulk_create` within the existing bulk creation loop. The relationships here are always 1-to-1, because of the nature of the backfill.

In other cases, where a single decision is being created for multiple media objects, then yes, setting it on the decision (or media) instance is more straightforward and doesn't require manually creating the through table instances. In this case, though, we should use bulk create, in line with the rest of the code.

krysal:
@sarayourfriend Good catch! I think the bulk updates of MediaDecision skip the MediaDecisionThrough table/model. Thanks for the handy diff, the changes have been applied.

stacimc (approved):
LGTM! No longer seeing issues with `confimed_sensitive`. I checked all action types and double checked the relationships are all set correctly, looks great 👍

sarayourfriend (approved):
LGTM!

Issue #4199, opened 2024-04-24
Remove and de-duplicate tags with leading/trailing whitespace


AetherUnbound:
## Description

We have some records in our data where there are duplicate tags, only the duplicate tag has leading or trailing whitespace. Here's an example: https://api.openverse.engineering/v1/images/2d454032-0cc1-48a5-8f40-e9235f1a4f12/

```json
"tags": [
        {
            "name": "abigfave",
            "accuracy": null
        },
        {
            "name": " abigfave",
            "accuracy": null
        },
        {
            "name": "artisanal",
            "accuracy": null
        },
        {
            "name": " artisanal",
            "accuracy": null
        },
        {
            "name": "color",
            "accuracy": null
        },
        {
            "name": " color",
            "accuracy": null
        },
        ...
    ],
```

This might need to be tackled in two steps, or a least an operation which covers both cases:
1. Remove leading/trailing whitespace from existing tags
2. Deduplicate tags which might previously have been separate ones

We will also want to check, similar to #1566, that any _new_ tags added always have extra whitespace stripped.

## Additional context

Related to #430

sarayourfriend:
@WordPress/openverse-catalog I'd like to take a shot at this issue. Am I correct to assume this should use the batched update DAG?

AetherUnbound:
If possible, it'd be best to combine both of those steps into a single batched update, that way we don't have to do two passes on the data! Might make for a tricky query, but then we only have to run it once 😄

sarayourfriend:
I guess `select distinct trimming_function(tag.name)`, or something along those lines would work?

Thanks for the input, y'all.

krysal:
Reopening for the pending execution of the `trim_and_deduplicate_tags` DAG.

krysal:
Solved in #4557.

PR #4366, opened 2024-05-21
Add catalog media properties documentation

obulat:
## Fixes

Fixes #2187 by @obulat

## Description

This PR adds the media properties documentation to the catalog docs.

I am not sure about `update_on`:  What is the difference between `last_sync_with_source` and `updated_on`?

## Testing Instructions

Read through the descriptions and confirm that they are accurate and full.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (`just catalog/generate-docs` for catalog
      PRs) or the media properties generator (`just catalog/generate-docs media-props`
      for the catalog or `just api/generate-docs` for the API) where applicable.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #4429, opened 2024-06-03
Add DAG to trim and deduplicate tags

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/4199 by @krysal

## Description

I've made use of the `batched_update_dag` to trim and deduplicate tags.

The approach is essentially: for each row, filter the list of tags distinct on the trimmed tag name, then set tags to the filtered list.
It's worth noting that I don't know whether doing this "for all rows" by using a `WHERE true` query is safe, and I'm asking for clarification on what the specific limitations of our data transformation are in https://github.com/WordPress/openverse/pull/4143#issuecomment-2143228283.

The update query produced by the batched update DAG looks like this:

```sql
UPDATE image
    SET tags = (
                        SELECT
                            jsonb_agg(
                                jsonb_set(
                                    deduped.tag,
                                    '{name}',
                                    to_jsonb(deduped.trimmed_name)
                                )
                            )
                        FROM (
                            SELECT DISTINCT ON (tag->>'name', tag->>'provider')
                                trim(tag->>'name') trimmed_name
                                tag
                            FROM jsonb_array_elements(tags || '[]'::jsonb) tag
                        ) deduped
                    )
    WHERE identifier in (
        SELECT identifier FROM trim_and_deduplicate_tags_image_rows_to_update
        WHERE row_id > 0 AND row_id <= 10000
        FOR UPDATE SKIP LOCKED
    );
```

There is one huge and significant caveat to this approach: it assumes that if the trimmed name and the provider match, that there is no other possibly meaningful difference. We can change that by adding additional clauses to `SELECT DISTINCT ON`. Are the name and provider the "unique" pair for a tag? I believe so, but I could be wrong!

The select query is just `WHERE true`, which selects every row. There's no other way to do this I can think of, because any _other_ approach would require doing the trimming and deduplication in the select query, which I don't believe is an option, and would require duplicating the transformation regardless (I think). Instead, this will just go through row-by-row and perform the transformation for each row, without causing a full table scan.

## Testing Instructions

To test this, you need to augment your local data. I've done this locally with this very hacky query. It duplicates the existing tags of a single image. It also adds one of the tags known to be on this image with an additional entry of that tag but from a distinct provider. That lets you test that the distinct on clause works.

```sql
UPDATE image
SET tags = (
  SELECT jsonb_strip_nulls(jsonb_agg(augmented) || jsonb_agg(augmented) || jsonb_build_array(jsonb_object('{name, "   background  ", provider, unmagical_stats_bot}')))
  FROM (
    SELECT
      '  ' || "name" || ' ' "name",
      provider,
      accuracy
    FROM jsonb_to_recordset(tags || '[]'::jsonb) as x("name" text, provider text, accuracy float)
  ) augmented
)
WHERE identifier = '4bc43a04-ef46-4544-a0c1-63c63f56e276';
```
This is an awful way to test it, but I don't know what the better way is. @WordPress/openverse-catalog if y'all can give guidance here, beyond modifying this query to make it create more cases, is there a better way to test this? Is there a way to write unit tests for this?

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -55,7 +55,7 @@ def trim_and_deduplicate_tags():
                                 )
                             )
                         FROM (
-                            SELECT DISTINCT ON (tag->>'name')
+                            SELECT DISTINCT ON (tag->>'name', tag->'provider')
                                 trim(tag->>'name') trimmed_name,
                                 tag
                             FROM jsonb_array_elements(tags || '[]'::jsonb) tag
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -47,15 +47,19 @@ def trim_and_deduplicate_tags():
                 "update_query": dedent(
                     """SET tags = (
                         SELECT
-                            jsonb_strip_nulls(jsonb_agg(trimed_and_deduped))
+                            jsonb_agg(
+                                jsonb_set(
+                                    deduped.tag,
+                                    '{name}',
+                                    to_jsonb(deduped.trimmed_name)
+                                )
+                            )
                         FROM (
-                            SELECT DISTINCT ON (trim("name"))
-                                trim("name") "name",
-                                provider,
-                                accuracy
-                            FROM
-                                jsonb_to_recordset(tags || '[]'::jsonb) as x("name" text, provider text, accuracy float)
-                        ) trimed_and_deduped
+                            SELECT DISTINCT ON (tag->>'name')
+                                trim(tag->>'name') trimmed_name,
+                                tag
+                            FROM jsonb_array_elements(tags || '[]'::jsonb) tag
+                        ) deduped
                     )"""
                 ),
                 "dry_run": False,
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -33,7 +33,7 @@ def trim_and_deduplicate_tags():
         trigger_dag_id=BATCHED_UPDATE_DAG_ID,
         wait_for_completion=True,
         execution_timeout=timedelta(hours=5),
-        max_active_tis_per_dag=1,
+        max_active_tis_per_dag=2,
         map_index_template="""{{ task.conf['table_name'] }}""",
         retries=0,
     ).expand(
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -1,6 +1,4 @@
 """
-# Trim and Deduplicate Tags
-
 See the issue for context and motivation: https://github.com/WordPress/openverse/issues/4199
 
 This DAG triggers a run of the batched update DAG. It generates a new list of tags by
@@ -25,6 +23,7 @@ DAG_ID = "trim_and_deduplicate_tags"
     schedule=None,
     start_date=datetime(2024, 6, 3),
     tags=["database"],
+    doc_md=__doc__,
     max_active_runs=1,
     default_args=DAG_DEFAULT_ARGS,
 )
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -41,25 +41,42 @@ def trim_and_deduplicate_tags():
             {
                 "query_id": f"{DAG_ID}_{media_type}",
                 "table_name": media_type,
-                # Just iterate through all the rows, don't bother sub-selecting as it's impossible to reasonably do so for trimming
-                "select_query": "WHERE true",
-                "update_query": dedent(
-                    """SET tags = (
-                        SELECT
-                            jsonb_agg(
-                                jsonb_set(
-                                    deduped.tag,
-                                    '{name}',
-                                    to_jsonb(deduped.trimmed_name)
+                "select_query": (
+                    "WHERE "
+                    + (
+                        ""
+                        if media_type == "audio"
+                        # We've identified these image providers in particular definitely have tags that need trimming, so we can narrow the query starting there at least
+                        else (
+                            "provider IN ('flickr', 'geographorguk', 'thingiverse', 'animaldiversity', 'clevelandmuseum', 'digitaltmuseum') AND "
+                        )
+                    )
+                    + (
+                        f"""EXISTS (SELECT 1 FROM jsonb_array_elements({media_type}.tags) AS t WHERE t->>'name' ILIKE ' %' OR t->>'name' ILIKE '% ')"""
+                    )
+                ),
+                "update_query": (
+                    "SET updated_on = now(), "
+                    + dedent(
+                        f"""
+                        tags = (
+                            SELECT
+                                jsonb_agg(
+                                    jsonb_set(
+                                        deduped.tag,
+                                        '{{name}}',
+                                        to_jsonb(deduped.trimmed_name)
+                                    )
                                 )
-                            )
-                        FROM (
-                            SELECT DISTINCT ON (tag->>'name', tag->'provider')
-                                trim(tag->>'name') trimmed_name,
-                                tag
-                            FROM jsonb_array_elements(tags || '[]'::jsonb) tag
-                        ) deduped
-                    )"""
+                            FROM (
+                                SELECT DISTINCT ON (tag->>'name', tag->'provider')
+                                    trim(tag->>'name') trimmed_name,
+                                    tag
+                                FROM jsonb_array_elements({media_type}.tags || '[]'::jsonb) tag
+                            ) deduped
+                        )
+                        """
+                    )
                 ),
                 "dry_run": False,
             }
--- a/catalog/dags/maintenance/trim_and_deduplicate_tags.py
+++ b/catalog/dags/maintenance/trim_and_deduplicate_tags.py
@@ -1,0 +1,68 @@
+"""
+# Trim and Deduplicate Tags
+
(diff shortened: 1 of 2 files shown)


AetherUnbound:
> The select query is just `WHERE true`, which selects every row. There's no other way to do this I can think of, because any _other_ approach would require doing the trimming and deduplication in the select query, which I don't believe is an option, and would require duplicating the transformation regardless (I think). Instead, this will just go through row-by-row and perform the transformation for each row, without causing a full table scan.

Given that the example result for this issue is a Flickr result and so it's safe to assume Flickr's data is affected (and also constitutes the largest set of data per provider), I wanted to check to see if any other providers were affected. Here's the results:

```sql
deploy@localhost:openledger> select count(*), provider from image
 where provider != 'flickr' and exists (
     select 1 from jsonb_array_elements(image.tags) as t where t->>'name' ilike ' %' or t->>'name' ilike '% ')
 group by 2;
 
+-------+-----------------+
| count | provider        |
|-------+-----------------|
| 90100 | geographorguk   |
| 4     | thingiverse     |
| 9     | animaldiversity |
| 5     | clevelandmuseum |
| 148   | digitaltmuseum  |
+-------+-----------------+
SELECT 5
Time: 7220.224s (2 hours 20 seconds), executed in: 7220.221s (2 hours 20 seconds)
```

With this, I actually think we can change `WHERE true` to `WHERE provider IN ('flickr', 'geographorguk', 'thingiverse', 'animaldiversity', 'clevelandmuseum', 'digitaltmuseum')`. This will help speed up both the select and the batched insert operation.

stacimc:
As I understand it, tags have already been deduplicated. So the only duplicates that remain will be for records that have tags that need to be trimmed, if once trimmed they match another existing tag. If that's correct, can't we select _only_ for records from that list of providers, which have at least one tag that needs trimming?

```
 f"""WHERE
 provider IN ('flickr', 'geographorguk', 'thingiverse', 'animaldiversity', 'clevelandmuseum', 'digitaltmuseum')
AND EXISTS (
    SELECT 1 FROM jsonb_array_elements({media_type}.tags) AS t WHERE t->>'name' ILIKE ' %' OR t->>'name' ILIKE '% '
)
"""
```

Technically we don't need to have a PR/DAG for this -- previously we've been running the batched updates manually with the exception of the popularity refresh. But for these large updates it's nice to go through the formal PR review process and testing! We do have [tests from the `batched_update` DAG](https://github.com/WordPress/openverse/blob/main/catalog/tests/dags/database/test_batched_update.py#L149) that could be used as a template for writing a few unit tests.

sarayourfriend:
Sure thing! I didn't know if querying on the contents of tags was actually safe and reliable, that's great to know. I've been significant confused and turned around with the different seemingly contradictory perspectives and information about what indexes matter when, that I've at this point essentially dropped any confidence in what performance characteristics to expect from the catalog database.

Thanks for the input both of you, I'll push an update to add the where clause. If you don't think this needs to be a DAG, we can close the PR and just run it manually. Up to you @stacimc and @AetherUnbound. We'd just delete this DAG after running it anyway (though I see other one-off DAGs that haven't been, so whether that's a reliable bet, I don't know).

sarayourfriend:
Here's a new query for creating testing data:

```sql
UPDATE image
SET tags = (
  SELECT jsonb_strip_nulls(jsonb_agg(augmented.tag) || jsonb_agg(augmented.new_tag))
  FROM (
    SELECT
        jsonb_build_object('name', '  ' || "name" || ' ', 'provider', provider, 'accuracy', accuracy) tag,
        jsonb_build_object('name', "name", 'provider', 'magical_computer_vision', 'accuracy', accuracy) new_tag
    FROM jsonb_to_recordset(tags || '[]'::jsonb) as x("name" text, provider text, accuracy float)
  ) augmented
)
WHERE identifier IN (SELECT identifier FROM image WHERE provider = 'flickr' LIMIT 10);
```

I've pushed the changes to filter the selected rows based on Staci's version, but I only select based on provider for image to avoid breaking audio. I'm not sure if audio needs this update at all, though. I suppose it doesn't hurt to run it there, and the dataset is small so there won't be any adverse effects.

@stacimc let me know if you prefer this to be a DAG or if you think it's better to just run it with the batched update DAG directly. I'm fine either way, just up to what process you want to follow :+1:

catalog/dags/maintenance/trim_and_deduplicate_tags.py
```diff
@@ -0,0 +1,88 @@
...
+                        else (
+                            "provider IN ('flickr', 'geographorguk', 'thingiverse', 'animaldiversity', 'clevelandmuseum', 'digitaltmuseum') AND "
+                        )
+                    )
+                    + (
+                        f"""EXISTS (SELECT 1 FROM jsonb_array_elements({media_type}.tags) AS t WHERE t->>'name' ILIKE ' %' OR t->>'name' ILIKE '% ')"""
+                    )
+                ),
+                "update_query": (
+                    "SET updated_on = now(), "
```
stacimc, line 59: Huh. _Wow_, good catch. This is explicitly set during ingestion, rather than using a trigger. I suspect this has gone uncaught in all previous batched updates, including the popularity refreshes. I'll make a PR to fix popularity refreshes, and we may need to think about adding a trigger or else enforcing it in the batched update DAG somehow.

stacimc (approved):
Looks great!

I think we should go ahead and merge the PR and delete the DAG afterward, at least this time. It makes it easy to search for the history in Github even after the DAG is deleted, and avoids any issues copy/pasting the parameters in. I don't think we should require a PR _every_ time (and I wouldn't have a blocking objection if you preferred to do even this one manually), but it seems like a good idea for more complex operations.

catalog/dags/maintenance/trim_and_deduplicate_tags.py
sarayourfriend: The `add_license_url` DAG includes it, which I was using as a reference for some of this work. I guess you have something like this in mind, with respect to a trigger?

https://stackoverflow.com/questions/9556474/automatically-populate-a-timestamp-field-in-postgresql-when-a-new-row-is-inserte/9556527#9556527

It would seem non-controversial that if a row is updated the `updated_on` should change, and a trigger would enforce that, and easily enough too.

sarayourfriend:
With approval from @AetherUnbound or another reviewer, I'll merge this and then discuss in Slack re: timing the run.

catalog/dags/maintenance/trim_and_deduplicate_tags.py
AetherUnbound: I actually raised this in the recent media properties PR for the catalog! https://github.com/WordPress/openverse/pull/4366#issuecomment-2136191545

I can go ahead and make an issue for adding a trigger for this, given you're on board Staci.

AetherUnbound (approved):
Looks good! My preference would be to run these manually rather than merge the DAG (so we don't have to take care of the cleanup steps), but until we have something more formalized (like #4458 or #4461), having a DAG run as history here is useful 🚀

catalog/dags/maintenance/trim_and_deduplicate_tags.py
```diff
@@ -0,0 +1,88 @@
...
+    max_active_runs=1,
+    default_args=DAG_DEFAULT_ARGS,
+)
+def trim_and_deduplicate_tags():
+    TriggerDagRunOperator.partial(
+        task_id=DAG_ID,
+        trigger_dag_id=BATCHED_UPDATE_DAG_ID,
+        wait_for_completion=True,
+        execution_timeout=timedelta(hours=5),
+        max_active_tis_per_dag=1,
```
krysal, line 36: Since this is iterating through the media tables it's safe to start both at the same time instead of waiting. That could speed up things a bit :)

```suggestion
        max_active_tis_per_dag=2,
```

krysal (approved):

catalog/dags/maintenance/trim_and_deduplicate_tags.py
AetherUnbound: Good thought! Although this is the `max_active_tis_per_dag`, which looks like [according to the docs](https://airflow.apache.org/docs/apache-airflow/stable/faq.html#how-to-improve-dag-performance) is:

> This parameter controls the number of concurrent running task instances across dag_runs per task

I think we also want to increase `max_active_runs` on L27 to `2`:

> Overrides [max_active_runs_per_dag](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#config-core-max-active-runs-per-dag)

catalog/dags/maintenance/trim_and_deduplicate_tags.py
stacimc: Agreed! Trigger is the right solution IMO, but I opened https://github.com/WordPress/openverse/pull/4460 to fix it in the popularity refresh in the interim. If we want to just jump ahead and go straight to adding the trigger instead that is also 100% fine with me.

catalog/dags/maintenance/trim_and_deduplicate_tags.py
sarayourfriend: Why do we need to increase the maximum number of DAG runs? Each run for all media tables is 2 tasks within a single run. I think Krystle's change alone would address that :thinking: Or does the max active dag run somehow also limit the max tasks running within a DAG?

AetherUnbound: Oh forgive me, I confused this for building up two separate DAGs rather than one DAG with two media types. That should be fine!

catalog/dags/maintenance/trim_and_deduplicate_tags.py
AetherUnbound: Issue created: https://github.com/WordPress/openverse/issues/4520

PR #4447, opened 2024-06-06
Capture thumbnails during europeana ingestion

### LINKED ISSUE #4403: Capture thumbnails during Europeana ingestion
## Description

We've been advised by the folks who maintain the Europeana API that it would be advisable to use the thumbnails they've already generated rather than trying to generate our own! Given that we have a _few_ issues with generating Europeana thumbnails (#3970, #3783, #3480), it would definitely be ideal to use the values they provide instead.

The thumbnails are available in the `edmPreview` field within the returned data, here's an example with our sample data: 

https://github.com/WordPress/openverse/blob/aedc9c16ce5ed11709e2b6f0b42bad77ea1eb19b/catalog/tests/dags/providers/provider_api_scripts/resources/europeana/europeana_example.json#L93-L95

It looks like there could be a list of thumbnails, we should probably just retrieve the first one from the list.

In order to accomplish this, we'll need to modify the [Europeana provider ingestion script](https://github.com/WordPress/openverse/blob/7f4fb7c91363747505d72a63dc7509617d16acfc/catalog/dags/providers/provider_api_scripts/europeana.py). We should add a new function to the `EuropeanaRecordBuilder` class to retrieve this value. A good example to use would be [`_get_filesize`](https://github.com/WordPress/openverse/blob/7f4fb7c91363747505d72a63dc7509617d16acfc/catalog/dags/providers/provider_api_scripts/europeana.py#L156-L157). We'll then need to capture this information in `get_record_data`, by adding it to the dictionary with the `"thumbnail_url"` key here:

https://github.com/WordPress/openverse/blob/7f4fb7c91363747505d72a63dc7509617d16acfc/catalog/dags/providers/provider_api_scripts/europeana.py#L73

We'll also need to add tests for this function and update any other Europeana tests that might be affected. An example test for `_get_filesize` can be [found here](https://github.com/WordPress/openverse/blob/33f9f7c54c2a27593716ce429b2b4b42f8df6e22/catalog/tests/dags/providers/provider_api_scripts/test_europeana.py#L299-L308).

madewithkode:
## Fixes

Fixes #4403 by @AetherUnbound 

## Description

This PR aims to use the thumbnails already generated by provider(Europeana) at the point of ingestion, as against trying to generate our own.

## Testing Instructions

See `test_get_thumbnail` in `catalog/tests/dags/providers/provider_api_scripts/test_europeana.py`

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/providers/provider_api_scripts/europeana.py
+++ b/catalog/dags/providers/provider_api_scripts/europeana.py
@@ -71,6 +71,7 @@ class EuropeanaRecordBuilder:
                 "license_info": self._get_license_info(data),
                 "filetype": self._get_filetype(item_data),
                 "filesize": self._get_filesize(item_data),
+                "thumbnail_url": self._get_thumbnail(data),
             } | self._get_image_dimensions(item_data)
 
             data_providers = set(record["meta_data"]["dataProvider"])
@@ -180,6 +181,14 @@ class EuropeanaRecordBuilder:
             return description[0].strip()
 
         return ""
+
+    def _get_thumbnail(self, data: dict) -> str | None:
+        # looks like edmPreview can either be a list or string
+        # this was inferred from observing the difference
+        # between sample data in item_full.json and europeana_example.json
+        # so it's best to handle both cases.
+        if preview := data.get("edmPreview", None):
+            return preview[0] if isinstance(preview, list) else preview
 
 
 class EuropeanaDataIngester(ProviderDataIngester):
--- a/catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
@@ -206,6 +206,7 @@ def test_record_builder_get_record_data(ingester, record_builder):
         "width": 381,
         "filesize": 36272,
         "filetype": "jpeg",
+        "thumbnail_url": "https://api.europeana.eu/api/v2/thumbnail-by-url.json?uri=http%3A%2F%2Fbibliotecadigital.jcyl.es%2Fi18n%2Fcatalogo_imagenes%2Fimagen_id.cmd%3FidImagen%3D102620362&type=IMAGE",
     }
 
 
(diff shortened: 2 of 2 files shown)


catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -71,6 +71,7 @@ def get_record_data(self, data: dict) -> dict | None:
                 "license_info": self._get_license_info(data),
                 "filetype": self._get_filetype(item_data),
                 "filesize": self._get_filesize(item_data),
+                "thumbnail_url": self._get_thumbnail(item_data),
```
krysal, line 74: The problem is passing `item_data` here, which doesn't contain the thumbnail, instead of `data`. Replacing it made the DAG collect the thumbs for me.

```suggestion
                "thumbnail_url": self._get_thumbnail(data),
```

catalog/dags/providers/provider_api_scripts/europeana.py
```diff
@@ -181,6 +182,19 @@ def _get_description(self, data: dict) -> str | None:
...
+        # looks like edmPreview can either be a list or string
+        # this was inferred from observing the difference
+        # between sample data in item_full.json and europeana_example.json
+        # so it's best to handle both cases.
+        if item_data.get("edmPreview", None):
+            return (
+                item_data.get("edmPreview")[0]
+                if isinstance(item_data.get("edmPreview"), list)
+                else item_data.get("edmPreview")
+            )
```
krysal, line 195: The number of lines can be reduced by using the **assignment expression operator**, a.k.a. the **_walrus operator_**. It comes in very handy for cases like this one.

```suggestion
        if preview := item_data.get("edmPreview", None):
            return preview[0] if isinstance(preview, list) else preview
```

catalog/tests/dags/providers/provider_api_scripts/test_europeana.py
krysal, line 1: You will need to update `test_record_builder_get_record_data` in this file as well, to include checking for the thumbnail.

krysal (changes requested):
@madewithkode Thanks for working on this! I ran the Europeana DAG but didn't get any thumbnails. Luckily, it's an easy and quick change. I commented on the exact line with the problem below.

catalog/dags/providers/provider_api_scripts/europeana.py
madewithkode: Thanks for pointing this out @krysal. It's funny how I wrote the "running DAGs locally" guide but still don't fully understand where to check for what, reason I couldn't verify this. At this point we might just need a video demo 😅

madewithkode:
PR has been updated with all requested changes.

obulat (approved):
Nice, thank you for your contribution, @madewithkode !

PR #4451, opened 2024-06-06
Update pr_automations.yml with missing character

zackkrida:
Fixes the PR automations so that artifacts are saved with the correct names.


dhruvkb (approved):
Great catch, thanks!

AetherUnbound (approved):

PR #4465, opened 2024-06-07
Add data flow diagram for various ETL steps in pipelines

### LINKED ISSUE #4455: Document current & desired ETL steps and data flow
## Description

Recent discussions about the projects #430, #431, and #3925 have brought up some interesting questions about _where_ in our general data flow certain operations are occurring. We have many different steps in our data pipeline where various ETL steps are occurring, and it would be useful to enumerate the current state and operations. We should also describe a _desired_ state that will be the result of the completion of the three aforementioned projects.

## Additional context

https://make.wordpress.org/openverse/2024/05/20/openverse-monthly-priorities-meeting-2024-06-05/#comment-282

AetherUnbound:
## Fixes

Fixes #4455 by @AetherUnbound

## Description

This PR adds a document which visualizes and describes the current and proposed data flows for the project. It also enumerates the ETL steps at each stage to make it easy to understand which operations are happening at which step.

I didn't include any transformations that happen between the client and the frontend because I wasn't aware of them, but please feel free to point them out and I can add them!

## Testing Instructions

Check out the document preview and make sure it makes sense.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/meta/media_properties/etls.md
+++ b/documentation/meta/media_properties/etls.md
@@ -24,8 +24,10 @@ flowchart TD
     end
 
     subgraph IS [Ingestion Server]
-    C --> ING(Cleanup, filtering, & deleted media removal):::ETL
-    ING --> A[(API DB)]
+    C --> ING(Deleted media removal):::ETL
+    ING --> TT[(Temp Table)]
+    TT --> CF(Cleaning & filtering):::ETL --> TT
+    TT --> A[(API DB)]
     A --> IC2(Index constraints):::ETL --> A
     A --> ESF(Index creation & deleted media filtering):::ETL
     ESF --> ES[(Elasticsearch)]
@@ -44,7 +46,8 @@ flowchart TD
     click M "#mediastore-transformations"
     click IC1 "#index-constraints-catalog"
     click B "#batched-update"
-    click ING "#cleanup-filtering-deleted-media-removal"
+    click ING "#deleted-media-removal"
+    click CF "#cleaning-filtering"
     click IC2 "#index-constraints-api"
     click ESF "#es-index-creation-filtering"
     click DL "#dead-link-filtering"
@@ -93,23 +96,13 @@ Updates against the catalog database may also be initiated using the
 updates may be automated (such as the popularity refresh process) or manual
 (such as data correction).
 
-### Cleanup, filtering, & deleted media removal
-
-During the ingestion server's data refresh process, a series of cleanup and
-filtering steps occur. These steps happen in two places:
-
-1. During the copy from the upstream database into the temporary table within
-   the API database.
-2. After the copy is complete by processing chunks of records in the temporary
-   table and updating values that were changed during the process.
-
-#### During the copy
+### Deleted media removal
 
 Presently, we
 [filter out deleted media from each of the primary media tables](https://github.com/WordPress/openverse/blob/b4ab20cdb220f442949d9c06a99ced5e80c1e1e1/ingestion_server/ingestion_server/queries.py#L173-L182)
 while the `SELECT {columns} FROM {upstream_table}` query is being run.
 
-#### After the copy
+### Cleaning & filtering
 
 We also iterate through the copied data in batches and update the values in the
 new table based on several cleaning/filtering steps (all of which can be seen in
@@ -138,8 +131,10 @@ the catalog database, many of which are for query optimization. These include
 ### ES index creation & filtering
 
 An additional filtering step happens when the records are indexed from the API
-database into Elasticsearch. In addition to the
-[filtering of sensitive terms using two separate indices](/catalog/reference/DAGs.md#create_filtered_media_type_index),
+database into Elasticsearch. In addition to the filtering of
+[sensitive terms using two separate indices](/catalog/reference/DAGs.md#create_filtered_media_type_index)
+and
+[deleted media](https://github.com/WordPress/openverse/blob/b1a1f6eb711a3d6999b2d2bbe4b8cc1d1435c3fa/ingestion_server/ingestion_server/indexer.py#L121),
 there are also transformations that occur when constructing the ES documents.
 The
 [exhaustive list can be seen in the `elasticsearch_models` module](https://github.com/WordPress/openverse/tree/main/ingestion_server/ingestion_server/elasticsearch_models.py),
@@ -170,3 +165,54 @@ which the API may have trouble accessing in order to prevent dead or invalid
 links from being surfaced in API results. The full set of dead link logic can be
 found in the
 [`check_dead_link` module](https://github.com/WordPress/openverse/tree/main/api/api/utils/check_dead_links/__init__.py).
+
+## Proposed
+
+The projects #430, #431, and #3925 all intent to modify the above process. Below
+is the diagram of what this process might look like after all steps are taken.
+Most blocks reference the same sections above, with the exception being
+[Deleted media & tag filtering](#deleted-media-tag-filtering).
+
+```{mermaid}
+flowchart TD
+
+%% Graph definitions
+    subgraph AF [Airflow]
+    P{{Provider}} --> M(MediaStore):::ETL
+    M --> T{{TSV}}
+    T --> IC1(Index constraints):::ETL
+    IC1 --> C[(Catalog DB)]
+    C --> B(Batched update):::ETL --> C
+    C --> ING(Deleted media & tag filtering):::ETL
+    ING --> A[(API DB)]
+    A --> IC2(Index constraints):::ETL --> A
+    A --> ESF(Index creation & deleted media filtering):::ETL
+    ESF --> ES[(Elasticsearch)]
+    end
+
+    A --> DL(Dead link filtering):::ETL
+    DL --> CL[Client]
+    CL --> F[Frontend]
+
+%% Style definitions
+    style AF fill:#00c7d4,margin-left:12em
+    classDef ETL fill:#fa7def
+
+%% Reference definitions
+    click M "#mediastore-transformations"
+    click IC1 "#index-constraints-catalog"
+    click B "#batched-update"
+    click ING "#deleted-media-tag-filtering"
+    click IC2 "#index-constraints-api"
+    click ESF "#es-index-creation-filtering"
+    click DL "#dead-link-filtering"
+```
+
+### Deleted media & tag filtering
+
+The data normalization project (#430) intends to remove the
+[cleanup steps](#cleaning--filtering) from the ingestion server. Subsequent
+ingestion server removal project (#3925) will remove the ingestion server
+entirely. This will be fleshed out more as part of #4456, but the _filtering_
+steps (e.g. demographic tags, tags below a confidence level, etc.) will be
+present within the pipeline prior to the API indexing values into Elasticsearch.
--- a/documentation/meta/media_properties/etls.md
+++ b/documentation/meta/media_properties/etls.md
@@ -1,0 +1,65 @@
+# Extract-Transform-Load flows
+
+The diagrams below show how data flows through Openverse, and the various
+transformation steps that might happen at each stage.
+
+**Guide**:
+
+- **Blue boxes** represent places where data sits at rest.
+- **Purple boxes** represent transformation steps. Each box links to a longer
+  description below.
+
+## Current flow
+
+```{mermaid}
+flowchart TD
+
+%% Graph definitions
+    subgraph AF [Airflow]
+    P{{Provider}} --> M(MediaStore):::ETL
+    M --> T{{TSV}}
+    T --> IC1(Index constraints):::ETL
+    IC1 --> C[(Catalog DB)]
+    C --> B(Batched update):::ETL --> C
+    end
+
+    subgraph IS [Ingestion Server]
+    C --> ING(Cleanup, filtering, & deleted media removal):::ETL
+    ING --> A[(API DB)]
+    A --> IC2(Index constraints):::ETL --> A
+    A --> ESF(Index creation & deleted media filtering):::ETL
+    ESF --> ES[(Elasticsearch)]
+    end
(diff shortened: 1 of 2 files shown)


sarayourfriend:
Excited to spend more time reviewing this!

Tiny error detail I noticed to begin with, but the API pulls from ES, and does dead link filtering on the ES hits (meaning, only has ES's data) + Redis (which is being both read from and written to during this process), not the API DB. The API only pulls the API DB results just before passing to the serializer, after all the `search_controller` routines are finished working on the ES hits. The difference there is that any dead link filtering that happens only has ES and Redis to work off of (currently), and to incorporate the API DB there would require a new API DB query for each ES query. Right now, there's only ever one single API DB request, and `1 ... n` ES queries (limited by the max iteration depth of `search_controller._post_process_results`).

The change to the graph would be to say, the API pulls from ES, filters and processes the ES results, and then only pulls the final presented subset from the API DB. So `ES + Redis -> dead link filtering (filter) -> API DB -> API serializers (transform) -> requester`, and then _if_ the frontend is the requester, the frontend also does some additional processing (like generating attributions), but those are not a _necessary_ (as in, constant) part of the data lifecycle. It might be self-evident, but worth keeping in mind that the frontend is not the only API consumer, and transformations created at the frontend level are not getting to anyone else. That's fine if they're just for the frontend to use, but not fine if those transformations could also matter to other clients (like the WP media inserter, or Openverse Slack Reaction app).

There's a cyclical relationship between the API DB and ES in that one informs the way the other is used, either to decide what data to load into ES to begin with, but then also what results to present from the API DB, with the API DB results being those that are presented, not the ES hits.

Additionally, it would be good to add the fact that Redis is both written to and read from during dead link filtering (which is basically itself an ETL process, I guess, we're at the very least _generating_ data, if not extracting it, though we do save the specific status code, not just a dead/not-dead boolean).

Really awesome, though, very excited to dig into this more, and to get a more cohesive view of how each part of our data lifecycle works with the other parts.

documentation/meta/media_properties/etls.md
```diff
@@ -0,0 +1,218 @@
...
+When records are retrieved from Elasticsearch, they also go through a process
+called "dead link filtering". This operation attempts to remove links to media
+which the API may have trouble accessing in order to prevent dead or invalid
+links from being surfaced in API results. The full set of dead link logic can be
+found in the
+[`check_dead_link` module](https://github.com/WordPress/openverse/tree/main/api/api/utils/check_dead_links/__init__.py).
+
+## Proposed
+
+The projects #430, #431, and #3925 all intent to modify the above process. Below
```
sarayourfriend, line 171:
```suggestion
The projects #430, #431, and #3925 all intend to modify the above process. Below
```

sarayourfriend (changes requested):
Okay! This LGMT aside from the issue I commented on before (the slight inaccuracy of leaving Redis out of the API picture, and clarifying where results come from, + adding serialization as a pre-client transformation step). Some of that is being a bit nit-picky, but I think it's worth especially clarifying _what_ data is sent in API responses, including what transformations the API itself may do to them in serialization. No need to outline the specifics of those transformations here. As you've done with the other sections, just mentioning that it happens with a link to the relevant code (serializers) it is more than enough and will prevent documentation drift :+1: 

With documentation drift in mind, though, I suppose we'll need to update this document once the three mentioned projects are complete!

Otherwise, this is great. It does not answer every question I have, but it's also purely descriptive, which I think is good for now. I personally would really like a _prescriptive_ version of this document in the future, a "vision" of the data lifecycle, but I think that will come with time as we all discuss and refine what our goal is with the data lifecycle and each part of it.

Staci and I had an interesting conversation earlier this week about the difference between approaching problems in Openverse (generally, not just with data eng. stuff), with a "keep the changes minimal" mindset. While that has some benefits (e.g., it makes some things go faster), it has the effect of us treating the way things are as the way they should be, and eliminating chances for us to talk about what we _want_ (our vision) for these systems in concrete terms. That, in turn, stops us from making moves towards that vision. It also, I believe, introduces significant opportunities for miscommunication and misalignment. It is probably the case that all of us have _some_ ideas of how things _should_ be (and each of us to varying degrees for each area depending on our areas of focus in Openverse), but because we very rarely have that clearly defined, we end up talking past each other, or getting confused about one choice or another. I often feel like we (meaning the whole team as individuals) end up ever so slightly (or sometimes even more) misaligned in our visions, leading to spinning wheels in some places. Here, for example, we see that there's a convergence of three understandings of the intention of each piece of our data lifecycle that were planned independently of each other, with only really offhand mentions of the others in planning, rather than true integration. Ideally, these three projects would all work together to move our data engineering and data lifecycle towards our shared vision. But that presupposed the shared vision is defined, and indeed shared between us all!

Anyway, just musing on the fact that this document is a step in that direction, and I'm really glad for that. The conclusion Staci and I came to with respect to how we've planned projects to aim for minimal changes is that if we had a grander vision overall, we could make decisions in project planning that set us up for better situations in the future. Especially with the ingestion server removal project, we're pulling things more inline with treating Airflow as _the_ coordinator of movement in our data infrastructure. Using the EC2 workers as workers is standard practice (I've learned this recently), but something like the ingestion server never was! It was doing Airflow's job, and way more tediously. Moving towards more Airflow-y way of coordinating our Catalog -> API DB -> ES ETL was not an explicit intention (we instead wanted to simplify data refresh to make it easier to work on, without ever explicitly saying that the reason it would be simpler is because it would be more in line with how Airflow should be used to coordinate these things!), but is a significant and meaningful realignment and readjustment of how the ETL pipelines are thought about. That's excellent, because it means we are making prescriptive changes towards a vision of "how things should be for Openverse", by saying "Airflow is the place for this".

That's powerful! I'm excited for the possibilities that brings up, especially with respect to other things we can use Airflow for aside from data engineering task management. My view of Airflow is personally expanded by learning more about it and speaking with Staci this week.

TL;DR: This is great, looking forward to more discussions about these things, and I'm very excited to learn more about all of this before and through those discussions with y'all. Always exciting to see what we come up with when we give ourselves permission to think beyond the minimal constraints of a specific single need.

zackkrida (approved):
@AetherUnbound this is so awesome! The structure of the document itself is lovely, with the graphic linking off to the the descriptive sections. :heavy_plus_sign: to Sara's suggestions but I'll approve now to streamline things.

documentation/meta/media_properties/etls.md
```diff
@@ -0,0 +1,218 @@
...
+    ING --> TT[(Temp Table)]
+    TT --> CF(Cleaning & filtering):::ETL --> TT
+    TT --> A[(API DB)]
+    A --> IC2(Index constraints):::ETL --> A
+    A --> ESF(Index creation & deleted media filtering):::ETL
+    ESF --> ES[(Elasticsearch)]
+    end
+
+    A --> DL(Dead link filtering):::ETL
+    DL --> CL[Client]
```
zackkrida, line 37:
```
    DL --> CL[Client]
```

Obviously from a computing and data flow standpoint this name makes sense, but we don't refer to the API as a "client" anywhere else, really. Is there a way we could expand this (maybe a parenthetical including "Django API") so that others can "connect the dots" between this graphic and our stack as understood by developers?

AetherUnbound:
> Especially with the ingestion server removal project, we're pulling things more inline with treating Airflow as _the_ coordinator of movement in our data infrastructure. Using the EC2 workers as workers is standard practice (I've learned this recently), but something like the ingestion server never was! It was doing Airflow's job, and way more tediously. Moving towards more Airflow-y way of coordinating our Catalog -> API DB -> ES ETL was not an explicit intention (we instead wanted to simplify data refresh to make it easier to work on, without ever explicitly saying that the reason it would be simpler is because it would be more in line with how Airflow should be used to coordinate these things!), but is a significant and meaningful realignment and readjustment of how the ETL pipelines are thought about. That's excellent, because it means we are making prescriptive changes towards a vision of "how things should be for Openverse", by saying "Airflow is the place for this".

It's interesting to hear you say this. I feel like I've been trying to communicate this point and push for exactly this way of thinking about Airflow for a while 🙂 It was part of the reason I wanted to see if we could use the `ECSRunTaskOperator` or something similar for the actual reindexing step rather than keep the Falcon server/indexer workers that we had. Airflow is _built_ for this kind of coordination and centralized management, and it's definitely a vision I've been working to communicate.

What I'm hearing, though, is that it might have been useful to make that intention explicit in some form of policy or "vision" document. That might have helped make it clearer to communicate "this is what we want to use Airflow for" rather than me saying it over and over during the occasions where an opportunity to leverage it cropped up organically. That's helpful feedback, and hopefully that can also be part of the things we make explicit:
- The catalog database is **the** data warehouse where we store as much information as we can/is reasonable.
- Airflow is **the** orchestrator for our data management and automation workflows, particularly those that interact with multiple systems or aspects of our stack.

documentation/meta/media_properties/etls.md
AetherUnbound: Ah, see I meant for "Client" here to be someone _using_ the API (which could be the Gutenberg integration, a WP plugin, etc). I'll try to clarify that!

AetherUnbound: I actually don't have the Django API on here anywhere, and I think I should! I'll expand on it a bit more 😄

documentation/meta/media_properties/etls.md
```diff
@@ -0,0 +1,303 @@
...
+    C --> ING
+    end
+
+    %% Connections for the data refresh
+    ESF --> ES
+    ING --> A
+    A --> IC2 --> A
+    A --> ESF
+
+    subgraph DJA [DjangoAPI]
```
zackkrida, line 257:
```suggestion
    subgraph DJA [Django API]
```

zackkrida:
@AetherUnbound the changes look great :+1:

documentation/meta/media_properties/etls.md
AetherUnbound: For some reason I remember this causing an error...but I just checked and it's fine! I'll fix the other reference too 🚀

sarayourfriend:
> Airflow is the orchestrator for our data management and automation workflows, particularly those that interact with multiple systems or aspects of our stack.

I think the specific thing that didn't connect to me is that Airflow isn't even a data engineering tool, per se. It's _just_ a task scheduler and you can do _anything_ with it. The data engineering happens with _other_ tools, like Postgres, or Python scripts. It didn't click to me until I read about data engineering tools generally and realised that Airflow doesn't have much to do with it aside from being a task coordinator.

With respect to the ingestion server removal, I did _not_ understand the difference. I still don't think ECS is a good option (FARGATE's pricing structure is actively hostile to resource cost/utilisation optimisations and wildly expensive), but AWS Batch with spot instances would be a tremendous change and I'd support going in that direction.

Yeah, for whatever reason it never clicked that Airflow is just a task coordinator :confused:. Sorry! I haven't been that involved in data things until the ingestion server removal project, so I'm not sure I encountered that idea before, and when reviewing that IP I did not understand the idea of removing the API from the indexer workers. We can still do that if we want, and swap out the ASG for AWS Batch, even without spot instances, and it would be an improvement and move towards a more airflow-y model for sure. I think I was also specifically hung up on ECS FARGATE, which as I said above I think is a specifically bad solution, but it's possible to structure the containerised workload such that the execution environment doesn't matter, so swapping ECS for kubernetes or AWS Batch becomes more or less trivial. I apologise for not understanding that, something just didn't click until I learned more about data engineering as a whole and Airflow in particular in the past two weeks.

documentation/meta/media_properties/etls.md
```diff
@@ -0,0 +1,303 @@
...
+    C --> ING
+    ING --> TT[(Temp Table)]
+    TT --> CF(Cleaning & filtering):::ETL --> TT
+    end
+
+    TT --> A
+    ESF --> ES
+    A --> IC2 --> A
+    A --> ESF
+    subgraph DJA [DjangoAPI]
```
AetherUnbound, line 45:
```suggestion
    subgraph DJA [Django API]
```

sarayourfriend (approved):
LGTM!

stacimc (approved):
This looks _great_ and is so nice to have! The only thing that struck me as potentially confusing is placing the `temp table` within the ingestion server visually, when this is really a temp table in the API database that later gets promoted. Not sure how best to reflect that via the diagram but maybe we could add a section explaining what the temp table is, and have it be clickable like the other elements? Not blocking.

This is fantastic @AetherUnbound :)

AetherUnbound:
> The only thing that struck me as potentially confusing is placing the `temp table` within the ingestion server visually, when this is really a temp table in the API database that later gets promoted. Not sure how best to reflect that via the diagram but maybe we could add a section explaining what the temp table is, and have it be clickable like the other elements?

This was definitely a tough thing to convey 😅 we already had enough connections to the API database that I wanted to try and have it be separate. I'll add a note for it at the bottom of the diagram, thanks!

sarayourfriend:
Not sure where or how to document this, but I just remembered the provider occurrance tallying we aggregate on a weekly basis into Redis: https://github.com/WordPress/openverse/blob/0891970b5252a00630f3cb971f1121264d55b1af/api/api/utils/tallies.py#L26

It doesn't fit into and current ETL wants (like dead link detection does) but it strikes me that it could be useful to have a place where we've described the data we're generating overall, regardless of whether it's currently part of or intended to be part of an entire ETL pipeline. Provider occurrance rate might be a useful relevancy metric, for example, and it'd be a shame if we forgot about it because we aren't referring to it regularly (that I know of).

Just mentioning in this PR because of it's tangential relevancy to the subject, not because I think it should be included in this documentation. I needed to write it down _somewhere_ and this was the easiest and most sensible place in the immediate.

PR #4471, opened 2024-06-11
Remove single quotes in values of Ingestion Server's TSV files

### LINKED ISSUE #3912: Upload Ingestion Server's TSV files to AWS S3

krysal:
## Fixes

Related to #3912 by @krysal

## Description

This PR fixes several issues on the Ingestion Server and prepares it to upload the files.
- Saves files of cleaned data to a specific temporary directory locally before being send to S3 (this was an @AetherUnbound's previous suggestion)
- Recreates the temp directory before cleaning to avoid mixing values between different data refresh processes
- **Removes single quotes on cleaned values**. Otherwise, the file has URLs surrounded by `'`, which isn't necessary and makes it quite complicated to save the data in a database table later
- Adds stocksnap to TLS_CACHE manually (for some reason, it was being registered as TLS not supported, which is not true)

## Testing Instructions

1. Spin up the services
```
just a && just c
```

2. Make some rows of the image table in the catalog "dirty" by removing the protocol in one or more of the URLs fields (url, creator_url, or foreign_landing_url)

3. Run an image data refresh
```
# Optionally, delete the image index before to avoid the error of "index already exists"
just docker/es/delete-index image-init

just ingestion_server/ingest-upstream "image" "init"
```

4. Check the files are in the container, and check for their content:

```
just exec ingestion_server bash
```

```
ls /tmp/cleaned_data/
cat /tmp/cleaned_data/url.tsv
```

and that the content of each file is the identifier and values without quotes, eg.:

```
0041b4ec-f55e-4cd0-a491-84f41be72232	https://cdn.stocksnap.io/img-thumbs/960w/YUJXHHKSMI.jpg
00252271-e1dc-4faf-b840-7fcc386f2529	https://cdn.stocksnap.io/img-thumbs/960w/7HHEXNL6AQ.jpg
```
While in our bucket, you can find the manually uploaded files containing rows like the following:
```
4e406052-5d90-4c30-87b6-6f1b6ba533e7	'http://musee-mccord.qc.ca/ObjView/MP-1986.85.4.jpg'
3eba813d-6e24-4aa3-9fdb-9242d4bdfc4e	'https://flora-on.pt/Carlina-vulgaris_ori_v6Wn.jpg'
```

## Checklist

- [N/A] I ran the DAG documentation generator (`just catalog/generate-docs` for catalog PRs) or the media properties generator (`just catalog/generate-docs media-props` for the catalog or `just api/generate-docs` for the API) where applicable.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -64,6 +64,8 @@ TLS_CACHE = {
     "www.eol.org": True,
     ".digitaltmuseum.org": True,
     "collections.musee-mccord.qc.ca": False,
+    ".stocksnap.io": True,
+    "cdn.stocksnap.io": True,
 }
 
 TMP_DIR = pathlib.Path("/tmp/cleaned_data").resolve()
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -8,6 +8,7 @@ import csv
 import logging as log
 import multiprocessing
 import pathlib
+import shutil
 import time
 import uuid
 from urllib.parse import urlparse
@@ -307,8 +308,9 @@ def clean_image_data(table):
     :return: None
     """
 
-    # Create directory to store cleaned data temporarily
-    TMP_DIR.mkdir(parents=True, exist_ok=True)
+    # Recreate directory where cleaned data is stored
+    shutil.rmtree(TMP_DIR, ignore_errors=True)
+    TMP_DIR.mkdir(parents=True)
 
     # Map each table to the fields that need to be cleaned up. Then, map each
     # field to its cleanup function.
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -7,6 +7,7 @@ This includes cleaning up malformed URLs and filtering out undesirable tags.
 import csv
 import logging as log
 import multiprocessing
+import pathlib
 import time
 import uuid
 from urllib.parse import urlparse
@@ -64,6 +65,8 @@ TLS_CACHE = {
     "collections.musee-mccord.qc.ca": False,
 }
 
+TMP_DIR = pathlib.Path("/tmp/cleaned_data").resolve()
+
 
 def _tag_denylisted(tag):
     """Check if a tag is banned or contains a banned substring."""
@@ -106,9 +109,9 @@ class CleanupFunctions:
                 log.debug(f"Tested domain {_tld}")
 
             if tls_supported:
-                return f"'https://{url}'"
+                return f"https://{url}"
             else:
-                return f"'http://{url}'"
+                return f"http://{url}"
         else:
             return None
 
@@ -141,6 +144,7 @@ class CleanupFunctions:
 
         if update_required:
             fragment = Json(tag_output)
+            log.debug(f"Tags fragment: {fragment}")
             return fragment
         else:
             return None
@@ -200,7 +204,7 @@ class TlsTest:
             https = url.replace("http://", "https://")
             try:
                 res = re.get(https, timeout=2)
-                log.info(f"{https}:{res.status_code}")
+                log.info(f"tls_test - {https}:{res.status_code}")
                 return 200 <= res.status_code < 400
             except re.RequestException:
                 return False
@@ -243,23 +247,27 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
             if clean:
                 cleaned_data[update_field] = clean
                 log.debug(
-                    f"Updated {update_field} for {identifier} "
-                    f"from '{dirty_value}' to '{clean}'"
+                    f"Updated {update_field} for {identifier}\n\t"
+                    f"from '{dirty_value}' \n\tto '{clean}'"
                 )
         # Generate SQL update for all the fields we just cleaned
         update_field_expressions = []
         for field, clean_value in cleaned_data.items():
-            update_field_expressions.append(f"{field} = {clean_value}")
-            # Save cleaned values for later
-            # (except for tags, which take up too much space)
             if field == "tags":
+                # The `clean_value` for tags already includes the single quotes,
+                # so it's not necessary to add them, and they're omitted in
+                # `cleaned_values` to save in files later because they take up
+                # too much disk space.
+                update_field_expressions.append(f"{field} = {clean_value}")
                 continue
+            update_field_expressions.append(f"{field} = '{clean_value}'")
             cleaned_values[field].append((identifier, clean_value))
 
         if len(update_field_expressions) > 0:
             update_query = f"""UPDATE {temp_table} SET
             {', '.join(update_field_expressions)} WHERE id = {_id}
             """
+            log.debug(f"Executing update query: \n\t{update_query}")
             write_cur.execute(update_query)
     log.info(f"TLS cache: {TLS_CACHE}")
     log.info("Worker committing changes...")
@@ -273,18 +281,17 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
 
 
 def save_cleaned_data(result: dict) -> dict[str, int]:
-    log.info("Saving cleaned data...")
     start_time = time.perf_counter()
 
     cleanup_counts = {field: len(items) for field, items in result.items()}
     for field, cleaned_items in result.items():
         # Skip the tag field because the file is too large and fills up the disk
-        if field == "tag":
+        if field == "tag" or not cleaned_items:
             continue
-        if cleaned_items:
-            with open(f"{field}.tsv", "a") as f:
-                csv_writer = csv.writer(f, delimiter="\t")
-                csv_writer.writerows(cleaned_items)
+
+        with open(TMP_DIR.joinpath(f"{field}.tsv"), "a", encoding="utf-8") as f:
+            csv_writer = csv.writer(f, delimiter="\t")
+            csv_writer.writerows(cleaned_items)
 
     end_time = time.perf_counter()
     total_time = end_time - start_time
@@ -299,6 +306,9 @@ def clean_image_data(table):
     :param table: The staging table for the new data
     :return: None
     """
+
+    # Create directory to store cleaned data temporarily
+    TMP_DIR.mkdir(parents=True, exist_ok=True)
 
     # Map each table to the fields that need to be cleaned up. Then, map each
     # field to its cleanup function.
--- a/ingestion_server/test/unit_tests/test_cleanup.py
+++ b/ingestion_server/test/unit_tests/test_cleanup.py
@@ -47,12 +47,12 @@ class TestCleanup:
         tls_support_cache = {}
         pook.get("https://flickr.com").reply(200)
         result = CleanupFunctions.cleanup_url(bad_url, tls_support_cache)
-        expected = "'https://flickr.com'"
+        expected = "https://flickr.com"
 
         bad_http = "neverssl.com"
         pook.get("https://neverssl.com").reply(500)
         result_http = CleanupFunctions.cleanup_url(bad_http, tls_support_cache)
-        expected_http = "'http://neverssl.com'"
+        expected_http = "http://neverssl.com"
         assert result == expected
         assert result_http == expected_http
 


ingestion_server/ingestion_server/cleanup.py
```diff
@@ -300,6 +308,10 @@ def clean_image_data(table):
     :return: None
     """
 
+    # Recreate directory where cleaned data is stored
+    shutil.rmtree(TMP_DIR)
```
sarayourfriend, line 312: TIL the difference between `Path::rmdir` and `shutil.rmtree`. The former can only remove empty directories (the bit I didn't know!).

krysal: It's a striking difference but I guess internally it should make sense to keep functions with unique responsibilities.

sarayourfriend: For sure!

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -273,25 +284,63 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+        return boto3.resource(
+            "s3",
+            endpoint_url=config("AWS_S3_ENDPOINT", default="http://s3:5000"),
+            aws_access_key_id=config("AWS_ACCESS_KEY_ID", default="test_key"),
+            aws_secret_access_key=config(
+                "AWS_SECRET_ACCESS_KEY", default="test_secret"
+            ),
+        )
+
+    return boto3.resource("s3", region_name=config("AWS_REGION", default="us-east-1"))
```
stacimc, line 316: Can you add some comments indicating why these connections are handled differently? I'm not sure I understand the nuance here.

stacimc (commented):
I'm able to see the rows I updated getting cleaned, and all the expected logs in the ingestion server locally -- but I don't see the actual file in MinIO 🤔 No errors of any kind. Any ideas what could be going wrong?

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -243,23 +252,25 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
         # Generate SQL update for all the fields we just cleaned
         update_field_expressions = []
         for field, clean_value in cleaned_data.items():
-            update_field_expressions.append(f"{field} = {clean_value}")
-            # Save cleaned values for later
-            # (except for tags, which take up too much space)
             if field == "tags":
+                update_field_expressions.append(f"{field} = {clean_value}")
                 continue
+            update_field_expressions.append(f"{field} = '{clean_value}'")
```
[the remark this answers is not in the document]
krysal: The URL fields had single quotes added twice. As you can see above in the file (lines 117 and 119), I removed the quotes from the cleaned value so they won't appear in the file either (single quotes as the quoting character are problematic for later loading in the DB).

ingestion_server/ingestion_server/cleanup.py
krysal: Sure, I added a comment. Let me know if more is necessary.

In live environments, the connection is allowed via the IAM role policy of the ingestion server instances. You can see that configuration in the following link (only available to maintainers): 

https://github.com/WordPress/openverse-infrastructure/blob/a3c24aefea3c73dcc7b1198a1e3b1c845c4a732b/modules/services/ingestion-server/iam.tf#L57-L78

krysal:
@stacimc That sounds like, potentially, you don't have the `openverse-catalog` bucket that was recently [changed as the default](https://github.com/WordPress/openverse/commit/7c6699d3fa5e2737fa658a94cab79708c2b96a5d) for the catalog. That is not good because it should have raised an error and been logged. I realized my way of checking for the bucket's existence wasn't working. Now I updated and confirmed it checks for the bucket.

You can manually create the bucket through MinIO, or running a `just recreate` should work. Could try again? If the bucket exists and you see no files please share what is in the logs to get any hint.

ingestion_server/ingestion_server/cleanup.py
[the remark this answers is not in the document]
sarayourfriend: I'm also confused about this :confused: 

Where were the duplicate quotes coming from, exactly? And why does the `clean_value` for tags not need to be quoted but everything else does?

Maybe avoiding the continue would help the clarity of this block:

```py
for field, clean_value in cleaned_data.items():
    if field != "tags":
        # Save cleaned values for later
        # (except for tags, which take up too much space)
        cleaned_values[field].append((identifier, clean_value))

    update_field_expressions.append(f"{field} = {clean_value}")
```

Never mind that I don't understand where the difference is coming from for the format of the string added to `update_field_expressions`, this version is a lot easier to understand, to me.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -273,25 +284,68 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
         # Skip the tag field because the file is too large and fills up the disk
-        if field == "tag":
+        if field == "tag" or not cleaned_items:
             continue
-        if cleaned_items:
-            with open(f"{field}.tsv", "a") as f:
-                csv_writer = csv.writer(f, delimiter="\t")
-                csv_writer.writerows(cleaned_items)
+
+        with open(TMP_DIR.joinpath(f"{field}.tsv"), "a", encoding="utf-8") as f:
```
sarayourfriend, line 295: Kind of just a nit, but I usually see usage of the `/` operator to join path segments, rather than `joinpath`, although we seem to have a mix of both in our project (maybe more `/` versions, though?). Anyway, just a nit/nothing.

```suggestion
        with open(TMP_DIR / f"{field}.tsv", "a", encoding="utf-8") as f:
```

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -273,25 +284,68 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
     total_time = end_time - start_time
     log.info(f"Finished saving cleaned data in {total_time:.3f},\n{cleanup_counts}")
     return cleanup_counts
 
 
+def _get_s3_resource():
+    """
+    Locally, it connects to a MinIO instance through its endpoint and test credentials.
+    On live environments, the connection is allowed via IAM roles.
+    """
```
sarayourfriend, line 309: Suggestion to clarify this comment. I wasn't sure what "it" referred to (might have both been MinIO and boto3?), and the reason for the difference between local and live wasn't clear (like, why MinIO matters).

```suggestion
    """
    Retrieve a correctly configured boto3 S3 resource.
    
    Locally, we use MinIO to emulate S3, so we must specify the endpoint and credentials.
    On live environments, S3 itself is used, and authentication works via the instance profile.
    """
```

Although, overall, I do think this would be clearer if we didn't use default values here, and instead required setting them, and allowed them to be `None` when not configured in the environment. Then the code wouldn't need to worry about the implementation detail _at all_. I'll share a suggestion for that code in my next comment.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -273,25 +284,68 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+        return boto3.resource(
+            "s3",
+            endpoint_url=config("AWS_S3_ENDPOINT", default="http://s3:5000"),
+            aws_access_key_id=config("AWS_ACCESS_KEY_ID", default="test_key"),
+            aws_secret_access_key=config(
+                "AWS_SECRET_ACCESS_KEY", default="test_secret"
+            ),
+        )
+
+    return boto3.resource("s3", region_name=config("AWS_REGION", default="us-east-1"))
```
sarayourfriend, line 320:
```suggestion
    if config("ENVIRONMENT", default="local") == "local":
        return boto3.resource(
            "s3",
            endpoint_url=config("AWS_S3_ENDPOINT", default="http://s3:5000"),
            aws_access_key_id=config("AWS_ACCESS_KEY_ID", default="test_key"),
            aws_secret_access_key=config(
                "AWS_SECRET_ACCESS_KEY", default="test_secret"
            ),
        )

    return boto3.resource(
        "s3",
        endpoint_url=config("AWS_S3_ENDPOINT", default=None),
        aws_access_key_id=config("AWS_ACCESS_KEY_ID", default=None),
        aws_secret_access_key=config("AWS_SECRET_ACCESS_KEY", default=None),
        region_name=config("AWS_REGION", default=None),
    )
```

All these kwargs are defined with None as the default in the implementation, so we can safely explicitly pass null (if these variables aren't defined in the environment), and they'll just be ignored.

https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/session.html#Session.resource

That way the code doesn't have to worry about implementation details of local/live environments, and we get rid of this confusion altogether...

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -273,25 +284,68 @@ def _clean_data_worker(rows, temp_table, sources_config, all_fields: list[str]):
...
+        bucket = s3.Bucket(bucket_name)
+        log.info(f"Connected to S3 and '{bucket_name}' bucket loaded.")
+    except Exception as e:
+        log.error(f"Files upload failed. Error connecting to S3.\n{e}")
+        return
+
+    for field in fields:
+        file_path = TMP_DIR.joinpath(f"{field}.tsv")
+        if not file_path.exists():
+            continue
```
sarayourfriend, line 339: In what case is this file missing not an error condition (in other words, why can we just `continue` here rather than raise an exception)? The intention for this loop would be a lot safer and more explicit if instead of looping through `fields`, we only looped through the specific ones we want to upload.

sarayourfriend (changes requested):
Requesting changes to simplify the `get_s3_resource` function (or even just remove it), clarify further the changes regarding `update_field_expressions`, and clarify the loop in `_upload_to_s3` (specifically the case where the file does not exist).

ingestion_server/ingestion_server/cleanup.py
krysal: I took the idea from your recent changes in managing the files for waveform. I like how explicit it is in the `with` line. 

https://github.com/WordPress/openverse/blob/d562f4719aefdf7d90c13ca6738a5e8c27ce352f/api/api/utils/waveform.py#L79

I found the form with ` / ` handy for longer paths in a separate line.

ingestion_server/ingestion_server/cleanup.py
krysal: Once we start fixing the rows, we will reach the point where there is nothing to clean, so there is no file to upload in that case, and the `tags` field won't have a file here either.

ingestion_server/ingestion_server/cleanup.py
krysal: That's a good observation! Thank you. I agree that we can simplify this a lot, though it'd require setting these environment variables locally for the ingestion server to work.

ingestion_server/ingestion_server/cleanup.py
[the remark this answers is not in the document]
krysal: > Where were the duplicate quotes coming from, exactly?

The cleaned values have quotes added in the return of the cleaning function:
https://github.com/WordPress/openverse/blob/d562f4719aefdf7d90c13ca6738a5e8c27ce352f/ingestion_server/ingestion_server/cleanup.py#L108-L111

>  And why does the clean_value for tags not need to be quoted but everything else does?

The `psycopg2.extras.Json` function for tags already adapts a Python object to :sql:`json` data type, so they don't need extra quotes wrapping.

The difference in the string added to `update_field_expressions` is the quotes. I assume this was previously done that way to avoid this confusion, but it's necessary to have the cleaned values without these quotes in the file so they don't interfere with the copy upload to a DB table later.

An alternative would be to perform more string manipulation to remove the quotes before saving the rows in files, but this seemed unnecessary and more prone to error to me.

ingestion_server/ingestion_server/cleanup.py
krysal: I ended up removing the function per Sara's suggestion so it should be simpler now.

ingestion_server/ingestion_server/cleanup.py
krysal: Minio is what we use to emulate and test some AWS operations. In this case, it's the replacement for S3. Otherwise, we would need to use some test live resources, which would complicate things a lot.

ingestion_server/ingestion_server/cleanup.py
krysal: Moved to #4529.

krysal:
@stacimc @sarayourfriend I split the PR into two parts since many things were happening here and it's easier to explain by simplifying it. Here, I left the fix for the quotes in the TSV files, updated the testing instructions, and opened #4529 for the upload to S3. I hope I have answered all your questions. Let me know if anything needs further clarification.

stacimc (changes requested):
@krysal I'm not sure what's happening, but I'm getting really strange behavior when testing this. Locally in my catalog I updated three different records, removing the protocol from the `url` on one, the `creator_url` on the second, and the `foreign_landing_url` on the third. When I ran the data refresh only the `foreign_landing_url` actually got cleaned, ~and I still am not seeing the file in minio even though I see no errors in my ingestion server logs~. Edit: I see that you moved the s3 uploading out of this PR so that part makes sense.

I didn't have much time to look further into this, but is anyone else able to replicate my results?

ingestion_server/ingestion_server/cleanup.py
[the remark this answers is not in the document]
sarayourfriend: I agree with Staci, but I also don't want to block on it because of this code being deleted soon anyway :shrug:

sarayourfriend (approved):
I wasn't able to replicate Staci's results. I modified an image in the catalog to remove the scheme from all three URLs, and all three have separate files correctly adding the scheme.

I think there are ways to make the intention of the code clearer, but as I said, I am unwilling to block the PR given the code will be gone soon anyway. That said, I would be more firm in requesting clarification comments or reorganisation of that loop if this code was long term, so if similar code will be introduced elsewhere, I would block on its clarification. Just wanted to clarify it's a matter of the circumstance, not that I would encourage a pattern of ignoring clarity in reviews.

ingestion_server/ingestion_server/cleanup.py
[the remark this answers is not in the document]
krysal: I updated the comment. If it's still not clear, it's open to suggestions. I didn't think it would bring so much confusion 😅

krysal:
@stacimc Did you try doing a full `just recreate` before all the steps? ... @dhruvkb are you able to test and review the PR?

---

@sarayourfriend unrelated but I was about to suggest using `./ov just recreate` but it didn't work for me. I got the following:
```
➜ ./ov just recreate
Unable to find image 'openverse-dev_env:latest' locally
docker: Error response from daemon: pull access denied for openverse-dev_env, repository does not exist or may require 'docker login'.
See 'docker run --help'.
```
The plain `just recreate` worked.

dhruvkb (approved):
The code changes make sense based on the PR description. I followed the testing instructions and verified the correct output in `/tmp`.

sarayourfriend:
@krysal that's because of @dhruvkb's PR to https://github.com/WordPress/openverse/pull/4526.

stacimc:
@krysal I did recreate, but if three other people are unable to reproduce I'm willing to believe something was going awry in my environment. No objections from me 👍

PR #4484, opened 2024-06-13
Make it easier to run Openverse from an offline media

### LINKED ISSUE #4478: Make repo installable from offline media
## Description

Openverse should be easy to install from offline media like a flash drive. This will make it extremely easy to set up Openverse for contributors by carrying the repository on a flash drive to various coding events. This requires multiple things:

The repo should be copyable onto a flash drive that can support multiple Unix-like operating systems (with support for file system permissions etc.).

We should have `just` recipes to
- download the Docker images offline
- load the Docker images from offline `.tar.gz` files

We should have commands and `just` recipes to ensure that the repo can be cleaned of all untracked files like environment variables and secrets.

## Additional context

This issue was identified at contributor day.

dhruvkb:
## Fixes

Fixes #4478 by @dhruvkb

## Description

This PR makes Openverse easier to install from an offline media. It does so by providing recipes that can

- clean all env files from the repo to make sure that are no more secrets:  `just unenv`
- dump all Docker images needed for Openverse into `.tar.gz` files: `just docker/dump-3p && just docker/dump-lint && just docker/dump-ov`
- load all Docker images from the images directory: `just docker/load`.

## Testing Instructions

0. Dump all Docker images using the command above.
1. Clear your Docker images using `docker image prune -a`.
2. Load the images back using the command above.
3. Run Openverse, it should not need to download anything from the Internet.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
@dhruvkb Is this paused/blocked into something? What is it missing to be ready?

dhruvkb:
@krysal it's moving pretty slowly. I'd like to keep it open for a couple more weeks, if that's okay.

dhruvkb:
I have marked this as ready for review but one of the issues with this is I cannot test it between two computer. There isn't a USB file format that'll natively support both Mac and Linux, which are the two computers I have.

If anyone has two Macs or two Linux computers, please create a USB drive (with APFS or ext4 file system respectively) and try it out.

sarayourfriend:
> There isn't a USB file format that'll natively support both Mac and Linux, which are the two computers I have.

exFAT should be the one to use here, what prevents it from being used?

zackkrida:
Yes, I personally have two usb drives formatted exFAT that I use to transfer files between multiple machines. I'll test this!

dhruvkb:
exFAT does not support UNIX-like permissions so I've seen in the past that copying files over it can cause permission issues. So it's probably better to use a format that's native to the OS.

sarayourfriend:
> exFAT does not support UNIX-like permissions so I've seen in the past that copying files over it can cause permission issues

This seems like an error in using exFAT? Are there file permissions that need to be maintained from the machine that generated the images onto the new machine? That seems _highly_ unlikely considering these are images meant to be shared?

KDE and GNOME both mount exFAT devices with the user and group options set to the logged in user, so I'm not sure when this would be relevant unless mounting the device via `mount` directly, in which cause surely we can assume the correct options will be known by the person mounting the device?

Are you suggesting that we would need to have at least two USBs, potentially three (if we allow for someone arriving relying on WSL), all with different filesystems native to the OS being used?

dhruvkb:
We have determined that this is not a complete solution because of compatibility issues affecting Docker images, compiled Python and Node packages and USB drive file systems. A comprehensive solution would be overengineered, unnecessary and unused except for 2-3 days a year.

sarayourfriend:
We also talked about looking into github.dev again, and seeing if we can make it work, perhaps building on the foundation of `ov` to create a Dockerised development environment.

Issue #4490, opened 2024-06-14
Refactor `ov` to create a persistent container


sarayourfriend:
## Problem

`ov` currently creates a new container each time it runs (and automatically removes the previous container). That mostly works, but we could get a lot more out of `ov` if we treated it more like a persistent virtual machine than an immutable one-off environment. Crucially, it also means that if you run `ov <some long running command>` in one terminal, and `ov bash` in another, you're looking at two distinct and separate containers. Theoretically most state should be maintained in bound directories _but that might not be the case_, for example, with `tmp` files. This can lead to some confusing behaviour, and makes the development environment more tedious to debug inside of.

Additionally, it makes it impossible to, for example, defer installing certain system dependencies only when needed. That would be useful in the case of Playwright, which requires a lot of host dependencies to run browsers, that shouldn't be installed in the base container.

## Description

To accomplish this, we need the container to stay running, and then use `exec` to run commands inside the container. This could actually allow us to significantly simplify how we handle certain situations, and would speed up repeated script execution in almost all cases.

To do this, implement the following:

1. Change the container's entrypoint to an infinite while loop that calls sleep 1 in each iteration
2. Move the old entrypoint script to `docker/dev_env/setup_env.sh` and add an entry to `ov` for `setup-env` to run that script
3. Update `docker/dev_env/run.sh` to do the following:
  1. Check if the container exists (even if it is stopped), If it does, and is stopped, start it. Then (after it is started), use `docker exec` to run the command in the container. You'll still need to set up some of the same args as `run`, so extract those shared args into a separate `shared_args` array
  2. If the container does not exist yet, use `docker run` to create and start it with the current `run_args` variables. Make sure to pass `-d` as well, so that the command detaches, as the entrypoint is now the infinite loop + sleep script. Then, defer to the same `docker exec` method used for the other branch.

## Alternatives

We could promote the idea of running `ov bash` to start a "session", and then execute all commands inside that bash prompt. This could have several benefits, including the ability to set up `just`'s auto-complete, easier alias definitions, etc. That might still be the way to go, but even then, I think it would be easier to reason about a persistent environment rather than an ephemeral one.

## Additional context

[...]

PR #4491, opened 2024-06-14
Use a persistent container for `ov`

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/4490 by @sarayourfriend 

## Description

Implements the solution described in the issue for using a persistent container to run `ov`. I've also added support on Linux for running commands as root (macOS already always runs commands in the container as root).

Anecdotally, `ov` commands do run noticeably faster for me. It's a matter of maybe 50-100ms at most, which is less so _faster_ and moreso just plain responsive feeling. If you happen to notice a change here one way or the other, please let me know. I'm curious to know how fast `docker ps` is on macOS hosts that have the virtual machine to contend with (my impression is they should be plenty fast?).

## Testing Instructions

To test this change:

1. Run `ov clean && ov init` to get a brand new environment set up
2. Make sure things still work, like `ov just api/up`, `ov just api/test`, `ov nuxt`, etc. Just pick some things you'd normally do with the development environment and try them out.
3. Prove the environment's persistence by running `ov sudo dnf update --downloadonly`. `dnf` should list some updates (hopefully some exist) and ask if you want to download them. Respond yes (type `y`, then return), and let them download. Now run the same command again, and respond yes again. dnf should report that all the packages are already downloaded. If you try the same process on `main`, the second time dnf will download the packages again, because the container environment was not persisted.

I will make sure the pinged reviewers include at least one macOS user and one Linux user. In doing so, the changes to accommodate `sudo` exercised in step 3 will be sufficiently tested cross platform.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker/dev_env/Dockerfile
+++ b/docker/dev_env/Dockerfile
@@ -60,8 +60,8 @@ RUN bash -c 'chmod -Rv 0777 /opt'
 # Avoid overwriting `.venv`'s from the host
 ENV PDM_VENV_IN_PROJECT="False"
 
-COPY entrypoint.sh /entrypoint.sh
+# Infinite sleep loop so that we create a persistent environment
+# Everything runs in the same container, rather than new one-off containers for each command
+ENTRYPOINT /bin/bash -c 'while true; do sleep 1; done;'
 
-ENTRYPOINT ["/entrypoint.sh"]
-
-CMD ["/bin/sh"]
+CMD ["/bin/bash"]
--- a/docker/dev_env/Dockerfile
+++ b/docker/dev_env/Dockerfile
@@ -60,8 +60,6 @@ RUN bash -c 'chmod -Rv 0777 /opt'
 # Avoid overwriting `.venv`'s from the host
 ENV PDM_VENV_IN_PROJECT="False"
 
-# Infinite sleep loop so that we create a persistent environment
-# Everything runs in the same container, rather than new one-off containers for each command
-ENTRYPOINT /bin/bash -c 'while true; do sleep 1; done;'
-
-CMD ["/bin/bash"]
+# The fedora container's default entrypoint + command behaviour is to keep running
+# so we do not need to modify entrypoint and command to use the container as a
+# persistent environment.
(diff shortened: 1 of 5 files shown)


stacimc (approved):
Works for me 👍

docker/dev_env/Dockerfile
```diff
@@ -60,8 +60,8 @@ RUN bash -c 'chmod -Rv 0777 /opt'
...
 
-COPY entrypoint.sh /entrypoint.sh
+# Infinite sleep loop so that we create a persistent environment
+# Everything runs in the same container, rather than new one-off containers for each command
+ENTRYPOINT /bin/bash -c 'while true; do sleep 1; done;'
 
-ENTRYPOINT ["/entrypoint.sh"]
-
-CMD ["/bin/sh"]
+CMD ["/bin/bash"]
```
dhruvkb, line 67: We can theoretically drop all of this. The `fedora` image runs continuously with

```
docker run -itd fedora
```

Screenshot 2024-06-15 at 1 54 14 PM

dhruvkb (commented):
I'm not getting into a formal review because I don't have enough concentration/time but I have a small suggestion.

docker/dev_env/Dockerfile
sarayourfriend: Oh, good to know! I will remove those entries for sure :+1:

sarayourfriend: Actually, I think I'd rather leave these explicit, both entrypoint and command, because it documents the behaviour we _expect_ without relying on the base image "incidentally" behaving that way. I'd rather explicitly instruct our image to loop and sleep. What do you think?

dhruvkb: I'll leave that up to you then. I personally would like to avoid instructions that can be omitted without altering the end experience but I can also get behind the value of being explicit.

sarayourfriend: I've pushed it. Part of why I was reluctant is because I felt the need to explain that we _wanted_ that behaviour, so it was documented in the Dockerfile. It's just as many lines as the explicit implementation, so I don't see a difference, really, but I suppose it's nice to use the default configuration as you said. It just doesn't make anything more concise or easier to understand overall, at least not to me.

dhruvkb (approved):
LGTM, looking forward to faster `ov` commands!

PR #4499, opened 2024-06-17
Include xml in frontend attribution options

### LINKED ISSUE #4430: Attribution: XML/RDF/Turtle please.
It would be handy if there was some additional formats for the "Credit the creator" section.
In particular - I would suggest at least it should include a simple well-formed XML record.

Better: one that corresponds to the Dublin Core specification : https://www.dublincore.org/specifications/dublin-core/dcmi-terms/

Or RDF in general - including a 'Turtle format'. 

- https://en.wikipedia.org/wiki/Resource_Description_Framework
- https://en.wikipedia.org/wiki/Turtle_(syntax)

(Although publishing in Dublin Core XML format would be enough for others to automatically translate this to other forms of RDF probably).

I would suggest this would also encourage more compliance with attribution , since it would be easier for the author to automatically credit creators.

madewithkode:
## Fixes

Fixes #4430 by @midijohnny!

## Description

Screenshot 2024-06-17 at 18 27 28

This PR seeks to add an extra Dublin-Core-Conforming XML attribution option on the frontend, allowing users to be able to credit creators in XML.

## Testing Instructions

This change can be observed by visiting the detail page of any media item on Openverse, under the "Credit the creator" section, you should see an extra XML option.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VMediaInfo/VCopyLicense.vue
```diff
@@ -39,6 +39,15 @@
...
         <p>{{ getAttributionMarkup({ isPlaintext: true }) }}</p>
       </VLicenseTabPanel>
+      <VLicenseTabPanel
+        :tab="tabs[3]"
+        :media-id="media.id"
+        :media-type="media.frontendMediaType"
+      >
+        <!-- Ignore reason: the interpolated string cannot have any whitespace around it when inside <p>, else there will be unwanted whitespace -->
+        <!-- prettier-ignore -->
+        <p id="attribution-xml" class="break-all font-mono" dir="ltr">{{ getAttributionMarkup({ isXml: true }) }}</p>
```
sarayourfriend, line 49: It might be good to use `pre` here, instead of `p`, so that the XML snippet retains its line breaks. Although, this approach with `p` matches the existing HTML version :shrug: Both are a difficult and annoying jumble to read, which might make sense for HTML (maybe more people copy/paste HTML without checking it) but for XML, it's definitely targeted at more technical users, so I think it could be formatted in presentation?

```suggestion
        <pre dir="ltr">{{ getAttributionMarkup({ isXml: true }) }}</pre>
```

frontend/src/utils/attribution-html.ts
```diff
@@ -124,6 +124,21 @@ const licenseElementImg = (licenseElement: LicenseElement): string => {
...
+ * Generate a Dublin Core conforming XML attribution snippet.
+ *
+ * @param mediaItem - the media item being attributed
+ * @returns the XML markup of the attribution
+ */
+const xmlAttribution = (mediaItem: Record<string, string>) =>
+  `<attribution xmlns:dc="http://purl.org/dc/elements/1.1/">
+	<dc:creator>${mediaItem.creator}</dc:creator>
+	<dc:title>${mediaItem.title}</dc:title>
+	<dc:rights>"${mediaItem.title}" ${mediaItem.byCreator} ${mediaItem.markedLicensed} ${mediaItem.license}. ${mediaItem.viewLegal}</dc:rights>
```
sarayourfriend, line 137: Could use `mediaItem.attribution` here for simplicity's sake. I think that catches some potential edge cases that exist? You'd need to add `attribution` to `AttributableMedia` as well (like you did for `frontendMediaType`).

```suggestion
	<dc:rights>${mediaItem.attribution}</dc:rights>
```

madewithkode:
Hi @sarayourfriend your suggestions have been resolved. However, I see that playwright visual regression tests continue to fail, I'm not exactly sure how to go about updating them, I've looked at the docs referenced by the bot on playwright but it's not exactly clear how to come up with the new comparison screenshots e.t.c. I'd appreciate if I could get a step by step instruction on how to tackle this.

dhruvkb:
@madewithkode here's a hack I've used in the past which can avoid the need to run the VR tests locally.

Go to the workflow summary page https://github.com/WordPress/openverse/actions/runs/9564951682 and look for the artifacts named `*_test_results`. Download them. Then extract the files in the artifact ZIP that you downloaded and look at the previous and new versions of the file and the diff. If the new version of the file looks correct, replace the file in the repo with the files from the artifact ZIP.

sarayourfriend:
@madewithkode here are instructions for updating snapshots in tests: https://docs.openverse.org/frontend/guides/test.html#updating-snapshots. Run `./ov just frontend/run test:playwright visual-regression -u` (leave off `./ov` if you haven't set that up yet).

We should probably combine that document with the one the bot links in its comment. I've long struggled to know what the precise difference was between reference and guides in our docs, and it's often muddied things up a bit like it has here. Sorry for the confusion there.

madewithkode:
Thanks @dhruvkb for the quick hack, going to try it out and hope it works, as running playwright locally appears to be a bit resource-intensive for me. @sarayourfriend Thanks for pointing to the actual documentation too and I totally agree that the information shared by the bot can really be improved to include the actual steps needed for updating snapshots. I might as well take up the linked issue to fix this after this one is done.

frontend/test/locales/ar.json
```diff
@@ -556,7 +556,8 @@
         "html": "لغة البرمجة",
         "plain": "نص عادي",
         "copyText": "نسخ النص",
-        "copied": "نسخ!"
+        "copied": "نسخ!",
+        "xml": "XML"
```
dhruvkb, line 560: Let's use the Arabic transliteration of "xml" here. Also noting, for a separate issue, that the translation of "html" is quite incorrect above.

```suggestion
        "xml": "إك سم إل"
```

zackkrida (approved):
This LGTM, thank you @madewithkode!

I do want to flag a concern for our designer @fcoveram: this change makes some of the tab labels occupy two lines on mobile instead of one. Here is an example:

https://github.com/WordPress/openverse/blob/4430_include_xml_in_frontend_attribution_options/frontend/test/playwright/visual-regression/pages/pages-single-result.spec.ts-snapshots/audio-ltr-from-search-results-with-additional-search-views-xs-linux.png

Do you think that looks okay, or should we change the design a bit, perhaps reducing the font size slightly on mobile or some other change? It doesn't have to happen in this PR in my opinion, but is something we could fix later.

dhruvkb (changes requested):
I have a bit of a small suggestion which will, unfortunately, necessitate new snapshots for Arabic. Mostly everything else looks fine and I would've approved the PR except for this change.

frontend/test/locales/ar.json
madewithkode: This has been updated.

frontend/src/utils/attribution-html.ts
```diff
@@ -163,15 +181,20 @@ export interface AttributionOptions {
...
+    isXml: false,
   }
 ): string => {
   if (!mediaItem) {
     return ""
   }
 
+  if (isXml) {
+    isPlaintext = true
+  }
```
dhruvkb, line 196: This might just be me, but this conflation of XML with plain text causes a bit of mental gymnastics in the logic below. I feel like using `isXml` and `isPlaintext` separately would've been clearer.

dhruvkb (commented):
There are a number of snapshots in the PR that are the same before and after, like the audio search and error pages, because they do not have the attribution box. The snapshot tests will likely still pass if these files are dropped from the PR.

I have a minor nit that seems okay to ignore.

frontend/src/utils/attribution-html.ts
sarayourfriend: Can we bypass the plaintext handling if we just use `mediaItem.attribution`, too, by the way?

dhruvkb: That could work, and it would cut down on the code the frontend duplicates from the API, but I'm not sure about localisation though. The frontend is capable of localising the attribution text in a way the API is not.

dhruvkb (approved):
After reversing the snapshots that had not been changed, I feel like this PR is good to go.

Thanks for the contribution @madewithkode! This is a very valuable feature and your implementation has been very clear and thorough.

frontend/src/utils/attribution-html.ts
sarayourfriend: Ah true :confused: In that case, should we undo the change from this comment before merging? https://github.com/WordPress/ openverse/pull/4499#discussion_r1643556075

dhruvkb: I don't a simple revert will be adequate. I'm personally not well versed with the i18n situation of the attribution box and feel like looking into what currently happens (and I expect should happen) will need a bit of time. Part of that's because I'm not entirely sure if XML, which is more geared towards machines than people, should even be translated.

I was only referring to the rich and plain text versions, which should be translated imo and using `mediaItem.attribution` is definitely not enough for those two types of attribution.

fcoveram:
> I do want to flag a concern for our designer @fcoveram: this change makes some of the tab labels occupy two lines on mobile instead of one. Here is an example:
>
> https://github.com/WordPress/openverse/blob/4430_include_xml_in_frontend_attribution_options/frontend/test/playwright/visual-regression/pages/pages-single-result.spec.ts-snapshots/audio-ltr-from-search-results-with-additional-search-views-xs-linux.png
>
> Do you think that looks okay, or should we change the design a bit, perhaps reducing the font size slightly on mobile or some other change? It doesn't have to happen in this PR in my opinion, but is something we could fix later.

I would follow the common practice of replacing with a dropdown element. Tailwind addresses [this](https://tailwindui.com/components/application-ui/navigation/tabs#component-de43ff625fee032d234b14989e88422f) pretty well.

zackkrida:
That's a great (and simple) idea, @fcoveram. I'll make a follow up issue for that!

frontend/src/utils/attribution-html.ts
sarayourfriend: @zackkrida suggested using just the license URL, and the DC standard recommendation matches that too:

> Typically, rights information includes a statement about various property rights associated with the resource, including intellectual property rights. Recommended practice is to refer to a rights statement with a URI. If this is not possible or feasible, a literal value (name, label, or short text) may be provided.

Then we don't need to worry about i18n at all for this (ideal!).

dhruvkb: @madewithkode would you be able to update this to use the license URL instead of the attribution text?

madewithkode: @dhruvkb this would mean just switching `<dc:rights>${mediaItem.attribution}</dc:rights>` in the XML attribution snippet to reference the CC license URL as in `<dc:rights>${mediaItem.license_url}</dc:rights>` right?

madewithkode: This has been updated.

sarayourfriend:
@madewithkode I pushed a change to revert some unneeded snapshot updates as well as to fix a bug I realised with unescaped elements in the title and other work meta data. Some works would have had broken XML due to unescaped XML (or XML-like) text. Take this work, for example, where the creator's included their email in their username: https://openverse.org/image/7be0e5cb-6308-4c11-b397-38baf2d0a4e7

The XML generated without escaping it (which is caused by setting `isPlaintext = true`), is as follows:

```xml
<attribution xmlns:dc="http://purl.org/dc/elements/1.1/">
	<dc:creator>Photographer: Richard Ling <wikipedia@rling.com></dc:creator>
	<dc:title>Grey Nurse Shark at Fish Rock Cave, NSW</dc:title>
	<dc:rights>https://creativecommons.org/licenses/by-sa/3.0/</dc:rights>
	<dc:identifier>https://commons.wikimedia.org/w/index.php?curid=130380</dc:identifier>
	<dc:type>StillImage</dc:type>
</attribution>
```

That won't pass XML validation. Fix was easy though, just removed the lines setting `isPlaintext = true` to enable the escaping (though really we should separate these concerns into more explicit variables, I think!).

I'm also rebasing and re-running the snapshot tests to fix the conflicts for you.

sarayourfriend (approved):
LGTM!

madewithkode:
> @madewithkode I pushed a change to revert some unneeded snapshot updates as well as to fix a bug I realised with unescaped elements in the title and other work meta data. Some works would have had broken XML due to unescaped XML (or XML-like) text. Take this work, for example, where the creator's included their email in their username: https://openverse.org/image/7be0e5cb-6308-4c11-b397-38baf2d0a4e7
> 
> The XML generated without escaping it (which is caused by setting `isPlaintext = true`), is as follows:
> 
> ```
> <attribution xmlns:dc="http://purl.org/dc/elements/1.1/">
> 	<dc:creator>Photographer: Richard Ling <wikipedia@rling.com></dc:creator>
> 	<dc:title>Grey Nurse Shark at Fish Rock Cave, NSW</dc:title>
> 	<dc:rights>https://creativecommons.org/licenses/by-sa/3.0/</dc:rights>
> 	<dc:identifier>https://commons.wikimedia.org/w/index.php?curid=130380</dc:identifier>
> 	<dc:type>StillImage</dc:type>
> </attribution>
> ```
> 
> That won't pass XML validation. Fix was easy though, just removed the lines setting `isPlaintext = true` to enable the escaping (though really we should separate these concerns into more explicit variables, I think!).
> 
> I'm also rebasing and re-running the snapshot tests to fix the conflicts for you.

Thank you for the changes and for spotting this edge case @sarayourfriend

sarayourfriend:
Thanks for your hard work, @madewithkode :pray:

PR #4524, opened 2024-06-19
Explicitly include Filter Data step in ingestion server removal IP

### LINKED ISSUE #4456: Update ingestion server removal IP to include plan for filtering tags
## Description

Recent discussions in data related projects have uncovered a need to separate the tag filtering step from the other cleanup operations occurring during the data refresh:

https://github.com/WordPress/openverse/blob/3747f9aa40ed03899becb98ecae2abf926c8875f/ingestion_server/ingestion_server/cleanup.py#L119-L150

Per discussion in https://github.com/WordPress/openverse/pull/4417#discussion_r1623674796, we want to **keep** the tags that do not meet the criteria they're currently being filtered on (denylist and tag accuracy) rather than remove them from the catalog database. Since the intent of the data normalization project is to remove the cleanup steps from the ingestion server, it's imperative we move this filtering option into the new data refresh process. As such, we'll need to modify the [implementation plan for that project](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html) to include where this filtering will take place.

It may be possible to add this as another step in the DAG, but this can be an intensive (and long-running) filtering operation. We also need to be clear at which step the filtering is happening: are we filtering the data within the temporary table prior to indexing and promoting it, or are we filtering the values as they go into Elasticsearch and leaving them in the API database? The former approach is the current one, but the latter provides more flexibility for rebuilding the index without having to copy the upstream table again. If we filter at the ES level, we could potentially leave the full list of tags in the API's response body for each record. This would mean we'd be returning the full list of tags (including denylisted and non-confident tags), but we would only be performing _searches_ against the filtered tags.

Another consideration would need to be made for where in the new data refresh DAG this would run. It could be run as a set of mapped tasks directly in the DAG, or (since we already have the indexer workers available for this kind of chunked operation) we could add it as a second task that's handled by the indexer workers within the DAG. The modifications to the IP will need to make this distinction explicit.

This step is necessary to define going forward, since we may want to filter out entire tag providers (e.g. Clarifai) or adjust the confidence interval down the line, and we want to preserve this data in the catalog.

## Additional context

See: #430, #431, and #3925

AetherUnbound:
## Fixes

Fixes #4456 by @AetherUnbound

## Description

This PR modifies the ingestion server removal IP to explicitly include a plan for handling the filtering of data which occurs during the current "cleanup" step and _will not_ be handled by the data normalization project.

While the change to the plan is minimal, I added a significant footnote just to provide context, describe what's needed, and then provide suggestions for the future. The suggestions made are **not necessary** at this moment, in my opinion. We can start with just slapping what used to be the "cleanup" step of the ingestion server into the new data refresh DAG and calling it good for now. I do think, though, that we could make an issue for optimizing this down the line and doing it in a more "Airflow-y" manner. It'll require more work to do that and it's worth doing eventually, but not needed now!

## Testing Instructions

View the rendered document and make sure the description looks okay!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
+++ b/documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
@@ -347,7 +347,24 @@ server's work:
 - `Filter Data`: initially, this should be a single python task which exactly
   mirrors the behavior of the
   [`clean_image_data` function of `cleanup.py`](https://github.com/WordPress/openverse/blob/47fe5df0e9b8ad3dba06021f4cd4af9139977644/ingestion_server/ingestion_server/cleanup.py#L295)
-  (only applying the tag-specific steps) on the ingestion server[^3].
+  (only applying the tag-specific steps) on the ingestion server[^3]. The
+  easiest way to do this would be to directly map the functionality of the
+  ingestion server on this step within a single Airflow task. The steps for this
+  task are as follows (see [Alternatives](#filtering-approach) for possible
+  future directions):
+  1. Get a batch of records from the database using `CLEANUP_BUFFER_SIZE`
+  2. Divide batch up into `multiprocessing.cpu_count()` subbatches
+  3. Split the filtering up into separate workers using multiprocessing
+  4. On each process
+     1. Create a new DB connection & cursor per worker
+     2. Iterate through each record
+        1. Remove tags below confidence level
+        2. Remove tags that need to be filtered (denylisted, machine-generated
+           filter list, provider, etc)
+        3. Only surface the record if it needs to be changed
+        4. Update each records one by one with a single `UPDATE`
+     3. Commit cursor and close connection
+  5. Repeat steps 1-4 until all batches are consumed
 - `Create Index`: we can use our existing
   [Elasticsearch tasks](https://github.com/WordPress/openverse/blob/05ff48d05f2163104151c5589cf352a156bc6a97/catalog/dags/common/elasticsearch.py#L82)
   to create the new elasticsearch index with the index suffix generated in the
@@ -565,6 +582,8 @@ No new tools or packages are required.
 
 <!-- Describe any alternatives considered and why they were not chosen or recommended. -->
 
+### ECS approach
+
 The alternative options of using an ECS approach or performing the reindex
 entirely in Airflow are discussed at length in the
 [Approach to the Distributed Reindex](#approach-to-the-distributed-reindex)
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -609,3 +617,52 @@ monitor the first production data refreshes closely.
 
 [^2]:
     https://towardsdatascience.com/using-apache-airflow-dockeroperator-with-docker-compose-57d0217c8219
+
+[^3]:
+    See #4456 for further context on this. The filtering is a _necessary_ step
+    of the data refresh we need to carry forward even after removing the other
+    cleanup steps. There are a number of ways we can accomplish this, but by far
```
krysal, line 624: This looks too long to be a footnote. Wouldn't it be better under the "Add remaining steps to the Data Refresh DAGs" section in the filtering step?

krysal (approved):
LGTM! I agree that the step being a copy of the current cleanup step is fine. Thanks for writing this addendum!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
```diff
@@ -609,3 +617,52 @@ monitor the first production data refreshes closely.
...
+          4. Update each records one by one with a single `UPDATE`
+       3. Commit cursor and close connection
+    5. Repeat steps 1-4 until all batches are consumed
+
+    The Airflow scheduler container has access to 4 cores, which is the same as
+    the ingestion server where this step was originally running. At present it
+    takes about 8 hours for all cleanup steps, but that includes the URL
+    cleaning which is certainly more time intensive than the tag filtering since
+    it makes outbound requests. Running the tag filtering on Airflow should not
+    impact any of the other running tasks or saturate the instance.
```
stacimc, line 648: > but that includes the URL cleaning which is certainly more time intensive than the tag filtering since it makes outbound requests.

Great point I totally hadn't thought of!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
stacimc: If we want to make the footnote smaller I think it would be better move the specific steps for the task into `Create the new data refresh DAG factory and move initial steps into Airflow` (since this is one of the initial steps, even though it's going to end up being implemented after the others) and the suggested future optimizations into `Alternatives` or something like that.

Not remotely a blocker for me but might make it easier to find when reading back through this document later.

stacimc (approved):
LGTM -- I was envisioning something much more involved (along the lines of the future optimizations you've documented here), but I'm sold on your plan. The simple approach is _so_ much simpler to implement and there's really no effort wasted if we end up wanting to optimize down the road. Let's not overcomplicate things too early!

documentation/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.md
AetherUnbound: Thanks you two for mentioning it - part of why I wanted to add it as a footnote was because this context has been added after-the-fact, and having it literally at the end of the document in the form of a footnote visually and conceptually matches that intent! I can move some of the information around as suggested, then merge this in 😄

AetherUnbound: Done in d84ddde!

Issue #1910, opened 2023-04-27
Rename docker-compose services with more obvious names


sarayourfriend:
## Problem

We have a few services that have non-obvious names for some of their containers, specifically "web" (Django API) and "webserver" (Airflow). This is annoying because not only do devs need to remember these (now) arbitrary names, certain containers are ambiguous enough that commands like "attach" don't work well. Try "just attach ingestion_server" and you'll be met with a nasty warning caused by the ambiguity of the worker and the main service name.

## Description

Disambiguate docker-compose service names to improve usability and reduce cognitive load for devs interacting with the stack.

dhruvkb:
While #1935 won't solve the problem of poorly named services, it will prevent the error you mention about attaching to the `ingestion_server`.

dhruvkb:
Applied the "💬 talk: discussion" label to allow us to settle on unambiguous names for the services.

sarayourfriend:
Proposed names:

- `web` -> `api` or `django`
- `webserver` -> `catalog` or `airflow`
- `scheduler` -> `catalog_scheduler` or `airflow_scheduler` to match above
- `proxy` -> `api_proxy` or `django_proxy` (or remove it altogether, is it ever used?)
- `s3` -> `catalog_s3` or `airflow_s3`
- `cache` -> `redis` (we use it for more than "caching", not the biggest deal, `redis` is easier to remember than `cache` imo)
- `upstream_db` -> `catalog_db` or `airflow_db`
- `db` -> `api_db` or `django_db`

I'm torn of whether to name things after the "service" or the technology used. I prefer the service name (`api`, `catalog`, etc) because that's usually how we talk about them. At the same time, "api" is ambiguous in a context where most of our services actually do have web APIs, whether it's Falcon's `api.py`, Nuxt's server endpoints, or our DRF API. Using the name of the technology used removes that ambiguity but would mix the langauge we use to talk about them in the docs, for example.

AetherUnbound:
I think those names look good and I agree, I think `api` and `catalog` should be sufficient prefixes (or names where that's warranted).

dhruvkb:
I like `api` and `cat` (🐈 as a cute shorthand for the `catalog`), but `catalog` works just fine.

PR #4526, opened 2024-06-19
Fix separators in catalog and dev-env images and dev-env volume

dhruvkb:
### Image conventions

Conventionally, our services in Docker Compose are named using snake case (e.g. `service_name`). This means that images built by Docker Compose are typically named `openverse-service_name`.

This PR updates the catalog and dev-env images to comply with this convention.

- `openverse_catalog` &rarr; `openverse-catalog`
- `openverse-dev-env` &rarr; `openverse-dev_env`

Screenshot 2024-06-24 at 12 04 32 PM

### Volume conventions

Similarly conventionally, our volumes in Docker Compose are named using kebab case (e.g. `volume-name`). This means that volumes built by Docker Compose are typically named (`openverse_volume-name`).

This PR updates the dev-env volume to comply with this convention.

- `openverse-dev-env` &rarr; `openverse_dev-env`

Screenshot 2024-06-24 at 12 06 57 PM

### Summary

This is only a change to the development environment. Also note that the catalog image name was [already `openverse-catalog` in GHCR](https://github.com/WordPress/openverse/pkgs/container/openverse-catalog).

## Note for users

This change to the image and volume names means that `ov` will need to be rebuild the image (which should be fast enough and utilise the cache) and pre-commit environments (which will be slow and necessitates a fast Internet connection).

Run `ov clean && ov init && ov just lint` to make sure the new dev environment is set up. You can remove the old image and volume after this step.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/compose.yml
```diff
@@ -7,7 +7,7 @@ x-airflow-common: &airflow-common
   depends_on:
     - postgres
     - s3
-  image: openverse_catalog
+  image: openverse-catalog
```
sarayourfriend, line 10: Why does this `image` entry exist at all on these? Every time I run `just up` these days, docker tries to retrieve these images and it fails with a (seemingly) random authentication error. What's the purpose of these image names on our services, exactly? If they aren't necessary, I'd much rather remove them from everywhere.

sarayourfriend (dismissed):
Fine by me, just left a non-blocking tangential comment for an alternative "fix"

catalog/compose.yml
dhruvkb: Having both [`image` and `build` in the Compose file](https://docs.docker.com/compose/compose-file/build/#using-build-and-image) means that Docker will first try to pull it and if the image is not found, it will be built. This is standard and expected behaviour so I am wondering if what you're seeing is an info/ warning message and not an error.

Also, we can theoretically control the pulling behaviour using [`pull_policy`](https://docs.docker.com/compose/compose-file/05-services/#pull_policy). I suppose setting the `pull_policy` to "never" might get rid of that message.

sarayourfriend: It is a warning log, not an error (I misspoke). I suppose it is recent behaviour that it actually tries to pull the images, or at least that it warns me that it wasn't able to. Anecdotally, it sometimes spends several seconds to a minute trying to pull the images. I suppose if someone published an image by these names to `docker.io` or whatever default registry a given container runtime uses it would actually try to pull that instead of building locally. That doesn't seem right? In what cases do any of these `image` entries do anything other than waste cycles trying to download images that aren't findable?

> I suppose setting the pull_policy to "never" might get rid of that message.

Why would that be different than just removing `image` altogether?

Does `image` tell Docker to try to reuse the _local_ image? And if so, why is that better than just the build instructions? If the local image is present, then Docker will use the cache, and it wouldn't spend nearly any time at all trying to build. If `pull_policy: never` combines the two (never pull, but still try to find it locally), then I guess that _is_ the best of both worlds. I'm surprised the image entry is needed at all for images that are nearly always locally built (in any meaningful case) :thinking:

dhruvkb: I think what `image` does for us is allow us to use a name for the image that's different from the service name (otherwise Docker will automatically tag the image based on the name of the service, which is not always what we want).

It also allows us to reuse a image for multiple services as once an image has been built for a service, other services can directly use it without building copies themselves.

sarayourfriend: Could we simplify all this by changing the name of the services, as recommended in #1910?

Does docker cache using the Dockerfile and context, so examples like the catalog's three services would still have the same images? Maybe that's wrong. I'll try it out tomorrow if you don't have a definitive answer.

It sounds like `pull_policy: never` would solve the warning, though? If not get rid of the image names?

dhruvkb:
The catalog tests are passing just fine for me locally so I have no clue why the tests are failing in CI. @AetherUnbound @stacimc if you have a moment, could you help me with debugging this? Thanks!

catalog/compose.yml
dhruvkb: Yes, simplifying the names of services will remove some of these `image` fields. Not all, but at least those where the service name is the same as that of the image.

For services with different names, different images will be created by Docker. Because of efficiency, they will share underlying images and layers, so space consumed will not go up, but it will still lead to multiple image tags.

`pull_policy: never` is definitely useful to avoid pulling a random, potentially malicious image from Docker, and will solve that warning. Should that be part of this PR?

sarayourfriend: This PR is fine as is to me, I was just wanting to know if there's a bigger change that helps address some other issues :slightly_smiling_face: However you feel is the best way to solve these devex problems is good to me! I suppose the only thing to keep in mind is that #1910 will affect how people run just scripts, just logs, in particular being one I use a lot and relies on the service names, but I always mess up and write `just logs api` anyway LOL. That just means it would need to be communicated more clearly than changes like this that don't affect how people interact with it.

But really, from my perspective, up to you how you think is the best way to solve these issues :+1: Aside from the specific details, whether in this PR, another one, whatever works!

AetherUnbound (approved):
LGTM, it's nice to have parity here now even if we go with some of the alternatives that Sara has suggested down the line 🙂

AetherUnbound:
Oh that's right, lemme look at the tests.

AetherUnbound:
The `index.json` within the `catalog` artifact is:

```json
{"schemaVersion":2,"manifests":[{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","digest":"sha256:b7e073ea8840f0922b82ef1fc6dada298e5216e0500e50efd76f2fd6f06b3d7d","size":5985,"annotations":{"io.containerd.image.name":"docker.io/library/openverse-catalog:latest","org.opencontainers.image.created":"2024-06-19T23:39:52Z","org.opencontainers.image.ref.name":"latest"},"platform":{"architecture":"amd64","os":"linux"}}]}
```

So it _looks_ like it's being saved as `openverse-catalog` correctly at least 🤔

dhruvkb:
I figured out what the issue was. Once we fixed the catalog image name to use hyphens, the test job started using the prod version of the image to run tests with. Earlier it would just not find it and build a dev version on demand.

I've updated the test job to build a dev image, identical to our process for the API. Thanks @sarayourfriend and @AetherUnbound for nudging me in the right direction!

docker/dev_env/run.sh
```diff
@@ -2,10 +2,10 @@
 
 set -e
 
-container_name="openverse-dev-env"
+container_name="openverse-dev_env"
```
AetherUnbound, line 5: Wait, why did this get changed _away_ from the hyphen now?

dhruvkb: Our style is hyphens between "openverse" and service name and underscores within the service name.

sarayourfriend: This is going to force everyone using `ov` to rebuild their container from scratch, including anything cached in `opt` (like pre-commit environments).

Is this really necessary? And is the use of underscores in the app name a prescriptive "must do no matter what"? (I personally find the mixed usage ugly, but just my opinion)

Mostly just concerned about the effect on every one currently using `ov`.

AetherUnbound (approved):
Was able to rebuild stuff locally and it worked great!

stacimc (approved):
LGTM, rebuilt everything locally. No issues with the indexer worker either :)

Edit: Approving but I don't want @sarayourfriend's comment to get missed, so boosting: https://github.com/WordPress/openverse/pull/4526#discussion_r1649391544

docker/dev_env/run.sh
dhruvkb: The only reason this PR exists is because I'm obsessing over a minor, largely inconsequential detail. I don't mind closing it if the consensus is that it's too much of a hassle. This was nagging at me since I noticed it and I had to make the PR just for closure. Nonetheless, I'll just give a bit of history that explains how we got here to this ugly mix of separators.

When we were using Docker Compose v1, the separator between the directory/repo/project name and the service name (hardcoded by Docker Compose) was `_`. So for consistency, we used service names with underscore separators (like "ingestion_server" or "upstream_db"). 

With Docker Compose v2, the separator was changed to `-`. By that time, I personally felt it was better to use a different separator with the "openverse" prefix so that we can conveniently split at `-` to separate the prefix.

dhruvkb:
I agree with @sarayourfriend's comment about the mixed usage being "ugly" and the name change causing a hassle for everyone using `ov`. Personally I value consistency over ugliness, and would justify the hassle worth it, _at the moment_, where we're ironing out the kinks with `ov`. 

So I won't merge the PR till we can be sure this is actually worth merging.

docker/dev_env/run.sh
sarayourfriend: Gotcha, thanks for the background.

> so that we can conveniently split at - to separate the prefix

I suppose `image_name.replace("openverse-", "")` is either just as or more reliable than `image_name.split("-")[-1]`. If we named an image without the hypen/underscore encoding in mind, the split version fails every time. The replace (or even better, substring, `image_name.substring("openverse-".length)` in js/`image_name[len("openverse-"):]` in python) always works.

Well, anyway, I just wanted to point out that this will cause everyone to need to rebuild the dev image (which may be cached? In which case are the explicit image names necessary?), but in particular any state that exists _after_ the dev image is built during the course of usage (and in particular the pre-commit envs, because they do take a long time to spin up unfortunately), will introduce some tedium.

At the very least it's something to be contientious of and maybe let the team know so that anyone who doesn't notice this PR or see the image name difference doesn't end up frustrated and confused why their environment is suddenly slow again while it initialises the first time around.

Overall my vote is just going to be: whatever is the best thing to have long term. If the mix of separators is the best thing, fine.

Looking at the documentation about image names, I guess we could use `openverse/<image name>` too, and that would actually fit the specific usage more closely. From [the Docker docs on image names](https://docs.docker.com/reference/cli/docker/image/tag/#extended-description):

>     `[NAMESPACE/]REPOSITORY`: The first, optional component is typically a user's or an organization's namespace. The second, mandatory component is the repository name. When the namespace is not present, Docker uses library as the default namespace.

If we're going to cause an upheaval, and there's a "more correct" or obvious or widely shared approach to this kind of thing, we could go ahead and take the opportunity to do it now, rather than multiple times. I'm personally a lot more likely to remember the nuance between `openverse/` then `openverse-`.

Just my two cents! If we're delving into the weeds of little details, might as well delve all the way in and ask if there's a "perfect" version. Anecdotally, I can't recall any public docker images I've used that use underscore separators _at all_. Hyphens seem much more common :thinking:

But it's just little details. I agree consistency would be best above anything else. Consistency _and_ matching the exact features of whatever implementation we're relying on? Maybe even better...

While we're at it, for the images that are published, should we use the GHCR reference for them instead? Maybe then pulling would actually _be_ a benefit. An individual might get some stages of the build cached on that first run of the stack?

But it would still be nice to move those towards the same namespace approach of `openverse/`, if we can do that? I can't remember of we control https://github.com/openverse/, but GHCR didn't mind when I pushed an image with multiple namespaces, so `wordpress/openverse/<image name>` would work (example here: https://github.com/users/sarayourfriend/packages/container/package/openverse%2Ffrontend).

The OCI spec explicitly allows multiple namespaces: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests. Docker's documentation mentions that most repositories only support at most two slashes, but we use GHCR and it handles them fine. If we ever had to change registries, we can make sure we only change to a spec-compliant registry (benefits all around there, I'd say).

Look, bottom line for me is: let's pick whatever is the best thing we actually _want_, especially if this will be somewhat disruptive anyway, so that we aren't left groaning or wanting to make another potentially disruptive change again in the future. If that's the current overloaded meaning of hyphens and underscores: fine, all good, let's go for it, and let's do some work to make sure we document and maybe lint for that. If it's the namespaces, same deal, document and educate/lint.

Whichever way we go is fine with me. Like the slashes because I think they look nice and are easy to remember what they mean, but I ultimately don't care about the specific thing we choose here, as I agree the more fundamentally important thing is consistency. But if we have an opportunity to change towards a better approach (if it is indeed better) then we can take that now, if we want to.

dhruvkb: I have no strong preferences either.

1. I like your ideas with slashes and namespaces, but that seems like a pretty big change across our code and infrastructure and docs. I also see there is some uncertainty around support for the multiple namespaces spec across repositories (the link in your example does not work).
2. This PR, my second preference, involves 2 changes and that too only to the local dev environment. Everything else remains unchanged, and I find that stability quite nice. Also I only made the PR to resolve small discrepancies, I did not expect to be making huge changes.

I don't know how one could lint for image names but regardless of the choice, it could be quite easy to document the naming scheme somewhere.

sarayourfriend: Sorry, I'm making a much of nothing in this thread. To clarify I am also not blocking this change (I would have requested changes if that was the case). The main important thing I want to stress is the disruption to `ov`, and to request that the disruption (and reason for it) is communicated clearly using our regular communication tools. Knowing a similar disruption won't happen again any time soon would also be good. So, if this hyphen/underscore meaning is meaningful enough for such a disruption, then yes, that should also be documented and made visible, which isn't inherent in just documenting. Linting image names to enforce an expectation could be done rather trivially using the output from `ov just dc config`. It doesn't cover instances like `ov` that are outside of the compose stack (but then I don't personally see why this image name needs to match anyway, as it is outside the application stack where any automation could reasonably be applied, but I'm not arguing one way or another here... like I said, I would have requested changes if I felt so strongly).

To me it is also worth questioning whether the hyphen/underscore meaning is valuable or reliable, if either of those are the goal, or what the consistency brings. This looks like just a trivial change in names but becuase of the effect on breaking the `ov` volume, it's unfortunately more than just a change in names :shrug:

For what it's worth, as far as I can tell the GHCR link does work (and I can `docker pull ghcr.io/sarayourfriend/openverse/frontend:testing`). Maybe you are referencing a different link?

dhruvkb: The link doesn't work for me.

https://github.com/WordPress/openverse/assets/16580576/6e6b30fd-78b6-4d16-8f84-4c1e5a585ef6

It might be a permissions issue.

```
❯ docker pull ghcr.io/sarayourfriend/openverse/frontend:testing
Error response from daemon: Head "https://ghcr.io/v2/sarayourfriend/openverse/frontend/manifests/testing": unauthorized
```

I feel like once this change is merged, the image names will be consistent enough that I might be satisfied and we could reliably keep them the same for a long time.

Also the effect on the `ov` volume is just that all the stuff will automatically be rebuilt. There will be a small increase in first run time post-merge, but mostly the workflow will remain unbroken and unchanged. I don't see this as a breaking change, and FWIW, we should be open to breaking changes with `ov` for a short while as it is fairly new, has not been widely adopted, and we're ironing out the kinks.

sarayourfriend: Ah, it must be a permissions issue, yes.

> Also the effect on the ov volume is just that all the stuff will automatically be rebuilt. There will be a small increase in first run time post-merge, but mostly the workflow will remain unbroken and unchanged. I don't see this as a breaking change, and FWIW, we should be open to breaking changes with ov for a short while as it is fairly new, has not been widely adopted, and we're ironing out the kinks.

It's the pre-commit environments that take the longest for me, at least, and it's a pain and extremely disruptive because it's always just as I am usually wrapping up a bit of work. The renovate config validator takes so long! Maybe it shouldn't be in the local pre-commit, and we could just validate it in CI.

I agree we can accept breaking changes for a while, but I also want to be mindful of the fact that because it's new and different, it can feel extra frustrating if things suddenly feel inconsistent or random. Like, if someone doesn't know about this PR and suddenly they have to rebuild the pre-commit env again just before merging, it will be frustrating, confusing, and seem like `ov` has introduced an inscruitable and frustrating lack of transparency and consistency into the development environment. That's why it's important to communicate the change and why it happened. I know I would be frustrated and sceptical if `ov` suddenly seemed to just make things more confusing and slower when the other way worked fine (for me), and I'd question it's value if I didn't have the context of why e.g., committing suddenly was taking a very long time because of the pre-commit environment creation happening _again_.

As I said:

> This looks like just a trivial change in names but becuase of the effect on breaking the ov volume, it's unfortunately more than just a change in names 🤷

I think that's mischaracterising it, and a small increase in the first run time is also mischaracterising it, ignoring a significant frustration. Maybe I'm the only one who finds a slow pre-commit a pain in the butt, but I really do find it almost intolerable (I almost always `--no-verify` the frontend's pre-push hooks, which are truly intolerably slow to me, and often run on rebased pushes). Anyway, I just don't want that to be brushed off and because `ov` is new it's good to make sure everyone knows about it.

Will you communicate the change, the reason for it, and the effect of it to everyone? In Slack, at the very least? I think the PR description should also include a note indicating that. I just don't think it's as trivial a change for that first touch as you've made it out to be.

dhruvkb: > It's the pre-commit environments that take the longest for me

That is true. pre-commit setup is a bit slow and any hooks that take a long time to be configured or run should be relegated to the CI workflow. In fact @AetherUnbound and I were discussing the making pre-commit hooks opt-in, so that there is one fewer barrier for new contributors.

> That's why it's important to communicate the change and why it happened.

That is definitely something that we will do. Once this PR is merged, we should post a message on Slack about the change and what actions a contributor needs to take to have everything working as before.

> I almost always --no-verify the frontend's pre-push hooks

I feel like `--no-verify` carries a negative stigma which it does not deserve. Since we have a fast-failing lint job in CI, we should feel free to use `--no-verify` if that speeds things up or reduces frustration.

> Will you communicate the change, the reason for it, and the effect of it to everyone?

Yes!

PR #4527, opened 2024-06-20
Set `ov` workdir to current working directory

sarayourfriend:
## Description

This makes `ov` contextual to the working directory, so you can rely on your host CWD inside `ov`. This works because the Openverse repository is mapped with the literal directory path from your host (rather than to a distinct container path), so the CWD will exist in the container, so long as `ov` is run from within the repository.

I considered making this conditional by testing whether `$(pwd)` is actually a subpath of `OPENVERSE_PROJECT`, but running `ov` outside of the repository seems weird to try to explicitly support :thinking: 

## Testing Instructions

Think of something you often do that is CWD contextual. Maybe, for example, running a just recipe from withing a subproject's directory. Or maybe using `pnpm exec`. Or `pdm run`. Now cd to the directory and prefix it with `ov`. Start with `pwd` and you'll see it work directly, hopefully. This also mostly only matters if you have `ov` on your PATH, but it should also work if you evoke it using a direct reference (e.g,, `../ov` from inside `/api`).

```
$ cd frontend && ov pwd
/path/to/openverse/frontend

$ cd ../api && ov pwd
/path/to/openverse/api
```

Then try other stuff that would be context dependent:
```
$ cd api && ov pdm info
PDM version:
  2.15.4
Python Interpreter:
  /opt/pdm/venvs/api-UJMDzYmD-3.11/bin/python (3.11)
Project Root:
  /home/sara/projects/openverse/api
Local Packages:

```

etc...

You get the idea, hopefully!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker/dev_env/run.sh
+++ b/docker/dev_env/run.sh
@@ -13,7 +13,7 @@ shared_args=(
   -i
   --env "OPENVERSE_PROJECT=$OPENVERSE_PROJECT"
   --env "TERM=xterm-256color"
-  --workdir "$OPENVERSE_PROJECT"
+  --workdir "$CWD"
 )
 
 run_args=(
--- a/docker/dev_env/run.sh
+++ b/docker/dev_env/run.sh
@@ -13,7 +13,7 @@ shared_args=(
   -i
   --env "OPENVERSE_PROJECT=$OPENVERSE_PROJECT"
   --env "TERM=xterm-256color"
-  --workdir "$CWD"
+  --workdir "$(pwd)"
 )
 
 run_args=(
@@ -104,4 +104,4 @@ else
   docker start "$existing_container_id" 1>/dev/null
 fi
 
-docker exec "${shared_args[@]}" "$container_name" python3 docker/dev_env/exec.py "${_cmd[@]}"
+docker exec "${shared_args[@]}" "$container_name" python3 "$OPENVERSE_PROJECT"/docker/dev_env/exec.py "${_cmd[@]}"


stacimc:
With these changes applied, I'm getting the error `python3: can't open file '//docker/dev_env/exec.py': [Errno 2] No such file or directory` every time I try to run any just recipe with `ov`, even ones that work on `main`.

Steps I'm doing:
* Run `./ov clean && ./ov init` on main
* Run `./ov just catalog/pgcli` and see that it works
* Check out branch, run `./ov clean && ./ov init`
* Run `./ov just catalog/pgcli` and see the error

sarayourfriend:
Fascinating! I think I was seeing a bunch of false positives when using `$CWD`, which seems to be context dependent, rather than the actual working directory :thinking:

I'm not sure why it worked on my desktop, but I get the same error on my laptop. The fix was to use `pwd` to get the actual working directory (`$CWD` seems to resolve to different things depending on the environment, not sure), and then to use an absolute reference to the `exec.py` script.

Can you try again, Staci? I've updated the testing instructions for things that will absolutely prove it works instead of false positives that I had before (whoops!)

stacimc (approved):
🥳 That's fixed it! LGTM :)

AetherUnbound (approved):
LGTM and works great!

PR #4538, opened 2024-06-21
Add low-level TS API client with generated types

sarayourfriend:
## Description

Based on my work exploring this concept in https://git.sr.ht/~sara/openverse-api-client.

There are major differences from that version:
1. I have only included the TypeScript client in this PR. (Previous versions included a Python client).
2. The TypeScript client uses [`openapi-fetch`](https://openapi-ts.dev/openapi-fetch/) and it's TypeScript support package to generate TypeScript interfaces from the OpenAPI schema, instead of generating them with Python and Jinja2 templates.

I've already demonstrated the value of the JavaScript client for even our own frontend. The Python client, on the other hand, would be more useful for things like the Openverse Slack Reaction app (which isn't part of the project yet). For that reason, I think it is necessary to scope this only to the JS client. It also makes this easier to review, because the JS client is substantially simpler than the Python client. That's thanks to the excellence of the `openapi-fetch` and `openapi-typescript` packages. Nothing of that quality exists for Python. We will need to do much more work for a generated Python client, and it deserves to be its own pull request.

To briefly document the approach:

- I've introduced one new package, `@openverse/api-client`. The package houses the generated types and exports a client wrapping those types.
- The package also implements an [`openapi-fetch` Middleware](https://openapi-ts.dev/openapi-fetch/middleware-auth) to handle the Openverse API authentication flow.
  - This flow is tested with `nock`. However, it might be worth adding another test suite of integration level tests that run against a "real" API instance (even if it's the local API). I think that would be something good for a follow-up PR, rather than in this one, as it will require additional tooling to make possible in CI/CD.
  - For an example of the type of tooling I'm talking about, [see this commit which removes similar tooling used to enable the previously included Python package to run tests against the local dev API instance](https://github.com/WordPress/openverse/pull/4538/commits/fb393ee83e8b9e6d208e644c6c9f66a8584df092). It is possible, and even desirable, to have this. But it should be in a separate PR to ensure ease of review.
- Package publication is _not_ included, but the package's `build` script already includes checks to ensure the package builds and passes basic checks required for publication. A separate PR should follow to document and implement a manual publication process. We can turn that into an automated process later on.

Further improvements can be made to create higher level APIs that build upon this "low-level" version, as discussed by @obulat (who raised the question in the first place) and myself in the comments of this PR. Additionally, we could consider creating an intermediate package `@openverse/types` to house the generated OpenAPI schema types, as well as further support types like we use in the frontend's `src/types` directory, such as definitions of media types, and so forth. That package could also export specific components from `components["schemas"]`, like a `type Media = components["schemas"].Audio | components["schemas"].Image`, and so forth.

## Testing Instructions

Review the documentation for the package (and the information above) to understand the goals and approach. Then review the code. Due to the nature of the `openapi-fetch` package, there is actually relatively little code to review in this PR. (This was not the case in previous versions which also included a Python version.) Feel free to review the generated code to get a feeling for how it looks.

I've left comments on some changes in the PR that I hope will clarify things I anticipated may have questions asked about. Please feel free to ask for further clarification in any case, of course!

The main code to review carefully is the authentication middleware. This middleware is based on the strategy proven in our own frontend application, [by the `api-token` Nuxt plugin](https://github.com/WordPress/openverse/blob/faf599ba7d1c1973e8e062a19f0b5989588ee13e/frontend/src/plugins/api-token.server.ts). As you read the middleware code, review the test cases for it as well, in the tests included in the package.

You can also check out the branch and try the client out in the unit tests, to see how it feels to work with. Nothing is set in stone, and it is only the "low-level" version. As Olga and I have discussed in the comments on this PR, it's worth considering a future where we have higher level functions like `client.image.search`, and so on, which may have better ergonomics and discoverability than `client.GET("/v1/images/", { params: { query: { q } } })`.

We should aim to use the package inside our frontend (and elsewhere, if appropriate) as much as possible to exercise it and understand what kinds of higher level wrapper functions would be beneficial.

## Checklist

- [N/A] I ran the DAG documentation generator (`./ov just catalog/generate-docs` for catalog

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (commented):
This project is very interesting and I am excited to review this PR once it is undrafted. I also want to discuss the use of the LGPL license because, while I agree with the choice, the reasons for it need to be documented somewhere.

sarayourfriend:
> I also want to discuss the use of the LGPL license because, while I agree with the choice, the reasons for it need to be documented somewhere

Totally! I'd forgotten I kept the LGPL from my original work. It's been a long overdue conversation for us to talk about when/if we'd go in line with WP foundation's GPL usage.

We can keep this MIT in the meantime for flexibility, though.

That's a good conversation for us to consider for the meetup, potentially (depending on how the maintainers all feel about it).

packages/js/api-client/src/client.ts
```diff
@@ -0,0 +1,219 @@
...
+    fetch,
+    getBinaryResponseBody: (res) => (res as Response).arrayBuffer(),
+    getJsonResponseBody: (res) => (res as Response).json(),
+  }
+}
+
+export type OpenverseClient = ReturnType<typeof OpenverseClient>
+export const OpenverseClient = (
+  {
+    baseUrl = "https://api.openverse.engineering/",
```
obulat, line 69:
```suggestion
    baseUrl = "https://api.openverse.org/",
```

Shows the age of the original PR :)

obulat (commented):
I will continue testing this PR, but I have a question regarding the API design:

With this design, for every method call, you have to specify the URL, the version and the request method. I guess the URLs can be autocompleted, but I still think it would be more user-friendly to abstract these details, similar to the original approach used in the frontend. What was the motivation for this more low-level approach?

My suggestion would be to move to something like below:

```
const openverse = OpenverseClient(version=1)

const images = await openverse.images.search({
  params: {
    q: "dogs",
    license: "by-nc-sa",
    source: ["flickr", "wikimedia"],
  },
})

const image = await openverse.images.get({ params: { id: "xxxx" } })
const related = await openverse.images.related({ params: { id: "xxxx" } })
```

obulat:
Sorry for accidentally closing this PR.

sarayourfriend:
I wondered about an API like that as well, Olga. Cloudflare's Python client library has a similar API to that (and types), however they handle versioning fundamentally differently (and indeed require specifying the version every time). For what it's worth, the JS client's API was heavily inspired by [GitHub's client's `request` function](https://github.com/octokit/octokit.js?tab=readme-ov-file#octokitrequest). Their documentation does a great job of highlighting how the string parameter with method and route correspond directly to the OpenAPI documentation.

Leaving aside the inspirations or the approaches others have taken, I disagree with the approach you suggested for the JS client for the following reasons:

1. The Openverse API does not have a versioning policy, but even if we did, I wouldn't expect it to have a static version supporting all endpoints. If we bumped a particular route to v2, for example, I don't think we would bump other routes if they didn't also change. A major change in the image search API is completely separate from the auth token endpoints (both versioned). Using a static version for the client (or even adding a default and facilitating overrides for individual calls) would: potentially imply that versioning _does_ work that, make it non-obvious that a single client instance can support multiple different versions, complicate the code to enable the support to override versions for an individual call (and then select types conditional on the version). We shouldn't try to support a single version per client instance either, because then we'd have to share state between client instances for authentication, a complicated situation already, made even worse if we tried to coordinate multiple clients that for some reason operate on different versions of the authentication API... which would by definition operate differently!
1. From a typing perspective, it's untenable or at the very least much more complicated than the current approach. Grant that we would need to support overriding the version of a particular call at the call site. That means the return type, which currently merely maps the type out of the Routes type using the string the client function is called with, would need to also map that based on the version. Ignoring how the route is determined, it would look like this: `client.search_images({ version: 2 })` and the return type mapped based on the version, either inline or in some super-type like the current Routes type.
1. From a maintainability perspective, it's far more code to write, and it would either be exceedingly verbose (generate every single endpoint as individually nested JavaScript objects and functions, which somehow share state for auth) or highly abstract (some kind of proxy object that builds the route based on the dotted path, and which is typed separately from the code itself that runs the proxy).
1. Finally, it doesn't translate well to any API that would be easy to generate and maintain for the Python clients (this is a minor concern, but a nice-to-have if they roughly align in some way).

Generally (for both the JS and Python client) I disagree that we should optimise for conciseness of calls. The difference in the number of characters required to call `openverse.images.search()` and `openverse("GET /v1/images/")` is effectively 3 characters (ignoring the quotes and the dot in each, which are just syntax). If we used `images.search` for search and `images.get` for the single result, then they are pretty easy to confuse either way, especially if you're reading the REST API's documentation. Using the routes, there is no possibility of confusion of which route you are working with, and translating the REST API documentation to a client call is one-to-one: `openverse("GET /v1/images/")` and `openverse("GET /v1/images/{identifier}/", { identifier: "..." })`.

From a discoverability perspective, TypeScript's language server is perfectly capable of autocompleting these strings and giving you all the options as you type, even if you type fragments.

image

For documentation, there is an easy improvement to make here I only thought of due to this discussion! We can `export delcare function Client(...)` with overloads for each route and docstrings on each endpoint.

I am writing on a flight and will continue writing the rest of my thoughts later. Thanks for the thoughtful suggestion!

sarayourfriend:
One more thing regarding the alignment between the REST API documentation and the JS client's API. If you're _not_ using TypeScript, then the REST API's documentation should be copy/paste-able to create the calls you need. That avoids us needing to write double documentation for each of the nested objects and functions like `.images.get` and so forth, to make the library discoverable without relying on TypeScript's language server.

I think that's pretty valuable.

I will admit, however, that the Python API is rather ugly. The method names are long and admittedly a bit rough to read. At one point I tried to implement the same string-literal based approach in Python, but the language server I was using does not automatically cast strings to `Literal[...]`, so a set of overloads of a single `request` function where the return type corresponds to the first string parameter as a literal value wasn't working very well. I _think_ mypy would support it (maybe even without the casts) and I'm not sure what Pylance or Pycharm do (I use Jedi's language server).

However, while writing this, I remembered trying something along these lines and never really gave it a full try:

```py
from openverse_api_client import Openverse, routes
openverse = Openverse()

openverse.request(routes.v1_images(q="...", source="..."))
```

In this version, the `request` method is overloaded like this:

```py
class Openverse:
    @overload def request(r: routes.v1_images) -> routes.v1_images.Response
    @overload def request(r: routes.v1_audio) -> routes.v1_audio.Response
    def request(r: OpenverseRequest) -> OpenverseResponse:
        ...
```

I did try _something_ like this, and ran into some kind of problem, I just cannot remember what. I may have gotten frustrated at the need for individual overloads instead of being able to map the generic type? And then forgotten to go back and try this again after I realised we would need to write a method signature for every individual route anyway?

Of course the benefit of this is that it removes the need for individual implementations of each method, the single request method just works over the `OpenverseRequest` instance. That would make the generated code (and really the _template_) a bit easier to comprehend and would remove the need to duplicate every single route method's implementation in the async and sync clients.

I would also be possible to eventually add back in overloads for individual routes, so that `routes.v1_images` could be defined like this (never mind the complexity of generating this from the OpenAPI schema, it's a bit of just dreaming at this point, the OpenAPI schema doesn't currently describe the true discriminated-union nature of the search endpoints):

```
class v1_images(OpenverseRequest):
    class Response:
        ...

    @overload def __init__(self, q: str) -> Response
    @overload def __init__(self, collection: Literal["tag"], tag: str) -> Response
    @overload def __init__(self, collection: Literal["source"], source: str) -> Response
    @overload def __init__(self, collection: Literal["creator"], source: str, creator: str) -> Response
```

Of course, there's the same literal-string casting problem again. I guess the other option is:

```
class v1_images(OpenverseRequest):
    class Response:
        ...

    @overload def __init__(q: str) -> Response
    @overload def __init__(tag_collection: str) -> Response
    @overload def __init__(source_collection: str) -> Response
    @overload def __init__(creator_collection: str, source: str) -> Response
```

But that's deviating _far_ from the REST API's documentation, and not even using the same parameter names anymore. Tag is easy, I suppose, because the `tag` parameter is used exclusively for the tag collection. Source can be used on its own and creator+source can be used without the collection parameter ([I've just learned this, I'm not sure what this code is doing different from if you add collection=creator, but I haven't looked yet](https://api.openverse.org/v1/images/?source=wikimedia&creator=Unknown%20author)).

Another idea is to abuse the TypedDicts indexer types, and have the clients inherit from a generic TypedDict where the keys are the `GET /v1/images/` style route names, with the value being a corresponding `Procotol` with `__call__` (and presumably `async def __call__`) defined with the args and return type for the route. Last I checked Jedi wasn't capable of auto-completing strings inside the indexer brackets for TypedDicts, but if other LSPs support that now (or soon), maybe that's the way to go. Then the APIs would align essentially 1-to-1.

```python
class v1_images(Protocol):
    class Response:
        ...

    def __call__(self, q: str, ...) -> Response:
        ...

class OpenverseRoutes(TypedDict, total=False):
    "GET /v1/images/": v1_images

class Openverse(OpenverseRoutes):
    def _request(self, route: v1_images | ...) -> ...:
        ...

    def __getitem__(self, key: str):
        # this is made up code ... maybe something like this is possible?
        return self._request(OpenverseRoutes.__annotations__.get())
```

Anyway, there is a lot of room to explore here. If anything though, I'd like to aim to get the Python client as close to the JS client as possible, in terms of how 1-to-1 the relationship is to the REST API documentation itself.

obulat:
I will take more time reading through your replies and the code, thank you for giving all the details and context. I understand the reasons, although I'm still not the biggest fun of the string-based approach :)

sarayourfriend:
That's fair! We could start with this, it's much simpler overall to program, and then create an alias layer over top it, similar to the GitHub library's API. The scope for this PR definitely needs to be defined _somehow_, a "low-level" version (as you called it) sets a foundation and lets us ignore more complex questions about what to name aliases and such.

sarayourfriend:
For reviewers: I plan to make at least one change, which is to add the various overloaded definitions of the TS client function type so that documentation for each route shows up in the function call (currently does not do that).

I will experiment with the Python client and the idea of using individual classes to construct a request for a particular route, and overloaded definitions to transform the response type accordingly.

I was also wondering, because this is such a large PR to review (and there aren't really good ways of splitting it up), maybe we can review it together at the maintainers' meetup next month? Otherwise, we could do that over a video chat earlier. Just if that would help get this reviewed. It certainly should be given some careful thought, but we aren't committing to anything in particular here other than the methodology. These packages won't be published yet (and the versions that are published are marked alpha anyway) so we can feel free to make any changes we want to align with our goals.

sarayourfriend:
After trying Pyright again (after much time not using it, and only using mypy instead), it turns out Pyright will implicitly cast a literal string passed to a function with overloads of `Literal["whatever"]` and find the right overload. And in fact, mypy will do so as well. So I don't know why I thought it wouldn't. Anyway, I'll get changes up as soon as possible to make the Python clients align with the "low-level" API of the JS client.

sarayourfriend:
Okay, I got _something_ like what I wanted, but not quite. Annoyingly, whilst mypy and Pyright will both select the correct override when the `endpoint` is the single argument, they will not do so when there are additional positional or key-word arguments. Instead, `client.endpoint` takes the endpoint string and returns a class whose constructor signature has the path, query, and body parameters, and you pass the instance of _that_ class to `client.request`.

```py
search_result = client.request(
    client.endpoint("GET /v1/images/")(q="dogs", source="flickr", page=2),
    headers={"User-Agent": "CustomApp/1.0"},
)
```

I don't really like this. At all. I can't remember if I tried overloads on `request` with the parameters, so you wouldn't need to do the awkward call off the result of `endpoint()` with the request parameters. Maybe it would be cleaner if `endpoints` was just a TypedDict and the syntax was `client.endpoints["GET /v1/images/"](q="dogs", ...)`? It at least feels more normal, and is effectively the same, perhaps even simpler because `endpoint` as a method requires a list of overloads, but as a TypedDict, that wouldn't be required. I'll try that in the next commit.

sarayourfriend:
I'm converting this back to a draft. I want to refocus this PR a bit by making it explicitly interested solely in implementing the "low-level" version, and facilitating a higher-level API built upon it. To me, that means making the following changes:

1. Go back to a dedicated `request` method on the Python client, and implement an explicit function named `request` on the JavaScript client. The latter will include converting the JS client to a class, rather than this weird pseudo-constructor function I for some reason used. Using a class with a `request` method will make it easier to implement any kinds of higher-level APIs over top that `request` method within the context of the class, just as we can do with the Python client.
2. Explore splitting the method and route into separate strings, and in Python, use `StrEnum`s for both (`http.HTTPMethod` for the method, of course).
3. Clean up the OpenAPI schema parsing code to use new `NamedTuples` for each OpenAPI schema type rather than Python _types_. This should make following the code a _lot_ simpler. I tried using <https://pypi.org/project/openapi-type/> for this, but it didn't immediately work, and required some potentially significant changes to support 3.0.3 and even some features of OpenAPI that we use which are technically optional, and therefore unimplemented in `openapi-type`. I forked it and briefly tried to make some of these changes, but the complexity of that library (it uses `typeit` to parse the schema) felt over-the-top for just handling the current OpenAPI schema features that the Openverse REST API uses. We don't need to implement a _general_ OpenAPI schema parser, just the subset of one that will handle the Openverse REST API.

.vale/styles/Openverse/TermCasing.yml
```diff
@@ -5,17 +5,19 @@ message: "Incorrect casing. Use '%s' instead of '%s'."
 level: error
 ignorecase: false
 scope:
-  - text
+  # Non-code sentence level text OR headings
+  - summary
+  - heading
```
sarayourfriend, line 10: Changes in this file I noticed while writing and editing the documentation for the new package. These should be more reliable and avoid false positives like variable identifiers inside of code blocks (which is what I ran into that prompted me to look into it).

.pre-commit-config.yaml
```diff
@@ -122,11 +122,11 @@ repos:
 
       - id: eslint
         name: eslint
-        files: (frontend|automations|packages/js).*?\.(js|ts|mjs|vue|json5|json)$
+        files: (frontend|automations|packages/js).*?\.(js|ts|mjs|vue|json5?)$
         "types": [file] # ESLint only accepts [javascript] by default.
         language: system
         pass_filenames: false
-        entry: bash -c 'pnpm run eslint --fix'
+        entry: just eslint
```
sarayourfriend, line 129: See the note on the new just recipe for an explanation of why I've moved this from the root package.json to just. TL;DR: just has better default argument handling than package.json scripts, so it's more convenient!

package.json
```diff
@@ -10,17 +10,16 @@
...
-    "preeslint": "pnpm --filter '@openverse/eslint-plugin' run build",
-    "eslint": "eslint --ext .js,.ts,.vue,.json,.json5 --ignore-path .gitignore --ignore-path .eslintignore --max-warnings=0 frontend automations/js packages/js .pnpmfile.cjs .eslintrc.js prettier.config.js tsconfig.base.json"
+    "prepare:nuxt": "pnpm --filter frontend run prepare:nuxt"
   },
   "packageManager": "pnpm@9.5.0",
   "engines": {
     "node": ">= 20.0.0 <21"
   },
   "devDependencies": {
+    "@types/node": "22.5.1",
```
sarayourfriend, line 20: `@types` shared by multiple libraries should ideally live at the workspace root, to ensure they have the same versions. Otherwise, you can run into nasty errors like `type Plugin is not compatible with argument type Plugin`, where each `Plugin` reference is just referencing a different version of the type by the same name from the same package. This only holds for devDependencies.

packages/js/api-client/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "author": "Openverse Contributors <openverse@wordpress.org>",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/WordPress/openverse",
+    "directory": "packages/js/openverse-api-client"
+  },
+  "license": "MIT",
+  "devDependencies": {
+    "esbuild": "^0.23.1",
+    "nock": "14.0.0-beta.11",
```
sarayourfriend, line 40: Nock only supports fetch as of version 14, which is still in beta. It works fine for these test cases, so I don't think it will be an issue to be pinned to a pre-release version for now.

packages/js/api-client/src/auth.ts
```diff
@@ -0,0 +1,191 @@
...
+   * can proceed with the existing token, as it is still valid.
+   *
+   * However, if the token is expired (or no token yet exists), the request
+   * must be held until the client has an API token.
+   *
+   * This is only relevant to clients configured with credentials. Non-credentialed
+   * clients bypass this process altogether.
+   *
+   * @returns Whether the client should wait.
+   */
```
sarayourfriend, line 89: Understanding this flow might be helped by reading the tests that exercise it, so check there first if you aren't following the approach.

As mentioned in the PR description, the same methodology is proven in the frontend's `api-token` plugin, except there it blocks the full SSR request until the token returns, rather than just individual API requests. The principle and effect is the same, and I managed to sort out an implementation that works without needing to rely on the third-party async-mutex library like we use in the frontend.

packages/js/api-client/src/types.ts
```diff
@@ -0,0 +1,12 @@
+import type { Client } from "openapi-fetch"
+
+import type { paths } from "./generated/openverse"
+
+export type { components } from "./generated/openverse"
```
sarayourfriend, line 5: Exporting components lets consumers of the package reference specific models. For example, to get the Image model (say, to define it in a store), you would use `components["schemas"].Image`, with `components` imported from `@openverse/api-client`.

packages/js/api-client/tsconfig.prod.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["./tests/**", "./vitest.config.mts"]
+}
```
sarayourfriend, line 4: This tsconfig is used when building types for the package. The excludes ensure that only the types appropriate for distribution are included in the package's type output.

documentation/conf.py
```diff
@@ -110,6 +110,10 @@ def add_ext_to_path():
     "catalog/guides/deploy": "/meta/release_and_deployment/runbooks/catalog.html",
     "catalog/guides/deployment": "/meta/release_and_deployment/index.html#catalog",
     "ingestion_server/guides/deploy": "/meta/release_and_deployment/runbooks/ingestion_server.html",
+    "packages/eslint_plugin/index": "/packages/js/eslint_plugin/index.html",
+    "packages/eslint_plugin/analytics_configuration": "/packages/js/eslint_plugin/analytics_configuration.html",
+    "packages/eslint_plugin/no-unexplained-disabled-test": "/packages/js/eslint_plugin/no-unexplained-disabled-test.html",
+    "packages/openverse_attribution/index": "/packages/python/openverse_attribution/index.html",
```
sarayourfriend, line 116: I originally moved these into language-specific sub-paths to accommodate the two `api-client` packages (one `@openverse/api-client` and the other `openverse-api-client`). That's no longer present in _this_ PR, but because we know we want a Python API client, this conflict will happen soon. In that case, I'd rather move these to language specific sub-paths sooner than later, thereby reducing the overall number of redirects we have to add in the future. This would also apply if we created a separate `@openverse/attribution` JS package, for example, and other things of that nature where we need both a JS and Python version.

.pre-commit-config.yaml
zackkrida: Great idea!

packages/js/api-client/src/types.ts
zackkrida: I think the paths should be exported as well, for example to allow the production of types like:

```ts
    type Method = paths["/v1/images/{identifier}/"]["get"]
    type SuccessResponse = Method["responses"][200]
    type SuccessBody = SuccessResponse["content"]["application/json"]
```

which can be quite useful to consumers who want to create type-safe client.GET arguments outside of the function call, as an example use case.

zackkrida (approved):
This is excellent, and looks great. I'm only blocking on the suggestion that we also export the `path` types to users. I think access to those will be important to any consumers writing type-safe applications using this client. 

Really excellent work! I so look forward to when we implement this on the frontend and delete a lot of code.

zackkrida (changes requested):
Apologies, I meant to request changes. See my first comment for details.

sarayourfriend:
Sounds good, thanks for the review @zackkrida! I'll add exporting the paths.

sarayourfriend:
@zackkrida `paths` export added in https://github.com/WordPress/openverse/pull/4538/commits/9da282d5a17c0a1a4245f77fc4c6ea24e7b69e76

zackkrida (approved):

Issue #3840, opened 2024-02-28
Create bulk moderation actions


stacimc:
## Description

Blocked by https://github.com/WordPress/openverse/issues/3636

Create the custom Django admin actions for bulk moderation decisions:
* bulk deindex sensitive
* bulk deindex copyright
* bulk mark sensitive
* bulk reverse mark sensitive
* bulk reverse deindex

Details can be found in the IP [here](https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20240122-implementation_plan_bulk_moderation_actions.html#actions).

Actions will be added to the appropriate views in subsequent issues.

sarayourfriend:
Adding a note here that we should make sure _not_ to rely on the existing methods for marking sensitive or deleting media, as those _do not do bulk operations_ and are wildly inefficient if used for a bulk operation (e.g., if used in a `for` loop over a set of sensitive or deleted media instances).

PR #4544, opened 2024-06-24
Create sensitive and deleted media models for decisions

### LINKED ISSUE #4513: Create `MediaDecision` have no effect on deindexed actions
## Description

Currently, when trying to deindex a reported item, nothing happens beyond creating the record of the decision. In these cases, items are supposed to be deindexed and removed from the media table, saving the identifier in the corresponding deleted media table. This function is critical to remove inappropriate content from Openverse and decisions created while this is done will need to be backfilled.

## Reproduction

1. Report an image (or audio)
2. Select the action to deindex it via the Django admin
3. Verify the deleted images and see the table is empty
4. Also observe the image is still available in search and in its details endpoint

## Record

https://github.com/WordPress/openverse/assets/9145885/4899ccf5-1e29-40e6-b892-696af0abad74

## Additional context

Related to project #383.

dhruvkb:
## Fixes

Fixes #4513 by @krysal 

## Description

This PR ensures that `SensitiveImage`/`SensitiveAudio` and `DeletedImage`/`DeletedAudio` models are created for every decision.

## Testing Instructions

0. Create a report for an image.
1. Take a deindexed-sensitive or deindexed-copyright action on the report.
   - You should see a new `DeletedImage` object.
   - The image should be deleted.
   - The report should be resolved with a linked decision.
2. Create a report for an image.
3. Take a marked-sensitive action on the report.
   - You should see a new `SenstiveMedia` object.
   - The image should be marked as sensitive.
   - The report should be resolved with a linked decision.

Repeat these steps for audio.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound (approved):
Makes sense, and looks good to me! I tested this locally and confirmed that the deindexed image gets deleted from the API database, and the sensitive image gets a record in the sensitive table.

api/api/management/commands/backfillmoderationdecision.py
```diff
@@ -92,14 +92,18 @@ def handle(self, *args, **options):
...
                     MediaDecisionThrough(
                         media_obj_id=report.media_obj_id, decision=decision
                     )
                     for report, decision in zip(reports_chunk, decisions)
                 ]
             )
+            # Bulk create does not call ``save()`` so the mark-sensitive/deindex
+            # actions will not be undertaken.
+            for through in throughs:
+                through.perform_action()
```
sarayourfriend, line 106: This loop is a disappointing solution to the backfill, and somewhat defeats the purpose of the bulk creates above :confused: They're still useful as is, but this really goes against that effort by executing an individual query for each through instance regardless, especially because `perform_action` not only does the queries to create the deleted/sensitive model instances, but also relies on `self.decision.action`, which implicitly queries for `self.decision`, as well as `self.media_obj` which implicitly queries the entire media object. So the `perform_action` function is potentially very inefficient, especially when we consider it in light of the known case of bulk moderation actions, which must be able to operate on many records on an ongoing basis.

We can avoid the `self.media_obj` query within the method pretty easily. `self.decision.action` in this particular case (and potentially all cases where it would be relevant) is actually implicitly avoided (though honestly only accidentally, and is only an accident, not necessary) because the model will already have `decision` loaded from when it was created on line 98. I don't think we should rely on that implicit (accidental) fact to avoid the implicit query.

Instead, we should go ahead and make `perform_action` presume a bulk operation, and can be trivially adapted to do a single record more easily if we wanted (though I don't think it's necessary). Because we know decisions will be created in bulk, and because the bulk operation should perform the same as the non-bulk operation, I think we need to go ahead and solve that, potentially even down to the ES level. Unfortunately, that's a big task, and probably worth separating into a new PR instead of including all the fixes here.

Maybe something like this would work:

```
MediaObjectIdentifiers = namedtuple("identifier", "pk")
def decide(
    media_object_identifiers: list[MediaObjectIdentifiers],
    report_pks: list[int],
    action: str,
):
    # create decision
    # bulk update the reports
    # bulk create the through records (using `decision_id=decision.pk` rather than the full decision object)
    # bulk create the appropriate sensitive/deleted media instances based on the action
    # bulk update the filtered and non-filtered indexes using `Elasticsearch::bulk`
```

Anyway, I think either we should do that in this PR, or leave out this fix to the backfill and address it in a separate PR. I am strongly against introducing this kind of operation into the codebase, especially if it could accidentally be taken as an approved pattern to use when implementing the actual bulk decision creation.

api/test/unit/management/commands/test_backfillmoderationdecision.py
```diff
@@ -78,12 +80,13 @@ def test_create_moderation_decision_for_reports(
...
     decision = MediaDecision.objects.first()
-    assert decision.media_objs.count() == 1
     assert decision.action == expected_action
     assert decision.moderator.username == username
 
+    assert MediaDecisionThrough.objects.filter(decision=decision).count() == 1
+
     decision_through = MediaDecisionThrough.objects.first()
-    assert decision_through.media_obj == report.media_obj
+    assert str(decision_through.media_obj_id) == report.media_obj_id
```
sarayourfriend, line 89: Aren't these statements equivalent? Why the change here? When would they not be equivalent? Django models essentially compare equality through `self.pk == other.pk`: https://github.com/django/django/blob/main/django/db/models/base.py#L593-L601

api/api/models/media.py
```diff
@@ -361,6 +363,22 @@ class AbstractMediaDecisionThrough(models.Model):
...
+        """Perform the action specified in the decision."""
+
+        if self.decision.action in {
+            DecisionAction.DEINDEXED_SENSITIVE,
+            DecisionAction.DEINDEXED_COPYRIGHT,
+        }:
+            self.deleted_media_class.objects.create(media_obj=self.media_obj)
+
+        if self.decision.action == DecisionAction.MARKED_SENSITIVE:
+            self.sensitive_media_class.objects.create(media_obj=self.media_obj)
```
sarayourfriend, line 376: Using `self.media_obj` will query the database, and isn't necessary here. (`_perform_es_action` queries it anyway to get the database identifier [truly an absurd and confusing distinction I will never understand why we don't do everything we can to get rid of], but we don't need to repeat bad patterns/habits and there it's unavoidable, unlike here where we can easily avoid it.)

```suggestion
        if self.decision.action in {
            DecisionAction.DEINDEXED_SENSITIVE,
            DecisionAction.DEINDEXED_COPYRIGHT,
        }:
            self.deleted_media_class.objects.create(media_obj_id=self.media_obj_id)

        if self.decision.action == DecisionAction.MARKED_SENSITIVE:
            self.sensitive_media_class.objects.create(media_obj_id=self.media_obj_id)
```

api/api/models/media.py
```diff
@@ -361,6 +363,22 @@ class AbstractMediaDecisionThrough(models.Model):
     class Meta:
         abstract = True
 
+    def perform_action(self):
+        """Perform the action specified in the decision."""
+
+        if self.decision.action in {
```
sarayourfriend, line 369: A suggestion to avoid the implicit `self.decision` query when calling `perform_action` directly (if `save` doesn't run). This will be necessary when we implement the bulk decisions anyway.

```suggestion
    def perform_action(self, action=None):
        """Perform the action specified in the decision."""
        
        action = self.decision.action if action is None else action
        
        if action in {
```

sarayourfriend (changes requested):
The blocking change is to fix the issue with non-performant bulk decision creation.

I think we should exclude fixing the backfill from this PR and instead just fix the media admin so that it's working again. And then address the backfill (and lay groundwork for #3840, which needs this anyway) in a separate PR.

obulat (dismissed):
After you delete a media object using a `deindexed_[...]` action, admin is trying to open the same `change_view` form with the same media object. However, because this media object has been deleted, you get an error on line `if tags := media_obj.tags:`. To prevent it, I added the redirect to the `change_view`:
```
        media_obj = self.get_object(request, object_id)
        if media_obj is None:
            return redirect(f"admin:api_{self.media_type}_changelist")
```

If you do not select any report, and submit a decision (e.g., `mark_sensitive`), you get no indication of the error except for the warning in the logs (which the moderator using the Admin UI will probably not see). This can be a follow-up issue since this PR is critical, but we should add the error display ("report_id" is required) to the form.

api/api/admin/media_report.py
```diff
@@ -440,6 +440,10 @@ def decision_create_view(self, request, object_id):
         if request.method == "POST":
             media_obj = self.get_object(request, object_id)
 
+            through_model = {
+                "image": ImageDecisionThrough,
+                "audio": AudioDecisionThrough,
+            }[self.media_type]
```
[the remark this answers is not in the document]
dhruvkb: The reason I write use the `= {}[]` pattern is because it is the most compact among all the alternatives and also raises an exception when none of the keys match the input.

I've done this a few times in this file itself. Would you prefer I change this pattern across the entire file, or keep this in in the interest of consistency, or just change it here?

api/api/models/media.py
dhruvkb: Updated in fb8495f.

api/api/models/media.py
dhruvkb: Updated in fb8495f.

api/api/management/commands/backfillmoderationdecision.py
dhruvkb: I've reverted the changes to the backfill command to implement this separately.

api/api/admin/media_report.py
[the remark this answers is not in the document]
sarayourfriend: If the dict-key-access approach is preferred for any reason at all (it is nice it implicitly throws if the key doesn't exist), I'd at the very least think defining the dictionary outside the runtime scope of the function is a reasonable requirement, if only so that it's in a shared location. It's far-fetched to me to establish a pattern of defining otherwise static dictionaries, especially one encoding relationships between static objects, entirely dynamically in the runtime of a function. From a performance perspective it's fine here, but in a tight loop it's just silly, right? From a shared data/relationship encoding perspective (and discoverability, clarity, etc) it's definitely the worst option I can think of :confused:. Just seems like an antipattern to me :shrug: I also don't think compactness is necessarily a virtue, and certainly not in Python, which actively resists compactness in my experience.

sarayourfriend:
@stacimc pointed out that the backfill doesn't need to "perform" the action at all, because it's just creating the decisions for actions that have already been performed. Glad we removed it already! I'll re-review this today.

sarayourfriend (approved):
LGTM, tests fine locally and I'm unblocking on my requested changes to get the fix into the admin.

Issue #4517, opened 2024-06-19
`ov` pre-commit hooks do not respond to sigterm (control + c)


sarayourfriend:
## Description

git pre-commit hooks run with ov do not respond to sigterm. This doesn't occur when running pre-commit directly in ` ov` (i.e., `ov j lint`).

## Reproduction

1.  Make a change and commit it. When the pre-commit hook start, send `control + c` to try to cancel it right away and see that it never gets cancelled.

PR #4556, opened 2024-06-26
Revert "Changes all sensible occurrences of the just commands to have them run using ov"

sarayourfriend:
I'm temporarily reverting this PR. Until we are confident `ov` is 100% ready to be _the_ Openverse development environment (as in, the only one that is supported), and until its general usage is documented in our getting started/entry point documentation, we should hold off on downstream documentation referencing it.

I fully expect to re-revert these changes in the new future, though, once `ov` is deemed ready by the Openverse maintainers.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_source_suggestion.yml
+++ b/.github/ISSUE_TEMPLATE/new_source_suggestion.yml
@@ -71,7 +71,7 @@ body:
         - The script must not use anything from `catalog/dags/retired/providers/provider_api_scripts/modules/etlMods.py`, since that module is deprecated.
         - If the provider API has can be queried by 'upload date' or something similar, the script should take a `--date` parameter when run as a script, giving the date for which we should collect images. The form should be `YYYY-MM-DD` (so, the script can be run via `python my_favorite_provider.py --date 2018-01-01`).
         - The script must provide a main function that takes the same parameters as from the CLI. In our example from above, we'd then have a main function `my_favorite_provider.main(date)`. The main should do the same thing calling from the CLI would do.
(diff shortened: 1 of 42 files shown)


dhruvkb:
I think it is fine to recommend `ov` usage. `ov` works perfectly fine for me and I am exclusively using it locally already, and I feel the revert (and soon revert-revert) is not necessary.

sarayourfriend:
Well, I'd like to confirm that all Openverse maintainers understand and can use `ov`. Based on notes I saw from the discussion between you, @obulat and @krysal, there might still be some confusion and testing out to do.

I am eager to get it merged as the official solution, and it also works almost perfectly for me (still some usability issues like #4517 but those are fixable and non-blocking). I'd _really_ like to make sure it works for WSL or for us to just say we don't support Windows based development in any form whatsoever.

@obulat had trouble using it on WSL and a contributor tried it on WSL and also had problems they shared in another issue. Considering how troublesome out development environment has been in the past, I want to make sure it's viable if we're going to call it the only development environment we support.

I am keen to get there though, especially because of the vast number of improvements we can make to scripts that currently use roundabout or hacky scripts to avoid using tools we'll be able to assume we have from now on.

zackkrida:
Just for another single data point, I personally am using `ov` locally without issue.

sarayourfriend:
Okay. I'll go ahead and close this PR and send a ping to the Openverse maintainers internally and see if anyone objects to calling `ov`  _the_ solution for development environment, and dropping support immediately for any other approaches, pending documentation changes.

My only concern really is WSL, and if we're not too worried about that breaking (maybe even if only temporarily) then we can go ahead and move forward anyway.

PR #4572, opened 2024-06-29
Add distributed reindex steps

### LINKED ISSUE #4148: Implement local distributed reindexing
## Problem

Blocked by #4146, #4147, https://github.com/WordPress/openverse-infrastructure/issues/849

This issue tracks adding the orchestration steps for the distributed reindex to the new data refresh DAGs.

## Description

In this step we will add tasks to the data refresh DAGs to orchestrate the distributed reindex. At the end of this step, it will be possible to run a distributed reindex locally, but because the infrastructure work to create the ASGs is not complete, it can not be run on production yet. The following code can all be refactored from [distributed_reindex_scheduler.py](https://github.com/WordPress/openverse/blob/main/ingestion_server/ingestion_server/distributed_reindex_scheduler.py).

- Use [describe_auto_scaling_groups](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/describe_auto_scaling_groups.html) and filter by tags to select the appropriate ASG for the desired environment. (Skips in local env.)
- Use [set_desired_capacity](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/set_desired_capacity.html) to increase the desired capacity of the ASG to the desired number of workers, depending on the environment. This will cause the ASG to begin spinning up instances. (Skips in local env.)
- Use [describe_auto_scaling_groups](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling/client/describe_auto_scaling_groups.html) to poll the ASG until all instances have been started, and get the EC2 instance IDs. (Skips in local env.)
- Use dynamic task mapping to distribute reindexing across the indexer workers by first calculating start and end indices that will split the records in the media table into even portions, depending on the number of workers available in the given environment. Then:
  - POST to each worker’s `reindexing_task` endpoint the `start_index` and `end_index` it should handle
  - Use a Sensor to ping the worker’s `task/{task_id}` endpoint until the task is complete, logging the progress as it goes
- Use [terminate_instance_in_auto_scaling_group](https://boto3.amazonaws.com/v1/documentation/api/1.26.86/reference/services/autoscaling/client/terminate_instance_in_auto_scaling_group.html) to terminate the instance. Make sure to set `ShouldDecrementDesiredCapacity` to `True` to ensure that the ASG does not try to replace the instance. This task should use the [NONE_SKIPPED TriggerRule](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules) to ensure that the instances are terminated, even if there are upstream failures. (Skips in local env.)
- Finally, after all tasks have finished (regardless of success/failure), we should have a cleanup task that calls `set_desired_capacity` to 0. Generally this should be a no-op, but if an instance crashes during reindexing (rather than simply failing during reindexing) the ASG will spin up a replacement and Airflow will not automatically clean it up. This task ensures that any dangling instances are terminated.

## Additional context

See [this section](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html#implement-distributed-reindexing-locally) of the IP.

stacimc:
## Fixes

Fixes #4148 by @stacimc 

## Description

Adds steps to the new data refresh DAGs to create the temporary ES index (which will later be promoted), and populating it via the distributed reindex:

Screenshot 2024-08-19 at 4 30 34 PM

## Testing Instructions

`./ov just api/init && ./ov just up` and then run a few DAGs in order to ingest some new data (I used Jamendo for audio and Cleveland for image, and manually marked them as success as soon as some records were recorded). Then run the new data refresh DAGs (eg `staging_audio_data_refresh`). They should run successfully, but skip the tasks to create/terminate EC2 instances. Go to [elasticvue](https://app.elasticvue.com/cluster/0/indices) and you should see a new index matching the output of the `generate_index_name` task.

It should have no aliases. If you refresh the index in Elasticvue, you'll also see that it has the expected number of records (so for example, if you started from sample data the main audio index will have 5000 records. If you ingested 100 records from Jamendo before running the data refresh, your new temp index will have 5100 records.)

Run `./ov just api/pgcli` and `select count(*) from audio` (or image, depending on which data refresh you ran). Then verify the count is the same as the number of docs in the new index in Elasticvue.

Also note that you'll see somewhat unexpected behavior if you run the DAGs multiple times in a row right now, because we don't yet delete the temp table. That's expected! If you need to run it a second time, just drop the temp import tables manually in the API DB first.

It's not simple to test this against the staging environment when running locally because of the requirement that we skip these tasks locally, but I did so by configuring the `AIRFLOW_CONN_AWS_DEFAULT` locally and commenting out the logic to skip. This should technically be safe even in production because no data is destroyed or promoted, but if you do want to test this you should ONLY run the staging DAGs to be safe, and we should clean up the dangling tables/indices afterward. There are additional issues in the milestone for testing this formally in those environments so I do not think it is a requirement for testers here.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (commented):
I tried running this DAG locally. The DAG created the index, but it was empty. After some investigation I saw that the `id` in my `temp_import_audio` start with 20001, not 1. Could that be the reason.
Another question: indexer logs say that the data was copied to ES from `audio` table. Shouldn't we copy from the temporary table?

AetherUnbound:
> I tried running this DAG locally. The DAG created the index, but it was empty.

I also tried and noticed this as well - I think it was because the index was not refreshed after the documents were added. I hit "Refresh index" in the Elasticvue UI and the 5000 documents showed up!

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+        target_environment=target_environment,
+    )
+
+    trigger_reindexing_task = TempConnectionHTTPOperator(
+        task_id="trigger_reindexing_task",
+        http_conn_id=worker_conn,
+        endpoint="task",
+        data={
+            "model_name": data_refresh_config.media_type,
+            "table_name": data_refresh_config.media_type,
```
AetherUnbound, line 350: Per @obulat's note, this should probably be the temporary table rather than the base table, right?

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    aws_conn_id: str = AWS_CONN_ID,
+):
+    """Perform the distributed reindex on a fleet of remote indexer workers."""
+    estimated_record_count = PGExecuteQueryOperator(
+        task_id="get_estimated_record_count",
+        conn_id=OPENLEDGER_API_CONN_ID,
+        sql=dedent(
+            f"""
+            SELECT id FROM {data_refresh_config.media_type}
+            ORDER BY id DESC LIMIT 1;
```
AetherUnbound, line 402: This will need to be the temporary table too I think. I'd also be worried about this not being super performant, but I guess we can see how long it takes in production before trying to do something hackier!

AetherUnbound (changes requested):
This is a huge bolus of code, great work on getting this off the ground Staci! I've got a number of comments here, the most blocking one being what @obulat pointed out - we should probably be referencing the temp table on the API rather than the base one for creating the indices. Other than that, I have a few code structure comments, but nothing else major!

AetherUnbound:
One other thought here - given our recent strategy around PR review requirements, it would be great to have some new tests for all of the relevant pieces here alongside this PR. I know much of this is using operators, but for any of the pieces that you think make the most sense :smile

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    if environment != PRODUCTION:
+        raise AirflowSkipException("Skipping instance creation in local environment.")
+
+    ec2_hook = EC2Hook(aws_conn_id=aws_conn_id, api_type="client_type")
+    launch_templates = ec2_hook.conn.describe_launch_templates(
+        LaunchTemplateNames=INDEXER_LAUNCH_TEMPLATES.get(target_environment)
+    )
+
+    if len(launch_templates.get("LaunchTemplates")) == 0:
+        raise Exception("Unable to determine launch template version.")
```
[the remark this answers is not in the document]
sarayourfriend: For what it's worth, boto3 raises an exception in this case anyway:

```
>>> ec2_client.describe_launch_templates(LaunchTemplateNames=["hello"])
  File "<stdin>", line 1, in <module>
  File "/home/sara/projects/openverse-infrastructure/.venv/lib/python3.12/site-packages/botocore/client.py", line 565, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sara/projects/openverse-infrastructure/.venv/lib/python3.12/site-packages/botocore/client.py", line 1021, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InvalidLaunchTemplateName.NotFoundException) when calling the DescribeLaunchTemplates operation: At least one of the launch templates specified in the request does not exist.
```

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+def get_launch_template_version(
+    environment: str,
+    target_environment: Environment,
+    aws_conn_id: str = AWS_CONN_ID,
+):
+    """
+    Get the latest version of the launch template. Indexer workers will all be created with this
+    version. Importantly, this allows us to retry an individual indexer worker and ensure that
+    it will run with the same version of the indexer worker as the others, even if code has been
+    deployed to the indexer worker in the meantime.
```
sarayourfriend, line 136: For posterity, might be worth mentioning that this retriability an intentional reason not to use the `$Latest` alias for launch template versions so that every request to use the launch template during this run will use the same launch template version.

This could say, "Get the latest version of the launch template at time of distributed reindex start", or something, to clarify that "latest" is relevant to a particular point in time. :thinking: 

Just a nit though, it might be over complicating it to try to explain it in this much detail, you've already covered the important bits.

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    ec2_hook = EC2Hook(aws_conn_id=aws_conn_id, api_type="client_type")
+    result = ec2_hook.conn.describe_instance_status(InstanceIds=[instance_id])
+
+    instance_status = result.get("InstanceStatuses", [])[0]
+    state = instance_status.get("InstanceState", {}).get("Name")
+    status = next(
+        status.get("Status")
+        for status in instance_status.get("InstanceStatus", {}).get("Details", [])
+        if status.get("Name") == "reachability"
+    )
```
sarayourfriend, line 215: Does a reachability status always exist? And if so, does that mean we shouldn't need those defaults in the `.get` method calls? Otherwise we'll get a `StopIteration` thrown here if the comprehension creates a no-value generator:

```
>>> next(status.get("Status") for status in {}.get("InstanceStatus", {}).get("Details", []) if status.get("Name") == "reachability")
  File "<stdin>", line 1, in <module>
StopIteration
```

Either `StopIteration` needs to be anticipated and caught, or the default values need to be safe, or we should pass `next` a default. For example, `.get("Details")` could default to:

```
.get("Details", {"Name": "reachability", "Status": "unknown"})
```

Or better yet, use `next`'s optional default argument:

```suggestion
    status = next(
        (
            status.get("Status")
            for status in instance_status.get("InstanceStatus", {}).get("Details", [])
            if status.get("Name") == "reachability"
        ),
        None,
    )
```

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    ec2_hook = EC2Hook(aws_conn_id=aws_conn_id, api_type="client_type")
+    reservations = ec2_hook.describe_instances(instance_ids=[instance_id]).get(
+        "Reservations"
+    )
+
+    # `Reservations` is a list of dicts, grouping instances by the launch request that
+    # started them. Because we are querying a single InstanceId we expect to have one
+    # Reservation.
+    if not (len(reservations) == 1 and len(reservations.get("Instances", {})) == 1):
+        raise Exception("Unable to describe worker instance.")
```
sarayourfriend, line 242: Here again, boto3 should have you covered. Here's an example of calling `describe_instances` with an instance ID that doesn't exist:

```
>>> ec2_client.describe_instances(InstanceIds=["i-0de0dc5cd17fc5f2c"])
  File "<stdin>", line 1, in <module>
  File "/home/sara/projects/openverse-infrastructure/.venv/lib/python3.12/site-packages/botocore/client.py", line 565, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sara/projects/openverse-infrastructure/.venv/lib/python3.12/site-packages/botocore/client.py", line 1021, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (InvalidInstanceID.Malformed) when calling the DescribeInstances operation: Invalid id: "i-0de0dc5cd17fc5f2c"
```

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    environment: str,
+    target_environment: Environment,
+    aws_conn_id: str = AWS_CONN_ID,
+):
+    """
+    Trigger a reindexing task on a remote indexer worker and wait for it to complete. Once done,
+    terminate the indexer worker instance.
+    """
+
+    launch_template_version = get_launch_template_version(
```
sarayourfriend, line 314: Because launch template versions are full resources, it may be helpful to disambiguate the names for this and the task the way AWS itself does, by specifying that it's a version _number_, rather than a full launch template version document.

sarayourfriend (changes requested):
Had some time to review this, so I've left some requested changes. Namely, the blockers for me are:

1. Use launch template version _number_ to reference what is retrieved in the `get_launch_template_version` task, so that it is disambiguated from the launch template version resource: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_launch_template_versions.html
2. Move retrieval of the launch template version number out of the mapped task group to avoid the race condition possible if each task group retrieves the version number independently of the others
3. Add a default value to the `next` call to avoid an unhandled StopIteration, or, if there are always values, then replace the `get` calls with indexer syntax and add a comment explaining why it is safe to proceed without default values.

Otherwise, I agree with Madison's comments, but the rest of mine are not blockers, just suggestions or extra context/information to help you make choices if you want.

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+from data_refresh.data_refresh_types import DataRefreshConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+WORKER_CONN_ID = "indexer_worker_{worker_id}_http_{environment}"
+
+
+class TempConnectionHTTPOperator(HttpOperator):
```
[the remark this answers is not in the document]
stacimc: I actually _meant_ for temp to stand for temporary, as these are temporary connections that are created and destroyed when the EC2 instance we're connecting to is destroyed! 🤣 However it does happen to also be templated, so if we want to pull these out to be reusable then changing the name to `TemplatedConnectionHTTPOperator` is a better name because it's not so specific to this use case.

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,425 @@
...
+    template_fields: Sequence[str] = (
+        "endpoint",
+        "request_params",
+        "headers",
+        # Extended to allow templating of conn_id
+        "http_conn_id",
+    )
+
+
+def response_filter_status_check_endpoint(response: Response) -> str:
```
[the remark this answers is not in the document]
stacimc: ^ Exactly, I don't want there to be any disentangling/refactoring involved in removing the ingestion server code.

catalog/dags/data_refresh/distributed_reindex.py
stacimc: My understanding was that there _can_ be more than one reservation, which is what I was trying to catch -- but actually I'm only passing one instance ID so that isn't necessary either! I think this was necessary in the previous design where I was describing multiple instances at once 😅 Awesome, good to get rid of some potentially confusing checks.

catalog/dags/data_refresh/distributed_reindex.py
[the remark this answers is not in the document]
AetherUnbound: Oh jeez 😆 Yea if you're okay with that, I think pulling them out would be ideal!

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -27,6 +25,10 @@
     PRODUCTION,
     Environment,
 )
+from common.operators.TemplatedConnectionHttpOperator import (
+    TemplatedConnectionHttpOperator,
+)
+from common.sensors.TemplatedConnectionHttpSensor import TemplatedConnectionHttpSensor
```
AetherUnbound, line 31: Tiniest nit: It seems a bit wonky to have a `CapsCase` module import with then a `CapsCase` class import 😅 Do you think we could change these to the following?

```suggestion
from common.operators.http import TemplatedConnectionHttpOperator
from common.sensors.http import TemplatedConnectionHttpSensor
```

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -35,37 +37,21 @@
 logger = logging.getLogger(__name__)
 
 
-WORKER_CONN_ID = "indexer_worker_{worker_id}_http_{environment}"
-
-
-class TempConnectionHTTPOperator(HttpOperator):
+def setup_ec2_hook(func: callable) -> callable:
```
AetherUnbound, line 40: Love this!

AetherUnbound (approved):
This is fantastic, we're getting closer to having this whole thing complete!! 🤩 Just a few small things but nothing blocking, local testing worked as expected 🚀

catalog/dags/data_refresh/data_refresh_types.py
```diff
@@ -96,17 +101,22 @@ def __post_init__(self):
         self.dag_id = f"{self.media_type}_data_refresh"
         self.filtered_index_dag_id = f"create_filtered_{self.media_type}_index"
 
+    def table_mappings(self) -> list[TableMapping]:
```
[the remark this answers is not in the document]
stacimc: Sure!

catalog/dags/data_refresh/distributed_reindex.py
stacimc: Ah, did this on autopilot and now you point it out I hate it! Strong agree, fixing immediately :)

sarayourfriend (approved):
LGTM! I've only re-reviewed this from a process perspective rather than testing it locally (didn't want to duplicate what Madison has done) and it makes good sense! Nice work, really excited to see this come together :grinning:

obulat:
I managed to run the audio refresh locally.
When running the image refresh, I ran into a couple of problems:
- `get_alter_batches` returns weird batches. I only have 15000 items in the temp table, but the batches returned are (5000, 105000)
- wait for reindex task is always yellow, and I don't know how to debug this. Here's what I see in the Airflow UI logs:
```
[2024-08-21, 16:23:10 UTC] {taskinstance.py:2905} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 460, in _execute_task
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 432, in _execute_callable
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 401, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/http/sensors/http.py", line 164, in execute
    super().execute(context=context)
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 401, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sensors/base.py", line 264, in execute
    raise e
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sensors/base.py", line 246, in execute
    poke_return = self.poke(context)
                  ^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/http/sensors/http.py", line 149, in poke
    return self.response_check(response, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/catalog/dags/data_refresh/distributed_reindex.py", line 79, in response_check_wait_for_completion
    raise ValueError("An error was encountered during reindexing.")
ValueError: An error was encountered during reindexing.
```

catalog/dags/data_refresh/copy_data.py
```diff
@@ -198,7 +198,7 @@ def copy_data(
 def copy_upstream_table(
     upstream_conn_id: str,
     downstream_conn_id: str,
-    environment: Environment,
+    target_environment: Environment,
```
obulat, line 201: `target_environment` is not used anywhere in this function, should it be removed?

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,403 @@
...
+    return True
+
+
+@task
+@setup_ec2_hook
+def get_worker_params(
+    estimated_record_count: int,
+    environment: str,
+    target_environment: Environment,
+    ec2_hook: EC2Hook = None,
```
obulat, line 91: `ec2_hook` is not used in this function

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -0,0 +1,403 @@
...
+    return reservations[0].get("Instances", {})[0].get("PrivateIpAddress")
+
+
+@task(
+    # Instance creation tasks are skipped locally, but we still want this task to run.
+    trigger_rule=TriggerRule.NONE_FAILED
+)
+def create_connection(
+    environment: str,
+    target_environment: str,
```
obulat, line 236: `environment` and `target_environment` are not used in this function

obulat (approved):
I think something is wrong with my setup: audio refresh ran fine, but running image refresh locally didn't work as expected. I don't want to block on it since you and Madison have successfully run the process locally. 
I went through the code in Pycharm and marked all places where it complained of unused function parameters. Are they going to be used in the future, or can they be removed?

stacimc:
@obulat is there any chance you already had a `temp_import_image` table in your API DB, perhaps from running the DAG previously (the DAG currently does not drop the import tables)? The errors you're seeing are what I would expect when the import table already exists and are a quirk of the DAG being half-completed.

catalog/dags/data_refresh/distributed_reindex.py
stacimc: Thanks for noting these!

stacimc:
I've retested this several times and cannot reproduce @obulat's errors; given that multiple other people are also unable to reproduce, and that I'm fairly sure the issue can be explained by the temp table already existing (which will be fixed as the remaining steps are added to the DAG), I'm going to go ahead and merge. This will be tested many more times before it goes live :)

PR #4567, opened 2024-06-27
Expand `ov` documentation and replace environment setup docs with `ov` instructions

sarayourfriend:
## Fixes

Related to #4343

## Description

`ov` has been adopted by the team, and we hope to make it _the_ Openverse development environment. In doing so, we would explicitly drop support for any alternative methods of working with Openverse. This would allow us to assume certain tools will exist within the `ov` context and can be reliably used without concern of adding further dependencies that developers need to individually manage.

I'm taking for granted that I have not 100% nailed the documentation here, and that we will iterate on this and improve it as we go, like all documentation.

This PR replaces all development environment setup documentation with instructions for using `ov`. It also documents parts of `ov` that were previously undocumented, and brings other bug fixes and improvements:
- Document `ov aliases` and improve the way they work to be more flexible and allow listing and documenting the available aliases
- Fix an issue where `--tty` was incorrectly used, causing carriage returns to exist in piped output
- Document the way `ov` works so that it's easier for maintainers to understand, contribute to and augment it themselves (and troubleshoot problems)
- Add `ov doctor` as a place to add common troubleshooting fixes that can be easily automated
- Replace relative references to `ov` with references that assume it's available on the `PATH`. While we will still not _require_ putting `ov` on the PATH (it may not be feasible for some), it is far too cumbersome in the documentation to use a relative path, especially if we want to be _accurate_ when we use it and clarify it must be done from the repository root. Instead, we can clarify this once in the general setup guide. Plus, `ov` is aware of your current working directory, and aside from just recipes, it's much easier to be able to `cd api && ov pdm add cool-new-dependecy` than `./ov pdm -p ./api install cool-new-dependency`.

As part of the improvements to aliases, I renamed `.ov_aliases.json` to `.ov_profile.json` to give it a more generic name. This opens up the possibility of using it for more advanced customisation, like defining a list of `dnf` packages to install in `setup_env.sh` (as an example). That would be nice for things like Playwright, where we _could_ run it directly in `ov`, but we shouldn't include it in the base system because installing its dependencies (chromium) significantly increases the download size (and time) of the `dnf install` in the Dockerfile.

## Testing Instructions

Read the documentation. Try out `ov aliases`, `ov help`, `ov aliases --help`, and piping `ov` output to `ov` or other tools (e.g., `ov just api/stats | ov jq '.[0]'` behaves _very_ strangely on `main` due to the carriage returns from the incorrect Docker tty usage, but is fixed on this branch).

## Checklist

[...]

Issue #1986, opened 2023-05-02
Find some way to apply prettier to admonition code-blocks


sarayourfriend:
## Problem

Prettier skips wrapping code blocks. This becomes a problem for Sphinx `admonition` directives. See this commit as an example:

https://github.com/WordPress/openverse/pull/1975/commits/4e46865f273f35181d0f8cc71bfbc5fd93e0cae1

## Description

Find some way to run Prettier _or some other auto-word-wrapping-tool_ against these code blocks. People have diverse editor configurations, including things like proportional coding fonts (which are an accessibility consideration), and manually wrapping anything, let alone several lines of prose which may be edited later and require re-wrapping, is not a long-term approach we should consider acceptable.

sarayourfriend:
It's not possible to use a prettier plugin without publishing it or without switching to a local hook for prettier like we use for eslint. That's not really an option, it seems, otherwise we'll be adding `pnpm` as a dev dependency for the entire monorepo, rather than just the frontend (which is currently the case).

PR #4585, opened 2024-07-01
Make running Playwright to update snapshots easier

### LINKED ISSUE #4535: Make running Playwright to update snapshots easier
## Problem

Playwright is very heavy to run. Any time we change versions of Playwright, to run the visual regression tests locally (and update snapshots accurately), you're forced to download a rather large Playwright-specific docker image (which incidentally also tends to come from a slow host, depending on where in the world you are). This makes it difficult for some contributors to run, especially if your development computer isn't a huge multi-core powerhouse.

## Description

Originally I thought about the alternative approach I've shared below (opening a new branch and PR), but that would require maintaining a few more new triggers and workflows to clean up branches and such. Instead, using `git diff --binary` might be a good approach for creating a shareable blob that could be applied by _any_ contributor, and not require any special permissions or use any more resources than we currently use already.

To do that, update the existing Playwright CI jobs to run Playwright with `-u` so that snapshots are enabled. Then, instead of only relying on a non-zero exit code, check for a non-zero exit code and whether snapshots changed. If either happened, consider the run a failure, and trigger the existing failure job.

However, instead of uploading the test results zip like we do now, we can instead upload a gzipped git patch blob:

```
git add .
git diff --staged --binary | gzip -f > /tmp/snapshot_diff.patch.gz
```

Upload `/tmp/snapshot_diff.patch.gz` as an artefact of the workflow run, and link to it directly from the comment. Add instructions for how to use it:

> ```gunzip -k -c snapshot_diff.patch.gz | git apply -```

`gunzip` looks to be available on macOS (and is usually available on most Linux distros), but we can skip compression if we don't think it's that important.

This would improve the situation for contributors and prevent needing to run the heavy visual regression tests locally if you're unable to for any reason. 

## Alternatives
Let's leverage CI more heavily for running Playwright. We need to make sure we use this responsibly (to prevent exhausting our free CI resources), but we could add a new GitHub Workflow accessible to core contributors (those with commit access to the monorepo) that runs the Playwright test suite with `-u` enabled (so snapshots get updated), and opens a PR targeting the PR that caused changes.

We could consolidate this into our existing playwright run, actually, by changing the playwright runs to always pass `-u` in CI, and then instead of relying on just a non-zero exit code from the test runs, check for a non-zero exit code _and_ see if snapshots updated. If either happened, then consider it a failed run. We can leave the comments we already leave, but if there are snapshot changes, then also open (or update) a secondary branch targeting the original PR's branch with a single commit containing the updated snapshots. The flow should go:

```
Run Playwright tests with `-u`

If snapshot changes:
    Create a local branch `${ PR branch }__playwright_ci_snapshots`
    Commit the snapshot changes `git add . && git commit -m 'Update snapshots'`
    Force push the branch
    If a PR does not exist yet for this new branch:
        Open the PR and save the link

If snapshot changes or non-zero exit code:
    Consider it a failure
    Leave the failed test comment
    If snapshot changes:
        include the link to the PR with the snapshot updates
```

An additional step needs to be added for when a PR is merged, to check for branches and PRs of the pattern `origin/${ PR branch }__playwright_ci_snapshots` and close those PRs & delete the branches on `origin`.

## Additional context

Coming out of thinking about https://github.com/WordPress/openverse/issues/4514 and @madewithkode's experience with the playwright tests recently.

madewithkode:
## Fixes

Fixes #4535 by @sarayourfriend 

## Description

This PR aims to make debugging Playwright test failures caused by snapshot changes easier by modifying the current Playwright CI workflow to run tests with the `-u` option which basically updates modified snapshots automatically. We then upload a gzipped git patch blob of these changes so the initiating user can download and apply the patch to their branch without having to run Playwright updates(which happen to be a resource intensive process) locally.

## Testing Instructions

This can be tested by triggering a failing Playwright workflow due to UI changes.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -803,11 +803,11 @@ jobs:
           - storybook
         include:
           - name: playwright_vr
-            script: "test:playwright visual-regression"
+            script: "test:playwright visual-regression -u"
           - name: playwright_e2e
             script: "test:playwright e2e"
           - name: storybook
-            script: "test:storybook"
+            script: "test:storybook -u"
 
     steps:
       - name: Checkout repository
@@ -822,12 +822,37 @@ jobs:
       - name: Run Playwright tests
         run: just frontend/run ${{ matrix.script }} --workers=2
 
+      - name: Check for changes after running tests with -u
+        id: check_snapshots_changed
+        if: failure() || success()
+        run: |
+          diff_names="$(git diff --binary --name-only)"
+          if [ -n "$diff_names" ]; then
+            git add . && git diff --staged --binary > /tmp/snapshot_diff.patch
+            echo "snapshots_changed=true" | tee "$GITHUB_OUTPUT"
+            printf "The following snapshots were updated:\n"
+            while IFS= read -r filename; do
+              printf "\t- %s\n" "$filename"
+            done <<< "$diff_names"
+            exit 1
+          else
+            echo "snapshots_changed=false" | tee "$GITHUB_OUTPUT"
+          fi
+
       - uses: actions/upload-artifact@v4
         if: failure()
         id: test-results
         with:
           name: ${{ matrix.name }}_test_results
           path: frontend/test-results
+
+      # This step only runs if there was a failure and snapshots indeed changed.
+      - uses: actions/upload-artifact@v4
+        if: failure() && steps.check_snapshots_changed.outputs.snapshots_changed == 'true'
+        id: snapshot-patch
+        with:
+          name: ${{ matrix.name }}_snapshot_diff
+          path: /tmp/snapshot_diff.patch
 
   # This job runs when `playwright` doesn't and always passes, thus allowing
   # PRs to meet the required checks criteria and be merged.
@@ -888,11 +913,9 @@ jobs:
           help_body<<$EOF
           **Playwright failure test results**: <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}>
 
-          It looks like some of the Playwright tests failed. You can download the trace output and image diffs for the failed tests under the "Artifacts" section in the above page.
+          It looks like some of the Playwright tests failed. If you made changes to the frontend UI without updating snapshots, this might be the cause. You can download zipped patches containing the updated snapshots alongside a general trace of the tests under the "Artifacts" section in the above page. They're named in the form \`*_snapshot_diff\` and \`*_test_results\` respectively.
 
-          You can read the [docs on how to use this artifact](https://docs.openverse.org/frontend/reference/testing_guidelines.html#debugging).
-
-          Additionally, if you made changes to the frontend UI, you may need to [update relevant playwright test snapshots](https://docs.openverse.org/frontend/guides/test.html#updating-snapshots).
+          You can read more [on how to use these artifacts in the docs](https://docs.openverse.org/frontend/reference/testing_guidelines.html#debugging).
 
           If the test is flaky, follow the [flaky test triage procedure](https://docs.openverse.org/general/test.html#flaky-tests).
           $EOF
--- a/documentation/frontend/reference/testing_guidelines.md
+++ b/documentation/frontend/reference/testing_guidelines.md
@@ -357,15 +357,44 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
 [Playwright Trace Viewer](https://playwright.dev/docs/trace-viewer) to inspect
 these (or open the zip yourself and poke around the files on your own).
 
-For Playwright tests failing in CI, a GitHub comment will appear with a link to
-download an artifact of the `test-results` folder.
+Playwright tests in CI are run with `-u` option by default, this means that
+snapshots will automatically be updated for modified parts of the UI if
+Playwright detects that. See
+[Updating snapshots](/frontend/guides/test.md#updating-snapshots) for more
+reading about this.
+
+When this happens, a GitHub comment will appear with a link to download zipped
+artifacts named in the form `*_snapshot_diff.zip`. Download and save this to the
+repository root. Once downloaded, decompress and apply them to your working
+branch by running:
+
+`ov unzip -p *_snapshot_diff.zip | git apply`
+
+The above command basically uses the `unzip` tool to unpack the contents of the
+downloaded archive file named `*_snapshot_diff.zip`, the `-p` option prevents it
+from actually creating any extracted file but rather prints the contents to the
+standard output which is then piped to `git apply`.
+
+```{note}
+Remember to replace `*_snapshot_diff.zip` with the actual downloaded filename.
+
+Keep in mind that you'd need to delete the downloaded file
+(`*_snapshot_diff.zip`) from the repository after successfully applying
+the changes to avoid committing them.
+
+```
+
+After successfully applying the patch, stage, commit and push the latest changes
+to your branch upstream and you should most likely have Playwright CI tests
+pass.
 
 Visual regression tests that fail to match the existing snapshots will also have
 `expected`, `actual` and `diff` files generated that are helpful for
 understanding why intermittent failures are happening. These are generated
 automatically by Playwright and will be placed in the `test-results` folder
 under the fully qualified name of the test that failed (with every parent
-describe block included).
+describe block included). This is also available for download in the Artifacts
+section after every failed Playwright test in the CI.
 
 Additionally, you can run the tests in debug mode. This will run the tests with
 a headed browser as opposed to a headless (invisible) one and allow you to watch
--- a/docker/dev_env/Dockerfile
+++ b/docker/dev_env/Dockerfile
@@ -50,6 +50,7 @@ RUN dnf -y install dnf-plugins-core \
     nodejs npm \
     python3.12 pipx \
     docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin \
+    unzip \
   && dnf clean all \
   && pipx install \
     httpie \


documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,10 +357,29 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+artifacts named in the form `*_snapshot_diffs`. The resulting artifacts would
+each contain a zipped patch of the snapshot changes named
+`snapshot_diff.patch.gz`, once downloaded, they can be unzipped and applied to
+your working branch by running:
+
+`gunzip -k -c snapshot_diff.patch.gz | git apply -`
+
+```{note}
+It is assumed that you'd already have `gunzip` installed as it looks to be available on macOS and is usually available on most Linux distros. If this is not your case, you might need to have that installed first.
+```
```
sarayourfriend, line 376: Good note. It is available in `ov`, so the instructions can be `cat snapshot_diff.patch.gz | ov gunzip -kc - | git apply`. gzip is usually available, you're right, but we can do away with the question altogether using `ov` :slightly_smiling_face:

documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,10 +357,29 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+Playwright tests in CI are run with `-u` option by default, this means that
+snapshots will automatically be updated for modified parts of the UI if
+Playwright detects that. See
+[Updating snapshots](/frontend/guides/test.md#updating-snapshots) for more
+reading about this.
+
+When this happens, a GitHub comment will appear with a link to download
+artifacts named in the form `*_snapshot_diffs`. The resulting artifacts would
+each contain a zipped patch of the snapshot changes named
+`snapshot_diff.patch.gz`, once downloaded, they can be unzipped and applied to
```
sarayourfriend, line 369: Very minor nitpick, just to be more precise about the compression used.

```suggestion
each contain a gzipped patch of the snapshot changes named
`snapshot_diff.patch.gz`, once downloaded, they can be decompressed and applied to
```

.github/workflows/ci_cd.yml
```diff
@@ -822,12 +822,24 @@ jobs:
...
+            echo "snapshots_changed=false" >> $GITHUB_OUTPUT
+          fi
+
       - uses: actions/upload-artifact@v4
-        if: failure()
+        if: failure() || steps.check_snapshots_changed.outputs.snapshots_changed == 'true'
         id: test-results
         with:
-          name: ${{ matrix.name }}_test_results
-          path: frontend/test-results
```
sarayourfriend, line 830: We should still upload the test-results file. The trace can still be useful for non-snapshot tests that fail, especially to debug if they fail in CI but pass locally. Would it still get uploaded by some other means I'm not understanding?

.github/workflows/ci_cd.yml
```diff
@@ -822,12 +822,24 @@ jobs:
...
+        id: check_snapshots_changed
+        run: |
+          # shellcheck disable=SC2086
+          if [ -n "$(git diff --binary --name-only)" ]; then
+            git add . && git diff --staged --binary | gzip -f > /tmp/snapshot_diff.patch.gz
+            echo "snapshots_changed=true" >> $GITHUB_OUTPUT
+            git diff --binary --name-only
+          else
+            echo "snapshots_changed=false" >> $GITHUB_OUTPUT
+          fi
```
sarayourfriend, line 835: Ideally we wouldn't disable shell check rules unless absolutely necessary. Is there a reason why `$GITHUB_OUTPUT` cannot be double quoted? If so, can you add a comment explaining why we need to disable the check for this script?

.github/workflows/ci_cd.yml
```diff
@@ -822,12 +822,24 @@ jobs:
...
+
       - uses: actions/upload-artifact@v4
-        if: failure()
+        if: failure() || steps.check_snapshots_changed.outputs.snapshots_changed == 'true'
         id: test-results
         with:
-          name: ${{ matrix.name }}_test_results
-          path: frontend/test-results
+          name: ${{ matrix.name }}_snapshot_diffs
+          path: /tmp/
```
sarayourfriend, line 842: Small nit/question: there's only a single diff file, so we don't need to pluralise the artefact name. But do we need to use the complete path to it, by chance?

```suggestion
          name: ${{ matrix.name }}_snapshot_diff
          path: /tmp/snapshot_diff.patch.gz
```

sarayourfriend (changes requested):
Looks good to me, just two small things to change:

1. Remove the need to disable SC2086 or add a comment explaining why it needs to be disabled.
2. Retain the previous behaviour of uploading the test-results trace file, as it serves a purpose beyond snapshot test debugging.

The rest are clarifications/questions/nit picks. Looks good otherwise though :slightly_smiling_face:

.github/workflows/ci_cd.yml
madewithkode: I agree. So what I'm picking from this is that we should upload both the trace like we already do, alongside the patches(if they exist)?

.github/workflows/ci_cd.yml
madewithkode: Good spot with the complete path. It should upload just the file and not the whole folder since we already have a folder(artifact) declared for it.

.github/workflows/ci_cd.yml
madewithkode: No reason at all. It didn't occur to me that that was the exact reason as I'd thought it had something to do with the compression command `gzip -f > /tmp/snapshot_diff.patch.gz`. Fixed that.

.github/workflows/ci_cd.yml
madewithkode: Suggested changes have been implemented

.github/workflows/ci_cd.yml
sarayourfriend: Yep :sligh

.github/workflows/ci_cd.yml
```diff
@@ -822,13 +822,32 @@ jobs:
...
       - uses: actions/upload-artifact@v4
         if: failure()
         id: test-results
         with:
           name: ${{ matrix.name }}_test_results
           path: frontend/test-results
 
+      # This step only runs if there was a failure and snapshots indeed changed.
+      - uses: actions/upload-artifact@v4
+        if: failure() && steps.check_snapshots_changed.outputs.snapshots_changed == 'true'
```
sarayourfriend, line 845: When passing `-u`, the exit code will be 0 (if there are no regular test failures), so in the case of when there are _only_ updated snapshots, checking `failure()` would only upload the snapshots if there were additional failing tests.

```suggestion
        if: steps.check_snapshots_changed.outputs.snapshots_changed == 'true'
```

.github/workflows/ci_cd.yml
```diff
@@ -822,13 +822,32 @@ jobs:
       - name: Run Playwright tests
         run: just frontend/run ${{ matrix.script }} --workers=2
 
+      - name: Check for changes after running tests with -u
+        id: check_snapshots_changed
```
sarayourfriend, line 826: It occurred to me when thinking about the condition on the step to upload the snapshot diff artefact, that we need _this_ step to run even if there testing step fails. By default, it wouldn't, so if there are failures in addition to changed snapshots, we'd miss the changed snapshots.

```suggestion
        id: check_snapshots_changed
        if: success() || failure()
```

sarayourfriend (approved):
Just two small things on conditions for the steps I am pushing changes for, but otherwise this looks good to me. If we find we need to adjust, this is a perfect base to iterate on (but for now I think this should work just fine and will be a good improvement).

documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,15 +357,34 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+[Updating snapshots](/frontend/guides/test.md#updating-snapshots) for more
+reading about this.
+
+When this happens, a GitHub comment will appear with a link to download
+artifacts named in the form `*_snapshot_diff`. The resulting artifacts would
+each contain a gzipped patch of the snapshot changes named
+`snapshot_diff.patch.gz`, once downloaded, they can be decompressed and applied
+to your working branch by running:
+
+`cat snapshot_diff.patch.gz | ov gunzip -kc - | git apply`
```
krysal, line 372: I [downloaded](https://github.com/WordPress/openverse/actions/runs/9798106098/job/27055999711?pr=4600) it as `playwright_vr_snapshot_diff.zip` so I had to decompress it twice (first from a .zip and then from a .gz). It seems that the GitHub actions always do the compression for downloads, so we can skip that part.

```suggestion
`cat playwright_vr_snapshot_diff.zip | ov unzip -kc - | git apply`
```

.github/workflows/ci_cd.yml
```diff
@@ -822,13 +822,33 @@ jobs:
...
+        if: success() || failure()
+        run: |
+          if [ -n "$(git diff --binary --name-only)" ]; then
+            git add . && git diff --staged --binary | gzip -f > /tmp/snapshot_diff.patch.gz
```
krysal, line 830: Let's skip this compression.
```suggestion
            git add . && git diff --staged --binary > /tmp/snapshot_diff.patch.gz
```

.github/workflows/ci_cd.yml
```diff
@@ -822,13 +822,33 @@ jobs:
...
+          fi
+
       - uses: actions/upload-artifact@v4
         if: failure()
         id: test-results
         with:
           name: ${{ matrix.name }}_test_results
           path: frontend/test-results
 
+      # This step only runs if there was a failure and snapshots indeed changed.
```
krysal, line 844: Not exactly accurate and mostly redundant. I'd omit the comment altogether.
```suggestion
      # This step only runs if snapshots are changed.
```

krysal (commented):
This is looking very cool! Generating snapshots is quite a consuming task. I tried the flow in https://github.com/WordPress/openverse/pull/4600 and (with some little changes) it was a breeze to get the updated snapshots.

My only concern is that I did not get failing tests from the changes. They're supposed to fail so you are aware that you need to update them. Are we on the same page here that this is still the expected behavior?

documentation/frontend/reference/testing_guidelines.md
sarayourfriend: Nice!

sarayourfriend:
You're right, @krysal, it should still consider it a failure in the check. I guess we can fix that by adding `exit 1` to the script that creates the patch diff when there are updated snapshots. Maybe it should also log the list of updated snapshots in that step too.

documentation/frontend/reference/testing_guidelines.md
madewithkode: Thanks for pointing this out @krysal. A slight change to your suggestion would be:
 `cat *_snapshot_diff.zip | ov unzip -kc - | git apply`
 This is because the prefix `*` represents the current playwirght test. It is dynamic and auto generated in the workflow using a matrix. I'd go ahead and note this on the doc too.

madewithkode:
@sarayourfriend I'd think that `git diff --binary --name-only` in the patch creation script itself already logs the list of updated snapshots. If this is not the case, perhaps something like `git diff --binary --name-only | echo` should suffice? Also, I agree with exiting with 1 to trigger a failure.

This would be my modification of the patch creation script(let me know if it makes sense):
         
          if [ -n "$(git diff --binary --name-only)" ]; then

            git add . && git diff --staged --binary > /tmp/snapshot_diff.patch

            echo "snapshots_changed=true" | tee "$GITHUB_OUTPUT"

            echo "The following snapshots were updated:"

            git diff --binary --name-only | echo

            exit 1

          else

            echo "snapshots_changed=false" | tee "$GITHUB_OUTPUT"

          fi

sarayourfriend:
Because `git diff --binary --name-only` is captured in the test expression, it's `stdout` doesn't go to the actual logs. Or at least, that's my understanding of how output capture in scripts work, and it's how it behaves when I try it locally.

You should force a playwright failure in this PR so we can see how it goes in practice :slightly_smiling_face:

madewithkode:
@sarayourfriend @krysal  I have just triggered a playwright failure with the latest workflow changes on [this build](https://github.com/WordPress/openverse/actions/runs/9871509431) on this PR. Everything on the workflow works as expected. An update to snapshots triggered a failure and both test results and the snapshot patch/diff were uploaded.

However, I have noted a few things that might require slight changes with the realization that we no longer need to `gzip`/`gunzip` as Github CI zips artifacts by default.

After trying to make this work for myself locally while trying as much as I can to assume the role of a new contributor, the following command set seems more intuitive and works out of the box.

`unzip -q *_snapshot_diff.zip && git apply snapshot_diff.patch`

- `unzip -q *_snapshot_diff.zip` basically uses the unzip tool to unpack the contents of *_snapshot_diff.zip.
The `-q` option tells unzip to operate quietly, suppressing informational messages during the extraction process. 

Note: The `ov` toolkit does not seem to have the `unzip` utility installed yet. FWIW, `unzip` is widely available on most Unix based distros. However if we decide to add this to the `ov` toolkit and use `ov` instead the command can easily be modified to `ov unzip -q *_snapshot_diff.zip && git apply snapshot_diff.patch`

Running the above allowed me to seamlessly apply the downloaded patch without any errors. If this is okayed across board i'd proceed to update the PR with the changes.

sarayourfriend:
Awesome, glad the workflow is working!

You can definitely add unzip to `ov`, that'd be great! Piping the unzipped output would be better, I think, because unzipping two archives that have the same name of file inside the archive will overwrite each other, won't they? If both the storybook and playwright snapshots have diffs, wouldn't they output to the same `snapshot_diffs.patch` file, and overwrite whichever one the glob chose as second?

To add unzip to `ov`, add it to the dnf-installed packages in `docker/dev_env/Dockerfile`. Fedora calls it `unzip`:

```
unzip.x86_64 : A utility for unpacking zip files
```

After that, this works well for me locally and is nice and succinct. We could probably add a just recipe (or `ov` alias) for the unzip part too, to make it less fiddly, but not required for this PR if it's a hassle or behaves strangely for whatever reason.

```
ov unzip -p *_snapshot_diff.zip | git apply
```

madewithkode:
@sarayourfriend you're indeed correct about the patch overriding possibilities as the actual patches for both tests would aways have same name, thanks for spotting that. This, alongside other suggestions have been addressed and I can confirm that the latest version works as it should locally.

madewithkode:
@sarayourfriend you're indeed correct about the patch overriding possibilities as the actual patches for both tests would aways have same name, thanks for spotting that. This, alongside other suggestions have been addressed and I can confirm that the latest version works as it should locally and can also be confirmed from [this test run.](https://github.com/WordPress/openverse/actions/runs/9895149129)

.github/workflows/ci_cd.yml
sarayourfriend: @madewithkode it looks like this change got lost in the mix, but I think we still need it in the case I mentioned where there are multiple types of failures.

documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,15 +357,43 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+
+`ov unzip -p *_snapshot_diff.zip | git apply`
+
+The above command basically uses the `unzip` tool to unpack the contents of the
+downloaded archive file named `*_snapshot_diff.zip`, the `-p` option prevents it
+from actually creating any extracted file but rather prints the contents to the
+standard output which is then piped to `git apply`.
+
+```{note}
+For the above command to work verbatim for you, make sure to have saved the downloaded artifact to the root of the project from where you're able to run git commands. Otherwise, you'd need to adjust the path to the downloaded file in the command accordingly. Should you settle for the former, keep in mind that you'd also need to delete the file(`*_snapshot_diff.zip`) after successfully applying the changes to avoid committing them.
```
sarayourfriend, line 378: Unfortunately #1986 means we need to manually wrap the text inside code blocks, to at least _around_ 80 characters. It's so tedious I really don't try too hard other than to get it around 80. Still needs to be done, unfortunately.

There's also a space missing between `file` and the parenthetical after it.

documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,15 +357,43 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+When this happens, a GitHub comment will appear with a link to download zipped
+artifacts named in the form `*_snapshot_diff.zip`. Once downloaded, they can be
+decompressed and applied to your working branch by running:
+
+`ov unzip -p *_snapshot_diff.zip | git apply`
+
+The above command basically uses the `unzip` tool to unpack the contents of the
+downloaded archive file named `*_snapshot_diff.zip`, the `-p` option prevents it
+from actually creating any extracted file but rather prints the contents to the
+standard output which is then piped to `git apply`.
```
sarayourfriend, line 375: Just a nit, and something we need to address more widely in our documentation (it's unfortunately pervasive, including in my own writing), but this would read a lot more clearly in active voice, rather than passive, and even more so as imperative instructions. "Once downloaded, apply them to your working branch by running", etc.

documentation/frontend/reference/testing_guidelines.md
sarayourfriend: Also: for the files to be in context for `ov`, they _must_ be downloaded to the repository. The repository is the only host filesystem location that ov explicitly mounts into the container. Probably worth just adding that to the instructions: download the zipped snapshot patches to the repository root, etc.

documentation/frontend/reference/testing_guidelines.md
```diff
@@ -357,15 +357,43 @@ output of failed tests by finding the `trace.zip` for the relevant test in the
...
+
+```{note}
+For the above command to work verbatim for you, make sure to have saved the downloaded artifact to the root of the project from where you're able to run git commands. Otherwise, you'd need to adjust the path to the downloaded file in the command accordingly. Should you settle for the former, keep in mind that you'd also need to delete the file(`*_snapshot_diff.zip`) after successfully applying the changes to avoid committing them.
+
+Remember to replace `*_snapshot_diff.zip` with the actual downloaded filename.
+
+Additionally, you might need to replace `ov` with a relative path `./ov` if you do not have `ov` added to your PATH yet.
```
sarayourfriend, line 382: You can leave this out, in favour of the general advice we've written in #4567. We've decided not to have to write this caveat everywhere and just have it in a single place.

sarayourfriend (changes requested):
LGTM. I have a few bits of feedback on the docs, but none are blocking. The only blocking part is the condition of the snapshot check job, which needs to run if the previous step fails. Otherwise this LGTM!

madewithkode:
> LGTM. I have a few bits of feedback on the docs, but none are blocking. The only blocking part is the condition of the snapshot check job, which needs to run if the previous step fails. Otherwise this LGTM!

@sarayourfriend If I get you correctly, you basically want a `failure()` check added to `check_snapshots_changed` step? If this is correct, this would mean that it'd only run when playwright test step above exits with 1, however, I don't think snapshot changes/updates are typically considered a failure by playwright, so this might mean that the snapshot check step would be skipped if there were no actual test failures but only snapshot changes, which I assume isn't the behavior we want. I stand to be corrected.

sarayourfriend:
It should be `failure() || success()`. If the previous step fails on non-snapshot tests, but also includes changes to the snapshots, then we still need the next step to run and generate the diff. It needs to run:

- When the test run step "succeeds", so we can check for changed snapshots, generate the diff, and fail if necessary
- When the test run step fails due to non-snapshot related errors, so that we can also check for changed snapshots, generate the diff, etc.

It needs to run in both cases. Right now it will only run in the first one, so that if there are both types of test failures, the CI run would only show non-snapshot related failures.

sarayourfriend (approved):
LGTM!

sarayourfriend:
Docs build failure was in the storybook build part of the job, not the docs changed in this PR. I've just kicked off a re-run of the job. Looks like we got rate limited in the request to GlotPress.

@dhruvkb Maybe we should consider caching translation results for 24 hours and only pulling new ones once a day instead of on every single CI/CD run. I'll write an issue for it.

obulat (approved):
Thank you for all the work to get this working, @madewithkode ! This will make the process of updating tests much simpler

PR #4597, opened 2024-07-03
Add internet archive to robots.txt denylist

zackkrida:
This PR blocks Internet Archive from crawling search and single result endpoints. Wayback Machine is an awesome tool for preserving the history of the web, but in the case of Openverse we only want historical snapshots of our home and static content pages (/about, /sources, etc.). We do not want to archive single and search results because:

- The latest, most accurate information about a work should always be on the live openverse.org
- If a work is _removed_ from Openverse, it is likely for copyright, sensitivity, or other, potentially legal, reasons. This isn't _always_ the case though. Sometimes, sadly, sources go down or stop working.

### DIFF
--- a/frontend/src/server-middleware/robots.js
+++ b/frontend/src/server-middleware/robots.js
@@ -13,6 +13,7 @@ const deniedUserAgents = [
   "Bytespider",
   "ImagesiftBot",
   "cohere-ai",
+  "ia_archiver", // Internet Archive for the Wayback Machine. Not malicious, but we don't want to allow single and search results to be indexed at the risk of preserving deindexed works.
 ]
 
 const aiDisallowRules = deniedUserAgents


frontend/src/server-middleware/robots.js
```diff
@@ -13,6 +13,7 @@ const deniedUserAgents = [
   "Bytespider",
   "ImagesiftBot",
   "cohere-ai",
+  "ia_archiver", // Internet Archive for the Wayback Machine. Not malicious, but we don't want to allow single and search results to be indexed at the risk of preserving deindexed works.
```
sarayourfriend, line 16: I'm not sure I completely understand how this file works, but won't this disallow IA entirely? Is the goal to only ask them to not archive search and single results pages or to stop archiving everything from Openverse?

Probably also worth noting that you've written that IA isn't malicious, implying the other entries in this list are. The list is used to construct the variable `aiDisallowRules`, so it seems that the list is meant to be AI specific? The variable names are a little odd. I don't know whether labelling AI crawlers as _malicious_ is correct, would generally "unwanted" be more accurate?

Furthermore, and perhaps most directly related to the method if implementation and what the intention is, we already disallow on search and single result pages (down where `contents` is constructed). Is something about that not working, such that IA is still crawling the search pages? Does IA's crawler ignore robots.txt directives telling robots to ignore certain pages? If it does... then why would any change here _at all_ make a difference? All add this entry to this list does (as far as I can tell) is disallow all routes, rather than only the ones we don't want archived.

sarayourfriend (changes requested):
Requesting changes because I believe the implementation diverges from the intention expressed in the PR description. I may be misunderstanding this file, but I don't see why this would solve the specific problem in the PR description if the existing route-specific disallow entries aren't already working.

zackkrida:
I threw this up pretty hastily forgetting how our existing rules work. Even more concerning is that I think IA isn't respecting our robots.txt.

Going to close this, reexamine their traffic, and look into a block at the firewall level if proven necessary.

Issue #1290, opened 2023-02-12
Librivox


sgdwn:
### Source Site

https://librivox.org/

### Value Provided

huge repository of public domain audiobooks

### Licenses Provided

Public Domain // CC0

### Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
Did a bit more digging, we can sort the API by ID and iterate through that way by increasing the offset parameter:

https://librivox.org/api/feed/audiobooks?sort_field=id&offset=2

The API responds quickly even with high offsets (https://librivox.org/api/feed/audiobooks?sort_field=id&sort_order=desc&offset=18000) so I'm not worried that we would cause a problem if we slowly crawled the API in this way. It also returns JSON responses if you set the content type header to `application/json`.

[There is code to check a rate limit but based on that code](https://github.com/LibriVox/librivox-catalog/blob/master/application/libraries/REST_Controller.php#L715), it doesn't seem like it applies to the API as there's no `limit` defined on the feed controller. In any case, we should still get an API key for this if we do it.

Note that the API returns results for works that haven't yet been reviewed by the LibriVox folks, so we would need to check that `url_librivox` exists for the work before including it.

Additionally, there are further complications regarding how we would present these results. Specifically, what will be considered the "unit" of a single "work" that we would show in results? One audiobook recording can include many readers for different sections (see https://librivox.org/der-abenteuerliche-simplicissimus-teutsch-teil-1-by-hans-jakob-christoffel-von-grimmelshausen/ as an example). That means we can't include individual books as the result without flattening the attribution.  To make matters worse, however, the API doesn't have any way (as far as I can tell) to retrieve the readers for sections of a work. You can get the list of sections for an audio book but it does not list the readers (https://librivox.org/api/feed/audiotracks?project_id=2999).

I don't know whether attribution of the recording would need the reader, however, or if it should be attributed to the author of the work being read.

Tenacious-E:
Hello,

I would like to implement this feature.

sarayourfriend:
@Tenacious-E @AetherUnbound I don't think this feature is ready to implement yet. We need to discuss the issues I brought up in this comment, and decide on the approach we will take:

> I don't know whether attribution of the recording would need the reader, however, or if it should be attributed to the author of the work being read.

[...]

PR #4599, opened 2024-07-04
WIP: Unify names of generic relations to media classes

### LINKED ISSUE #4553: Replace function-local dict media type -> model relationship configurations with static module dicts or an otherwise shared approach
## Current Situation

See this thread for context: https://github.com/WordPress/openverse/pull/4544#discussion_r1653696199

This pattern exists widely in our code, and does not need to, and has the following current and potential problems:

- The pattern encourages a local definition of static, global relationships between entities (in the case of the example, the media type and a decision model for that media type). These relationships are _not_ local relationships, and should be defined once in a single place.
- Creating a dict only to immediately retrieve a key from it, and then throw the dict away, is potentially costly in the context of a tight loop. That isn't a problem in this particular example, but the pattern is very easily accidentally used in this way, and should under no circumstances be used indiscriminately, especially in code paths that will inevitably inform other code.
- The pattern is relatively obscure compared to an if/else or match/case or statically defined dict. While it is compact, compactness is not by itself a virtue in code, and certainly not in Python which actively resists compactness.

## Suggested Improvement

Either:
- Define the relationship between these objects in a single dict once in a shared location to be referenced by all places that care about this relationship; or
- Ensure the relationships can be derived directly from the models themselves, either via class-property introspection or by renaming the fields (or aliasing them) so that they can easily be retrieved generically from any media object regardless if the media type of that object

To me the latter is preferred because it creates a much cleaner and more consistent code, without the need to manually maintain dicts of these relationships at all.

## Benefit

Avoid the anti-pattern qualities of the function-local inline dict approach.

sarayourfriend:
## Fixes

Fixes #4553 by @sarayourfriend 

## Description

This is still WIP because I need to test a bit more thoroughly. So far things appear to work with manual testing, but I haven't run unit tests, and I suspect there are edge cases in the model admins that I have missed.

Heads up through, @dhruvkb, if you want to take a look at the approach I've taken. It's a bit abstract, and I could definitely see wanting to use a more concrete, straightforward, declarative approach that just manually creates the back references onto the media classes.

The approach overrides the metaclass for Django models with one specifically for models related to a media type. The abstract models reference that metaclass, along with a related name, and then subclasses are processed to get assigned to the media class as `{related_name}_class`, and the related name from the abstract class is passed down to the `Meta` of the final models (which Django doesn't do on its own, only a small set of fields are inherited from `Meta` by default).

I initially tried this with `__init_subclasses__` but you can kind of see what that results in with the admin models and forms... but a bit more unwieldy because of the related name. It ended up being a lot of specific, fiddly implementation details on each of the related abstract classes, which the metaclass was able to tidy up into a single implementation pretty nicely. Of course, it is a bit magical, so if we want to just, say, hand-maintain a nested dictionary of these relations, kind of like the `media_type_config` fixture in the tests (which could be replaced or reduced by some of this work), and prefer that for its straightforwardness (if being slightly more clunky and requiring more manual configuration of relation names) then I can probably be convinced.

My goal was to prevent any SQL migrations, and indeed, sqlmigrate shows the current migration as a complete noop.

Anyway, let me know what you think, Dhruv, if you get a chance to look at this while it's still a draft. Very curious to see what you'll think of this introspective/meta-programming based approach, and whether you can think of an easier way to do it or think a more concrete and less meta approach would be better.

## Testing Instructions

TBD.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (commented):
This approach is quite novel and interesting, and the final code undoubtedly looks much cleaner.

However, I feel like metaprogramming is a bit of a huge sledgehammer to wield against what was a relatively small nut of being able to connect media models with each other.

Using common related names like `sensitive_media` and `report` (instead of media-specific ones like `sensitive_image` and `audio_report`) already simplifies the code massively by removing almost all cases where we needed to handle media types separately.

What I'm trying to say is that I don't feel like the metaprogramming part adds much to this PR and the simplifications would remain even if that particular aspect of this PR was dropped.

I am making these statements based on a quick pass through the PR changes and so it is possible that I am not realising the full importance of the metaprogramming aspect. So I will definitely look into the PR more closely soon. But I wanted to leave my first impressions so that you can also have time to explore other directions if you are in an exploratory mood in the coming days.

sarayourfriend:
Yeah, totally agree metaprogramming is a pretty intense approach for this. I suppose the motivation of using metaprogramming to solve this, as opposed to explicitly configuring the related names on each real model (and additional explicit `__init_subclass__` or other type methods) was to make the wiring of these happen automatically, so new media related models (including other subclasses for new future media types) behave identically without needing to repeat the wiring everywhere.

The alternatives I considered were:
- as mentioned, `__init_subclass__` on each of the related media models. It is still metaprogramming, just maybe metaprogramming "lite", and required manually assigning related names on fields, because `__init_subclass__` happens after the Django metaclass initialisation, and so after `_meta` is constructed, and critically after `contribute_to_class` is called on each of the fields... meaning `__init_subclass__` cannot be the source of `default_related_name` on the subclasses' `Meta` (it doesn't happen early enough in the lifecycle). So ultimately this required a lot of manual configuration still, for every abstract related model and for each of the subclasses (which needed `Meta.default_related_name` defined individually _or_ to manually inherit `Meta` from the parent class).
- The same as above, but instead of `__init_subclass__` on each related model, create a caching `@property \ @classmethod` on `AbstractMedia` for each of the related media models, which would just call `<Related media model>.__subclassess__()` and find the one for the media type of the abstract media subclass (and then cache it so the runtime cost of accessing the classes this way is amortised). As with `__init_subclass__`, this still requires manually configuring the `related_name`s on the related models fields or `Meta` class.

What I like about the metaclass approach is that it fully centralises the configuration of the related name and the backreferences of the actual classes from the media models. so that subclasses of the related models don't have to do anything (other than define `media_class`, which most already did anyway). It removes the need to manually configure anything beyond the abstract classes, except for edge cases like the `moderator` reference, where the default related name clashed.

So, the metaclass approach looks pretty intense at face value, but it does, I think, ultimately simplify the configuration and make it less fragile or prone to random accidental variations (for example, when new related models are added).

--- 

_(Here's an aside about our data model, inspired by thinking about the different approaches and where this complexit comes from)._

To be honest though, I think it's worth us considering whether the per-media type segmentation of the models (and tables) is entirely useful, and whether there is a generic approach that would actually eliminate the need to repeat structures for each media type, especially when considering aspects of the data model like the complexities I touched on in #1290, where our current model is so focused on individual works, it doesn't allow for the actually complex relationships that sometimes exist. Audio having Audioset as a distinct audio-specific implementation is interesting, for example. We already have images in our catalogue that are parts of real, meaningful sets (which could, for example, be extremely useful for related media queries). Mostly these are scans of pages from the same books (Biodiversity Heritage Library being the source that immediately comes to mind), but there are other forms of images-in-series or collections... not to mention the idea of being able to create Openverse-user curated collections, which we've touched on frequently.

So where metaprogramming in this PR seems like it could be overly complex or abstract, it might also (or instead) point to a problem with our data model, which pre-encodes this level of abstraction and complexity that might neither be necessary or even _real_ in a fundamental sense, and for which meaningful differences between media types could be accounted in less complex ways.

Notably, other systems for modelling the metadata of individual works all work under the assumption of linked data built off core metadata shared by most works regardless of the kind. For example DublinCore and the more complex RDF (which both came up in #4430) work from the same set of basic terms shared by all works, and enabling much more flexible modeling and metadata storage than we currently have, demonstrated by [Fedora Repository[^notlinux]](https://wiki.lyrasis.org/display/FF/Fedora+Repository+Home) as an example. Our per-media-type approach is not necessarily the best way of doing things generally, and other systems that pre-date ours and which have been manually tested and applied to many millions of digital and analogue works do things a little differently.

Crucially broken in our system, for example, is how a multimedia work or collection of works would be described. We don't have any way of modelling audio that are related to images, for example, that wouldn't require creating a separate container for those works from containers of audio (which we have as audio set) and containers of images (which we do not have).

Anyway, that's not really important to the question of what the best way of solving this problem of generic relations between these Python classes... I just thought it was interesting to consider why these relations exist, whether they're necessary and whether they're even _real_ (in the sense of, does this division actually exist when works are reduced down to an identifier pointing to a binary blob, with a few to many of potentially infinite number of data points associated). If we step back from the model we've implemented and especially if we look at how others have tried to solve this problem, our approach is rather distinct. Whether that's good, purposeful, or serves our use case in a meaningful way that is worth the trade-offs... good for us to think about!

[^notlinux]: Unrelated to the Linux distro.

api/api/models/audio.py
```diff
@@ -277,7 +276,7 @@ class DeletedAudio(AbstractDeletedMedia):
         primary_key=True,
         db_constraint=False,
         db_column="identifier",
-        related_name="deleted_audio",
+        related_name="deleted_media",
```
sarayourfriend, line 279: Oops, these actually do not need to be defined anymore! I accidentally left these in when I was trying one of the other approaches I mentioned that still required manually configuring the related names. The metaclass approach intentionally removes the need for this.

api/api/models/base.py
```diff
@@ -12,3 +16,74 @@ def __iter__(self):
         for field_name in self._meta.get_fields():
             value = getattr(self, field_name, None)
             yield field_name, value
+
+
+class OpenverseMediaModel(Protocol):
+    media_type: MediaType
```
sarayourfriend, line 22: This is unused, left over from another experimental approach I tried (one that didn't leverage metaclass kwargs and instead generated a new metaclass per-abstract related media model, or used a decorator, both of which were a little ugly or didn't work).

Anyway, it's meant to be removed.

api/api/models/media.py
```diff
@@ -147,6 +147,18 @@ def attribution(self) -> str | None:
         except ValueError:
             return None
 
+    @property
+    def index(self) -> str:
+        return settings.MEDIA_INDEX_MAPPING[self.media_type]
+
+    @property
+    def filtered_index(self) -> str:
+        return f"{self.index}-filtered"
```
sarayourfriend, line 156: These should be cached classmethod properties.

api/api/models/media.py
```diff
@@ -275,17 +289,29 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
-class AbstractMediaDecision(OpenLedgerModel):
+# Do not subclass or it will cause an unnecessary migration due to the change in class name
+def MediaModeratorForeignKey(related_name: str):
```
sarayourfriend, line 293: This should probably be a static method on the `AbstractMediaDecision` instead.

api/api/models/audio.py
dhruvkb: This is the approach that I was actually preferring over the metaclass because it is much more explicit and clearer.

sarayourfriend: It doesn't solve the back-references to the class though. You can go through the related _set_ but if it's a one-to-one, there's no good way to do it, as far as I can tell, that won't cause an implicit query.

The complexity comes from trying to solve both problems:
- Make it possible to use generic references to instance relations
- Make it possible to use generic references to _model_s of those relations

api/api/models/base.py
krysal: Could you leave a comment in the code saying it is meant to be removed?

krysal:
This is looking quite interesting. Similar to Dhruv's opinion, I like how much code this saves, and the resulting one looks clean. I also worry about the metaprogramming part, it diverges quite a lot from our current approach. If we can reach an intermediate stage that would be wonderful. Perhaps with your second alternative:

> The same as above, but instead of __init_subclass__ on each related model, create a caching @property \ @classmethod on AbstractMedia for each of the related media models, which would just call <Related media model>.__subclassess__() and find the one for the media type of the abstract media subclass (and then cache it so the runtime cost of accessing the classes this way is amortised). As with __init_subclass__, this still requires manually configuring the related_names on the related models fields or Meta class.

---

> So where metaprogramming in this PR seems like it could be overly complex or abstract, it might also (or instead) point to a problem with our data model, which pre-encodes this level of abstraction and complexity that might neither be necessary or even real in a fundamental sense, and for which meaningful differences between media types could be accounted in less complex ways.

Regarding data modeling, I find our current approach still simple, as we currently have a few central tables. Actually, most of the complexities I usually find come more from maintaining a sort of mirror with the catalog DB. Of course, I'd be interested in exploring new ways to store/represent the data, but I don't see this as a problem for our current use case but with _expanding_ the use cases that we could potentially include.

> Crucially broken in our system, for example, is how a multimedia work or collection of works would be described. We don't have any way of modelling audio that are related to images, for example, that wouldn't require creating a separate container for those works from containers of audio (which we have as audio set) and containers of images (which we do not have).

The key is that the current models weren't exactly planned for collections, so it's understandable that they cannot be fit currently. It's important to note this because relational models are designed with very specific requirements so they can work better. It's not set in stone, of course, but trying to change it one way or another has several implications.

>  If we step back from the model we've implemented and especially if we look at how others have tried to solve this problem, our approach is rather distinct.

I'd like to know more about these cases you mentioned so I can understand them better! And see how they relate to what we are trying to resolve. I agree we can reevaluate all of these, but we also need to define the requirements/constraints beforehand.

api/api/models/base.py
sarayourfriend: It needs to be removed from this PR. I'll do so when I get back around to making updates to this based on feedback for the overall approach.

PR #4603, opened 2024-07-05
Limit access for content moderators

### LINKED ISSUE #3640: Finalise access control for Django admin content moderator users
## Description

See the following for implementation description and rationale: https://docs.openverse.org/projects/proposals/trust_and_safety/content_report_moderation/20231208-implementation_plan_django_admin_moderator_access.html#finalise-access-control

**This issue only covers the final point in the linked section.** Necessary infrastructure changes are covered in https://github.com/WordPress/openverse-infrastructure/issues/748

dhruvkb:
## Fixes

Fixes #3640 by @sarayourfriend 

## Description

This PR
- changes all moderator access to just "view".

  This means they can only view the model instances from the admin view. Actions like edit, delete and make changes are not available to them. They can still file reports and take decisions on pending reports using the form within the single media item view in the admin as those are implemented through custom views.

- adds checks on the custom views for taking decision

  The decision models, and the corresponding through models, are created through the custom decision making view. This PR also adds a check to this view that only allows `POST` requests from people with `is_superuser` flag or "Content Moderators"-group membership.

## Testing Instructions

0. Check out the PR.
1. Re-init the API DB so that the old moderators group is replaced with new permissions.

   ```bash
   just down -v && just api/init
   ```

2. Login into the Django admin as the user "moderator" (password: "deploy"). You should not see any controls to edit or delete any models.
3. Go to the single media item view and try to file report and perform moderation. You should still be able to do both of these things.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
@dhruvkb I don't really understand the decision to not use Django's permissions model. I can see that it does work, but it's a surprising deviation from Django's intended usage. Can you elaborate further what the trade-offs were that led to the decision not to use the permissions model as intended and described by the implementation plan? Without the specific context of this code and PR, if I looked at the permissions of the content moderator group in production and only saw view permissions rather than the expected edit/create permissions, I would be very surprised and assume there was a bug. This is especially the case because the implementation plan explicitly outlines using Django's permissions model for this.

Furthermore, if we added additional user groups in the future, avoiding Django's intended permissions model would likely complicate the integration of those groups and the management of their permissions and access.

If there are significant trade-offs that make not using Django's permission's model the right call, they are surely unusual enough to be documented in the code that causes those trade-offs. Likewise, we should update the implementation plan to avoid confusion here.

Notably, using the user permissions as intended would allow model admin permission checks to ignore details like the specific user group names, checking `is_superuser`, and so forth. It's, to my mind and I presume to the regular Django user, given Django's built-in tooling, much more natural and expected to call `user.has_perm(request, "decision_change")`, for example, rather than try to remember all the particular details of when a user would have one permission or other. In other words, relying on the data model and configuration to express the permitted actions rather than business logic if/else checks.

At the very least, we can absolutely still use the model admin permissions methods, even if we do not rely on the Django user groups for some reason: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_view_permission

Using those methods makes the intended permissions much clearer than burying them in a branch of an action method. They also present a natural place to document the decision for how permissions are determined (if indeed we will not use the user groups permissions as intended by the implementation plan).

api/api/admin/media_report.py
```diff
@@ -441,69 +441,84 @@ def decision_create_view(self, request, object_id):
...
+                messages.info(
+                    request,
+                    "The media object has been marked as sensitive.",
                 )
+        else:
+            logger.warning(
+                "Form is invalid",
+                **form.cleaned_data,
+                errors=form.errors,
+            )
```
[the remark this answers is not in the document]
dhruvkb: That's a good idea, updated the code to use messages in 8100e24.

api/api/admin/media_report.py
```diff
@@ -441,69 +441,84 @@ def decision_create_view(self, request, object_id):
...
+                media_obj=media_obj.id,
+            )
+
+            reports = form.cleaned_data["reports"]
+            count = reports.update(decision=decision)
+            logger.info(
+                "Decision recorded in reports",
+                report_count=count,
+                decision=decision.id,
+            )
```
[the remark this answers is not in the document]
dhruvkb: Fair point, changed the logs to `DEBUG` in 8074cca.

api/api/admin/media_report.py
```diff
@@ -441,69 +442,79 @@ def decision_create_view(self, request, object_id):
...
-                    decision=decision,
-                    media_obj=media_obj,
-                )
-                logger.info(
-                    "Through model created",
-                    through=through.id,
-                    decision=decision.id,
-                    media_obj=media_obj.id,
-                )
+        if not request.user.has_perm(f"api.add_{self.media_type}decision"):
```
dhruvkb, line 450: @sarayourfriend your review was accurate. I was operating under a limited understanding of Django's permission model.

I have updated the PR to be more idiomatic and use `user.has_perm` instead of checking for `is_superuser` or membership to a specific group.

sarayourfriend: Sounds good, glad that was easy to clarify.

One more question: shouldn't this check be in the `has_add_permission` method of the model admin?

https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_add_permission

I'm actually surprised it isn't automatically handled as part of the model admin permissions :thinking:. I thought these two aspects of the base ModelAdmin would automatically handle these checks:

https://github.com/django/django/blob/0e49a8c3bd9119795525d9f076f73740741479b7/django/contrib/admin/options.py#L1844-L1845

and 

https://github.com/django/django/blob/0e49a8c3bd9119795525d9f076f73740741479b7/django/contrib/admin/options.py#L1836-L1837

If we're manually implementing the checks, we should have unit tests for them. Ideally we can rely on the base class behaviour... 

At the very least I think this line can be changed to:

```suggestion
        if not self.has_add_permission(request):
```

sarayourfriend (changes requested):
One more clarification/improvement, but otherwise this LGTM. I won't get to do the final review as I'll be AFK starting tomorrow, but please please make absolutely certain that we are not duplicating functionality from Django Admin. Django's documentation on permissions is extensive and in my experience things "just work" if they are configured correctly. As I said before, any deviations from that definitely need documentation to explain why they must exist and why the base class behaviour is insufficient. On top of that, they also need unit tests.

api/api/admin/media_report.py
dhruvkb: This view (`decision_create_view`) exists in the `MediaListAdmin` class which is the admin-model class for media models (i.e. `Image` and `Audio`) so the `has_add_permission` in that class refers to the user's permission to add a new media item, and not a new decision, which is the intent of this statement.

obulat (approved):
The requested changes were implemented, and locally, works as expected. Nice logging clean up.

api/api/admin/media_report.py
```diff
@@ -441,69 +442,79 @@ def decision_create_view(self, request, object_id):
...
+                decision=decision.id,
+            )
+
+            if decision.action in {
+                DecisionAction.DEINDEXED_COPYRIGHT,
+                DecisionAction.DEINDEXED_SENSITIVE,
+            }:
+                messages.debug(
+                    request,
+                    "The media object has been deindexed from ES and deleted from DB.",
```
krysal, line 502: It would be more accurate to say it's the deleted from the media table, as deleted records are kept in a separate table.

```suggestion
                    f"The media object has been deindexed from ES and deleted from the {self.media_type} table.",
```

krysal (commented):
I was able to moderate a report created via a curl. Overall, that main flow looks good to me. However, I was **not** able to create a report for another image using the moderator user given the image list appears empty until a report is created. I think we did something on this end to prevent problems in the production API.

**Default admin image list view**

Screenshot 2024-07-24 at 11-31-25 Select image to view Django site admin

**Admin image list view with a reported image**

Screenshot 2024-07-24 at 11-41-36 Select image to view Django site admin

I don't see much use in creating reports for media that is already reported. If it's still considered important that moderators should be able to create reports internally then this is pending, but it can be addressed separately. What do you think @dhruvkb?

dhruvkb:
> the image list appears empty

The default filter on the right side is set to "Moderation queue" that's why the list is empty. If you switch to "All", all the media items will become visible and you can report and/or moderate any media item.

api/api/admin/media_report.py
dhruvkb: That's true, but technically all the data is gone. Only the UUID of the media item is left in the other table.

krysal (approved):
That's an excellent note and detail! In that case, this g2g. Thanks for clarifying.

Issue #4108, opened 2024-04-14
Reduce unused services in Docker Compose


dhruvkb:
## Problem

The following services in Docker Compose are largely unused.

- <s>`load_to_s3`</s>

  <s>I could not find any occurrences of this string in the project except for the `docker-compose.yml` file. This leads me to suspect it might actually be completely unused.</s>

- <s>`nginx`</s>

  <s>This runs the `nginx` target from the API `Dockerfile`. It's purpose in the local dev environment is unclear and building it when initialising the API slows down CI.</s>

- `proxy`

  `proxy` was added long back to test for HTTPS with the API server during local development. It also introduces a [dependency on `mkcert`](https://docs.openverse.org/general/general_setup.html#mkcert) so removing it will have twice the benefit.

They should be removed or, if they are in fact used, comments should be added to list down what they are used for and where.

## Additional context

This is part of a bigger improvement of the Docker Compose configuration that involves upgrading to v3 #4046 and is closely linked to #1910. This was also raised by @krysal in #1009.

AetherUnbound:
> * `load_to_s3`
>       I could not find any occurrences of this string in the project except for the `docker-compose.yml` file. This leads me to suspect it might actually be completely unused.

This service [loads some testing data into the local minio instance](https://github.com/WordPress/openverse/blob/eae8c19ead6c02369d5add22cd8c49c39f93e095/docker-compose.yml#L89-L104), which is used currently for iNaturalist but could be expanded in the future. It should not be removed.

dhruvkb:
@AetherUnbound thanks for clarifying! There might be ways to do this that don't involve a dedicated service but at the very least documenting this on the service itself would be helpful. It's also my bad that I did not look closely enough before making the assumption.

sarayourfriend:
The purpose of the API nginx image is to test the Nginx configuration we run in the live environments. We can make it optional by removing it from the api compose profile, (maybe put it in something like `api-extra`?), but it absolutely should not be removed.

We've discussed several times in passing getting rid of the `proxy` one for the API. I would support doing so: we don't need to regularly test the local API with HTTPS. We need to keep the other API nginx one for being able to easily verify changes to the API's nginx configuration, which do indeed happen on occasion, especially logging changes. Example: #4569

dhruvkb:
@sarayourfriend the `proxy` service was removed about 2 months ago in #4605. Good to know that the `nginx` image should not be removed.

PR #4605, opened 2024-07-06
Remove `proxy` service and make API NGINX setup similar to the frontend

dhruvkb:
## Fixes

Partially addresses #4108 by @dhruvkb

## Description

This PR
- removes the service `proxy` which was meant to test the API running under HTTPS locally. This was rarely or maybe even never used.
  - also removes the documentation and scaffolding files related to this service
- makes the API NGINX configuration identical to the frontend
  - also uses the `snippets/` pattern used in the frontend for common config in API locations
- updates Docker Compose profiles and `just` recipes so that the `api_extra` profile depends on the `api` profile

## Testing Instructions

Check out this PR and verify that the `nginx` service builds and runs as expected.

```bash
just build nginx && just api/up-extra && just logs nginx
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/justfile
```diff
@@ -50,6 +50,7 @@ up *flags:
 
 # Bring up services specific to the API profile, in addition to the API server
 up-extra *flags:
+    just up {{ flags }}
```
AetherUnbound, line 53: If this is also calling `just ../up` with the correct compose profile, why should this also be called? If we needed it, shouldn't we just add `api` to the list of `COMPOSE_PROFILES` to spin up?

api/nginx/snippets/django_location.conf
```diff
@@ -0,0 +1,6 @@
+proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
AetherUnbound, line 1: This is great!

AetherUnbound (commented):
The code changes look good to me and everything (but the removed container) still runs fine for me. My concern is though that, even if we don't use that service often, we still added it for a reason. I seem to remember it being related to an incident or some issue. Can we make sure we still have no use for the HTTPS setup before removing it?

dhruvkb:
@AetherUnbound yes, we had an issue where the links to the thumbnail and detail endpoints were `http://` instead of `https://`. But in the end, we fixed that using a custom field that automatically rewrites the URL scheme based on a value from the settings.

https://github.com/WordPress/openverse/blob/90f87b8712b50cf05401de91eb5805b067efeb85/api/api/serializers/fields.py#L9

Since that issue was resolved, I have not once run the API locally as HTTPS and do not anticipate any future situation where we would want or need to.

Additionally removing this removes a dependency on `mkcert` and brings us one step closer to `ov` having everything we need to run Openverse locally.

api/justfile
dhruvkb: You are right! I had not realised that. Fixed in 8379b8b.

dhruvkb: You are right! I had not realised that. Fixed in 8379b8b.

AetherUnbound:
> @AetherUnbound yes, we had an issue where the links to the thumbnail and detail endpoints were `http://` instead of `https://`. But in the end, we fixed that using a custom field that automatically rewrites the URL scheme based on a value from the settings.

Got it, thank you for giving me context on that! This definitely gives me more confidence in being able to remove this service 😄

AetherUnbound (approved):

PR #4610, opened 2024-07-12
Add `catalog_cleaner` DAG

### LINKED ISSUE #3415: Use the `batched_update` DAG with stored CSVs to update Catalog URLs

krysal:
## Fixes

Fixes #3415 by @krysal 

## Description

This PR adds a DAG similar to the `bactched_update` but is dedicated to using the files from AWS S3 generated from the ingestion server. I created a new DAG since the `batched_update` would need more tweaks than just loading the data from S3 into a table since the update is performed by joining the newly created table with `image`, and also does the batch updates in parallel using the Dynamic Task Mapping feature of Airflow. Given several `batched_update` runs have been performed with the popularity calculations, it suggests it's possible to parallelize this task as long as the rows to be modified do not overlap. The ingestion server also similarly applies the updates.

This is a proof of concept to see the feasibility of parallelizing the DB updates, making a separate DAG we don't interfere with those that depend on the `batched_update` DAG, and if it's successful (which I don't see why it wouldn't be) we can discuss how to generalize it, whether if modiying the `batched_update` of creating another to use with temporary tables from S3 files.

## Testing Instructions

You will need to have some "dirty" rows in the catalog and their corresponding TSV files containing the correct data to fix them. 

1. To create a TSV, spin up the catalog stack, `./ov just catalog/up`, and if you have `psql` locally try this:
```sh
PGPASSWORD=deploy psql -h localhost -p 50255 openledger deploy -c "COPY (SELECT identifier, url FROM image ORDER BY created_on DESC LIMIT 15) TO STDOUT WITH DELIMITER E'\t' CSV;" >> url.tsv
```
You should then have a file of `<identifier>	<url>` without quotes or a header like the following:
```
3074629a-9934-464b-aca6-b517f4b7cf80	https://pd.w.org/2024/06/8616676f5d15f44f7.79186143-2048x1366.jpg
e3637802-7ff5-4972-b1b4-554e97d921f1	https://pd.w.org/2024/06/85667a9763e32215.10445933-1152x2048.jpg
aec4bdef-6315-4a5d-a997-727891d2f00a	https://pd.w.org/2024/06/121666b021b681b06.28758133-1536x2048.jpg
...
```

2. Then modify the URLs for the same rows. You can use a query like the following to remove the protocol:
```sql
UPDATE image SET url = ltrim(url, 'https://') WHERE identifier IN (
	SELECT identifier FROM image ORDER BY created_on DESC LIMIT 15
);
```
3. Upload your new file to MinIO: http://localhost:5011/. Save the path to the bucket as you'll need it as a setting later.

4. Go to the Airflow U, unpause the `catalog_cleaner` DAG, and trigger it with a configuration that fits your test case. E.g.:
```json
{
    "batch_size": 3,
    "column": "url",
    "s3_bucket": "openverse-catalog",
    "s3_path": "shared/data-refresh-cleaned-data/url_test.tsv"
}
```

5. Verify the changes were applied in the DB to the rows.
```sh
./ov just catalog/pgcli
```

```sql
SELECT identifier, url FROM image ORDER BY created_on DESC LIMIT 15;
```

Try different configurations for `batch_size` and/or the `CLEANER_MAX_CONCURRENT_DB_UPDATE_TASKS` Airflow variable to see how it works. You can add `time.sleep(seconds)`  before returning in the `update_batch` function to better appreciate the concurrency.

## Checklist

- [x] I ran the DAG documentation generator (`./ov just catalog/generate-docs` for catalog PRs) or the media properties generator (`./ov just catalog/generate-docs media-props` for the catalog or `./ov just api/generate-docs` for the API) where applicable.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/database/catalog_cleaner/catalog_cleaner.py
+++ b/catalog/dags/database/catalog_cleaner/catalog_cleaner.py
@@ -50,9 +50,10 @@ def count_dirty_rows_and_notify(temp_table_name: str, task: AbstractOperator = N
         handler=single_value,
         task=task,
     )
+    column = temp_table_name.split("_")[-1]
     notify_slack.function(
-        text=f"Starting the cleaning process in upstream DB. Expecting {count:,} rows"
-        f" affected given `{temp_table_name}` table."
+        text=f"Starting the cleaning process in upstream DB for column `{column}`."
+        f" Expecting {count:,} rows affected given `{temp_table_name}` table."
     )
     return count
 
--- a/catalog/dags/database/catalog_cleaner/catalog_cleaner.py
+++ b/catalog/dags/database/catalog_cleaner/catalog_cleaner.py
@@ -12,10 +12,15 @@ from airflow.decorators import dag, task
 from airflow.models import Variable
 from airflow.models.abstractoperator import AbstractOperator
 from airflow.models.param import Param
-from airflow.utils.trigger_rule import TriggerRule
+from airflow.operators.python import get_current_context
 
 from common.constants import DAG_DEFAULT_ARGS, POSTGRES_CONN_ID
-from common.sql import PGExecuteQueryOperator, single_value
+from common.sql import (
+    RETURN_ROW_COUNT,
+    PGExecuteQueryOperator,
+    PostgresHook,
+    single_value,
+)
 from database.batched_update.batched_update import run_sql
 from database.catalog_cleaner import constants
 
@@ -38,6 +43,47 @@ def count_dirty_rows(temp_table_name: str, task: AbstractOperator = None):
     return count
 
 
+@task
+def get_batches(total_row_count: int, batch_size: int) -> list[tuple[int, int]]:
+    # batch_list = []
+    # batch_start = 0
+    # while batch_start <= total_row_count:
+    #     batch_end = min(batch_start + batch_size, total_row_count)
+    #     batch_list.append({
+    #         "batch_start": batch_start,
+    #         "batch_end": batch_end
+    #     })
+    #     batch_start += batch_size
+    # return batch_list
+    return [(i, i + batch_size) for i in range(0, total_row_count, batch_size)]
+
+
+@task(map_index_template="{{ index_template }}")
+def update_batch(
+    batch: tuple[int, int],
+    temp_table_name: str,
+    column: str,
+    task: AbstractOperator = None,
+):
+    batch_start, batch_end = batch
+    logger.info(f"Going through row_id {batch_start:,} to {batch_end:,}.")
+    context = get_current_context()
+    context["index_template"] = f"{batch_start}__{batch_end}"
+
+    pg = PostgresHook(
+        postgres_conn_id=POSTGRES_CONN_ID,
+        default_statement_timeout=PostgresHook.get_execution_timeout(task),
+    )
+    query = constants.UPDATE_SQL.format(
+        column=column,
+        temp_table_name=temp_table_name,
+        batch_start=batch_start,
+        batch_end=batch_end,
+    )
+    count = pg.run(query, handler=RETURN_ROW_COUNT)
+    return count
+
+
 @dag(
     dag_id=DAG_ID,
     default_args={
@@ -49,6 +95,7 @@ def count_dirty_rows(temp_table_name: str, task: AbstractOperator = None):
     catchup=False,
     tags=["database"],
     doc_md=__doc__,
+    render_template_as_native_obj=True,
     params={
         "s3_bucket": Param(
             default="openverse-catalog",
@@ -61,7 +108,9 @@ def count_dirty_rows(temp_table_name: str, task: AbstractOperator = None):
             description="The S3 path to the TSV file within the bucket.",
         ),
         "column": Param(
-            type="string", description="The column of the table to apply the updates."
+            type="string",
+            enum=["url", "creator_url", "foreign_landing_url"],
+            description="The column of the table to apply the updates.",
         ),
         # "table": Param(type="str", description="The media table to update."),
         "batch_size": Param(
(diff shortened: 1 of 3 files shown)


catalog/dags/database/catalog_cleaner/catalog_cleaner.py
```diff
@@ -0,0 +1,216 @@
...
+    """Get the number of rows in the temp table before the updates."""
+    count = run_sql.function(
+        dry_run=False,
+        sql_template=f"SELECT COUNT(*) FROM {temp_table_name}",
+        query_id=f"{temp_table_name}_count",
+        handler=single_value,
+        task=task,
+    )
+    logger.info(f"Found {count:,} rows in the `{temp_table_name}` table.")
+    return count
```
AetherUnbound, line 54: Nit: since `run_sql` is already a `@task` and this merely adds a `logger.info` line (with information which could be pieced together from the XComs and arguments), maybe it makes more sense to use `run_sql` directly with a `.override(task_id="count_dirty_rows")`?

AetherUnbound:
@krysal For the testing instructions on this one, should we run that command against the upstream database or the API one? And what format should the TSV be in (e.g. what columns do we need, does it need to be quoted, etc.). If possible, do you mind supplying a set of commands we can use (similar to the one you provided for altering the records) to generate and upload the TSV so testing across contributors can be consistent? Thank you!

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
```diff
@@ -0,0 +1,216 @@
...
+    column: str,
+    task: AbstractOperator = None,
+):
+    batch_start, batch_end = batch
+    logger.info(f"Going through row_id {batch_start:,} to {batch_end:,}.")
+
+    # Includes the formatted batch range in the context to be used as the index
+    # template for easier identification of the tasks in the UI.
+    context = get_current_context()
+    context["index_template"] = f"{batch_start}__{batch_end}"
```
AetherUnbound, line 76: Woah, this is neat!! I didn't know this could be evaluated _after_ the task had started 😮

AetherUnbound (commented):
Adding some preliminary thoughts before testing this locally!

krysal:
> @krysal For the testing instructions on this one, should we run that command against the upstream database or the API one? And what format should the TSV be in (e.g. what columns do we need, does it need to be quoted, etc.). If possible, do you mind supplying a set of commands we can use (similar to the one you provided for altering the records) to generate and upload the TSV so testing across contributors can be consistent? Thank you!

This is for cleaning the catalog so all the operations are performed over the `upstream` DB. I added a handy command that should get you the file if you have `psql` locally. If not, then it can be applied inside one of the docker containers, it just would be one or two more steps to get it to the host. Let me know if that works for you.

The columns that need to be cleaned in production (that is, for which we have files) are specifically `URL`, `creator_url`, and `foreign_landing_url`. The generated files are quite simple: Paris of identifiers and the fixed URL value, no quotes and no header.

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
krysal: Good observation! This remembers me that I left this as a quick replacement for the initial message. I can do that change and add the slack notification before the update.

AetherUnbound (approved):
Thanks _so_ much for the testing instructions, I was able to execute them perfectly! And that's a good thing to note about `psql`, sounds like it's a tool we should add to the `ov` container as well for this sort of thing 😄 I'll make an issue for it.

I was able to test this from end-to-end and it worked just as expected, including the batch mapped indices which is very neat. Great work on this!

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
```diff
@@ -0,0 +1,219 @@
...
+    """Get the number of rows in the temp table before the updates."""
+    count = run_sql.function(
+        dry_run=False,
+        sql_template=f"SELECT COUNT(*) FROM {temp_table_name}",
+        query_id=f"{temp_table_name}_count",
+        handler=single_value,
+        task=task,
+    )
+    notify_slack.function(
+        text=f"Starting the cleaning process in upstream DB. Expecting {count:,} rows"
```
obulat, line 54: Do we still call the DB upstream, or is it `catalog` DB (vs the API db) ?

obulat: Would it be possible to add the field being cleaned (`url`, `foreign_landing_url` etc) to this notification?

obulat (approved):
Thank you for the detailed testing instructions, @krysal ! It worked well locally (once I figured out the connection details). Hope this works in production the first time we run it 🤞

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
stacimc: Same, TIL!!

stacimc (approved):
Looks great to me -- thanks for the very clear description in addition to the testing instructions. The justification seems sound to me, and I'm also excited to test parallelization of the batches; I don't recall any reason for avoiding it. 🚀

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
krysal: > Do we still call the DB upstream, or is it `catalog` DB (vs the API db) ?

That's a good question. I use both terms intermittently, which can be confusing for sure 😅  Which one do you prefer?

> Would it be possible to add the field being cleaned (url, foreign_landing_url etc) to this notification?

It's part of the temporary table name but thinking again, it's better to make it explicit. Done! :)

krysal:
Thank you all for the review folks! ✨ I'll be starting the DAG on Monday.

catalog/dags/database/catalog_cleaner/catalog_cleaner.py
obulat: > Which one do you prefer?
I would prefer `catalog`.
Although you could say that the API is also a "catalog", but I guess I look at it more from the point of view of our stacks. On another hand, since this database (upstream/catalog) is the single source of truth that contains all information which can later be filtered in the API database, it makes sense to consider it "**the** catalog"

PR #4623, opened 2024-07-17
Migrate `utilities/dead_link` to PDM

obulat:
## Fixes

Related to #4166 by @dhruvkb

## Description

This PR migrates the `dead_links` utility to PDM.

To make the redis connection parameters (host and port) configurable, it converts the utility to a `click` script with `--port` and `--host` parameters.

It also moves away from a py_redis library that was last updated 3 years ago to `redis`, sets the port for redis connection and also uses `tqdm.write` to make sure that the progress bar is always at the bottom of the terminal.

## Testing Instructions

`cd utilities/dead_links`
`../../ov pdm install`
`../../ov pdm run dead_link_tally.py`
You should see the error message saying that you couln't connect to Redis.

Maintainers can also connect to Redis dev server and use the dev port to run `../../ov pdm run dead_link_tally.py --port=DEV_REDIS_PORT` to test running this script with staging data.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
> Maintainers only: connect to redis prod.

It's not really necessary for testing, to be clear, and best to avoid if not absolutely necessary to directly connect. You can get the same data locally by requesting search with `filter_dead=True` and setting the link validation timeout to something very low _or_ modifying the code to specially force the data to appear.

At the very least, connecting to `dev` Redis should be preferred when production data isn't strictly necessary but it cannot be reproduced locally for some reason.

utilities/dead_links/dead_link_tally.py
```diff
@@ -14,7 +14,7 @@
 from tqdm import tqdm
 
 
-redis = Redis("localhost", decode_responses=True)
+redis = Redis("localhost", port=6399, decode_responses=True)
```
sarayourfriend, line 17: Really should be configurable. I don't like the idea of encoding this to always and specifically read from the default port we map the prod redis to :frowning:. The script is fine in its current implementation, but if it changed for some reason, we shouldn't be using _production redis_ to test and iterate on those changes.

Can you either: revert this from this PR and open a separate issue to make this configurable as needed (preferably using argument parser for CLI args), or else go ahead and make that change in this PR?

To make it configurable, it should be enough to create an ArgumentParser with `--host` and `--port` arguments that default to `localhost` and `None`, and then use those. Don't even need to wrap this in `def main()`, and `if __name__ ...` if you want to keep the line change low.

I think it would be better to put it in a separate PR though.

sarayourfriend (approved):
Left one comment but it isn't a blocker. If you'd rather merge this as is, that's fine by me, but a fast-follow issue + PR to make the port and host configurable to discourage using production unless that's very intentional is, I think, a must for this code.

krysal (approved):
This looks more like a full refactor 😄 It worked for me locally with:

```
../../ov pdm run dead_link_tally.py --port=50263
```

Great!

PR #4628, opened 2024-07-17
Update pnpm and use corepack

### LINKED ISSUE #3807: Replace `pnpm/action-setup` with Corepack
## Problem

We use Corepack to manage pnpm locally and it works really well. But in CI we use `pnpm/action-setup` to provision pnpm for us.

## Description

It should be possible, even simpler, to use Corepack to provision pnpm in CI as well. We should explore that as an option.

## Additional context

This idea was provided by @obulat in https://github.com/WordPress/openverse/pull/3801#discussion_r1494737001. The linked commment includes a code reference too!

obulat:
## Fixes

Fixes #3807 by @dhruvkb

## Description

This PR replaces `pnpm/action-setup@v3` with `corepack enable` for setting up `pnpm` in the CI and Docker containers.
For the Playwright Docker containers, corepack looks into the closest `package.json` file to find the pnpm version, so I added it in the `frontend/package.json`

Since we no longer need to provide the pnpm version to the set up script (it's extracted from `package.json`'s `packageManager` field), I removed the related env variables.

## Testing Instructions

The CI should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -41,11 +41,11 @@ runs:
         pip install -U pip
         pip install pipenv pdm
 
-    # Node.js
     - name: Setup pnpm
       if: inputs.setup_nodejs == 'true'
-      uses: pnpm/action-setup@v3
-      # Automatically reads `packageManager` field from root `package.json`.
+      shell: bash
+      # Corepack automatically reads `packageManager` field from root `package.json`.
+      run: corepack enable pnpm
 
     - name: Setup Node.js
       if: inputs.setup_nodejs == 'true'
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -44,7 +44,6 @@ runs:
     - name: Setup pnpm
       if: inputs.setup_nodejs == 'true'
       shell: bash
-      # Corepack automatically reads `packageManager` field from root `package.json`.
       run: corepack enable pnpm
 
     - name: Setup Node.js
--- a/frontend/Dockerfile.playwright
+++ b/frontend/Dockerfile.playwright
@@ -4,9 +4,10 @@ ARG PLAYWRIGHT_VERSION
 
 FROM mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION}-jammy
 
-ARG PACKAGE_MANAGER
-
 COPY package.json /
 
+# Ensure the Playwright container's pnpm cache folder exists and is writable
+RUN mkdir -p /.cache/node/corepack/ && chmod -R 777 /.cache/node/corepack/
+
 # Requires `packageManager` field to be present in `frontend/package.json`.
-RUN npm install -g $PACKAGE_MANAGER
+RUN corepack enable pnpm
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,6 +3,7 @@
   "description": "Openverse frontend",
   "private": true,
   "version": "2.2.1",
+  "packageManager": "pnpm@9.5.0",
   "scripts": {
     "predev": "pnpm install && pnpm i18n:no-get",
     "dev": "run-p dev:only 'i18n:en --watch'",
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -10,7 +10,6 @@ ARG FRONTEND_NODE_VERSION
 FROM docker.io/node:${FRONTEND_NODE_VERSION}-alpine AS builder
 
 # Install system packages needed to build on macOS
-# Corepack automatically reads `packageManager` field from closest `package.json`.
 RUN apk add --no-cache --virtual .gyp python3 make g++ \
   && corepack enable pnpm
 
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -9,12 +9,10 @@ ARG FRONTEND_NODE_VERSION
 
 FROM docker.io/node:${FRONTEND_NODE_VERSION}-alpine AS builder
 
-# Automatically use the right version of pnpm.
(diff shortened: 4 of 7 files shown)


frontend/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Openverse frontend",
   "private": true,
   "version": "2.2.1",
+  "packageManager": "pnpm@9.5.0",
```
dhruvkb, line 6: Please correct me if I'm wrong but I was under the impression that `packageManager` is only read from the `package.json` in the repo root.

frontend/Dockerfile.playwright
```diff
@@ -4,9 +4,10 @@ ARG PLAYWRIGHT_VERSION
 
 FROM mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION}-jammy
 
-ARG PACKAGE_MANAGER
-
```
dhruvkb, line 8: If the arg is no longer used, it can be removed from a number of other places.

https://github.com/WordPress/openverse/blob/25a80213ebc13e8284668d8d86e53a0dee129103/frontend/docker-compose.playwright.yml#L7

https://github.com/WordPress/openverse/blob/25a80213ebc13e8284668d8d86e53a0dee129103/frontend/bin/playwright.sh#L12-L13

.github/actions/setup-env/action.yml
```diff
@@ -41,11 +41,11 @@ runs:
...
         pip install pipenv pdm
 
-    # Node.js
     - name: Setup pnpm
       if: inputs.setup_nodejs == 'true'
-      uses: pnpm/action-setup@v3
-      # Automatically reads `packageManager` field from root `package.json`.
+      shell: bash
+      # Corepack automatically reads `packageManager` field from root `package.json`.
+      run: corepack enable pnpm
```
dhruvkb, line 48: 👍 TIL that `corepack enable` takes the name of the package manager as an argument!

dhruvkb (approved):
Approving even though it's a draft because it's great. I love Corepack and happy to see it being used more.

frontend/package.json
obulat: This is necessary for the Playwright Docker setup, not the frontend (I think frontend can use the root package.json). The [node corepack docs](https://nodejs.org/api/corepack.html#corepack_configuring_a_package) say "_The Corepack proxies will find the **closest** [package.json](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions) file in your current directory hierarchy to extract its ["packageManager"](https://nodejs.org/api/packages.html#packagemanager) property._"

.github/actions/setup-env/action.yml
obulat: It doesn't need to have the package name, but there were warnings in some places without it.

.github/actions/setup-env/action.yml
```diff
@@ -41,11 +41,11 @@ runs:
         pip install -U pip
         pip install pipenv pdm
 
-    # Node.js
     - name: Setup pnpm
       if: inputs.setup_nodejs == 'true'
-      uses: pnpm/action-setup@v3
-      # Automatically reads `packageManager` field from root `package.json`.
+      shell: bash
+      # Corepack automatically reads `packageManager` field from root `package.json`.
```
sarayourfriend, line 47: This kind of comment is sort of just documenting what Corepack _is_ isn't it? Do we need this kind of comment? I'd argue it wasn't necessary for `pnpm/action-setup` either.

What's the benefit of using corepack here instead of the action, by the way? Just consistency or is there something else? The action does _quite_ a lot of things, including basic clean up steps that are apparently necessary for GitHub actions (store prune at the end of the job, for example).

Is corepack by itself, without any additional support steps or jobs, sufficient to actually replace what `pnpm/action-setup` does? It's not only installing pnpm, after all: https://github.com/pnpm/action-setup/blob/master/src/index.ts#L8-L17

obulat: The main reason is consistency and the removal of one extra dependency that needs to be updated.

I've seen many large projects use `corepack enable` instead of `pnpm/action-setup`. Do you think we really need all of the steps that the action does? Is store pruning important during the CI, if we are discarding the containers at the end anyway?

sarayourfriend: I have no idea what the purpose of store pruning is at the end of the action, all I know is that `pnpm/action-setup` does it, and lots of other actions have similar clean up type steps. I have not read about why that is or what the advice is for this (I've never written of this nature). To clarify, store pruning is on the host, not inside the docker images, so relevant for when we run `pnpm install` in the next step after this one.

Based on a quick search, I can't find anything clear or definitive on why clean up steps are necessary, other than suggestions that while runners are intended to be ephemeral, clean up helps prevent any accidental leaks between runners. In our case, the pnpm store shouldn't have an issue in that regard, not that I can think of off the top of my head. We don't download and proprietary packages from private repositories, for example, so it isn't like there is IP that could leak between runners if ephemerality didn't work as expected. It also seems like it might be more relevant for self-hosted runners (perhaps those have different behaviour in this regard).

frontend/Dockerfile.playwright
obulat: Removed in [6f3665e](https://github.com/WordPress/openverse/pull/4628/commits/6f3665eff3ee02ca17c41517abebd5c2a57e91ce)

AetherUnbound (approved):
Don't have much to add beyond what others have shared, but I'm glad we can get rid of some of the env vars here! 🙂

PR #4653, opened 2024-07-24
Fix HTML creator crediting incorrect image tags

zackkrida:
Fixes the incorrect HTML for `` tags in our "Credit the Creator" section HTML. Previously, image tags were rendered as `</img>` which is invalid HTML. 

Now, `` and other self-closing HTML tags will be rendered as `` (note the closing "/" at the end of the tag) which is valid.


AetherUnbound (approved):
LGTM, I was able to verify the fix locally!

dhruvkb (approved):

PR #4659, opened 2024-07-25
Nuxt 3 robots.txt fixes

zackkrida:
Adds a dynamically-generated robots.txt using the Nuxt SEO `robots` module. The module automatically prefixes robots.txt routes with our locale paths as well, which is nice. It is more verbose than the wildcard approach currently on https://openverse.org/robots.txt, but still works well.

It also automatically adds robots meta tags to pages based on these rules.

The module also automatically disables crawling in all non-production environments.  

## Testing instructions

- `ov j f` and visit localhost:8443/robots.txt to verify that the robotx.txt blocks all traffic in non-production setttings (as determined by the `site.indexable` property in the nuxt config.
- `ov j frontend/run build && ov j frontend/run start` and visit localhost:8443/robots.txt to verify that the production file contains the correct rules, including localized paths.
- Check that pages have the correct meta tags


frontend/nuxt.config.ts
```diff
@@ -28,6 +28,42 @@ const favicons = [
...
+ * @see {@link https://nuxtseo.com/robots/guides/nuxt-config|Robots Config Rules}
+ */
+const robots = {
+  userAgent: "*",
+  disallow: [
+    "/search/audio/",
+    "/search/image/",
+    "/search/",
+    "/image/",
+    "/audio/",
```
zackkrida, line 57:
```suggestion
```

zackkrida:
@obulat I've added the changes from #4660 here as well.

AetherUnbound:
> @obulat I've added the changes from #4660 here as well.

@zackkrida how did you want to handle this, should we close that PR or is it still worth merging? 😅

frontend/nuxt.config.ts
```diff
@@ -107,6 +137,7 @@ export default defineNuxtConfig({
     },
   },
   site: {
+    indexable: isProdNotPlaywright,
```
zackkrida, line 140: I _think_ this is right? We want `indexable` to be true on production but not on staging.

PR #4663, opened 2024-07-26
Replace retry with backoff, add backoff to all Freesound requests

### LINKED ISSUE #4594: Flaky 503s when trying to fetch Freesound audioset info
## Description

Several production DagRuns of the Freesound workflow have failed due to intermittent 503s when trying to get audio set info, which succeed when the DAG is later manually retried. See July 2024 runs of the DAG for examples.

We could address this a couple of ways, given that the problem is intermittent:
* Add an additional `retry` to [get_set_info](https://github.com/WordPress/openverse/blob/f556ccfb6b108a76add2d448576feeea60b4df82/catalog/dags/providers/provider_api_scripts/freesound.py#L132) similar to the existing special handling of flaky exceptions on [get_audio_file_size](https://github.com/WordPress/openverse/blob/f556ccfb6b108a76add2d448576feeea60b4df82/catalog/dags/providers/provider_api_scripts/freesound.py#L160)
* Increase the number of timeouts overall for Freesound
* Keep track of records which 503 and skip them during ingestion, then automatically retry them once the rest of ingestion has finished. This is likely more effort than is merited.
* Worst case, discard individual records whose audio set info can't be obtained, and report the number of skipped records to Slack. This is better than continuing to use `SKIP_INGESTION_ERRORS`, which will cause an entire batch to be skipped when one record exceeds retries.

AetherUnbound:
## Fixes

Fixes #4594 by @stacimc

## Description

> [!WARNING]
> This **should not** be merged until a deployment can be made immediately after, since it introduces new dependencies.

This PR makes two major changes:
- Replaces the `retry` with `backoff`
- Adds an exponential backoff to all requests made to Freesound during ingestion

The latter change is relatively straightforward: we encounter tons of exceptions during the Freesound ingestion process that seem to be transient and ultimately get resolved when checked later. However this causes significant disruptions in the execution of the DAG itself, since it often requires that maintainers restart the DAG with `SKIP_INGESTION_ERRORS` enabled for it. This change allows requests to be retried with an exponential backoff up to a max time (2 minutes for now) if they match the HTTP codes 400 or 503. On all other codes, the exception is raised and retries (outside of the normal ones for the `DelayedRequester`) are not attempted.

The former change is motivated by the above. [`retry`](https://pypi.org/project/retry/) is a Python library that hasn't been updated since 2016, and only supports fairly basic retry conditions. [`backoff`](https://pypi.org/project/backoff/) on the other hand has been updated as recently as 2022 and supports much more complexity in retry conditions. With the above case, we only wanted to retry on `HTTPErrors` with a specific return code; that was not possible with `retry` but is now possible with `backoff`. This should give us some additional flexibility moving forward too!

## Testing Instructions

I added new tests specifically for the conditions mentioned above. You can also run the `freesound_workflow` DAG locally to ensure everything behaves normally (and maybe encounter one of the above cases if you're lucky!).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/providers/provider_api_scripts/freesound.py
+++ b/catalog/dags/providers/provider_api_scripts/freesound.py
@@ -36,6 +36,7 @@ class FreesoundDataIngester(ProviderDataIngester):
     endpoint = f"https://{host}/apiv2/search/text"
     providers = {"audio": prov.FREESOUND_DEFAULT_PROVIDER}
     flaky_exceptions = (SSLError, ConnectionError)
+    flaky_error_codes = {400, 503}
     preferred_preview = "preview-hq-mp3"
     preview_bitrates = {
         "preview-hq-mp3": 128000,
@@ -52,6 +53,20 @@ class FreesoundDataIngester(ProviderDataIngester):
         }
 
         super().__init__(*args, **kwargs)
+
+    # Freesound often tends to be flaky when making requests during ingestion.
+    # For the flaky HTTP errors above, we wait some time before trying the request
+    # again in the hopes that the upstream issues resolve in that time:
+    #   * 400 - Although framed as bad requests, these appear to succeed later
+    #   * 503 - Service unavailable, so we should just try to wait
+    get_response_json = backoff.on_exception(
+        backoff.expo,
+        HTTPError,
+        max_time=60 * 2,
+        # Raise all other errors
+        giveup=lambda e: e.response.status_code
+        not in FreesoundDataIngester.flaky_error_codes,
+    )(ProviderDataIngester.get_response_json)
 
     def get_next_query_params(self, prev_query_params: dict | None) -> dict:
         if not prev_query_params:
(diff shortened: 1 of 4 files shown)


catalog/dags/providers/provider_api_scripts/freesound.py
```diff
@@ -53,6 +54,20 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+    # Freesound often tends to be flaky when making requests during ingestion.
+    # For the flaky HTTP errors above, we wait some time before trying the request
+    # again in the hopes that the upstream issues resolve in that time:
+    #   * 400 - Although framed as bad requests, these appear to succeed later
+    #   * 503 - Service unavailable, so we should just try to wait
```
dhruvkb, line 61: This comment would've been more helpful before L39.

dhruvkb (approved):
LGTM, except a nit about the position of a comment.

krysal (approved):
Nice to prune and keep updated dependencies! LGTM.

stacimc (approved):
Great idea! It would be wonderful if this eases some of the flakiness we've experienced🤞

Issue #4541, opened 2024-06-21
Add filter data step to the data refresh DAG


AetherUnbound:
## Description

In addition to the steps outlined in #4149 and #4146, we also need to add the `Filter Data` step to the data refresh DAG. Initially, this should be a single python task which exactly mirrors the behavior of the [`clean_image_data` function of `cleanup.py`](https://github.com/WordPress/openverse/blob/47fe5df0e9b8ad3dba06021f4cd4af9139977644/ingestion_server/ingestion_server/cleanup.py#L295) (only applying the tag-specific steps) on the ingestion server[[3]](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html#id6). The easiest way to do this would be to directly map the functionality of the ingestion server on this step within a single Airflow task. The steps for this task are as follows:

1.  Get a batch of records from the database using `CLEANUP_BUFFER_SIZE`
2.  Divide batch up into `multiprocessing.cpu_count()` subbatches
3.  Split the filtering up into separate workers using multiprocessing
4.  On each process
    1.  Create a new DB connection & cursor per worker
    2.  Iterate through each record
        1.  Remove tags below confidence level
        2.  Remove tags that need to be filtered (denylisted, machine-generated filter list, provider, etc)
        3.  Only surface the record if it needs to be changed
        4.  Update each records one by one with a single `UPDATE`
    3.  Commit cursor and close connection
5.  Repeat steps 1-4 until all batches are consumed

## Alternatives

_See [Alternatives in the IP](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html#filtering-approach) for possible future directions._

There are a number of ways to accomplish the data filtering, including several ways to improve the approach mentioned.

The Airflow scheduler container has access to 4 cores, which is the same as the ingestion server where this step was originally running. At present, it takes about 8 hours for all cleanup steps, but that includes the URL cleaning which is certainly more time intensive than the tag filtering since it makes outbound requests. Running the tag filtering on Airflow should not impact any of the other running tasks or saturate the instance.

There are a few ways this process could be improved, but none of them are required *at this moment*. We can follow up after this project is complete to assess what further optimizations might be necessary at this step. Some potential suggestions for that time:

[...]

PR #4667, opened 2024-07-29
Preemptively filter out Rekognition tags

### LINKED ISSUE #4644: Preemptively filter the Rekognition tags
## Description

Before inserting the Rekognition tags, we want to make sure they are appropriately filtered during the data refresh. This filtering can either be the more complete set of exclusions described above for both the labels themselves and their accuracy. This, however, depends on the completion of #4541 and the ingestion server removal project in general (#3925).

In order to work on this effort in parallel with #3925, we can add a check to the [existing tag filtering step](https://github.com/WordPress/openverse/blob/3747f9aa40ed03899becb98ecae2abf926c8875f/ingestion_server/ingestion_server/cleanup.py#L119-L150) which will exclude *all* tags with the provider `rekognition`. That way we can add all of the tags to the catalog with impunity, and allow those tags to be exposed when #3925 is finished and turned on.

## Additional context

See [this section](https://docs.openverse.org/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.html#preemptively-filter-rekognition-tags) of the IP.

AetherUnbound:
## Fixes

Fixes #4644 by @AetherUnbound

## Description

This PR adds an additional layer of filtering during the "tags cleanup" step of the (current) ingestion server which filters out tags that correspond to a specific provider. For the time being, only `rekognition` tags are filtered.

## Testing Instructions

I have been having some trouble getting `ov` to run these commands, but it might just be my desktop.

Running `just ingestion_server/test-local` should run the tests, which all should pass!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/cleanup.py
+++ b/ingestion_server/ingestion_server/cleanup.py
@@ -55,6 +55,9 @@ TAG_CONTAINS_DENYLIST = {
 # Filter out low-confidence tags, which indicate that the machine-generated tag
 # may be inaccurate.
 TAG_MIN_CONFIDENCE = 0.90
+# Filter out tags that match the following providers (either because they haven't
+# been vetted or because they are known to be low-quality).
+TAG_SKIP_PROVIDERS = {"rekognition"}
 
 # We know that flickr and wikimedia support TLS, so we can add them here
 TLS_CACHE = {
@@ -123,7 +126,7 @@ class CleanupFunctions:
     @staticmethod
     def cleanup_tags(tags):
         """
-        Delete denylisted and low-accuracy tags.
+        Filter denylisted, low-accuracy, and unverified provider tags.
 
         :return: an SQL fragment if an update is needed, ``None`` otherwise
         """
@@ -133,12 +136,14 @@ class CleanupFunctions:
         if not tags:
             return None
         for tag in tags:
-            below_threshold = False
+            alt_filtered = False
             if "accuracy" in tag and float(tag["accuracy"]) < TAG_MIN_CONFIDENCE:
-                below_threshold = True
+                alt_filtered = True
+            if "provider" in tag and tag["provider"] in TAG_SKIP_PROVIDERS:
+                alt_filtered = True
             if "name" in tag and isinstance(tag["name"], str):
                 lower_tag = tag["name"].lower()
-                should_filter = _tag_denylisted(lower_tag) or below_threshold
+                should_filter = _tag_denylisted(lower_tag) or alt_filtered
             else:
                 log.warning(f'Filtering malformed tag "{tag}" in "{tags}"')
                 should_filter = True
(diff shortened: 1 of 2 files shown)


ingestion_server/ingestion_server/cleanup.py
```diff
@@ -133,12 +136,14 @@ def cleanup_tags(tags):
         if not tags:
             return None
         for tag in tags:
-            below_threshold = False
+            alt_filtered = False
```
obulat, line 139: This code has been bothering me for a long time because it's difficult to understand all branches. What if we convert it to something like this:

```py
if not isinstance(tag.get("name"), str) or "accuracy" in tag and float(tag["accuracy"]) < TAG_MIN_CONFIDENCE or
tag.get("provider") in TAG_SKIP_PROVIDERS:
    should_filter = True
else:
    lower_tag = tag["name"].lower()
    should_filter = _tag_denylisted(lower_tag)

```
This will also be a bit better for performance because if the tag is `{}`, we won't check for it's provider and accuracy.

ingestion_server/ingestion_server/cleanup.py
```diff
@@ -55,6 +55,9 @@
 # Filter out low-confidence tags, which indicate that the machine-generated tag
 # may be inaccurate.
 TAG_MIN_CONFIDENCE = 0.90
+# Filter out tags that match the following providers (either because they haven't
+# been vetted or because they are known to be low-quality).
+TAG_SKIP_PROVIDERS = {"rekognition"}
```
obulat, line 60: The name is kind of confusing to me. Could we change it to something like 
`EXCLUDED_TAG_PROVIDERS` or `FILTERED_TAG_PROVIDERS` ?

AetherUnbound: I had the hardest time naming this 😆 I wanted `TAG` at the front to match the previous constant, but this fits much better!

ingestion_server/ingestion_server/cleanup.py
AetherUnbound: This is a good question - hopefully this code is going away long-term, but I'll try and see if I can simplify/optimize it without spending too much time on it :)

AetherUnbound: Looking at this again, we log specifically if the tag is _malformed_ based on the current set of conditions, but then don't log anything if it just gets filtered. So I think, given this will eventually be revised, I'll just leave it as-is for now.

obulat: Do we really need to log that? Malformed in this case means that a tag is either not a dictionary or doesn't have a string name value, right?
I think we've already removed those from the catalog db, haven't we?

AetherUnbound: You're right that I think we have - that said, I'm already working on #4541 and have altered the logic in these steps anyway for that change. If you have a blocking objection here than I can change it, otherwise I'd rather wait until we have to rework these cleanup steps entirely as part of the data refresh removal!

krysal (approved):
I tested faking some recognition tags in the catalog, then ran a data refresh, and they were successfully removed on the API DB 👍  Nice!

PR #4677, opened 2024-07-30
Fix waveform generation for short tracks

### LINKED ISSUE #4676: Unable to generate waveform for short audio files
## Description

Very short audio (in the order of less than a second) fails the waveform generation due to the high "pixels-per-second" required. When trying to generate it locally for the example below, I got the following output:

```
$ audiowaveform --input-filename /tmp/audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3 \
    --output-format json --pixels-per-second 22223
Input file: /tmp/audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3
Format: Audio MPEG layer III stream
Bit rate: 256000 kbit/s
CRC: no
Mode: single channel
Emphasis: no
Sample rate: 44100 Hz
Encoding delay: 1105
Padding: 341
Invalid zoom: minimum 2
```

This track's duration is only 45 milliseconds. Trying with a reduced "pixels-per-second" made the command succeed.

```
$ audiowaveform --input-filename /tmp/audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3     --output-format json --pixels-per-second 223
Input file: /tmp/audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3
Format: Audio MPEG layer III stream
Bit rate: 256000 kbit/s
CRC: no
Mode: single channel
Emphasis: no
Sample rate: 44100 Hz
Encoding delay: 1105
Padding: 341
Generating waveform data...
Samples per pixel: 197
Input channels: 1
Output channels: 1
Done: 100%
Frames decoded: 3 (0:00.078)
Generated 12 points
Output file: (stdout)
{"version":2,"channels":1,"sample_rate":44100,"samples_per_pixel":197,"bits":16,"length":12,"data":[-14225,14321,-14371,13900,-14179,14275,-13994,14017,-13787,13946,-13064,12678,-12218,10846,-10666,10624,-9127,9160,-5146,628,-11,11,-2,2]}
Done
```

## Reproduction

1. Go to http://api.openverse.org/v1/audio/e0742ff3-f909-4e1d-9555-8e6399157452/waveform/
2. See there are no peaks.

## Additional context

- Sentry Issue: [OPENVERSE-API-F3](https://openverse.sentry.io/share/issue/17855b1cf1e44e95b3b1298a59692dc2/).
- Related to #4651.

krysal:
## Fixes

Fixes #4676 by @krysal 

## Description

This reduces the number of `pixels-per-second` required for tracks with 100 milliseconds of duration or less.

## Testing Instructions

I did the following to test this:

1. Insert this track in the catalog DB (it's the same report in Sentry). Run `j catalog/pgcli` and

```sql
INSERT INTO audio (
	identifier, 
	created_on, updated_on, 
	ingestion_type, 
	provider, source, 
	foreign_identifier, foreign_landing_url, 
	url, license, license_version, 
	creator, creator_url,
	duration,
	removed_from_source
) VALUES (
	'e0742ff3-f909-4e1d-9555-8e6399157452',
	NOW(),
	NOW(),
	'provider_api',
	'freesound',
	'freesound',
	'196927',
	'https://freesound.org/people/AlienXXX/sounds/196927',
	'https://cdn.freesound.org/previews/196/196927_97763-hq.mp3',
	'by',
	'4.0',
	'AlienXXX',
	'https://freesound.org/people/AlienXXX',
	45,
	false
);
```

2. Run the audio data refresh locally. `just a` & start the DAG from the Airflow UI.
3. Observe the peaks at the API endpoint for the track: http://localhost:50280/v1/audio/e0742ff3-f909-4e1d-9555-8e6399157452/waveform/

## Checklist

- [n/a] I added or updated documentation (if applicable).
- [n/a] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog PRs) or the media properties generator (`ov just catalog/generate-docs media-props` for the catalog or `ov just api/generate-docs` for the API) where applicable.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
Could you add a test for this change? Also, there are a couple of other places where we have both `logger.error` and `sentry_sdk.capture_exception`.

Do you think we don't need to have the waveform logs on the server? It would be much better to setup Sentry with the structlog so that the `logger.error` sent the traceback to Sentry. But I don't know how to do that :) Maybe adding `[attach_stacktrace](https://docs.sentry.io/platforms/python/configuration/options/#attach-stacktrace)` to the Sentry setup would help?

sarayourfriend:
@krysal the second issue you linked is just this PR. Did you mean to link something else?

Can you clarify what you checked when you looked at duplicate issues in Sentry compared to the logger error? Does calling `capture_exception` also produce a log line, or is the issue that `logger.error` is creating multiple Sentry issues (and only a single log line)?

We definitely should _not_ remove the log unless for some reason `capture_exception` also produces a log-line, which I can't find any evidence to that end (neither in the code nor Sentry docs).

Olga's suggestion makes more sense, going the other way around, integrating Sentry into our logger, and then relying on `logger.error` (etc) to capture an exception event rather than explicitly calling `capture_exception`.

The issue is that the logs must show a full record of what happened in the event of an error, specifically with an error line in place to record the error. Otherwise it becomes very difficult to debug errors in production and context is potentially lost.

Can you split this work into separate PRs, instead? Solving the short waveform seems straightforward based on the solution you've come up with. The sentry/logger integration, on the other hand, is a separate problem that requires more careful changes and ensuring the integration with Sentry, structlog, and so forth is cohesive, and that the API codebase as a whole reflects the correct usage of those tools.

api/api/utils/waveform.py
```diff
@@ -102,7 +102,8 @@ def generate_waveform(file_name, duration):
 
     logger.debug("waveform_generation_started")
 
-    pps = math.ceil(1e6 / duration)  # approx 1000 points in total
+    width = 1e6 if duration > 100 else 1e5
```
zackkrida, line 105:
```suggestion
    # Determine the width of the waveform based on the duration of the audio.
    # The width varies to improve the appearance and "resolution" of the waveform.
    # For long audio files, we set the width to 1,000,000 pixels.
    #
    # For short audio files, we set the width to 100,000 pixels.
    # This prevents the waveform from appearing "stretched out" and sparse.
    width = 1e6 if duration > 100 else 1e5
```

Can we add a comment like this?

zackkrida (approved):
LGTM! Just a suggestion to explain the width conditional for our future selves :smile:

api/api/utils/waveform.py
krysal: Added. Thanks!

api/api/utils/waveform.py
```diff
@@ -102,6 +102,12 @@ def generate_waveform(file_name: str, duration: int):
 
     logger.debug("waveform_generation_started")
 
+    # Determine the width of the waveform based on the duration of the audio.
+    # The width varies to improve the appearance and "resolution" of the waveform.
+    # For long audio files, we set the width to 1,000,000 pixels.
+    #
+    # For short audio files, we set the width to 100,000 pixels.
+    # This prevents the waveform from appearing "stretched out" and sparse.
```
obulat, line 110: Maybe also add something like "if there are too few data points per pixel, audiowaveform throws an errror, see issue #"?

obulat (approved):
Nice to have this fixed! I have a small non-blocking doc suggestion.

api/api/utils/waveform.py
krysal: Added.

Issue #3336, opened 2023-11-10
Implementation Plan: Use a single index for all API results


AetherUnbound:
## Description

After consulting with some Elasticsearch experts, we've determined that we may be able to improve Elasticsearch performance by using a _single_ index with a `sensitive` keyword rather than two indices (filtered & unfiltered). The existence of two indices that are being regularly queried means that Elasticsearch must be loading records off disk from both indices much more frequently than it would do so on its own. The difference in disk ops between querying two indices and only one index can be seen in this graph:

![image](https://github.com/WordPress/openverse-infrastructure/assets/10214785/df9c43d2-1eb5-4d64-9157-b647709f5d87)

The easiest way to accomplish this might be an extra step during the initial indexing when building the document to feed into Elasticsearch which uses the sensitive terms list to determine a `sensitive` keyword field on the document.

It might also be possible to build and index the documents in a first step and default to `sensitive=False`, then process all the documents in a second pass once the index is built. This could query the documents matching the sensitive terms list and update the `sensitive` field for those documents (since we've seen they're about 1% of the total index size). We could use the [bulk update API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) for this, although we've been discouraged from doing so since that would be a very long-running request against the cluster.

## Alternatives

We can leave the cluster as-is with the two index setup; it has not been shown to affect the periodic response time spike we've been seeing recently.

## Additional context

The original outline for the two index approach can be found here: https://docs.openverse.org/projects/proposals/trust_and_safety/detecting_sensitive_textual_content/20230330-implementation_plan_filtering_and_designating_results_with_sensitive_textual_content.html

sarayourfriend:
@AetherUnbound I was wondering if you had any thoughts about when we might prioritise this? I think it would have benefits for service reliability, particularly with index redundancy, if we had fewer indexes spread across the data nodes (we could increase the number of replicas, for example, so that no single node carries a significant amount of any one index).

I also wanted to record some thoughts I had today about how this could work. In particular, regarding the new `sensitive` field on the index. I'm not sure if you used "keyword" intentionally, Madison, but it would be great to have it actually be a keyword list field, mirroring the "sensitivity_reasons" that we return from the API. So if the work is sensitive due to a confirmed use report, we would add `user_reported` to the `sensitive` keyword list. If the work has sensitive text, detected during ingestion, we would add `sensitive_text` to the `sensitive` keyword list on the work.

[...]

PR #4684, opened 2024-07-31
Add alter data step to the data refresh DAG 

### LINKED ISSUE #4541: Add filter data step to the data refresh DAG

AetherUnbound:
## Fixes

Fixes #4541 by @AetherUnbound

## Description

This PR adds a data filtering step (named `alter_data`) to the data refresh DAG. These steps are largely based on those of the `cleanup.py` file from the ingestion server.

However, unlike what's stated in #4541, the code is **not** copied over directly. I initially tried to do this to reduce the amount of work necessary to set this step up, but I ran into https://github.com/apache/airflow/issues/14896 and was unable to leverage the `multiprocessing` library within the context of a single Airflow task. I opted to take an approach similar to #4572 (indeed, producing a lot of similar code!) and instead use Airflow's mapped tasks to split the data across a number of workers. This actually simplified the code _significantly_ (along with removing the other cleanup steps and the generic field handling). Turns out: using Airflow for the thing Airflow was intended for makes things easier! Yay!

All that to say, this adds a task group to the data refresh after the `copy_data` step and before the distributed reindexing which filters out any tags that meet the following criteria:
- They're malformed
- They match or are included in our denylists
- They are below a certain accuracy
- They are from an unvetted provider

> [!IMPORTANT]
> This is a divergence from current ingestion server behavior because we will **also be filtering data for the `audio` table now**. Previous filtering was done only on the image table, so now all our data refresh DAGs will receive the same filtering (which is good!).

image

## Testing Instructions

> [!WARNING]
> A new variable, `AIRFLOW_VAR_DATA_REFRESH_ALTER_BATCH_SIZE=1000`, has been added to the `env.template` file. Be sure to add this to your own `catalog/.env` file before testing!

1. Run `ov just c && ov just i` to bring the catalog stack up (along with the downstream API database)
2. Enable the `production_audio_data_refresh` DAG (this should kick off a run automatically)
3. Observe that the `alter_table_data` block runs and succeeds
4. If you've set the above variable appropriately, you should see **5** mapped tasks for `alter_data_batch`
5. Click on mapped index **3** and view the logs, you should see `1` returned as the XCom and a note in the logs that 1 row was affected
6. (Optional) Remove the aforementioned Variable from your environment and run steps 1-5 again. See that only 1 mapped task is created, and the default batch size (100,000) is used instead of 1,000

Tests should also pass!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_refresh/alter_data.py
+++ b/catalog/dags/data_refresh/alter_data.py
@@ -1,9 +1,9 @@
 """
-# Filter Data Task
+# Alter Data Task
 
-This module contains the Airflow task used for filtering data from the
-newly created temporary table in the downstream (API) database. This
-provides an entrypoint for filtering tags that do not meet a certain
+This module contains the Airflow tasks used for filtering, transforming, and augmenting
+data from the newly created temporary table in the downstream (API) database.
+Currently, this provides an entrypoint for filtering tags that do not meet a certain
 minimum accuracy, are malformed, or are from unvetted providers.
 """
 
@@ -116,7 +116,7 @@ def generate_tag_updates(tags) -> list[dict] | None:
 
 
 @task
-def get_filter_batches(
+def get_alter_batches(
     id_bounds: tuple[int, int], batch_size: int | None
 ) -> list[tuple[int, int]]:
     start, stop = id_bounds
@@ -125,14 +125,14 @@ def get_filter_batches(
 
 
 @task
-def filter_data_batch(
+def alter_data_batch(
     batch: tuple[int, int],
     temp_table: str,
     postgres_conn_id: str,
     task: AbstractOperator = None,
     timeout: float = None,
 ) -> int:
-    logger.info(f"Starting data filtering on batch: {batch}")
+    logger.info(f"Starting data altering on batch: {batch}")
     batch_start, batch_end = batch
     postgres = PostgresHook(
         postgres_conn_id=postgres_conn_id,
@@ -147,17 +147,17 @@ def filter_data_batch(
         handler=fetch_all_handler,
     )
     worker_conn = postgres.get_conn()
-    logger.info("Data filtering worker connected to database")
+    logger.info("Data altering worker connected to database")
     with worker_conn.cursor() as write_cur:
         logger.info(f"Filtering {len(rows)} rows")
 
-        filtered_count = 0
+        altered_count = 0
         for row in rows:
             _id, identifier, tags = row
             tags_fragment = generate_tag_updates(tags)
             if not tags_fragment:
                 continue
-            filtered_count += 1
+            altered_count += 1
             logger.debug(
                 f"Updated tags for {identifier}\n\t"
                 f"from '{tags}' \n\tto '{tags_fragment}'"
@@ -170,7 +170,7 @@ def filter_data_batch(
     logger.info("Worker committing changes...")
     worker_conn.commit()
     worker_conn.close()
-    return filtered_count
+    return altered_count
 
 
 @task
@@ -180,12 +180,12 @@ def report(counts: list[int]):
     return total_count
 
 
-@task_group(group_id="filter_table_data")
-def filter_table_data(
+@task_group(group_id="alter_table_data")
+def alter_table_data(
     environment: Environment,
     data_refresh_config: DataRefreshConfig,
 ):
-    """Perform data filtering across a number of tasks."""
+    """Perform data altering across a number of tasks."""
     postgres_conn_id = POSTGRES_API_CONN_IDS.get(environment)
     temp_table = data_refresh_config.table_mappings[0].temp_table_name
 
@@ -197,13 +197,13 @@ def filter_table_data(
         return_last=True,
     )
 
-    batches = get_filter_batches(
+    batches = get_alter_batches(
         estimated_record_count.output,
-        batch_size="{{ var.value.get('DATA_REFRESH_FILTER_BATCH_SIZE', none) }}",
+        batch_size="{{ var.value.get('DATA_REFRESH_ALTER_BATCH_SIZE', none) }}",
     )
 
-    filter_data = filter_data_batch.partial(
+    alter_data = alter_data_batch.partial(
         temp_table=temp_table, postgres_conn_id=postgres_conn_id
     ).expand(batch=batches)
 
-    report(filter_data)
+    report(alter_data)
--- a/catalog/dags/data_refresh/dag_factory.py
+++ b/catalog/dags/data_refresh/dag_factory.py
@@ -44,6 +44,7 @@ from common.sensors.single_run_external_dags_sensor import SingleRunExternalDAGs
 from common.sensors.utils import wait_for_external_dags_with_tag
 from data_refresh.copy_data import copy_upstream_tables
 from data_refresh.data_refresh_types import DATA_REFRESH_CONFIGS, DataRefreshConfig
+from data_refresh.filter_data import filter_table_data
 from data_refresh.reporting import report_record_difference
 
 
@@ -147,6 +148,10 @@ def create_data_refresh_dag(
 
         # TODO Cleaning steps
 
+        filter_data = filter_table_data(
+            environment=environment, data_refresh_config=data_refresh_config
+        )
+
         # Disable Cloudwatch alarms that are noisy during the reindexing steps of a
         # data refresh.
         disable_alarms = PythonOperator(
@@ -198,7 +203,13 @@ def create_data_refresh_dag(
         )
 
         # Set up task dependencies
-        before_record_count >> wait_for_dags >> copy_data >> disable_alarms
+        (
+            before_record_count
+            >> wait_for_dags
+            >> copy_data
+            >> filter_data
+            >> disable_alarms
+        )
         # TODO: this will include reindex/etc once added
         disable_alarms >> [enable_alarms, after_record_count]
         after_record_count >> report_counts
--- a/catalog/dags/data_refresh/dag_factory.py
+++ b/catalog/dags/data_refresh/dag_factory.py
@@ -116,6 +116,7 @@ def create_data_refresh_dag(
         default_args=default_args,
         start_date=data_refresh_config.start_date,
         schedule=data_refresh_config.schedule,
+        render_template_as_native_obj=True,
         max_active_runs=1,
         catchup=False,
         doc_md=__doc__,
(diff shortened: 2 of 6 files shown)


krysal (commented):
This looks good to me! Very nice that the translation with dynamic tasks is so straightforward. A little issue I see is that for some reason, the unvetted provider cleaning didn't work for me. It's strange because in the catalog, I changed the provider of some tags to `recognition`, and in the API DB, after the data refresh they still appear with the old provider. Does this happen to any of you?

sarayourfriend:
> It's strange because in the catalog, I changed the provider of some tags to `recognition` [...]

@krysal just a heads up that the provider is `rekognition`, with a `k` rather than a `c`. Just in case that may be the issue, if you used it as typed in your comment ([AWS Rekognition](https://aws.amazon.com/rekognition/))

sarayourfriend (commented):
Cool!

I won't review carefully the code, but I have a question about the long term intent of this DAG, and how it would fit into another part of the data refresh overall ETL pipeline (the DB -> ES part).

Will this "filter data" step be _the_ place for catalog DB to API DB transformations of data during the data refresh? If that is the case, would a name more generic than "filter" be appropriate? What happens if we want to try an augmentation of data without committing it directly to the catalog media tables? Or if we want to join data into the API DB tables for presentation from other tables or data sources? I don't think we currently have the capabilities for it, but if you or @stacimc have a sense about whether that kind of thing would happen here or as a separate DAG (and why, either way) I'd be keen.

The reason I am asking is because from a DB to ES perspective, the #3336 work will need to at some point augment the data we have now. It's a good example where we want the flexibility of having the transformation happen every time during the data refresh so that we can more easily make changes in preparation of the next data refresh, if we need to adjust the terms list and such. But that's just one approach, and it might not make sense to happen at the catalog DB -> API DB level, it might make more sense if it's at the API DB -> ES stage.

On the other hand, there's no discounting that it might be better and perfectly flexible enough for our needs to augment the media tables with this information as part of an on-demand batched update similar to the popularity data, if that's a reliable way to do it. It might also be possible to use ES's [document enrichment step](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-enriching-data.html) rather than have it part of the data refresh at all (but whether that approach fits with the requirements I'm not sure, so the options need to be somewhat open).

Anyway, to summarise, essentially a single question in this vein:

- Is the "filter data" DAG meant specifically only for filtering data? If so, where would we enrich data going into the API DB during the data refresh, if not at this stage? If not, then should the name of the DAG change to be more broad, to reflect its possibilities, and avoid e.g., an entirely new step added that could go alongisde filter, rather than having a separate independent iteration through the data?

Beyond that I have an technical/architectural question about the approach taken in this DAG. If I understand it correctly, it happens _after_ the data is loaded into the API DB? The task executes after `copy_data`, according to the PR description and my understanding of the DAG code. Was the possibility of filtering at the insertion step, rather than as an update after the fact, considered? All of these tag filters could be implemented as SQL in the `copy_data_query`.

I can understand this approach is convenient moving from the previous implementation, but I wanted to ask if there are specific intentions for it long term to remain as a Python task that iterates through each row of the new table, or if it could make sense to move some of these essentially copy-time filter steps to `copy_data`?

AetherUnbound:
Thanks for your thoughts and questions @sarayourfriend! I appreciate the broader way you're thinking about this piece of the refresh process and where it could go from here. While the intent was to move this with an exact implementation of what's on the ingestion server currently, Airflow's own internals required a bit of adjustment which has made this step appear more generic.

I certainly think it could be used down the line for the kinds of data operations you describe, namely augmentation & other filtering. However since this particular PR is part of the ingestion server removal project, I'm hesitant to try and change the scope _on this particular PR_ right now. I think I'd want to explore the performance implications of any transformations we do out into the `copy_data` step rather than as a separate filtering/augmentation step later in the DAG. There's also an aspect of maintainability there - we have few experts in SQL on the team and if we make these operations purely in SQL it may be harder to maintain or adjust down the line (whereas almost all of the team is familiar with performing this kind of filtering/augmentation in Python). That to say that I'm hesitant to make that sort of change _at this moment_, especially given we can always change that down the line as long as the effect on the data is the same.

On the naming piece though, I could see us adding more steps to filter or augment the data right at this step. Let me see if I can come up with a more generic term that still communicates the intent while allowing room to grow in the future.

AetherUnbound:
Thought on it more and I think "alter" is the way to go! It captures filtering, transforming, and augmenting, and is generic enough for us. Going to make that change today.

catalog/dags/data_refresh/alter_data.py
```diff
@@ -0,0 +1,209 @@
...
+        task_id="get_estimated_record_count",
+        conn_id=postgres_conn_id,
+        sql=QUERY_ROW_ESTIMATE.format(temp_table=temp_table),
+        handler=fetch_one_handler,
+        return_last=True,
+    )
+
+    batches = get_alter_batches(
+        estimated_record_count.output,
+        batch_size="{{ var.value.get('DATA_REFRESH_ALTER_BATCH_SIZE', none) }}",
```
stacimc, line 202: Shame it's tricky to default this, but I think the way you've done it is the best option 🤔

catalog/dags/data_refresh/alter_data.py
```diff
@@ -0,0 +1,209 @@
...
+    "pdm",
+}
+
+# Filter out low-confidence tags, which indicate that the machine-generated tag
+# may be inaccurate.
+TAG_MIN_CONFIDENCE = 0.90
+# Filter out tags that match the following providers (either because they haven't
+# been vetted or because they are known to be low-quality).
+FILTERED_TAG_PROVIDERS = {"rekognition"}
+# Select the data within a given batch
```
stacimc, line 59: Do you think there'd be any utility in having these come from Airflow variables? My feeling is that if we wanted to adjust these it would probably be worth a PR and approval process anyway, but I thought I'd throw the question out there. Any reason we might want to be able to filter out different tag providers more quickly, for example?

stacimc (approved):
Looks great to me! Very clean this way as well :) +1 for `alter_data`, and any other changes we want to make in the future would be easy to do; I don't feel this locks us into anything unnecessarily. I also agree that I'd prefer to keep the simplest, minimal-change approach _for now_, while there's a lot happening with the data refresh already.

I was seeing some weird results for the batch ids (like ranging from 5,000 - 10,000) but I didn't have the latest commits pulled, and now I can't reproduce it so I assume my recent environment issues were at fault :)

sarayourfriend:
Sounds good, thanks for taking the time on my questions! I like the new name as well, and definitely agree to keep the scope as is for this PR; mostly I was wanting to understand the intention for the process in the future... but it's probably misguided to think _too_ much about that now or treat any current understanding of any particular step as the way we'd keep things, especially with the intention we have of talking about the entire Openverse "data story" and understanding better where we are now (and why), and where we would like to be (and why), and some paths we might take to get there. Which is to say, the step is important, but talking about it in isolation like I was in my question only has so much utility. It'd be good to think about these concepts (filtering, augmentation, transformation) in the context of the whole concept of the data refresh and its actual goals, rather than the current implementation of that process, the architecture of which we inherited and some aspects of which do not have concrete explanations.

Anyway, that's just to say, the thinking right now makes sense to me, clarifies the question I understand that _can_ be clarified right now, and the new name does a good job doing that to at least communicate intent for near term needs even if ongoing discussions may yield greater insights into the process as a whole.

Thanks again!

catalog/dags/data_refresh/alter_data.py
AetherUnbound: Yea...I agree 😞 This seemed like the best way to make sure that it was 1) only filled in by Airflow at runtime and 2) easiest for testing different batch sizes.

catalog/dags/data_refresh/alter_data.py
AetherUnbound: Since there's _so much_ heavy lifting with the data refresh process (and probably as you mention, the act of having to make a PR for the change will serve as documentation in some form for the change itself), I think it's best not to have this be configured by a Variable. It's a good thing to be thinking about! And if we find we're changing this sort of thing frequently (_after_ the Rekognition project, at least haha), then we can add that functionality 😊

PR #4508, opened 2024-06-18
Log dead link verification request timings

sarayourfriend:
## Fixes

Part of https://github.com/WordPress/openverse/issues/4507 by @sarayourfriend 

## Description

See the issue for rationale of both the `get_aiohttp_session` change and the new timing logs.

The PR does both, using `time.time` to calculate the request timing.

## Testing Instructions

Run the API locally with `./ov just api/up` (and `./ov just api/init` if you don't have data). Tail the logs with `./ov just logs web` and make an image request with `filter_dead=True` (this matches the production default): http://localhost:50280/v1/images/?filter_dead=True

In your logs, you should see a line like this for each verified link:

```
web-1  | 2024-06-18T03:30:03.753846Z [info     ] dead_link_validation_timing    [api.utils.check_dead_links] filename=__init__.py func_name=_head ip=172.18.0.1 lineno=59 request_id=50cdc9a5-cf63-4f1e-9df6-366ce3cce800 status=200 time=0.45412659645080566 url=https://live.staticflickr.com/65535/51747932079_de3b535418_b.jpg
```

Note that if you make the request again, the verification will be cached, and you will not see it on subsequent requests (this is just how dead link verification works). That also means that if you've made a verification request recently, you'll need to either choose a query that pulls records that you haven't verified yet, or run `./ov just exec cache redis-cli FLUSHALL` to wipe Redis clean.

Do the same on main and notice that instead of a timing log, you have a `get_aiohttp_session` log for each verified link. That's fixed in this PR too (see issue for rationale).

I did not add unit tests because I didn't think it made sense (far more complex to do that, I think), but existing tests will confirm that the original functionality hasn't changed, and a manual test will confirm the new logs are correct. This code will be removed as soon as we decide what to reduce the timeout to.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [N/A] I added or updated tests for the changes I made (if applicable).
- [N/A] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (`just catalog/generate-docs` for catalog
      PRs) or the media properties generator (`just catalog/generate-docs media-props`
      for the catalog or `just api/generate-docs` for the API) where applicable.

[best_practices]:
[...]

PR #4695, opened 2024-08-01
Decrease thumbnail upstream timeout and add timing logs for thumbnail requests

### LINKED ISSUE #4570: Investigate and decrease image proxy upstream request timeout
## Problem

Image proxy requests time out at 15 seconds. Similar to #4507, thumbnail requests are highly susceptible to upstream provider problems. If Flickr has a bad day, we'll have a bad day because we can't generate thumbnails (or validate dead links, to tie it back to that example).

## Description

We should perform a similar analysis as I did for #4507. Add logging to the thumbnail requests to time how long they take so we can analyse them. Log the time, status, url, and maybe go ahead and log the `provider` slug specifically so we can more easily segment the analysis along those lines (I wish we had done so for the dead link validation).

Deploy the logs for a week and follow the same approach to analysing the times as I took for the link validation. Find out how long (average, p95, p99) it takes for successful thumbnails to come back. Find out if there are meaningful differences between providers that we could modulate the timeout based on. etc etc.

The goal is to figure out if we can lower the 15-second timeout without creating significant negative consequences. Some failed thumbnails must be accepted, but something to consider is whether if a request times out (specifically times out) if we could try increasing the timeout the next time that thumbnail gets requested, and do so progressively until we hit some maximum (whether that's 15 seconds or otherwise).

## Additional context

This is mostly a way to investigate whether thumbnail timeouts really need to be so high. If they do, that's fine, but it'd be good to know.

It is not, however, likely to be an alternative to segmenting our API request timing monitors, as thumbnail requests will probably always take a little longer than our other requests, and along with search, will always be the most important to know without having the speedy outliers or volume of one or the other affect the visibility of problems with the other.

sarayourfriend:
## Fixes

Fixes #4570 by @sarayourfriend

## Description

This PR decreases the thumbnail upstream request timeout as discussed in the linked issue. It also decreases the timeout for the HEAD request for the extension to the same default value. Finally, it adds timing logs for both these requests so we can further evaluate appropriate timeouts for each. For example, hopefully we can decrease the HEAD request timeout significantly (like 0.5 seconds at least). But, because the analysis in the linked issue was done with timings of the entire thumbnail request, it's not possible to know which part of the request actually takes 4 seconds. Therefore, the total timeout of outbound requests is now 8 seconds, which is still way lower than the 25 total seconds and still a good improvement in line with the goals of the issue.

To facilitate request timing in any circumstance we might want it, I've created an [aiohttp `TraceConfig` to log timing with context variables we've used so far](https://docs.aiohttp.org/en/stable/client_advanced.html#client-tracing). This only affects requests that configure themselves to activate the timing logs, as verified in the unit tests for the new trace config. This generalises the functionality added in #4508 for dead link checks, and therefore is able to replace that original implementation. As such, you'll see the code to manually time those requests is gone now, replaced by the trace config version. Because of that, I've also removed the separate timeout status we were caching before. This is safe for existing data because we check for specific _live_ status codes when using the cached data, and never actually referred to the difference between the timeout and general error statuses returned from `_head` except when inserting them into the cache.

## Testing Instructions

All existing unit tests must continue to pass without modification.

Run the API locally with `ov just api/up`. Open the logs with `ov just logs web` and make thumbnail requests and see the timing logs are present. Follow the testing instructions from #4508 and confirm they still work. The logs produced in this PR should be identical to the ones described in the linked PR, including the same contextual information, as well as `provider`, which was added later in #4598.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/compose.yml
+++ b/api/compose.yml
@@ -53,3 +53,12 @@ services:
       DJANGO_NGINX_UPSTREAM_URL: web:50280
     depends_on:
       - web
+
+  httpbin:
+    profiles:
+      - api
+    # Use go-httpbin because it's only 15 MB compared to upstream HTTPBin's 204 MB
+    # That helps reduce the network burden of spinning up a development environment
+    image: docker.io/mccutchen/go-httpbin:v2.14.0
+    expose:
+      - "8080"
--- a/api/compose.yml
+++ b/api/compose.yml
@@ -54,6 +54,7 @@ services:
     depends_on:
       - web
 
+  # Used in API unit tests
   httpbin:
     profiles:
       - api
--- a/api/api/utils/aiohttp.py
+++ b/api/api/utils/aiohttp.py
@@ -81,11 +81,15 @@ async def get_aiohttp_session() -> aiohttp.ClientSession:
 
 
 class LogTiming(aiohttp.TraceConfig):
+    TIMEOUT_STATUS = -2
+    ERROR_STATUS = -1
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         self.on_request_start.append(self._start_timing)
-        self.on_request_end.append(self._end_timing)
+        self.on_request_end.append(self._log_timing)
+        self.on_request_exception.append(self._log_timing)
 
     async def _start_timing(
         self,
@@ -95,12 +99,14 @@ class LogTiming(aiohttp.TraceConfig):
     ):
         trace_config_ctx.start_time = time.perf_counter()
 
-    async def _end_timing(
+    async def _log_timing(
         self,
         session: aiohttp.ClientSession,
         trace_config_ctx,
-        params: aiohttp.TraceRequestEndParams,
+        params: aiohttp.TraceRequestEndParams | aiohttp.TraceRequestExceptionParams,
     ):
+        """Log timing at end of request or when request results in an exception."""
+
         if not (
             trace_config_ctx.trace_request_ctx
             and "timing_event_name" in trace_config_ctx.trace_request_ctx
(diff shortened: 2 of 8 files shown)


api/compose.yml
```diff
@@ -53,3 +53,12 @@ services:
...
       - web
+
+  httpbin:
+    profiles:
+      - api
+    # Use go-httpbin because it's only 15 MB compared to upstream HTTPBin's 204 MB
+    # That helps reduce the network burden of spinning up a development environment
+    image: docker.io/mccutchen/go-httpbin:v2.14.0
+    expose:
+      - "8080"
```
sarayourfriend, line 64: I tried a few approaches to testing the trace config, and this was the one that finally worked. Here's what I tried and why they didn't work:

1. I tried pook first, of course. However, the way pook hooks into aiohttp stops it from executing its trace config altogether, so it was useless for testing this.
2. I tried using pytest-httpbin, which runs an httpbin Python server during the pytest suite run, and provides some useful fixtures. However, httpbin itself is very badly out of date and incompatible with other dependencies of the API: https://github.com/postmanlabs/httpbin/issues/712
3. Finally, just running httpbin inside the compose stack, which is what I've done here.

I'm open to suggestions for other approaches that will work and avoid adding this dependency into the stack.

api/compose.yml
```diff
@@ -53,3 +53,12 @@ services:
       DJANGO_NGINX_UPSTREAM_URL: web:50280
     depends_on:
       - web
+
+  httpbin:
```
sarayourfriend, line 57:
```suggestion
  # Used in API unit tests
  httpbin:
```

obulat (approved):
Tested well locally. The thumbnail request time drops sharply when the thumbnail is ready in Photon: from ~1.5 to 0.3.
Is there a plan for monitoring these logs and actions we should take based on the results?

sarayourfriend:
I thought I had created an issue to investigate the extension timings, but I'll do so after merging, and include looking at both timings, not just the extension timings.

PR #4697, opened 2024-08-01
Bump the minimum required PDM version for the API

AetherUnbound:
## Description

I was seeing an issue on `main` where when I tried to run `ov just build`, I got the following error:

```
------
 > [web builder 5/5] RUN pdm install --check --frozen-lockfile --dev:
1.475 WARNING: Lockfile is not compatible with PDM
1.476 Please run `pdm lock` to update the lock file
------
failed to solve: process "/bin/sh -c pdm install --check --frozen-lockfile $PDM_INSTALL_ARGS" did not complete successfully: exit code: 1
error: Recipe `dc` failed on line 209 with exit code 17
error: Recipe `build` failed on line 213 with exit code 17
```

@dhruvkb pointed out that this version might need to be bumped, and indeed it helped resolve the issue on my end!

## Testing Instructions

Check out this branch and run `ov just build`, ensure that everything works as expected.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -43,7 +43,7 @@ RUN apt-get update \
     python3-dev \
   && apt-get autoremove -y \
   && rm -rf /var/lib/apt/lists/* \
-  && pip install pdm~=2.14
+  && pip install pdm~=2.17
 
 # Copy subpackages from additional build-context 'packages'
 # hadolint ignore=DL3022


dhruvkb (approved):
Since the `lock_version` got bumped to the latest version, I suspect Renovate uses the latest version of PDM. So I suppose we could also drop the version and let `pip` install the latest one automatically?

AetherUnbound:
> Since the `lock_version` got bumped to the latest version, I suspect Renovate uses the latest version of PDM. So I suppose we could also drop the version and let `pip` install the latest one automatically?

Only trouble with that approach is that, because the line doesn't change in the `Dockerfile`, I'm not sure when Docker would necessarily know it needs to re-run the install of PDM (and thus resolve issues like this which occur in the future). Hopefully the structure of the lockfile doesn't change too frequently.

zackkrida (approved):

sarayourfriend:
I agree with Dhruv, dropping the version might be the way to go. I think that `apt-get ...` line runs every single time I run `just build web`, but we could copy the lockfile in before it if we wanted to make sure it definitely ran anytime the lockfile changed.

Not sure what PDM's backwards compatibility approach is though. Pinning to a major version is probably a good idea...

AetherUnbound:
> I think that `apt-get ...` line runs every single time I run `just build web`

Those layers should almost certainly be cached, it seems odd that would be happening every time 🤔 That said, I'm going to go ahead and merge this, but the idea of moving the `pdm.lock` copy step to before the install could work. We should probably move the `pip install pdm` line to a different `RUN` layer so the `apt-get install` line can be cached more frequently, since the lock file can change quite often and it might be a nuisance to reinstall _all_ dependencies even if only the python ones have changed.

Issue #432, opened 2023-02-18
Masking sensitive words in Sentry


obulat:
| Start Date | ETA        | Project Lead | Actual Ship Date |
| ---------- | ---------- | ------------ | ---------------- |
| 2023-11-01 | 2023-11-30 | TBD          | TBD              |

## Description

Openverse uses Sentry to track runtime errors in the API and the frontend. To make the Sentry issues safer for the reviewer, we need to add Sentry scrubbing in the API and Frontend against our sensitive term list. 

### Prerequisites
- Determining a list of sensitive terms
- #377
## Documents

- [ ] Project Proposal
- [ ] Implementation Plan

sarayourfriend:
Has anyone confirmed this is realistically possible without, for example, sending the entire sensitive terms list to the frontend (or bundling it into the source code)? I can't find anything in Sentry's documentation that makes it seem like we could do this on the _Sentry_ side of things (https://docs.sentry.io/product/data-management-settings/scrubbing/) so I think we'd have to do it before we send the event to Sentry itself. That could be prohibitively expensive/potentially impossible to do without hacking deep into Sentry's inner workings. Sentry assumes you'd want to scrub specific _fields_ rather than terms-in-fields, so there isn't a facility for running everything through a filter, as far as I can tell. There's a non-trivial performance question there too. If every Sentry event needs to run through a huge list of sensitive terms (or a massive regex statement), there's a cost we need to consider to that.

AetherUnbound:
I looks like we should have some capacity to do this on the client side, for both Javascript and Python:
- [Python scrubbing instructions](https://docs.sentry.io/platforms/python/data-management/sensitive-data/?original_referrer=https%3A%2F%2Fgithub.com%2FWordPress%2Fopenverse%2Fissues%2F432#scrubbing-data)
- [Python scrubbing internals for Sentry](https://github.com/getsentry/sentry-python/blob/1.18.0/sentry_sdk/scrubber.py)
- [Vue scrubbing instructions](https://docs.sentry.io/platforms/javascript/guides/vue/data-management/sensitive-data/?original_referrer=https%3A%2F%2Fgithub.com%2FWordPress%2Fopenverse%2Fissues%2F432)

Sentry provides the entire event in both those cases, so we should be able to modify any aspect of it (including text within fields). That said though, that is a fair amount of processing that'd happen on our services _before_ the event is even sent, which leaves more room for failure IMO. I'm going to look into Sentry server-side processing next.

[...]

PR #4704, opened 2024-08-02
Setup structlog Sentry integration

### LINKED ISSUE #4651: Waveform issues are reported to Sentry twice
## Description

We call both `logger.error` and `sentry_sdk.capture_exception` for all errors during waveform generation:

https://github.com/WordPress/openverse/blob/7cfb689b20e08d4dfc2b2ac31051ea21b558184b/api/api/utils/waveform.py#L120-L121

Both of these calls send the error to Sentry:
Screenshot 2024-07-23 at 12 46 11 PM

The `logger.error` shows where the error was generated in the code:

```
{'file_name': 'audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3', 'e': CalledProcessError(1, ['audiowaveform', '--input-filename', 'audio-e0742ff3-f909-4e1d-9555-8e6399157452.mp3', '--output-format', 'json', '--pixels-per-second', '22223']), 'event': 'waveform_generation_failed', 'timestamp': '2024-07-22T05:49:08.862359Z', 'request_id': '', 'ip': 'xx.xx.xx.xx', 'logger': 'api.utils.waveform', 'level': 'error', 'lineno': 120, 'func_name': 'generate_waveform', 'filename': 'waveform.py'}
```

However, it's easier to understand and handle the stack trace from the `sentry_sdk.capture_exception` call in Sentry UI:

Screenshot 2024-07-23 at 12 47 05 PM

## Additional context

To solve this, we can drop the `capture_exception` call. In this case, it would be nice to make the `logger.error` calls show the stack trace in the Sentry UI.
Another solution would be to add a child logger to this module that is ignored by Sentry, and only use the `capture_exception` results.

The changes were introduced in #4444.

obulat:
## Fixes

Fixes #4651 by @obulat

## Description

This PR sets up the `structlog-sentry` library suggested by @sarayourfriend to fix our structlog-Sentry integration. 

With this, all `logger.error` calls should also be sent to Sentry automatically, and we no longer need to call `sentry_sdk.capture_exception` in addition to logging the error to get the error stack trace in Sentry.

By default, the error was being sent twice: as an error, and as a JSON log. This PR added an integration during Sentry setup to prevent JSON logs from being sent to Sentry[^1] as described in the [docs](https://github.com/kiwicom/structlog-sentry?tab=readme-ov-file#logging-as-json)

This is a sample error with added stack trace: https://openverse.sentry.io/share/issue/f21fdb1d1067424eafd648dea13d4b77/

[^1]: [sample Sentry error as a JSON log](https://openverse.sentry.io/share/issue/e2bc26efc51048a981a0da629a34930d/)
## Testing Instructions

- Follow @krysal's the instructions in #4677 to add a very short audio file locally.
- Set the correct Sentry DSN in `api/.env` file
- Add the following patch to stop saving the waveform and always produce the error locally:

<details>
<summary>Patch</summary>

```diff
Subject: [PATCH] Always send waveform error to Sentry
---
Index: api/api/models/audio.py
<+>UTF-8
===================================================================
diff --git a/api/api/models/audio.py b/api/api/models/audio.py
--- a/api/api/models/audio.py	(revision d813510c86a029e7cb419cd7ea44d281a30d2264)
+++ b/api/api/models/audio.py	(date 1722661866457)
@@ -240,8 +240,8 @@
     def get_or_create_waveform(self):
         add_on, _ = AudioAddOn.objects.get_or_create(audio_identifier=self.identifier)
 
-        if add_on.waveform_peaks is not None:
-            return add_on.waveform_peaks
+        # if add_on.waveform_peaks is not None:
+        #     return add_on.waveform_peaks
 
         add_on.waveform_peaks = generate_peaks(self)
         add_on.save()
Index: api/conf/settings/sentry.py
<+>UTF-8
===================================================================
diff --git a/api/conf/settings/sentry.py b/api/conf/settings/sentry.py
--- a/api/conf/settings/sentry.py	(revision d813510c86a029e7cb419cd7ea44d281a30d2264)
+++ b/api/conf/settings/sentry.py	(date 1722661641127)
@@ -17,8 +17,8 @@
     DjangoIntegration(),
     LoggingIntegration(event_level=None, level=None),
 ]
-
-if not DEBUG and SENTRY_DSN:
+if True:
+    # if not DEBUG and SENTRY_DSN:
     sentry_sdk.init(
         dsn=SENTRY_DSN,
         integrations=INTEGRATIONS,
```
</details>

- go to http://0.0.0.0:50280/v1/audio/e0742ff3-f909-4e1d-9555-8e6399157452/waveform/ to raise an error
- check the error in Sentry

This is a sample error: https://openverse.sentry.io/share/issue/f21fdb1d1067424eafd648dea13d4b77/
Note the stack trace and all error information, no duplicate errors.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/api/utils/image_proxy/extension.py
```diff
@@ -47,7 +46,7 @@ async def get_image_extension(image_url: str, media_identifier) -> str | None:
 
             await _cache_extension(cache, key, ext)
         except Exception as exc:
-            sentry_sdk.capture_exception(exc)
+            logger.error("upstream_thumbnail_exception", e=exc)
```
sarayourfriend, line 49: I don't think this works as expected, or at least, it doesn't work the way `structlog-sentry` wants you to pass exc info:

> This won't automatically collect sys.exc_info() along with the message, if you want to enable this behavior, just pass exc_info=True.

I think it might be this that we need instead?

```suggestion
            logger.error("upstream_thumbnail_exception", exc_info=True)
```

api/conf/settings/logging.py
```diff
@@ -140,7 +143,8 @@ def suppress_unwanted_logs(record: LogRecord) -> bool:
     + shared_processors
     + [
         structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
-    ],
+    ]
+    + [SentryProcessor(event_level=logging.ERROR)],
```
sarayourfriend, line 147: This needs to come before the shared processors otherwise `exc_info` won't work:

> When you want to use structlog's built-in [format_exc_info](http://www.structlog.org/en/stable/api.html#structlog.processors.format_exc_info) processor, make that the SentryProcessor comes before format_exc_info! Otherwise, the SentryProcessor won't have an exc_info to work with, because it's removed from the event by format_exc_info.

api/justfile
```diff
@@ -40,6 +40,9 @@ export PGCLI_VERSION := `grep -A 1 'name = "pgcli"' pdm.lock | grep 'version' |
 install *args:
     pdm install {{ args }}
 
+add *args:
+    pdm add {{ args }}
```
sarayourfriend, line 44: Maybe just a general `pdm` recipe, if this is really necessary? Why not run `ov pdm add` from within the api directory, though?

```suggestion
pdm *args:
    pdm {{ args }}
```

Adding this in just the API justfile is kind of tricky though, it makes everything else that uses PDM inconsistent, right?

You can pass `-p <project directory>` to pdm from the repo root to target a project. I do this often: `ov pdm -p api [...whatever]`. A root justfile recipe `pdm` that requires a target project would be more consistent and work for everything using PDM without creating inconsistencies of recipe availability in different projects.

api/pyproject.toml
```diff
@@ -10,33 +10,34 @@ authors = [
...
-  "future >=0.18.3, <0.19",
-  "limit >=0.2.3, <0.3",
-  "pillow >=10.2.0, <11",
-  "psycopg >=3.1.18, <4",
-  "python-decouple >=3.8, <4",
-  "python-xmp-toolkit >=2.0.2, <3",
-  "sentry-sdk >=1.40.6, <2",
-  "uvicorn[standard] >=0.30, <0.31",
-  "openverse-attribution @ file:///${PROJECT_ROOT}/../packages/python/openverse-attribution",
+    "adrf >=0.1.4, <0.2",
```
sarayourfriend, line 13: Why the whitespace change? The rest of this file uses two spaces for indentation. Did PDM automatically change this?

sarayourfriend (commented):
To test this, you should be able to enable Sentry locally, trigger one of these errors that should go to Sentry, and then confirm it shows up in Sentry.

api/justfile
obulat: Thank you for the `-p` suggestion! I actually added this as a quick workaround and forgot to remove it before pushing the PR

api/pyproject.toml
obulat: Yes, this was an automatic change. I'll try reverting it.

api/conf/settings/sentry.py
```diff
@@ -12,10 +12,16 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DJANGO_DEBUG_ENABLED", default=False, cast=bool)
+
+INTEGRATIONS = [
+    DjangoIntegration(),
+    LoggingIntegration(event_level=None, level=None),
```
obulat, line 18: This is what prevents the error being sent to Sentry twice: once with the correct information, and the other - with the logged json as the error name (as in https://openverse.sentry.io/share/issue/e2bc26efc51048a981a0da629a34930d/)

sarayourfriend:
Just wanted to add a note to myself when I review (or for anyone else reviewing) that because this will include all `logger.<whatever>` calls as breadcrumbs to Sentry, we should make sure those are safe to do. I'm pretty sure they are, but for example we need to make sure no email addresses are logged in OAuth registration routes, if they could accidentally end up in the Sentry breadcrumbs unmasked and therefore publicly available on any issue with a "share"/public link.

sarayourfriend:
The breadcrumbs from previous logs are _definitely_ working: 
image

Now we know @obulat runs the local API with SQL query logging enabled 😅 

I'm reviewing this now with an eye towards seeing if those breadcrumbs could cause us any issues. We don't do query logging in production _but if we did turn it on, for debugging, this would be an issue!_. @obulat at least a preliminary requested change will be to add a filter to exclude Django's query logger from the Sentry logger integration. Here are the Sentry docs on ignoring a logger: https://docs.sentry.io/platforms/python/integrations/logging/#ignoring-a-logger

sarayourfriend:
One thing that could be a problem:

https://github.com/WordPress/openverse/blob/add/structlog-sentry-integration/api/api/controllers/search_controller.py#L355-L362

There we log the query directly, which is just use input. We have the issue #432 which would be relevant here. If these logs are showing up in Sentry as breadcrumbs, it's one more place in addition to the URL where we'd need to mask that.

I don't think we need to change anything here _right now_, but we should keep that in mind when we're deciding what to log. Perhaps logging the query here is duplicative of the request_start log for that request lifecycle anyway, as it already includes the URL and therefore the query. If we want to change something at that location for this PR, it would just be to remove the query property but keep everything else. Even that I'm not sure about though 🙂 We could also ignore breadcrumbs named `query`?

sentry-structlog has the option:

> ignore_breadcrumb_data A list of data keys that will be excluded from [breadcrumb data](https://docs.sentry.io/platforms/python/enriching-events/breadcrumbs/#manual-breadcrumbs). Defaults to keys which are already sent separately, i.e. level, logger, event and timestamp. All other data in event_dict will be sent as breadcrumb data.

sarayourfriend:
Only one more place we might want to think carefully, with by brief review of our API logging:

https://github.com/WordPress/openverse/blob/add/structlog-sentry-integration/api/api/controllers/elasticsearch/helpers.py#L34-L40

We could also increase the structlog-sentry log level to warning instead of info, and avoid most of the potential difficulties here for now. All the places I have questions about are info level, so we'd bypass that at least.

sarayourfriend (changes requested):
Two requested changes:

1. Ignore the Django SQL query logger. We don't have it enabled in production, but we have in the past used it for debugging, and we definitely don't want that ending up in Sentry breadcrumbs.
2. Increase the log level for the integration to warning, so we avoid the questionable breadcrumbs in our info-level logs. **Although**, I see that it's already set to ERROR. @obulat did you create that Sentry example issue before setting it to ERROR? It includes info-level logging, which either indicates the configuration was different when you triggered it or a bug in structlog-sentry that isn't correctly filtering events on log level.

api/conf/settings/sentry.py
sarayourfriend: Maybe you could just link to https://github.com/kiwicom/structlog-sentry?tab=readme-ov-file#logging-as-json in a comment for documentation? The reason it prevents the duplicate logs is specific and important for context and maintainability of the logging configuration (i.e., that we specifically use JSON logging).

sarayourfriend:
@obulat have you had a chance to look at my questions? The main question is just: when you created the example Sentry issue, were you running the PR with different settings or with ERROR set as the only level for Sentry to capture? If they were different settings, then we should re-test to make sure Sentry does not capture any other log events as breadcrumbs. If they were the same, then something is bugged in the library and we need to make sure that we ignore loggers we definitely do not want Sentry creating breadcrumbs from (django's DB logger).

sarayourfriend:
I'm upgrading this to critical because we need to address it ASAP, as innocuous changes to the API that `logger.error` tank our Sentry integrations usefulness until the issue of these logs not being understood by Sentry is resolved.

AetherUnbound:
I've set the breadcrumb level to `logging.WARNING` which should hopefully handle all of the concerns @sarayourfriend raised for now, going to see about testing this and then we can merge it and deploy.

sarayourfriend (approved):
LGTM, thanks @AetherUnbound for testing and the fix

AetherUnbound:
Fix was tested and the result can be seen here: https://openverse.sentry.io/issues/5712727883/?environment=local&project=6107216&query=&referrer=issue-stream&statsPeriod=14d&stream_index=1 (the `Query` lines are a red herring entirely due to local configuration, what's important is that no `logging.info` lines are present in the breadcrumbs).

PR #4705, opened 2024-08-02
Remove the use of `import.meta.env`

obulat:
## Fixes

Fixes the incident caused by incorrectly setting buildtime/runtime variables

## Description

The logger in `utils` does not work as expected: only logging on the server or on the client when running locally.
When we call `getLogger`, we cannot get the information about the current deployment environment (`useNuxtApp` returns `null`)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/nuxt.config.ts
```diff
@@ -127,20 +116,18 @@ export default defineNuxtConfig({
...
+      deploymentEnv: "local",
+      apiUrl: "https://api.openverse.org/",
       providerUpdateFrequency: 3600000,
       savedSearchCount: 4,
       sentry: {
         dsn: "",
-        environment: deploymentEnv,
-        release: import.meta.env.SEMANTIC_VERSION,
+        environment: "local",
+        release: "",
```
sarayourfriend, line 126: Release should be kept reading from `import.meta.env.SEMANTIC_VERSION`, it is definitively a build-time variable. If there's a way to configure it outside of runtime config in order to make that unambiguously clear, we should. It must not be set at runtime (it makes no sense to do so).

frontend/nuxt.config.ts
```diff
@@ -172,7 +159,6 @@ export default defineNuxtConfig({
     cssPath: "~/styles/tailwind.css",
   },
   i18n: {
```
sarayourfriend, line 161: It looks like i18n can be configured with runtime config as well, and potentially a mix, but might require code changes to do so: https://i18n.nuxtjs.org/docs/options/runtime-config

frontend/src/stores/feature-flag.ts
```diff
@@ -204,9 +211,11 @@ export const useFeatureFlagStore = defineStore(FEATURE_FLAG, {
      * are read in the corresponding `initFromCookies` method.
      */
     writeToCookie() {
+      const secure = useRuntimeConfig().public.deploymentEnv === PRODUCTION
```
sarayourfriend, line 214: What's the logic here? Shouldn't it also be applied to staging? Why not, if not?

frontend/.env.template
```diff
@@ -1,58 +1,13 @@
...
-# DEPLOYMENT_ENV=staging
-# NUXT_API_CLIENT_ID=""
-# NUXT_API_CLIENT_SECRET=""
-# NUXT_PUBLIC_API_URL=https://api-staging.openverse.org/
-# NUXT_PUBLIC_SENTRY_DSN=""
-# NUXT_PUBLIC_SENTRY_RELEASE=""
-# SITE_URL=https://staging.openverse.org
-# SITE_DOMAIN=staging.openverse.org
-# PLAUSIBLE_SITE_URL=https://staging.openverse.org
+DEPLOYMENT_ENV="local"#shouldn't be used, probably
```
sarayourfriend, line 1:
```suggestion
```

Definitely shouldn't be used 👍

frontend/test/playwright/playwright.config.ts
```diff
@@ -31,7 +31,9 @@ const config: PlaywrightTestConfig = {
       NUXT_PUBLIC_API_URL: API_URL,
       UPDATE_TAPES: UPDATE_TAPES,
       DEPLOYMENT_ENV: PRODUCTION,
```
zackkrida, line 33:
```suggestion

```

Can be safely removed.

frontend/.env.template
```diff
@@ -1,58 +1,12 @@
...
+NUXT_PUBLIC_SENTRY_DSN=""
+NUXT_PUBLIC_SENTRY_ENVIRONMENT="local"
+NUXT_PUBLIC_SENTRY_RELEASE="latest"
+NUXT_PUBLIC_PLAUSIBLE_ENABLED="true"
+NUXT_PUBLIC_PLAUSIBLE_DOMAIN="localhost"
+NUXT_PUBLIC_PLAUSIBLE_API_HOST="http://localhost:50290"
+NUXT_PUBLIC_AUTO_PAGEVIEWS="true"
+NUXT_PUBLIC_I18N_BASE_URL="http://localhost:8443"
+NUXT_PUBLIC_SITE_URL="http://localhost:8443" #can be read by Nuxt SEO, @see https://nuxtseo.com/site-config/getting-started/how-it-works#_6-runtime-config-and-environment-variables
+NUXT_PUBLIC_SITE_INDEXABLE="true"
```
zackkrida, line 12:
```suggestion
NUXT_API_CLIENT_ID=
NUXT_API_CLIENT_SECRET=

NUXT_PUBLIC_API_URL="https://api.openverse.org/"
NUXT_PUBLIC_DEPLOYMENT_ENV="local"
NUXT_PUBLIC_I18N_BASE_URL="http://localhost:8443"

NUXT_PUBLIC_PLAUSIBLE_API_HOST="http://localhost:50290"
NUXT_PUBLIC_PLAUSIBLE_AUTO_PAGEVIEWS="true"
NUXT_PUBLIC_PLAUSIBLE_DOMAIN="localhost"
NUXT_PUBLIC_PLAUSIBLE_ENABLED="true"

NUXT_PUBLIC_PROVIDER_UPDATE_FREQUENCY=3600000
NUXT_PUBLIC_SAVED_SEARCH_COUNT=8

NUXT_PUBLIC_SENTRY_DSN=""
NUXT_PUBLIC_SENTRY_ENVIRONMENT="local"
NUXT_PUBLIC_SENTRY_RELEASE="latest"

NUXT_PUBLIC_SITE_INDEXABLE="true"
NUXT_PUBLIC_SITE_URL="http://localhost:8443"
```

I'd like to suggest some grouping for readibility and adding the non-public variables to the config as well.

frontend/test/playwright/playwright.config.ts
obulat: Removed

frontend/nuxt.config.ts
```diff
@@ -190,11 +132,6 @@ export default defineNuxtConfig({
     vueI18n: "./src/vue-i18n",
   },
   plausible: {
-    enabled: !isTest,
-    logIgnoredEvents: !isProductionBuild,
-    trackLocalhost: !isProdNotPlaywright,
-    autoPageviews: isProdNotPlaywright,
-    domain: import.meta.env.SITE_DOMAIN,
-    apiHost: import.meta.env.PLAUSIBLE_SITE_URL,
+    trackLocalhost: true,
```
sarayourfriend, line 135: This setting isn't mentioned in the module documentation: https://nuxt.com/modules/plausible

How does it work?

zackkrida:
@obulat we should also be able to delete /server/plugins/indexable.ts as part of this and rely on the runtime env var.

frontend/nuxt.config.ts
zackkrida: @sarayourfriend it's deprecated but still works, and the replacement functionality ignoredHostnames has a bug: https://github.com/nuxt-modules/plausible/issues/30

@obulat let's add a comment referencing the bug.

obulat: I opened an upstream issue: https://github.com/nuxt-modules/plausible/issues/30
I don't know why, but `options.ignoredHostnames` always contains `localhost`, and it's not possible to remove it. So, I used the _deprecated_ setting that still works.

sarayourfriend: Gotcha. A comment would be great, yes!

frontend/src/stores/feature-flag.ts
obulat: I think the Secure cookies didn't work well locally, so the check was for production build (production/staging).

Checked again, and the cookies work fine on localhost, so I used `secure` for all envs.

frontend/nuxt.config.ts
```diff
@@ -127,20 +68,18 @@ export default defineNuxtConfig({
...
-      apiUrl,
+      deploymentEnv: "local",
+      apiUrl: "https://api.openverse.org/",
       providerUpdateFrequency: 3600000,
       savedSearchCount: 4,
       sentry: {
         dsn: "",
-        environment: deploymentEnv,
+        environment: "local",
         release: import.meta.env.SEMANTIC_VERSION,
```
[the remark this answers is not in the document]
obulat: We have to Sentry plugins: one for the nitro server, and one for Nuxt itself (client). In the second one, I don't think we can get the value of an env variable. So, I got the release from `runtimeConfig` there, too.

frontend/src/stores/feature-flag.ts
obulat: Setting cookies to `secure` broke a unit test, but I mocked the value to set it to `secure: false`.

frontend/nuxt.config.ts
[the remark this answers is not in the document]
zackkrida: I'll test, but I believe the client plugin would also be able to reference import.meta.env.SEMANTIC_VERSION and it would be statically replaced + added to the bundle at build time.

zackkrida (approved):
@obulat I have tested this extensively locally and I'm able to set values at build or runtime and see the expected output. Using `ov env` like `ov env NUXT_PUBLIC_SENTRY_RELEASE=1.2.3.4 NUXT_PUBLIC_SITE_INDEXABLE="false" NUXT_PUBLIC_SITE_URL="http://localhost:8443" just frontend/run start` made this easy to test.

sarayourfriend (changes requested):
The only specifically blocking change I am requesting is to move semantic version to app.config.ts, which I've got working locally with the diff below. The diff also includes suggestions for changes to the documentation, but you can take them or leave them based on your opinion of how and what to document. I prefer not to duplicate Nuxt's documentation regarding the purpose and usage of particular variables, but do agree we can be explicit when deliniating which variables go where. Luckily, Nuxt's documentation is actually pretty clear about this, and the _purpose_ for runtimeConfig and app.config are explicit and straightforward in the documentation. We can point to that, but from my perspective it isn't necessary to repeat the documentation. Up to you and other reviewers where the line is there, though.

As I said, and just to clarify, the only blocking change is to move semantic version to app.config (and use it from there). The rest are just suggestions.

**Edit:** Oh, I forgot about part of the diff, the removal of type overrides from typings/nuxt... We don't need those anymore, because `defineNuxtConfig` and `defineAppConfig` hook into Nuxt's build-time type generation.

```diff
diff --git a/frontend/nuxt.config.ts b/frontend/nuxt.config.ts
index dfac91444..a2addf8db 100644
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -5,28 +5,6 @@ import locales from "./src/locales/scripts/valid-locales.json"
 
 import type { LocaleObject } from "@nuxtjs/i18n"
 
-/**
- * This configuration sets values necessary at *build time*
- * to build the app and prerender our pages.
- *
- * Pay special attention when setting runtimeConfig values
- * in this file.
- *
- * Any key in `{runtimeConfig: {}}` can be
- * overwritten with a NUXT_ environment variable,
- * with the camelCase key converted to SCREAMING_SNAKE_CASE.
- *
- * The runtimeConfig values here are either defaults for local
- * development, or placeholders used to register the NUXT_
- * environment variables.
- *
- * See our .env.template for a definitive list of runtime values.
- *
- * Do not use import.meta.env to retrieve environment variables
- * here without careful consideration.
- *
- * @see {@link https://nuxt.com/docs/guide/going-further/runtime-config#example}
- */
 export default defineNuxtConfig({
   srcDir: "src/",
   serverDir: "server/",
@@ -39,11 +17,17 @@ export default defineNuxtConfig({
   },
   compatibilityDate: "2024-07-23",
   css: ["~/assets/fonts.css", "~/styles/accent.css"],
-  // Remember: Can be overwritten by NUXT_* env vars
+  /**
+   * Define available runtime configuration values and their defaults.
+   *
+   * See linked documentation for details, including how to override defaults
+   * with runtime values using environment variables.
+   *
+   * @see {@link https://nuxt.com/docs/api/nuxt-config#runtimeconfig-1}
+   */
   runtimeConfig: {
     apiClientId: "",
     apiClientSecret: "",
-    // Remember: Can be overwritten by NUXT_PUBLIC_* env vars
     public: {
       deploymentEnv: "local",
       apiUrl: "https://api.openverse.org/",
@@ -52,7 +36,7 @@ export default defineNuxtConfig({
       sentry: {
         dsn: "",
         environment: "local",
-        release: import.meta.env.SEMANTIC_VERSION,
+        // Release is a build time variable, and as such, is defined in app.config.ts
       },
     },
   },
diff --git a/frontend/server/plugins/sentry.ts b/frontend/server/plugins/sentry.ts
index dc467a10b..ae4f3cfb5 100644
--- a/frontend/server/plugins/sentry.ts
+++ b/frontend/server/plugins/sentry.ts
@@ -1,4 +1,4 @@
-import { useRuntimeConfig } from "#imports"
+import { useRuntimeConfig, useAppConfig } from "#imports"
 
 import { defineNitroPlugin } from "nitropack/runtime"
 import * as Sentry from "@sentry/node"
@@ -10,13 +10,18 @@ export default defineNitroPlugin((nitroApp) => {
     public: { sentry },
   } = useRuntimeConfig()
 
-  Sentry.init({
+  const appConfig = useAppConfig()
+
+  const sentryConfig = {
     dsn: sentry.dsn,
     environment: sentry.environment,
-    release: sentry.release,
-  })
+    release: appConfig.semanticVersion,
+  }
+
+  Sentry.init(sentryConfig)
+
   Sentry.setContext("render context", { platform: "server" })
-  logger.success("Initialized sentry on the server with config\n", sentry)
+  logger.success("Initialized sentry on the server with config\n", sentryConfig)
 
   nitroApp.hooks.hook("request", (event) => {
     event.context.$sentry = Sentry
diff --git a/frontend/src/app.config.ts b/frontend/src/app.config.ts
new file mode 100644
index 000000000..35052894b
--- /dev/null
+++ b/frontend/src/app.config.ts
@@ -0,0 +1,5 @@
+import { defineAppConfig } from "#imports"
+
+export default defineAppConfig({
+  semanticVersion: import.meta.env.SEMANTIC_VERSION as string,
+})
diff --git a/frontend/src/plugins/01.sentry.client.ts b/frontend/src/plugins/01.sentry.client.ts
index 97c8f2994..d5208c78b 100644
--- a/frontend/src/plugins/01.sentry.client.ts
+++ b/frontend/src/plugins/01.sentry.client.ts
@@ -1,4 +1,4 @@
-import { defineNuxtPlugin, useRuntimeConfig } from "#imports"
+import { defineNuxtPlugin, useRuntimeConfig, useAppConfig } from "#imports"
 
 import * as Sentry from "@sentry/vue"
 
@@ -7,6 +7,8 @@ export default defineNuxtPlugin((nuxtApp) => {
     public: { sentry },
   } = useRuntimeConfig()
 
+  const { semanticVersion } = useAppConfig()
+
   if (!sentry.dsn) {
     console.warn("Sentry DSN wasn't provided")
   }
@@ -14,7 +16,7 @@ export default defineNuxtPlugin((nuxtApp) => {
   Sentry.init({
     dsn: sentry.dsn,
     environment: sentry.environment,
-    release: sentry.release,
+    release: semanticVersion,
     app: nuxtApp.vueApp,
   })
   Sentry.setContext("render context", { platform: "client" })
diff --git a/frontend/typings/nuxt__types/index.d.ts b/frontend/typings/nuxt__types/index.d.ts
index c446f88c4..31b05109c 100644
--- a/frontend/typings/nuxt__types/index.d.ts
+++ b/frontend/typings/nuxt__types/index.d.ts
@@ -21,14 +21,3 @@ declare module "@nuxtjs/i18n" {
     translated: number
   }
 }
-
-declare module "nuxt/schema" {
-  interface RuntimeConfig {
-    apiSecret: string
-  }
-  interface PublicRuntimeConfig {
-    apiUrl: string
-    providerUpdateFrequency: number
-    savedSearchCount: number
-  }
-}
```

frontend/nuxt.config.ts
```diff
@@ -1,116 +1,33 @@
...
-  "Bytespider",
-  "ImagesiftBot",
-  "cohere-ai",
-]
-
 /**
- * Robots.txt rules are configured here via the \@nuxtjs/robots package.
- * @see {@link https://nuxtseo.com/robots/guides/nuxt-config|Robots Config Rules}
+ * This configuration sets values necessary at *build time*
+ * to build the app and prerender our pages.
```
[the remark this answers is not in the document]
zackkrida: > In that sense, the "definitive" list of supported variables shouldn't be .env.template, but the runtimeConfig property of this nuxt configuration and whatever is defined in app.config.ts and pulled from import.meta.env.

If the .env.template is the union of the runtimeConfig values in nuxt.config.ts and the values is defined in app.config.ts, isn't it actually a pretty good "definitive" list of supported variables? Perhaps then a comment in the .env.template explaining that the values come from those two locations would be a better replacement for these docs.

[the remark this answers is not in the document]
sarayourfriend: I suppose so, I just don't see why `.env.template` is necessary, especially because it encourages/requires us to define defaults in two places: that .env.template file _and_ in nux.config.ts's runtimeConfig object.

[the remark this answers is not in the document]
zackkrida: That's a fair point. We probably don't even document it well, and since the frontend doesn't _require_ a `.env` for local development, we might not even need the template. Provided we have some documentation about `ov env` or other mechanisms for setting local runtime variables.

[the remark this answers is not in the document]
obulat: I've added the patch, @sarayourfriend.

I also removed the `env.template` because it's not necessary, as is stated in the comments here.

Runtime config in `nuxt.config` should have the values used for the dev environment, if they are different from the default values. These should be overriden by the runtime variables in the infrastructure, specific for each environment (staging/production).

Build time variables are in set in `appConfig`.

frontend/nuxt.config.ts
```diff
@@ -167,13 +66,39 @@ export default defineNuxtConfig({
...
+        /* Nuxt i18n fields */
+
+        code: "en", // unique identifier for the locale in Vue i18n
+        dir: "ltr",
+        file: "en.json",
+        iso: "en", // used for SEO purposes (html lang attribute)
+
+        /* Custom fields */
+
+        name: "English",
```
sarayourfriend, line 97: BTW, I noticed this incidentally when looking through the `AppConfig` and `RuntimeConfig` types yesterday, the i18n module actually defines `name` as a property now, so it isn't a custom field anymore :open_mouth:

Nothing to change in this PR, these comments aren't new, but just thought I would mention it.

frontend/src/utils/focus-management.ts
```diff
@@ -36,7 +36,7 @@ const focusableSelector = [
   "textarea:not([disabled])",
 ]
   .map(
-    import.meta.env.NODE_ENV === "test"
+    import.meta.env?.NODE_ENV === "test"
```
sarayourfriend, line 39: I wonder if this should use `import.meta.test` instead? Documented here: https://nuxt.com/docs/api/advanced/import-meta#runtime-app-properties

Not necessary for this PR, but if you think so, maybe an easy follow up issue for later on.

frontend/test/playwright/playwright.config.ts
sarayourfriend: It looks like this got added back.

sarayourfriend (approved):
LGTM! I noticed two small things, but they are not blockers. Feel free to resolve them now (remove `DEPLOYMENT_ENV` from the playwright.config, use `import.meta.test` instead of `NODE_ENV === 'test'`) or as a follow up issue; neither need to block this PR any further.

Thanks for hanging in with the several rounds of review and feedback @obulat :pray: Glad we've sorted out how this works. The Nuxt documentation ended up being pretty good in the end, at least I just needed to actually read the various pages on configuration in whole, lol! :slightly_smiling_face:

frontend/test/playwright/playwright.config.ts
obulat: Replaced with `NUXT_PUBLIC_DEPLOYMENT_ENV`, which is necessary for the feature flags to work correctly.

frontend/nuxt.config.ts
obulat: Removed 👍

frontend/src/utils/focus-management.ts
obulat: Replaced 👍

sarayourfriend:
LGTM @obulat! In any case, we can easily verify things are working correctly in staging.

PR #4716, opened 2024-08-05
Update nuxtjs/plausible

obulat:
## Fixes

When analytics is disabled, the `fetch` call throws an error that cannot be caught in the previous versions of Plausible. 
Additionally, the new option for not ignoring `localhost` wasn't working with an empty `[]`.

## Description

This PR updates the `nuxtjs/plausible` package to [v1.0.2](https://github.com/nuxt-modules/plausible/releases/tag/v1.0.2) fix the above issues.

## Testing Instructions

Run the app and open http://localhost:8443 in a browser with privacy focus (Firefox).
Open the browser console and clear it.
Select "Image" in the search type dropdown. This should attempt to send the event to Plausible.

On main, you can see `Uncaught (in promise) TypeError: NetworkError when attempting to fetch resource. ` when you select a search type from the homepage (when sending this event to Plausible). In this branch, you will not see this error.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/nuxt.config.ts
```diff
@@ -192,7 +192,7 @@ export default defineNuxtConfig({
   plausible: {
     enabled: !isTest,
     logIgnoredEvents: !isProductionBuild,
-    trackLocalhost: !isProdNotPlaywright,
+    ignoredHostnames: [],
```
zackkrida, line 195: @obulat I was thinking about this, do you think "staging.openverse.org" should be in here? And if "localhost" is in here, I am curious; are the events still logged? If yes, it might make sense to hardcode this as `["staging.openverse.org", "localhost"] here and set it as an empty array in Playwright's config. That assumes we need to turn on tracking in Playwright to properly test that the events are firing.

obulat, line 195:
```suggestion
    ignoredHostnames: [],
    trackLocalhost: true,
```

frontend/nuxt.config.ts
```diff
@@ -191,10 +190,14 @@ export default defineNuxtConfig({
   },
   plausible: {
     enabled: !isTest,
+    // In dev, the events are not sent anywhere, only logged to the console.
+    // In production builds, ignore events from staging.
+    ignoredHostnames: !isProductionBuild
+      ? ["localhost"]
+      : ["staging.openverse.org"],
```
zackkrida, line 197: @obulat Since we remove "isProductionBuild" in #4705, should we use it here? I tried to ask this in my earlier comment but wouldn't `["localhost", "staging.openverse.org"]` be a safe value in _all_ environments, regardless of the build type?

frontend/test/playwright/playwright.config.ts
```diff
@@ -27,10 +27,12 @@ const config: PlaywrightTestConfig = {
...
     env: {
-      API_URL,
+      UPDATE_TAPES,
       NUXT_PUBLIC_API_URL: API_URL,
-      UPDATE_TAPES: UPDATE_TAPES,
-      DEPLOYMENT_ENV: PRODUCTION,
+      NUXT_PUBLIC_DEPLOYMENT_ENV: PRODUCTION,
+      NUXT_PUBLIC_PLAUSIBLE_API_HOST: "http://localhost:50290",
+      NUXT_PUBLIC_PLAUSIBLE_DOMAIN: "localhost",
+      NUXT_PUBLIC_PLAUSIBLE_AUTO_PAGEVIEWS: "false",
```
obulat, line 35:
```suggestion
      NUXT_PUBLIC_PLAUSIBLE_AUTO_PAGEVIEWS: "false",
      NUXT_PUBLIC_SITE_INDEXABLE: "true",
```

frontend/nuxt.config.ts
```diff
@@ -140,7 +139,7 @@ export default defineNuxtConfig({
     },
   },
   site: {
-    indexable: deploymentEnv === PRODUCTION,
+    indexable: false,
```
obulat, line 142:
```suggestion
```

frontend/nuxt.config.ts
obulat: I added the env variable to enable logging the ignored events when localhost is ignored.

frontend/nuxt.config.ts
```diff
@@ -191,10 +189,11 @@ export default defineNuxtConfig({
...
   plausible: {
     enabled: !isTest,
-    logIgnoredEvents: !isProductionBuild,
-    trackLocalhost: !isProdNotPlaywright,
-    autoPageviews: isProdNotPlaywright,
-    domain: import.meta.env.SITE_DOMAIN,
-    apiHost: import.meta.env.PLAUSIBLE_SITE_URL,
+    ignoredHostnames: ["localhost", "staging.openverse.org"],
+    // Set `NUXT_PUBLIC_PLAUSIBLE_LOG_IGNORED_EVENTS` to "true" to log in the dev environment.
+    logIgnoredEvents: false,
```
zackkrida, line 194: Any reason not to log by default? Since these logs are always on the client it doesn't seem too intrusive or disruptive. And in production since the events are not ignored, they wouldn't log. Is that right?

obulat: If set to true, it will log in two cases:
- the plausible domain is on the list of ignored domains (so, it will log in staging, and in dev mode)
- the user has switched off analytics (this might confuse some users if they open console and see events logged after they've turned off the analytics)

I think that we should only be logging in the dev mode; not on staging or when the user has switched off analytics.

After having more thought about that, I think the best way to fix this would be to set `logIgnoredEvents` to `true` in `nuxt.config`, but add the `NUXT_PUBLIC_PLAUSIBLE_LOG_IGNORED_EVENTS=false` to staging and production. What do you think, @zackkrida?

obulat: I added this variable to https://github.com/WordPress/openverse-infrastructure/pull/997

obulat, line 194:
```suggestion
    logIgnoredEvents: true,
```

frontend/.env.template
```diff
@@ -56,3 +56,5 @@ PORT=8443
 # SITE_URL=https://staging.openverse.org
 # SITE_DOMAIN=staging.openverse.org
 # PLAUSIBLE_SITE_URL=https://staging.openverse.org
+
+NUXT_PUBLIC_PLAUSIBLE_LOG_IGNORED_EVENTS=true
```
obulat, line 60:
```suggestion
```

frontend/nuxt.config.ts
zackkrida: @obulat thank you for the extra info, I didn't consider the case of users who have disabled analytics! 

The new approach sounds good to me with one small exception: why wouldn't we want log the events on staging?

obulat: Since the staging is partly public-facing, I thought it would be better to disable logging there. But disabling it on staging would not allow us to understand if Plausible events are being fired. I'll update this change.

frontend/nuxt.config.ts
```diff
@@ -115,8 +115,9 @@ export default defineNuxtConfig({
     vueI18n: "./src/vue-i18n",
   },
   plausible: {
-    // `trackLocalhost` is deprecated, but the replacement `ignoredHostnames: []`
-    // has a bug, @see https://github.com/nuxt-modules/plausible/issues/30
-    trackLocalhost: true,
+    ignoredHostnames: ["localhost", "staging.openverse.org"],
+    logIgnoredEvents: true,
+    apiHost: "http://localhost:50290",
+    domain: "localhost",
```
sarayourfriend, line 121: Shouldn't this all be moved to runtime config?

sarayourfriend (changes requested):

sarayourfriend (approved):

PR #4725, opened 2024-08-07
Fix the sources table crash

### LINKED ISSUE #4724: Repeated sorting of the sources table causes page freeze
## Description

Clicking on the table header more than once results in the site hanging or even crashing. Must be some faulty logic or recursion in the sorting logic.

Caused by a circular `computed` in VSourcesTable when sorting the list. Because the computed pulls a reference from the store, and uses Array.prototype.sort to sort the list, the computed itself mutates the state that it depends on, causing the loop.

Copying the list in the computed is very slow. Instead, we can either use a `watch`, so we can explicitly define the dependencies of the `sortedProviders`, or sort the list in the event callback whenever the table headers are clicked and store the sorted list in a plain `ref`, instead of the computed.

## Reproduction

_Be careful: this will hang your browser_
1. Go to https://openverse.org/sources
2. Click on one of the sources table heading twice
3. See error. The browser hangs.

## Additional context

First reported by @zackkrida

obulat:
## Fixes

Fixes #4724 by @obulat

## Description

This PR replaces a computed for providers with a ref that is updated on a table header click.
It also changes the sorting order for the items count column: when you first click it, the items are sorted in a descending order instead of ascending order as in prod. This feels like an expected behavior for this column.

The providers data is fetched on every request in a plugin when starting the app, and is available when the user opens the sources page.

This fixes both the page hanging and the server-client mismatch that is available in production.

## Testing Instructions

Go to the sources page and click on the last column header several times. The rows should be sorted, and page should not freeze. You should also not see any `client-server hydration mismatch` logs in the console.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/e2e/sources.spec.ts
```diff
@@ -16,3 +16,17 @@ test("sources table has links to source pages", async ({ page }) => {
 
   await expect(getH1(page, "Flickr")).toBeVisible()
 })
+
+// Tests the fix for https://github.com/WordPress/openverse/issues/4724
+test("sources table can be sorted", async ({ page }) => {
```
dhruvkb, line 21:
```suggestion
test("sources table can be sorted multiple times", async ({ page }) => {
```

dhruvkb (approved):
LGTM, the sorting works and clicking it multiple times does not crash anymore. It also fixes a nit I had with the production version that sorts in ascending order by default.

PR #4730, opened 2024-08-08
Fix mobile recent searches modal with banners

### LINKED ISSUE #2603: Prevent the header overlapping the main content when the banners are visible and skip-to-content is clicked
## Description

In #2538 we've added simplified skip-to-content links. The problem is that when the users click on the links, the page scrolls under the header. To prevent that, we've added the `scroll-mt-20` class. However, when the banners are visible, the header area becomes taller, and the content is still hidden beneath the header when you click on the "skip-to-content" link.

## Solution
We should use a `--header-height` css variable that is computed in the layouts to account for all of the banners and the header. This CSS variable should be used to set the `scroll-margin` of the `skip-to-content` link.

## Additional information
Originally noted in 2538 by @sarayourfriend :
> LGTM. The changes to prevent the header overlap work when there are no banners. With one banner at least part of the word is still visible, so while not ideal, there's the visual queue to scroll up a little to see the heading. Ideally we'd be able to correct the amount of margin for the number of banners present. A CSS custom property that gets updated by the banner components when they render is the only way I can think to do that, but it would need to be a manual approach rather than using the tailwind utilities for this.

_Originally posted by @sarayourfriend in https://github.com/WordPress/openverse/pull/2538#pullrequestreview-1521222386_

### LINKED ISSUE #4729: Input is not visible when recent searches are open on mobile and banners are visible
## Description

The recent searches modal hides part of the search bar when a banner is present.

## Reproduction

1. Clear `ui` cookies in your browser to make the analytics banner appear.
2. Go to https://openverse.org/search?q=cat
3. Click on the search bar to open the recent searches modal.
4. See error: more than half of the search bar is hidden.

## Screenshots

Screenshot 2024-08-08 at 7 32 07 AM

obulat:
_Marked as critical because this issue does not allow search on mobile when the banner is not dismissed_

## Fixes

Fixes #4729 by @obulat

## Description

This PR adds a `--header-height` CSS variable that is then used to compute the height of the modal backdrop, and the modal offset from the top to make sure that any banners inside the header element are taken into account.

I tried to also fix #2603 in this PR, but because clicking "Skip to content" _navigates_ to a `#content` hash route, it requires a [router options](https://nuxt.com/docs/guide/recipes/custom-routing#router-options) [scroll behavior changes](https://router.vuejs.org/guide/advanced/scroll-behavior).

## Testing Instructions

1. Clear the `ui` cookies in your browser, or use _incognito_ mode.
2. Use the _mobile, narrow_ screen width.
3. Run the app and go to `http://localhost:8443/search?q=cat` or `http://localhost:8443/ru/search?q=cat`
4. You should see a banner or two banners at the top.
5. Click on the search bar. The recent searches modal should open, and it should not hide any part of the search bar.

Now do the same in production: on https://openverse.org/ru/search?q=cat you should see 2 banners, and the recent searches modal hides the bottom banner as well as the search bar.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/layouts/search-layout.vue
+++ b/frontend/src/layouts/search-layout.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { computed, onMounted, provide, ref, watch } from "vue"
-import { useScroll } from "@vueuse/core"
+import { useElementSize, useScroll } from "@vueuse/core"
 
 import { useUiStore } from "~/stores/ui"
 import { isSearchTypeSupported, useSearchStore } from "~/stores/search"
@@ -45,6 +45,7 @@ const showScrollButton = ref(false)
  * Note: template refs do not work in a Nuxt layout, so we get the `main-page` element using `document.getElementById`.
  */
 let mainPageElement = ref<HTMLElement | null>(null)
+let headerElement = ref<HTMLElement | null>(null)
 
 const { y: mainPageY } = useScroll(mainPageElement)
 watch(mainPageY, (y) => {
@@ -54,6 +55,14 @@ watch(mainPageY, (y) => {
 
 onMounted(() => {
   mainPageElement.value = document.getElementById("main-page")
+  headerElement.value = document.getElementsByClassName(
+    "header-el"
+  )[0] as HTMLElement
+})
+
+const { height } = useElementSize(headerElement)
+watch(height, (height) => {
+  uiStore.setHeaderHeight(height)
 })
 
 provide(IsHeaderScrolledKey, isHeaderScrolled)
--- a/frontend/test/playwright/e2e/recent-searches.spec.ts
+++ b/frontend/test/playwright/e2e/recent-searches.spec.ts
@@ -2,6 +2,7 @@ import { expect, test, type Page } from "@playwright/test"
 
 import {
   goToSearchTerm,
+  isPageDesktop,
   preparePageForTests,
   searchFromHeader,
   sleep,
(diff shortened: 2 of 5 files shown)


frontend/src/layouts/search-layout.vue
```diff
@@ -54,6 +55,14 @@ watch(mainPageY, (y) => {
...
 onMounted(() => {
   mainPageElement.value = document.getElementById("main-page")
+  headerElement.value = document.getElementsByClassName(
+    "header-el"
+  )[0] as HTMLElement
+})
+
+const { height } = useElementSize(headerElement)
+watch(height, (height) => {
+  uiStore.setHeaderHeight(height)
```
sarayourfriend, line 65: This is miserable :disappointed: I wish there was another easy way to fix this, but we've put ourselves into such a hole by managing the layout of these elements with JavaScript instead of CSS :face_with_diagonal_mouth: 

At the very least, it would be nice to clean this up in a fast-follow by creating a component to encapsulate this behaviour instead of selecting on the classname in a random part of the app. 

Not blocking this due to criticality.

frontend/test/playwright/e2e/recent-searches.spec.ts
```diff
@@ -100,53 +133,76 @@ breakpoints.describeMobileXsAndDesktop(({ breakpoint }) => {
     const recentSearchesText = await getRecentSearchesText(page)
     expect(recentSearchesText).toContain(noRecentLabel)
   })
+})
 
-  test("can open recent searches with keyboard", async ({ page }) => {
-    await executeSearches(page)
-    await tabToSearchbar(page)
-    await page.keyboard.press("ArrowDown")
+breakpoints.describeMobileXsAndDesktop(({ breakpoint }) => {
```
sarayourfriend, line 138: In a fast-follow PR, we should add a test that ensures the search box can be clicked when the recent searches is open at each breakpoint.

sarayourfriend (approved):
LGTM tested well locally. I have two comments that, were this PR not critical, I would block on. Instead, they should be done in a follow-up PR.

PR #4731, opened 2024-08-08
Fix Playwright test failures when Nuxt context is not available

### LINKED ISSUE #4737: CI failing due to Nuxt context being unavailable during feature flag store setup
## Description

Playwright tests fail intermittently due to `[nuxt] context` being unavailable.
This was introduced when we replaced all instances of `import.meta.env` with `useRuntimeConfig` in the feature flags store state initialization. Sometimes, feature flag store is called before the Nuxt context is available, so `useRuntimeConfig` throws an error.

## Reproduction

See error in this CI run: https://github.com/WordPress/openverse/actions/runs/10300287434/job/28509472846?pr=4716. Note that there are different failing tests during each CI run, so the failure is intermittant.

obulat:
## Fixes

Fixes #4737 by @obulat

## Description

The failures were caused by race conditions when setting up the feature flag store. 

When setting up the state, we need the `runtimeConfig`'s `deploymentEnv` to compute the feature flag statuses and states. Sometimes it was called when the Nuxt context necessary for `useRuntimeConfig` wasn't available, and this caused errors down the line. This only happened on the server.

This PR adds a check if the Nuxt app context is available to the feature flags store. If it's not available, it is possible to get the `deploymentEnv` from the runtime env variable (`NUXT_PUBLIC_DEPLOYMENT_ENV`). It also throws an error when the env variable is not available - but that should never happen.

Some code is also refactored to avoid calling feature flag store from deeply nested functions (e.g., function that is called by the function returned from `createApiService`, which is called by the media store) as that could cause loss of Nuxt context (more on this in the [Nuxt docs](https://nuxt.com/docs/guide/concepts/auto-imports#vue-and-nuxt-composables))

## Testing Instructions

See that the latest Playwright CI runs in this PR ran successfully. I will run the tests several times.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/plugins/03. analytics.ts
+++ b/frontend/src/plugins/03. analytics.ts
@@ -2,7 +2,6 @@ import { defineNuxtPlugin, useTrackEvent } from "#imports"
 
 import type { Events, EventName } from "~/types/analytics"
 import { useUiStore } from "~/stores/ui"
-import { useFeatureFlagStore } from "~/stores/feature-flag"
 
 type SendCustomEvent = <T extends EventName>(
   name: T,
@@ -20,9 +19,6 @@ export default defineNuxtPlugin(() => {
   }
 
   const uiStore = useUiStore()
-  const featureFlagStore = useFeatureFlagStore()
-
-  featureFlagStore.syncAnalyticsWithLocalStorage()
 
   const sendCustomEvent: SendCustomEvent = (name, payload) => {
     useTrackEvent(name, {
--- a/frontend/src/stores/feature-flag.ts
+++ b/frontend/src/stores/feature-flag.ts
@@ -1,4 +1,4 @@
-import { useCookie, useRuntimeConfig } from "#imports"
+import { tryUseNuxtApp, useCookie } from "#imports"
 
 import { defineStore } from "pinia"
 import { useStorage } from "@vueuse/core"
@@ -131,8 +131,20 @@ const FEATURE_FLAG = "feature_flag"
 
 export const useFeatureFlagStore = defineStore(FEATURE_FLAG, {
   state: () => {
-    const deploymentEnv = useRuntimeConfig().public.deploymentEnv as DeployEnv
-    return initializeFlagState(deploymentEnv)
+    const nuxtApp = tryUseNuxtApp()
+    if (nuxtApp) {
+      return initializeFlagState(
+        nuxtApp.$config.public.deploymentEnv as DeployEnv
+      )
+    }
+    if (!process.env.NUXT_PUBLIC_DEPLOYMENT_ENV) {
+      throw new Error(
+        "Cannot initialize feature flags store. Deployment environment not set in the Nuxt app context and the env variable is unavailable."
+      )
+    }
+    return initializeFlagState(
+      process.env.NUXT_PUBLIC_DEPLOYMENT_ENV as DeployEnv
+    )
   },
   getters: {
     /**
--- a/frontend/src/stores/feature-flag.ts
+++ b/frontend/src/stores/feature-flag.ts
@@ -1,4 +1,4 @@
-import { tryUseNuxtApp, useCookie } from "#imports"
+import { useCookie, useRuntimeConfig } from "#imports"
 
 import { defineStore } from "pinia"
 import { useStorage } from "@vueuse/core"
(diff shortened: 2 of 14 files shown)


frontend/test/playwright/visual-regression/pages/search-with-banners.spec.ts
```diff
@@ -14,7 +14,7 @@ breakpoints.describeEvery(({ breakpoint, expectSnapshot }) => {
       dismissBanners: false,
       dismissFilter: false,
     })
-    await page.goto("/ru/search/?q=birds&referrer=creativecommons.org")
+    await page.goto("/ru/search?q=birds&referrer=creativecommons.org")
```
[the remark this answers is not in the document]
obulat: The snapshot did not have the referrer banner :) Thank you for catching this!

frontend/test/playwright/utils/page.ts
```diff
@@ -4,18 +4,6 @@
 
 import type { Page } from "@playwright/test"
 
-export const removeHiddenOverflow = async (page: Page) => {
```
obulat, line 7: We no longer have `embedded` mode, and this function was not used for a long time

frontend/src/stores/feature-flag.ts
```diff
@@ -131,8 +131,23 @@ const FEATURE_FLAG = "feature_flag"
 
 export const useFeatureFlagStore = defineStore(FEATURE_FLAG, {
   state: () => {
-    const deploymentEnv = useRuntimeConfig().public.deploymentEnv as DeployEnv
-    return initializeFlagState(deploymentEnv)
+    const nuxtApp = tryUseNuxtApp()
```
obulat, line 134: `useRuntimeConfig` throws an error when the Nuxt context isn't available. The same is true for `useNuxtApp`. This is why we use `tryUseNuxtApp` here, which returns `null` instead of throwing an error.

frontend/src/plugins/03. analytics.ts
```diff
@@ -20,9 +19,6 @@ export default defineNuxtPlugin(() => {
   }
 
   const uiStore = useUiStore()
-  const featureFlagStore = useFeatureFlagStore()
```
obulat, line 23: The sync is done in `app.vue` and `error.vue` `onMounted`, no need to call it here

frontend/src/stores/feature-flag.ts
```diff
@@ -131,8 +131,23 @@ const FEATURE_FLAG = "feature_flag"
...
+    if (nuxtApp) {
+      return initializeFlagState(
+        nuxtApp.$config.public.deploymentEnv as DeployEnv
+      )
+    }
+
+    const deploymentEnv = process.env.NUXT_PUBLIC_DEPLOYMENT_ENV as DeployEnv
```
sarayourfriend, line 141: I guess this is _technically_ safe, because as you said, this loss of context occurs on the server, so client code will never reach this line... but isn't `import.meta.env` the correct way to read an environment variable, for idempotent code (which this technically is, as it isn't explicitly scoped)?

I also wonder, after reading the Nuxt docs, whether this fallback approach actually paints over a deeper problem of misused composables and asynchronous code. Based on the Nuxt docs you linked, it seemed like the real fix was to prevent this code running in an unsupported context, rather than papering over the issue with this fallback.

Is the fallback still necessary after the other refactors you made to remove the deep nesting? It would be nice to do that refactor first, and then see if the problem still happens, and if so trying and find the root cause rather than a band-aid fix. At least, that's if I've understood the docs correctly when it comes to what causes this lack of Nuxt context in certain code paths.

obulat: > Is the fallback still necessary after the other refactors you made to remove the deep nesting?

Debugging this is kind of tricky since this happens so rarely. But from what I can see, we _**don't need**_ this anymore. Before removing the deep nesting, we had 2-3 test failures per Playwright test suite run. Now, after adding these changes, and rebasing another PR onto this one, I don't see the failures AND there are no logs that would happen if nuxt content wasn't available in the store.

frontend/src/stores/feature-flag.ts
zackkrida: Could we reformat this a bit?

```js
  const nuxtApp = tryUseNuxtApp();
  
  if (!nuxtApp) {
    throw new Error(
      "Could not set up feature flag store because Nuxt content isn't available."
    );
  }
  
  return initializeFlagState(
    nuxtApp.$config.public.deploymentEnv as DeployEnv
  );
```

zackkrida (approved):
It's a bit of a shame to have to repeat the fake_sensitivity logic in four different stores. I hate to block this but I think a light abstraction, like the following composable would help prevent us from missing one of those in the future, and also making it easier to pass future values to the api client. 

```ts
import { useNuxtApp } from '#imports'
import { useFeatureFlagStore } from '~/stores/feature-flag'
import { createApiClient } from '~/data/api-service'

export function useApiClient() {
  const { $openverseApiToken: accessToken } = useNuxtApp()
  const featureFlagStore = useFeatureFlagStore()

  const fakeSensitive =
    featureFlagStore.isOn('fake_sensitive') &&
    featureFlagStore.isOn('fetch_sensitive')

  return createApiClient({ accessToken, fakeSensitive })
}
```

obulat:
> It's a bit of a shame to have to repeat the fake_sensitivity logic in four different stores. I hate to block this but I think a light abstraction, like the following composable would help prevent us from missing one of those in the future, and also making it easier to pass future values to the api client.
> 
> ```ts
> import { useNuxtApp } from '#imports'
> import { useFeatureFlagStore } from '~/stores/feature-flag'
> import { createApiClient } from '~/data/api-service'
> 
> export function useApiClient() {
>   const { $openverseApiToken: accessToken } = useNuxtApp()
>   const featureFlagStore = useFeatureFlagStore()
> 
>   const fakeSensitive =
>     featureFlagStore.isOn('fake_sensitive') &&
>     featureFlagStore.isOn('fetch_sensitive')
> 
>   return createApiClient({ accessToken, fakeSensitive })
> }
> ```

Awesome idea, added here. Thanks!

PR #4735, opened 2024-08-09
Disallow crawlers from high cardinality collection pages

sarayourfriend:
## Fixes

Related to ongoing issue with crawlers going through the collection endpoints

## Description

Collection pages for creator and tag have unlimited cardinality. Crawlers end up spending heaps of time quickly going through the tag and creator pages. Source collections on the other hand have bounded cardinality based on the number of sources we've defined for each media type, so that is safe. However, we do not want to disallow bots _at all_ from these pages, so we cannot add them to robots.txt. Instead, we use the meta tags (and header, thanks to `@nuxtjs/robots` for handling this transparently) to tell crawlers it's okay to visit the page, but not to follow the links. This makes sure that OpenGraph tags are always accessible on relevant pages, and that pages we want indexed can still be indexed without causing a flood of new requests due to crawling results on those pages.

I've also added a new Playwright snapshot test so we can easily audit changes to the robots.txt in diffs of PRs. The locales reflect those enabled for Playwright, which are a subset used for testing. The actual file uses the real locales, and you can see it by following the testing instructions below. This new test supplements the route specific tests.

Finally, we had unnecessarily duplicative robots rules. Robots.txt rules are prefix based, so there's no need to have `/search` and `/search/images`. `/search` already covers everything that starts with search.

https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt?csw=1

## Testing Instructions

Review the new robots.txt by:

1. Checking out the new snapshot in the playwright tests in the diff of this PR
2. Run the Nuxt dev server locally (`ov nuxt`) and visit <http://127.0.0.1:8443/robots.txt?mockProductionEnv>; this will show you the full robots.txt with all locales (if you have them downloaded, which you should if using `ov nuxt`).

Review the updated Playwright tests that prove the meta tags are appropriate for each of the relevant scenarios. You can also test it locally by visiting the pages and adding `?mockProductionEnv` to them. Otherwise you will get the default for local environments on pages that do not explicitly use `useRobotsRule` which is noindex,nofollow (import for checking the static pages, for example).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
I will update this PR to only disallow for tag and creator collections. The tag and creator parameters have unlimited cardinality, and crawlers are going very fast through those pages. Source collections are fine because source has a bounded cardinality.

frontend/nuxt.config.ts
```diff
@@ -68,14 +68,14 @@ export default defineNuxtConfig({
   },
   /**
    * Robots.txt rules are configured here via the \@nuxtjs/robots package.
-   * @see {@link https://nuxtseo.com/robots/guides/nuxt-config|Robots Config Rules}
+   * @see {@link https://nuxtseo.com/robots/guides/nuxt-config}
    */
   robots: {
-    disallow: ["/search", "/search/audio", "/search/image"],
+    disallow: ["/search"],
```
sarayourfriend, line 74: Robots rules are prefix based. As such, there's no need to define `/search/audio` (and image) because `/search` already matches everything that starts with `/search`.

frontend/nuxt.config.ts
```diff
@@ -68,14 +68,14 @@ export default defineNuxtConfig({
   },
   /**
    * Robots.txt rules are configured here via the \@nuxtjs/robots package.
-   * @see {@link https://nuxtseo.com/robots/guides/nuxt-config|Robots Config Rules}
+   * @see {@link https://nuxtseo.com/robots/guides/nuxt-config}
```
sarayourfriend, line 71: This `|Robots ...` bit broke the link when clicking it from the editor.

frontend/test/playwright/e2e/seo.spec.ts
```diff
@@ -8,6 +8,7 @@ const DESCRIPTION =
 const NO_INDEX = "noindex, nofollow"
 const INDEX =
   "index, follow, max-image-preview:large, max-snippet:-1, max-video-preview:-1"
+const INDEX_NO_FOLLOW = "index, nofollow"
```
sarayourfriend, line 11: I wonder if this should have the `max-` directives as well, if index is enabled?

obulat (approved):
Nice to have the robots text in the tests to compare the values!

I _think_ the Playwright failures can be fixed if rebased onto #4731.

zackkrida (approved):
Great idea for the snapshot test, kudos! This LGTM.

frontend/src/middleware/collection.ts
```diff
@@ -78,6 +83,14 @@ export const collectionMiddleware = defineNuxtRouteMiddleware(async (to) => {
     })
   }
 
+  const s = useSiteConfig()
+
+  if (s.indexable && collectionParams.collection === "source") {
```
zackkrida, line 88:
```suggestion
  const siteConfig = useSiteConfig()

  if (siteConfig.indexable && collectionParams.collection === "source") {
```

Super nit: from what I've seen we usually favor explicit verbosity over brevity in most of the other frontend code.

sarayourfriend: Whoops, definitely, I wasn't sure `useSiteConfig` was going to work and meant to change it if it did! Thanks for catching that.

frontend/src/pages/audio/collection.vue
```diff
@@ -30,6 +33,12 @@ const {
   pageTitle,
 } = useCollection({ mediaType: AUDIO })
 
+// Collection params are not nullable in the collections route, this is enforced by the middleware
+// Question: should this non-nullability be filtered in the type and enforced in runtime by `useCollection`?
```
obulat, line 37: Good idea

obulat (approved):
LGTM

frontend/src/pages/audio/collection.vue
sarayourfriend: https://github.com/WordPress/openverse/issues/4740

PR #4747, opened 2024-08-13
Only cache branch protection within the scope of the function

### LINKED ISSUE #4734: PR review reminder branch protection cache is indefinite
## Description

The PR review reminder branch protection cache is indefinite, it has no TTL or expiration. That means that even when we've changed the branch protection rules to reduce the required number of reviewers, the DAG continues to use the branch protection rules from when it first retrieved them, presumably the first time the DAG ran when the Airflow instance last restarted.

We can "fix" this by restarting Airflow, but it'd be better to cache these responses for a set amount of time, like 1 day, or perhaps only during the lifetime of the DAG run. (Note: those are equivalent now, because the DAG runs once every weekday, but caching it for the duration of the DAG run would prevent a weird race condition where the response expired during the next DAG run and different PRs received different treatment; would be better to just cache it _for the DAG run_.

## Reproduction

1.  See this PR where I've requested changes, and therefore the required number of reviews is fulfilled, but the DAG still pinged the other reviewer to give the second review: https://github.com/WordPress/openverse/pull/4704#issuecomment-2276925121

## Additional context

Marked as high because this will cause noisy and unnecessary pings so long as it isn't fixed, which reduces the value and reliability and worsens sentiments towards pings over time.

AetherUnbound:
## Fixes

Fixes #4734 by @sarayourfriend

## Description

This PR moves the branch protection cache from the module to within the scope of the `get_min_required_approvals` function. Based on the way Airflow manages its workers and processes, this cache was created at module import time and persisted throughout the duration of the Airflow scheduler process. Scoping this to the function ensures that it's recreated on every DAG run.

## Testing Instructions

Tests should continue to pass!

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
+++ b/catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
@@ -182,8 +182,9 @@ def get_branch_protection(
     return cache[repo][branch_name]
 
 
-def get_min_required_approvals(gh: GitHubAPI, pr: dict) -> int:
-    branch_protection_cache = defaultdict(dict)
+def get_min_required_approvals(
+    gh: GitHubAPI, pr: dict, branch_protection_cache: dict
+) -> int:
     repo = base_repo_name(pr)
     branch_name = pr["base"]["ref"]
 
@@ -218,6 +219,7 @@ def get_min_required_approvals(gh: GitHubAPI, pr: dict) -> int:
 @task(task_id="pr_review_reminder_operator")
 def post_reminders(maintainers: set[str], github_pat: str, dry_run: bool):
     gh = GitHubAPI(github_pat)
+    branch_protection_cache = defaultdict(dict)
 
     open_prs = []
     for repo in REPOSITORIES:
@@ -271,7 +273,9 @@ def post_reminders(maintainers: set[str], github_pat: str, dry_run: bool):
         existing_reviews = gh.get_pull_reviews(base_repo_name(pr), pr["number"])
 
         approved_reviews = [r for r in existing_reviews if r["state"] == "APPROVED"]
-        if len(approved_reviews) >= get_min_required_approvals(gh, pr):
+        if len(approved_reviews) >= get_min_required_approvals(
+            gh, pr, branch_protection_cache
+        ):
             # if PR already has sufficient reviews to be merged, do not ping
             # the requested reviewers.
             continue
--- a/catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
+++ b/catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
@@ -173,30 +173,32 @@ def base_repo_name(pr: dict):
     return pr["base"]["repo"]["name"]
 
 
-_BRANCH_PROTECTION_CACHE = defaultdict(dict)
+def get_branch_protection(
+    gh: GitHubAPI, repo: str, branch_name: str, cache: dict
+) -> dict:
+    if branch_name not in cache[repo]:
+        cache[repo][branch_name] = gh.get_branch_protection(repo, branch_name)
 
-
-def get_branch_protection(gh: GitHubAPI, repo: str, branch_name: str) -> dict:
-    if branch_name not in _BRANCH_PROTECTION_CACHE[repo]:
-        _BRANCH_PROTECTION_CACHE[repo][branch_name] = gh.get_branch_protection(
-            repo, branch_name
-        )
-
-    return _BRANCH_PROTECTION_CACHE[repo][branch_name]
+    return cache[repo][branch_name]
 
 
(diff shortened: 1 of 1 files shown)


catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
```diff
@@ -173,38 +173,42 @@ def base_repo_name(pr: dict):
...
-        _BRANCH_PROTECTION_CACHE[repo][branch_name] = gh.get_branch_protection(
-            repo, branch_name
-        )
-
-    return _BRANCH_PROTECTION_CACHE[repo][branch_name]
+    return cache[repo][branch_name]
 
 
 def get_min_required_approvals(gh: GitHubAPI, pr: dict) -> int:
+    branch_protection_cache = defaultdict(dict)
```
sarayourfriend, line 186: This doesn't resolve the issue, unfortunately, at least not the detail I mentioned about using a cache per-DAG run for purposes of consistency within the run. Maybe that isn't important, but this also doesn't persist the cache within the DAG run.

`get_min_required_approvals` is called in a loop over all `urgent_prs` in the main task function. This cache is recreated for each of those PRs within the context of this function, and basically doesn't do any caching. Each time this function runs, the cache will be empty, and it will fetch the branch protections for the queries tested in each of the calls to `get_branch_protection` within this function each time, each of which use different branches for the query, and so do not re-use cached values.

Solving this problem requires a wider refactor of this DAG so that the cache can be shared within the DAG run but ephemeral outside of it. The easiest way that comes to mind to do that is to move these individual functions and the body of the task function into methods on a class, and store the cache as an instance attribute of the class. Then the task function should create a new instance of the class, and call whatever method encapsulates the logic currently in the task function.

That's short of re-writing this into multiple tasks with distributed state (including the cache).

sarayourfriend (changes requested):
Requesting changes because this solves the problem of caching for too long, but at the expense of essentially not caching at all. If that's acceptable, then we can just delete the caching code altogether so the code doesn't say it is doing any caching when in fact it isn't.

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
AetherUnbound: Gah, okay thanks for expounding on this. I didn't quite think through the steps when making the change here. Let me take another look at it and try to retain the cache across the DAG.

AetherUnbound: There's actually a simpler solution here which is just to move the cache one more level up! A small change within the primary function which doesn't require any significant refactoring or changes to tests. I added some logging locally to ensure the cache is being used appropriately, and it was 😄 Going with that approach for now!

sarayourfriend: Cool! I hadn't looked at it a lot and was worried it would require passing through to lots of functions (like we do with the `gh` instance), but glad it was that easy :grin:

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
```diff
@@ -214,6 +219,7 @@ def get_min_required_approvals(gh: GitHubAPI, pr: dict) -> int:
 @task(task_id="pr_review_reminder_operator")
 def post_reminders(maintainers: set[str], github_pat: str, dry_run: bool):
     gh = GitHubAPI(github_pat)
+    branch_protection_cache = defaultdict(dict)
```
sarayourfriend, line 222: Maybe worth adding a comment to explain why we cache within the context of the task rather than at the module level, even just a link to the issue if you don't want to reproduce the rationale, but I think something like this is simple enough to explain it... otherwise it's a little cryptic :sweat_smile: 

```suggestion
    # Build a new cache for each DAG run so that changes to repository branch settings
    # are reflected "by the next DAG run". Caching them at the run level also ensures
    # that all evaluations for pings happen with the same settings, preventing changes
    # during a run from causing PRs to receive different treatment, which could
    # produce confusing results we might interpret as a bug rather than just a delay.
    branch_protection_cache = defaultdict(dict)
```

sarayourfriend (approved):
Yep, great! One suggestion for an explanatory comment on the rationale behind the cache strategy. Adding the comment is a soft-blocker, but I don't have too strong an opinion on what precisely you write for it, so I'm approving now so that you can add the comment and then merge without needing another round of review.

stacimc (approved):
+1 for Sara's comment suggestion, but looks great 🥳

catalog/dags/maintenance/pr_review_reminders/pr_review_reminders.py
AetherUnbound: Love the suggestion, thanks!

PR #4748, opened 2024-08-13
Correct response handling for checking dead links

sarayourfriend:
Reverts WordPress/openverse#4746

I'll copy my explanation of why `raise_for_status` caused the issue from our incident report here:

> As [Staci] noted, we added `raise_for_status` in the check dead links HEAD request. This caused non 2xx responses to raise a ClientResponseError. These are caught with ClientError. But the except block in the function calls logger.error with the exception object. Adding raise_for_status incorrectly classified dead link responses as errors rather than expected responses that we wanted to catch. The except block is only written to handle unintentional errors, i.e., errors in making the request at the networking level, rather than the response we receive. That’s why it catches ClientError instead of specifically ClientResponseError. The latter is raised for “bad” status codes when raise_for_status=True. We don’t want that, so we shouldn’t have set raise_for_status=True, and we missed the significance of this change during review. Simply removing raise_for_status=True will resolve the issue so that any response from upstream has their status code used regardless of if it’s a “bad” one, and only errors in making the request will raise when calling head.

This PR removes that line and adds an explanation for how to understand the error handling of that particular method. It comes down to the fact that any status code from upstream is useful information for dead link checking, but broken networking is something we need to be made aware of in the error logs.

As noted in the incident report, the issue with multiple Sentry issues created for this has the same root cause addressed by https://github.com/WordPress/openverse/pull/4704, which we should also prioritise reviewing and getting merged.

### DIFF
--- a/api/api/utils/aiohttp.py
+++ b/api/api/utils/aiohttp.py
@@ -1,4 +1,5 @@
 import asyncio
+import time
 import weakref
 
 import aiohttp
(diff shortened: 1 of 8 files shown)


stacimc (approved):
Appreciate the comment as well 👍

sarayourfriend:
I've validated this in staging by doing some queries that resulted in dead link checks and confirming that status codes that are "bad" do not raise an error and are not subsumed into the `-1` status code:

```json
{
  "event": "Deleting broken image from results id=07085894-f870-4701-808d-df435464c4ca status=404 provider=flickr ",
  "timestamp": "2024-08-14T10:18:36.914741Z",
  "ip": "2.223.112.126",
  "request_id": "e6144182-9e36-410e-a9a8-89d68fcde809",
  "logger": "api.utils.check_dead_links",
  "level": "info",
  "filename": "__init__.py",
  "func_name": "check_dead_links",
  "lineno": 184
}
```

Note that status is 404 in the "event" key.

PR #4752, opened 2024-08-13
Always use only latest provider tags when reingesting

### LINKED ISSUE #4732: Change tag upsert strategy to drop old provider tags
## Problem

As part of the discussion in #4452, we've decided that our historical strategy of merging old and new provider tags when reingesting a work is problematic. To quote @stacimc in that discussion:

> Although I agree we should retain data even if we are initially skeptical of its usefulness (deleted/modified tags were likely low quality tags, as you point out), the issue to me remains privacy/ethical concerns with deliberately retaining this historical data that is intentionally changed at source. I can't think of a use case for this data that would avoid those issues.

To expand on what the ethical and privacy concerns are: Openverse has an ethical responsibility to represent works as they are represented by the upstream provider, and to make clear when we are intentionally augmenting the description of a work. To that end, if the set of tags change in a provider, removed or modified tags must be reflected in Openverse's dataset. For example, if a museum decides on a new way of describing a work, and previous tags were culturally insensitive, it is important for Openverse not to reproduce that insensitivity, especially as we are representing those tags as specifically from the provider. Similarly, considering a privacy perspective, if tags at the upstream source originally include privacy invading information (e.g., the name of a pictured individual, sensitive location information, etc), and the upstream source removes them, it's critical that Openverse also no longer retains those tags after reingestion.

Keep in mind that Openverse does not current reingest most of its data (namely for dated provider DAGs like Flickr, Wikimedia), so for the vast majority of works, we will still have the problem of potentially retaining stale/incorrect tag information from the provider. Future work may be planned to selectively reingest works on a periodic basis (e.g., works returned in search queries, works for which the metadata may actually be seen by individuals).

Furthermore, as Staci pointed out in the priorities meeting yesterday, every other piece of metadata follows a replace, rather than merge approach. Provider tags have been unique in this way, and there's no reason to maintain this, and ample reason to change it.

## Description

Change the jsonb_array column strategy to drop all existing provider tags. Non-provider tags (e.g., machine generated tags) must be retained. Existing provider tags should be dropped entirely in favour of incoming provider tags.

https://github.com/WordPress/openverse/blob/de0079d70b7759bdeb52702c26eca6474e084cc6/catalog/dags/common/storage/columns.py#L70-L78

The current strategy is written in a generic form, with column parenthesised, but I think we need to replace it with a tags-specific merge strategy, because the `provider` field on the tags is tags specific. Other jsonb array columns should use their own approach.

Something like the following might be a good starting point:

```python
def _merge_tags(*args) -> str:
    return dedent(
        """
        tags = COALESCE(
            (
                SELECT jsonb_path_query_array(
                    old.tags,
                    '$ ? (@.provider == "$provider")',
                    jsonb_build_object('provider', old.provider)
                ) || EXCLUDED.tags
            ),
            EXCLUDED.tags,
            old.tags
        )
        """
    ).strip()
```

Though I'm still confused about whether `EXCLUDED` is definitely the _new_ tags or something else (the name is confounding).

## Additional context

Blocks #4452.

sarayourfriend:
## Fixes

Fixes #4732 by @sarayourfriend 

## Description

Changes the tag upsert strategy to match the requirements described in the issue. Namely: provider tags must always reflect the latest tags from the provider, and never retain deleted or changed tags in our data.

The PR does this by adding a new column upsert strategy exclusively for the tags column which filters the provider tags out from the column's existing tags before merging it with the incoming tags. This has the effect of preserving all non-provider tags whilst simply ignoring any previously held provider tags in favour of always only using the incoming tags.

## Testing Instructions

CI and unit tests should pass, of course. Review the changes I made to the test and ensure they are clear at explaining the intention of how the tags merging should work, and that I've covered the edge cases.

You can also manually test. Run Airflow with `ov just catalog/up` and visit localhost:9090 and log in with airflow as the username and password.

Add the `INGESTION_LIMIT` variable through the Admin variable page (http://0.0.0.0:9090/variable/list/) and set it to a low number like 10.

Enable and let the `wordpress_workflow` DAG run (http://0.0.0.0:9090/dags/wordpress_workflow/grid). You will now have 10 WordPress Photo Directory results in your local data.

Run `ov just catalog/pgcli` and execute the following SQL to mock some enriched tags as well as tags that "used to be" in the provider (and which will disappear after reingestion):

```sql
update image set tags = tags || '[{"name": "deleted_upstream", "provider": "wordpress"}, {"name": "chicken", "provider": "magic_vision"}]'::jsonb where identifier in (select identifier from image where provider = 'wordpress' limit 1) returning identifier, tags;
```

Record the identifier returned by the query.

Then re-run the WordPress workflow DAG by manually running it (http://0.0.0.0:9090/dags/wordpress_workflow/trigger?origin=%2Fdags%2Fwordpress_workflow%2Fgrid). You do not need to change the configuration at all. Let it finish running, then in the pgcli, select the tags for the image you modified before (use the identifier returned by the `update` query). You should see the "chicken" tag retained (because it is from "magic_vision" rather than "wordpress" provider), but should no longer see "deleted_upstream" in the tags list, because the incoming tags when you reingested no longer included those. Confirm this is the case.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/tests/dags/common/loader/test_sql.py
```diff
@@ -1050,22 +1050,48 @@ def test_upsert_records_merges_tags(
     IMG_URL = "https://images.com/a/img.jpg"
     LICENSE = "by"
 
-    TAGS_A = json.dumps(
-        [{"name": "tagone", "provider": "test"}, {"name": "tagtwo", "provider": "test"}]
+    unchanged_provider_tag = {"name": "unchanged_provider_tag", "provider": PROVIDER}
+    deleted_provider_tag = {"name": "deleted_provider_tag", "provider": PROVIDER}
+    brand_new_tag = {"name": "brand_new_tag", "provider": PROVIDER}
+    shared_machine_tag = {"name": "shared_tag", "provider": "magic_vision_ai"}
+    incoming_shared_provider_tag = shared_machine_tag | {"provider": PROVIDER}
```
krysal, line 1057: A thorough set of tags to test, nice 💯

krysal (approved):
LGTM! It looks like it's missing a docs update but the relevant change makes sense.

AetherUnbound (approved):
Tested locally and was able to capture the behavior mentioned! I've updated the media props as well, which should hopefully let CI pass.

sarayourfriend:
Thanks! I had no idea the media properties would change in this, apologies for marking that checkbox N/A.

Issue #4717, opened 2024-08-05
Image table contains `.ogv` videos


obulat:
## Description

Wikimedia script ingests `.ogv` video files as images. We should filter them out and probably also drop the ones we already added from the catalog database.

## Reproduction

1. Go to https://openverse.org/image/2dc93cee-34d6-4b98-a76e-92ebc8d4ae73?q=phytoplankton+bloom
2. The main image is unavailable, as are most of the related images.
3. If you go to the source, you will see that this is actually a video of `.ogv` extension.

## Additional context

These images were noticed by @sarayourfriend in #4665

sarayourfriend:
The structured data for this result on Wikimedia shows the media type as `application/ogg`.

That's actually interesting, because I don't know how you would disambiguate that from an audio file if you don't have the extension in the URL!

https://commons.wikimedia.org/w/index.php?curid=59122955

At least in this case, we do have the direct file URL and can use the extension from it.

obulat:
> That's actually interesting, because I don't know how you would disambiguate that from an audio file if you don't have the extension in the URL!

In the Wikimedia API result, this media is probably marked as image in the `mediatype` key of the response:
https://github.com/WordPress/openverse/blob/23467c8d57b593a758aac43cd20211e49a927fdb/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L450

We [accept as image](https://github.com/WordPress/openverse/blob/23467c8d57b593a758aac43cd20211e49a927fdb/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L145) all media that is marked as `"BITMAP"` or `"DRAWING"`. It appears that some of the media marked with those values are of `application/ogg` type, so we should be filtering them out.

sarayourfriend:
It's good to note that "image", in some metadata standards (DublinCore, for example) includes both moving and still images. Something to keep in mind, especially when interfacing with GLAM institutions that may be using terms like that in their metadata.

I'm surprised wikimedia considers those bitmaps or drawings, though 🙂 bitmap doesn't seem to me as generic as "image", though I suppose it can be if interpreted loosely, which it already is anyway!

krysal:
Code changes are prepared in #4772, and I confirmed there are 45397 of these items in Wikipedia incorrectly classified.

```sql
openledger=> SELECT COUNT(*) FROM image WHERE source = 'wikimedia' AND url ~ '^.*.(ogv)$';
 count
-------
 45397
```

[...]

PR #4772, opened 2024-08-16
Filter `filetype` of Wikimedia's images

krysal:
## Fixes

Part of #4717 by @obulat 

## Description

Adds a validation step for image file types in the Wikimedia provider DAG. I considered making this validation global (in the MediaStore class), but we must be really sure of the file types that will be accepted as valid on each media type (maintaining the `EXTENSIONS` dictionary up to date or validated in some way). So, for now, I limited the scope to Wikimedia images, which have been the most problematic.

## Testing Instructions

Run the Wikimedia DAG and check the filetypes in the upstream database.

## Checklist

- [x] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog PRs) or the media properties generator (`ov just catalog/generate-docs media-props` for the catalog or `ov just api/generate-docs` for the API) where applicable.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
@@ -399,6 +399,24 @@ def test_extract_creator_info_handles_link_as_partial_text(wmc):
     assert expect_creator_url == actual_creator_url
 
 
+@pytest.mark.parametrize(
+    "url, media_type, expected",
+    [
+        ("https://example.com/image.jpg", "image", "jpg"),
+        ("https://example.com/image.JpeG", "image", "jpeg"),
+        ("https://example.com/image.Png", "image", "png"),
+        ("https://example.com/image.GIF", "image", "gif"),
+        ("https://example.com/image.ogv", "image", None),
+        ("https://example.com/image.xyz", "image", None),
+        ("https://example.com/audio.mp3", "audio", "mp3"),
+        ("https://example.com/audio.ogg", "audio", "ogg"),
+        ("https://example.com/audio.WAV", "audio", "wav"),
+    ],
+)
+def test_extract_file_type(wmc, url, media_type, expected):
+    assert wmc.extract_file_type(url, media_type) == expected
+
+
 def test_extract_license_info_finds_license_url(wmc):
     image_info = _get_resource_json("image_info_from_example_data.json")
     expect_license_url = "https://creativecommons.org/licenses/by-sa/4.0/"
--- a/catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
+++ b/catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
@@ -402,12 +402,15 @@ def test_extract_creator_info_handles_link_as_partial_text(wmc):
 @pytest.mark.parametrize(
     "url, media_type, expected",
     [
+        # Valid images
         ("https://example.com/image.jpg", "image", "jpg"),
         ("https://example.com/image.JpeG", "image", "jpeg"),
         ("https://example.com/image.Png", "image", "png"),
         ("https://example.com/image.GIF", "image", "gif"),
+        # Invalid (for our sake) images
         ("https://example.com/image.ogv", "image", None),
         ("https://example.com/image.xyz", "image", None),
+        # Valid audio
         ("https://example.com/audio.mp3", "audio", "mp3"),
         ("https://example.com/audio.ogg", "audio", "ogg"),
         ("https://example.com/audio.WAV", "audio", "wav"),
--- a/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
+++ b/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
@@ -115,6 +115,7 @@ from types import MappingProxyType
 import lxml.html as html
 
 from common.constants import AUDIO, IMAGE
+from common.extensions import EXTENSIONS
 from common.licenses import LicenseInfo, get_license_info
 from common.loader import provider_details as prov
 from providers.provider_api_scripts.provider_data_ingester import ProviderDataIngester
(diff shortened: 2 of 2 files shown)


catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
```diff
@@ -399,6 +399,24 @@ def test_extract_creator_info_handles_link_as_partial_text(wmc):
...
+        ("https://example.com/image.jpg", "image", "jpg"),
+        ("https://example.com/image.JpeG", "image", "jpeg"),
+        ("https://example.com/image.Png", "image", "png"),
+        ("https://example.com/image.GIF", "image", "gif"),
+        ("https://example.com/image.ogv", "image", None),
+        ("https://example.com/image.xyz", "image", None),
+        ("https://example.com/audio.mp3", "audio", "mp3"),
+        ("https://example.com/audio.ogg", "audio", "ogg"),
+        ("https://example.com/audio.WAV", "audio", "wav"),
+    ],
```
AetherUnbound, line 414: Nit: do you mind adding comments to separate these?
```suggestion
    [
        # Valid images
        ("https://example.com/image.jpg", "image", "jpg"),
        ("https://example.com/image.JpeG", "image", "jpeg"),
        ("https://example.com/image.Png", "image", "png"),
        ("https://example.com/image.GIF", "image", "gif"),
        # Invalid (for our sake) images
        ("https://example.com/image.ogv", "image", None),
        ("https://example.com/image.xyz", "image", None),
        # Valid audio
        ("https://example.com/audio.mp3", "audio", "mp3"),
        ("https://example.com/audio.ogg", "audio", "ogg"),
        ("https://example.com/audio.WAV", "audio", "wav"),
    ],
```

AetherUnbound (approved):
LGTM, seems reasonable to put this here! Your point about moving this to the `MediaStore` class is interesting...but I agree that it's worth targeting Wikimedia specifically in this case. Thanks for addressing this!

catalog/tests/dags/providers/provider_api_scripts/test_wikimedia_commons.py
krysal: Good notes 👍

sarayourfriend:
@krysal this almost seems like something that should be implemented for all DAGs? Maybe at the store level? Wondering what the use case is for a provider ingestion DAG to _not_ check this, if `filetype IS NOT NULL` (i.e., it is known). @WordPress/openverse-catalog, would this warrant a more abstract approach to handle this check for all providers?

obulat:
@sarayourfriend, we do extract URL extension to use as the file type in the `MediaStore` if the file type is `None`. However, we only return the file type if it's a valid type for the media type. So, for an `.ogv` file, this would return `None`, but would still use the file. This is why we do need to filter Wikimedia items in the script itself. For other providers, the URL might not have an extension, so returning `None` as the filetype should be continued.

For this PR, though, we should re-use the existing utility for extracting extensions from here: https://github.com/WordPress/openverse/blob/5e2e0e308a3f26cf87edc731287aaac8b17117d4/catalog/dags/common/extensions.py#L7

sarayourfriend:
If that's true then how does this PR address the issue of OGV files present in the images table? It returns `None` if the filetype is ogv (if the extract filetype is not in the supported extensions it doesn't return the filetype).

Doesn't that mean the MediaStore would just see None for the filetype and then continue including the file (treating it like when the URL doesn't have an extension)? Wouldn't this change actually make the data _worse_ because now we won't even see when the `filetype` is OGV without doing a pattern match query on the URL?

I guess I don't understand how this is intended to validate anything. It filters metadata out of the final image row without affecting whether that image is kept. In these cases we ostensibly _know_ the filetype (at least as much as any other place where we know the filetype based on the URL extension) but we're just ignoring it. Why? What's the ultimate goal of this PR? @krysal's description explains it just as "validation", but I don't understand to what end, if we aren't filtering these results entirely out of ingestion by the time we see they are not of a supported filetype.

obulat:
@sarayourfriend, I assumed that this PR returned `None` from `get_record_data` if the extracted extension is `None`. That's why I wrote a very long explanation of how Wikimedia script and `MediaStore` worked together.

But then I saw that this PR only extracts a valid extension from the URL, and does not drop the image if the extension is invalid.

@krysal, this PR should be updated to:
1. use the existing function for extracting extension (note: it needs to be updated to `.lower()` the URL, otherwise it won't work for capitalized extensions in some Wikimedia URLs)
2. return `None` from `get_record_data` if the extension could not be extracted. 

I think we should also log the URLs with the valid Wikimedia (`BITMAP` and `AUDIO`?) from which we could not extract a valid extension, to see what other extensions we might be missing.

AetherUnbound:
> I assumed that this PR returned `None` from `get_record_data` if the extracted extension is `None`. 

Ahh, I had made the same assumption. I agree with Olga's suggested approach here.

krysal:
I don't know how that "tiny" detail was overlooked 🙃 My bad! I created #4785 to follow up. @obulat, it uses an approach quite similar to what you describe, with improvements to the common function to get the file type. It needs to update more tests, but you all let me know if it's what is expected.

sarayourfriend:
Should we revert this, considering it is potentially actively creating worse metadata for all new Wikimedia ingestion runs, rather than actually doing any filtering or augmenting metadata in a way we could use?

krysal:
@sarayourfriend It's not doing anything much different that before so I wouldn't say it creating _worse_ metadata. If anything, it's detecting and logging when the DAG might be ingesting an unexpected type. In any case, it's being changed in #4785.

sarayourfriend:
It is actively excluding the file type column for works we do not wish to include. To me that is worse metadata, we have less information to work from to fix our Wikimedia data than we did otherwise.

But fine, we will fix it in more significant ways in #4785 and #4717.

Issue #4643, opened 2024-07-22
Determine which labels to exclude from Rekognition’s label set


AetherUnbound:
## Description

This will involve a manual process of looking through each of the [available labels for Rekognition](https://docs.aws.amazon.com/rekognition/latest/dg/samples/AmazonRekognitionLabels_v3.0.zip) and seeing if they match any of the criteria to be filtered. This process should be completed by two maintainers, and their list of exclusions discussed & combined. The excluded labels should then be saved in an accessible location, either on S3 or within the [sensitive terms repository](https://github.com/WordPress/openverse-sensitive-terms) as a new file. Consent & approval should be sought from two other maintainers on the accuracy of the exclusion list prior to publishing.

## Additional context

See [this section](https://docs.openverse.org/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.html#determine-excluded-labels) of the IP.

## Final exclusions

This is the list of exclusions we've determined, based on the discussion below.

<details> <summary> Exclusions </summary>

- Adult
- Alien
- Angry
- Arguing
- Baby
- Baby Crawling
- Baby Laughing
- Ballerina
- Beard
- Bishop
- Blonde
- Blue Hair
- Boy
- Bridal Veil
- Bride
- Bridegroom
- Bridesmaid
- Brown Hair
- Buddha
- Building Flooding
- Bun (Hairstyle)
- Car Back - Damaged
- Car Dent
- Car Front - Damaged
- Car Mirror - Broken
- Car Scratch
- Car Window - Broken
- Child
- Childbirth
- Corrosion
- Crucifix
- Crying
- Curly Hair
- Family
- Female
- Fireman
- Frown
- Girl
- Green Hair
- Happy
- Hippie
- Hoe
- Home Damage
- Lady
- Laughing
- Male
- Man
- Mohawk Hairstyle
- Mold
- Mold Damage
- Mustache
- Newborn
- Pain
- Pink Hair
- Pope
- Prayer
- Prayer Beads
- Priest
- Red Hair
- Romantic
- Roof Damage
- Rust
- Sad
- Senior Citizen
- Shouting
- Shrine
- Smile
- Surprised
- Teen
- Temple
- Termite Damage
- Tribe
- Triumphant
- Window - Broken
- Woman
- Yawning

</details>

## Corrections

For various reasons (removing gender, capitalization correction, etc.) we plan on mapping the following terms to the corrected values.

<details> <summary>Corrections</summary>

[...]

PR #4784, opened 2024-08-19
Add specifics of Rekognition tag filtering to implementation plan

AetherUnbound:
## Description

This was prompted by a discussion in #4643.

This PR adds some clarity around the Rekognition tag filtering step, including adding an explicit step to implement the filtering using an inclusion-based list after the tags are inserted. See the discussion linked above for rationale, and the text changed for what's described.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
+++ b/documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
@@ -198,9 +198,10 @@ In order to incorporate accomplish the goals of this plan, the following steps
 will need to be performed:
 
 1. [Determine which labels to exclude from Rekognition's label set](#determine-excluded-labels)
-2. [Preemptively filter the Rekognition tags](#preemptively-filter-rekognition-tags)
+2. [Preemptively filter all Rekognition tags](#preemptively-filter-rekognition-tags)
 3. [Generate and insert the new Rekognition tags](#insert-new-rekognition-tags)
-4. [Filter and assess the existing Clarifai tags](#filter-clarifai-tags)
+4. [Filter the Rekognition tags based on the criteria determined in step 1](#filter-rekognition-tags)
+5. [Filter and assess the existing Clarifai tags](#filter-clarifai-tags)
 
 ## Step details
 
@@ -257,6 +258,12 @@ in bulk in a similar manner in the future, having a clear and repeatable process
 for doing so will make those operations easier down the line. It also allows us
 to test the insertion process locally, which feels crucial for such a
 significant addition of data.
+
+```{note}
+This step will add _all_ of the available Rekognition tags to the catalog. The
+tags with an insufficient confidence value or that were determined to be excluded
+will be filtered out of the API database during the data refresh.
+```
 
 #### Context
 
@@ -506,6 +513,28 @@ For local testing, a small sample of the Rekognition data could be made
 available in the local S3 server
 [similar to the iNaturalist sample data](https://github.com/WordPress/openverse/blob/82282a00abdaed21e8381052a874d8ab9a4f7e0a/catalog/compose.yml#L98-L101).
 
+### Filter Rekognition tags
+
+Once the new Rekognition tags have been inserted into the catalog, we will want
+to remove the blanket-filter for all Rekognition tags that was put in place in
+the [preemptive filtering step](#preemptively-filter-rekognition-tags). We will
+also need to add the logic for filtering out the tags that were determined
+should be excluded in the
+[determine excluded labels step](#determine-excluded-labels).
+
+For all machine-generated labels, we will employ an inclusion-based filtering
+process. This means that we will only filter out labels that match the list of
+approved labels, which prevents labels that are unreviewed from appearing in the
+downstream dataset. This can be added to the `alter_data` step of the data
+refresh (see #4684) and would only be applied to tags where the `provider` was
+not the record's `provider`.
+
+We will also add a step for recording if a label was not in the inclusion list
+_and_ if it did not exist in a full list of all reviewed labels from the
+provider. These "unreviewed" labels can be surfaced as part of the data refresh,
+so maintainers can review them and decide if they should be included in the
+inclusion list.
+
 ### Filter Clarifai tags
 
 While this project seeks to add new magine-generated labels to the database, we
@@ -527,16 +556,19 @@ will want to filter the existing Clarifai tags until we can perform the same
 analysis on the set of available tags
 [as will be done for the Rekognition ones](#determine-excluded-labels). This can
 be done using the same steps described for the Rekognition filtering, based on
-the status of this project and #3925.
+the status of this project and #3925. An alternative approach could be to use
+the "unreviewed" label approach described in the
+[filter Rekognition tags](#filter-rekognition-tags) section to surface the
+Clarifai tags for review.
 
 Once the filtering is in place, we can construct an exhaustive set of Clarifai
 labels and determine exclusions for that provider using the approach
-[described above](#label-criteria). Then the Clarifai label exclusions can be
-added to #4541 in the same way Rekognition's are added and the blanked exclusion
-for all tags from that provider can be lifted. These exclusion lists could be
-combined into a single filtering step, or we could have individual filter lists
-based on the label provider. My preference is former, since that way the single
-list serves as a more exhaustive exclusion list.
+[described above](#label-criteria). Then the Clarifai label inclusions can be
+added to the `alter_data` step (#4541) in the same way Rekognition's are added
+and the blanked exclusion for all tags from that provider can be lifted. These
+exclusion lists could be combined into a single filtering step, or we could have
+individual filter lists based on the label provider. My preference is former,
+since that way the single list serves as a more exhaustive exclusion list.
 
 ## Dependencies
 
@@ -711,3 +743,6 @@ xdf.groupby("image_uuid").median("Confidence").median()
 
 - 2024-07-25 - (#4662) Clarified policy around initially included demographic
   labels to ensure they're reviewed on a case-by-case basis.
+- 2024-08-19 - () Added a note about the inclusion-based filtering process and
+  the "unreviewed" label capture, and a note to insert all available Rekognition
+  data and filter it afterwards.
--- a/documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
+++ b/documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
@@ -520,7 +520,8 @@ to remove the blanket-filter for all Rekognition tags that was put in place in
 the [preemptive filtering step](#preemptively-filter-rekognition-tags). We will
 also need to add the logic for filtering out the tags that were determined
 should be excluded in the
-[determine excluded labels step](#determine-excluded-labels).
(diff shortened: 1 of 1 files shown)


documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
```diff
@@ -506,6 +513,28 @@ For local testing, a small sample of the Rekognition data could be made
...
+also need to add the logic for filtering out the tags that were determined
+should be excluded in the
+[determine excluded labels step](#determine-excluded-labels).
+
+For all machine-generated labels, we will employ an inclusion-based filtering
+process. This means that we will only filter out labels that match the list of
+approved labels, which prevents labels that are unreviewed from appearing in the
+downstream dataset. This can be added to the `alter_data` step of the data
+refresh (see #4684) and would only be applied to tags where the `provider` was
+not the record's `provider`.
```
sarayourfriend, line 530: I think it would be generally helpful to clarify that the filtering of the tags is in relation to their use in search?

```suggestion
For all machine-generated labels, we will employ an inclusion-based filtering
process. This means that we will only include labels in search that match the list of
approved labels, which prevents labels that are unreviewed from appearing in the
downstream dataset. We will add this to the `alter_data` step of the data
refresh (see #4684) and will only be applied to labels with the provider corresponding to the include list's provider.
```

The inclusion lists should be per-provider, right? The methodology of these labels are not all equal and we may not find the same value or benefits from one provider to the other. Especially considering we know _nothing_ about how or where the Clarifai dataset comes from (which really gives me pause _viz_ our continued use of them at all... but that's a separate question).

Should also clarify that the comparisons used for filtering must be case insensitive, due to some labels receiving orthographic changes upon ingestion into the catalogue. Mostly to clarify the reason here, than the fact of the particular implementation detail. But also, we've made changes to the contents of the labels. So that means for example that we actually need to modify the inclusion list from the provider anyway, like to change "fireman" to "firefighter", "RAM memory" to "RAM"? Those won't match compared to the provider list even if case insensitive.

Does that mean we need to maintain those orthographic translations in multiple places, in the inclusion list and in the code itself? Or should we keep them in a mapping that can be used to retrieve our "fixed" version of the label as well as the provider's original label from the fixed version? Then the ingestion transformation can use the "provider -> fixed" mapping, and the data refresh filtering can use the "fixed -> provider" mapping to retrieve the original provider label from any fixed ones (and just using the label as is if a mapping doesn't exist, of course).

documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
```diff
@@ -506,6 +513,28 @@ For local testing, a small sample of the Rekognition data could be made
...
+
+We will also add a step for recording if a label was not in the inclusion list
+_and_ if it did not exist in a full list of all reviewed labels from the
+provider. These "unreviewed" labels can be surfaced as part of the data refresh,
+so maintainers can review them and decide if they should be included in the
+inclusion list.
```
sarayourfriend, line 536: Just a nit about can/should/must etc. "can" reads very much as an "option" to me, rather than a certain "will be".

```suggestion
provider. These "unreviewed" labels should be surfaced as part of the data refresh,
so maintainers can review them and decide if they should be included in the
inclusion list.
```

sarayourfriend (changes requested):

documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
AetherUnbound: The inclusion list should cover all providers IMO, I'm not sure what benefit we would get from maintaining provider-specific inclusion lists. We don't do this for sensitive terms, we've simply applied a blanket across all of our data regarding term sensitivity. Given the paradigm of "include + all reviewed" labels, this would actually prevent us from having to exclude the Clarifai labels wholesale in the same way we're currently doing for Rekognition, since the only Clarifai labels that would pass through the alter data steps are labels which we've already reviewed and approved. We could then go in and capture all the "unreviewed" labels from Clarifai, and either add them to the reviewed list _or_ the inclusion list. Having per-provider lists would allow us finer-grain control, but re-reviewing label sets that may have significant cross over seems like adding unnecessary extra work on our part. I think we can treat this in the same way as the sensitive terms list.

Yes, I can add more information about the case insensitive comparison. And for the cases of "fireman" or "ram memory", those two would go into the "all reviewed labels" list, but only "firefighter" and "ram" would be in the "included labels" list. I don't think we need to maintain a mapping in perpetuity within the code.

sarayourfriend: Sensitive terms are different though, the dataset they are compared against is by definition "all English language words" because we don't know the extent of all terms present in title, description, and provider tags. Tags from a machine vision model are not that way, they are bound in their scope of terms, and have some particular methodology (or probably at best "pattern") for deriving those terms.

I agree there will be significant overlap and for some classes of terms (indeed, maybe for all the terms we have now) we can say we will not include from any provider (like gender terms like man and woman). But one provider may have labels we think are okay based on their methodology and not okay based on others. One model might confidently mislabel people as animals whereas another may not, distinguishing between those would be useful. That's especially important because we are specifically using an inclusion list rather than an exclusion list. We're in some way saying "we definitely want these labels", applying that as a blanket to all models doesn't seem wise.

For example, we are including the label "Barbie" based on it being about the toy, not a descriptor of a person. What if the Clarfai model also has that label but it isn't part of the "toys" category of labels? It's a fine distinction in this case and one we would want control over; or else accept we will always be working from the lowest common denominator of all relevant models' capabilities.

I don't feel too strongly here, though. We don't know the contents of the Clarifai labels so it's all theoretical. If we are trying to be careful (I believe we are), I don't think it's a big ask to review each label from each source independently of each other. We may even come to new conclusions about labels generally by doing so. I mean, _especially_ because we do not know the provenance of the Clarifai labels, literally _at all_. We have no information about where they came from, how they ended up in the catalogue, or what model was used to generate them. Why wouldn't we want to review all of them, even if we just end up including the same set? What does it cost us other than a bit of time? It's even a good opportunity to have different people than you, me, and Krystle involved in reviewing those labels, perhaps even someone from outside the core maintainers.

But, if you feel strongly there is no issue here, I won't push for separate lists beyond what I've said here.

> And for the cases of "fireman" or "ram memory", those two would go into the "all reviewed labels" list, but only "firefighter" and "ram" would be in the "included labels" list. I don't think we need to maintain a mapping in perpetuity within the code.

Shouldn't we maintain the mapping in code at least for provenance? I would hate to be the person in 2 years who tries to figure out how Openverse ended up with labels ostensibly from Rekognition which Rekognition does not itself say it generates. And how would we even do the mapping without it being in code? Wouldn't it be in the label ingestion DAG, which we would theoretically use as at least the building point if not exactly for ingesting future labels from Rekognition?

Is the idea that we would transform the labels in the inclusion list _and_ in the all labels list? We'll need to document the transformations there too, in that case, for the same reason. Per-provider lists seems especially important if we're doing this kind of mapping and then not explicitly documenting the intention.

Maybe I am being too cautious. To be fully transparent, I am terrified by the use of these labels and despite the apparent benefits still maintain extreme reservations about all of this. Perhaps that is motivating an overly cautious approach from me on these. Input from someone else would be good to have here.

krysal: I see the process described in the proposal as a good incremental approach. If we weren't filtering Clarify tags before by other criteria than the accuracy (and the general excluded terms) then we can continue with it. Mainly to limit the scope of the PR and the project itself, which goes around Recognition. If we see specific problems with the Clarify tags then we can prioritize that work you're proposing @sarayourfriend. This is not to say that your suggestion on keeping a list by provider would not be helpful, I agree that it will give us more control, but I think it's beneficial to start with Recognition and continue to see some progress.

krysal (approved):
LGTM! The ideas emerging from the conversation are interesting and worth continuing exploring them. We can continue at the same time.

documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
sarayourfriend: We have already added issues to evaluate Clarifai tags to the project's scope, for what its worth. Past history of not doing due diligence with these machine generated labels (or not documenting it, if it was done, we have no way of knowing), isn't a reason to continue ignoring the need for it.

https://docs.openverse.org/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.html#filter-clarifai-tags

> If we see specific problems with the Clarify tags then we can prioritize that work you're proposing

Sure. We've already committed to looking at the tags anyway. It isn't a big thing to add later anyway, unless I'm missing something about this that would make it more complex than a for loop over a list of provider lists :shrug:

AetherUnbound: I _do_ think it incurs a greater maintenance burden, both by way of review and code, to be maintaining separate lists. If we find there is significant enough semantic or cultural distinction between the tags that different providers use, then we can always move to maintaining separate lists down the line. Right now we are only concerned with Rekognition and Clarifai labels, since we don't have any other clear plans for incorporating other sources of machine-generated labels. As such, I feel comfortable taking this approach knowing it's simpler on the maintenance end, _and_ it's a reversible decision should we choose to change it down the line.

> Shouldn't we maintain the mapping in code at least for provenance?

The mapping has been codified, along with the other labels we're excluding and the full list of reviewed labels, in #4795. It will then also be present in the DAG code which is ingesting the labels from the Rekognition dataset and inserting them to the catalog.

The transformed label would be in the inclusion list, and both it _and_ the original label would be present in the all reviewed labels list.

sarayourfriend: > It will then also be present in the DAG code which is ingesting the labels from the Rekognition dataset and inserting them to the catalog.

Okay. I didn't understand that would be the case. You also said:

> I don't think we need to maintain a mapping in perpetuity within the code.

> and both it and the original label would be present in the all reviewed labels list

Okay, I see. Including it in the all reviewed _alongside the original_, means it won't get flagged as unreviewed. That bit, including _both_ in the reviewed list, was the part I was missing to understand how it would be consistent.

sarayourfriend (approved):
I didn't realise I was still set requesting changes, nothing to add beyond the discussion, no changes to the text for it from me :+1:

stacimc (approved):
This LGTM -- I agree strongly with the initial approach laid out here; any additional per-provider considerations discovered in the future should be easily iterated upon, without introducing complexity prematurely. Interesting discussion, thanks for the clarifications.

documentation/projects/proposals/rekognition_data/20240530-implementation_plan_augment_catalog_with_rekognition_tags.md
AetherUnbound: Ahh, apologies for communicating that in a confusing way! Including alongside the original was the point I was trying to convey 🙂

Issue #2421, opened 2023-06-16
Pdfs in images from Wikimedia


obulat:
## Description

We have some pdf file data in the image table. It passes the dead link validation, but shows up as "Image Not Available" in the frontend.

If cleaning up the catalog is quick, we should do that. Then these items will be fixed after the image refresh.

I'm not sure if we want to also add a hot-fix in the API layer to prevent files with `.pdf` in the URL from showing up in the results. If we added a file type check to the dead link filter, we would stop sending unnecessary requests to the source. The status is currently "200" because the file _is_ available, even though it doesn't show up in the UI.

## Reproduction

See: https://api.openverse.engineering/v1/images/dffcffbe-3980-4cc7-bc61-804851cb9121/
and https://openverse.org/image/dffcffbe-3980-4cc7-bc61-804851cb9121

## Additional context

@cc @krysal, I think you have done a similar cleanup in the catalog, haven't you?

krysal:
> @krysal, I think you have done a similar cleanup in the catalog, haven't you?

Yes, to remove audio files from the image table. But this is an interesting case, the link actually does return an image. It opens a debate: should we consider accepting PDF images or just remove them? Can Wikimedia provide a different format more traditionally accepted as an image?

AetherUnbound:
Thanks for surfacing this Olga - I think if we _can_ get an alternative from Wikimedia we should use that instead (the frontend shows that as "png", interestingly). But otherwise, we should probably remove PDFs from the image media set, I don't think most folks would consider PDFs as images 😅

obulat:
> Thanks for surfacing this Olga - I think if we _can_ get an alternative from Wikimedia we should use that instead (the frontend shows that as "png", interestingly). But otherwise, we should probably remove PDFs from the image media set, I don't think most folks would consider PDFs as images 😅 

I know why the frontend "considers" it PNG: we load the "Image not available" placeholder, and replace the image type with the placeholder type 😀 I'll open an issue for that.

@krysal , I didn't notice that it was considered an "image" in Wikimedia, that's interesting! We might need to add a check to the Wikimedia provider script if the pdf images are marked as images in the API responses, and we decide to not accept such pdfs as "image"s. I agree with @AetherUnbound that the users wouldn't expect an image to be a pdf.

PR #4785, opened 2024-08-20
Raise exception for misclassified items in MediaStore and get filetype from Wikimedia's API

krysal:
## Fixes

Related to #4717 by @obulat
Follow up #4772

## Description

Here, I broadened the number of file types the existing `commons.extensions.extract_filetype` function can guess using the `mimetypes` library. It is based on the same logic of guessing it from the URL. The `MediaStore` class will raise an error if it finds an item misclassified within the types we support.

The Wikimedia Commons DAG now gets the `filetype` from the API, giving us more accurate data in a straightforward way. You can take a look at their [MimeMap](https://doc.wikimedia.org/mediawiki-core/master/php/MimeMap_8php_source.html) file shared by @obulat below in comments. This helped complete the list of mime types and aliases.

## Testing Instructions

Check tests and confirm everything makes sense and the types that matter are covered. Optionally, run the `wikimedia_commons_workflow` DAG to verify it continues to work correctly.

## Checklist

- [x] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog PRs) or the media properties generator (`ov just catalog/generate-docs media-props` for the catalog or `ov just api/generate-docs` for the API) where applicable.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -448,6 +455,22 @@ def get_value_by_names(key_value_list: list, prop_names: list):
...
+    @staticmethod
+    def get_filetype(url, media_type) -> str | None:
+        """
+        Extract the filetype from extension in the media url.
+
+        Returns only if the media type guessed matches the expected media type.
+        """
+        filetype, extracted_media_type = extract_filetype(url)
+        if extracted_media_type != media_type:
+            # Prevent ingesting files like .ovg as images
```
obulat, line 467:
```suggestion
            # Prevent ingesting files like .ogv as images
```

catalog/dags/common/extensions.py
```diff
@@ -1,12 +1,30 @@
...
-    if possible_filetype in EXTENSIONS.get(media_type, {}):
-        return possible_filetype
-    return None
+def extract_filetype(url: str) -> tuple[str | None, str | None]:
+    """
+    Extract the filetype from a media url extension.
+
+    Returns only if the media type guessed is a supported type.
+    """
+    if mime_type := mimetypes.guess_type(url)[0]:
```
obulat, line 21: The [`types_map`](https://github.com/python/cpython/blob/8edfa0b0b4ae4235bb3262d952c23e7581516d4f/Lib/mimetypes.py#L432) in Python mimetypes library is missing some types that we use, including `.mid`, `.ogg` and `.ogv` types. This is also causing the CI tests to fail.
 
We can probably add new types after initializing, using `add_type` function:
 https://github.com/python/cpython/blob/8edfa0b0b4ae4235bb3262d952c23e7581516d4f/Lib/mimetypes.py#L86
 
 I think we should also add logging to identify extensions that don't return a supported filetype here to find any other extensions that are not supported but available in our data.

obulat (changes requested):
Nice test coverage! Could we also add a case that an upper-cased file type is converted to a lower cased one (when it overrides the type extracted from the url)?

My blocking requests here are:
- adding missing types to the mimetypes
- adding logging to identify any other extensions/types we are missing.

catalog/dags/common/storage/media.py
```diff
@@ -304,17 +303,25 @@ def _format_raw_tag(self, tag: str) -> dict:
...
+        if filetype is not None:
+            filetype = filetype.lower()
+            return FILETYPE_EQUIVALENTS.get(filetype, filetype)
+
+        filetype, extracted_media_type = extract_filetype(url)
+        if extracted_media_type is not None and extracted_media_type != self.media_type:
+            raise ValueError(
+                f"Extracted media type `{extracted_media_type}` does not match "
+                f"expected media type `{self.media_type}`."
+            )
```
sarayourfriend, line 323: I suppose these docs should be updated to reflect the new exception case:

https://github.com/WordPress/openverse/blob/0ea4c6cbfa6037c5883207279e891ecbd3e00c83/catalog/dags/common/storage/media.py#L115-L116

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -448,6 +455,22 @@ def get_value_by_names(key_value_list: list, prop_names: list):
...
 
+    @staticmethod
+    def get_filetype(url, media_type) -> str | None:
+        """
+        Extract the filetype from extension in the media url.
+
+        Returns only if the media type guessed matches the expected media type.
+        """
+        filetype, extracted_media_type = extract_filetype(url)
+        if extracted_media_type != media_type:
```
sarayourfriend, line 466: Is this something that can be generically applied to all providers?

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -317,10 +317,17 @@ def get_record_data(self, record):
...
             return None
 
+        try:
+            filetype = self.get_filetype(url, valid_media_type)
+        except ValueError:
+            logger.warning(
+                f"Skipping record due to media type mismatch. Media type: {valid_media_type}, url {url}"
+            )
+            return None
+
```
sarayourfriend, line 327: Kind of a nit, but this uses exceptions for control flow, and rather generically at that. A `ValueError` could occur for potentially any reason inside `self.get_filetype` and we aren't checking the error string to confirm that its happened for the reason we think. This could swallow a serious bug in `self.get_filetype`, for example, if `extract_filetype` started throwing a ValueError due to `self.get_filetype` giving it invalid inputs.

This warning should go in the place where we actually validated the fact that the media type of the work didn't match the media type of the ingestion; and then `get_filetype` can just return None, and this function and check the return value the exact same way we've done for `valid_media_type`, `license_info`, `url`, and `foreign_landing_url`.

sarayourfriend (commented):
What is the relationship between these proposed changes and it's media type detection and the `extract_media_type` function?

https://github.com/WordPress/openverse/blob/0ea4c6cbfa6037c5883207279e891ecbd3e00c83/catalog/dags/providers/provider_api_scripts/wikimedia_commons.py#L474-L489

This PR seems to duplicate the functionality of that existing method. Should it replace it? Why doesn't the existing method already solve this problem to begin with? Is there a bug in that code that could inform a wider understanding of what's gone wrong in our Wikimedia ingestion?

catalog/dags/common/storage/media.py
krysal: The `filetype` is still not required generally. I'm just ensuring it raises when a type conflict is found so we know which DAG is ingesting those and adjust it, like the case with Wikimedia. If

krysal:
> Nice test coverage! Could we also add a case that an upper-cased file type is converted to a lower cased one (when it overrides the type extracted from the url)?

@obulat Sure! I can add those. I'm aware that several tests need updates. Before continuing with the rest of the changes, I was waiting for confirmation that this approach aligns with what is expected, namely, **adding the filetype check to all the media ingestion and preventing Wikimedia from ingesting wrong rows in the image (or audio) table**.

Since your comments are on adding more log records, I assume you agree with this solution, and those are good notes. Thanks!

---

> What is the relationship between these proposed changes and it's media type detection and the extract_media_type function?

@sarayourfriend That function is specific to Wikimedia and detects which kind of item (an image or audio track) we're dealing with when extracting the details. 

> This PR seems to duplicate the functionality of that existing method. Should it replace it? Why doesn't the existing method already solve this problem to begin with?

Since we've found [PDFs](https://github.com/WordPress/openverse/issues/2421) and OGV videos in the image table, we cannot rely on Wikimedia's classification to prevent ingesting those items. Maybe this is already fixed on their end, I don't know, but the only way we can be sure is to double-check on our end.

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
krysal: That is my intention in including it in the `MediaStore` class (which you also commented on). The difference is that there, it should raise an Exception (so we're aware of the problem), and here on Wikimedia, it should simply skip the miscategorized record. That's my view, at least. Do you picture another approach to this?

catalog/dags/common/storage/media.py
sarayourfriend: I mean more that "raises an error" part. The function raises in an error in more than just the conditions related to the required fields now.

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
sarayourfriend: > here on Wikimedia, it should simply skip the miscategorized record

Using a ValueError in this comment to communicate that makes it difficult to understand that that is the intention; just follows into the other comment I shared about how this control flow is managed compared to the other checks.

sarayourfriend:
If we don't trust wikimedia's media type detection, then why use it at all? It duplicates the functionality, at least in name, if not in implementation. Keeping two places that "check the media type" in two completely different ways with no clear relationship between the two, and when one of them is apparently an unreliable source of truth doesn't make sense to me. If detecting the media type based on file type is reliable, then we should use that. If it's unreliable (like if somehow some URLs do not have extensions), then we need to document that case and make it clear that the fallback for media type detection relies on the other method. Combining them into a single routine that detects the file format if available and the media type (either based on the file format if it was available or based on Wikimedia's metadata) would make things _much_ clearer.

obulat:
> If we don't trust wikimedia's media type detection, then why use it at all?

I spent some time trying to understand why the Wikimedia's media type detection didn't work. Only to realize that it worked perfectly, but we had some data ingested before [we started filtering](https://github.com/cc-archive/cccatalog/pull/453) images by the Wikimedia's `mediaType`. And most of the non-image media items that were ingested before that PR were removed from the database in [[Bug] Non-image content is being stored in the image table (original #438)](https://github.com/WordPress/openverse/issues/1796#top).

So, my request for change here is to revert all changes, the Wikimedia script was filtering out the `.ogv` files perfectly well.

catalog/dags/common/extensions.py
```diff
@@ -1,12 +1,32 @@
...
+    # Image extensions
+    "jpeg": "jpg",
+    "tif": "tiff",
+    # Audio extensions
+    "midi": "mid",
 }
 
+mimetypes.add_type("audio/midi", ".mid")
+mimetypes.add_type("audio/midi", ".midi")
+mimetypes.add_type("audio/x-matroska", ".mka")
```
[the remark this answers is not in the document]
krysal: Querying the API DB returns nothing.

<details>
<summary>query</summary>

```sql
SELECT * FROM audio WHERE filetype = 'mka' OR url ILIKE '%.mka' LIMIT 1;
```
</details>

I have yet to come across a track like this to have an example. Do you have reservations about it? It's included in Wikimedia's mediaType.

catalog/dags/common/extensions.py
krysal: It's curious. The .mid/.midi extensions are included, and the script works well on Mac but not in the container. It seems it depends on the OS's local files. 

https://github.com/python/cpython/blob/8edfa0b0b4ae4235bb3262d952c23e7581516d4f/Lib/mimetypes.py#L592-L593

In any case, it's good to have a workaround!

krysal:
> If we don't trust wikimedia's media type detection, then why use it at all? It duplicates the functionality, at least in name, if not in implementation. Keeping two places that "check the media type" in two completely different ways with no clear relationship between the two, and when one of them is apparently an unreliable source of truth doesn't make sense to me. If detecting the media type based on file type is reliable, then we should use that. If it's unreliable (like if somehow some URLs do not have extensions), then we need to document that case and make it clear that the fallback for media type detection relies on the other method. Combining them into a single routine that detects the file format if available and the media type (either based on the file format if it was available or based on Wikimedia's metadata) would make things _much_ clearer.

@sarayourfriend That makes sense. The double-check also felt odd to me. Thanks a lot to @obulat for clarifying that the Wikimedia classification is correct. That simplified the PR a lot 🙌  and I think I addressed all the comments. This is ready for another review. Thanks for the patience, folks. Let me know your thoughts!

catalog/dags/common/extensions.py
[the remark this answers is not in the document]
sarayourfriend: [Matroska contrainers are not supported by Chrome, Firefox, or Safari](https://jellyfin.org/docs/general/clients/codec-support/#container-compatibility). We would need to extract the audio from the container and potentially convert it to support playing them. We should probably ignore these files and create an issue to follow up on adding some kind of support for them if we think it is valuable to include them. If there are currently 0 of them, then it's probably not worth the effort.

catalog/dags/common/extensions.py
```diff
@@ -1,12 +1,44 @@
...
+        self.expected_filetype = expected_filetype
+        if not message:
+            message = (
+                f"Extracted media type `{self.actual_filetype}` does not match "
+                f"expected media type `{self.expected_filetype}`."
+            )
+        super().__init__(message)
+
+
+def extract_filetype(url: str) -> tuple[str | None, str | None]:
```
sarayourfriend, line 31: Is the return type more correctly put as `tuple[str, str] | tuple[None, None]`? It's hard to tell from the code, but it doesn't look like there are any cases where it would be mixed `str, None` or `None, str`, but this signature would include those possibilities.

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
```diff
@@ -320,7 +320,7 @@ def get_record_data(self, record):
         creator, creator_url = self.extract_creator_info(media_info)
         title = self.extract_title(media_info)
         filesize = media_info.get("size", 0)  # in bytes
-        filetype = self.extract_file_type(url, valid_media_type)
+        filetype, _ = extract_filetype(url)
```
sarayourfriend, line 323: I still don't really understand why we should throw away the media_type value from this and keep the `::extract_media_type` method. It would be a lot clearer if these calls were next to each other and an explanation in comment were added to clarify their relationship to one another. Even knowing the context of this PR, I would struggle to explain the reasoning behind continuing to use the Wikimedia media types.

Especially because `::extract_media_type` _already extracts the mimetype Wikimedia has for the file_, which should in theory be more reliable than extracting it from the URL extension.

I really strongly feel these should not be separate methods. Either we should just use the Wikimedia version (and extract the "filetype" we want from it) or only use `extract_filetype` and ignore the Wikimedia metadata (the latter of which feels very wrong to me, like we are ignoring the authority). At worst we should fall back to `extract_filetype` if for some reason the Wikimedia metadata doesn't make sense.

Two separate methods for this is very confusing and at minimum warrants a comprehensive explanation. But really I think we should consolidate these into a single routine. The code is really confusing as it is now (and already was before, FWIW!).

sarayourfriend (changes requested):
The changes are an improvement, but I still think this is not the right way to proceed.

We already have the file type and media type metadata from Wikimedia themselves. We are ignoring the file type information, however, which leads me to read the code and think it must not be reliable (and somehow extracting from the URL is better). If that is the case, then why use Wikimedia's metadata to determine the media type, but not the file type? Why not just use our own methods for determining the media type as well as the file type?

But then, we need to explain why we are ignoring Wikimedia's metadata, and why we believe the URL is more reliable for this.

In any case, two separate methods that have identical results (with very different methods) and no explanation for why we are doing this in the code, is a recipe for confusion in the future. Unless there is a clear explanation for the current approach of using Wikimedia's metadata for the media type but not the file type and using our bespoke approach for the file type but not the media type, we consolidate these into a single routine using a single approach for finding the information we need. If there is a clear explanation, it must be included in the code, and the code organised to emphasise and clarify that relationship.

catalog/dags/providers/provider_api_scripts/wikimedia_commons.py
krysal: > I still don't really understand why we should throw away the media_type value from this and keep the ::extract_media_type method. It would be a lot clearer if these calls were next to each other and an explanation in comment were added to clarify their relationship to one another. Even knowing the context of this PR, I would struggle to explain the reasoning behind continuing to use the Wikimedia media types.

I changed the assumption in that Wikimedia classification is correct after [Olga's comment](https://github.com/WordPress/openverse/pull/4785#issuecomment-2301909971), and then the media_type from `extract_filetype` theoretically will always be the same than the returned from the Wikimedia specific method, `::extract_media_type`.

> Especially because ::extract_media_type already extracts the mimetype Wikimedia has for the file, which should in theory be more reliable than extracting it from the URL extension.

That Wikimedia method only extracts the media type, whether the record is an image or audio, not so the mime type. But this gave me an idea! It would be a lot easier if we could get the mime type [directly from the API](https://commons.wikimedia.org/w/api.php?action=help&modules=query%2Bimageinfo), and actually we can! I think that would be more reliable. 

I made the changes for this locally, just need to modify tests. Wait for it 🤞

sarayourfriend: I see, okay. I misunderstood the code in `extract_media_type`. I thought the check, `if media_type in image_mediatypes:`, was looking at the MediaWiki list of image mime types, but it's _media_ types. Sorry for the confusion I've caused.

I'm glad there is a way to get the mime type from Wikimedia, though. Hopefully it will be reliable and help consolidate this logic.

catalog/dags/common/extensions.py
krysal: That seems correct though I kind of prefer the shorter general signature still.

catalog/dags/common/extensions.py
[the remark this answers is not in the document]
krysal: Providing a wide range of formats should be valuable to users; even if they cannot reproduce them in Openverse right now, they should be able to download them from the source. The funny is we don't have any of this currently.

```sql
openledger=> SELECT COUNT(*) FROM audio WHERE filetype = 'mka' OR url ILIKE '%.mka';
 count 
-------
     0
(1 row)
```

So, for now, we're safe. We could add a check in some DAG to alert if this type is detected.

sarayourfriend (approved):
LGTM! Glad there was a nice way to simplify the Wikimedia DAG in the end, and the MediaStore changes are great.

As a side note: it is wild to me that we have `filetype` as a single word, despite being two words, while also having `meta_data` as two words, despite itself only being a single word. I always have a moment of "oh yeah, weird" when I read these column names :joy: The consequences of old decisions :sweat_smile:

catalog/dags/common/storage/media.py
```diff
@@ -113,8 +116,9 @@ def clean_media_metadata(self, **media_data) -> dict | None:
         - add `provider`,
         - add default `category`, if available.
 
-        Raises an error if missing any of the required fields:
-        `license_info`, `foreign_identifier`, `foreign_landing_url`, or `url`.
+        Raises an error if missing any of the required fields: `license_info`,
+        `foreign_identifier`, `foreign_landing_url`, or `url`. Or if an extracted
```
obulat, line 120:
```suggestion
        `foreign_identifier`, `foreign_landing_url`, or `url`); or if an extracted
```

catalog/dags/common/storage/media.py
```diff
@@ -113,8 +116,9 @@ def clean_media_metadata(self, **media_data) -> dict | None:
         - add `provider`,
         - add default `category`, if available.
 
-        Raises an error if missing any of the required fields:
-        `license_info`, `foreign_identifier`, `foreign_landing_url`, or `url`.
+        Raises an error if missing any of the required fields: `license_info`,
```
obulat, line 119:
```suggestion
        Raises an error if missing any of the required fields (`license_info`,
```

catalog/dags/common/storage/media.py
```diff
@@ -113,8 +116,9 @@ def clean_media_metadata(self, **media_data) -> dict | None:
...
+        `foreign_identifier`, `foreign_landing_url`, or `url`. Or if an extracted
+        media type does not match the media type of the class.
```
obulat, line 121: I don't have a suggestion here, but this part of the sentence is confusing. "extracted media type" refers to what we extract from the "filetype" in the provider script, or from the URL extension, right? So, if, for example, the API returns `jpg` as `filetype` or if the URL ends with `.jpg` (when the API does not return `filetype`) for an audio item?

catalog/dags/common/storage/media.py
```diff
@@ -304,17 +308,22 @@ def _format_raw_tag(self, tag: str) -> dict:
 
     def _validate_filetype(self, filetype: str | None, url: str) -> str | None:
         """
-        Extract filetype from the media URL if filetype is None.
+        Extract filetype from the media URL if filetype is None and check
+        that it corresponds to the media type of the class, while also unifies
```
obulat, line 312: It's better to split this into two sentences, because there's a mismatch now (first part of the sentence is imperative, "extract", and the second is not, "unifies")

catalog/dags/common/extensions.py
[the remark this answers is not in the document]
obulat: I agree with providing all available formats on the API level. Since it is an audio, we should add a check for this format to the Nuxt app, to display a message similar to what we currently do _for ogg on iOS_, IIRC, that this format cannot be played on the user's device. If we ever get any media with this format 😆

obulat (approved):
I love more clarity on the file and mime types that this PR adds 👍 
Added several suggestions for more clarity in the docs comments.

catalog/dags/common/storage/media.py
krysal: From the URL extension! That's a good question. The comment was not clear, so I updated it with more details.

catalog/dags/common/storage/media.py
krysal: Fixed the grammar! Good eye 😆 Thanks!

krysal:
Thanks for the reviews, you both!

PR #4088, opened 2024-04-10
Upgrade eslint-plugin-playwright to get new checks; fix small checkbox bug

sarayourfriend:
## Description

These are changes I made while investigating https://github.com/WordPress/openverse/issues/4087. The issue turned out not to be actually an issue, but these new tests are useful, and I found a different bug anyway, which is fixed in this PR. I also upgraded us to the latest playwright ESLint plugin, and I will need to fix the many new warnings this has produced. I did the upgrade originally hoping that https://github.com/playwright-community/eslint-plugin-playwright/blob/main/docs/rules/prefer-web-first-assertions.md would catch `toHaveAttribute("checked", "checked")`, but of course, that's a different assertion than `toBeChecked()`. Very easy to get wrong! I'm sure I learned this at some point, but it's a trivial thing to forget.

There was also an issue with the TypeScript version we use in the application being outside the supported range of the older TS ESLint plugin we used. I've updated the TS ESLint plugin, which also required updating ESLint, which in turn required updating the other ESLint plugins. In the end I decided to just update all the ESLint stuff using `pnpm -r up 'eslint*'`.

## Testing Instructions

Run storybook using `just p frontend storybook` and visit the checkbox story: http://localhost:54000/iframe.html?id=components-vcheckbox--default-story

Confirm you can still click to toggle the checkbox on and off.

Run the new storybook tests (or look at them in CI) and confirm the assertions pass. If you run them locally, you can remove `pointer-events-none` from the checked icon to see the bug I fixed.

Run `pnpm run eslint` and see that there are no errors. CI will also confirm this.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [N/A] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
[...]

PR #4804, opened 2024-08-23
Replace hard-coded strings in Playwright tests

### LINKED ISSUE #619: Update e2e tests to not rely on specific text copy, instead use translation tokens
## Problem

Take this test as an example:

https://github.com/WordPress/openverse-frontend/blob/26af972f90f3f9b8081131d8c61a2aefb8329991/test/e2e/search.spec.js#L29

If we try to run it in another language that is translated, it will fail. But we shouldn't have to write unique tests for every language we'd like to test. Likewise, if the copy is changed then the test will fail. This is even worse for languages outside of the token language of English because the copy is controlled externally and can change at any time, thus causing the e2e tests to fail if we write them against another language (which we should, for example, write them for RTL languages and other non-latin alphabet languages).

## Description

Create some method for the e2e tests to only rely on static tokens for the selectors rather than actual page copy that can change arbitrarily (when compared to the behavior, which should stay the same unless intentionally changed and should not change due to a simple copy update).

Using the translation token might be way around this but I'm not sure how that would work for RTL languages or testing non-latin alphabet languages.

## Alternatives

Use `data-test-id` style selectors instead of copy or tokens. Maybe that's better when there aren't any other accessibility based selectors to use.

## Implementation

obulat:
## Fixes

Fixes #619 by @sarayourfriend

## Description

This PR audits the Playwright tests to not use hard-coded strings, replacing them with i18n translations

It also replaces some `page.locator` with `page.getByRole` selectors, and uses better selectors for checkboxes.

## Testing Instructions

The CI should pass, and the changes should make sense. The Playwright tests should not have hard-coded strings left.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/playwright/e2e/image-detail.spec.ts
+++ b/frontend/test/playwright/e2e/image-detail.spec.ts
@@ -16,7 +16,6 @@ const imageObject = {
   id: "e9d97a98-621b-4ec2-bf70-f47a74380452",
   provider: "flickr",
 }
-
 const goToCustomImagePage = async (page: Page) => {
   // Test in a custom image detail page, it should apply the same for any image.
   await page.goto(`image/${imageObject.id}`)
@@ -26,7 +25,7 @@ const goToCustomImagePage = async (page: Page) => {
 const errorPageLocator = (page: Page) =>
   page.getByRole("heading", {
     level: 1,
-    name: /The content you’re looking for seems to have disappeared/,
+    name: new RegExp(t("404.title")),
   })
 
 test.describe.configure({ mode: "parallel" })
@@ -77,7 +76,7 @@ test("sends GET_MEDIA event on CTA button click", async ({ context, page }) => {
 
   await goToCustomImagePage(page)
 
-  await page.getByRole("link", { name: /get this image/i }).click()
+  await page.getByRole("link", { name: t("imageDetails.weblink") }).click()
 
   const getMediaEvent = analyticsEvents.find((event) => event.n === "GET_MEDIA")
 
--- a/frontend/test/playwright/e2e/report-media.spec.ts
+++ b/frontend/test/playwright/e2e/report-media.spec.ts
@@ -32,10 +32,9 @@ export const openReportModal = (page: Page) =>
 
 const checkOption = async (page: Page, option: ReportReason) => {
   const name = t(`mediaDetails.contentReport.form.${option}.option`)
-  if (!(await page.getByRole("radio", { name }).isChecked())) {
-    // `.select()` doesn't work for radio buttons because Playwright
-    // incorrectly states that the element is covered by the icon SVG.
-    await page.getByText(name).click()
+  const radio = page.getByRole("radio", { name })
+  if (!(await radio.isChecked())) {
+    await radio.check()
   }
 }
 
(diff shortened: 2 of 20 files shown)


frontend/test/playwright/e2e/image-detail.spec.ts
```diff
@@ -26,7 +25,7 @@ const goToCustomImagePage = async (page: Page) => {
 const errorPageLocator = (page: Page) =>
   page.getByRole("heading", {
     level: 1,
-    name: /The content you’re looking for seems to have disappeared/,
+    name: new RegExp(t("404.title")),
```
obulat, line 28:
```suggestion
    name: t("404.title"),
```

frontend/test/playwright/e2e/report-media.spec.ts
```diff
@@ -23,8 +25,19 @@ test.describe.configure({ mode: "parallel" })
...
+
 export const openReportModal = (page: Page) =>
-  page.click('text="Report this content"')
+  page.getByRole("button", { name: reportButtonLabel }).click()
+
+const checkOption = async (page: Page, option: ReportReason) => {
+  const name = t(`mediaDetails.contentReport.form.${option}.option`)
+  if (!(await page.getByRole("radio", { name }).isChecked())) {
+    // `.select()` doesn't work for radio buttons because Playwright
+    // incorrectly states that the element is covered by the icon SVG.
```
sarayourfriend, line 37: If there is a bug report or something to link to to explain why Playwright _incorrectly_ reports this is the case, it would be helpful to document it here. Otherwise, it seems to easy to write Playwright's error off as "incorrect" when there may actually be an underlying issue we can solve. IIRC there is a bug report about something related to SVGs and click events.

I vaguely remember something like this being the case when working on https://github.com/WordPress/openverse/pull/4088, and it ended up being a bug in our pointer-events configuration for the checkboxes, leading to this change: https://github.com/WordPress/openverse/pull/4088/files#diff-9cab0d134035b3db0d42b453e18b3fff640f27e4155601385449cbe17b7788f8R50

When I was working on that, I remember coming across a Playwright bug report that seemed to indicate there was a general problem with Playwright's "click" and SVGs and that the issue seemed to imply it was just broken, rather than there being a real fix for it (in this case, passing through pointer events). In the end there was a fix in our code that did actually resolve a real (potential) problem. Apparently click events "happened" to work, but didn't technically always appear like they would work (which could affect accessibility tools maybe?).

If either is the case here, if it is a Playwright bug that we cannot do anything to fix, or if there is something we can change about the radio buttons to fix it, like there was for the checkboxes, then we should include that here explicitly.

sarayourfriend (changes requested):
LGTM. Just one small requested change to expand on the comment about Playwright's behaviour. It's a blocker because of the potential for error or confusion with it, but otherwise this is great. Love all the changes to `getByRole` (and similar)!

frontend/test/playwright/e2e/report-media.spec.ts
obulat: I thought there was a bug because I was adding `pointer-events-none` to the `VIcon.vue` component, and it was still getting the click event. Only to realize that the element in the report component is `VSvg`! 🤦 

Added `pointer-events-none` to `VRadiomark` and removed the comment [4cf211e](https://github.com/WordPress/openverse/pull/4804/commits/4cf211e2de9a1f910feae1274987e8ca9d8ad33a)

sarayourfriend (approved):
Nice! :rocket:

PR #4823, opened 2024-08-27
Update Nuxt and deps to v3.13.0; fix server dir location

### LINKED ISSUE #4820: Nuxt 3 update broke the healthcheck route
## Description

On the Nuxt 3 branch, the healthcheck route doesn't work/exist. This caused the last 2 staging deploys to fail. I couldn't notice this because there was a flaky Playwright test.

We should revert the branch. In follow up PRs, we need to
- add a Playwright test for the healthcheck route
- find a fix for the server route

zackkrida:
## Fixes

Fixes #4820 by @obulat 

## Description

This PR fixes the underlying issue with the serverDir, the healthcheck route, and Nuxt 3.13.0. Here we un-revert the dependency upgrades, add the E2E test for the healthcheck from #4822, and move the frontend `/server` directory to its default location in `/src/server`.

Prior to Nuxt 3.13.0, Nuxt was accidentally overwriting the user-supplied serverDir property in Nuxt.config.ts, and I suspect we were accidentally relying on that bug in some way which was then rectified in the switch to 3.13.0. 

## Testing Instructions

CI should pass, and the healthcheck can be viewed locally at `/healthcheck` after running `ov just frontend/run dev`. 

## Checklist

- [x] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (approved):
Approving ass this is a critical change.
However, I would prefer to correctly set the `serverDir`, maybe using the relative path as described here (the [PR that updated it](https://github.com/nuxt/nuxt/pull/28249/files)):

https://github.com/nuxt/nuxt/blob/74ef96d096d0045fd00fccd880c1dbbfea51e22a/packages/schema/src/config/common.ts#L154

This would be more compatible with Nuxt 4, where the default src directory is `/app`, and `/server` is on the same level with it:

https://nuxt.com/docs/getting-started/upgrade#new-directory-structure

zackkrida:
@obulat I actually went with the default _because_ the Nuxt 4 codemod will automatically move it for us if we're already adhering to the default location. So despite not following the Nuxt 4 convention _yet_, we're following the expected Nuxt 3 convention which allows for easier migration when the time comes.

PR #4831, opened 2024-08-28
Update appearance of sensitive image thumbnails

### LINKED ISSUE #4777: Improvement over image component
As per the [design](https://github.com/WordPress/openverse/issues/4563#issuecomment-2243368772) introduced in #4563 for an enhanced version of browsing results with sensitive images, the image component requires some changes.

## Description

[Both versions](https://www.figma.com/design/GIIQ4sDbaToCfFQyKMvzr8/Openverse-Design-Library?node-id=4050-9020&m=dev) of the component documented in the Design Library have the following changes:

### Image large
* Blur layer over image has the `bg-blur` variable
* In hover and focus, the component shows the text "Sensitive content" inside. Center and bottom aligned

image large variant in Figma with specs

### Image small
* Blur layer over image has the `bg-blur` variable
* The text "Sensitive content" replaces the CC license icons
* In resting state, the label has the `text-secondary` variable applied
* In hover and focus states, the label has the `text` variable applied

image small variant in Figma with specs

The alias mentioned above go as (copied from [this comment](https://github.com/WordPress/openverse/issues/4563#issuecomment-2244785363)): 

#### Colors
* `white-88`: `#FFFFFF` at 88%
* `black-72`: `#000000` at 72%

#### Alias

```
# Light theme

:root {
	bg-blur: var(white-88);
}

# Dark theme

:root {
	bg-blur: var(black-72);
}
````

## Additional context

The tokens applied are according to #3592 project

### PR DESCRIPTION
**Pull request by dhruvkb**
## Fixes

Fixes #4777 by @fcoveram 

## Description

This PR
- adds a new background token "blur" that corresponds to white-88 or black-72 depending on the theme
  - removes an incorrect, and redundant, use of `bg-blur`
- updates the image cell to show the text "Sensitive content" and an icon indicating that the image has been blurred because of sensitivity

## Testing Instructions

0. Check out the PR and run the Nuxt server.
1. Go to `/preferences` and check the "fake_sensitive" preference.
2. Perform a search for something.
3. Verify that a number of results have been blurred and show the "hide" icon.
4. Hover or focus over these results to reveal the bottom-centered "Sensitive content" note instead of the license info.
5. Switch to a mobile view.
6. Verify that the results have the "hide" icon and show the "Sensitive content" note in place of the license info.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida (changes requested):
- I don't think we can nest `<div>` in an `<a>`, so perhaps it would be better to wrap the icons and label in `<span>`s
- The text on hover is inheriting an underline from the parent `<a>` tag
- The icon is the wrong color in dark mode (looks like there is a fill hardcoded on the svg path)

The updated blurs look gorgeous!

obulat (changes requested):
Nice! ✨ 
Two problems I found:
1. There is a 1px non-blurred stripe at the bottom of the image:
Screenshot 2024-08-28 at 10 06 57 PM
2. The eye icon in the dark theme is not visible, should be of a different color.

dhruvkb:
> I don't think we can nest `<div>` in an `<a>`

@zackkrida, I thought so too, but I just learned that in HTML5 `<a>` [now allows](https://stackoverflow.com/a/1828032) `<div>`s. Nonetheless, switched to `<span>` in 3e057ba.

> There is a 1px non-blurred stripe at the bottom of the image

@obulat I tried both Firefox and Chrome but I am not able to reproduce the 1px line at the bottom, can you share your browser info?

zackkrida (dismissed):
Just two more small pieces of feedback, one we likely have to ask @fcoveram about.

1. I think we should remove the `title` attribute from the anchor tag, as the tooltip feels redundant:
Screenshot from 2024-08-29 07-28-01

2. The light mode overlay color is very overpowering, and removes much of the blurred quality and the underlying image:

> With
Screenshot from 2024-08-29 07-36-07

> Without: 
Screenshot from 2024-08-29 07-36-01

To the point that it almost looks like a plain grey background. I wonder if it's possible to adjust that?

fcoveram:
To your points:
1. Agree. It is redundant.
2. You are right. I lean on keeping the blur layer in a light tone to then shift it to dark in dark mode. We could try `#ffffff` at 80%, 72%, and 64%, and see which one feels better. Once we decided the % value, I update the Figma variable.

How does that sound?

obulat:
> @obulat I tried both Firefox and Chrome but I am not able to reproduce the 1px line at the bottom, can you share your browser info?

This must have something to do with the size calculations in the image grid. It only happens on the image grid where the sizes are variable. I could reproduce it in Chrome and Safari. Here's a Safari screenshot. Lines at the bottom are not so common, but the corners often have a different-color dot. I don't want to block on this because I think it won't be that noticeable.

Safari screenshot, see the first image in the row:

Screenshot 2024-08-30 at 5 49 36 PM

Chrome screenshot, look at the bottom row:

Screenshot 2024-08-30 at 5 47 21 PM

dhruvkb:
Dropped the tooltip in 4d4c81e.

@fcoveram would you like to see some screenshots at the different opacity levels?

@obulat I am still not able to see the lines. I'll keep trying to get them to show up, and will fix them in post.

dhruvkb:
Update: I got the line to show up! It seems like a strange rendering issue because the dimensions and position of both elements is exactly the same upto 5 decimal places.

Screenshot 2024-08-30 at 8 58 01 PM

fcoveram:
> @fcoveram would you like to see some screenshots at the different opacity levels?

That sounds great. Thanks @dhruvkb

dhruvkb:
image

@fcoveram these are 88% (current), 80%, 72%, 64% and no overlay respectively. As @zackkrida mentioned the current version shows almost no features of the underlying image, which only start to appear at much lower opacity (even 64% might be too much).

dhruvkb:
I will merge this PR once @fcoveram's preferred opacity level is implemented.

zackkrida:
@dhruvkb I think 72% might be the best option, while we await Francisco's opinion :laughing: Thank you for showing the different options.

fcoveram:
I agree that 72% looks good.

I simulated having an image with pure black area beneath the sensitive content text to see if it meet a11y requirements, and it seems safe to use. Here is a screenshot

Screenshot of image component with contrast requirements for accessibility notes on the right

So +1 to go with 72%

PR #4833, opened 2024-08-29
Add filtered index and promotion steps to data refresh

### LINKED ISSUE #4149: Add remaining steps to the Data refresh DAGs
## Problem

Blocked by #4148 

This issue tracks adding the final steps of the data refresh to the new DAGs.

## Description

The final steps will be added to the DAGs:

- Create and Populated Filtered Index: this should be implemented as a reusable TaskGroup. This work can already be implemented using our existing Elasticsearch tasks and replaces [this function](https://github.com/WordPress/openverse/blob/226aed0890a19ace1e0b54c1e784c86e9f26b4cb/ingestion_server/ingestion_server/indexer.py#L437).
- Reapply Constraints and Promote Table: these SQL queries can be performed with the PostgresHook, and replaces [this function](https://github.com/WordPress/openverse/blob/226aed0890a19ace1e0b54c1e784c86e9f26b4cb/ingestion_server/ingestion_server/ingest.py#L340).
- Promote Index and Delete Index: these can be implemented using our existing Elasticsearch tasks, and replaces [these](https://github.com/WordPress/openverse/blob/226aed0890a19ace1e0b54c1e784c86e9f26b4cb/ingestion_server/ingestion_server/indexer.py#L311) [functions](https://github.com/WordPress/openverse/blob/226aed0890a19ace1e0b54c1e784c86e9f26b4cb/ingestion_server/ingestion_server/indexer.py#L377).

## Additional context

See [this section](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html#add-remaining-steps-to-the-data-refresh-dags) of the IP.

stacimc:
## Fixes

Fixes #4149 by @stacimc 

## Description

This PR adds the final steps to the data refresh DAG:
* create and promote the filtered index
* remap indices from the media table to the temp table
* remap constraints from the media table to the temp table
* promote the temp table & delete the original
* promote the 'main' and filtered indices

Screenshot 2024-08-30 at 4 46 40 PM

This was much less straightforward than I imagined and this PR is fairly large as a consequence (although there are tests, and an `es_mapping.py` configuration file that's just copied from the ingestion server adding to the line count). I will attempt to break it into smaller pieces if reviewers like; however I'll point out that it's much easier to manually test in its _current_ state, because you can actually run the entire data refresh (even multiple times in a row).

While working on this PR I also noticed several small issues that I fixed at the same time. I have left comments at the relevant code changes to make these easier to spot:
* I changed the `create_index` steps to create the new index according to a hard-coded index config, rather than basing it off the config of the current index (see comment inline for reasoning)
* I fixed a bug where index calculations were off when multiple data refreshes were run concurrently
* I fixed *several* bugs with error handling and dependency chains, where the failure of a task would not cause presumed downstream tasks to skip. See commit messages
* I fixed a bug in the ingestion server with the way constraints are mapped. See [commit message](https://github.com/WordPress/openverse/pull/4833/commits/4d04e3b9e61897d5a8674b27866c1a914143f8ef)

I've tried to make commit messages and comments in the code as descriptive as possible, but please ping me if clarification is needed anywhere.

## Testing Instructions

Add the following to your `catalog/.env`:
```
AIRFLOW_CONN_SENSITIVE_TERMS="http://catalog_indexer_worker:8003/static/mock_sensitive_terms.txt"
```

Run `ov just recreate` to start with fresh sample data.

Run an image and an audio DAG for a few minutes to get some new records. Wait until you see that a couple 100 lines have been written in the `pull_data` logs, then mark the task as a success. I used Cleveland and Jamendo.

### Establish a baseline

Go to https://app.elasticvue.com/cluster/0/indices and verify you see the four initial indices (`image-init`, `image-init-filtered`, `audio-init`, `audio-init-filtered`), and note the number of records in each. Now run `ov just api/pgcli` and run `describe image;` and `describe audio;`. Save the results somewhere.

### Run the new data refreshes and compare

Now enable the `staging_audio_data_refresh` and `staging_image_data_refresh` DAGs. They should both complete successfully. Once done, check out elasticvue again. You should see that the `-init` indices have all been dropped, and replace with new indices with a uuid (example `audio-f61ab0e757d244ab9abeca0a0613d9e6`, `audio-f61ab0e757d244ab9abeca0a0613d9e6-filtered`, and so on). You should have *more* records in each of these new indices than in the initial ones. To be extra safe, you can go check the logs of the provider DAGs you ran to see how many records were ingested; that's the number of extra records you should see in the non-filtered indices.

In pgcli, again run `describe image;` and `describe audio;`. Compare the results to what you saw earlier; pay particular attention to the indices and constraints. They should be identical, including the index names.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


catalog/dags/data_refresh/create_and_populate_filtered_index.py
```diff
@@ -0,0 +1,116 @@
+"""
+# Create and Promote Index
+
+This file contains TaskGroups related to creating and populating the filtered Elasticsearch indices
+as part of the Data Refresh.
+
+TODO: We'll swap out the create and populate filtered index DAG to use this instead
+of the one defined in the legacy_data_refresh.
```
stacimc, line 8: Noting that there is a separate issue for this, (and moreover the filtered index is going away).

catalog/dags/data_refresh/create_index.py
```diff
@@ -30,17 +31,14 @@ def create_index(
...
-    index_config = es.get_index_configuration_copy.override(
-        task_id="get_index_configuration"
-    )(
-        source_index=data_refresh_config.media_type,
-        target_index_name=temp_index_name,
+    # Create a new index
+    es.create_index(
+        index_config={
+            "index": temp_index_name,
+            "body": index_settings(data_refresh_config.media_type),
```
stacimc, line 38: This change ensures that the data refresh always creates a new index based off the hard-coded index configuration in `es_mapping.py`, rather than by copying the configuration of the existing index.

This is how the ingestion server does it currently, and was just a miss in the previous PR. It's necessary to ensure that temporary or experimental changes to the indices (for example, created using our `create_new_es_index` DAGs or similar) are not carried forward in the data refresh without making the changes permanent via a code change/PR.

catalog/entrypoint.sh
```diff
@@ -58,8 +58,8 @@ while read -r var_string; do
   # set the environment variable
   export "$var_name"="$new_value"
 
-  # only include Slack airflow connections
-done < <(env | grep "^AIRFLOW_CONN_SLACK*")
+  # only include Slack airflow connections and the Sensitive Terms connection
+done < <(env | grep "^AIRFLOW_CONN_SLACK*\|AIRFLOW_CONN_SENSITIVE_TERMS")
```
stacimc, line 62: This change is necessary to allow testing with the real sensitive terms list when desired, since it uses https.

indexer_worker/indexer_worker/api.py
```diff
@@ -116,6 +117,7 @@ def create_api():
     _api.add_route("/healthcheck", HealthcheckResource())
     _api.add_route("/task", IndexingJobResource(task_tracker))
     _api.add_route("/task/{task_id}", TaskStatusResource(task_tracker))
+    _api.add_static_route("/static", (Path(".") / "static").absolute())
```
stacimc, line 120: The indexer worker serves the mock sensitive terms list, similar to how the ingestion server currently does so. This is okay because in local development, the indexer worker is always running, and in production we use the *real* sensitive terms list rather than the mock one. It is helpful to have the mock list served this way to make the local `filtered_index` steps behave as similarly as possible to the way it does in production, just using a different endpoint.

stacimc:
@sarayourfriend I added you as an additional reviewer because you expressed interest in this PR (and your feedback is always appreciated 😄) but don't feel obligated!

catalog/dags/data_refresh/promote_table.py
obulat, line 1: This is such a great example of the fact that the productivity should not be measured by the line count :) A couple of lines changed needed so much debugging and explanation. Thank you for great detailed notes, as always. Great to have the local environment close to the production.

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -274,6 +279,24 @@ def drop_connection(worker_conn: str):
...
 
+@task(trigger_rule=TriggerRule.NONE_FAILED)
+def report_reindexing_status():
+    """
+    Fail if the direct upstream tasks, which perform the actual reindexing,
+    fail. Otherwise simply pass.
+
+    This is necessary because we have some tasks after the reindex step
+    which always run even on reindexing failure (to drop the connection
+    and terminate the worker instance). However in the even of failures,
```
sarayourfriend, line 290:
```suggestion
    and terminate the worker instance). However in the event of failures,
```

catalog/dags/data_refresh/remap_table_indices.py
```diff
@@ -0,0 +1,137 @@
...
+from airflow.utils.trigger_rule import TriggerRule
+
+from common.sql import fetch_all, run_sql
+from data_refresh import queries
+
+
+logger = logging.getLogger(__name__)
+
+
+class TableIndex(NamedTuple):
```
sarayourfriend, line 20: Question: is the usage of `NamedTuple` instead of dataclasses due to a limitation in Airflow? I seem to recall limitations in task return types. Otherwise, do you have a preference for NamedTuples over dataclasses? And if so, I'd be curious to know the reason :slightly_smiling_face:

indexer_worker/indexer_worker/api.py
sarayourfriend: If we wanted to do it a different way (to avoid adding this functionality to the indexer worker), we could put the file into minio, and still retrieve it over HTTP. [According to this guide about using minio to host a website](https://gist.github.com/harshavardhana/f05b60fe6f96803743f38bea4b565bbf) it shouldn't be too difficult to do later on.

For the sake of simplicity, though, would it be better if the DAG just swapped the location it reads the sensitive terms from based on an additional environment variable? Like:

```py
if USE_MOCK_SENSITIVE_TERMS:
  return ["running", "water", "bird"]
else:
  # retrieve the regular list
```

I know it's not as seamless as just swapping out the HTTP connection string in the environment file, but it avoids a bit of this hoop-jumping required to make the DAG code "seamless". I guess really we've just hidden the seam somewhere else by doing it this way :stuck_out_tongue: (Also: because I wrote the original code for this in the ingestion server, I can say I think it was a bit cheeky/too clever to do it this way and wish I'd done this if/else there to begin with!)

catalog/dags/data_refresh/create_and_populate_filtered_index.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    sensitive_terms = HttpOperator(
+        task_id="get_sensitive_terms",
+        http_conn_id=SENSITIVE_TERMS_CONN_ID,
+        method="GET",
+        response_check=lambda response: response.status_code == 200,
+        response_filter=response_filter_sensitive_terms_endpoint,
+        trigger_rule=TriggerRule.NONE_FAILED,
+    )
```
sarayourfriend, line 92: Did you mean for this to use the `get_sensitive_terms` function above? It looks like the `HttpOperator` is identical in both. If we need both this version and the factory function, can you add a comment explaining the difference? Otherwise, consolidating them would be good.

sarayourfriend (commented):
Far out! I haven't tested this locally, just read over the code and process. Very cool! It's so exciting to have this laid out in the graph view, it really helps show the complexity of the data refresh, while also giving a clearer narrative to the different steps. I'm really glad we went forward with this project to remove the ingestion server.

If you'd like me to test this out locally, I'm happy to do so as well, just let me know. Otherwise, I'll leave it to the catalogue reviewers to do that part of it. For my part, this is very cool and exciting.

indexer_worker/indexer_worker/api.py
stacimc: We couldn't use the HttpOperator if we wanted to do that, or else we'd have to add branching tasks to get around selecting the right task per environment. Not a big deal but I'd prefer to keep the operators clean. We _could_ use minio, something similar to what @AetherUnbound did in #4836 maybe 🤔

catalog/env.template
```diff
@@ -78,6 +78,11 @@ AIRFLOW_CONN_DATA_REFRESH=http://ingestion_server:8001
 # server.
 AIRFLOW_CONN_STAGING_DATA_REFRESH=http://ingestion_server:8001
 
+# Connect to sensitive terms list used for filtered index creation in the data refresh.
+# For local development, the mock terms list is served by the catalog indexer worker. Change to the
+# full production URL in order to use the production sensitive terms list.
+AIRFLOW_CONN_SENSITIVE_TERMS="http://catalog_indexer_worker:8003/static/mock_sensitive_terms.txt"
```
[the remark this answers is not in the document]
stacimc: It's not required, since we can set this through the Admin UI (as we have with several other connections/variables). We could make an infrastructure change to set it as an environment variable though; actually, I'm not clear how we've drawn the line between which variables should be configured where 🤔 What do you think?

catalog/dags/data_refresh/create_and_populate_filtered_index.py
stacimc: Ah yep, that's a miss -- `get_sensitive_terms` was from a previous implementation and used to be more complicated, as well. That should be deleted.

catalog/dags/data_refresh/remap_table_indices.py
stacimc: I believe anything json serializable can be passed through XCOMs, if I'm not mistaken a dataclass would work. No strong preference beyond following convention in the project, and given that we don't require anything more complex. I don't think the pros/cons of either option are particularly significant in this use case.

catalog/env.template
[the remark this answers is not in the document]
sarayourfriend: Using the Admin UI sounds good, I believe that's in line with decisions we've made in the past (https://github.com/WordPress/openverse-infrastructure/issues/215). I forgot that environment variables aren't necessary for configuring these. I realised that I didn't personally have a clear picture of environment variables vs Airflow variables/connections, and that some of our code muddies that distinction... and presents problems for advancements in our Airflow configuration, management, and execution. I've opened #4863 to address those inconsistencies.

I've started [a discussion here](https://github.com/WordPress/openverse/discussions/4865) to work towards an approach more in line with the rest of our applications (using infrastructure as code).

indexer_worker/indexer_worker/api.py
sarayourfriend: > We couldn't use the HttpOperator if we wanted to do that, or else we'd have to add branching tasks to get around selecting the right task per environment

Oh :confused: We can't call the HttpOperator conditionally in a Python `@task` operator? I figured something like this would be possible:

```py
@task
def get_sensitive_terms(**context):
    if USE_MOCK_SENSITIVE_TERMS:
        return ["running", "water", "bird"]
    else:
        http_op = HttpOperator(...)
        return http_op.execute(**context)
```

obulat:
I ran the steps from the testing instructions.

The audio refresh ran first. The numbers in the ES (seen in ElasticVue) increased, but the `get_after_record_count` returned the old counts.
Then, the image refresh ran. The numbers in the ES did not change at all, even though the DAG seemed to have finished successfully

`wait_for_reindexing_task` finished successfully, but logs an error stating that you cannot pickle a module i in the end. [This SO answer](https://stackoverflow.com/questions/57560813/airflow-falls-with-typeerror-cant-pickle-module-objects-when-task-returns-kwa) suggests that something is wrong with the kwargs object.

<details>

<summary> The error messages </summary>

```
[2024-09-05, 15:17:29 UTC] {taskinstance.py:3916} ERROR - Error scheduling downstream tasks. Skipping it as this is entirely optional optimisation. There might be various reasons for it, please take a look at the stack trace to figure out if the root cause can be diagnosed and fixed. See the issue https://github.com/apache/***/issues/39717 for details and an example problem. If you would like to get help in solving root cause, open discussion with all details with your managed service support or in Airflow repository.
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 3912, in schedule_downstream_tasks
    return TaskInstance._schedule_downstream_tasks(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/utils/session.py", line 94, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 3861, in _schedule_downstream_tasks
    partial_dag = task.dag.partial_subset(
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/dag.py", line 2646, in partial_subset
    t.task_id: _deepcopy_task(t)
               ^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/dag.py", line 2643, in _deepcopy_task
    return copy.deepcopy(t, memo)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 143, in deepcopy
    y = copier(memo)
        ^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 1388, in __deepcopy__
    setattr(result, k, copy.deepcopy(v, memo))
                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 162, in deepcopy
    y = _reconstruct(x, memo, *rv)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 259, in _reconstruct
    state = deepcopy(state, memo)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 136, in deepcopy
    y = copier(x, memo)
        ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 221, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
                             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/copy.py", line 151, in deepcopy
    rv = reductor(4)
         ^^^^^^^^^^^
TypeError: cannot pickle 'module' object
```
</details>

There are also two warnings in the beginning of this task's logs:

```
[2024-09-05, 15:17:29 UTC] {baseoperator.py:405} WARNING - TemplatedConnectionHttpSensor.execute cannot be called outside TaskInstance!
```



```
[2024-09-05, 15:17:29 UTC] {logging_mixin.py:190} WARNING - /home/***/.local/lib/python3.12/site-packages/***/task/task_runner/standard_task_runner.py:70 DeprecationWarning: This process (pid=27862) is multi-threaded, use of fork() may lead to deadlocks in the child.

```

catalog/dags/common/sql.py
```diff
@@ -38,6 +40,14 @@ def single_value(cursor):
         raise ValueError("Unable to extract expected row data from cursor") from e
 
 
+def fetch_all(cursor):
+    try:
+        rows = cursor.fetchall()
+        return [row[0] for row in rows]
+    except Exception as e:
+        raise ValueError("Unable to extract expected row data from cursor") from e
```
AetherUnbound, line 48: If we're running `fetch_all` on the cursor, why do we also need to return `row[0]` for every result? 😮 Also just noting that Airflow has a [`fetch_all_handler`](https://github.com/apache/airflow/blob/278dd31228f73f926eff68124374d0db1f4d70b8/airflow/providers/common/sql/hooks/sql.py#L93) that we could use if that operation isn't necessary.

catalog/dags/common/sql.py
```diff
@@ -172,3 +182,25 @@ def get_db_hook(self):
...
+@task
+def run_sql(
+    postgres_conn_id: str,
+    sql_template: str,
+    task: AbstractOperator = None,
+    timeout: float = None,
+    handler: callable = RETURN_ROW_COUNT,
+    **kwargs,
+):
+    # TODO this can potentially be used in other DAGs (batched_update, delete_records, add_license_url)
```
AetherUnbound: I have a note in my TODOs to make an issue for this 😄 I'll take care of that soon! haha

catalog/dags/data_refresh/distributed_reindex.py
```diff
@@ -274,6 +279,24 @@ def drop_connection(worker_conn: str):
     session.commit()
 
 
+@task(trigger_rule=TriggerRule.NONE_FAILED)
+def report_reindexing_status():
```
AetherUnbound, line 283: If we really only need this for grouping, maybe we should use an [`EmptyOperator`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/empty/index.html) instead? According to the docs, it might be more performant:

> It can be used to group tasks in a DAG. The task is evaluated by the scheduler but never processed by the executor.

catalog/dags/data_refresh/promote_table.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    return queries.GO_LIVE_QUERY.format(
+        table_name=table_name,
+        temp_table_name=temp_table_name,
+        restore_index_names=restore_index_names,
+    )
+
+
+@task_group(group_id="promote_table")
```
AetherUnbound, line 51: It looks like [`group_id` defaults to the function name](https://github.com/apache/airflow/blob/278dd31228f73f926eff68124374d0db1f4d70b8/airflow/decorators/task_group.py#L75), so we don't need to do anything more than `@task_group` both here and below!

catalog/dags/data_refresh/remap_table_indices.py
AetherUnbound: +1 to what Staci said, according to [the documentation](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html):

> They can have any (serializable) value

I'm assuming "serializable" in this case means JSON serialization. I was able to [return a `NamedTuple` in the `add_rekognition_labels` DAG](https://github.com/WordPress/openverse/blob/41479444c02b394f991d566a6c0e819b706447be/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py#L133-L139) and Airflow [hydrated it back into the `NamedTuple` just fine](https://github.com/WordPress/openverse/blob/41479444c02b394f991d566a6c0e819b706447be/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py#L143). In practice (via the UI), it looks like it serializes it to a list: 

image

image

Dataclasses, although they have more functionality, are a little heavier-weight. If we essentially just need an object to attach a bunch of data to and reference it more easily/programmatically/in a typed way than a dictionary, `NamedTuple` is preferred. That said, I _did_ just find [this documentation on dataclasses!!](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html#passing-arbitrary-objects-as-arguments) 🤩:

> As mentioned TaskFlow uses XCom to pass variables to each task. This requires that variables that are used as arguments need to be able to be serialized. Airflow out of the box supports all built-in types (like int or str) and it supports objects that are decorated with `@dataclass` or `@attr.define`.

stacimc:
Thanks @obulat! Good catch on the record count -- it was getting the counts for the _old_ index rather than the promoted one, there's a quick fix which I'll push shortly!

I cannot reproduce your issues with the image data refresh, though I'll keep trying some more things. Are you positive that the provider DAG had succeeded and finished loading additional records into the catalog DB before the refresh started? 🤔 

The error/warning you mentioned seemed like it's related to the way I've overridden templated fields in the Sensor, but the sensor is _working_ so that's strange. I'll play around with it to see if I can get the warnings cleared.

catalog/dags/data_refresh/remap_table_indices.py
AetherUnbound: Update: I made an upstream PR to make that clearer here! 
https://github.com/apache/airflow/pull/42045

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+from data_refresh import queries
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConstraintInfo(NamedTuple):
+    constraint_table: str
+    constraint_name: str
+    constraint_statement: str
```
AetherUnbound, line 23: Nit: since these are already attached to an object with `Constraint` in the name (and will likely be referenced by a variable named `constraints` or similar, could these maybe just be `table`, `name`, and `statement`?

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def fetch_all_tuples(cursor):
+    try:
+        rows = cursor.fetchall()
+        return [ConstraintInfo(row[0], row[1], row[2]) for row in rows]
+    except Exception as e:
+        raise ValueError("Unable to extract expected row data from cursor") from e
```
AetherUnbound, line 31: NEAT

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def _is_foreign_key(_statement, table):
```
AetherUnbound, line 34: Is the proceeding underscore necessary here? Usually that's used to denote private variables

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+    return f"REFERENCES {table}(" in _statement
+
+
+def _remap_constraint(
+    constraint_name: str,
+    constraint_table: str,
+    constraint_statement: str,
```
AetherUnbound, line 41: These arguments directly mirror the `NamedTuple` above - even if we construct a new tuple or replace the values, would it be easier to pass around one object here and in the other places they're referenced, rather than 3 separate arguments?

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+    table_name: str,
+    temp_table_name: str,
+) -> list[str]:
+    """
+    Given information about an existing CONSTRAINT involving the original media table,
+    produce a list of `ALTER TABLE...` statements to remove the constraint from the
+    existing table and apply it to the temp table.
+
+    The constraint to remap may be a constraint directly on the live media table, in
+    which case it is dropped and then added directly to the temp table. Alternatively,
```
AetherUnbound, line 51: Just to be clear, does that mean the live tables won't have those constraints for some time, even while they're still live? That's _probably_ fine, but if something fails part way through, could the constraints be missing on the live table the next time they're tried to be moved over?

AetherUnbound (commented):
This is wonderful! I'm sorry this last piece took more than expected but it's so nice to finally be there 🥳 

I'm not quite done with my review yet, still about 10 more files to edit, but here's what I have so far. 

One thing I wanted to note too - we send Slack notifications for nearly every step of the data refresh in its current implementation (that just all happens on the to-be-removed ingestion server). Maybe it might make sense to have an `on_success_callback` for the data refresh which does something similar? That way we don't have to add it to every step, but still get some reporting. Just something to think about, if you like the idea or didn't already have a plan for Slack updates, we can make an issue for it!

obulat (approved):
I re-ran the testing instructions, and the image refresh ran well and saved 10,000 images to the main image index.

catalog/dags/data_refresh/remap_table_indices.py
```diff
@@ -0,0 +1,137 @@
...
+        f"temp_import_{old_index_name}"
+        if not is_pk
+        else f"temp_import_{table_name}_pkey"
+    )
+    tokens[index_name_idx] = temp_index_name
+
+    # Update the table name to the temp table. The table name always follows
+    # the word 'ON' in the def, e.g. `ON public.audio`
+    table_name_idx = tokens.index("ON") + 1
+    schema_name, table_name = tokens[table_name_idx].split(".")
```
AetherUnbound, line 57: It's unlikely (maybe impossible?) for this to result in more than two elements, but it might be good to ensure we only ever get two elements:
```suggestion
    schema_name, table_name = tokens[table_name_idx].split(".", maxsplit=1)
```

catalog/dags/data_refresh/remap_table_indices.py
```diff
@@ -0,0 +1,137 @@
...
+        # Create the index
+        run_sql.function(
+            postgres_conn_id=postgres_conn_id, sql_template=index_config.index_def
+        )
+
+    return index_configs
+
+
+def _is_foreign_key(_statement, table):
+    return f"REFERENCES {table}(" in _statement
```
AetherUnbound, line 97: Nit: would it be possible to move this function above the `@task` and `@task_group` definitions? IDEs make these functions much easier to find but it does seem odd sandwiched here in between operators 😅

catalog/env.template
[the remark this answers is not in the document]
AetherUnbound: I think this is fine for now! One disadvantage of declaring this in the Admin UI is that the password (if needed) for the services isn't codified in the same way in our infrastructure repo. Noting that for now, I'll add more thorough thoughts to that discussion when I get a chance to look at that discussion.

catalog/tests/dags/data_refresh/test_remap_table_constraints.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    "all_constraints, table_name, temp_table_name, expected_alter_statements",
+    [
+        # No constraints
+        ([], "audio", "temp_import_audio", []),
+        (
+            [
```
AetherUnbound, line 17: I know it's a little pedantic, but all of these constraints are bundled into the same test case. Would it be possible to separate these out into individual test cases, so they can each fail individually rather than as a single test if one should fail? We can use `audio` and `temp_import_audio` as the `table_name` and `temp_table_name` respectively for every test, so those at least wouldn't need to be parametrized.

catalog/tests/dags/data_refresh/test_remap_table_indices.py
```diff
@@ -0,0 +1,117 @@
...
+
+@pytest.mark.parametrize(
+    "existing_index_defs, expected_index_defs",
+    [
+        # Empty
+        ([], []),
+        # Transform all indices, including pk
+        (
+            [
+                "CREATE INDEX image_provider_7d11f847 ON public.image USING btree (provider)",
```
AetherUnbound, line 20: (Same note here RE: separate tests)

catalog/tests/dags/data_refresh/test_remap_table_indices.py
```diff
@@ -0,0 +1,117 @@
...
+            [
+                "CREATE INDEX temp_import_image_provider_7d11f847 ON"
+                " public.temp_import_image USING btree (provider)",
+                "CREATE UNIQUE INDEX temp_import_image_identifier_key ON"
+                " public.temp_import_image USING btree (identifier)",
+            ],
+        ),
+    ],
+)
+def test(index_configs, expected_sql):
```
AetherUnbound, line 107: Should this have a more descriptive name?

indexer_worker/indexer_worker/api.py
AetherUnbound: We might be able to initialize the operator inside a `@task`, but it's probably easier to use the `HttpHook` inside a `@task` than have an operator-in-operator situation by using `HttpOperator`.

I think I'm fine going with the new static route approach suggested here, but @stacimc do you mind adding a comment about what the route is for, and that it's only intended to be used for local testing?

AetherUnbound (commented):
Okay, here's the rest of my feedback!

catalog/dags/common/sql.py
stacimc: Yes, I would've preferred to use the existing handler but the results are returned as tuples. Unpacking/accessing the first element within `transform_index_defs` felt much less clean to me.

catalog/dags/data_refresh/remap_table_constraints.py
stacimc: Correct. This _does_ mirror the current implementation, so it's not a regression and the (minor) improvement here is that we'd have immediate insight into where a failure happened/if those steps need to be manually undone.

I'll make an issue about this, because it's not simple to work around. We could do something similar to the way we remap the indices, where we give the new constraints a temporary name and then map the names back over in the promote step (if that's even necessary 🤔). But the `delete_orphans` step and remapping foreign key constraints to the temp table can't be easily postponed or undone.

I don't think we should be _too_ worried about this because it's the very last step before promotion, anyway -- but definitely good to call out, and something we'd _need_ to be aware of if for some reason we wanted to start the data refresh over from the beginning after a failure between this step and promotion.

catalog/dags/data_refresh/remap_table_constraints.py
```diff
@@ -0,0 +1,233 @@
...
+    return constraint_statements
+
+
+@task(map_index_template="{{ task.op_kwargs['table_name'] }}")
+def apply_constraints_to_table(
+    postgres_conn_id: str, table_name: str, constraints: list[str]
+):
+    logger.info(f"Applying constraints for `{table_name}`.")
+    for constraint in constraints:
+        run_sql.function(postgres_conn_id=postgres_conn_id, sql_template=constraint)
```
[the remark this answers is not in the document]
stacimc: You can't dynamically map tasks within a dynamically mapped taskgroup, unfortunately (at least yet). I suppose we could have `generate_constraints_for_table` generate constraints for _all_ tables, and then map over table + constraint together when applying, although I don't like that control flow as much... 🤔 I think we could take this into consideration as part of the same [issue](https://github.com/WordPress/openverse/pull/4833#discussion_r1747813964) I'll be creating for the other constraint concerns, particularly as neither are regressions from existing behavior.

catalog/dags/data_refresh/remap_table_indices.py
stacimc: This should actually be deleted, it's part of `remap_table_constraints` and was accidentally duplicated/left in when I refactored them out into separate files! I'll remove it :)

catalog/dags/data_refresh/remap_table_constraints.py
stacimc: Ah, I was confused because I thought you were questioning the preceding underscore in the function name 😅 This was copied directly from the ingestion server and I didn't even notice `_statement`, not sure why it's there! Can definitely be removed.

catalog/dags/data_refresh/remap_table_constraints.py
stacimc: With the variable names now shortened, certainly!

catalog/tests/dags/data_refresh/test_remap_table_constraints.py
stacimc: Part of what this is testing is that the generated constraints are ordered correctly when there are multiple, but I can add additional tests for individual constraints.

catalog/dags/data_refresh/remap_table_indices.py
stacimc: I don't think it's technically impossible, but would require us to have a table name containing a period, in which case we probably have bigger problems 😆 Absolutely no harm in checking, though, you never know!

catalog/dags/data_refresh/copy_data.py
```diff
@@ -36,28 +36,7 @@
 DEFAULT_DATA_REFRESH_LIMIT = 10_000
```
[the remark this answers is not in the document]
stacimc: Well spotted! Fixed.

stacimc:
@AetherUnbound I've run out of time today but written myself a TODO to create issues for the constraint error handling concerns you raised as well as Slack reporting. I'm not sure what Slack reporting we'll want yet, since unlike the ingestion server we'll already have much easier visibility into what's happening through Airflow -- but I think we ought to have reporting at least when the copy data and reindexing steps finish, for sure.

indexer_worker/indexer_worker/api.py
sarayourfriend: Ah, maybe I'm thinking of the hook.

All good, I was just suggesting alternatives _if wanted_, this approach obviously has worked fine for us :slightly_smiling_face: 

It will go away in #3336 anyway, which I didn't know when I wrote my original comment or I would have mentioned it!

catalog/dags/data_refresh/remap_table_constraints.py
AetherUnbound: Thanks for the explanation, I think that sounds good! Not blocking in this case because it mirrors the previous implementation.

catalog/tests/dags/data_refresh/test_remap_table_constraints.py
AetherUnbound: Ahhh okay, thanks for that context!

AetherUnbound (approved):
Fantastic work on this, looks great to me! I didn't get a chance to test it but given that other reviewers have, I feel comfortable approving it 😄

PR #4864, opened 2024-09-05
Fix frontend memory leak

zackkrida:
## Description

This PR fixes the current frontend memory leak by addressing an issue with the initialization of the ui store. For increased ease of testing, I created an object outside of the store to capture the default state (so it could be imported into the test spec) and then spread the properties of that object into the store's state object. Apparently, that is very bad. I'd like to figure out exactly why, so I'll try and look at Pinia's internals tomorrow. It might be very obvious to someone else. 

## Testing Instructions

This should be fairly straightforward to prove it works but there are many steps. You also might need the change I made locally to `utilities/load_testing/Dockerfile` before running these steps, if you're not able to pull the load testing Docker image. 

- With the `main` branch checked out, build the Openverse frontend with: `ov just frontend/run build`.
- Run the frontend with `ov node --inspect frontend/.output/server/index.mjs`
- In a chromium-based browser, visit `chrome://inspect`. Choose the app from the "target" list to open a debugger (by clicking the link "inspect" below the app name, probably `frontend/.output/server/index.mjs`.
- Press the record icon in the top left of the "profile" tab to take an initial heap snapshot
- Back in your terminal `cd utilities/load_testing` and run (replacing the IP with the correct host for your environment, probably `host.docker.internal` on MacOS) `FRONTEND_URL=http://172.17.0.1:3000/ just k6-frontend en`.
- Press the record icon in the top left to take a new heap snapshot.
- Observe that the 2nd heap is much, much larger, in the hundred of MBs.
- Checkout this PR and repeat. The 2nd snapshot should only be a little bit larger than the first now :smile: (we must have some smaller but much less consequential leaks still)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/src/stores/ui.ts
+++ b/frontend/src/stores/ui.ts
@@ -69,21 +69,19 @@ export interface UiState {
 
 export const breakpoints = Object.keys(ALL_SCREEN_SIZES)
 
-export const defaultUiState: UiState = {
-  instructionsSnackbarState: "not_shown",
-  innerFilterVisible: false,
-  isFilterDismissed: false,
-  isDesktopLayout: false,
-  breakpoint: "sm",
-  dismissedBanners: [],
-  shouldBlurSensitive: true,
-  revealedSensitiveResults: [],
-  headerHeight: 80,
-  colorMode: "system",
-}
-
 export const useUiStore = defineStore("ui", {
-  state: (): UiState => ({ ...defaultUiState }),
+  state: (): UiState => ({
+    instructionsSnackbarState: "not_shown",
+    innerFilterVisible: false,
+    isFilterDismissed: false,
+    isDesktopLayout: false,
+    breakpoint: "sm",
+    dismissedBanners: [],
+    shouldBlurSensitive: true,
+    revealedSensitiveResults: [],
+    headerHeight: 80,
+    colorMode: "system",
+  }),
 
   getters: {
     cookieState(state): OpenverseCookieState["ui"] {
--- a/frontend/test/unit/specs/stores/ui-store.spec.ts
+++ b/frontend/test/unit/specs/stores/ui-store.spec.ts
@@ -4,7 +4,7 @@ import { vi, describe, beforeEach, it, expect, test } from "vitest"
 
 import { setActivePinia, createPinia } from "~~/test/unit/test-utils/pinia"
 
-import { defaultUiState, UiState, useUiStore } from "~/stores/ui"
+import { UiState, useUiStore } from "~/stores/ui"
 import { BannerId } from "~/types/banners"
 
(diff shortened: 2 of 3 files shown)


sarayourfriend:
I've been playing with this locally a bunch. I don't know why this is the case, but removing the object spread fixes the leak.

```json
diff --git a/frontend/src/stores/ui.ts b/frontend/src/stores/ui.ts
index 4144d799d..949b75e8a 100644
--- a/frontend/src/stores/ui.ts
+++ b/frontend/src/stores/ui.ts
@@ -69,19 +69,21 @@ export interface UiState {
 
 export const breakpoints = Object.keys(ALL_SCREEN_SIZES)
 
+export const defaultUiState: UiState = Object.freeze({
+  instructionsSnackbarState: "not_shown",
+  innerFilterVisible: false,
+  isFilterDismissed: false,
+  isDesktopLayout: false,
+  breakpoint: "sm",
+  dismissedBanners: [],
+  shouldBlurSensitive: true,
+  revealedSensitiveResults: [],
+  headerHeight: 80,
+  colorMode: "system",
+})
+
 export const useUiStore = defineStore("ui", {
-  state: (): UiState => ({
-    instructionsSnackbarState: "not_shown",
-    innerFilterVisible: false,
-    isFilterDismissed: false,
-    isDesktopLayout: false,
-    breakpoint: "sm",
-    dismissedBanners: [],
-    shouldBlurSensitive: true,
-    revealedSensitiveResults: [],
-    headerHeight: 80,
-    colorMode: "system",
-  }),
+  state: (): UiState => defaultUiState,
 
   getters: {
     cookieState(state): OpenverseCookieState["ui"] {
```

I spent a long time trying to find a reason for this, digging through heap snapshots, allocation timelines... I didn't come up with anything. I looked through Pinia's code a lot, in particular the parts that use `state()`, and didn't see anything suspicious.

The only thing that jumped out to me when looking at the snapshots on the broken code, is that there are heaps of objects with large shallow size referencing the i18n strings, and that are allocated over and over again and never freed. These objects have an allocation size of 8160. There are also many i18n objects as well, with allocation size 3424. After that, there are many Pinia objects at allocation size 3160, and those appear to just be pinia store references.

I don't see any of that with the fix applied. There are still some seemingly redundant i18n objects allocated, but I don't see those Pinia objects.

I can't figure out more than this from the snapshots and timelines. We can try looking at it together sometime, that helped when @obulat and I were individually stuck with debugging the first big frontend memory leak.

obulat (approved):
This does fix the memory leak, even though it's not clear why.

Issue #4774, opened 2024-08-16
Do not log timeout errors for upstream thumbnail exception


stacimc:
## Current Situation

#4748 decreased the timeout for upstream thumbnail requests. Upstream timeout errors are being recorded as issues in Sentry along with all other [UpstreamThumbnailExceptions](https://github.com/WordPress/openverse/blob/5e2e0e308a3f26cf87edc731287aaac8b17117d4/api/api/utils/image_proxy/extension.py#L62).

## Suggested Improvement

We shouldn't logger.error TimeoutErrors. See #4757 for example where we omit these.

## Additional context

https://openverse.sentry.io/share/issue/309a511cd431430aa2123ee7d1b37d91/

sarayourfriend:
I mentioned this in Slack, but we should look at the possible errors that can happen and decide which ones we actually think are potentially actionable vs always noise, and then only logger.error the subset we want in Sentry, and logger.warning the rest.

While we're at it, we can delete the tallying code for these exceptions in the thumbnails. We have never consulted them and as written the keys are pretty useless and potentially take up a huge amount of space in Redis anyway. It would be far more useful in the logs where we can actually query them and we have an in-built rotation mechanism so they aren't just indefinitely piling up.

I'd like the fix for this to take that into account, or else I can create a separate issue for all that if we want to just exclude timeout errors... it'd be far more useful to us in our Sentry volume if we went ahead and addressed the entire scope of the problem though (in all places where we logger.error for upstream provider requests).

sarayourfriend:
@WordPress/openverse-api does anyone have input here? This feels important to take care of to avoid Sentry filling up with unactionable issues, and to clarify the intention in the thumbnail and dead link request areas.

stacimc:
+1 to everything you've said, @sarayourfriend. As to the tallying, I've never looked at it -- pinging @AetherUnbound just in case, I think she might have? Although I could be thinking of the other tallies.

Should we go through Sentry to build the list of possible errors and make the call on noise/actionable?

AetherUnbound:
I have seen this and will respond to this discussion when I have time!

AetherUnbound:
We just received notice from Sentry that our quota is filling up as a result of these and that we may hit our limit soon. I think removing the tally code can be a separate issue & discussion. 

On which errors to ignore, I think we can make a change to ignore the following to start (with more to potentially be added later):
[...]

PR #4875, opened 2024-09-06
Fix handful of networking issues including silence errors

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/4774 by @stacimc 

## Description

Okay, this is a bit of a big one. We can split it up, but I'm running out of time before I need to end working for the week, and I needed to get this all down at least somewhere. The changes here (except the PIL ones) are all related to the linked issue, either by just ignoring the errors as non-actionable the way we discussed, or by actually fixing some of them (hopefully).

The main difference there being the refactor to treat aiohttp client responses as async context managers. This, apparently, is necessary when using aiohttp. I wrote the original code that introduced aiohttp, and I don't know how I missed this, or how the application was able to keep working so long with it missed, especially in such heavily trafficked endpoints like the thumbnail. In any case, I believe this is the root cause of the RuntimeErrors we've seen persistently. They "went away" for a while, but only because we stopped logging them. I was really worried about silencing them again, so spent some time pouring over the issues list for aiohttp, uvloop, uvicorn, httpx (to compare), asgiref, and django. I came up with nothing, and then sort of _randomly_ read some of the aiohttp code trying to understand the request flow better... and yeah, noticed that it returns a context manager from `get` (et co). The docs also say this rather clearly if you read the client reference, but do _not_ on the quickstart (and some instances of network calls just use await without closing).

[This section, for example, does not use async with, just await](https://docs.aiohttp.org/en/v3.9.1/client_quickstart.html#post-a-multipart-encoded-file).

[But if you look here, on the actual `get` method's docs, it's clear about returning a context manager that needs closing](https://docs.aiohttp.org/en/v3.9.1/client_reference.html#aiohttp.ClientSession.get).

So anyway, we got away with it for a long time, but hopefully this fixes it! It certainly makes sense that not using the context manager there would cause weird state issues with the client session being open but the underlying transport being closed and not recycled.

The rest of the changes around the aiohttp code are to get at the rest of the errors described by the issue, the ones we feel are non-actionable. I'm not entirely sure I agree with the list, and if we had more time, I would like to spend it looking at `ClientOSError` in particular. But we don't have time, and considering we're hopefully fixing a large portion of errors that otherwise did not need to happen, I'm fine going forward with this list as "non-actionable". I've made it clear in the comments on the lists of non-actionable exception classes that we might change how we approach this in the future. So this isn't closing any doors.

Finally, when looking at the context manager bits, I realised there were PIL Image objects never getting closed. They are only used in the oembed and watermark endpoints, both of which are known to be underused. Nevertheless, this should fix whatever small potential memory leak might have come about through that.

## Testing Instructions

CI should pass. It took a while to get tests in a good state again, but only because the tests themselves are somewhat complex and obscure the underlying errors. The image_proxy.get function's tendency to just swallow errors makes debugging failing tests really difficult.

Test the following routes by visiting them on your local computer, after running `ov just api/up`:

- image and audio search (test with and without `filter_dead=true`)
- thumbnails
- watermark
- oembed

All of these should return correct responses without errors.

Lastly, try to think of places where there might be dangling context managers, not necessarily to add to this PR, but to open issues to follow up for. If you search the codebase and find any aiohttp references that aren't doing this correctly, we should include them here, I think.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/check_dead_links/__init__.py
+++ b/api/api/utils/check_dead_links/__init__.py
@@ -44,11 +44,28 @@ _timeout = aiohttp.ClientTimeout(total=settings.LINK_VALIDATION_TIMEOUT_SECONDS)
 _ERROR_STATUS = -1
 
 
+# Used to filter network errors during liveness checks that we believe
+# we should get errors (rather than warnings) for, e.g., a Sentry issue.
+# As such, this should exclude errors where we don't think we have the
+# ability to control the outcome, whether now or in the future.
+# Some things, like SSL errors and timeouts, may be used for detecting
+# more confident liveness checks or useful catalog information in the future.
+# However, for now, we just need to avoid filling up our error backlog
+# with things we aren't able to actually do anything about right now.
+_NON_ACTIONABLE_NETWORK_EXCEPTIONS = (
+    aiohttp.ClientConnectorCertificateError,
+    aiohttp.ClientConnectorSSLError,
+    aiohttp.ClientConnectorError,
+    aiohttp.ServerDisconnectedError,
+    aiohttp.ClientOSError,
+)
+
+
 async def _head(
     url: str, session: aiohttp.ClientSession, provider: str
 ) -> tuple[str, int]:
     try:
-        response = await session.head(
+        async with session.head(
             url,
             allow_redirects=False,
             headers=HEADERS,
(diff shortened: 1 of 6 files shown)


dhruvkb (approved):
The changes make total sense to me. LGTM!

PR #4886, opened 2024-09-06
Create S3 client for smart_open from session

AetherUnbound:
## Problem

When I kicked of the Rekognition DAG in production, we saw the following exception:

```
botocore.exceptions.SSLError: SSL validation failed for https://migrated-cccatalog-archives.s3.s3.amazonaws.com/kafka/image_analysis_labels-2020-12-17.txt hostname 'migrated-cccatalog-archives.s3.s3.amazonaws.com' doesn't match either of '*.s3.amazonaws.com', 's3.amazonaws.com'
```

Upon investigating the internals of the `S3Hook`, it looked like the `get_client_type` call was explicitly setting the `endpoint_url` for the `s3` client to be `s3.amazonaws.com`, which something (either `smart_open` or `boto3` itself) was _also_ tacking on another `s3` after the bucket name. 

https://github.com/apache/airflow/blob/1da4b146e954f78280dbe7bbbef452d58c8f728c/airflow/providers/amazon/aws/hooks/base_aws.py#L704

We didn't identify this sooner because we weren't using the production URL and didn't have any trouble when interacting with Minio on this front.

## Description

This PR fixes this issue by using the `S3Hook::get_session` method and creating the `s3` client _from that instead_, without explicitly setting any extra defaults. I tested this locally with the production S3 file and I was able to reproduce the original error, then verity that this change fixed the issue.

In local testing I also ran into an issue where there appeared to be multiple labels for a single ID:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "rekognition_label_insertion_identifier_idx"
DETAIL:  Key (identifier)=(262c626a-a714-4cd3-8159-abc54305d26e) already exists.
```

I've changed the insert SQL to use `replace=True` so that these conflicts are just replaced wholesale rather than raising an exception.

## Testing Instructions

**Maintainers only**

1. Set up a new AWS connection in your `.env` which points to production (I had mine set to `AIRFLOW_CONN_AWS_PROD`)
2. Remove any override you have in your `.env` for `AIRFLOW_VAR_REKOGNITION_DATASET_PREFIX` (so the prod default is used)
3. Alter the connection ID on L79 to use `"aws_prod"` rather than `AWS_CONN_ID`
4. Kick off the `add_rekognition_labels` DAG - if you're on `main`, this should fail with the above exception, if you're on this branch it should pass and processing should continue! (you can also add an `if total_processed > 2000: break` line in the `while` loop to prevent processing from running away)

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py
+++ b/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py
@@ -76,7 +76,9 @@ def parse_and_insert_labels(
         deserialize_json=True,
     )
 
-    s3_client = S3Hook(aws_conn_id=AWS_CONN_ID).get_client_type("s3")
+    # Create the client from the session so that Airflow doesn't override any
+    # defaults we want on the S3 client
+    s3_client = S3Hook(aws_conn_id=AWS_CONN_ID).get_session().client("s3")
     with smart_open.open(
         f"{s3_bucket}/{s3_prefix}",
         transport_params={"buffer_size": file_buffer_size, "client": s3_client},
--- a/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py
+++ b/catalog/dags/data_augmentation/rekognition/add_rekognition_labels.py
@@ -76,9 +76,15 @@ def parse_and_insert_labels(
         deserialize_json=True,
     )
 
-    # Create the client from the session so that Airflow doesn't override any
-    # defaults we want on the S3 client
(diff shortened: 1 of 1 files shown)


AetherUnbound:
Drafting to fix the test issue

PR #4889, opened 2024-09-09
Fix unbound access of `log` (when the captured exception is a timeout)

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse/issues/4888

## Description

`log` will be unbound if the exception is a timeout error. We intentionally do not log timeouts, as noted in the code comment in this block, so the fix is to ensure `log` is only used when it is bound, which is inside the block guarded by the exception being anything but a timeout error.

## Testing Instructions

Cause a timeout locally by lowering `THUMBNAIL_EXTENSION_REQUEST_TIMEOUT` to something very low, and request a thumbnail. You should not see an exception about the unbound local `log`.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/utils/image_proxy/extension.py
+++ b/api/api/utils/image_proxy/extension.py
@@ -91,7 +91,7 @@ async def get_image_extension(media_info: MediaInfo) -> str | None:
                 else:
                     log = logger.error
 
-            log("upstream_thumbnail_exception", exc=exc, exc_info=True)
+                log("upstream_thumbnail_exception", exc=exc, exc_info=True)
 
             raise UpstreamThumbnailException(
                 "Failed to render thumbnail due to inability to check media "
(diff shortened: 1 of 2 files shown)


obulat (approved):
Approving because this works and makes sense.
I wonder why this wasn't caught by linting or tests?

dhruvkb (approved):

sarayourfriend:
>  I wonder why this wasn't caught by linting or tests?

Tests did not catch this because they did not exercise the code path of a timeout for the extension request.

For some reason, F823 (unbound local) from ruff doesn't catch this. I think it's because it's hard to detect unbound locals statically when the local is defined in the scope but only _sometimes_, as it is in this case? mypy and pyright both had (maybe still have) trouble with this as well, from what I've seen.

obulat:
Interesting, I just saw this error in the CI, @sarayourfriend: https://github.com/WordPress/openverse/actions/runs/10766152416/job/29851556998?pr=4841 :) Good that this is merged now.

PR #4892, opened 2024-09-09
Log when API client apps authorize

sarayourfriend:
## Description

While investigating certain strange request patterns today, I've found that it would be nice to be able to see the client application information when one creates a token. This isn't available now because we only log `client_application_authenticated` when the client _authenticates_ (in the authentication area), and clients do not "authenticate", strictly speaking, and at the very least don't go through the authentication flow, when they request a token. In any case, they do gain authorisation, and the oauth2_provider has a signal which provides the `AccessToken` object, which we can use to easily log in that case.

This will make it easier to audit what applications are doing things like sending heaps of token requests and nothing else that we can see from that source (to give a real example of what I'm trying to investigate).

## Testing Instructions

I tried to add a unit test for this, but for some reason the signal does not get processed in the `test_auth` suite. Instead, I've tested it manually using `http`, which is available in `ov`:

1. Checkout the branch and run the API with `ov just api/up`
2. Tail the API container logs with `ov just logs web`
3. In another terminal, register a new application with `ov http localhost:50280/v1/auth_tokens/register/ name=testing description=sara-testing email=user@example.com`. You can change the inputs if you like.
4. Switch back to the logs and you will see a verification link, which you should open to verify your app so you can request tokens.
5. After verifying the app, create a token with `ov http localhost:50280/v1/auth_tokens/token/ grant_type=client_credentials client_id=<client-id> client_secret=<client-secret> -f`
6. Switch back to the logs again, and you should see a line like this:

```
web-1  | 2024-09-09T05:17:39.498883Z [info     ] client_application_authorized  [api.views.oauth2_views] application_id=148 application_name=testing application_verified=True filename=oauth2_views.py func_name=log_app_authorized ip=172.18.0.1 lineno=189 request_id=21e4ef1f-1831-4af9-8338-d86e99d17666
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/api/views/oauth2_views.py
+++ b/api/api/views/oauth2_views.py
@@ -17,6 +17,8 @@ import structlog
 from drf_spectacular.utils import extend_schema
 from oauth2_provider.contrib.rest_framework.permissions import TokenHasScope
 from oauth2_provider.generators import generate_client_secret
+from oauth2_provider.models import AccessToken
+from oauth2_provider.signals import app_authorized
 from oauth2_provider.views import TokenView as BaseTokenView
 from redis.exceptions import ConnectionError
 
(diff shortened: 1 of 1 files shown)


zackkrida (approved):
Works as advertised. This will be really nice to have in the logs.

Issue #4706, opened 2024-08-02
Tag a subset of Playwright tests for running against live environments immediately after deployments to verify the deployment


sarayourfriend:
## Problem

We do not have any automated verification of live environment functionality after deploying them. We take for granted that things like configuration issues and such will cause problems sufficiently bad to surface in other ways (health checks, alarms, etc). That isn't necessarily the case though (https://github.com/WordPress/openverse/issues/4703).

## Description

Identify a subset of our Playwright tests that can be used against live environments to verify core workflows like, SSR search, CSR search, page metadata (head), and so forth.

These can be part of the regular suite, but focusable as a group (without needing to be in the same module) by a new tag on those tests, `@deployment-smoketest`.

We need to add a new dispatchable workflow to this repository to run these tests against an input that specifies staging or production. That new workflow should be dispatched at the end of the deployment workflows in the infrastructure repository, and the result should be reported to Slack.

## Additional context

One idea coming out of the incident related to #4703 discussed by @zackkrida and me.

sarayourfriend:
https://github.com/WordPress/openverse/pull/5075 adds HMAC signing for requests in Playwright, and facilities for running the smoke tests after staging deployments. To fully close out this issue, more tests should get tagged with `@deployment-smoketest` to fully cover at least the critical paths of the application.

PR #4908, opened 2024-09-11
Sign k6 requests with HMAC to enable WAF bypass

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/1031 by @obulat

## Description

This PR adds signing requests with an HMAC to enable bypassing WAF rules, so that k6 requests do not get rate limited.

## Testing Instructions

Follow the test instructions in this infrastructure PR: https://github.com/WordPress/openverse-infrastructure/pull/1048

These must be tested together.

You should also be able to run the k6 tests against your local frontend, without passing `-e signing_secret`. Run your frontend with `ov j p frontend prod` and then in another terminal, run the k6 tests:

```shell
ov j k6 frontend static-en -e FRONTEND_URL=https://staging.openverse.org/ -e scenario_vus=1 -e scenario_iterations=1
```

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/packages/js/k6/rollup.config.ts
+++ b/packages/js/k6/rollup.config.ts
@@ -7,6 +7,8 @@ import { defineConfig } from "rollup"
 import { glob } from "glob"
 
 import typescript from "@rollup/plugin-typescript"
+import { nodeResolve } from "@rollup/plugin-node-resolve"
+import commonjs from "@rollup/plugin-commonjs"
 
 function getConfig(testFile: string) {
   return defineConfig({
@@ -18,7 +20,7 @@ function getConfig(testFile: string) {
       preserveModules: true,
       preserveModulesRoot: "src",
     },
-    plugins: [typescript()],
+    plugins: [typescript(), nodeResolve(), commonjs()],
   })
 }
 
--- a/packages/js/k6/src/crypto.ts
+++ b/packages/js/k6/src/crypto.ts
@@ -1,0 +1,15 @@
+// Courtesy of @mbforbes via https://gist.github.com/robingustafsson/7dd6463d85efdddbb0e4bcd3ecc706e1?permalink_comment_id=4884925#gistcomment-4884925
+
+import { createHMAC } from "k6/crypto"
+
+import type { Algorithm } from "k6/crypto"
+
+export function sign(
+  data: string | ArrayBuffer,
+  hashAlg: Algorithm,
+  secret: string | ArrayBuffer
+) {
+  const hasher = createHMAC(hashAlg, secret)
+  hasher.update(data)
+  return hasher.digest("base64rawurl")
+}
--- a/packages/js/k6/src/frontend/scenarios.ts
+++ b/packages/js/k6/src/frontend/scenarios.ts
@@ -1,8 +1,7 @@
-import { group } from "k6"
-import exec from "k6/execution"
-import http from "k6/http"
+import { check } from "k6"
 
-import { getRandomWord, makeResponseFailedCheck } from "../utils.js"
+import { getRandomWord } from "../utils.js"
+import { http } from "../http.js"
 
 import { FRONTEND_URL, PROJECT_ID } from "./constants.js"
 
@@ -10,23 +9,10 @@ import type { Options, Scenario } from "k6/options"
 
 const STATIC_PAGES = ["about", "sources", "privacy", "sensitive-content"]
 const TEST_LOCALES = ["en", "ru", "es", "fa"]
-const TEST_PARAMS = "&license=by&extension=jpg,mp3&source=flickr,jamendo"
+const TEST_PARAMS = "license=by&extension=jpg,mp3&source=flickr,jamendo"
 
 const localePrefix = (locale: string) => {
   return locale === "en" ? "" : locale + "/"
-}
-
-const visitUrl = (url: string, action: Action) => {
-  // eslint-disable-next-line import/no-named-as-default-member
-  const response = http.get(url, {
-    headers: { "User-Agent": "OpenverseLoadTesting" },
-  })
-  const checkResponseFailed = makeResponseFailedCheck("", url)
-  if (checkResponseFailed(response, action)) {
-    console.error(`Failed URL: ${url}`)
-    return 0
-  }
-  return 1
 }
 
 const parseEnvLocales = (locales: string) => {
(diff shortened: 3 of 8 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -1182,6 +1182,22 @@ jobs:
           wait_time: 60 # check every minute
           max_time: 1800 # allow up to 30 minutes for a deployment
 
+      - name: Setup k6
+        uses: grafana/setup-k6-action@v1
+
+      - name: Login to k6 cloud
+        env:
+          K6_CLOUD_TOKEN: ${{ secrets.GC_K6_TOKEN }}
+        run: k6 login cloud --token "$K6_CLOUD_TOKEN"
```
zackkrida, line 1191: According to the docs the mere presence of a `K6_CLOUD_TOKEN` environment variable is sufficient for authentication:

https://grafana.com/docs/grafana-cloud/testing/k6/author-run/tokens-and-cli-authentication/#authenticate-with-an-environment-variable

So, the login step could be removed and the env block here could be merged with that of the following "Run k6 script" step.

sarayourfriend: Ah! I thought that may be the case but didn't look hard enough in the docs! I think I was looking at the k6 docs, rather than Grafana Cloud k6 docs (lesson to be learned there).

Great call, thanks for checking it out.

sarayourfriend:
The infrastructure side of things is underway in this draft PR: https://github.com/WordPress/openverse-infrastructure/pull/1048

Nothing is working _yet_... :slightly_smiling_face: but it will soon!

packages/js/k6/src/http.ts
```diff
@@ -0,0 +1,95 @@
...
+    return key
+  }
+
+  const encodedSecret = stringToArrayBuffer(signingSecret)
+  key = await crypto.subtle.importKey(
+    "raw",
+    encodedSecret,
+    { name: "HMAC", hash: "SHA-256" },
+    false,
+    ["sign", "verify"]
```
sarayourfriend, line 37: We don't need to verify with this key, so this can just be `sign`

```suggestion
    ["sign"]
```

packages/js/k6/rollup.config.ts
```diff
@@ -18,7 +20,7 @@ function getConfig(testFile: string) {
       preserveModules: true,
       preserveModulesRoot: "src",
     },
-    plugins: [typescript()],
+    plugins: [typescript(), nodeResolve(), commonjs()],
```
sarayourfriend, line 23: New rollup plugins used to support bundling npm dependencies, because k6 doesn't support them unless they are bundled in. In our case, we need it for `core-js` to provide a sensible URL API.

I chose this as an alternative to the jslib URL implementation because (a) that still requires importing `URL` and (b) there are no means to support types for it. The suggested approach for supporting types from k6 jslib libraries is to vendor the library code.

I did not try declaring the types using `declare module "...jslib..."` so I'll try that for good measure (which would be nice for the other jslib function we use).

_However_, I've found that so far the k6 and jslib implementations of certain Web APIs have very subtle and annoying differences to the specification, which mostly just causes a bit of mental overload trying to keep track of. Which is to say, it's nice to be able to bundle in polyfills from core-js that will behave according to the specification!

packages/js/k6/src/frontend/scenarios.ts
```diff
@@ -16,59 +15,57 @@ const localePrefix = (locale: string) => {
...
       }
-    })
+    }
   }
 }
 
-export function visitSearchPages() {
+export async function visitSearchPages() {
   const locales = parseEnvLocales(__ENV.LOCALES)
-  const params = __ENV.PARAMS
```
sarayourfriend, line 55: todo: add back the params functionality for the "search with filters" scenarios that use them.

sarayourfriend:
I'm fairly certain the issue with this is on the Cloudflare side with my usage of their hmac validation function (made a slight error). Excited to test later today and maybe get this in a state where we can start running these after each staging deploy starting next week :grin: 

cc @obulat who created the issue and I suppose will be interested to see this become available.

Once this HMAC pattern is established, we can use it to run Playwright tests as well #4706

sarayourfriend:
Putting this back as draft because I realised I still had a todo comment here, and can undo the async redactor now that I've switched this back to using the k6/crypto library instead of WebCryptoAPI. K6 really isn't meant for async yet, so its okay if we stick to sync code here for now.

Aside from that, if you're reading this and care to test the WAF bypass bit out, that's all working well.

I also decided to leave the GitHub Workflow bits out of this PR to avoid it getting too big/complex... There are a couple of things I need to think about for those. Main thing being... We should run all test scenarios in CI/CD anytime something relevant to a test namespace changes, which is also important to avoid pushing changes to the load tests to main that haven't at least run against the local frontend in the PR CI checks.

Anyway, that will be a separate PR. This one will stay scoped to only getting the HMAC signing working... Which it does! Just need to do a bit more cleanup/undo some of to unnecessary changes in this PR.

packages/js/k6/rollup.config.ts
sarayourfriend: k6-jslib-url just exports core-js: https://github.com/grafana/k6-jslib-url/blob/main/index.src.js

For the sake of making sure bundling in node dependencies is sorted for anyone else working in these tests, I think it's best to go ahead and use core-js directly in this PR, as a means of integrating the bundling process before it's needed otherwise.

obulat:
I'm going to review this PR tomorrow morning

packages/js/k6/src/http.ts
zackkrida, line 1: This whole file is _great_ :man_cook: :kiss:

packages/js/k6/src/crypto.ts
```diff
@@ -0,0 +1,15 @@
+// Courtesy of @mbforbes via https://gist.github.com/robingustafsson/7dd6463d85efdddbb0e4bcd3ecc706e1?permalink_comment_id=4884925#gistcomment-4884925
```
zackkrida, line 1: Random thought but it'd be kinda cool to make folks co-authors of the PR when we reuse code like this! The comment is totally sufficient, of course.

zackkrida (approved):
LGTM. Verified against staging and the local frontend.

packages/js/k6/src/crypto.ts
sarayourfriend: Hmmm, I'd be worried about going to that extent, it might imply they were involved in the PR more broadly, because we squash commits onto main? I'd be worried about doing that without their consent.

Issue #4305, opened 2024-05-10
Setup and add initial Dark Mode visual regression test screenshots


zackkrida:
> This issue is part of the Dark Mode project: #3592. Please see the [implementation plan](https://docs.openverse.org/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.html) for additional detail and context. 

> [!IMPORTANT]  
> This issue is blocked until #4303 and #4304 are completed.  

## Description

Once the dark mode palette is added and can be forced via feature flag, update our visual regression tests to create new dark mode snapshots. Existing screenshots should be renamed to <snapshot_name>_light and the new dark mode screenshots should be named <snapshot_name>_dark.

For setting dark mode in the tests, update `frontend/test/playwright/utils/breakpoints.ts` so that each breakpoint produces and expects a dark mode screenshot to pass as well as the existing light mode screenshot. Emulating dark mode can be easily handled via [Playwright's emulation controls](https://playwright.dev/docs/emulation#color-scheme-and-media).

This will produce a large file diff. When this PR lands @fcoveram  and @WordPress/openverse-frontend  should review the full dark mode appearance for correctness and sufficient color contrast (see the [“Accessibility”](https://docs.openverse.org/projects/proposals/dark_mode/20240325-implementation_plan_dark_mode.html#accessibility) section for more details).

zackkrida:
@obulat and I discussed this a bit an actually came up with a seemingly much simpler approach:

- Add a new step to CI "Run Dark Mode Playwright tests (playwright_vr)", that runs the test suite but with global configuration that enables dark mode, like:

```yaml
         - name: playwright_vr
            script: "ENABLE_DARK_MODE=true test:playwright visual-regression -u"
```

Then, we change the snapshot directory to one for dark mode screenshots using the https://playwright.dev/docs/api/class-testproject#test-project-snapshot-path-template config based on the presence of ENABLE_DARK_MODE

Finally, we setup the dark mode config with something like this:

export default defineConfig({
  use: {
    colorScheme: ENABLE_DARK_MODE !== "true" ? 'light' : 'dark',
  },
});

sarayourfriend:
[...]

PR #4914, opened 2024-09-11
Rename all snapshots to "light"

obulat:
## Fixes

## Description

Most of the changes in this PR are renaming of the snapshots: from "...-linux.png" to "...-light.png". In the follow-up PR, the "-dark.png" snapshots will be added.

~~This PR removes the `-linux` platform suffix from the snapshot names. This will enable someone on a different system to easily compare the snapshots to the linux ones in the repository, instead of creating new snapshots that cannot be compared.~~ this change was removed to prevent flakiness in the future, as per @sarayourfriend's comment below.

## Testing Instructions

The code changes should make sense, and the snapshots should only be renamed, not changed.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/playwright.config.ts
```diff
@@ -21,6 +21,8 @@ const pwCommand = process.env.FASTSTART !== "false" ? "dev" : "prod:playwright"
 
 const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
+  snapshotPathTemplate:
```
obulat, line 24: By default, this is `'{snapshotDir}/{testFileDir}/{testFileName}-snapshots/{arg}{-projectName}{-snapshotSuffix}{ext}'`

`snapshotSuffix` is the platform (`linux`), `project` is a logical group of tests running in the same browser (context). Since we only have one project, we don't need this.

https://github.com/microsoft/playwright/blob/a4bd551597f0161a853f7bdb725e738544db2130/packages/playwright/src/common/config.ts#L166

sarayourfriend (changes requested):
Add `light` to the snapshot names makes sense to me :+1: 

However, I do not believe we should remove the platform indication from the filename. You mention in the PR body that they make it hard to compare snapshots generated across platforms... but that is intentional, it's the reason the platform is indicated. All the major browsers behave differently between platforms, especially and importantly in font rendering. Indicating the platform the snapshot was generated under ensures that when comparisons do happen, there is no possibility for platform-specific variations to cause noise in the diff.

We run playwright tests inside the Docker container for that reason, to eliminate that noise. We really should not compare screenshots from different OSs. If there are devex issues with the Playwright container that make it necessary to run the visual regression tests outside the Playwright container such that the platform indication would not be linux, then we need to fix those issues. Making it easy to do a faulty comparison (screenshots across OSs) is not an appropriate solution to that problem.

frontend/test/playwright/playwright.config.ts
obulat: Removed this in [6395b2c](https://github.com/WordPress/openverse/pull/4914/commits/6395b2c7c73d697b8e1e7c8138e7e6d9c81df788)

sarayourfriend (changes requested):
Just reviewing the diffs of screenshots. Some storybook ones are actually changed, rather than just renamed. Almost all of these with changes have visible differences, and some seem incorrect, so I wanted to confirm they are intentional changes.

The button screenshots have a variety of differences. Some have significantly different colours, like `frontend/test/storybook/visual-regression/v-button.spec.ts-snapshots/filled-pink-8-resting-linux.png`:

image

Some have the same colours, but show ~2px of difference in the width of the button, like `frontend/test/storybook/visual-regression/v-button.spec.ts-snapshots/filled-dark-resting-linux.png` and `frontend/test/storybook/visual-regression/v-button.spec.ts-snapshots/filled-pink-8-hovered-linux.png` (although the latter also has the colour difference issue).

By the name `frontend/test/storybook/visual-regression/v-button.spec.ts-snapshots/filled-white-hovered-linux.png` and `frontend/test/storybook/visual-regression/v-button.spec.ts-snapshots/transparent-dark-hovered-linux.png` would appear to be white and dark-transparent, but both don't match my expectations of what a white or transparent button would be (including in their versions on `main`, so something is wrong here overall, maybe).

`frontend/test/storybook/visual-regression/v-icon-button.spec.ts-snapshots/v-icon-button-sizes-linux.png` is completely broken and doesn't show the components anymore:

image

obulat:
> Just reviewing the diffs of screenshots. Some storybook ones are actually changed, rather than just renamed. Almost all of these with changes have visible differences, and some seem incorrect, so I wanted to confirm they are intentional changes.

Sorry for clicking "rerequest review" too early. I noticed that the storybook snapshots are off, and I extracted #4919 out of this PR. Since the Storybook tests had `maxDiffPixelRatio` set, they were not updated when the colors were updated during the dark theme project. It would be good to audit all the colors/focus rings/snapshots, but I think it should be done later. This specific PR is necessary to set up the light/dark tests, which will make fixing that easier, later :)

sarayourfriend (approved):

sarayourfriend:
Sounds good! LGTM

PR #4922, opened 2024-09-13
Add new feature notice for dark mode

### LINKED ISSUE #4157: Dark mode feature introduction flow
Designs for Dark mode project (#3592)

## Description
In order to have users updated with the latest improvements on the search, I have the following idea.

Feature introduction modal in homepage

The "what's new" message is displayed in a modal with a blurred background that shows an image or gif, text content introducing the feature, and one action to enable/explore the feature and one to close the overlay dialog.

For the dark mode feature, all texts are pending and need comms revision, and the visual placed above text content will show how the feature works in production (discussed in #4155).

Here is a prototype that simulates all interactions expected.

https://github.com/WordPress/openverse/assets/895819/28f784f6-58f0-4cd6-a978-1f81317292e2

## Mockups

Figma file: [Dark mode](https://www.figma.com/file/OP3r0AJWo1k7U3K2hgRwwK/Dark-mode?type=design&node-id=1899-46617&mode=design)

dhruvkb:
## Fixes

Fixes #4157 by @fcoveram

## Description

This PR
- adds a new `VFeatureNotice` component that can be used for notifying people of new features
- adds a component `VDarkModeFeatureNotice` that uses the above to notify people that we now have a dark mode
- adds a spinning-glowing effect to the `VSelectField` based on the `showNewHighlight` boolean prop
- adds some logic to hide the notice and glow once the user has interacted with the feature
- _does not_ add logic to remove the notice based on time because we can do that with cleanup in a subsequent deployment

## Testing Instructions

1. Enable the "dark_mode_ui_toggle" feature.
2. Navigate to the homepage. You should see the notice and the glow effect around theme selector.
3. Click on the "See more" button in the notice or the theme selector. The notice and glow will go away, forever (or until you clear cookies).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VFeatureNotice/VFeatureNotice.vue
zackkrida, line 1: This looks _amazing_!

I think it would be preferable for the entire element to be the link. This will give a larger touch target on mobile. The "see more" text can still be underlined on hover for increased clarity.

The focus state looks nice this way as well:



image

frontend/src/components/VFeatureNotice/VFeatureNotice.vue
```diff
@@ -0,0 +1,57 @@
...
+}>()
+
+const handleClick = (e: MouseEvent) => {
+  emit("click", e)
+}
+</script>
+
+<template>
+  <div
+    class="border-gradient flex h-10 w-[23.125rem] flex-row items-center gap-2 rounded-full border border-double border-tx bg-origin-border pe-4 ps-2"
```
zackkrida, line 19:
```w-[23.125rem]
```

This will need a media query and some adjustment for smaller mobile viewports:

image

dhruvkb: I added the change to make this take up the full width on mobile phones (with x-margins) which aligns it with the hero text on both sides. Now it also wraps text inside without overflowing.

frontend/src/components/VFeatureNotice/VFeatureNotice.vue
dhruvkb: I changed the `VLink` to wrap the entire component. This is how it looks.

Hover:
Screenshot 2024-09-14 at 8 58 55 AM

Focus:
Screenshot 2024-09-14 at 8 57 10 AM

We can adjust these based on @fcoveram 's input.

fcoveram: > what do you think for a hover state for this component? Is underlining "see more" sufficient?

Yes. It's sufficient.

---

> I think it would be preferable for the entire element to be the link. This will give a larger touch target on mobile…

I agree with this, but the focus ring around the whole area looks strange.

"VLink" has its own ring, and it could be great to keep it while making the entire component clickable. Similar to how Gutenberg does it, where the ring wraps the label but the tab is the target area.

https://github.com/user-attachments/assets/0128e7fe-2491-4bb8-a431-f568434b02b6

What do you think of the above?

fcoveram: I shared an idea for a darker homepage in #4936 that could dismiss my suggestion above and instead apply the focus ring to the whole component.

dhruvkb:
In accordance with #4936, the feature notice border has been changed to a single color instead of a gradient. The homepage color change should be a separate PR as that will involve some snapshot changes.

frontend/src/components/VSelectField/VSelectField.vue
```diff
@@ -71,32 +74,58 @@ const splitAttrs = computed(() => {
...
+      v-if="showNewHighlight"
+      class="new-highlight absolute -inset-1.5px animate-new-highlight rounded-[calc(0.125rem_+_1.5px)]"
+      aria-hidden="true"
+    ></div>
+    <div
+      class="relative box-content block w-fit rounded-sm border bg-curr-page text-sm focus-within:m-0 focus-within:border-1.5 focus-within:border-focus hover:border-hover focus-within:hover:border-focus"
       :class="[
-        showSelected ? 'w-full' : 'w-0 max-w-0',
-        hasStartContent ? 'ps-10' : 'ps-2',
+        splitAttrs.classAttrs,
```
obulat, line 86: At first I thought that the snapshot tests were finicky, but they actually did catch a problem. previously, we were binding the `class` attributes to the outer-most div. Now, we are binding them to the inner div instead. However, we have two kind of classes added to `VLanguageSelect`: the one related to the element width (`max-w-full`) and one related to the border (`border-secondary`), which need to be passed to two different elements now (first - to the outer element, and the second - to this div).

I tried moving `max-w-full` from the `VLanguageSelect` in `VFooter` to the `VSelectField`s outer `div`, and it seemed to work fine. But maybe there's a better solution?

obulat (changes requested):
Sorry you had to spend so much time with the snapshots, but I'm glad that they did find a [small] problem with the language select size. Fixing it is the blocking request for change.

frontend/src/components/VSelectField/VSelectField.vue
dhruvkb: I switched up the approach in `VSelectField` to add a new element inside the existing `<div>` instead of adding a wrapper `<div>` around it. Hopefully that will avoid the bug you have pointed above.

obulat (commented):
Perfect ✨  
Could you add a note in the Dark mode project to replace the link in the feature notice?

obulat (approved):

frontend/src/components/VFeatureNotice/VFeatureNotice.vue
```diff
@@ -0,0 +1,37 @@
...
+
+const handleClick = (e: MouseEvent) => {
+  emit("click", e)
+}
+</script>
+
+<template>
+  <VLink
+    :href="seeMoreHref"
+    class="group flex h-10 w-full flex-row items-center gap-2 rounded-full border border-secondary pe-4 ps-2 hover:no-underline sm:w-[23.125rem]"
```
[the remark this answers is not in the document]
dhruvkb: I did this because it was "Fixed" in Figma. When Francisco uses "Fixed" over "Hug", I usually see that as a sign that text is meant to wrap in those scenarios.

PR #4924, opened 2024-09-13
Run k6 against local Nuxt on PRs

sarayourfriend:
## Fixes

Part of https://github.com/WordPress/openverse-infrastructure/issues/1031 by @obulat
Depends on and is blocked by #4908.

## Description

Run k6 against a locally running Nuxt in PRs. This acts as an integration test of the k6 tests themselves, as well as an informal performance test of the local Nuxt application (albeit, without constraints or anything to normalise the results).

I'm not comfortable yet failing this check if any of the k6 requests fail, because I don't know how reliable it will be to run these tests in the GitHub runner with its specific resources constraints. We can evaluate that in this PR and add it now or later if we like.

## Testing Instructions

Check out the new CI step in this PR and see that it makes sense. It should leave a comment with the k6 text summary.

## Checklist

- [N/A] I tried running the project locally and verified that there are no visible errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Okay, the tests are getting throttled when they make API calls. That makes sense. Because these are only local tests meant to test the frontend (and to ensure the tests themselves work), I think it's appropriate to incorporate talkback here. I'll generate new tapes, and have a commit soon to add that.

sarayourfriend:
Finally! It works! Mostly! Yay! I'll push some last tidy-up changes and then undraft this, though it is still blocked on #4908. @obulat any chance you will have an opportunity to review these monorepo PRs, if not also the infrastructure one?

obulat (changes requested):
I tried running these tests locally, and all of the `fa` locale tests have failed because the `/fa` url shows the "This content has disappeared" error page. Does it run fine for you?

I remember I added the `fa` locale for RTL instead of the `ar` locale to these tests, but don't remember the reason. Will have to look it up.

My main blocker to this PR is the number of lines changed. I know those are due to the added tapes, so I opened a PR to only save the first 200 peaks of the audio response.

sarayourfriend:
Another method of reducing the tape size, is we could minify parts of them, or perhaps the entire tape. It's nice for them to be inspectable _when that's needed_, but really we could probably leave them minified by default... It would certainly make the line counts less surprising when adding new tapes!

obulat:
> Another method of reducing the tape size, is we could minify parts of them, or perhaps the entire tape. It's nice for them to be inspectable _when that's needed_, but really we could probably leave them minified by default... It would certainly make the line counts less surprising when adding new tapes!

I do use the tape contents sometimes, e.g. when trying to find which id is being opened from the search page - to make sure that the tape exists, or to add exactly that tape.

> Yes, but you need to install production translations (which the CI step does using setup-env with locales: "production"). This is a pre-existing issue with these test scripts. If you remember why and we can change it to use the ar so we can use the test locales (or we can import the current fa strings as a new testing locale?) works for me.

Oh, I didn't realize it was using the production locales :)
I _think_ I might have wanted to change a locale that doesn't exist. Maybe we should try to change it back to `ar` and see if anything breaks? I don't think it would break anything, and I don't think we need to test non-existent locale.

.github/workflows/ci_cd.yml
```diff
@@ -668,6 +668,86 @@ jobs:
         env:
           DEPLOYMENT_ENV: production
 
+  nuxt-load-test:
```
obulat, line 671: Maybe the load testing could be split to a separate action?

obulat (approved):
Awesome, excited to have the tests running soon 🚀

.github/workflows/ci_cd.yml
sarayourfriend: Potentially, and it's something I plan to look into when adding the job to run k6 against staging in the next PR after this one. The steps to run k6 on a local environment are so different than running it against a remote environment, I wanted to make sure I didn't abstract too early :slightly_smiling_face:

sarayourfriend:
I'll push a change to use `ar` instead of `fa` and then merge this provided things pass :slightly_smiling_face: 

We can also opt to make locales configurable if we want to run against more or different locales in live environments.

PR #4928, opened 2024-09-13
Add dark snapshots to storybook tests

obulat:
## Fixes

## Description

This PR adds the dark snapshots to the Storybook tests.

It also adds a 100ms wait after taking the light snapshot, before taking the dark mode snapshot, to make sure that the dark theme has fully rendered (otherwise, the white line underneath the filters tabs was saved as anywhere between gray and white, depending on how much of the theme switching was finished by the time of the snapshot).
The `"slim-filled-borderless"` focus test was removed since we no longer use this class.

## Testing Instructions

Look through the added snapshots. Only dark mode storybook snapshots should be added. Some elements might need fixing, but not in scope of this PR. The CI should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/utils/expect-snapshot.ts
```diff
@@ -89,10 +89,13 @@ export const expectSnapshot: ExpectSnapshot = async (
...
+  if (!useColorMode) {
     return
   }
+
   await turnOnDarkMode(page, dir ?? "ltr")
+  // Wait for the theme to change.
+  await sleep(100)
 
   // Wait for the theme to change.
   await sleep(100)
```
sarayourfriend, line 101: Is it intentional to have two sleeps? It's unfortunate to need to use one at all... could we instead do something in `turnOnDarkMode` that probes the page for the expected classname on the root element?

If we absolutely need two separate sleeps, then it would be better to document them explicitly in reference to each other (and maybe move them into `turnOnDarkMode`, if it's always needed anyway).

sarayourfriend (changes requested):
LGTM except for the two sleeps. Left a comment asking for clarification.

frontend/test/playwright/utils/expect-snapshot.ts
obulat: The two sleeps are caused by the faulty merge conflict resolution :( I'm removing the extra one.
The existence of class name does not guarantee that the rendering has caught up and all colors have changed. I've seen a range of white-to-gray colors in the separator lines in the flaky snapshots when working on this, so I think a sleep is the best solution here.

frontend/test/playwright/utils/expect-snapshot.ts
```diff
@@ -89,10 +89,13 @@ export const expectSnapshot: ExpectSnapshot = async (
...
     .toMatchSnapshot(getSnapshotName(name, "light"), snapshotOptions)
 
-  if (!(useColorMode === true)) {
+  if (!useColorMode) {
     return
   }
+
   await turnOnDarkMode(page, dir ?? "ltr")
+  // Wait for the theme to change.
+  await sleep(100)
```
obulat, line 98:
```suggestion
```

frontend/test/playwright/utils/expect-snapshot.ts
sarayourfriend: Okay, understood.

I'm surprised it doesn't happen in a single rendering pass :thinking:. What causes the browser to take multiple passes to render in response to the classname change, I wonder.

sarayourfriend (approved):
LGTM, it makes sense that fixing component issues it outside the scope of this PR. Just because this was a good point at which to review them (all in one PR, easy to find) I looked through the snapshots to see if anything should be noted as needing to fix later. Nothing obviously wrong is there, but perhaps @zackkrida or @dhruvkb are better suited to at least glance through them to make sure. The goal being just to note any issues so we can fix them later.

In any case, it doesn't block merging this PR, review of the snapshots can happen afterwards :slightly_smiling_face:

PR #4942, opened 2024-09-17
Simplify audiowaveform installation in API Dockerfile

### LINKED ISSUE #4680: Fix `audiowaveform` isolation issues in API dockerfile
## Current Situation

Despite [realies/audiowaveform building audiowaveform using `BUILD_STATIC=1`](https://github.com/realies/audiowaveform-docker/blob/master/Dockerfile#L94), the output is still a dynamically linked binary. From within the `latest` image:

```
/ # ldd /usr/local/bin/audiowaveform
        /lib/ld-musl-x86_64.so.1 (0x7f17b78f9000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f17b765a000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f17b7636000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f17b78f9000)
```

The API Dockerfile copies these four libraries directly from the `realies/audiowaveform` image and overlays them on the base system. The `audiowaveform` binary we copy out of that image was built and tested with those libraries. However, as noted in our API Dockerfile, because we run `apt update` after copying these libraries, [`audiowaveform` may not necessarily be running with the version of those libraries it was built and tested against](https://github.com/WordPress/openverse/blob/main/api/Dockerfile#L84-L91).

## Suggested Improvement

I propose we resolve this once and for all as follows:

Instead of copying the binary as built from `realies/audiowaveform`, download the Debian Bookworm .deb (the `1.10.1-1-12_<arch>.deb` one, 12 indicating the Debian version) from https://github.com/bbc/audiowaveform/releases/tag/1.10.1 and install it alongside other system dependencies in our existing `RUN apt [...]` line. We should also specify the Bookworm variant of the python `-slim` tag, even though that is the default. The correct .deb needs to be pulled based on the architecture of the build host.

The main complexity of this is ensuring the correct architecture .deb is downloaded depending on the build host. On the other hand, it would allow us to delete a huge portion of the API Dockerfile dedicated to copying (and making it possible to copy) the `audiowaveform` binary out of `realies/audiowaveform`. It also removes an intermediate middle dependency between us and `audiowaveform`.

## Benefit

Resolve this minor edge case in API build stability.

## Additional context

I also considered the following alternatives:

1. Continue copying the binary, but isolate it and its libraries in `/opt/audiowaveform`, and then set `LD_LIBRARY_PATH` when invoking `audiowaveform`. Do this without needing to modify the Python code that calls the binary by creating a shim `/usr/local/bin/audiowaveform` along the lines of...: `LD_LIBRARY_PATH="/opt/audiowaveform/lib:/opt/audiowaveform/usr/lib" /opt/audiowaveform/usr/local/bin/audiowaveform "$@"`
2. Switch from the `-slim` to `-alpine` variant for the final Python image. The (partially) "static" build only leaves out standard library dependencies. These are essentially never going to change unless `audiowaveform` changes to a new language. We could therefore assume that so long as we ensure libstdc++ and libgcc are installed with `apk` when we update the base system (trivial), we wouldn't need to copy these final linked libraries anymore. libc.musl will already be in the alpine base system. 

The first of these alternatives is, in theory, a smaller change, but might take some fiddling (I haven't actually tried it but it should work in theory as long as `ld` works the way it should!). The second might work, but I don't know whether the rest of our Python application runs on Alpine without issue.

Overall, the approach I chose to propose moving forward with is the simplest both in the immediate way it can be implemented and in that it significantly simplifies our API Dockerfile. It is the safest and cleanest option, all things considered.

sarayourfriend:
## Fixes

Fixes #4680 by @sarayourfriend 

## Description

Simplify the audiowaveform installation in the API Dockerfile by using audiowaveform's published deb packages from their GitHub releases, as discussed in the linked issue.

Pinging @dhruvkb for review because of his familiarity with this to begin with, plus I need to make sure that someone on macOS (an arm64 computer, really) tests this.

## Testing Instructions

Checkout the branch and run `ov j build web && ov j up --force-recreate`. Then `ov j api/dj generatewaveforms --max_records=2` and confirm that waveforms are successfully generated for 2 audio files. If you already have waveforms for all audio files, you'll need to remove them manually (left up to you) or start over with API test data using `ov j down -v && ov j api/init`.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -4,27 +4,6 @@
 ARG API_PY_VERSION
 
 ##################
-# Audio waveform #
-##################
-
-# Pull image `realies/audiowaveform` from Docker Hub and alias as `awf`
-FROM docker.io/realies/audiowaveform:1.10.1 AS awf
-
-# Identify dependencies of the `audiowaveform` binary and move them to `/deps`,
-# while retaining their folder structure
-
-# Enable pipefail before `RUN` that contains pipes
-SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
-
-# The script is intentionally single quoted below so that it is not
-# expanded too eagerly and in the wrong context.
-# hadolint ignore=SC2016
-RUN ldd /usr/local/bin/audiowaveform \
-    | tr -s '[:blank:]' '\n' \
-    | grep '^/' \
-    | xargs -I % sh -c 'mkdir -p $(dirname deps%); cp % deps%;'
-
-##################
 # Python builder #
 ##################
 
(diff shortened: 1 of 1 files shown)


dhruvkb (approved):
Tested locally, and it works. It's nice that we can't eliminate the hacky installation and replace it with a more elegant approach.

krysal (approved):
Confirmed it works great in my case too. Awesome!

PR #4943, opened 2024-09-17
Conditional removal of changelogs

### LINKED ISSUE #4726: Sphinx/MYsT dev build is very slow
## Current Situation

The various Sphinx extensions we use combined with the very large number of files (particularly the changelogs) results in a huge overhead for starting up the Sphinx dev site (`ov just documentation/live`).

## Suggested Improvement

It would be nice to find a workaround for this so that most cases of building the dev site could be faster, with optional configuration available to enable a "full" build.

One approach could be to just ignore the changelog directory by default, changelogs account for close to 50% of the files in the documentation site, noticeable during the "Write output" step of the build.

I've tried disabling some extensions locally, like the two custom extensions we have (link issues and link usernames), but doing so does not noticeably speed up the build.

Ignoring the changelogs could be accomplished by conditionally adding them to the `exclude_patterns` list in `documentation/conf.py` based on an environment variable. Something like this could be a good start:

```python
import os

CI = os.getenv("CI") is not None
INCLUDE_CHANGELOGS = bool(os.getenv("INCLUDE_CHANGELOGS", CI))

_default_exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv", "README.md"]
exclude_patterns = _default_exclude_patterns + ([] if INCLUDE_CHANGELOGS else ["changelogs"])
```

Ideally we'd discover the source of the slowness and see whether there are official workarounds. We destroy the build cache in the `documentaiton/live` recipe. Perhaps this isn't always necessary, and we could be more judicious about when we do it, rather than doing it _every_ time.

## Benefit

Reduce the overhead of most documentation live site builds, speeding up work on documentation, encouraging proper local review of changes, make the documentation more pleasant to work with.

kb-0311:
## Fixes

Fixes #4726 by @sarayourfriend

## Description

Based on an environment variable conditionally removes the changelogs

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -43,7 +43,25 @@ todo_include_todos = True
 
 source_suffix = {".rst": "restructuredtext", ".md": "markdown"}
 
-exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv", "README.md"]
+
+import os
+
+# CI will be True if the CI environment variable is defined, otherwise False
+CI = os.getenv("CI") is not None
+
+# INCLUDE_CHANGELOGS will be True if the INCLUDE_CHANGELOGS environment variable is set to a truthy value.
+# If INCLUDE_CHANGELOGS is not set, it will fall back to the value of CI.
+INCLUDE_CHANGELOGS = bool(os.getenv("INCLUDE_CHANGELOGS", CI))
+
+_default_exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv", "README.md"]
+
+# If INCLUDE_CHANGELOGS is True, no additional patterns are excluded, so changelogs will be included.
+# If INCLUDE_CHANGELOGS is False, "changelogs" will be added to the exclusion list.
+exclude_patterns = _default_exclude_patterns + ([] if INCLUDE_CHANGELOGS else ["changelogs"])
+
+# Result: exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv", "README.md"]
+# If INCLUDE_CHANGELOGS is False, the result will also include "changelogs".
+
 suppress_warnings = [
     # Pygments does not fully support language features we use
     # in code blocks in the documentation. Unfortunately the best
@@ -115,5 +133,10 @@ redirects = {
     "packages/eslint_plugin/no-unexplained-disabled-test": "/packages/js/eslint_plugin/no-unexplained-disabled-test.html",
     "packages/openverse_attribution/index": "/packages/python/openverse_attribution/index.html",
 }
+
+# Check if "changelogs" is not excluded, and add a redirect for them
+if "changelogs" in exclude_patterns:
+    # temporary placeholder page
+    redirects["changelogs/"] = "/general/contribution/dev_flow.html"
 
 myst_enable_extensions = ["linkify"]
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -43,7 +43,25 @@ todo_include_todos = True
 
 source_suffix = {".rst": "restructuredtext", ".md": "markdown"}
 
-exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", ".venv", "README.md"]
+import os
+
+# CI will be True if the CI environment variable is defined, otherwise False
+CI = os.getenv("CI") is not None
+
+# INCLUDE_CHANGELOGS will be True if the INCLUDE_CHANGELOGS environment variable is set to a truthy value.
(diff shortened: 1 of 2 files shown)


kb-0311:
@sarayourfriend I am not sure if this is the right approach for this issue. 

For some reason, while setting up the project locally for some reason, after following the general development guidelines , the local test runs keep failing, should I ask for a solution in the slack channel ?

sarayourfriend:
@kb-0311 Can you confirm you have you set up `ov` for linting? It comes pre-loaded with pre-commit so shouldn't have that issue. The output of `cat .git/hooks/pre-commit` should be this, if you have:

```bash
#! /usr/bin/env bash

export TERM=xterm

./ov hook pre-commit "$@"
```

If not, please [follow the `ov` setup instructions on this documentation page, starting with `./ov init` from the repository root](https://docs.openverse.org/general/general_setup.html#up-and-running), and then committing again.

Let me know if you run into any issues with setting up `ov`.

kb-0311:
Thanks a lot , fixed the env setup error : )

Coming to how the redirects are being handled here in my PR:

https://github.com/WordPress/openverse/blob/c95cc0777c53d76e0362a8f88d24eb5937d292f3/documentation/conf.py#L139-L144

I am not really sure if this is the right way to go about it. Also, I could not really find the "Page not found" file location so just added a temporary value for now.

sarayourfriend:
This LGTM! I have two small requests, and then this will be good to go from my perspective.

First is to add the following to `documentation/conf.py` so that we don't see non-actionable warnings in local builds:

```py
nitpick_ignore_regex = () if INCLUDE_CHANGELOGS else (
    ("myst", ".*/changelogs/index.*"),
)
```

This uses the [Sphinx configuration option `nitpick_ignore_regex`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpick_ignore_regex) to ignore "nitpick" warnings about the invalid cross-reference [which is the only targeted method to ignore this error that MyST's documentation describes](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#handling-invalid-references).

---

The other thing to do is address the redirect. I like the idea of using a redirect, and it simplifies things a lot to do so. However, it would be nice to redirect to a page that explains why the changelogs are missing, if someone happens to run across a link to the changelogs that doesn't resolve when they are working in docs locally. Here is a diff you can work from that does that, by adding an orphan page under `meta` to redirect to when `INCLUDE_CHANGELOGS` is false:

```diff
diff --git a/documentation/conf.py b/documentation/conf.py
index 86035b925..a4d5e33ac 100644
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -137,9 +137,6 @@ redirects = {
 }
 
 if "changelogs" in exclude_patterns:
-    # temporary placeholder for now
-    redirects["changelogs/"] = (
-        "/general/contribution/github_contribution_practices.html"
-    )
+    redirects["changelogs/index"] = "/meta/missing_changelogs.html"
 
 myst_enable_extensions = ["linkify"]
diff --git a/documentation/meta/missing_changelogs.md b/documentation/meta/missing_changelogs.md
new file mode 100644
index 000000000..c98635e87
--- /dev/null
+++ b/documentation/meta/missing_changelogs.md
@@ -0,0 +1,13 @@
+---
+orphan: true
+---
+
+# Changelogs missing from local build
+
+The documentation build excludes changelogs unless it is in a CI environment or explicitly instructed to include changelogs. This significantly reduces the number of files processed for local preview builds, which cuts startup time of the documentation preview by half and improving the ergonomics of working with preview documentation locally.
+
+To force the inclusion of the changelogs, run the documentation build with `INCLUDE_CHANGELOGS=1` in your environment. For example, the following will run the preview build, including changelogs:
+
+```shell
+ov env INCLUDE_CHANGELOGS=1 just documentation/serve
+```
```

---

Let me know if these requests make sense, or if you would like help implementing them, and thanks as always for the contribution @kb-0311 :pray:

sarayourfriend (changes requested):
I meant to add that feedback to a request changes comment, sorry for the double ping.

kb-0311:
@sarayourfriend Added the suggestions as well in the new commit, please do take a look when you are free. 

And thanks a lot for the help : )

sarayourfriend (approved):
LGTM! I was able to test this locally by building with `CI=1` and seeing the changelogs show up still and everything works perfectly. This is such a significant improvement, thanks again @kb-0311 :rocket:

PR #4954, opened 2024-09-18
Add envfile rotation DAG

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/968 by @sarayourfriend 

## Description

Adds a DAG to rotate the envfiles in the environment files buckets. The DAG finds the 3 most recent envfiles for each service, and then deletes everything else in the buckets.

Drafted because I'm still trying to understand how best to test this. @stacimc @AetherUnbound if either of y'all have any input/ideas about how to test this, please let me know. I've tried changing the AWS connection locally to test the listing logic, but Airflow expects to be able to write logs to S3 using those same connection credentials, so the tasks fail immediately when they can't write to the logs bucket. Should I give up trying to test against the real buckets and just mock responses out in unit tests? That's about as good as just trying to run it in production and seeing what happens, as it'll depend entirely on whether I get the mocks right... and there's no sure-fire way to test that with our current tools, as far as I know. I've seen [this `moto` library mentioned around before as a good way to mock AWS resources](https://github.com/getmoto/moto) but we don't use it at all, and it's missing implementation for `describe_launch_template_versions` anyway... it might still be useful for other parts.

Let me know what y'all think. I'll leave this drafted until then/until I come up with something on my own. Thanks in advance.

## Testing Instructions

TBD.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/dags/maintenance/rotate_envfiles.py
+++ b/catalog/dags/maintenance/rotate_envfiles.py
@@ -1,0 +1,249 @@
+"""
+Check daily for envfiles in need of removal, following standard criteria for each service.
+
+Manages envfiles based on references in tags of task definitions and launch templates.
+Envfiles are deleted from S3 if they are not referenced by the 3 most recent task definitions or launch templates for a given service.
+
+It does this by constructing a list of "recent" envfiles, and then deleting objects from the envfile buckets
+that are not in the list of recent envfiles.
+
+Resources that reference envfiles are tagged in the form `envfile:<container>` with the value
+set to the hash of the envfile referenced by that resource/version.
+
+For the sake of brevity, this module uses "lt" as an abbreviation for launch template and "task def" as an abbreviation for task definition.
+"""
+
+import logging
+from collections import defaultdict
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.providers.amazon.aws.hooks.ec2 import EC2Hook
+from airflow.providers.amazon.aws.hooks.ecs import EcsHook
+from airflow.providers.amazon.aws.hooks.s3 import S3Hook
+
+from common.constants import AWS_CONN_ID, DAG_DEFAULT_ARGS
+
+
+logger = logging.getLogger(__name__)
+
+DAG_ID = "rotate_envfiles"
+MAX_ACTIVE = 1
+
+ENVFILE_TAG_PREFIX = "envfile:"
+TEMPLATE_LAUNCH_TEMPLATE_SUFFIX = "-tpl"
+TEMPLATE_TASK_DEF_SUFFIX = "-template"
+
+
+@task
+def retrieve_recent_launch_template_envfiles() -> dict[str, set[str]]:
+    ec2_hook = EC2Hook(aws_conn_id=AWS_CONN_ID, api_type="client_type")
+
+    lt_response = ec2_hook.conn.describe_launch_templates()
+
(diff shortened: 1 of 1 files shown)


AetherUnbound:
Here are my suggestions for testing:
- Use a connection ID which is specific to this DAG, but defaults to `AWS_CONN_ID`. This allows the conn ID to be defined locally in your `.env` file and used exclusively for the actions of this DAG, while all using the same conn ID in production. An example for this is the [`AWS_RDS_CONN_ID`](https://github.com/WordPress/openverse/blob/2d364ed7c86437f4a6314e76e461e433901fa82c/catalog/dags/common/constants.py#L49-L48) which is used in the [staging database restore DAG](https://github.com/WordPress/openverse/blob/d500b7764c411f7d228ae12c57dce519c8709610/catalog/dags/database/staging_database_restore/staging_database_restore.py#L176). 
- If you want to mock interactions with S3 specifically, you can use the `minio` instance that gets spun up with the catalog for preloading and operating on S3 file. The [add Rekognition labels](https://github.com/WordPress/openverse/blob/41479444c02b394f991d566a6c0e819b706447be/catalog/tests/dags/data_augmentation/rekognition/test_add_rekognition_labels.py#L12) is an example of this (files in `s3-data` folder, like the [Rekognition one](https://github.com/WordPress/openverse/blob/749259effdc9f6695ef13711b499d32f53dcd9ea/catalog/tests/s3-data/migrated-cccatalog-archives/image_analysis_labels.jsonl), get loaded in when the stack is first created).
- Beyond those needs, I would simply mock the responses from the hook functions. We do this in numerous places throughout the codebase: [1](https://github.com/WordPress/openverse/blob/4d8e8f5f15bee32aad2abfeb40df9ba9cd063c74/catalog/tests/dags/maintenance/test_rotate_db_snapshots.py#L18), [2](https://github.com/WordPress/openverse/blob/90f87b8712b50cf05401de91eb5805b067efeb85/catalog/tests/dags/providers/provider_api_scripts/test_inaturalist.py#L181), [3](https://github.com/WordPress/openverse/blob/7f5baa29f1c5f0e09266946bd2c1a2f321d6f07a/catalog/tests/dags/common/test_slack.py#L24), etc.

sarayourfriend:
Okay, thanks! I'll go with the specific-connection approach first, I do remember now using that to test the RDS snapshot rotation DAG before.

catalog/dags/maintenance/rotate_envfiles.py
```diff
@@ -0,0 +1,320 @@
...
+                        for error in errors
+                    ]
+                ),
+            )
+
+        if deleted := delete_response.get("Deleted", []):
+            deleted_objects[env] = [obj["Key"] for obj in deleted]
+            logger.info(
+                "Deleted the following objects: %s", "; ".join(deleted_objects[env])
+            )
```
krysal, line 271: Same as previous lists.

```suggestion
            logger.info(
                "Deleted the following objects: %s", ";\n\t".join(deleted_objects[env])
            )
```

krysal (approved):
I've tested it and worked as mentioned and expected. Great!

I was wondering how you set up the AWS connection. I used my own credentials, but I think there should be an alternative with more specific permissions and limitations, as you describe.

Issue #4911, opened 2024-09-11
Dark mode visual bugs


zackkrida:
## Description

This is a meta issue to track all dark mode bugs identified in the testing #4783.

- [x] #4829 
- [x] Content report form #5026
    - Textarea text is incorrect color
    - Report button disabled state is incorrect
- Some focus rings are missing/incorrect/using the browser default:
  - [x] Fixed in #4983
    - Source in audio details page
    - Footer items
    - Header items
  - Clear button in search input in Search header
  - ...and more, potentially
- Some focus rings have a white halo in dark mode:
  - [x] #4959 components using `focus-bold-filled` style: image and audio on All content results 
  - [x] Fixed in #4983 Links in body content in all internal pages, Link in filter safe browsing section
  - [x] Fixed in #4983 Links in Attribution box
  - [x] #4939 `VSelectField` (theme switcher, language switcher)
  - ...and more, potentially

- [x] #4910
- In light and dark mode: The search input changes the background color when it should just change the placeholder color.
- Audio playhead is incorrect color

obulat:
While trying to fix the "white halo" around the focus outlines, I found [this article](https://css-tricks.com/copy-the-browsers-native-focus-styles/) about `outline-style: auto`: it's not a bug, but a feature in Chrome.

Of course, since we **do** change the colors of outlines based on the background color, we can override it using `outline-style: solid` or `outline-style:none` (when using the `box-shadow` from Tailwind `ring`).

PR #4958, opened 2024-09-19
Fix link focus rings

obulat:
## Fixes

Related to #4911 by @zackkrida

## Description

This PR fixes most of the focus ring problems for `VLink` elements.

Most of `VLink` elements use the default `outline` that the browser automatically activates on focus. This PR correctly sets the color of the outline. In Chrome, the default `auto` outline style generates two-tiered outline (dark/light), to make sure it works in both dark and light themes). However, since we set the color of the outline with an appropriate contrast, we can use a single-color outline. To do that, we set the default outline style for `focus-visible` to `solid`, which creates a ring of single color.

To make the outline rounded, if a `VLink` does not set a `rounded-` class, then we add `rounded-sm` (border-radius: 2px) to it.

## Testing Instructions

1. Run the app using `ov just frontend/run dev`
2. Go to `http://localhost:8443/` and navigate around using <kbd>Tab</kbd>. The links in the header/footer, Openverse link and the WordPress link, as well as other links throughout the site should have a rounded outline of correct color: pink in light theme, and yellow in dark theme. 
3. Narrow your screen to mobile width, and open the pages modal by clicking on the hamburger in the top right corner (on LTR). The focus rings here should always be yellow, even if the app's color theme is light.

Use the Storybook to test the links, e.g., Internal header: https://docs.openverse.org/_preview/4958/storybook/?path=/story/components-vheader-vheaderinternal--default&globals=viewport:xs

Note that in the light mode, the Openverse link and the close button in the modal still has a pink focus outline, it will be fixed in a later PR.

Check different browsers.

This PR also contains a fix to an e2e test that failed twice in the CI. The fix was to check that the sensitive switch was indeed changed before checking for the cookies.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/tailwind.config.ts
```diff
@@ -98,6 +98,12 @@ export default {
       pink: shadeScale("pink"),
       yellow: shadeScale("yellow"),
 
+      // Focus ring colors
+      focus: {
+        DEFAULT: "var(--color-border-focus)",
+        yellow: "var(--color-yellow-3",
```
sarayourfriend, line 104:
```suggestion
        yellow: "var(--color-yellow-3)",
```

sarayourfriend (commented):
@obulat @zackkrida are there plans to refactor the WP logo component to use the CSS-only methods of colour scheme detection? I suppose it retains the JS approach in order to support the mobile nav, which is always in dark mode... but could the mobile nav add the dark mode classname to its root element, and then follow the same approach for colours as the rest of the application? It will be confusing long term to have the JS-based pattern still present in the app, for anyone trying to understand how to correctly style components.

frontend/tailwind.config.ts
obulat: How did it work for me locally 🤔

obulat:
> @obulat @zackkrida are there plans to refactor the WP logo component to use the CSS-only methods of colour scheme detection? I suppose it retains the JS approach in order to support the mobile nav, which is always in dark mode... but could the mobile nav add the dark mode classname to its root element, and then follow the same approach for colours as the rest of the application? It will be confusing long term to have the JS-based pattern still present in the app, for anyone trying to understand how to correctly style components.

The mobile pages modal is always black, in dark and in light theme. We use a CSS variable to set the background color of the current page, and in this modal, we set it to black (even though in the main app it can be yellow on the homepage for light, and dark gray for dark theme). We could set a similar "contrast text color" variable when we set the page background color variable for the colors of the WordPress link and Openverse logo. Would that be a better approach?

frontend/tailwind.config.ts
sarayourfriend: CSS is surprisingly flexible when it wants to be :grin:

zackkrida:
@sarayourfriend that's a great idea. @obulat is that suggestion make sense to you as well could you create an issue for it?

obulat:
> @sarayourfriend that's a great idea. @obulat is that suggestion makes sense to you as well could you create an issue for it?

Issue opened: #4998

obulat:
These changes were merged with #4983.

PR #4961, opened 2024-09-19
Fix just argument splitting on pass-through recipes with complex args

sarayourfriend:
## Description

This PR fixes a problem we've run into several times, and never really found a suitable ("nice") solution for. 

Without these changes, when we pass just's collected arguments lists (`+args` and `*args` type things), the underlying bash call that just makes will cause arguments to be split on _any space_ and follow difficult to track rules about quotation mark escaping. Commands like this, were simply impossible, without abstract levels of quote escaping:

```shell
ov env DC_USER="airflow"  just exec scheduler airflow dags trigger staging_audio_data_refresh --conf '{"index_suffix": "init"}'

ov just api/test -k "test_image_proxy or test_check_dead_links"
```

The `--conf` argument would have its quotes stripped and turn into `{"index_suffix": "init"}`, passed to the underlying command, which `just` executes using the default shell (`bash` in the case of `ov`). Bash would interpret this as two separate positional arguments, `{index_suffix:` and `init}`, with quotation marks stripped [following bash's quotation handling logic](https://www.gnu.org/software/bash/manual/html_node/Quoting.html).

By using [`just`'s `positional-arguments` setting on these recipes](https://github.com/casey/just?tab=readme-ov-file#positional-arguments), we're able to pass arguments through using bash's argument array expansion, which preserves quoted arguments as single individual arguments, rather than splitting on all strings.

With these changes, the commands above work perfectly and exactly as expected. The quoted arguments are preserved as a single argument, rather than split when `just` compiles the recipe template which drops the argument groups.

I chose not to enable positional-arguments globally in each justfile, despite using it almost everywhere with gather args, because I didn't want to cause side effects on any recipes that weren't written with that particular setting in mind. Plus, setting it explicitly on each recipe should make it easier for folks to understand why those recipes behave the way they do compared to others (or a new one they are writing), and will give them something easy to search for in the just documentation. The global setting would be slightly harder to find and know that it was the effective difference.

## Testing Instructions

Check out main and try to run the commands above to see what kind of error you get back. This helps make sure you know what goes wrong without these changes. Then, check out this branch, and run the commands above again, and they should work.

CI should pass, which is a decent indicator that I haven't broken anything :sweat_smile: 

Make sure to take an extra look at the recipes that use positional arguments for a final gather argument, but that also have individual named just arguments. For example, the `k6/run` recipe. The `$@` split should capture only the positional arguments that aren't handled as regular just template variables.

You can test this by echoing the final command instead of running it, to see the expanded arguments. For example, the diff below yields this output when running the `k6/run` recipe (via the root justfile alias `k6`):

```shell
$ ov j k6 frontend static-en --summary-export ./k6-summary.json -e scenario_vus=1 -e scenario_iterations=1 --console-output ./k6-summary.txt
pnpm run build --input src/frontend/static-en.test.ts
k6 run --summary-export ./k6-summary.json -e scenario_vus=1 -e scenario_iterations=1 --console-output ./k6-summary.txt ./dist/frontend/static-en.test.js
```

Here we can see that the `pnpm run build` command has the correct namespace and test-file interpolated. Likewise in the `k6 run` command. The `k6 run` command also needs to have the additional CLI args passed through between `run` and the script location. This is achieved with `"${@:3}"`, which takes a slice of the `$@` array starting at the 4th position forward. Recall that bash automatically drops from `$@` what would be assigned to `$0` (that is, the command that was run), unless you address it directly. However, the index still remains. If you echo `"${@:0}"` and `"${@:3}" in the recipe, you will see these lines respectively

```
run frontend static-en --summary-export ./k6-summary.json -e scenario_vus=1 -e scenario_iterations=1 --console-output ./k6-summary.txt

--summary-export ./k6-summary.json -e scenario_vus=1 -e scenario_iterations=1 --console-output ./k6-summary.txt
```

As you can see, while `$@` on its own does not include argv0, when slicing it as an array, you still need to consider it when deciding the index to slice from. As such, the _third_ argument passed to `just` will be in index 3 (position 4) of `$@` when sliced, so we slice from 3 onward.

<details>
<summary>Here is a diff you can use to work off and explore the concepts above to better understand <code>$@</code> if you would like</summary>

```diff
diff --git a/packages/js/k6/justfile b/packages/js/k6/justfile
index dc6c2e609..1d390b056 100644
--- a/packages/js/k6/justfile
+++ b/packages/js/k6/justfile
@@ -1,5 +1,7 @@
 # Build and run K6 script by namespace and scenario
 [positional-arguments]
-run namespace scenario *extra_args:
-    pnpm run build --input src/{{ namespace }}/{{ scenario }}.test.ts
-    k6 run "${@:3}" ./dist/{{ namespace }}/{{ scenario }}.test.js
+@run namespace scenario *extra_args:
+    echo "${@:0}"
+    echo "${@:3}"
+    echo pnpm run build --input src/{{ namespace }}/{{ scenario }}.test.ts
+    echo k6 run "${@:3}" ./dist/{{ namespace }}/{{ scenario }}.test.js
```

</details>

## Checklist

- [x] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/justfile
+++ b/catalog/justfile
@@ -99,13 +99,14 @@ pgcli db_user_pass="deploy" db_name="openledger": up
 #########
 
 # Run a command in a test container under `SERVICE`
-_mount-test command: up-deps
+[positional-arguments]
+_mount-test *command: up-deps
     env DC_USER="airflow" just ../run \
         -e AIRFLOW_VAR_INGESTION_LIMIT=1000000 \
         -w /opt/airflow/catalog \
-        --volume \'{{ justfile_directory() }}/../docker\':/opt/airflow/docker/ \
+        --volume {{ justfile_directory() }}/../docker:/opt/airflow/docker/ \
         {{ SERVICE }} \
-        {{ command }}
+        "$@"
 
 # Launch a Bash shell in a test container under `SERVICE`
 # Run pytest with `--pdb` to workaround xdist breaking pdb.set_trace()
@@ -115,7 +116,7 @@ test-session:
 # Run pytest in a test container under `SERVICE`
 [positional-arguments]
 test *args:
-    just _mount-test "bash -c \'pytest ""$@""\'"
+    just _mount-test bash -c "pytest $@"
 
 #############
 # Utilities #
--- a/catalog/justfile
+++ b/catalog/justfile
@@ -56,12 +56,14 @@ install *args: check-py-version
 ######
 
 # Bring up services specific to the catalog profile
+[positional-arguments]
 up *flags:
-    env COMPOSE_PROFILES="catalog" just ../up {{ flags }}
+    env COMPOSE_PROFILES="catalog" just ../up "$@"
 
 # Bring up services specific to the catalog profile, except Airflow
+[positional-arguments]
 up-deps *flags:
-    env COMPOSE_PROFILES="catalog_dependencies" just ../up {{ flags }}
+    env COMPOSE_PROFILES="catalog_dependencies" just ../up "$@"
 
 # Load sample data into upstream DB
 init: up
@@ -111,8 +113,9 @@ test-session:
     just _mount-test bash
 
 # Run pytest in a test container under `SERVICE`
+[positional-arguments]
 test *args:
-    just _mount-test "bash -c \'pytest {{ args }}\'"
+    just _mount-test "bash -c \'pytest ""$@""\'"
 
 #############
 # Utilities #
--- a/catalog/justfile
+++ b/catalog/justfile
@@ -84,11 +84,12 @@ shell:
     env DC_USER="airflow" just ../exec {{ SERVICE }} /bin/bash
 
 # Launch an IPython shell in a new container under `SERVICE`
+[positional-arguments]
 ipython *args: up-deps
     env DC_USER="airflow" just ../run \
         --workdir /opt/airflow/catalog/dags \
         {{ SERVICE }} \
-        bash -c \'ipython {{ args }}\'
+        bash -c "ipython ${@:2}"
 
 # Launch a `pgcli` shell in the PostgreSQL container
 pgcli db_user_pass="deploy" db_name="openledger": up
@@ -124,7 +125,7 @@ test *args:
 
 # Generate the documentation (either "dag" or "media-props")
 generate-docs doc="dag" fail_on_diff="false":
-    #!/bin/bash
+    #! /usr/bin/env bash
     set -e
     if [ "{{ doc }}" == "dag" ]; then
         SCRIPT_PATH="catalog/utilities/dag_doc_gen/dag_doc_generation.py"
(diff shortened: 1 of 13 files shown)


justfile
```diff
@@ -4,9 +4,6 @@ set dotenv-load := false
 # @ - Quiet recipes (https://github.com/casey/just#quiet-recipes)
 # _ - Private recipes (https://github.com/casey/just#private-recipes)
 
-IS_PROD := env_var_or_default("PROD", env_var_or_default("IS_PROD", ""))
-# `PROD_ENV` can be "ingestion_server" or "catalog"
-PROD_ENV := env_var_or_default("PROD_ENV", "")
```
sarayourfriend, line 9: These variables were used long ago when we used to deploy the ingestion server and catalog using `just`. We don't do that any more, and they just add noise/complexity to this file now, so are good to remove.

justfile
```diff
@@ -352,10 +349,12 @@ f:
     just frontend/run dev
 
 # alias for `pnpm --filter {package} run {script}`
-p package script +args="":
-    pnpm --filter {{ package }} run {{ script }} {{ args }}
+[positional-arguments]
+p package script *args:
```
sarayourfriend, line 353: A handful of scripts used `+args=""`, which works, but is technically redundant. `*args` expresses the same thing (optional variadic arguments to gather), and perhaps much more intentionally, especially to relay that they are _optional_, rather than the `+` prefix which indicates at least one is required (a requirement nullified by the default blank that is passed).

sarayourfriend:
Hmmm, I don't know why those render-release-drafter scripts fail in CI. They pass locally :thinking:. I thought at first it was because the `sh` implementation in the GitHub runner was different than the one in `ov`, but setting it to `bash` hasn't fixed it. I will give it a deeper look now.

stacimc (changes requested):
I haven't fully tested everything yet, but on an `ov just recreate`, although recreate runs apparently successfully, I see this error:
```
just up "--force-recreate --build"
env COMPOSE_PROFILES="api,ingestion_server,frontend,catalog" docker compose -f docker-compose.yml "$@"
unknown flag: --force-recreate --build
error: Recipe `dc` failed on line 204 with exit code 16
```

I also get errors trying to run the catalog tests, which I was able to fix by removing the single quotes in L106 for the `_mount-test` recipe.

sarayourfriend:
@stacimc recreate fixed in https://github.com/WordPress/openverse/pull/4961/commits/e83422953ff64717273b8b8bcaf057b7cee6a2d8 and catalog/test fixed in https://github.com/WordPress/openverse/pull/4961/commits/f97a5e612ba6046e718a617b919ca817ee3d5da0

It's surprising the lengths we ended up going to work around this in just!

sarayourfriend:
Yes, it's failed in CI as well. I'm working on a fix... shouldn't be long, sorry! I will draft this until CI is passing.

sarayourfriend:
Okay, this is ready to go now, hopefully!

stacimc (approved):
Amazing! What an improvement, I hope this will save us many headaches in the future :)

catalog/justfile
```diff
@@ -82,11 +84,12 @@ shell:
     env DC_USER="airflow" just ../exec {{ SERVICE }} /bin/bash
 
 # Launch an IPython shell in a new container under `SERVICE`
+[positional-arguments]
 ipython *args: up-deps
     env DC_USER="airflow" just ../run \
         --workdir /opt/airflow/catalog/dags \
         {{ SERVICE }} \
-        bash -c \'ipython {{ args }}\'
+        bash -c "ipython ${@:2}"
```
AetherUnbound, line 92: @sarayourfriend why was `${@:2}` used here? There doesn't appear to be a positional argument that we need to ignore, and it seems like the first argument after the command provided would be swallowed (for instance `ov just catalog/ipython  --TerminalInteractiveShell.editing_mode=vi` would no longer work as expected because the `editing_mode` argument would be ignored). Was that intentional?

PR #4962, opened 2024-09-20
Update init scripts to use new data refresh

### LINKED ISSUE #4153: Update the `load_sample_data` scripts to remove ingestion server usage
## Current Situation

Blocked by #4150

Our load_sample_data.sh scripts currently use just commands to [run parts of the data refresh](https://github.com/WordPress/openverse/blob/226aed0890a19ace1e0b54c1e784c86e9f26b4cb/load_sample_data.sh#L98) on the local ingestion server, as part of setting up the local development environment.

## Suggested Improvement

We should update the scripts to instead use the Airflow CLI to unpause the two new data refresh DAGs and await their completion.

## Additional context

See [this section](https://docs.openverse.org/projects/proposals/ingestion_server_removal/20240328-implementation_plan_ingestion_server_removal.html#update-the-load-sample-data-scripts) of the IP.

stacimc:
## Fixes

Fixes #4153 by @stacimc 

## Description

Updates the `load_sample_data` script to load sample data by actually triggering the new data refreshes, removing this dependency on the ingestion server and also helping to test the new DAGs! If AIRFLOW_CONN_SENSITIVE_TERMS is not configured, it will set this up automatically by pulling the value from the `env.template` to prevent setup from breaking for folks who have not yet added this env variable (see [discussion](https://github.com/WordPress/openverse/pull/4962#pullrequestreview-2331663844)).

It also fixes a few dependency issues that were noticed as part of this work, explained in commit messages.

Note that this change also updates CI, so the data refreshes are run during the setup for the api tests. This should go a long way to integration testing the data refreshes; if they break, or don't populate the sample data correctly, we should see test failures in CI.

The downside is that the overhead of Airflow, while it should be negligible in an actual production data refresh compared to the amount of time a refresh takes, _does_ slow down the init scripts and CI by extension. I've seen the api tests in CI take 9-10 minutes, up from 5-7. **To compensate for this I've taken @sarayourfriend's suggestion to add an option to allow the data refreshes to run concurrently, which is enabled in the init scripts.**

## Testing Instructions

First, remove `AIRFLOW_CONN_SENSITIVE_TERMS` from your `.env` to verify that the init scripts will still work without it.

Run `ov j recreate`. Once the catalog stack is up, you can open http://localhost:9090/dags/staging_audio_data_refresh/grid and watch as the DAG is actually triggered locally. Verify that the DAG (and the related image one) both pass, and that the `recreate` works. Check out the image data refresh at the same time to see them running concurrently.

Then try `ov j api/init` and verify this works successfully as well.

CI should also pass, of course.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (changes requested):
This LGTM and I was able to get everything working as expected locally. I am requesting changes for a very small, but I think significant issue with the development experience of this change.

The data refresh DAGs will fail if the sensitive terms Airflow connection is not configured. A default configuration is in `catalog/env.template`, but nothing in this change indicates a hard requirement to get new .env values (whether manually or by recreating your .env altogether). When the `wait-for-index` check times out in load_sample_data, even if you check Airflow, all you will see is that the connection is missing.

I was able to figure it out and remembered how the sensitive terms connection works in the new data refresh, but that was with my context and experience on the project. Newer contributors will run into this, and will not have any direct indication of how to resolve the problem. In fact, they may not even know to check the DAG status.

I believe there is a relatively simple fix for this: in the load sample data script, grep `catalog/.env` for the sensitive terms connection envvar. If it is not present, log a message and either:

1. Automatically append the one from env.template to .env. In a local environment, for a script used only in CI (where this check will never be relevant) or locally, I feel this is fine.
2. Just log the message but with instructions for adding the new connection manually, and then exit 1.

Both of these could eventually be removed, say in 2 or 3 months. There is no hard "backwards compatibility" requirement of our development environment (it would be untenable), but transitions should ideally to be smooth and transparent at the time.

The diff below solves this problem for me locally. I tested by removing the connection variable from my .env and re-running the init script:

```diff
diff --git a/load_sample_data.sh b/load_sample_data.sh
index 46df60c8f..702bba335 100755
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -6,6 +6,18 @@ CACHE_SERVICE_NAME="${CACHE_SERVICE_NAME:-cache}"
 UPSTREAM_DB_SERVICE_NAME="${UPSTREAM_DB_SERVICE_NAME:-upstream_db}"
 DB_SERVICE_NAME="${DB_SERVICE_NAME:-db}"
 
+# `true` resolves to an empty string, and prevents the script from failing
+# due to `set -e` and grep's non-zero status code if the pattern isn't found
+has_sensitive_terms_airflow_conn=$(grep "AIRFLOW_CONN_SENSITIVE_TERMS" catalog/.env || true)
+
+if [[ ! "$has_sensitive_terms_airflow_conn" ]]; then
+  echo "Adding new Airflow connection environment variable required for sample data loading"
+  grep "AIRFLOW_CONN_SENSITIVE_TERMS" catalog/env.template >> catalog/.env
+
+  echo "Restarting Airflow to populate the new connection variable"
+  just dc restart webserver scheduler triggerer
+fi
+
 while getopts 'c' OPTION; do
   case "$OPTION" in
   c)
```

Otherwise, this PR LGTM. I have some questions about the DevEx, but nothing that is blocking or can't be iterated on. Those are:

1. Loading sample data takes noticeably longer than it did with the ingestion server. Is the DAG concurrency check necessary for local usage? Could we run audio and image data refresh concurrently when loading sample data? Given the regularity of recreating sample data locally (at least when working on some tickets), if there's anything we can do to cut down the amount of time this takes, it would be a great benefit.
2. We trigger the DAGs in order of audio then image. Is it reliable that audio will always happen first? If not, if image could go first somehow, then the load sample data script would sit there waiting for audio much longer. I don't think it would reach the 5-minute time out, but, speaking personally, I would almost certainly kill the process if it was taking longer than 2 minutes.
  - If it's even possible for the DAG order to be inconsistent, running the DAGs concurrently obviates any concern about it.
3. There are four rather visible warnings emitted by Airflow CLI: "/home/airflow/.local/lib/python3.12/site-packages/airflow/cli/commands/dag_command.py:48 UserWarning: Could not import graphviz. Rendering graph to the graphical format will not be possible.". Can we suppress these? However unlikely it may seem, warnings could be taken as the cause if something else goes wrong in the load sample data script, and could confuse someone unfamiliar with how this all works.

To clarify again, none of those 3 are blockers, but I would create issues for any you agree are valid improvements to make after the fact.

The only blocker is making sure the sensitive terms connection doesn't cause unexpected and niche failures during this transitionary period.

stacimc:
@sarayourfriend Great call about the sensitive terms Airflow connection. I added your suggestion in [d9b2a70](https://github.com/WordPress/openverse/pull/4962/commits/d9b2a7033e8d96df37af5cdf60f775be8eeb218a); I agree, we should handle this proactively since it could break all the init scripts for any dev who is unaware of the need to add the new connection, and this is only a temporary measure.

As to running the DAGs concurrently during the sample data script -- I see no reason why not! There were a number of naming collisions (connection ids, FDW, etc) that we didn't need to worry about when the data refreshes were guaranteed to be isolated, but that can be handled. I think it's worth getting this working in this PR because it should help with cutting down that time in the recreate/init scripts and CI. I've added a `param` to the DAG to allow data refreshes to run concurrently, and enabled that during the `load_sample_data` script :)

sarayourfriend (approved):
Amazing!

PR #4991, opened 2024-09-25
Run k6 after staging deployments

sarayourfriend:
## Fixes

Fixes https://github.com/WordPress/openverse-infrastructure/issues/1031 by @obulat 

## Description

This PR:
- Adds an intentionally flexible new run-k6 GitHub Workflow for usage both with manual dispatch and programmatic dispatch via the GitHub CLI during other workflows (we could also leverage that to e.g., run some subset of k6 against production on a regular basis irrespective of deployments, managed using Airflow)
- Changes `FRONTEND_URL` to use a generic `service_url` variable so prepare for also having API tests in k6... this allows us not to need to change the variable name based on the other inputs.
- Adds an `ls` recipe to the k6 justfile for listing the namespaces and scenarios.
- Dispatch that workflow after the staging frontend deploys.

## Testing Instructions

Review the testing runs here: https://github.com/WordPress/openverse/actions/workflows/k6.yml

Note that the three failures were due to issues with passing the signing token. That was fixed by https://github.com/WordPress/openverse-infrastructure/pull/1060 and by correctly passing the token in the workflow :slightly_smiling_face: 

If you have access to Openverse's Grafana, you can also see those runs here: https://openverse.grafana.net/a/k6-app/tests/926189

Those runs were triggered in this PR by me setting the workflow to dispatch on `pull_request` and hard-coding input parameters. I've removed those hard coded parameters and put the workflow back to dispatch.

The only remaining thing to test is the staging deployment bit that triggers this workflow, and of course we can dispatch the workflow manually from the repository to test that further as well (e.g., to test the inputs).

Otherwise, there is not much real testing possible in this PR.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1320,6 +1320,7 @@ jobs:
           token: ${{ secrets.ACCESS_TOKEN }}
           workflow_file_name: k6.yml
           wait_interval: 60
+          propagate_failure: false
           client_payload: |
             {
               "namespace": "frontend",
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -708,7 +708,7 @@ jobs:
       - name: Run k6 frontend all against local Nuxt
         run: |
           just k6 frontend all \
-            -e FRONTEND_URL=http://127.0.0.1:8443/ \
+            -e service_url=http://127.0.0.1:8443/ \
             -e text_summary=/tmp/k6-summary.txt
 
       - name: Upload artifacts
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1312,6 +1312,22 @@ jobs:
           wait_time: 60 # check every minute
           max_time: 1800 # allow up to 30 minutes for a deployment
 
+      - name: Trigger staging k6 load test
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
+        with:
+          owner: WordPress
+          repo: openverse
+          token: ${{ secrets.ACCESS_TOKEN }}
+          workflow_file_name: k6.yml
+          wait_interval: 60
+          client_payload: |
+            {
+              "namespace": "frontend",
+              "scenario": "all",
+              "service_url": "https://staging.openverse.org/",
+              "report": true
+            }
+
   deploy-api:
     name: Deploy staging API
     runs-on: ubuntu-latest
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1320,6 +1320,8 @@ jobs:
           token: ${{ secrets.ACCESS_TOKEN }}
           workflow_file_name: k6.yml
           wait_interval: 60
+          # TODO: Set to true once we see that this test is stable, and we can fail the deployment if it fails.
+          # @see https://github.com/WordPress/openverse/pull/4991
           propagate_failure: false
           client_payload: |
             {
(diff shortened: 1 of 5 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -1262,6 +1262,22 @@ jobs:
...
+          token: ${{ secrets.ACCESS_TOKEN }}
+          workflow_file_name: k6.yml
+          wait_interval: 60
+          client_payload: |
+            {
+              "namespace": "frontend",
+              "scenario": "all",
+              "service_url": "https://staging.openverse.org/",
+              "report": true
+            }
```
sarayourfriend, line 1279: This "propagates failures" so if the load testing workflow fails, it will mark the staging deployment as failed (which will send a note in our alerts channel about a failed CI/CD run).

This might prove to be a mistake, if these tests are flaky in any way. But, we can try it for now, and easily add the `propagate_failure: false` option to the action.

We can also choose not to propagate the failure and instead use Grafana's alerting system to send a message when a load test run is uploaded with bad stats (failures).

.github/workflows/ci_cd.yml
```diff
@@ -1312,6 +1312,22 @@ jobs:
...
           max_time: 1800 # allow up to 30 minutes for a deployment
 
+      - name: Trigger staging k6 load test
+        uses: convictional/trigger-workflow-and-wait@v1.6.5
+        with:
+          owner: WordPress
+          repo: openverse
+          token: ${{ secrets.ACCESS_TOKEN }}
+          workflow_file_name: k6.yml
+          wait_interval: 60
```
sarayourfriend, line 1322: @obulat I'd recommend adding this line initially, to prevent the workflow sending those CI failures if on first merging it isn't working well.

Then, after merging, monitor the new k6 workflow itself and see how it goes. Later on, you can make it a stipulation of staging deployment "success" by setting `propagate_failure` to `true`.

```suggestion
          wait_interval: 60
          propagate_failure: false
```

obulat: That was exactly the change I was going to add, with a doc comment explaining it!

Thank you for your input. I want to get this PR merged next week 🤞

obulat (approved):
:shipit:

sarayourfriend:
@obulat the workflow runs are failing, turns out `ar` locale is not available, presumably due to low-translation percentage? Switching to `skr` might be a good option!

obulat:
> @obulat the workflow runs are failing, turns out `ar` locale is not available, presumably due to low-translation percentage? Switching to `skr` might be a good option!

Yes, turns out _that_ was the reason for not including `ar` locale in the load testing, there are 0 translations for it, so we filter it out and not show it in the language switcher. I thought about adding `fa` for RTL, but I don't think we really need RTL for load testing, so decided to remove `ar` entirely for now: [Exclude ar from k6 load testing](https://github.com/WordPress/openverse/pull/5070)

sarayourfriend:
For the sake of long-term cost control and simplicity, I'd recommend scaling the staging load tests down until they reliably pass, rather than staging tasks up. Reducing the VUs or the number of iterations both sound like good options. You could also try [using the `constant-arrival-rate` executor](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/constant-arrival-rate/) to simplify finding the right balance. You could try something like `timeUnit: 1s, rate: 2, duration: 5m` and then manually ramp `rate` up by +1 until staging starts to fall over. The goal would be to find the highest rate where it consistently passes so you don't end up with flaky tests, but still have an aggressive enough load test that something like a memory leak would show up.

On that note: I'd also recommend intentionally merging a memory leak and seeing whether this new testing mode catches that :slightly_smiling_face:. Trying the same with something CPU bound would also be good.

PR #5027, opened 2024-10-09
Add tests for data refresh task dependencies

### LINKED ISSUE #4340: Test the copy data steps of Data Refresh
## Problem

While much of the new data refresh DAGs is orchestration code that cannot easily be tested, we should add some tests for the copy data SQL and connections between the upstream and downstream database.

stacimc:
## Fixes

Fixes #4340 by @stacimc

## Description

The new data refresh has already been added to the `load_sample_data` scripts, such that it is run each time a contributor runs `ov j api/init` and also in CI. This adds a lot of coverage for the "happy path" of the data refresh.

In addition to this, however, it's important to test that certain vital task dependencies are maintained. While developing the data refresh I found that it's surprisingly easy to accidentally break the dependency graph in a manner such that, while a normal data refresh works as expected, an upstream task failure may not cascade properly. TL;DR: it's easy to introduce a change that seems fine, but causes elasticsearch reindexing/promotion to occur _even if the `copy_data` step fails_. This is a huge concern because it could result in bad or incomplete data going live.

This PR adds tests to ensure that those vital dependencies are maintained.

## Testing Instructions

Tests should pass; take a look and see if there are any other dependencies you feel should be asserted.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/catalog/tests/dags/data_refresh/test_task_dependencies.py
+++ b/catalog/tests/dags/data_refresh/test_task_dependencies.py
@@ -1,0 +1,97 @@
+"""
+Tests vital dependencies for the data refresh: for example, ensuring that the indices have been
+created and refreshed before index promotion runs. These dependencies can be broken in
+unexpected ways so it is important to verify that the most critical ones are preserved.
+"""
+
+import logging
+
+from airflow.models import DagBag
+
+
+logger = logging.getLogger(__name__)
+
+
+def test_staging_schedule():
+    dagbag = DagBag()
+
+    # Assert that all of the staging data refresh DAGs
+    # have a schedule of None
+    for dag in [
+        dag for dag in dagbag.dags.values() if "staging_data_refresh" in dag.tags
+    ]:
+        assert dag.schedule_interval is None, (
+            "Staging data refresh DAGs must be on a `None` schedule"
+            " or risk breaking the load_sample_data script."
+        )
+
+
+def _assert_dependencies(downstream_task_id: str, upstream_task_ids: list[str]):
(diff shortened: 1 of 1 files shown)


obulat (approved):
Approving because the tests pass, but I'm not sure I understand this:
> it's easy to introduce a change that seems fine, but causes elasticsearch reindexing/promotion to occur even if the copy_data step fails. This is a huge concern because it could result in bad or incomplete data going live.

Could you give an example of how it can occur or when it occurred? And why is it not possible to use Airflow to somehow prevent this from happening?

stacimc:
> Could you give an example of how it can occur or when it occurred? And why is it not possible to use Airflow to somehow prevent this from happening?

Sure, here's an [example of a fix](https://github.com/WordPress/openverse/pull/4962/commits/b4b958fcbb90b238f5b917977dae4c6755abe052) for one I noticed in the last PR. In that case the issue was that I altered the return statement of a decorated `@task_group` so that we could access its value in XCOMs; under the covers, though, this also changes the default way that Airflow sets up task dependencies (normally the last task in a task group is set upstream of downstream tasks, but if you have an explicit return, _that task and that task alone_ is set upstream). It's also not possible, to my knowledge, to manually/explicitly set dependencies between tasks inside task groups and outside of task groups. In a DAG as complicated as the data refresh, it's easy to not notice this sort of dependency being broken, particularly because it only "matters" in this case if there's an error.

PR #4363, opened 2024-05-20
Fix frontend to include languages that do not have iso-639-1 codes

zackkrida:
## Description

As discovered by a contributor in [this issue](https://github.com/WordPress/openverse/pull/4224#issuecomment-2119152158), our translations of the Saraiki language were not appearing in production. I realized that the language was being filtered out of our valid translations because it does not have a `langCodeIso_639_1` property in our code. This is because the language is part of the https://en.wikipedia.org/wiki/ISO_639-3 standard, which is essentially v3 of the list of international language codes, and our code was only including languages present in v1. 

Saraiki appears to be the only language impacted by this change

## Testing Instructions

`just frontend/run i18n` and `just frontend/run dev` locally and observe that the `/skr` path now works. You could also count and confirm that the locale chooser now has 61 languages instead of the 60 in production, with running something like `[...document.querySelectorAll('.language option')].length` in your browser's console. 

## Checklist

- [x] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [x] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.
- [x] I ran the DAG documentation generator (`just catalog/generate-docs` for catalog
      PRs) or the media properties generator (`just catalog/generate-docs media-props`
      for the catalog or `just api/generate-docs` for the API) where applicable.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
[...]

PR #5036, opened 2024-10-11
Correctly set locale `language` property

### LINKED ISSUE #5024: Incorrect value for `language` in Nuxt i18n
## Description

We use the three-letter ISO values for `language` properties of the locale instead of using the language-region code (e.g., `es-VE` or `en-ZA`). This prevents Nuxt i18n from correctly setting the alternative links for the different language and region pairs.

## Reproduction

1. Go to http://openverse.org and check the alternate links in the head. You will see that they only use the language code, and do not use the region. So, for South African English, `hreflang` should be set to `en-ZA`, but it's incorrectly set to `en`.
2. Notice that there are only single values for the regional language varieties: one for Spanish, one for English, etc.

obulat:
## Fixes

Fixes #5024 by @obulat

## Description

The Nuxt i18n module sets the `lang` value of the `html` element using the `language` property of the locale. Previously, we were using one of the ISO properties of the locale, but this is incorrect. The value we need is in `slug` property.

Previously, we set `language` as `langCodeIso_639_1 || langCodeIso_639_2 || langCodeIso_639_3`.

This means that all variants of Spanish would overwrite each other because they all had the same ISO values (`{ "langCodeIso_639_1": "es", "langCodeIso_639_2": "spa", "langCodeIso_639_3": "spa" }`), even though their slugs were different (`es`, `es-mx`, etc). If you look in the header language links on https://staging.openverse.org, you will see that we only have a single language for all regions.

[Fix frontend to include languages that do not have iso-639-1 codes](https://github.com/WordPress/openverse/pull/4363) was the latest PR that touched this code. Before that, we would discard any language that did not have the `iso` set, and `iso` was one of the `iso` codes. Now that we use `slug`, we do not discard any languages (including Saraiki mentioned in that issue) solely based on the `iso`(`language`) code. We only discard locales that have low translated percentage.

Because of the way the English is set up separately, its values (url) were being overwritten by other region's English (South African `en-za`).
In this PR, you can see that we have different translations for different regions.

## Testing Instructions

Run the app with `NUXT_PUBLIC_I18N_BASE_URL` env variable set to `https://openverse.org` (so that `href` in the links is set correctly):   `ov env NUXT_PUBLIC_I18N_BASE_URL=https://openverse.org just frontend/run dev`. Check that the alternate language links are set correctly

Screenshot 2024-10-11 at 9 20 20 PM

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -102,6 +102,7 @@ export default defineNuxtConfig({
         dir: "ltr",
         file: "en.json",
         language: "en", // used for SEO purposes (html lang attribute)
+        isCatchallLocale: true, // the catchall locale for `en` locales
 
         /* Custom fields */
 
--- a/frontend/src/locales/scripts/get-validated-locales.js
+++ b/frontend/src/locales/scripts/get-validated-locales.js
@@ -29,14 +29,8 @@ const getValidatedLocales = async () => {
     code: locale.slug,
     dir: locale.textDirection || "ltr",
     file: `${locale.slug}.json`,
-    // Check for a language in all three versions of the ISO 639 spec,
-    // defaulting to the v1 two-character codes before checking for the
-    // three-character codes in the v2 and v3 specs.
-    language:
-      locale.langCodeIso_639_1 ??
-      locale.langCodeIso_639_2 ??
-      locale.langCodeIso_639_3 ??
-      undefined,
+    // Used for the html lang attribute.
+    language: locale.slug,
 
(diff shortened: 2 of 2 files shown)


krysal (approved):
Nice that it has an easy fix, and thank you for explaining the problem and the solution in detail, @obulat. It works perfectly!

PR #5046, opened 2024-10-18
Initialize the stores only on the server

### LINKED ISSUE #5037: Duplicate `/stats` requests
## Description

Every frontend request now sends 4 requests to the `/stats` endpoint: 2 requests from the server and 2 requests from the client during hydration (for image and audio). This is excessive since the server results should be reused on the client.

## Reproduction

1. Look at the network tab when opening openverse.org. 
2. You will see 2 XHR requests for `/stats`: images and audio. These requests should not be sent because the stats are already available in the server.

obulat:
## Fixes

Fixes #5037 by @obulat

## Description

This PR makes the `init-stores` plugin run only on the server.

We don't need to run it both on the server and on the client because the cookie values don't change from the server to client rendering. We also do not need to re-fetch the providers since they were fetched on the server.

This PR also adds numbers to the plugin names to ensure that the order of execution is correct on the server: `01.api-token` - `02.init-stores` - other plugins. The token is necessary for fetching the providers, and the stores need to be initialized before anything else

The other plugins do not depend on each other, so there I removed numbers from their names.

## Testing Instructions

Run the app and check the browser devtools Network tab for requests. There should be no `stats` requests since they are sent on the server (and thus, don't show up in the browser devtools).

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound (approved):
I was able to confirm that requests to the `/stats` endpoints aren't being made when I visit the homepage locally 🥳 

> This PR makes the `init-stores` plugin run only on the server.

Just curious, was this done simply by changing `.ts` to `.server.ts` in the plugin name? Will doing this have any affect on any of the other data in the store? Things like dark mode or sensitivity preferences?

obulat:
> Just curious, was this done simply by changing `.ts` to `.server.ts` in the plugin name? Will doing this have any affect on any of the other data in the store? Things like dark mode or sensitivity preferences?

Yes, simply by adding the `.server`. The plugins without a suffix run both on the server and on the client. If you want them to run either on the server or on the client, you can add the suffix. 
For dark mode or sensitivity, we get the cookies on the server - and initialize the stores based on them. I don't think there is a scenario where the cookies from the server request would be different from the cookies during the hydration (from the broser/client).

AetherUnbound:
Thanks for that explanation!

PR #5075, opened 2024-10-25
Add baseline functionality for running Playwright tests against live environments

sarayourfriend:
## Fixes

Related to https://github.com/WordPress/openverse/issues/4706 by @sarayourfriend 

## Description

This PR introduces the same HMAC request signing strategy used for #4908 tests against live environments to use with Playwright. This allows Playwright tests to run against the live environment and bypass Cloudflare's WAF, otherwise it gets blocked as automated traffic.

To enable the signing transparently, I've created an override of the `page` fixture that automatically adds the HMAC headers to requests. This requires using `test` imported from `~~/test/playwright/utils/test` so the fixture is always available. I've updated all imports of `test` to use the new location and added an ESLint rule to encourage importing from the fixture-enabled location.

Additionally, I've added a new workflow file similar to the k6 workflow for running the Playwright smoke tests. The smoke tests are identified via the [test annotation](https://playwright.dev/docs/test-annotations#tag-tests) `@deployment-smoketest`. The workflow is dispatched after staging deployments, before the load test runs.

For this initial integration, I've just tagged some keyboard navigation tests with `@deployment-smoketest`, but to fully close the issue, more tests should be included.

## Testing Instructions

CI should pass, regular Playwright test runs should be unaffected.

You should be able to run the deployment smoke tests locally using the following command:

```bash
env HMAC_SIGNING_SECRET=<signing secret> PLAYWRIGHT_BASE_URL=https://staging.openverse.org just p frontend test:playwright:local --grep '@deployment-smoketest'
```

Maintainers can pull the signing secret as needed.

I also added temporary triggers for the new workflow file to run on pull_requests, and you can check out the successful run of that [at this link](https://github.com/WordPress/openverse/actions/runs/11510009898/job/32041101434?pr=5075). I've removed the pull_request trigger now, so it will only run when dispatched, just like the k6 workflow.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/test/playwright/utils/test.ts
+++ b/frontend/test/playwright/utils/test.ts
@@ -1,0 +1,61 @@
+// eslint-disable-next-line no-restricted-imports
+import { test as base } from "@playwright/test"
+
+const encoder = new TextEncoder()
+
+const signingSecret = process.env.HMAC_SIGNING_SECRET
+
+const { subtle } = globalThis.crypto
+
+let signingKey: null | CryptoKey = null
+
+async function getSigningKey() {
+  if (!signingSecret) {
+    return null
+  }
+
+  if (!signingKey) {
+    const encodedSecret = encoder.encode(signingSecret)
+    signingKey = await crypto.subtle.importKey(
+      "raw",
+      encodedSecret,
+      { name: "HMAC", hash: "SHA-256" },
+      false,
+      ["sign"]
+    )
+  }
+
+  return signingKey
+}
+
+export const test = base.extend({
+  page: async ({ page }, use) => {
+    // Only match staging; it'll just be ignored for local testing
+    await page.route(
+      (url) => url.host === "staging.openverse.org",
+      async (route) => {
+        const key = await getSigningKey()
+        if (!key) {
+          return route.continue()
+        }
+
+        const request = route.request()
+        const url = new URL(request.url())
+        const timestamp = Math.floor(Date.now() / 1000)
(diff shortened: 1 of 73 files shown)


frontend/test/playwright/utils/test.ts
```diff
@@ -0,0 +1,61 @@
+// eslint-disable-next-line no-restricted-imports
+import { test as base } from "@playwright/test"
+
+const encoder = new TextEncoder()
+
+const signingSecret = process.env.HMAC_SIGNING_SECRET
```
obulat, line 6: We need to add this env variable to the infrastructure repository to get this to work, right?

obulat (approved):
Thank you so much for adding this! 
I like the idea of adding this in the `test` fixture.

frontend/test/playwright/utils/test.ts
sarayourfriend: It's the same secret used for signing requests in k6, so no need for a new secret (though you could want to change the name to make it generic instead of k6 specific). See here in the new workflow file that I passed `K6_SIGNING_SECRET` for `HMAC_SIGNING_SECRET`:

https://github.com/WordPress/openverse/pull/5075/files#diff-c0447faa897e95e3c385302da0a0b3ea051e6c90be93e10cf0914eb8589a881dR34

PR #5135, opened 2024-11-01
Fix Playwright VR test failures

### LINKED ISSUE #5134: Flaky test: `frontend/test/playwright/visual-regression/components/filters.spec.ts-snapshots/filters-modal-checked-ltr-md-dark-linux.png`
## Description

The test is flaky because the tab bottom border color in the content switcher modal doesn't render correctly.

## Reproduction

See failed run: https://github.com/WordPress/openverse/actions/runs/11623248087/job/32370067485

obulat:
## Fixes

Fixes #5134 by @obulat
Note: the added lines are lines from thumbnail request tapes for rendering the search results page in Playwright

## Description

The problem in #5134 was that the dark theme content switcher modal would sometimes display a "loading" state, even though the loading was finished before the light snapshot was taken.

In the Playwright visual regression tests, we have `expectSnapshot` function that takes the light snapshot, and then switches the theme to dark and takes the dark snapshot. The theme switching should not cause any network requests such as re-loading the search results, but in these tests for some reason that was happening.

Originally, this PR added a wait for `networkidle` event before taking a snapshot of the filters modal. I also tried various options of using the UI state to detect that all of the images were loaded (so that there are no changes in the snapshot backgrounds) but nothing works reliably. 
Even with this check, the app was re-fetching data when the theme was changed from light to dark in `expect-snapshot`. I couldn't understand why that was, and couldn't reproduce it in the real app. It was only happening in Playwright trace.

So, instead of tweaking the wait for the content settings modal being rendered, I made the test load the filter on SSR (so, instead of searching without filters, and then selecting a filter and snapshotting the filters grid, we now go to the page with `license_type=commercial` query and load the filtered search results on the SSR).

### Other changes
- Instead of taking the picture of the whole page with the content settings modal, we take a picture of only the content settings modal.
- The border above the tab panel in the content settings modal in the dark theme wasn't being set correctly: it was using `border-default` instead of `border-overlay`, and one of the "dark" theme CSS variable lists didn't have a value for `border-overlay` color.
- initialize the theme switcher icon based on the store theme, if it's available (rather than undefined) to make it render faster
- unset aria-hidden instead of using `aria-hidden=false` for "Show results" button in the modal
- save search result thumb tapes because if the thumb doesn't exist, the page tries to load the image from the provider, making network requests.

## Testing Instructions

The CI should pass even if re-ran several times

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/test/playwright/utils/expect-snapshot.ts
```diff
@@ -96,7 +137,7 @@ export const expectSnapshot: ExpectSnapshot = async (
   await turnOnDarkMode(page, dir ?? "ltr")
 
   // Wait for the theme to change.
-  await sleep(200)
+  await sleep(100)
```
obulat, line 140: This was the original value. I thought we needed to increase it to fix the modal rendering, but turns out that switching the theme was done incorrectly before. When the modal is open, the theme switcher is covered by it. We need to close the modal, switch the theme, and then re-open the modal before taking the snapshot.

dhruvkb (approved):
The changes to the tests and the updated snapshots look good and make sense. I'm approving because it's critical but I do have one question:

Can you explain why there are new tapes as part of this PR? Additionally, if there are new tapes added are some old tapes no longer useful and can be removed?

obulat:
Thank you!

> Can you explain why there are new tapes as part of this PR? Additionally, if there are new tapes added are some old tapes no longer useful and can be removed?

When the search results page loads, if the thumbnail is not available (in one of the tapes), then the component makes a request for the image's direct URL in on image load error handler, so the rendering is further delayed. If we have all tapes for all of the results on the page, then we don't make any requests to the providers.

As for removing tapes, maybe we should remove one of the search terms we use in the Playwright tests. We currently use "cat", "birds" and "galah", which is not necessary. If we decide that this is a good idea, we can update the test files to only have one search terms, and remove unused tapes.

Issue #592, opened 2022-03-21
Switch to `fetch` away from `axios`


sarayourfriend:

sarayourfriend:
@obulat Any thoughts on whether this is worth doing? It looks like most of our axios usage is concentrated in the API service of the frontend, along with usage in the i18n utils. Axios is 30.2 kB minified: https://bundlephobia.com/package/axios@1.6.2

Could be a good "help wanted" issue?

dhruvkb:
The i18n utils are quite decoupled from the frontend so switching them could very well be a standalone "good first"/"help wanted" issue. But it would be good to have consistency in the library used across the codebase or at least within a stack layer so we should only switch if we plan on going all in with it.

sarayourfriend:
Would it help to separate this into two issues, then? One for the application code and one for the i18n code? The i18n utilities would need to use `node-fetch` anyway, so it's good to handle as a separate issue, I think.

sarayourfriend:
Actually, we need `node-fetch` for SSR as well, so maybe that point is moot.

We do need to make sure `node-fetch` doesn't accidentally get used in the client bundle, so we might need to have separate files exporting the fetch function to use:

```ts
// utils/fetch.client.ts
export const fetch = window.fetch;


// utils/fetch.server.ts
export fetch from 'node-fetch'
```

Something along those lines.

PR #5177, opened 2024-11-22
Rewrite the i18n scripts

### LINKED ISSUE #629: [i18n] Add irregular pluralizations
## Problem

Some languages have more than one plural noun form that correspond to different numbers. We need to set up the pluralization for languages other than Russian.  

## Description

Vue-i18n uses the pluralization config to select the correct form, and here is the example implementation for Russian:
https://github.com/WordPress/openverse-frontend/blob/0f784f4df9fdf4b460dab387f0877229420efebc/src/plugins/vue-i18n.js#L11-L30
This corresponds to the following config in the `.po` format:
`"pluralExpression": "(n % 10 == 1 && n % 100 != 11) ? 0 : ((n % 10 >= 2 && n % 10 <= 4 && (n % 100 < 12 || n % 100 > 14)) ? 1 : 2)",`

It is OK to create one PR per single locale group with the same pluralExpression, one from the following list:

### Locales:
- [ ] Arabic `ar`, Algerian Arabic `arq`, Moroccan Arabic `ary`
- [ ] Bengali (India) `bn_IN`, Breton `bre`, Persian (Afghanistan) `fa_AF`, Persian `fa_IR`, French (Belgium) `fr_BE`, French (Canada) `fr_CA`, French (France) `fr_FR`, Arpitan `frp`, Borana-Arsi-Guji Oromo `gax`, Indonesian `id_ID`, Kabyle `kab`, Lingala `lin`, Maori `mri`, Occitan `oci`, Picard `pcd`, Portuguese (Brazil) `pt_BR`, Saraiki `skr`, Tahitian `tah`, Turkish `tr_TR`, Turkmen `tuk`, Tamazight (Central Atlas) `tzm`
- [ ] Cornish `cor`
- [ ] Czech `cs_CZ`, Slovak `sk_SK`
- [ ] Welsh `cy`
- [ ] Lower Sorbian `dsb`, Upper Sorbian `hsb`, Slovenian `sl_SI
- [ ] Irish `ga`
- [ ] Scottish Gaelic `gd`
- [ ] Icelandic `is_IS`', 'Macedonian `mk_MK`
- [ ] Lithuanian `lt_LT`
- [ ] Latvian `lv`
- [ ] Montenegrin `me_ME`
- [ ] Polish `pl_PL`
- [ ] Romanian `ro_RO`
- [ ] Silesian `szl`
- [ ] Tamazight `zgh`

These locales use a single version, so probably don't need any pluralization set up?
- [ ] Emoji `art_xemoji`, Balochi Southern `bcc`, Tibetan `bo`, Dzongkha `dzo`, Igbo `ibo`, Japanese `ja`, Georgian `ka_GE`, Kyrgyz `kir`, Khmer `km`, Korean `ko_KR`, Lao `lo`, Mauritian Creole `mfe`, Malay `ms_MY`, Rohingya `rhg`, Sundanese `su_ID`, Thai `th`, Tigrinya `tir`, Tatar `tt_RU`, Uzbek `uz_UZ`, Vietnamese `vi`, Wolof `wol`, Chinese (China) `zh_CN`, Chinese (Hong Kong) `zh_HK`, Chinese (Singapore) `zh_SG`, Chinese (Taiwan) `zh_TW`

These locales should use the same rules as Russian:
- [ ] Belarusian `bel`, Ukrainian `uk`, Serbian `sr_RS`, Croatian `hr`, Bosnian `bs_BA` 

## Additional context

To get the `.po` format pluralization rules, run `pnpm run dev`. This should create a `src/locales/scripts/wp-locales.json` file, with an object with locale properties for all available locales. Only the locales with irregular plurals have the `"pluralExpression"` property.

## Implementation

### LINKED ISSUE #4979: Flatten i18n message keys in message JSON
## Problem

Our current vue/i18n messages files utilise nested JSON objects to discriminate keys. A sample from our `en.json5` looks like this:

```json5
{
  "404": {
    title: "The content you’re looking for seems to have disappeared.",
    main: "Go to {link} or search for something similar from the field below.",
  },
  hero: {
    subtitle: "Explore more than 800 million creative works",
    description: "An extensive library of free stock photos, images, and audio, available for free use.",
    search: {
      placeholder: "Search for content",
    },
    disclaimer: {
      content: "All {openverse} content is under a {license} or is in the public domain.",
      /**
      Interpolated into hero.disclaimer.content:
       _{license}_ part of "All Openverse content is under a {license} or is in the public domain."
       */
      license: "Creative Commons license",
    },
  },
}
```

Code that references these keys does so in a dot-delimited path format. For example, `hero.disclaimer.content` references the key at that path in the nested object.

While nested objects may provide a slight advantage to authorship of the messages file, it presents a severe disadvantage when trying to find the message from a key in the code referencing it. Using the example above, if you wanted to see the message associated with the key `hero.disclaimer.content`, and tried to search the codebase for that literal string, you would not find it. Instead, you would have to know to navigate to the `en.json5`, and then find the `hero` key, find its `disclaimer` key, and then finally the `content` key. Sometimes you can shortcut this by searching for just the final part of the key, in this case `content`. However, we have 31 keys that end with the segment `.content`, so searching `content:` in the file would still require looking through individual instances to find it. The example above also uses relatively small and shallow nesting and does not illustrate the additional difficulty of navigating deeply nested keys in large collections of messages, like `sensitive.designations.userReported.title.description.a`.

## Description

Instead, I propose we "flatten" the messages objects where the keys are the full path to the message. In other words, remove all nesting.

For example, the messages excerpt above would turn into this, instead:

```json5
{
  "404.title": "The content you’re looking for seems to have disappeared.",
  "404.main": "Go to {link} or search for something similar from the field below.",

  "hero.subtitle": "Explore more than 800 million creative works",
  "hero.description": "An extensive library of free stock photos, images, and audio, available for free use.",
  "hero.search.placeholder": "Search for content",
  "hero.disclaimer.content": "All {openverse} content is under a {license} or is in the public domain.",
  /**
  Interpolated into hero.disclaimer.content:
    _{license}_ part of "All Openverse content is under a {license} or is in the public domain."
    */
  "hero.disclaimer.license": "Creative Commons license",
}
```

This format is backwards compatible with our existing messages objects. If you replace the 404 and hero objects with the flattened version above and run your local frontend, there is zero issue.

**The benefit of this approach is that it is easily searchable in both directions**. From the messages file, it is easier to find uses of the keys. From runtime code, it is easier to find the content of the translation string. Additionally, our `json-to-pot` script could be simplified, as right now it has to collapse keys when converting to POT, because `POT` is a flat-format.

Our POT-to-JSON conversion re-explodes the keys into the nested object. We should retain this behaviour in the final output messages files. A meaningful downside to the flat format is an increase in the total character size of the keys, due to the repeated strings.

<details>
<summary>The nested example minifies to 509 characters (expand for minified version).</summary>

```json
{"404":{"title":"The content you’re looking for seems to have disappeared.","main":"Go to {link} or search for something similar from the field below."},"hero":{"subtitle":"Explore more than 800 million creative works","description":"An extensive library of free stock photos, images, and audio, available for free use.","search":{"placeholder":"Search for content",},"disclaimer":{"content":"All {openverse} content is under a {license} or is in the public domain.","license":"Creative Commons license"}}}
```

</details>

<details>
<summary>The flattened example minifies to 527 characters (expand for minified version).</summary>

```json
"404.title":"The content you’re looking for seems to have disappeared.","404.main":"Go to {link} or search for something similar from the field below.","hero.subtitle":"Explore more than 800 million creative works","hero.description":"An extensive library of free stock photos, images, and audio, available for free use.","hero.search.placeholder":"Search for content","hero.disclaimer.content":"All {openverse} content is under a {license} or is in the public domain.","hero.disclaimer.license":"Creative Commons license"}
```

</details>

Because this change is proposed to improve authorship (not transport or anything else relevant to the final generated files), and because it is backwards compatible with the nested format, we should retain the nested format for the produced JSON files.

To implement this change, we will need to flatten the `en.json5`'s keys. This can be done by hand, or using something like [this online JSON flattening tool](https://www.coderstool.com/flatten-json), except that tool and others like it strip comments and make other unwanted transformations, so would still require manual changes to address those... Because `json-to-pot` already has to flatten the keys for use in the POT files, it shouldn't be too hard to adapt our existing `json-to-pot` script into a `json-to-flattened-json` script that preserves comments, single-quoted strings (which we use to avoid needing to escape double quotes in some strings), etc. I would recommend that approach, it seems like the least tedious option to me!

### LINKED ISSUE #5059: Incorrect slug for Kyrgyz locale
## Description

Kyrgyz locale should have `ky` as slug, but in Openverse we incorrectly use `kir`.

According to [Chrome for Developers documentation](https://developer.chrome.com/docs/lighthouse/seo/hreflang#guidelines_for_hreflang_values), `hreflang` should use ISO 639-1 code for language, which is `ky` [^1], [^2].

[^1]: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
[^2]: https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?iso_639_1=ky

This value comes from GlotPress:

https://github.com/GlotPress/GlotPress/blob/0de4f9656a954132becffbccca8cc336143c93e2/locales/locales.php#L1604-L1614

## Screenshots

The error as reported by Lighthouse:

Image

## Additional context

The latest change to Kyrgyz locale properties in GlotPress were done in [Update Kyrgyz locale information](https://github.com/GlotPress/GlotPress/pull/550). However, the changes don't fully match the [locale editor request](https://make.wordpress.org/polyglots/2014/10/24/hi-can-someone-change-kyrgyz-languages-country-code/#comment-261769).

obulat:
## Fixes

Fixes #4979 by @sarayourfriend
Fixes #629 by @obulat
Fixes #5059 by @obulat
Related to #592 by @sarayourfriend 

## Description

This PR re-writes the i18n locale metadata and translation files generation.

### Folder structure
The recent update in the Nuxt i18n has introduced a new top-level i18n folder. Now all i18n-related files are located there:
- `i18n/data` folder contains the source-of-truth original `en.json5` file. The locale metadata (`valid-locales.json` and `untranslated-locales.json`) are saved there as well.
- `i18n/locales` folder is where all downloaded translations are saved.
- `i18n/scripts` folder contains the scripts that download  locale metadata, translations, convert `en.json5` to `.pot` file that is used for translations on https://translate.wordpress.org

### Code modernization
Now that it's easy to run modules in Node, the scripts can be converted to `.mjs` and use `import` instead of `require`. This PR contains the conversion, as well as migration from `axios` to native `fetch`. 

### Flattening of the keys
To simplify finding static and dynamic keys in the source code, as well as the conversion from and to `.pot` files, the keys of the translation strings have been flattened. So now, `en.json5` is a flat JSON file, where all keys are top-level. This required the update to the ESLint rule for the json5 keys.

### Description of changed files (what to review)
The main entry point for the i18n scripts is now `i18n/scripts/setup.mjs`. This file reads the CLI options, and creates the necessary i18n files. All modes (English-only, production - downloading all locale metadata and translations, testing - copying of the local files from the test folder) are now combined in a single workflow.

The order of the steps for production was changed because the `translated` property for the locale metadata is now generated based on the number of keys in the locale's json file, so we need to have the translations before creating the list of available locales.

Previously, the steps were:
1. get the list of locales and their properties from the GlotPress GitHub repository
2. parse the translation statistics from the Openverse project page on https://translate.wordpress.org and add the `translated` property to the locales from step 1.
3. download all translations in a zip and convert the contained flat `json` files into nested json.
4. save the list of locales for which the translations were downloaded to `valid-locales.json` file, and the locales without translations to `invalid-locales.json`.

In this PR, step 2 was eliminated, and now the process is:
1. (previously, step 3) download all translations and save the _non-empty_ json files. We no longer need to convert flat json files, and we don't need to save empty files as we are doing now.
2. (previously, step 1) get the list of locales and their properties from the GlotPress GitHub repository, and compute the translated percentage (comparing the number of keys in the original English en.json5 file and the translated json file.
3. (previously, step 4) save the list of locales for which the translations were downloaded to `valid-locales.json` file, and the locales without translations to `invalid-locales.json`.

The code for the new step 1 is located in `i18n/scripts/translations.mjs`, for the new step 2 and 3 - in `i18n/scripts/metadata.mjs`.

The code for generating the pot file was also simplified now that we don't need to handle nested json, and is now located in `i18n/scripts/generate-pot.mjs`.

All of the paths for files are now located in a single file, `i18n/scripts/paths.mjs`, so it will be easier to safely move a file and update its paths in a single place.

The i18n check script was removed from `frontend/package.json` because it wasn't working, and the library doesn't seem to be maintained.

## Testing Instructions

Run the i18n scripts, deleting the locale data between the runs to ensure that existing data does not interfere.
- "i18n" - the main script, by default generates all the data needed for production. Can also be used with flags.
- "i18n:debug" - the same as `i18n`, but has more verbose logging, including the logging of keys/placeholders that have been modified to fix one of the linked GitHub issues 
- "i18n:en" - only generates the `en.json` file and an empty `valid-locales.json`. This can be used for quickly running local environment, including with the `--watch` flag (that watches the `en.json5` file and updates `en.json` when the original file changes)
- "i18n:test" - copies the files from `test/locales` into correct folders
- "i18n:no-get" - used to generate `valid-locales.json` file based on existing translation files, without downloading translations from GlotPress
- "i18n:generate-pot" - converts `en.json5` file to `pot` file. The generated file should be the same as the one generated on `main`.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


documentation/frontend/reference/i18n.md
```diff
@@ -78,7 +75,7 @@ processed and loaded into Nuxt via the Nuxt i18n module.
 ## Test locales
 
 Three locales are maintained in the code base (as opposed to downloaded from
-Glotpress) for use in testing, alongside the default English locale. These
+GlotPress) for use in testing, alongside the default English locale. These
```
sarayourfriend, line 78: tip: You can add GlotPress to the Vale `TermCasing.yaml` so these errors are found automatically in the future.

packages/js/eslint-plugin/src/rules/key-name-casing.ts
```diff
@@ -0,0 +1,161 @@
...
+import { JSONProperty } from "jsonc-eslint-parser/lib/parser/ast"
+
+import { OpenverseRule } from "../utils/rule-creator"
+
+export type CasingKind = "camelCaseWithDot" | "snake_case_with_dot"
+
+export const allowedCaseOptions: CasingKind[] = [
+  "camelCaseWithDot",
+  "snake_case_with_dot",
+]
```
sarayourfriend, line 11: Just a tip, if you want to avoid duplicating the names in this definition, you can derive `CasingKind` from `allowedCaseOptions`:

```suggestion
export const allowedCaseOptions = [
  "camelCaseWithDot",
  "snake_case_with_dot",
] as const

export type CasingKind = typeof allowedCaseOptions[number]
```

Alternatively, you can derive both from the `checkersMap`:

```typescript
const checkersMap = {
  camelCaseWithDot: isCamelCase,
  snake_case_with_dot: isSnakeCase,
} as const

export type CasingKind = keyof typeof checkersMap

export const allowedCaseOptions: CasingKind[] = Object.keys(checkersMap)
```

packages/js/eslint-plugin/src/rules/key-name-casing.ts
```diff
@@ -0,0 +1,161 @@
...
+  snake_case_with_dot: isSnakeCase,
+}
+
+/**
+ * Checks whether the given string has symbols.
+ */
+function hasSymbols(str: string) {
+  return /[\u0021-\u0023\u0025-\u002c./\u003a-\u0040\u005b-\u005e`\u007b-\u007d]/u.test(
+    str
+  ) // without " ", "$", "-" and "_"
```
sarayourfriend, line 24: Are these ranges inclusive of the bounds? When I check https://symbl.cc/en/unicode-table/, 0025 is the space character. Should the the second range be updated to 0026?

Also: I'm not sure whether the inclusion of `.`, `/` and `` ` `` separate from the unicode ranges is intentional, but if it is, it might be helpful to move them to the start or end of the bracket group so they do not get lost.

sarayourfriend, line 24: Because this rule is only used for locale json files, it might be easier to understand this particular check by inverting it to check the characters that are allowed in a locale key. Would that be, `[a-zA-Z._]`, for example? If this rule is only used for locale key strings, and that would allow simplifying aspects of it by making it more exact, then it's worth doing. This regex is a little obscure as it is :sweat_smile: 

In that case, it would also be good to change the name of the rule to `@openverse/locale-key-name-casing`.

packages/js/eslint-plugin/src/rules/key-name-casing.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Checks whether the given string is snake_case.
+ */
+export function isSnakeCase(str: string): boolean {
+  return !(hasUpper(str) || hasSymbols(str) || /-|__|\s/u.test(str))
+}
+
+/**
+ * Return case checker ('camelCaseWithDot', 'snake_case_with_dot')
+ */
+export function getChecker(name: "camelCaseWithDot"): (str: string) => boolean {
```
sarayourfriend, line 49:
```suggestion
export function getChecker(name: CasingKind): (str: string) => boolean {
```

sarayourfriend (commented):
This is cool! Thanks for the ping on the issues closing. Did a quick little review and have some simple suggestions. What a nice clean up and reorganisation :slightly_smiling_face:

packages/js/eslint-plugin/src/rules/key-name-casing.ts
obulat: I copied the code as-is from the key name casing rule, only adding the `.` :) Great suggestions, thank you!

frontend/i18n/scripts/po/parse-vue-files.mjs
```diff
@@ -1,32 +1,40 @@
...
 
-const BASE_PATH = path.join(
-  path.dirname(path.dirname(path.dirname(__dirname))),
-  "src"
-)
-
-function readVueFiles(src) {
-  const targetFiles = glob.sync(src)
+/**
+ * Parses all vue files found in the glob paths, and returns an
```
dhruvkb, line 11: Maybe we should add to `TermCasing.yml` hehe:
```json
"vue": "Vue"
```

dhruvkb (approved):
The changes make sense, and it's quite nice that we're able to use `.mjs` files now and all the benefits they come with.

I have one (semi-serious) suggestion, but the PR is good enough to merge as-is.

Issue #5254, opened 2024-12-05
Need to Add https instead of http in design channel slack link


wpeople-dev:
## Description

- Need to Add https instead of http in design channel slack link in documentation/general/contributing.md file to avoid unnecessary redirection 

## Reproduction

1. Open documentation/general/contributing.md file
2. Go to Line 121 & 135
3. See error.

PR #5255, opened 2024-12-05
Added https instead of http in wordpress design channel slack link 

wpeople-dev:
Fixes: https://github.com/WordPress/openverse/issues/5254


dhruvkb (approved):
Good catch!

Issue #3473, opened 2023-12-06
Proxy frontend API requests through Nuxt


sarayourfriend:
## Problem

Our approach to anonymous rate limiting inadvertently affects single IP address institutions like schools and libraries. In these cases, all users on their networks are rate limited as a single user. We've observed this happening especially when the referrer is openverse.org, meaning these are likely to be regular frontend users getting rate limited.

## Description

The Openverse frontend should proxy API requests and forward the SSR server's API key with requests. This would prevent anonymous rate limiting applying to API requests from the frontend and would allow us to safely turn anonymous rate limiting back on in the API.

If it's quick to do, using a CSRF token + cookie combination would make this slightly more resilient to attacks, but we can do that in a follow up PR if it's too complex to implement right away.

## Documents

_Due to the small size of this project, a project plan can be skipped and only an implementation plan should be necessary_

- [ ] Implementation Plan(s)

sarayourfriend:
This is _way_ harder than I thought it would be, mostly because of the api service object pattern.

I think things could be simplified quite a bit (type wise and code wise) if we refactored to get rid of the service object pattern and instead had a single configured axios instance injected into the Nuxt context (which would have the base URL and the access token when needed), and then helper functions to make specific requests as needed. The current pattern is a weird mash up of functional and OOP that's resulted in some really confusing patterns. Most of that comes from creating the stateful api service object in multiple different places with slightly different configurations. A set of helper functions when needed that take in a configured axios instance and make the call as needed, returning the types as needed as well, would be a big cleaner and would eliminate the multiple-entrypoint issue with axios right now.

I don't know what the right thing to do here is though. @obulat @dhruvkb and @krysal what do y'all think? How would y'all proceed?

sarayourfriend:
Closing this in favour of https://github.com/WordPress/openverse/issues/2037. I'll update this comment later with an explanation, just wanted to get this off the project tracker for while we're looking through projects after the first few months of 2024.

sarayourfriend:
I am reopening this as an independent project, because it will ultimately be necessary to solve this for the frontend no matter what. There is, unfortunately, no way around it. However, I think the work with #4908 sets a decent precedent of what something _could_ look like for light-weight proxying of API requests, that doesn't let someone easily hijack the frontend to make unlimited requests to the API (and which, anyway, we will continue to have WAF-level rate limits on the frontend to prevent the most egregious examples of that).

PR #5265, opened 2024-12-11
Implementation Plan: Proxying frontend API requests through the Nuxt server

obulat:
## Due date:

2024-01-11

## Assigned reviewers

- [ ] @krysal
- [ ] @dhruvkb 

## Description

Related to  #3473

## Current round

This discussion is following the Openverse decision-making process. Information about this process can be found

This discussion is currently in the **Decision round**. The deadline for review of this round is **2024-01-11**

## Revision details
[2024-01-08] I updated the plan based on the comments. The main changes are:
1. **Proxy All Frontend Traffic:** Updated the plan to proxy all frontend routes, not just the /api routes. This will make the rate-limiting setup easier and eliminate the need to integrate modules like Nuxt Turnstile. Handling Cloudflare challenges directly in the Nuxt server avoids the complexity of requiring the Pinia media store to process challenge responses instead of result responses.
2. **Use `h3` session for identifying users**: The plan now uses h3 sessions to create a verifiable session header/cookie. Cloudflare will validate that the header/cookie exists, allowing requests to pass even if the cookie is invalid (e.g., generated by automated requests). The Nuxt server will then decode the cookie using the secret and verify its validity. Invalid cookies will trigger a 401 Unauthorized response from the Nuxt server, preventing these requests from reaching the Django API. This ensures robust protection against automated abuse.
3. **Shared IP scenario testing:** Detailed a method to test shared IP scenarios using two different browsers or an incognito window to simulate multiple users. This approach demonstrates how authenticated users bypass stricter limits while anonymous users face challenges if limits are exceeded.
4. **Excluding static asset routes:** Explicitly stated that static asset requests (e.g., `/_nuxt/`, `.css`, `.jpg`) are excluded from rate limiting to streamline traffic and reduce unnecessary challenges.

Here's a branch with a draft of changes implementing this project (single commit for all changes): https://github.com/WordPress/openverse/tree/add/api-proxying

### DIFF
--- a/documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
+++ b/documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
@@ -5,7 +5,7 @@
 ## Reviewers
 
 - [ ] @dhruvkb
-- [ ] @krysal
+- [x] @krysal
 
 ## Project Links
 


documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+- Nuxt CPU and memory usage. In case the server is under heavy load, we might
+  need to adjust the rate limits and/or frontend task values.
+
+## Rejected Alternatives
+
+- **Always Issuing a Challenge Without Thresholds**: Issuing a managed challenge
+  to every new user, regardless of load or conditions, would inconvenience users
+  behind NATs and degrade the initial user experience. While simpler to
+  implement (we would not need to calculate the exact rate limit), it fails our
+  design goal of minimizing friction for normal users.
```
sarayourfriend, line 343: > Issuing a managed to every new user, regardless of load or conditions

Can you explain further why every new user would be issued a challenge? For what it's worth, the managed challenge "should" most of the time be a JavaScript challenge, which requires no user interaction.

To reduce friction without needing to introduce the turnstile module and involving the application code, you could have a separate (or no) rate limit for the home page, relying on that being the most common entry point for users where they would receive the original cookie to carry for the rest of their session.

To clarify the situation behind a NAT, it might be overly cautious to think that the rate limit will be tripped so quickly. Imagine a large school with maybe 3 computer labs having a simultaneous class, 50 students each. That is 150 fully unauthenticated requests from a single IP that need to happen for everyone to get a cookie. If the rate limit of a cookie-less request is set to 150 requests / 30 minute period, and it just takes a managed challenge to clear it for the IP, I don't think this would introduce very much friction at all.

For users on low spec Android phone, they would probably avoid ever seeing the managed challenge and get the cookie, and then just be subject to the regular rate limit.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+        NS ->> A: authenticated request
+        A ->> NS: response
+        NS ->> O: response
+    end
+```
+
+## Cloudflare Rate Limiting
+
+Originally, I thought of using Cloudflare managed challenges to protect the Nuxt
+server routes from abuse. However, this would have required issuing a challenge
```
sarayourfriend, line 88: A clarification here: "to protect the Nuxt server routes from abuse".

For the solution to work, it's necessary for _all_ routes to be behind the same Cloudflare rate limit rules, not just the server routes. That way, if someone gets rate limited on the server routes, the frontend can reload the page and the user will actually get the managed challenge page for clearing the rate limiting (or the explanation page if they are going too fast and need to actually be limited).

If only the server routes are limited by the new configuration, then there's no way to present the user with a challenge that will clear that rate limit, and the user would be stuck.

I mention in another comment that I don't think this would require issuing a challenge to every new user and would not meaningfully increase friction in the vast majority of even extreme cases. Regardless, it is necessary for all routes to have the same rate limit rules for this Cloudflare-rate-limit strategy to work with the safe guard of users being able to clear the rate limit as needed via the challenge.

sarayourfriend (commented):
I've left some comments suggesting clarifications or alternatives. Happy to chat more about this. Excited to see this project come to light, I think it will massively improve the user experience for a major category of Openverse users (in education) as well as make the project's ingress more sustainable and manageable for the maintainers. Kudos on the IP!

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+- Create a helper function that accepts the H3 `event` object, and extracts the
+  request details to forward to the API:
+  - media type
+  - the API media type slug (i.e., the special case of "images" for "image")
+  - media id, if applicable
+  - search query
+  - headers that should be passed on to the API
+- From the
+  [k6 implementation](https://github.com/WordPress/openverse/pull/4908), copy
+  the helper function that signs the request with HMAC.
```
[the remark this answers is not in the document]
obulat: I wrote a long reply first, and then realized that my thoughts were valid if only the `/api/` routes are behind the rate limiting.  If _all_ frontend routes are behind the rate limit rules, we don't need to do anything special for protecting the `/api/` routes (that can send unlimited requests to the API) from automatic abuse, as they would be behind the frontend rate limiting rules.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+
+The server route handler should:
+
+- generate the relevant API request URL (i.e., converting `/api/image/?q=cat` to
+  `<apiUrl>/v1/images/?q=cat`)
+- add the headers (headers from the original client request that can be proxied,
+  authentication, HMAC)
+- send the request using `ofetch` (or `ofetch.raw` to extract the response
+  headers for `SEARCH_RESPONSE_TIME` Plausible event)
+- handle errors and return the appropriate response to the client
```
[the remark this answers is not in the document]
obulat: Good idea. I had to write the draft of the implementation to see how it would work.

- in the beginning of the server setup (e.g., in the `init-stores.server` plugin, if it's renamed `init.server`), set the `sessionId` cookie that is `httpOnly` (if it does not exist).

- send the request from the store to the `/api` route. 
  - On the client, the `httpOnly` will automatically be sent. 
  - On the server, the request will be sent between the different parts of the Nuxt app, without the browser passing the cookies. We need to specifically pass the cookies using `getProxyRequestHeaders` from `h3` to extract them and pass them on with the `$fetch` request. Since we are on the server, the `httpOnly` cookies are also available.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+The action should be `managed_challenge`, which will issue a challenge to the
+user when the rate limit is exceeded.
+
+#### Testing the rate limits
+
+When testing this feature, we can adjust the staging limits using the UI. I
+envisage that we will need to adjust the limits many times because testing the
+Cloudflare rate limiting locally is not possible.
+
+### Use server routes in the frontend when the feature flag is on
```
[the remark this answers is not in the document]
obulat: I think using the Nuxt-native methods would be better for logging and error handling of the requests on the server.
During the transition period, there would need to be separate branches for requesting the proxied/non-proxied urls. However, it would be more convenient to handle errors within the server route, where we can log all requests and errors on the server.

We could use the same naming for the server routes as the API has (so, instead of `/api/search/image/?q=cat` have `/v1/images/?q=cat`), which would make the conversion of the URLs simpler.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
sarayourfriend: Gotcha! I was not connecting the dots on the turnstile handling the flow of presenting the challenge for the subset of rate limited routes.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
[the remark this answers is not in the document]
sarayourfriend: Fair enough! For what it's worth, it's possible to swap the underlying `fetch` implementation to pass options through to Nuxt's version, if it has better logging and other options. [Check out the `fetch` option to `createClient` in the docs](https://openapi-ts.dev/openapi-fetch/api#createclient).

The main benefit of using `@openverse/api-client` is unifying the API and types, and it provides a convenient way to swap out the target URL as needed. It would be possible to re-write the paths dynamically in the function passed to `fetch`, so that the Nuxt routes don't need to reproduce exactly the same paths as the upstream API.

Anyway, just an alternative idea in case it is useful. The plan you've got doesn't look bad to me at all :slightly_smiling_face:

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+When the Cloudflare rate limiting triggers a managed challenge, the client will
+receive an HTML Cloudflare challenge response. To handle this response, we can
+use the [Nuxt turnstile module](https://nuxt.com/modules/turnstile).
+
+This is the part of the plan I'm most unsure about. I couldn't understand the
+details of how this module works, so we would probably need to do a lot of
+testing to make sure it works as expected. Unfortunately, we cannot test
+Cloudflare challenge responses locally because the proxy is only set up in
+staging/production. We will have to use staging for this, and to make the
+testing easier, we can set the staging rate limits to be very low.
```
krysal, line 304: Understandably, this module needs further research. Also, preferably, there should be a way to keep the frontend working locally, so maybe turning this module off or bypassing it?

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+Currently, client requests[^1] from the frontend to the API are sent without
+authentication. As a result, strict API rate limiting on unauthenticated
+requests would inadvertently throttle legitimate traffic from openverse.org
+users. By proxying all client requests[^1] through the Nuxt server, we can
+authenticate them and thus distinguish them from the anonymous traffic in the
+API.
+
+This implementation plan outlines how we will route all frontend traffic to the
+API through Nuxt server routes — effectively “authenticating” these requests at
+the API level — and configure Cloudflare rate-limiting to protect these routes
```
krysal, line 29: Would requests from the same user also be identified at the API level? I understand that everything will be seen as "coming from the frontend." For example, is the cookie value passed to the API, too?

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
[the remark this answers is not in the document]
krysal: @sarayourfriend, your bullet points helped me picture this and made me realize that the cookie is probably an important part I didn't get. Thank you for summarizing it.

krysal (commented):
Thank you for writing this detailed plan, @obulat. It's a nice explanation of the steps to incorporate Cloudflare-provided rate-limiting for the frontend and better control the use of Openverse. It sounded quite challenging at first, but I now have a clearer idea of what it entails after reading it. I only have a couple of loose points, namely:
- The cookie that will help identify each user
- How will be the local usage of Openverse after these changes are applied

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+- **Always Issuing a Challenge Without Thresholds**: Issuing a managed challenge
+  to every new user, regardless of load or conditions, would inconvenience users
+  behind NATs and degrade the initial user experience. While simpler to
+  implement (we would not need to calculate the exact rate limit), it fails our
+  design goal of minimizing friction for normal users.
+
+- **Proxying thumbnail requests** The thumbnail requests from the frontend will
+  remain anonymous on the API level. Since controlling search and related
+  requests naturally throttles excessive thumbnail retrieval, we consider it
+  unnecessary to further load the Nuxt server.
```
dhruvkb, line 348: Is this also similarly applicable for audio waveforms?

dhruvkb (commented):
I don't have much to add beyond what @sarayourfriend and @krysal have already covered. So 👍 from me to move into the next round of incorporating their feedback.

The turnstile module was a confusing aspect of the plan for me but it seems from existing feedback that a second version of the IP will not be needing it. (I will still check out the module, in case it still ends up being used for the plan).

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
obulat: Yes, I forgot about the waveforms 🤦

krysal: I wonder the same as Dhruv.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
...
+    send directly to Openverse API without any authentication, because using the
+    API credentials in the client code would expose them to the public.
+
+[^2]:
+    [Cloudflare multi-language support](https://developers.cloudflare.com/waf/reference/cloudflare-challenges/#multi-language-support)
+
+[^3]:
+    [Cloudflare rate limiting rules](https://developers.cloudflare.com/waf/rate-limiting-rules/)
+
+[^4] [Nuxt $fetch](https://nuxt.com/docs/api/utils/dollarfetch)
```
krysal, line 450:
```suggestion
[^4]: [Nuxt $fetch](https://nuxt.com/docs/api/utils/dollarfetch)
```

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,396 @@
...
+    end
+```
+
+## Cloudflare Rate Limiting
+
+Originally, I thought of using Cloudflare managed challenges to protect the Nuxt
+server routes from abuse. However, this would have required issuing a challenge
+to every new user, which would have been a poor user experience. Instead, we
+will use Cloudflare rate limiting to protect the Nuxt server routes from abuse.
+The rate limiting will be based on the IP and cookie combination, and will
```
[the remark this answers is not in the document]
krysal: For the record, this is replied in the new version, specifically, in the "Add session to the init plugin" section, as the session cookie is generated and handled by the `unjs/h3` utility.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
krysal: Ditched in the new version. Great! 💯

krysal:
I'm glad the proposal has been simplified by removing the Nuxt Turnstile module requirement 😄 I have a small question about the number of sessions, other than that, there are minor comments/suggestions. Once clarified, I expect a quick approval. Thanks for writing this, @obulat. It looks very promising.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
...
+To test that the rate limits work as expected in the shared IP scenario, we can
+simulate two users accessing the service from the same IP address:
+
+1. Use two different browsers (or one browser with an incognito window) to
+   represent two users.
+2. User 1: In one browser, simulate a user with a valid session cookie,
+   confirming they are treated as authenticated and bypass stricter rate limits.
+3. User 2: In the other browser, simulate an anonymous user without a session
+   cookie. Verify that their requests are subject to stricter IP-based rate
+   limits.
```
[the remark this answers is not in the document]
obulat: Right, simply using the frontend app should add the session cookie. So, what we can test here is the fact that even if the same IP sent more requests than allowed by the rate limit, the next user with the same IP can see the challenge and use Openverse after the challenge is passed.

Here's how it can be done:
1. Set the stagomg rate limit to something like 2 requests per 10 minutes
2. Open several tabs in incognito mode, and visit the staging app. After the first two tabs, you should see the Cloudflare challenge. It should be easy to pass the challenge.

To test that it's not easy to pass the rate-limiting, we can send 3 requests to staging. The 3rd request should return the Cloudflare challenge, and not pass through the Cloudflare proxy.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
...
+- initialize the
+  [`OpenverseApiClient`](https://github.com/WordPress/openverse/tree/347878acf7620548a0628d6833885a479c30b9c1/packages/js/api-client).
+  If the API secrets are set, use authenticated client. Otherwise, in
+  development environment, use unauthenticated client, and in
+  staging/production, throw an error, as the API requests from the frontend
+  should always be authenticated. To access the secrets, the plugin should use
+  the `process.env.[VARIABLE_NAME]`, not `runtimeConfig.VARIABLE_NAME` since the
+  `runtimeConfig` is not available in the plugin body.
+- add a middleware using `defineEventHandler` that adds the `OpenverseApiClient`
+  as to the event context as `event.context.client`.
```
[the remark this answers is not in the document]
obulat: I like `api_client` as it's specific, clear and still short.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
...
+
+### Set up Nuxt Server proxy routes
+
+- Create the Nuxt server routes under `/api/`:
+
+  - `/api/search/[type]` - for search requests
+  - `/api/[type]/[id]` - for single result requests
+  - `/api/[type]/related/[id]` - for related media requests
```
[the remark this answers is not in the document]
obulat: This might make it more difficult to handle. With the proposed structure, the single and related pages would be `/api/[type]/[id].ts` and `/api/[type]/related/[id].ts`. If we move the `[id]` in the related route, the single result route might also catch the related requests because it could match on `/api/[type]/[id].ts`. Can we keep this as is, and change it when implementing, if possible, @krysal ?

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
...
+  IP see the managed challenge more often, if the IP is sending many requests.
+  However, after the first request such a user will be subject to the first rate
+  limiting counts (IP and session cookie).
+
+```
+(http.host eq "staging.openverse.org"
+and (
+    (http.request.headers["x-h3-session"][0] ne "")
+    or
+    (http.request.headers["Cookie"] contains "ovSession")
```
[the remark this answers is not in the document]
obulat: There will be only one session cookie. Did I use different names throughout the IP? That was a mistake, if so. I'll fix it.
However, there should be both the header and the cookie, because `h3` `useSession` sets both, and the priority is given to the header.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
obulat: I just realized that we request the waveforms when we add the `peaks=true` param to the audio requests. They are added to each audio object in the `peaks` array. So, there is no separate request.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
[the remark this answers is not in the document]
krysal: I see. I thought Nuxt routing could handle it somehow, but sure, let's keep it as is and evaluate it during the implementation 👍

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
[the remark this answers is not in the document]
krysal: Perfect, thanks for clarifying!

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
krysal: @obulat There is also the waveform endpoint, `https://api.openverse.org/v1/audio/<uuid>/waveform/` Isn't it used in the single audio page?

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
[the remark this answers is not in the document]
obulat: I tried it locally, and Nuxt _can_ handle the routes like this:
- `/api/[type]/search.ts` - for search
- `/api/[type]/[id]/index.ts` - for individual results
- `/api/[type]/[id]/related.ts` - for related results.
I will update this, thank you for the suggestion!

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
krysal: Understood, great! Thanks for clarifying.

documentation/projects/proposals/proxy_frontend_api_requests/20241202-implementation_plan_proxy_frontend_api_requests_ip.md
```diff
@@ -0,0 +1,450 @@
+# 2024-12-09 Implementation Plan: Proxy Frontend API Requests Through Nuxt
+
+**Author**: @obulat
+
+## Reviewers
+
+- [ ] @dhruvkb
+- [ ] @krysal
```
krysal, line 8:
```suggestion
- [x] @krysal
```

krysal (approved):
No more questions from me. The plan looks great!

PR #5288, opened 2024-12-18
Add sources cache to Nuxt app

### LINKED ISSUE #5287: Cache sources stats in Nuxt
## Problem

Currently, we send a `/stats` request to the API on every frontend incoming request. This is unnecessary and wasteful.

## Description

Add a Nuxt server cache to only load the stats once an hour.  

## Alternatives

- Save the data during the build time. This will only allow us to update it once a week during deploy, which is not often enough.

obulat:
## Fixes

Fixes #5287 by @obulat

## Description

This PR uses the server route with `defineCachedFunction` to cache the `/stats` response for an hour. This is following SWR (stale-while-revalidate) principle, so if a request to update the stats fails, the cached (even outdated) version is still returned.

This PR also simplifies the provider store to combine the fetch state for images and audio into a single object since we don't really need the distinction between the media types.

The requests don't add the auth token to the API request, but it should be okay because the request is only sent once an hour.

## Testing Instructions

Run the app locally, and open the `localhost:8443/sources` page. You should see the log about `/stats` requests (for audio and images) in the terminal (not the browser console). Now, reload the page several times. There should be no other requests logged, since the data is cached in the nitro server.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/frontend/server/api/sources.ts
+++ b/frontend/server/api/sources.ts
@@ -1,76 +1,51 @@
-import { ofetch } from "ofetch"
 import { consola } from "consola"
-import { defineEventHandler } from "h3"
-import { useRuntimeConfig } from "nitropack/runtime"
-import { useStorage } from "nitropack/runtime/storage"
+import { useRuntimeConfig, defineCachedFunction } from "nitropack/runtime"
+import { defineEventHandler, getProxyRequestHeaders, type H3Event } from "h3"
 
-import { supportedMediaTypes } from "#shared/constants/media"
-import type { SupportedMediaType } from "#shared/constants/media"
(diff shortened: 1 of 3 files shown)


dhruvkb (approved):
LGTM! I was able to verify that this works as expected locally. The cache seems to persist across frontend server sessions as it did not make the API call after I terminated the server with <kbd>⌘</kbd>+<kbd>C</kbd> and started it again.

PR #5323, opened 2025-01-08
Simplify media `FetchState` and fix fetching errors

### LINKED ISSUE #5322: Simplify `fetchState` in the Nuxt app
## Problem

The current structure of `fetchState` uses multiple flags (e.g., `isFetching`, `hasStarted`, `isFinished`, `fetchingError`) to manage the state of API requests. This approach:

- Increases complexity and redundancy.
- Makes it harder to reason about the current state of a request.
- Lacks explicit type safety, leading to potential inconsistencies (e.g., `isFinished: true` with a non-null `fetchingError`).

## Description

Refactor all `fetchState` objects to use a **discriminated union**, which explicitly defines possible states and ensures type safety. The new structure will look like this:

```typescript
export type FetchState =
  | { status: 'idle' | 'loading' | 'success'; error: null }
  | { status: 'error'; error: FetchingError };
```

### Benefits

- Simplified state management: The status field alone determines the fetch state, eliminating the need for redundant flags.
- Explicit error handling: The error field is only available when status is 'error', reducing ambiguity and improving clarity.
- Type safety: Ensures that invalid combinations (e.g., status: 'success' with error) are impossible.

This change will apply to all instances of fetchState objects across the codebase.

## Alternatives

Keep the Current Structure

- Pros:
  - No immediate changes required.
- Cons:
  - Complexity and redundancy remain.
  - Potential for invalid state combinations persists.

### Derived states (getters) for the UI

This change will also simplify the following common derived states:

#### `canLoadMore`
Determines whether to display the "Load More" button:
```typescript
const canLoadMore = (currentPage: number, totalPages: number) => {
  return currentPage < totalPages;
};
```

#### `showLoadingSkeleton`

Loading skeleton is only displayed for the first page being loaded. This getter will be used to determine whether to display it:
```typescript
const showLoadingSkeleton = (fetchState: FetchState, currentPage: number) => {
  return fetchState.status === 'loading' && currentPage === 0;
};
```

#### `showNoResultsPage`

Determines if the “No Results” page should be displayed:

```typescript
const showNoResultsPage = (fetchState: FetchState, totalResults: number) => {
  return fetchState.status === 'success' && totalResults === 0;
};
```

### Derived state for the store
To determine whether to send a request for the next page for specific media type, the following getter can be used:
```typescript
const canFetch = (type: MediaType) => {
  const fetchState = this.mediaFetchState[type];

  return (
    fetchState.status !== 'error' && // No unretriable error
    fetchState.status !== 'loading' && // Not already fetching
    !(this.results[type].page >= this.results[type].pageCount) // Not finished
  );
};
```

### LINKED ISSUE #5325: Results replaced with No result error page when clicking Load more if second page API result has no results
## Description

When the API returns 240 results for page 1, and then no results for subsequent pages (possibly due to dead link filtering), the frontend on "Load more" button click replaces the initial results with No results page. This is a very confusing user experience.
Instead, we should remove the "Load more" button and replace the result count with the updated one.

### All media view

1. Visit https://openverse.org/search?q=bamberg
2. Click on "Load more" button.
3. See error: the "Image" content link at the top shows "No result"
_Expected behavior: the number of results should be updated to the count of currently loaded results._

Image

### Images view

1. Visit https://openverse.org/search/image?q=bamberg
2. Click on "Load more" button.
3. See the No result error page (_We couldn't find anything for "Bamberg"_)
_Expected behavior: the first page should stay open, the Load more button should be hidden, and the result count in the search bar should be updated._

obulat:
## Fixes
Fixes #5322
Fixes #5325

## Description
This PR replaces the flags used in `FetchState` with single `status` property, and an additional `error` property. To prevent invalid states where status is not `error`, but there's error, the type uses TypeScript discriminated union.

Some other changes:
- getters for the derived states added or updated in the media store to decide:
  - whether to show the load more button
  - whether to show the result count or "Loading" in the results labels
- `fetchMedia` returns the `Results` type that is a discriminated union (so, the `type` property determines the type of the items in the `items` list)
- moved the `canLoadMore` and i18n label computation out of the component to the parent components. This simplifies the component rendering in the Storybook.

## Testing instructions
Run the app using `ov just frontend/run dev`
Search for "bamberg" and try the steps in #5325 - the errors from the issue should be fixed. 
Fetching the media should work properly, both in the client and in SSR.
When the first page of results is loading on All media content, the content links should not say "No results", but should show "Loading..." label instead.

---

I tried creating a PR completely using the GitHub copilot, but I had to add a lot of changes on the first version: [Copilot Workspace session](https://copilot-workspace.githubnext.com/WordPress/openverse/pull/5323?shareId=3a9b9d4f-365a-49db-9e87-3879f88e90c0).


frontend/.storybook/main.ts
```diff
@@ -1,7 +1,7 @@
 import type { StorybookConfig } from "@storybook-vue/nuxt"
 
 const config: StorybookConfig = {
-  stories: ["../src/**/*.mdx", "../src/**/*.stories.@(js|jsx|mjs|ts|tsx)"],
+  stories: ["../src/**/*.stories.@(js|jsx|mjs|ts|tsx)"],
```
obulat, line 4: We don't have any `.mdx` files, so Storybook was always printing warnings about this pattern.

dhruvkb (approved):

PR #5261, opened 2024-12-10
Clean repeated static string interpolation

tejaswarathe:
## Fixes

Fixes #4980 by @sarayourfriend 

## Description

The changes here is to remove the repeated interpolation strings as mentioned in #4980 

I have changed the method replacer() in translations.mjs which is used to replace the string with {} brackets while the pot files are converted to json for each language from Glotpress.

I have also added an object targetWords where we can add the world which we do not want to translate. Further work required is just to change the en.json5 file and change the strings like `_{github}_ `to `_{'github'}_`  and adding that string to targetWords object. This will make sure that the translations also will keep the string within single quotes ('') to consider them as string literal.

Currently the changes have only been done for `openverse` string.

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (`ov just catalog/generate-docs` for catalog
      PRs) or the media properties generator (`ov just catalog/generate-docs media-props`
      for the catalog or `ov just api/generate-docs` for the API) where applicable.

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
[...]

PR #5343, opened 2025-01-22
Fix i18n placeholder replacement and make tests robust

obulat:
## Fixes

Follow up fixes after #5261 and #5323

## Description

The main reason for this PR was to fix the CI failures. The visual regression tests failed because the locale fallback mechanism broke, so when the locale (Arabic for RTL) string wasn't available, instead of falling back to the English, the i18n keys were used. 

The cause was the update of Nuxt i18n and Nuxt in the CI. Playwright tests in CI did not use the locked dependencies from `pnpm-lock.yaml`, and were resolving dependencies differently. So, instead of the Nuxt 3.15.1 and `@nuxtjs/i18n` 9.1.1., 3.15.2 and 9.1.2 were being installed (check the installed dependencies in this failed run: https://github.com/WordPress/openverse/actions/runs/12894308773/job/35973352713?pr=5341). This combination somehow broke the fallback mechanism.

In this PR, I added a step to copy `pnpm-lock.yaml` to the Playwright container to ensure that the locked dependencies are installed.

### i18n static placeholders
#5261 has replaced `{openverse}` placeholders with `{'Openverse'}`, but failed making the corresponding changes to the pot creation script. So, instead of `###openverse###`, GlotPress displayed `{'Openverse'}`. The converted locale `json` files contained illegal characters. This was caught by eslint in PRs.

### fetchState props
#5323 left out the `canLoadMore` prop from `audio/collection.vue`, didn't update the `CollectionLink` story props, and left a stray console.log statement.

### App setup on server request
Locally, sometimes you can get a text stats response from the API instead of json, and this breaks the setup because stats are called on every SSR request. To handle that, I updated the plugin order:
- `init` plugin only sets up the app using available data (cookies)
- `init-stores` plugin is moved later in the initialization chain, so that the error handling and analytics plugins are already available.

This PR also makes the Playwright tests more robust by adding the color mode cookies updating the selectors for changing the color mode, and replacing the getter in the watcher with a computed (to make the change faster).

## Testing Instructions

The CI should pass.

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/src/components/VThemeSelect/VThemeSelect.vue
```diff
@@ -64,7 +64,7 @@ const updateRefs = () => {
 }
 
 onMounted(updateRefs)
-watch(() => [darkMode.colorMode.value, darkMode.osColorMode.value], updateRefs)
+watch([darkMode.colorMode, darkMode.osColorMode], () => updateRefs())
```
dhruvkb, line 67: How is `() => updateRefs()` different from `updateRefs`? Does it fix a problem or is it a stylistic/conventional thing?

frontend/test/playwright/playwright.config.ts
```diff
@@ -54,7 +54,7 @@ const config: PlaywrightTestConfig = {
     baseURL,
     trace: "retain-on-failure",
   },
-  timeout: 60 * 1e3,
+  timeout: 10 * 1e3, // 5 seconds in enough to see if the test is stuck
```
dhruvkb, line 57: Did you mean...?
```suggestion
  timeout: 5 * 1e3, // 5 seconds in enough to see if the test is stuck
```

frontend/test/proxy.js
```diff
@@ -189,9 +189,9 @@ const opts = /** @type {Partial<TalkbackOptions>} */ ({
   summary: false,
   tapeNameGenerator,
   tapeDecorator,
-  responseDecorator: (tape, req, context) => {
+  responseDecorator: (tape) => {
     // Log responses to make debugging easier
-    console.log(req.method, req.url, tape.res?.status, context.id)
+    // console.log(req.method, req.url, tape.res?.status, context.id)
```
dhruvkb, line 194: Nit: Why not remove the line?

dhruvkb (commented):
I'm still wrapping my head around the lockfile moving and copying but here are some small nits and questions so far.

frontend/src/components/VThemeSelect/VThemeSelect.vue
obulat: I'm not certain. I added this change trying to fix the cases where Playwright would select an option, but the selected option was still the same on the next line. I think the previous way has one more layer of indirection than what is here, and maybe the change happens in a different tick of the event loop so that Playwright sees the not-yet-changed version?

frontend/test/playwright/playwright.config.ts
obulat: Definitely! I tried 5 seconds, but the audio search page VR test needs more time.

frontend/Dockerfile.playwright
```diff
@@ -9,6 +9,9 @@ WORKDIR /frontend
 
 COPY package.json .
 
+# Copy the pnpm-lock.yaml from the root of the monorepo
+COPY ../pnpm-lock.yaml ./pnpm-lock.yaml
+
```
dhruvkb, line 14: Why does this copy `pnpm-lock.yaml` file from the parent directory when the `ci_cd.yml` file has copied it into `frontend/`? Also how is this Dockerfile able to access files from one level up? As per my understanding, it would not be part of the Docker context.

dhruvkb (commented):
I have one question about the lock-file movement.

frontend/Dockerfile.playwright
obulat: Updated to copy from frontend directory to docker.

frontend/test/proxy.js
obulat: Updated to only log if `verbose` is set to true. This could be very useful for debugging CI failures.

PR #4099, opened 2024-04-12
Add a redirect on 503 to frontend nginx

obulat:
## Fixes

Fixes #2996 by @obulat

## Description

This PR rewrites the error code when Plausible is unavailable (503) to a 424 error to allow us to count frontend 5xx errors separately from Plausible 5xx errors.

## Testing Instructions

I don't know how to test this.

I tried adding the following patch:
```
Subject: [PATCH] Add a redirect on 503 to frontend nginx
---
Index: frontend/nginx.conf.template
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/frontend/nginx.conf.template b/frontend/nginx.conf.template
--- a/frontend/nginx.conf.template	(revision 19cde7cdff2d8b785ebb4dee50966a37be5dd4eb)
+++ b/frontend/nginx.conf.template	(date 1712906826225)
@@ -73,7 +73,8 @@
     resolver $OPENVERSE_NGINX_DNS_RESOLVER;
     set $plausible_event_url $OPENVERSE_NGINX_PLAUSIBLE_EVENT_URL;
     location = /api/event {
-        proxy_pass $plausible_event_url;
+        return 503;
+        # proxy_pass $plausible_event_url;
         proxy_set_header Host plausible.io;
         proxy_buffering on;
         proxy_http_version 1.1;
```
And running `just frontend/up` and `just frontend/dev` (or `just frontend/build` & `just frontend/start`), but the event was still sent. I also updated the Plausible settings in the `nuxt.config` to use the local url instead of plausible.io, but could not mock a 503 error.
@sarayourfriend, do you know how we could test this?

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like`Update index.md`).
- [ ] My pull request targets the _default_ branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.
- [ ] I ran the DAG documentation generator (if applicable).

[best_practices]:
  https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

[...]

PR #5373, opened 2025-02-04
Replace Plausible 502 errors with 424

### LINKED ISSUE #3309: Transform 5xx responses from Plausible in frontend Nginx into 424s
## Problem

We proxy requests to Plausible from the `/api/event` endpoint on our frontend service. Our frontend Nginx image handles this. If the request to Plausible results in a 5xx response (e.g., if Plausible has degraded service or an outage) then this gets translated in our dashboards and monitoring to a 5xx response from our frontend service. However, our Plausible integration, while important for our understanding of how the application is used, is not critical to its core functionality, and therefore does not need to be treated as a "full unhandled breakage" the way a 5xx status code implies.

## Description

Rather than returning the 5xx status code from our proxied endpoint, we should transform it to [424 Failed Dependency](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/424), which accurately reflects the nature of the issue, and downgrades to a 424 status code, which clients are responsible for handling, rather than the service. Our frontend already handles _any_ status code from `/api/event` by ignoring all responses anyway, so downgrading these to 424 primarily benefits our infrastructure monitoring statistics by not creating false alarms of sudden increases in 5xx responses from our frontend service.

We can accomplish this with the `error_page` directive: http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page

We will also need to turn on `proxy_intercept_errors`: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_intercept_errors

Additionally, it'd be nice if we didn't completely swallow the response from Plausible, so if it's possible to create a new location that dynamically returns the upstream status code, that would be great. Maybe something like this (pseudo code!):

```
location ~= /424-plausible-(?<plausible_status_code>\d{3}) {
    default_type "application/json";
    return 424 '{"status":424, "plausible_status": $plausible_status_code}'
}

location = /api/event {
    // ...
    proxy_intercept_errors on;
    error_page 503 502 =424 /424-plausible-$upstream_status_code.json;
}
```

obulat:
## Fixes

Fixes #3309 by @sarayourfriend

## Description

This PR addresses the problem when we get an alarm for high number of 500 responses on the frontend. Most of those are 503 and 502 responses from Plausible. While the 503 have been handled (converted to 424 Failed dependency) by nginx, 502 still trigger alarms. 
This PR is based on the functionality from #4099, and adds 502 to the response codes that are converted to 424.

## Testing Instructions

Similar to #4099

## Checklist

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


frontend/nginx/nginx.conf.template
```diff
@@ -100,10 +100,10 @@ server {
         proxy_set_header X-Forwarded-Host  $host;
 
         proxy_intercept_errors on;
-        error_page 503 = @custom_424;  # Redirect 503 errors when Plausible is down
+        error_page 502 503 = @custom_424;  # Redirect 502 and 503 errors when Plausible is down
```
krysal, line 103: Should it include [504 errors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)? It's very similar to those two.

krysal (approved):
This will be greaaaaat if it gets rid of the 5XX alarms of the frontend due to Plausible!

frontend/nginx/nginx.conf.template
obulat: We only had 5 of 504s when there were thousands of others:
Screenshot 2025-02-05 at 7 02 27 AM
But I'll include them anyways, since they are in the same category.

By the way, this time, the problem with Plausible is reflected on the https://status.plausible.io/
Screenshot 2025-02-05 at 7 06 56 AM

Previously, when we had alarms, the status was green

PR #5585, opened 2026-06-03
Clarify API key registration fields

ekamran:
## Summary
- Clarifies that API key registration names must be unique across registered Openverse API applications
- Adds clearer schema help text for registration fields
- Fixes minor grammar in the registration endpoint documentation

Fixes #1803

Issue #4658, opened 2024-07-25
Convert usages of Pipenv to PDM after the ingestion server removal


obulat:
## Description

We have opted to use PDM as our package manager for the API and various other Python libraries in the monorepo. Most of usages of Pipenv were replaced in #4166. 

The remaining usages are necessary for the ingestion server. These will need to be converted after the [Removal of the ingestion server](https://github.com/WordPress/openverse/issues/3925) project is completed.

These items will be automatically removed during the implementation of the project.
- [ ] `ingestion_server/`
- [ ] `ingestion_server/Dockerfile`
- [ ] `ingestion_server/.dockerignore`
- [ ] `ingestion_server/justfile`

Pipenv should be removed from the following files after the project is completed:
- [ ] `.pre-commit-config.yaml`
- [ ] `.github/actions/setup-env/action.yml`

PR #5602, opened 2026-08-20
Pin CI tooling to restore build, lint, and image loading

ekamran:
Every open PR in this repository currently carries red CI for reasons unrelated to its changes. I traced them while trying to get #5585 reviewed. The common pattern is CI tooling installed unpinned at runtime, so quiet upstream releases changed CI behaviour under the repository.

This PR pins three of those runtime tools.

## 1. ingestion_server Docker build, failing since pipenv 2026.4.0

The Dockerfile installs whatever pipenv is newest. pipenv 2026.4.0 started normalizing package names when computing the Pipfile hash, so the existing Pipfile.lock no longer validates and `pipenv install --deploy` aborts.

I bisected the releases: 2026.2.2 is the last one that accepts the current lock, so this pins to it. Regenerating the lock would also work but touches every pinned dependency, and the long term fix is the planned move of this service off Pipenv in #4658.

Verified: the two-stage build fails on current main and passes with the pin.

## 2. Dockerfile lint job, failing since a recent hadolint release

The hadolint pre-commit hook pins the hook repository at v2.12.0 but leaves the docker image tag floating, so every run pulls the newest hadolint. A newer release added DL3066, an informational note that fires on six pre-existing Dockerfiles, and with no failure-threshold configured hadolint fails on any finding.

Two changes: the image is pinned to the same version as the hook rev so results are reproducible, and `failure-threshold` is set to `warning` so informational advice stays visible in the lint output without failing the job.

Verified with both the pinned and the latest image, and a warning-level violation still fails.

## 3. Load Docker images, failing since @actions/artifact v6

`.github/actions/load-img` installed `@actions/artifact` unpinned and then loaded it with `require()`.

`@actions/artifact` v6 is ESM-only and its exports map has no CommonJS entry, so the action failed with:

```text
ERR_PACKAGE_PATH_NOT_EXPORTED
```

This pins `@actions/artifact` to 5.0.3, the last release that still provides the CommonJS entry point used by the existing script.

Verified on this PR: the previous `Load Docker images` failure is gone, Docker image loading now proceeds, and `Run tests for ingestion-server` now passes.

## Current CI status

This PR still has failures, but they are no longer the three CI-tooling failures fixed here.

Code-level failures that remain:

- `Run tests for the API` fails because `audiowaveform` is missing. This one is not new: it also fails on the previous commit of this PR and on #5585, which only touches documentation.
- `Run Nuxt checks (frontend_init)` fails while Corepack downloads `pnpm-10.2.0` with `AssertionError [ERR_ASSERTION]: assert(!this.paused)`.
- `Run tests for the catalog` fails on missing S3 bucket `migrated-cccatalog-archives`.

The last two could not run at all before this PR, because their jobs stopped earlier at `Load Docker images`.

There are also maintainer-only label check failures.

So this PR does not make the full matrix green, but it removes three independent CI rot issues and gets the affected jobs past the blockers they previously could not pass.

### DIFF
--- a/.github/actions/load-img/action.yml
+++ b/.github/actions/load-img/action.yml
@@ -27,7 +27,9 @@ runs:
     - name: Install `@actions/artifact`
       shell: bash
       run: |
-        pnpm install @actions/artifact -w
+        # Pinned: v6 is ESM-only and its exports map has no CommonJS entry, so
+        # the `require()` in the script below fails with ERR_PACKAGE_PATH_NOT_EXPORTED.
+        pnpm install @actions/artifact@5.0.3 -w
 
     - name: Download images
       uses: actions/github-script@v7
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -49,8 +49,9 @@ runs:
     - name: Setup pnpm
       if: inputs.setup_nodejs == 'true'
       shell: bash
-      # Fix for https://github.com/nodejs/corepack/issues/612
-      run: npm install -g corepack@0.31.0 && corepack enable pnpm
+      # Pinned so pnpm downloads are verified with current npm registry
+      # signing keys (https://github.com/nodejs/corepack/issues/612).
+      run: npm install -g corepack@0.35.0 && corepack enable pnpm
 
     - name: Setup Node.js
       if: inputs.setup_nodejs == 'true'
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -8,7 +8,9 @@ ARG API_PY_VERSION
 # Python builder #
 ##################
 
-FROM docker.io/python:${API_PY_VERSION} AS builder
+# Pin the Debian release: the audiowaveform .deb below is built for
+# Debian 12 (bookworm) and silently fails to install on newer releases.
+FROM docker.io/python:${API_PY_VERSION}-bookworm AS builder
 
 # Get the version from the main justfile
 ARG OV_PDM_VERSION
@@ -45,7 +47,7 @@ RUN pdm install --check --frozen-lockfile $PDM_INSTALL_ARGS
 # API #
 #######
 
-FROM docker.io/python:${API_PY_VERSION}-slim AS api
+FROM docker.io/python:${API_PY_VERSION}-slim-bookworm AS api
 
 LABEL org.opencontainers.image.source="https://github.com/WordPress/openverse"
 
--- a/catalog/compose.yml
+++ b/catalog/compose.yml
@@ -67,7 +67,7 @@ services:
     profiles:
       - catalog_dependencies
       - catalog
-    image: docker.io/minio/minio:latest
+    image: docker.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
     ports:
       - "5010:5000"
       - "5011:5001"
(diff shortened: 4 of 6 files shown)


obenland:
Thanks for tracking these down @ekamran — this PR correctly identifies all three unpinned-tooling failures. A few notes from reviewing it:

**The hadolint `entry:` pin is the complete lint fix on its own.** I verified locally that hadolint v2.12.0 with the repo's existing `.hadolint.yaml` produces zero findings across all seven Dockerfiles (exit 0). The `failure-threshold: warning` addition is therefore unnecessary — and its comment is slightly misleading, since the version pin already prevents new rules from future releases appearing. I'd suggest dropping that hunk; if an info-level rule ever does need waiving, the repo's existing per-rule `ignored:` pattern keeps it visible.

**The lint fix is split out as #5605** so it can land immediately: because this PR touches `ingestion_server/Dockerfile`, it triggers the full API/catalog/Nuxt suites, which are currently failing for unrelated reasons and are required checks — the lint blockage affects every open PR (including several stuck Renovate security updates), so it shouldn't wait on those. Once #5605 merges, rebasing this PR will shrink it to the pipenv and `@actions/artifact` pins.

**On the `@actions/artifact@5.0.3` pin:** the reasoning (v6 is ESM-only) is correct. Longer term the dependency would be better declared in the root `package.json` so Renovate manages it and the lockfile pins it — right now it lives only in a shell string, the same unmanaged-drift class this PR fixes elsewhere. Fine as a follow-up, since `load-img` currently skips the install recipe and would need a small change to use it.

**Remaining red on this PR** (API tests, catalog tests, Nuxt `frontend_init` — the latter is the Corepack signature-verification breakage when downloading pnpm) is pre-existing and unrelated to these changes; it needs separate fixes before this can merge.

obenland:
Rebased onto `main` and extended this PR so the whole pipeline can go green (pushed with maintainer edit access — your two pins are unchanged, @ekamran):

- **Kept:** the pipenv pin and the `@actions/artifact@5.0.3` pin, as-is.
- **Dropped in the rebase:** the hadolint hunk — the image pin landed via #5605, and with hadolint back at v2.12.0 there are no findings to suppress, so the `failure-threshold` isn't needed.
- **Added three more pins for the failures that lint's breakage had been masking** (each root cause verified locally before committing):
  - `api/Dockerfile`: pin base images to `-bookworm`. `python:3.12-slim` now resolves to Debian 13, where the audiowaveform 1.10.1 .deb (built for Debian 12) fails to install — masked by the `|| apt-get --fix-broken install` fallback, so images build without the binary and the waveform tests fail. (Recent runs sometimes pass when the GHA build cache still holds pre-drift layers.)
  - `catalog/compose.yml` + `docker/minio/load_to_s3_entrypoint.sh`: pin `minio/minio` and `minio/mc` and switch the removed `mc config host add` to `mc alias set` — newer `mc` dropped the legacy command, so test buckets were never created (`NoSuchBucket` in the catalog tests).
  - `.github/actions/setup-env`: corepack `0.31.0` → `0.35.0` — 0.31.0 predates the npm registry's latest signing-key rotation, so its pnpm download fails signature verification (the Nuxt job's `ERR_ASSERTION`).

ekamran:
Thanks @obenland, and for splitting the lint half out so it could land on its own.

You were right about the failure-threshold hunk. With the image pinned to v2.12.0 there are no findings to suppress, so it was only adding noise.

The three extra pins are the useful part here. I had traced the failures lint was still reporting, not the ones its breakage was hiding.

I have rebased [#5585](https://github.com/WordPress/openverse/pull/5585) onto main, so the two temporary pins I was carrying there are gone and it is back to just the docs change.

